PR #296, opened 2024-02-02
[RFC] Design principles for testing

hardikjshah:
Scribbled some thoughts on how we should be setting up the tests

#### Context
- When i added a new test, i encountered some issues and Also spoke to Kartikay on some ideas. 

#### Changelog
- Suggestions on how we could manage our test setup 

#### Test plan
- N/A


RdoubleA:
Covered by #537

PR #362, opened 2024-02-09
Removing section on linking

kartikayk:
#### Context
- As per title

#### Changelog
- As per title

#### Test plan
image


RdoubleA:
@kartikayk I believe this has already been covered and is no longer applicable, I'll leave it to you to close this

PR #478, opened 2024-03-09
QLoRA

rohan-varma:
#### Context
- This PR enables e2e support for training with QLoRA technique, in which we implement ideas from https://arxiv.org/abs/2305.14314 specifically for single device LoRA finetuning use cases.

#### Changelog

I've broken down the set of changes into different components since this is quite a large change:

##### LoRA related changes
- I added 2 new ops via __torch__dispatch__ which represent a minimal set of changes needed to enable our usage of NF4Tensor as a parameter in `LoRALinear`. These are a `clone` implementation to make copies of the tensor (needed to register the Nf4tensor as parameter) as well as an inplace copy  method which is needed when loading state_dict.
- Inplace copy works by creating a reference nf4 tensor from the incoming bf16 tensor. Then, we set all attributes for the destination tensor to attributes for the bf16 tensor, and test that this indeed performs the proper copy.

##### LoRA related changes
- - Separate creating the base weight and bias into a helper function
- QLoRA does not support biases at the moment
- In forward, call regular F.linear or NF4's linear operator accordingly
- Add `quantize_base` and plumbing throughout the LoRA builders.

##### Checkpointing changes
- For load, we currently support taking in a bf16 checkpoint and quantizing it on the fly to NF4tensor.
- For loading checkpoint, `to_copy` dispatch implementation takes care of the above.
- For saving checkpoint, when saving full weights, and supporting LoRA weight merge, we implement inplace add and sub torch dispatch operators. This ensures that we use the bf16 restored weight when adding / subtracting the LoRA adapter deltas.
- We add a model-level state_dict post hook to reparametrize the model as bf16 and offload the bf16 parameters to CPU on the fly so as to not increase peak memory usage while checkpointing.

##### Caveats and follow ups needed
- The nf4 dispatch ops should live in AO and be thoroughly tested there. We'll upstream as part of follow up work
- Applying QLoRA to output proj results in higher memory usage, due to NF4tensor high reserved memory issue when working with very large tensors. So, not enabling quantize_base for output_proj of model yet, and explicitly passing in `quantize_base=False`
- Checkpoints saved are all in bf16. Need to investigate if there is any use in just saving the nf4 tensors themselves, maybe for efficiency reasons when loading checkpoint back in to torchtune for continued training.
- Docstrings and tutorials need to be updated appropriately. Will do this in follow up PRs.
- FP32 compute dtype support.

#### Test plan
Bunch of different testing to do so I've broken it down again.

##### Unittests
- LoRALinear tests are updated for quantize_base=True cases
- LoRALinear save/load are explicitly tested in QLoRA case
- LoRA builder tests are updated
- Model returned by `lora_llama2`  with quantize_base=True is unittested for save/load with expected dtypes, ensuring forwarad parity after save / load.

##### Checkpointing 
Save checkpoint with weight merging: `tune lora_finetune_single_device --config recipes/configs/alpaca_llama2_lora_finetune_single_device.yaml --override model_checkpoint=/home/rvarm1/local/dev/assets/llama2-bf16-latest seed=18 tokenizer_checkpoint=/home/rvarm1/local/dev/assets/tokenizer.model output_dir=/tmp/lora_debug device=cuda batch_size=2 enable_activation_checkpointing=True full_bf16=True quantize_base=True max_steps_per_epoch=1 &> out`

Results in `Model checkpoint of size 12852 MB saved to /tmp/lora_debug/model_0.ckpt`. 

Load that back in for continued training works: `tune lora_finetune_single_device --config recipes/configs/alpaca_llama2_lora_finetune_single_device.yaml --override model_checkpoint=/home/rvarm1/local/dev/assets/llama2-bf16-latest seed=18 tokenizer_checkpoint=/home/rvarm1/local/dev/assets/tokenizer.model output_dir=/tmp/lora_debug device=cuda batch_size=2 enable_activation_checkpointing=True full_bf16=True quantize_base=True max_steps_per_epoch=1 model_checkpoint=/tmp/lora_debug/model_0.ckpt`.

##### Memory analysis

- Baseline LoRA that FT's in ~16GB: `tune lora_finetune_single_device --config recipes/configs/alpaca_llama2_lora_finetune_single_device.yaml --override model_checkpoint=/home/rvarm1/local/dev/assets/llama2-bf16-latest seed=18 tokenizer_checkpoint=/home/rvarm1/local/dev/assets/tokenizer.model output_dir=/tmp/lora_debug device=cuda batch_size=2 enable_activation_checkpointing=True full_bf16=True`
- setting quantize_base in the config to True and same cmd, nvidia-smi reports ~9.8 GB peak RAM w/bf6 as our compute dtype. Will probably increase on T4's since fp32 needs to be our compute dtype. This is ~35% memory savings.
- Update: memory numbers with newly added memory logging - 

```
Memory Stats:, GPU peak memory allocation: 5.77175808 GB,
    GPU peak memory reserved: 9.294577664 GB,
    "GPU peak memory active: 5.77175808 GB"
```

##### Correctness comparison to LoRA - 

- Loss curves are on par (visually)
- LoRA bf16
image
- QLoRA - 
image

Eval results: 

- Trained a QLoRA model e2e: `tune lora_finetune_single_device --config recipes/configs/alpaca_llama2_lora_finetune_single_device.yaml dtype=bf16 device=cuda` with `quantize_base=True`. Ran eval with `tune eval --config torchtune/_cli/eval_configs/default_eval_config.yaml model_checkpoint=/home/rvarm1/local/dev/assets/qlora_trained.pt`, results: `truthfulqa_mc2: {'acc,none': 0.491684249068877, 'acc_stderr,none': 0.014723538004649834, 'alias': 'truthfulqa_mc2'}`
- Followed similar process for LoRA without quantization, but ran LoRA distributed recipe to get results faster. `tune --nproc_per_node 8 lora_finetune_distributed --config recipes/configs/alpaca_llama2_lora_finetune_distributed.yaml dtype=bf16 batch_size=32 enable_activation_checkpointing=True &> out_dist &`. eval result: `truthfulqa_mc2: {'acc,none': 0.46393356726865514, 'acc_stderr,none': 0.014448827209276354, 'alias': 'truthfulqa_mc2'}`
- These results are comparable to the lit-gpt QLoRA results https://lightning.ai/pages/community/lora-insights/#toc12 (see "LoRA Hyperparameter Tuning Part 3: Changing Alpha". highest result is for `r=256, alpha=32` with 0.5055 on truthfulqa_mc2), with batch_size=128 and micro batch size = 1
- Running with lit-gpt's best QLoRA config (r=256, a=32, effective batch size of 128 with micro batch size=1) - results in eval results of `truthfulqa_mc2: {'acc,none': 0.48860782692137417, 'acc_stderr,none': 0.014737964298503407, 'alias': 'truthfulqa_mc2'}`, compared to their 0.5055.


tests/torchtune/models/test_lora_llama2.py
```diff
@@ -171,3 +176,46 @@ def test_lora_llama2_state_dict_parity(
         )
         assert not unexpected
         assert all(["lora" in key for key in missing])
+
```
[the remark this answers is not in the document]
rohan-varma: Added

recipes/configs/lora_finetune_single_device.yaml
```diff
@@ -8,15 +8,16 @@
...
   _component_: torchtune.models.llama2.lora_llama2_7b
-  lora_attn_modules: ['q_proj', 'v_proj']
-  apply_lora_to_mlp: False
-  apply_lora_to_output: False
+  lora_attn_modules: ['q_proj', 'v_proj', 'k_proj', 'output_proj']
+  apply_lora_to_mlp: True
+  apply_lora_to_output: True
   lora_rank: 8
   lora_alpha: 16
+  quantize_base: True
```
ebsmothers, line 16: Wonder if this is a case where it makes sense to keep a single recipe and split the configs. E.g. `qlora_finetune_single_device.yaml`. This would also be nice cause QLoRA is one case where we do not (yet) support distributed, so by having only the single-device QLoRA config we make that more explicit. 

Separately, do we want to set `apply_lora_to_mlp` and `apply_lora_to_output` to True by default? No strong opinion on the former, but would consider setting the latter to False, just because it does use quite a bit more mem (even without QLoRA)

recipes/lora_finetune_single_device.py
```diff
@@ -96,6 +96,7 @@ def __init__(self, cfg: DictConfig) -> None:
         self.total_training_steps = 0
 
         self._resume_from_checkpoint = cfg.resume_from_checkpoint
+        self._gradient_accumulation_steps = cfg.gradient_accumulation_steps
```
ebsmothers, line 99: Thanks for adding this! Now only one recipe remains.. (though it's the one that actually needs this feature the least)

recipes/lora_finetune_single_device.py
```diff
@@ -361,6 +369,17 @@ def save_checkpoint(self, epoch: int) -> None:
             intermediate_checkpoint=(epoch + 1 < self.total_epochs),
         )
 
+    def _should_update_weights(self, current_iteration: int) -> bool:
+        """
+        Determines whether the weights should be updated on the current iteration or not.
+        True is returned either if we've accumulated gradients for enough steps or if this
+        is the last step in the epoch.
```
ebsmothers, line 376: nit: I don't think the second half applies here anymore. (I know it's the same in other recipes too)
```suggestion
        True is returned either if we've accumulated gradients for enough steps.
```

torchtune/modules/low_precision/_state_dict_hooks.py
```diff
@@ -0,0 +1,43 @@
...
+        state_dict (Dict[str, Any]): the state dict to modify
+        *args (Tuple[Any, ...]): Unused args passed when running this as a state_dict hook.
+        offload_to_cpu (bool): whether to offload the restored weight to CPU
+        **kwargs (Dict[Any, Any]): Unused keyword args passed when running this as a state_dict hook.
+    """
+    for k, v in state_dict.items():
+        if isinstance(v, NF4Tensor):
+            state_dict[k] = v.get_original_weight()
+            if offload_to_cpu:
+                state_dict[k] = state_dict[k].cpu()
```
ebsmothers, line 43: I think we discussed this but wanna make sure I understand. In the case that `offload_to_cpu=True`, why not do the CPU offload before the upcast? I guess the memory savings are minimal anyways because we do the offload one parameter at a time?

torchtune/modules/peft/lora.py
```diff
@@ -77,11 +83,40 @@ def initialize_parameters(self):
...
+        Creates a linear weight and bias tensor, using NF4 dtype if we're quantizing
+        (indicated via quantize_base=True).
+        """
+        in_dim, out_dim, use_bias = self.in_dim, self.out_dim, self.use_bias
+        linear = (
+            nn.Linear(in_features=in_dim, out_features=out_dim, bias=use_bias)
+            if not self._quantize_base
+            else FrozenNF4Linear(in_dim, out_dim, bias=False)
+        )
+        weight_tensor = linear.weight
```
ebsmothers, line 97: nit: just call it weight to line up with bias below

torchtune/modules/low_precision/_register_nf4_dispatch_ops.py
```diff
@@ -0,0 +1,66 @@
...
+
+@nf4_tensor_impl([torch.ops.aten.add_.Tensor])
+def add_bf16_tensor(func, *args, **kwargs):
+    """
+    Adds a bf16 tensor from an NF4Tensor, by add it from the
+    unquantized weight and re-casting to NF4.
+    """
+    nf4_tensor = args[0][0]
+    sub_tensor = args[0][1]
+    assert sub_tensor.dtype == torch.bfloat16
```
ebsmothers, line 65: add_tensor?

torchtune/modules/peft/lora.py
```diff
@@ -9,7 +9,11 @@
 import torch.nn.functional as F
 
 from torch import nn, Tensor
-
+from torchao.dtypes.nf4tensor import linear_nf4
+from torchtune.modules.low_precision import (  # noqa: F401
+    _register_nf4_dispatch_ops,
```
ebsmothers, line 14: Trying to understand how this works: so basically we import `_register_nf4_dispatch_ops` at the first import of lora.py, which will then ensure any NF4Tensor used henceforth in the same run will have all those overrides in the ops table? Assuming that is the correct understanding, is there a way to do this with less of a hammer? (I don't think it matters for this particular PR, just curious if we try to extend things down the line)

torchtune/modules/low_precision/_register_nf4_dispatch_ops.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+@nf4_tensor_impl([torch.ops.aten.copy_.default])
+def inplace_copy(func, *args, **kwargs):
+    """
+    Performs an inplace copy of an incoming tensor into the tensor
+    being copied into. The inplace tensor is given by args[0][1] and the
+    tensor being copied into is given by args[0][0]. The copy is performed
+    by copying over all attributes. This method would have to be updated
+    if additional attributes are added to NF4Tensor.
```
ebsmothers, line 29: Can we do an assert on e.g. the attributes of `ref_tensor` to address this?

torchtune/models/llama2/_lora_llama2_builders.py
```diff
@@ -34,6 +34,7 @@ def lora_llama2_7b(
     lora_rank: int = 8,
     lora_alpha: float = 16,
     max_batch_size: Optional[int] = None,
+    quantize_base: bool = False,
```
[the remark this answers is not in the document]
ebsmothers: My vote would be for yes. I don't see it in the latest version (unless I missed it), can we add it? Also thinking about how best to expose LoRA in our configs (see my [other comment](https://github.com/pytorch/torchtune/pull/478#discussion_r1535865690)), do you think it makes sense to expose e.g. `qlora_llama2_7b` there? Tbh I'm a bit torn cause I think it somewhat obscures how easy it is to switch LoRA <-> QLoRA with just the `quantize_base` flag. Either way, not a blocker for this PR.

torchtune/modules/low_precision/_register_nf4_dispatch_ops.py
```diff
@@ -0,0 +1,66 @@
...
+    unquantized weight and re-casting to NF4.
+    """
+    nf4_tensor = args[0][0]
+    sub_tensor = args[0][1]
+    assert sub_tensor.dtype == torch.bfloat16
+    return to_nf4(nf4_tensor.get_original_weight().sub_(sub_tensor))
+
+
+@nf4_tensor_impl([torch.ops.aten.add_.Tensor])
+def add_bf16_tensor(func, *args, **kwargs):
```
ebsmothers, line 58: This will apply for both weight merge and adding bias, right?

torchtune/models/llama2/_component_builders.py
```diff
@@ -225,76 +238,28 @@ def lora_llama2(
...
     tok_embeddings = nn.Embedding(vocab_size, embed_dim)
 
+    # TODO: quantize_base is not applied to final output_proj currently.
     output_proj = (
         LoRALinear(embed_dim, vocab_size, rank=lora_rank, alpha=lora_alpha)
         if apply_lora_to_output
         else nn.Linear(embed_dim, vocab_size, bias=False)
     )
-    return TransformerDecoder(
+    ret_model = TransformerDecoder(
```
ebsmothers, line 247: nit: just call `model`

torchtune/models/llama2/_component_builders.py
```diff
@@ -19,8 +21,12 @@
     TransformerDecoderLayer,
 )
 
-from torchtune.modules.peft import LoRALinear, LORA_ATTN_MODULES
+from torchtune.modules.low_precision import reparametrize_as_bf16_state_dict_post_hook
+
+from torchtune.modules.peft import LORA_ATTN_MODULES, LoRALinear
 
+# Modules from CausalSelfAttention that LoRA can be applied to
+LORA_ATTN_MODULES = Literal["q_proj", "k_proj", "v_proj", "output_proj"]
```
ebsmothers, line 29: import from peft_utils

ebsmothers, line 29: remove, you're already importing this

tests/torchtune/models/test_lora_llama2.py
```diff
@@ -171,3 +180,100 @@ def test_lora_llama2_state_dict_parity(
...
+            qlora = self.get_lora_llama2(
+                lora_modules=["q_proj", "v_proj", "k_proj", "output_proj"],
+                apply_lora_to_mlp=True,
+                apply_lora_to_output=False,
+                vocab_size=50,
+                quantize_base=True,
+                embed_dim=512,
+            )
+            qlora.load_state_dict(bf16_sd)
+            # LoRALinear weights should be nf4 still
```
ebsmothers, line 249: nit: base weights

tests/torchtune/models/test_lora_llama2.py
```diff
@@ -171,3 +180,100 @@ def test_lora_llama2_state_dict_parity(
...
+                vocab_size=50,
+                quantize_base=True,
+                embed_dim=512,
+            )
+
+        qlora_sd = qlora.state_dict()
+        # Ensure checkpoint merging produces bf16 tensors
+        merged_ckpt = get_merged_lora_ckpt(deepcopy(qlora_sd), rank=RANK, alpha=ALPHA)
+        for v in merged_ckpt.values():
+            # paranoid check for both, as NF4Tensor had issue where NF4Tensor.dtype would return bf16
```
ebsmothers, line 273: Is this resolved?

tests/torchtune/models/test_lora_llama2.py
```diff
@@ -171,3 +180,100 @@ def test_lora_llama2_state_dict_parity(
...
+
+    def test_lora_linear_quantize_base(self):
+        model = self.get_lora_llama2(
+            lora_modules=["q_proj", "v_proj", "k_proj", "output_proj"],
+            apply_lora_to_mlp=True,
+            # quantize_base
+            apply_lora_to_output=False,
+            vocab_size=50,
+            quantize_base=True,
+            embed_dim=512,
```
ebsmothers, line 192: Just curious: do you need a diff embed_dim for some particular reason?

ebsmothers (approved):
This is a really impressive PR. I left a bunch of comments but no huge concerns from my side. As always, make sure to clean up any debug code before landing. Otherwise looks good!

recipes/configs/lora_finetune_single_device.yaml
rohan-varma: set `apply_lora_to_output` to False for now. Although one thing to think about is whether our training config should be as close as possible to some sort of reference, the lit-gpt one does have this as true.

torchtune/modules/low_precision/_state_dict_hooks.py
```diff
@@ -0,0 +1,43 @@
...
+import torch.nn as nn
+from torchao.dtypes.nf4tensor import NF4Tensor
+
+
+def reparametrize_as_bf16_state_dict_post_hook(
+    model: nn.Module,
+    state_dict: Dict[str, Any],
+    *args: Tuple[Any, ...],
+    offload_to_cpu: bool = True,
+    **kwargs: Dict[Any, Any],
```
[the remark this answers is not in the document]
rohan-varma: So I'm not sure of the best way to design this sort of API. This API is meant to be a function that's passed into `_register_state_dict_hook`, and the function is invoked with some arguments that we don't use when taking state_dict in nn.Module. Since these are unused arguments but still provided to the function, I need to take them as unused with *args and **kwargs. Though, I want to offer `offload_to_cpu` as a keyword arg with a default, so I can't put it before *args. So this is what I was able to come up with, not sure if there's somethign better.

torchtune/modules/low_precision/_state_dict_hooks.py
rohan-varma: Not sure if `.cpu()` on NF4Tensor is supported OOTB (might need a dispatch op to copy from device to host). The memory saving is pretty minimal because we upcast / offload one at a time, as you've mentioned.

torchtune/modules/low_precision/_register_nf4_dispatch_ops.py
rohan-varma: Actually I think we can entirely get rid of the add and sub dispatch ops now that we're merging only bf16 weights (we upcast the nf4 tensor first with the state_dict post hook that I've added).

torchtune/modules/peft/lora.py
rohan-varma: Yep, when this is imported these ops get registered on the NF4Tensor. 

>  is there a way to do this with less of a hammer

Hmm, not sure what would be less of a hammer - I thought this was already pretty light in that we just handle importing something and the registration automatically happens, we don't have to call any specific `register_ops` API or anything like that. Did you have any example of what you'd like this to be?

torchtune/modules/low_precision/_register_nf4_dispatch_ops.py
rohan-varma: This might be a little tricky as I don't want this CPU overhead everytime this op is called. We could add a unittest that alerts us to check these attributes if nf4 tensor changes?

torchtune/modules/low_precision/_register_nf4_dispatch_ops.py
```diff
@@ -0,0 +1,66 @@
...
+@nf4_tensor_impl([torch.ops.aten.sub_.Tensor])
+def sub_bf16_tensor(func, *args, **kwargs):
+    """
+    Subtracts a bf16 tensor from an NF4Tensor, by subtracting it from the
+    unquantized weight and re-casting to NF4.
+    """
+    nf4_tensor = args[0][0]
+    sub_tensor = args[0][1]
+    assert sub_tensor.dtype == torch.bfloat16
+    return to_nf4(nf4_tensor.get_original_weight().sub_(sub_tensor))
```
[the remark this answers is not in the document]
rohan-varma: So I don't think we need add and sub anymore, at least for now, since when we're merging the lora_state_dict, the state_dict is already in full bf16.

torchtune/modules/low_precision/_register_nf4_dispatch_ops.py
rohan-varma: Probably not needed anymore - see other comment

torchtune/models/llama2/_component_builders.py
rohan-varma: Oops, might'be been bad merge

tests/torchtune/models/test_lora_llama2.py
rohan-varma: Yeah, the defaults for nf4 tensor is that the tensor dim needs to be divisible by the block size, which is 256. So just set 512 here (so the division isn't trivial and we actually end up with multiple blocks to quantize).

In the future, we can probably have some way where we override the NF4Tensor defaults / configure them to something more sensible for torchtune's workloads.

tests/torchtune/models/test_lora_llama2.py
rohan-varma: Not currently - 
image

recipes/lora_finetune_single_device.py
rohan-varma: Thanks for catching!

torchtune/modules/low_precision/_register_nf4_dispatch_ops.py
rohan-varma: Added unittest for now.

PR #527, opened 2024-03-19
Full finetune < 16GB

rohan-varma:
#### Context
- We'd like to enable a variant of full finetune that trains in < 16GB of RAM for users with consumer grade GPUs that have limited GPU RAM. This PR enables the full finetune to fit into that memory requirement.

#### Changelog
- Add support for running optimizer in backward. This is done by following this tutorial: https://pytorch.org/tutorials/intermediate/optimizer_step_in_backward_tutorial.html in `_setup_optimizer`.
- Add `OptimInBackwardWrapper` to checkpoint the optimizer states when running in backward + unittests.
- Update relevant portions of training code such as not calling `step`, `zero_grad` when using optimizer in backward.
- Disable optim state dict checkpointing for optimizer in backward, as this is nontrivial to implement and not supported yet.
- Warn when user is loading in an optimizer state checkpoint but optimizer in backward is enabled. In this case, the checkpoint is not loaded in.
- Change configuration to use PagedAdamW optimizer, when using this, memory is < 16GB:
> GPU peak memory allocation: 13.993540096GB, GPU peak memory reserved: 15.974006784GB, GPU peak memory active: 13.993540096GB

#### Test plan
- Run with bnb PagedAdamW + optim in backward + full bf16: `tune full_finetune_single_device --config recipes/configs/full_finetune_single_device.yaml &> out &` 

Memory does not spike above 16GB reserved - 
```
Memory Stats::
 3450     GPU peak memory allocation: 13.92 GB
 3451     GPU peak memory reserved: 15.97 GB
 3452     GPU peak memory active: 13.92 GB
```

Loss curves:
image

Eval result: 

```
truthfulqa_mc2: {'acc,none': 0.4372944697596623, 'acc_stderr,none': 0.014862338279863522, 'alias': 'truthfulqa_mc2'}
```

#### Checkpoint save/load of optimizer states
- Run and save recipe state: `tune full_finetune_single_device --config recipes/configs/full_finetune_single_device.yaml max_steps_per_epoch=1 device=cuda:1`
- Load recipe state, including optimizer state, back in for training: set `resume_from_checkpoint: True` and `recipe_checkpoint: /tmp/llama2/recipe_state.pt`, then run: `tune full_finetune_single_device --config recipes/configs/full_finetune_single_device.yaml device=cuda:1` --> training resume works

#### Discussion points
- Configuring bnb optimizers. Currently we are just able to do this in the config itself, user can just swap `torch.optim.AdamW` for `bitsandbytes.optim.PagedAdamW` for example. Can definitely have some documentation around this, but wondering if we need to do anything further for configuration.
- Currently, I'm mostly testing w/PagedAdamW but the 8bit adam from BNB works as well and can be similarly configured. PagedAdamW results in more memory savings (~14GB peak memory vs 27GB for 8 bit adam). 
- Checkpointing of optimizer in backward. For a fully functional recipe, checkpointing optimizer states for optimizer in backward is critical. We can do this with distributed's `_apply_optimizer_in_backward` API + `distributed_state_dict`, though the downsides are that the latter API is only avail. in more recent versions of torch, and the former API is private. OTOH, to implement checkpointing using the technique in https://pytorch.org/tutorials/intermediate/optimizer_step_in_backward_tutorial.html is not well understood from a technical perspective (hasn't been done before, and there are tricky things to figure out like how to key the optimizers by a UUID instead of a numerical index, which won't work when there are multiple optimizers). Using the former implementation is also useful since it is better tested and designed than building something ourselves.
- UX / exposing this memory efficient config. Currently I've just modified the config in place, but we can easily land a separate memory efficient config. We could even author an entirely separate recipe for users who don't want to enable optimizer in backward and deal with the UX complexity of using this.

#### Follow up work
- Documentation for using bitsandbytes optimizers, update documentation for memory efficiency in general


torchtune/utils/checkpoint.py
```diff
@@ -22,6 +22,59 @@
 )
 
 
+class OptimizerInBackwardWrapper:
```
kartikayk, line 25: I'd move this to _checkpointer/_checkpoint_utils.py since I plan to eventually nuke this file

recipes/configs/alpaca_llama2_full_finetune_single_device.yaml
```diff
@@ -32,8 +32,9 @@ resume_from_checkpoint: False
 batch_size: 2
 epochs: 3
 optimizer:
-  _component_: torch.optim.SGD
+  _component_: bitsandbytes.optim.PagedAdamW
   lr: 2e-5
+optimizer_in_bwd: True
```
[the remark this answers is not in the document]
rohan-varma: Nvm, we do support it now.

recipes/configs/llama2/7B_full_single_device.yaml
```diff
@@ -69,7 +70,7 @@ dtype: bf16
 
 # Logging
 metric_logger:
-  _component_: torchtune.utils.metric_logging.DiskLogger
-  log_dir: ${output_dir}
+  _component_: torchtune.utils.metric_logging.WandBLogger
+  project: foo
```
RdoubleA, line 74: if this is user facing, let's provide a real project name?

recipes/full_finetune_single_device.py
```diff
@@ -80,11 +80,18 @@ def __init__(self, cfg: DictConfig) -> None:
...
         self._output_dir = cfg.output_dir
         self._log_every_n_steps = cfg.log_every_n_steps if cfg.log_every_n_steps else 1
-        self._log_peak_memory_every_n_steps = 100
+        self._log_peak_memory_every_n_steps = 1
         # Training cfg
         self._resume_from_checkpoint = cfg.resume_from_checkpoint
         self._gradient_accumulation_steps = cfg.gradient_accumulation_steps
 
+        # TODO: find a better place / way to perform validation of args that don't yet
+        # compose with each other.
```
RdoubleA, line 89: after the dataclasses were removed we lost a convenient place to perform validation - my thinking is we provide a convenient utility function(s) that does this for users that they just import in their recipe, but curious about your thoughts

recipes/full_finetune_single_device.py
```diff
@@ -221,18 +229,55 @@ def _setup_model(
         return model
 
     def _setup_optimizer(
-        self, cfg_optimizer: DictConfig, opt_state_dict: Optional[Dict[str, Any]] = None
-    ) -> Optimizer:
+        self,
+        cfg_optimizer: DictConfig,
+        optimizer_in_bwd: bool = False,
+        opt_state_dict: Optional[Dict[str, Any]] = None,
+    ) -> Union[Optimizer, None]:
```
RdoubleA, line 236: `Optional[Optimizer]`?

recipes/full_finetune_single_device.py
```diff
@@ -221,18 +229,55 @@ def _setup_model(
...
         """
-        optimizer = config.instantiate(cfg_optimizer, self._model.parameters())
-
-        if opt_state_dict:
-            optimizer.load_state_dict(opt_state_dict)
-
-        log.info("Optimizer is initialized.")
-        return optimizer
+        if optimizer_in_bwd:
+            self._optimizer_in_bwd = True
```
RdoubleA, line 241: set this in init?

recipes/full_finetune_single_device.py
```diff
@@ -221,18 +229,55 @@ def _setup_model(
...
+            )
 
-        log.info("Optimizer is initialized.")
-        return optimizer
+            def optim_step(param) -> None:
+                optim_dict[param].step()
+                optim_dict[param].zero_grad()
+
+            for p in self._model.parameters():
+                p.register_post_accumulate_grad_hook(optim_step)
```
RdoubleA, line 257: I would say put the optimizer in backward logic in a separate utility, it uses some non-intuitive logic that may confuse users

recipes/full_finetune_single_device.py
```diff
@@ -221,18 +229,55 @@ def _setup_model(
...
+            # run, these need to have been saved with the same setting. Cannot restore from runs that did not
+            # use optimizer in backward.
+            if opt_state_dict is not None:
+                try:
+                    self._optim_ckpt_wrapper.load_state_dict(opt_state_dict)
+                except BaseException as e:
+                    raise RuntimeError(
+                        "Failed loading in-backward optimizer checkpoints."
+                        "Please make sure run being restored from was using in-backward optimizer."
+                        f"Original error {str(e)}"
```
RdoubleA, line 269: the `from e` should take care of surfacing the original error

recipes/full_finetune_single_device.py
```diff
@@ -281,13 +326,16 @@ def save_checkpoint(self, epoch: int) -> None:
...
                     utils.SEED_KEY: self.seed,
                     utils.EPOCHS_KEY: self.epochs_run,
                     utils.TOTAL_EPOCHS_KEY: self.total_epochs,
                     utils.MAX_STEPS_KEY: self.max_steps_per_epoch,
                 }
             )
+            if not self._optimizer_in_bwd:
+                ckpt_dict[utils.OPT_KEY] = self._optimizer.state_dict()
+            else:
+                ckpt_dict[utils.OPT_KEY] = self._optim_ckpt_wrapper.state_dict()
```
RdoubleA, line 338: if the APIs will be the same for the optim ckpt wrapper, you could just call it `self._optimizer` and remove the if else

recipes/full_finetune_single_device.py
```diff
@@ -344,20 +392,28 @@ def train(self) -> None:
                     self._metric_logger.log_dict(
                         {
                             "loss": loss.item(),
-                            "lr": self._optimizer.param_groups[0]["lr"],
+                            # NOTE: for optim in backward, this assumes all optimizers have the same LR. This is currently
+                            # true since we don't expose the ability to configure this yet.
+                            "lr": list(self._optim_ckpt_wrapper.optim_map.values())[
```
RdoubleA, line 397: can you make a method or property in the wrapper that retrieves this so this can be cleaner?

RdoubleA (commented):
If there are opportunities to make the optimizer wrapper have identical APIs to a normal optimizer, we can hide a lot of logic within the wrapper class and we won't need to check `if self._optimizer_in_bwd` in so many locations. I'm worried this will confuse new users esp since this is one of the main flagship recipes

recipes/full_finetune_single_device.py
rohan-varma: I actually tried this, and it didn't - i.e. I didn't see the exception from `e` in this error.

rohan-varma:
> If there are opportunities to make the optimizer wrapper have identical APIs to a normal optimizer, we can hide a lot of logic within the wrapper class and we won't need to check `if self._optimizer_in_bwd` in so many locations. I'm worried this will confuse new users esp since this is one of the main flagship recipes

So I'm actually reluctant to have the optimizer wrapper have identical APIs as a normal optimizer. This would mean it would need to implement a `.step()`, which would likely be a noop since the step actually happens in the backward pass. In general I'm a bit reluctant to add these sort of pass through abstractions that are just no-ops, as the user needs to click through the implementation to find it doesn't do anything (as expected). 

I do agree that all the if/else'ing here is not ideal.

recipes/full_finetune_single_device.py
rohan-varma: I'm personally not an advocate for this due to reason explained in the other comment. If folks feel like this is better UX though, I'm happy to just add it.

recipes/full_finetune_single_device.py
rohan-varma: I agree and see the reasoning in general, but IMO a downside of something like a `utils.setup_optim_in_backward(optim_config, model)` is that there is a side-effect of hooks getting registered on the model. I want things that modify state to be as explicit as possible, so I could do something like `register_optimizer_in_backward_hooks` and `make_optim_checkpoint_wrapper` - more utilities / components than a monolithic thing that configures the entire optimizer in backward. @kartikayk , @ebsmothers what do you think?

RdoubleA:
> In general I'm a bit reluctant to add these sort of pass through abstractions that are just no-ops, as the user needs to click through the implementation to find it doesn't do anything (as expected).

I see how this is odd but imo this is preferable to adding more if-else statements depending on certain parameters in the config. I've seen the runaway version of this where more features were added and the entire training recipe was bloated with these if-else checks that it was hard to even parse the core logic. Only advanced users may need to parse the logic of the optimizer in backward - I'd rather keep that more hidden so we don't risk muddying the core logic of a recipe that most beginner users will use as a starting point. Maybe there are alternative ways to do this by creating more modular methods?

rohan-varma:
> > In general I'm a bit reluctant to add these sort of pass through abstractions that are just no-ops, as the user needs to click through the implementation to find it doesn't do anything (as expected).
> 
> I see how this is odd but imo this is preferable to adding more if-else statements depending on certain parameters in the config. I've seen the runaway version of this where more features were added and the entire training recipe was bloated with these if-else checks that it was hard to even parse the core logic. Only advanced users may need to parse the logic of the optimizer in backward - I'd rather keep that more hidden so we don't risk muddying the core logic of a recipe that most beginner users will use as a starting point. Maybe there are alternative ways to do this by creating more modular methods?

This makes sense. I'd also like inputs from @kartikayk and @ebsmothers on this. One alternative I considered was writing an entirely new recipe, but decided not to go in this direction after discussing with @kartikayk offline. For now, I have an idea on how to separate this out into reusable components, so will put up something with that for folks to take a look at.

recipes/full_finetune_single_device.py
rohan-varma: Let me continue checking this though

torchtune/utils/_checkpointing/_checkpointer_utils.py
```diff
@@ -15,6 +15,101 @@ class ModelType(Enum):
...
+
+    Returns:
+        ``OptimizerInBackwardWrapper``: Wrapper for optimizer states running in backward.
+    """
+    return OptimizerInBackwardWrapper(
+        {n: optim_dict[p] for n, p in model.named_parameters()}
+    )
+
+
+def register_optim_in_bwd_hooks(
```
rohan-varma, line 92: Probably move this stuff to an optim_utils at this point

recipes/configs/llama2/7B_full_single_device_low_memory.yaml
```diff
@@ -0,0 +1,76 @@
...
+# To launch on a single device, run the following command from root:
+#   tune --nnodes 1 --nproc_per_node 1 full_finetune_single_device \
+#   --config llama2/7B_full_single_device \
+#
+# You can add specific overrides through the command line. For example
+# to override the checkpointer directory while launching training
+# you can run:
+#   tune --nnodes 1 --nproc_per_node 1 full_finetune_single_device \
+#   --config llama2/7B_full_single_device \
+#   checkpointer.checkpoint_dir=<YOUR_CHECKPOINT_DIR>
```
ebsmothers, line 19: missing `_low_memory` in these

recipes/full_finetune_single_device.py
```diff
@@ -221,18 +229,47 @@ def _setup_model(
...
-        optimizer = config.instantiate(cfg_optimizer, self._model.parameters())
-
-        if opt_state_dict:
-            optimizer.load_state_dict(opt_state_dict)
-
-        log.info("Optimizer is initialized.")
-        return optimizer
+        if optimizer_in_bwd:
+            # Maintain a dict of optims for every parameter.
+            # TODO (rohan-varma): check foreach arg
```
ebsmothers, line 242: I don't understand what this means in this context. Isn't foreach kind of a no-op in this case?

torchtune/utils/_checkpointing/_checkpointer_utils.py
ebsmothers: For now can also put in utils/_memory.py? If my interpretation of that file is correct I guess it includes memory-saving techniques given that AC wrapper is in there

torchtune/utils/_checkpointing/_checkpointer_utils.py
```diff
@@ -15,6 +15,101 @@ class ModelType(Enum):
...
+    model: torch.nn.Module, optim_dict: Dict[torch.nn.Parameter, torch.optim.Optimizer]
+) -> None:
+    """
+    Register hooks for optimizer step running in backward.
+
+    Args:
+        model (torch.nn.Module): Model whose parameters will be optimized. Note that currently
+            hooks for ALL parameters in the model will be registered.
```
ebsmothers, line 100: Implication being that this won't work for e.g. LoRA? Curious why we don't just gate on requires_grad when registering the post-accumulate-grad hooks, or is there more to it than that?

torchtune/utils/_checkpointing/_checkpointer_utils.py
```diff
@@ -15,6 +15,101 @@ class ModelType(Enum):
...
+    }
+    wrapper = OptimInBackwardWrapper(placeholder_optim_dict)
+    # load_state_dict expects a dict produced by this class's
+    # state_dict method.
+    wrapper.load_state_dict(torch.load("/tmp/optim_ckpt"))
+    # placeholder_optim_dict now has updated optimizer states.
+
+    NOTE: This wrapper is only meant to be used for single-device use cases.
+        Distributed use cases such as FSDP, which require specialized
+        optimizer state checkpointing, are not supported.
```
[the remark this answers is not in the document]
ebsmothers: Nice docstring btw. Do we plan to expose this as a public API? If so I think pydoclint would complain about missing explanation of `optim_map`

recipes/configs/llama2/7B_full_single_device_low_memory.yaml
```diff
@@ -0,0 +1,76 @@
...
+  recipe_checkpoint: null
+  output_dir: /tmp/llama2
+  model_type: LLAMA2
+resume_from_checkpoint: False
+
+# Fine-tuning arguments
+batch_size: 2
+epochs: 1
+optimizer:
+  _component_: bitsandbytes.optim.PagedAdamW
```
ebsmothers, line 53: bitsandbytes is not officially in our core requirements, right? What's our plan for handling things gracefully here?

recipes/full_finetune_single_device.py
```diff
@@ -80,11 +80,18 @@ def __init__(self, cfg: DictConfig) -> None:
         # logging attributes
         self._output_dir = cfg.output_dir
         self._log_every_n_steps = cfg.log_every_n_steps if cfg.log_every_n_steps else 1
-        self._log_peak_memory_every_n_steps = 100
+        self._log_peak_memory_every_n_steps = 1
```
ebsmothers, line 83: undo

recipes/full_finetune_single_device.py
ebsmothers: I'm inclined to agree with @RdoubleA. Definitely don't want to hide the actual behavior too much, but in this case we already have state_dict and other APIs defined, we might as well just cut down on branching (aside from where it's really needed, like in train). But honestly not taking a super strong stance here so fine either way

ebsmothers (commented):
Overall the changes look good and the memory reduction is great. Agree with your point in the summary, we should definitely put together a quick tutorial on some of the memory-saving tricks being applied here.

Two asks before stamping: (1) figure out a plan for bitsandbytes dep (right now we are not doing any checks anywhere as far as I can tell), and (2) tests 😃

recipes/full_finetune_single_device.py
rohan-varma: You're right! Foreach should not really matter here - might just run a quick test to confirm.

torchtune/utils/_checkpointing/_checkpointer_utils.py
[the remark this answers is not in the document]
rohan-varma: Yes, this won't work with FSDP optimizer checkpointing. The plan to checkpoint FSDP optimizer states in backward will be via `distributed_state_dict`.

recipes/configs/llama2/7B_full_single_device_low_memory.yaml
rohan-varma: Good call out, I think we'll simply add bitsandbytes to our core deps.

torchtune/utils/_checkpointing/_checkpointer_utils.py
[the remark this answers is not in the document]
rohan-varma: hmm this is a public API right now via `utils.OptimizerInBackwardWrapper`, not sure why pydoclint isn't making any issue. Will add docs to the rest of the functions as well.

torchtune/utils/_checkpointing/_checkpointer_utils.py
rohan-varma: Yeah can do that also. My vote for memory saving utils would be to put them in a directory for each components, i.e. utils.act_ckpt, utils.optim, etc. But happy to go with this approach as well.

torchtune/utils/_checkpointing/_checkpointer_utils.py
```diff
@@ -15,6 +15,101 @@ class ModelType(Enum):
     LLAMA2 = "llama2"
 
 
+class OptimizerInBackwardWrapper:
```
rohan-varma, line 18: cc @janeyx99

recipes/full_finetune_single_device.py
```diff
@@ -344,20 +384,30 @@ def train(self) -> None:
...
                             "loss": loss.item(),
-                            "lr": self._optimizer.param_groups[0]["lr"],
+                            # NOTE: for optim in backward, this assumes all optimizers have the same LR. This is currently
+                            # true since we don't expose the ability to configure this yet.
+                            "lr": (
+                                list(self._optim_ckpt_wrapper.optim_map.values())[
+                                    0
+                                ].param_groups[0]["lr"]
+                                if self._optimizer_in_bwd
+                                else self._optimizer.param_groups[0]["lr"]
```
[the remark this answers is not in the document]
rohan-varma: Added helper

recipes/configs/llama2/7B_full_single_device_low_memory.yaml
rohan-varma: Added bitsandbytes as a core dep

rohan-varma: Removed bnb as a core dep after discussion

recipes/full_finetune_single_device.py
ebsmothers: bump

recipes/full_finetune_single_device.py
ebsmothers: bump

tests/recipes/test_full_finetune_single_device.py
```diff
@@ -51,15 +51,16 @@ def _fetch_expected_loss_values(self):
         return [10.5074, 10.5563, 10.5152, 10.4851]
 
     @pytest.mark.integration_test
-    def test_loss(self, tmpdir, monkeypatch):
+    @pytest.mark.parametrize("config", ["full_single_device_low_memory"])
```
ebsmothers, line 54: Should still be running on the original recipe too, no?

ebsmothers (approved):
Looks great! One comment on the recipe test, otherwise just a couple nits

tests/recipes/test_full_finetune_single_device.py
rohan-varma: ah yes, good catch!

recipes/full_finetune_single_device.py
rohan-varma: This makes sense. I've added some documentation in the recipe class for now. @RdoubleA , let's chat about Evan's suggestion here?

recipes/full_finetune_single_device.py
rohan-varma: Makes sense - let's discuss in follow up PRs.

chauhang:
@rohan-varma Is there a way to achieve the same using torchao? BnB support on AMD is requires [complex setup](https://github.com/TimDettmers/bitsandbytes/issues/107#issuecomment-2016586255), compile from source did not work for me.

Issue #505, opened 2024-03-15
Explicitly install torch in Github workflows


ebsmothers:
We dropped torch from our requirements in #442. This means our CI _should_ need to install it explicitly now. But it's not. And nothing is failing. Why? I'm pretty sure because we conveniently took a torchao dependency just before that in #465. So we are now installing pytorch as a transitive dep for our CI, which we probably should not do.

Issue #341, opened 2024-02-08
Add nightly CI for pytorch nightly


rohan-varma:
- Would be good to support latest pytorch nightlies so we're not surprised when pytorch version updates. Should add some CI that runs w / pytorch nightly in addition to the pytorch existing CI.

joecummings:
Resolved by #663

PR #537, opened 2024-03-20
Clean up our testing

ebsmothers:
### Context
Our tests are a mess. This PR is an attempt to make things better.

**What would a good testing setup look like?**

There are a couple things we need to consider: (1) we have recipes that are not part of our core package and we still want to test them. (2) A lot of our recipes are distributed, and we need to test them on 2+ GPUs. We also need to set things up so that (a) we are reasonably confident that our test suite will catch breakages on bad PRs, and (b) we do not waste a ton of dev time (waiting for CI) and compute to accomplish that. 

So I propose the following delineation:

1) Unit tests. This is what's in `tests/torchtune` currently. Tbh I haven't done much with these in this PR, mainly because they're in decent (enough) shape already.
2) ~~Integration~~ Recipe tests. These should directly call the CLI. They should use the prod config, but *deliberately* override (a) the model (we should not use the full 7B) and (b) other configs as needed within the test class (e.g. setting `max_steps_per_epoch=2`). Recipe test files should maintain a 1:1 mapping with recipes.
3) Regression tests. In this PR it's still pretty minimal, but these are where we will actually run decently-sized fine-tunes and/or evals of full-size models and check the performance. These should only run on a nightly cadence, no need to block everyone's PRs with these heavyweight tests.

**We can now also handily mark and run each of these categories of tests** (as originally suggested in #296). Just add `@pytest.mark.integration_test` to your integration tests, and `@pytest.mark.slow_integration_test` to your regression tests. This means we can easily run e.g.

- Only unit tests: `pytest tests --without-integration --without-slow-integration`
- Only recipe tests: `pytest tests -m integration_test`
- Only regression tests: `pytest tests -m slow_integration_test`

(or really any other subset of these 3 you want). 

Other stuff: 
- (My fav topic) When testing a new component **DO NOT USE init_weights_with_constant**. Use `fixed_init_model` instead. Did you know that `A*x != x*A`? If you use `init_weights_with_constant` you will in fact not know that.
- There are a lot of grey areas between unit tests and recipe tests. As much as possible, please try to test logic in unit tests. The current setup is that unit tests are single-device only, so anything that may behave differently on multiple devices (e.g. stuff with FSDP `FULL_SHARD` shard strategy). That being said, there are still ways to mock basic distributed functionality in unit tests (see e.g. [single_box_init](https://github.com/pytorch/torchtune/blob/240cff850c637b5bd02b6ba80dc0957a6981bd55/tests/test_utils.py#L97) or uses of [get_pet_launch_config](https://github.com/pytorch/torchtune/blob/240cff850c637b5bd02b6ba80dc0957a6981bd55/tests/conftest.py#L15)).
- If you do not add a test for a new piece of functionality you're adding or (god forbid) you remove an existing test because "it's annoying/hard to understand" I will yell at you. If it's hard to understand please make it better, I promise on a case-by-case basis it is really not that hard.

### Changelog
- Set up workflow for distributed testing of recipes on 2x GPUs
- Separate out our recipe tests into single device and distributed to match the new structure of our recipes.
- Get our recipe tests to _actually_ run properly in a distributed setting. No more monkey patching, no more reading from stdout, no more `single_box_init`, etc. etc. There is probably more here but this is at least a minimal set of things we absolutely need to change.
- Add `gpu_test` decorator to define tests than should only run when >= n GPUs are available.
- Add back a bunch of test cases that got deleted. There are still areas missing coverage, but we are at least no worse here than we were previously.
- Say it with me: "No more separate test configs". These hurt my brain and are a maintenance nightmare. Just define overrides for each test class as needed via a list of strings. 
- Get rid of our "integration tests" altogether (along with corresponding workflow). This is just a parity check with HF. Tbh I have half a mind to delete our generation test as well, but haven't gotten there yet.
- Explicitly install pytorch in our workflow files. See #505 for more info on this. I know [this comment](https://github.com/pytorch/torchtune/issues/341#issuecomment-1963951352) mentions we should only test against nightlies, tbh I am still not 100% sold on it, but left it out of this first version for convenience more than anything.
- Remove 7B tests from `tests/recipes`. This allows us to scrap all the `--large-scale` args that we have littered everywhere (one exception is the checkpoint test, which can be done in a follow-up).
- Add two initial regression tests. The first just checks a few loss values for our llama2_7b full finetune with FSDP on 2 devices (similar to what we had previously in our single device test). The second runs LoRA FT on 7B for 500 steps, then 100 steps of eval on truthfulqa_mc2 and confirms that the accuracy is above 0.4. This is just a first pass, as we get more rigorous we can consider tightening the constraints and adding more here.
- Autodownload assets as needed. 
- Add tags for recipe tests and regression tests (discussed above).
- Combine the above two so that you can run e.g. `pytest tests --without-slow-integration` and pytest will download only small model files plus tokenizer (assuming they aren't already available locally). 

### Test plan

```
pytest tests
...
============================= 194 passed, 9 warnings in 676.25s (0:11:16) ========================
```

**Note:** Due to some challenges with the distributed CI I am gonna scrap the regression test github workflow for the time being. But the tests run and pass when run on our machines. We can make adding the CI job a fast follow.

CI is green (finally). And all I had to do was remove that pesky regression test job..

Screenshot 2024-03-22 at 5 15 13 PM


tests/assets/tiny_llama2_checkpoint.pt
joecummings, line 1: Why delete?

ebsmothers: Are we using it anywhere?

tests/regression_tests/test_llama2_7b.py
```diff
@@ -0,0 +1,119 @@
...
+        tune --nnodes 1 --nproc_per_node 2 lora_finetune_distributed
+            --config lora_finetune_distributed \
+            output_dir={tmpdir} \
+            checkpointer=torchtune.utils.FullModelTorchTuneCheckpointer
+            checkpointer.checkpoint_dir='{ckpt_dir}' \
+            checkpointer.checkpoint_files=[{ckpt_path}]\
+            checkpointer.output_dir={tmpdir} \
+            checkpointer.model_type=LLAMA2 \
+            tokenizer.path=/tmp/test-artifacts/tokenizer.model \
+            max_steps_per_epoch=10 \
```
ebsmothers, line 97: TODO: change before landing

tests/regression_tests/test_llama2_7b.py
```diff
@@ -0,0 +1,119 @@
...
+        runpy.run_path(TUNE_PATH, run_name="__main__")
+
+        eval_cmd = f"""
+        tune eval \
+            --config {EVAL_CONFIG_PATH} \
+            model_checkpoint={tmpdir}/torchtune_model_0.pt \
+            tokenizer._component_=torchtune.models.llama2.llama2_tokenizer \
+            tokenizer.path=/tmp/test-artifacts/tokenizer.model \
+            tasks=['truthfulqa_mc2']
+            limit=10 \
```
ebsmothers, line 110: TODO: change before landing

.github/workflows/regression_test.yaml
```diff
@@ -0,0 +1,57 @@
+name: Regression Tests
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+  workflow_dispatch:
```
ebsmothers, line 7: TODO: update before landing

.github/workflows/recipe_test_multi_gpu.yaml
```diff
@@ -49,23 +51,20 @@ jobs:
...
+          python -m pip install torch
           python -m pip install -r requirements.txt
           python -m pip install -r dev-requirements.txt
           python -m pip install -e .
-      - name: Run integration tests with coverage
-        env:
-          HUGGING_FACE_HUB_TOKEN: ${{ secrets.HUGGING_FACE_HUB_TOKEN }}
-        run: pytest integration_tests/ --cov=. --cov-report=xml --durations=5 -vv -s --llama2-path /tmp/test-artifacts/llama2-7b-01242024 --tokenizer-path /tmp/test-artifacts/tokenizer.model
+      - name: Run unit tests with coverage
+        run: pytest tests/recipes --cov=. --cov-report=xml --durations=20 -vv
```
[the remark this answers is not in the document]
ebsmothers: Yeah update: I have removed large-scale stuff from our recipe tests altogether and put into regression tests.

tests/assets/tiny_llama2_checkpoint.pt
joecummings: I mean we were using it in the recipe tests...

tests/recipes/conftest.py
```diff
@@ -0,0 +1,19 @@
...
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+import subprocess
+from pathlib import Path
+
+import torchtune
+
+root = Path(torchtune.__file__).parent.parent.absolute()
+CACHE_ARTIFACTS_SCRIPT_PATH = Path.joinpath(
```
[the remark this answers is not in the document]
rohan-varma: Curious, why? just so i iunderstand

tests/assets/tiny_llama2_checkpoint.pt
ebsmothers: We aren't currently on [main](https://github.com/search?q=repo%3Apytorch%2Ftorchtune%20tiny_&type=code) though, right?

tests/recipes/test_alpaca_generate.py
```diff
@@ -13,37 +13,31 @@
...
+    @pytest.mark.integration_test
+    def test_alpaca_generate(self, tmpdir, monkeypatch):
+        ckpt = "small_test_ckpt_tune"
+        model_checkpoint = CKPT_MODEL_PATHS[ckpt]
         cmd = f"""
         tune alpaca_generate
-            --config {_CONFIG_PATH} \
-            model=torchtune.models.{ckpt} \
-            model_checkpoint={self._fetch_ckpt_model_path(ckpt)} \
+            --config alpaca_generate \
```
rohan-varma, line 32: Not for this PR, but we should probably have this recipe hook into our config system + be generally better supported. Or, just rip it out entirely. cc @RdoubleA @joecummings

tests/recipes/test_full_finetune_single_device.py
```diff
@@ -0,0 +1,235 @@
...
+            "seed=9",
+            "epochs=1",
+            "max_steps_per_epoch=1",
+            "optimizer=torch.optim.AdamW",
+            "optimizer.lr=2e-5",
+            "log_every_n_steps=1",
+        ]
+
+    @pytest.mark.integration_test
+    @pytest.mark.parametrize("full_batch_size, micro_batch_size", [(2, 1), (4, 1)])
```
rohan-varma, line 171: What's the runtime of this test? Is there value on testing on multiple pairs of full, micro batch size?

tests/recipes/test_lora_finetune_distributed.py
```diff
@@ -0,0 +1,221 @@
...
+        lora_model.load_state_dict(base_model_sd, strict=False)
+        baseline_out = lora_model(inputs)
+
+        # Load merged final ckpt directly into llama2 and call fwd
+        with open(f"{tmpdir}/torchtune_model_1.pt", "rb") as f:
+            sd = torch.load(f, weights_only=True)
+        llama2_model.load_state_dict(sd)
+        merged_ckpt_out = llama2_model(inputs)
+
+        torch.testing.assert_close(baseline_out, merged_ckpt_out, rtol=1e-5, atol=1e-5)
```
rohan-varma, line 221: So overall I think the organization of these tests are currently fine (LoRA, full, and distributed in different recipes), but what do you think about the ability for these tests to scale? Wonder if we should eventually just generalize these to be more recipe agnostic than what they currently are.

tests/regression_tests/test_llama2_7b.py
```diff
@@ -0,0 +1,68 @@
...
+
+# TODO: remove this once we have eval configs exposed properly
+pkg_path = Path(torchtune.__file__).parent.absolute()
+EVAL_CONFIG_PATH = Path.joinpath(
+    pkg_path, "_cli", "eval_configs", "default_eval_config.yaml"
+)
+
+
+@gpu_test(gpu_count=2)
+class TestLoRA7BDistributedFinetuneEval:
```
rohan-varma, line 28: Nice! Any reason we do this only for LoRA?

rohan-varma (approved):
Super awesome, so excited to have our tests in a much better state! Mostly small comments and a few requiring discussion, thanks!

.github/workflows/conversion_script_test.yaml
```diff
@@ -62,6 +62,7 @@ jobs:
             aws s3 cp s3://pytorch-multimodal/llama2-7b/consolidated.00.pth /tmp/test-artifacts/llama2-7b-fair
       - name: Install dependencies
         run: |
+          python -m pip install torch
```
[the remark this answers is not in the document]
ebsmothers: Yeah this part I have not yet taken a stance on here. In practice, we can just add different versions to the matrix. In my experience running CI on nightlies means it will be red decently often. If we are OK with that I am happy to do it, but personally wanna discuss with folks a bit more on the pros and cons here.

.github/workflows/regression_test.yaml
```diff
@@ -0,0 +1,57 @@
...
+          role-to-assume: arn:aws:iam::308535385114:role/gha_workflow_torchtune_pytorch-multimodal
+          aws-region: us-east-1
+      - name: Install S3 CLI
+        run: |
+          python3 -m pip install awscli==1.32.6
+      - name: Install dependencies
+        run: |
+          python -m pip install torch
+          python -m pip install -r requirements.txt
+          python -m pip install -r dev-requirements.txt
```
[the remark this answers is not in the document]
ebsmothers: Honestly no, this is a good point. We can add a check for this, though not sure what the best way to do it is tbh. Might need a bit more research here

tests/cache_assets.sh
```diff
@@ -0,0 +1,82 @@
+#!/bin/bash
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
```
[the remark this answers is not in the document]
ebsmothers: Crap this should've been removed. Naming is hard and I couldn't make up my mind

tests/recipes/test_alpaca_generate.py
ebsmothers: My vote is for the latter 😃

tests/recipes/test_full_finetune_single_device.py
ebsmothers: Just ran this one and it actually only took about 5 seconds (probably because the model is small and we're on single device). Re multiple pairs it is kind of a holdover from the original version, and actually there was a point where the `(2, 1)` case was passing as a false positive (negative?) due to too high of a tolerance. But now that we've cleaned it up maybe we can remove it? Honestly given that the test is relatively quick I don't have a strong preference here either way.

tests/regression_tests/test_llama2_7b.py
ebsmothers: Ugh this is the one piece I have not ironed out yet. I had the full finetune test_loss in here on 7B but [took it out](https://github.com/pytorch/torchtune/pull/537/commits/b60e9d41eab0062a35e44a791ea8818453454227) just for testing. Basically all our regression tests are OOMing on the CI hardware, so I am trying to sort that out. Once it's figured out I will add the full finetune loss test back (but probably not an eval test, given that the eval results on many full finetunes are questionable anyways)

tests/recipes/test_lora_finetune_distributed.py
ebsmothers: Yeah I can see that being valuable down the line. The main question is how our recipes scale, depending on that it may be easier or harder to generalize our recipe tests. But in any scenario I would still like to maintain at least some version of a 1:1 mapping between recipes and recipe tests (just maybe the tests can eventually come from more of a factory pattern or something).

.github/workflows/regression_test.yaml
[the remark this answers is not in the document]
joecummings: Also, we shouldn't need to explicitly install the requirements. Instead we should use `pip install -e ".[dev]"`

.github/workflows/conversion_script_test.yaml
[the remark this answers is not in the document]
kartikayk: I guess it would depend on whether we gate on any nightly features? I think ultimately we won't have an option, but would love to see if we can hold off for as long as possible. The only caveat to this is that we also want to catch pushes to torch which break us.

.github/workflows/recipe_test.yaml
```diff
@@ -43,26 +43,14 @@ jobs:
...
-            python3 -m pip install awscli==1.32.6
-            mkdir -p /tmp/test-artifacts
-            aws s3 cp s3://pytorch-multimodal/llama2-7b/tokenizer.model /tmp/test-artifacts
-            aws s3 cp s3://pytorch-multimodal/small-ckpt-01242024 /tmp/test-artifacts
-            aws s3 cp s3://pytorch-multimodal/small-ckpt-hf-03082024.pt /tmp/test-artifacts
-            aws s3 cp s3://pytorch-multimodal/small-ckpt-tune-03082024.pt /tmp/test-artifacts
-            aws s3 cp s3://pytorch-multimodal/small-ckpt-meta-03082024.pt /tmp/test-artifacts
+      - name: Install S3 CLI
+        run: |
+          python3 -m pip install awscli==1.32.6
```
kartikayk, line 48: Curious if there's a reason to pin to this version? If so, maybe a comment to make sure no one inadvertently changes this?

.github/workflows/recipe_test.yaml
```diff
@@ -43,26 +43,14 @@ jobs:
...
       - name: Install dependencies
         run: |
-          python -m pip install -r requirements.txt
-          python -m pip install -r dev-requirements.txt
-          python -m pip install -e .
-      - name: Run unit tests with coverage
-        run: pytest tests/recipes --cov=. --cov-report=xml --durations=20 -vv
+          python -m pip install torch
```
kartikayk, line 51: this is an extreme nit, but I see a few ```python3 pip installs``` and some ```python pip installs```. Let's consolidate unless there's a reason not to?

tests/cache_artifacts.sh
```diff
@@ -0,0 +1,94 @@
...
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+# This script will handle caching of all artifacts needed for a given test run.
+# ./cache_artifacts.sh alone is a no-op
+# ./cache_artifacts --run-recipe-tests will fetch tokenizer and small model checkpoints.
+# ./cache_artifacts --run-regression-tests will fetch tokenizer and 7B model checkpoint.
+# ./cache_artifacts --run-recipe-tests --run-regression-tests will fetch all of the above.
+# In all cases, if the files already exist locally they will not be downloaded from S3.
```
kartikayk, line 13: YES!!!!

tests/test_utils.py
```diff
@@ -142,3 +152,38 @@ def captured_output() -> Generator[Tuple[TextIO, TextIO], None, None]:
...
+    Given an output directory containing metric logger .txt file,
+    parse the .txt and return a list of losses from each logged iteration.
+    """
+    with open(log_file_path, "r") as f:
+        logs = f.read()
+    losses = [float(x) for x in re.findall(r"loss:(\d+\.\d+)", logs)]
+    return losses
+
+
+def gen_log_file_name(tmpdir, suffix: Optional[str] = None) -> str:
```
kartikayk, line 178: I've come to appreciate how useful pathlib is for all of this. I think you can simplify this a lot of if you convert tmpdir into a ```Path``` object (I forgot if it's already a ```Path```). In that cass this begins:

```
filename = Path.joinpath(tmpdir, suffix).with_suffix('txt')
```

tests/recipes/test_full_finetune_single_device.py
kartikayk: If we feel confident, I'd be up for chipping away at the total running time by removing some tests

kartikayk (approved):
This is such an AWESOME PR! Thanks so much for fixing our tests. Conceptually this framework makes a lot of sense as well. Just a few nits. But this looks great!

.github/workflows/recipe_test.yaml
ebsmothers: Honestly I have no idea, I am just following along with what we had previously here. Can take a look though

tests/recipes/test_full_finetune_single_device.py
ebsmothers: Sounds good, will remove the (2, 1) case then.

tests/test_utils.py
ebsmothers: Sorry, actually gonna leave this as-is. The point here is to generate a unique and deterministic filename per test case here. So I am actually deliberately taking tmpdir and removing all the `/`s to do this. In that case I think Path actually makes things look a bit more complicated, since I am doing some pretty non-Pathy operations

tests/regression_tests/test_llama2_7b.py
ebsmothers: Leaving as-is as we are not running this as a nightly CI job yet anyways

tests/regression_tests/test_llama2_7b.py
ebsmothers: Same comment as [above](https://github.com/pytorch/torchtune/pull/537/files#r1536429488)

PR #538, opened 2024-03-20
Update doc build job to output to gh-pages

svekars:
Add a job to output to gh-pages branch.


NicolasHug (approved):
Thanks a lot @svekars . As discussed offline the only way to really test that is to merge the PR and see how it behaves on subsequent PRs, so let's try.

For anyone seeing this in case the job is red on the CI: ignore it, we'll fix it (but don't revert the PR please).

PR #549, opened 2024-03-21
Add eleuther_eval as recipe

joecummings:
#### Context

Adding EleutherAI Eval Harness as a recipe and removing it as a hard dependency. 

#### Changelog

* Add eval recipe interface
* Remove hard dependency on Eleuther
* Move eval to recipes/
* Update tests to account for additional functionality
* Update Github workflow tests to install lm-eval

#### Testing

1. CI

2. Local testing
```
(torchtune-2) [jrcummings@devvm050.nha0 ~/projects/torchtune (update-eleuther-eval)]$ pytest tests/recipes/test_eleuther_eval.py
========================================================================================== 2 passed, 1 warning in 26.51s ===========================================================================================
```

4. Speed & acc

Ours with `meta-llama/Llama-2-7b`:  1.41s, 0.39 acc, command: `tune eleuther_eval --config eleuther_eval tasks=["truthfulqa_mc2"]`
Eleuther Harness directly with `meta-llama/Llama-2-7b-hf`: 1.50s, 0.39 acc, command: `lm_eval --model hf     --model_args pretrained=meta-llama/Llama-2-7b-hf     --tasks truthfulqa_mc2     --device cuda:0     --batch_size 32`


recipes/eleuther_eval.py
```diff
@@ -0,0 +1,189 @@
...
+
+
+logger = utils.get_logger("DEBUG")
+
+try:
+    import lm_eval
+    from lm_eval.evaluator import evaluate
+    from lm_eval.models.huggingface import HFLM
+    from lm_eval.tasks import get_task_dict
+except ImportError:
```
joecummings, line 29: This catches if the user has an incorrect version installed or if they don't have any version installed.

recipes/eleuther_eval.py
```diff
@@ -0,0 +1,189 @@
...
+        device (torch.device): The device to use.
+        max_seq_length (int): The maximum sequence length to use.
+        batch_size (int): The batch size per GPU to use.
+    """
+
+    def __init__(
+        self,
+        model: TransformerDecoder,
+        tokenizer: Tokenizer,
+        *,
```
joecummings, line 52: KWARGS!

recipes/eleuther_eval.py
```diff
@@ -0,0 +1,189 @@
...
+        return self._tokenizer.encode(text=string, add_bos=False, add_eos=False)
+
+    def tok_decode(self, tokens, **kwargs):
+        return self._tokenizer.decode(tokens)
+
+    def _model_call(self, inps):
+        return self._model(inps)
+
+    def _model_generate(self, *args, **kwargs):
+        raise RuntimeError(
```
joecummings, line 98: Found this out the hard way. In a rough estimate, 85% of all tasks in Eleuther are **not** free generation so we have the majority of our bases covered. However, if people open a bunch of issues asking for this, we can add a generation method.

recipes/eleuther_eval.py
```diff
@@ -0,0 +1,189 @@
...
+        t1 = time.time()
+
+        model_eval_wrapper = _EvalWrapper(
+            self._model,
+            self._tokenizer,
+            device=self._device,
+            max_seq_length=self._cfg.max_seq_length,
+        )
+
+        # Task initialization API changed between v0.4.1 and 0.4.2
```
joecummings, line 161: Copied this from gpt-fast

recipes/eleuther_eval.py
ebsmothers: So this is basically our workaround so that (a) we can still run eleuther eval as a recipe and (b) we do not have to take every dep on god's green earth in our package?

joecummings: Oui

.github/workflows/recipe_integration_test.yaml
```diff
@@ -65,6 +65,7 @@ jobs:
         run: |
           python -m pip install -r requirements.txt
           python -m pip install -r dev-requirements.txt
+          python -m pip install lm-eval==0.4.*
```
ebsmothers, line 68: How did we decide on this particular version?

README.md
```diff
@@ -25,7 +25,8 @@ The library provides:
 - Native-PyTorch implementations of popular LLMs
 - Support for checkpoints in various formats, including checkpoints in HF format
 - Training recipes for popular fine-tuning techniques with reference benchmarks and comprehensive correctness checks
-- Integration with HuggingFace Datasets for training and EleutherAI's Eval Harness for evaluation
+- Evaluation of trained models with EleutherAI Eval Harness
```
ebsmothers, line 28: 😎

recipes/configs/llama2_eleuther_eval.yaml
```diff
@@ -0,0 +1,30 @@
+# Config for EleutherEvalRecipe in eluether_eval.py
```
ebsmothers, line 1:
```suggestion
# Config for EleutherEvalRecipe in eleuther_eval.py
```

recipes/configs/llama2_eleuther_eval.yaml
```diff
@@ -0,0 +1,30 @@
...
+#
+# To launch, run the following command from root torchtune directory:
+#    tune eleuther_eval --config llama2_eleuther_eval
```
ebsmothers, line 4: Might even add tasks arg here explicitly given that (a) it's the most commonly used, and (b) list + CLI is annoying

tests/recipes/test_eleuther_eval.py
```diff
@@ -0,0 +1,98 @@
...
+models.small_test_ckpt_tune = llama2_small_test_ckpt
+
+
+class TestEleutherEval:
+    tokenizer_pth = "/tmp/test-artifacts/tokenizer.model"
+    model_ckpt = "small_test_ckpt_tune"
+
+    import os
+
+    os.environ["CUDA_VISIBLE_DEVICES"] = ""
```
ebsmothers, line 34: Do not do this

tests/recipes/test_eleuther_eval.py
```diff
@@ -0,0 +1,98 @@
...
+
+import torchtune
+from tests.common import TUNE_PATH
+from tests.recipes.utils import fetch_ckpt_model_path, llama2_small_test_ckpt
+from torchtune import models
+
+pkg_path = Path(torchtune.__file__).parent.parent.absolute()
+EVAL_CONFIG_PATH = Path.joinpath(
+    pkg_path, "recipes", "configs", "llama2_eleuther_eval.yaml"
+)
```
ebsmothers, line 23: Why do we need this now? Just use the recipe name only, no?

tests/recipes/test_eleuther_eval.py
```diff
@@ -0,0 +1,98 @@
...
+
+models.small_test_ckpt_tune = llama2_small_test_ckpt
```
ebsmothers, line 25: Merge with testing PR changes and u will live a happy and fulfilling life

torchtune/__init__.py
```diff
@@ -12,13 +12,15 @@
...
     "lora_finetune_distributed.py",
+    "eleuther_eval.py",
 ]
 _CONFIG_LISTS = {
     "full_finetune_single_device.py": ["full_finetune_single_device.yaml"],
     "full_finetune_distributed.py": ["full_finetune_distributed.yaml"],
     "lora_finetune_single_device.py": ["lora_finetune_single_device.yaml"],
     "lora_finetune_distributed.py": ["lora_finetune_distributed.yaml"],
     "alpaca_generate.py": ["alpaca_generate.yaml"],
+    "eleuther_eval.py": ["llama2_eleuther_eval.yaml"],
```
ebsmothers, line 23: Given we scrapped all the prefixes in our FT recipes, maybe just `eleuther_eval.yaml`? Thoughts?

tests/recipes/test_eleuther_eval.py
```diff
@@ -0,0 +1,98 @@
...
+
+        monkeypatch.setattr(sys, "argv", cmd)
+        with pytest.raises(SystemExit):
+            runpy.run_path(TUNE_PATH, run_name="__main__")
+
+        log_out = caplog.messages[-1]
+        assert "'acc,none': 0.3" in log_out
+
+    @pytest.fixture
+    def hide_available_pkg(self, monkeypatch):
```
ebsmothers, line 63: nice trick

.github/workflows/recipe_integration_test.yaml
joecummings: It is used by lit-gpt and gpt-fast and is the most up-to-date release, plus there would be significant "hacks" to be put in place between 0.3 and 0.4 to make it work e.g. using BaseLM instead of HFLM.

tests/recipes/test_eleuther_eval.py
joecummings: Merge bad, rebase good.

tests/recipes/test_eleuther_eval.py
joecummings: TIHI

tests/recipes/test_eleuther_eval.py
joecummings: : (

recipes/configs/llama2_eleuther_eval.yaml
```diff
@@ -0,0 +1,30 @@
...
+
+# Model Arguments
+model:
+  _component_: torchtune.models.llama2.llama2_7b
+
+checkpointer:
+  _component_: torchtune.utils.FullModelTorchTuneCheckpointer
+  checkpoint_dir: /tmp/llama/
+  checkpoint_files: [finetuned_model.pt]
```
ebsmothers, line 13: Where is this coming from? Might be nice to align with our output checkpoint file format so that it works out of the box

recipes/eleuther_eval.py
```diff
@@ -0,0 +1,205 @@
...
+
+    @property
+    def batch_size(self):
+        return self._batch_size
+
+    @property
+    def device(self):
+        return self._device
+
+    def tok_encode(self, string: str, **kwargs):
```
ebsmothers, line 83: This choice of param name hurts me

recipes/eleuther_eval.py
```diff
@@ -0,0 +1,205 @@
...
+        # see https://github.com/Lightning-AI/lit-gpt/blob/main/eval/lm_eval_harness.py#L66,
+        # though notably fast-gpt does the opposite
+        # https://github.com/pytorch-labs/gpt-fast/blob/main/eval.py#L123.
+        return self._tokenizer.encode(text=string, add_bos=False, add_eos=False)
+
+    def tok_decode(self, tokens, **kwargs):
+        return self._tokenizer.decode(tokens)
+
+    def _model_call(self, inps):
+        return self._model(inps)
```
ebsmothers, line 95: Add some type hints where relevant (also do we not run type checker on recipes or something?)

recipes/eleuther_eval.py
```diff
@@ -0,0 +1,205 @@
...
+
+    def _model_generate(self, *args, **kwargs):
+        raise RuntimeError(
+            "This recipe does not currently support tasks that evaluate free generation,"
+            "e.g. `truthfulqa_gen` or `bigbench_color_generate_until`."
+        )
+
+
+_DEFAULT_TASKS = ["hellaswag"]
```
ebsmothers, line 104: Tbh I am wondering if we should even have this. Like yes it's convenient to not have to write it out but it's already caused us some confusion. Like in what case is a user just gonna say "yolo I'll just run eval without even thinking about the task"

recipes/eleuther_eval.py
```diff
@@ -0,0 +1,205 @@
...
+    def load_checkpoint(self, cfg: DictConfig) -> Dict[str, Any]:
+        checkpointer = config.instantiate(cfg)
+        checkpoint_dict = checkpointer.load_checkpoint()
+        return checkpoint_dict
+
+    def setup(self) -> None:
+        self._device = utils.get_device(device=self._cfg.device)
+        self._dtype = utils.get_dtype(dtype=self._cfg.dtype)
+        self._limit = self._cfg.limit
+        self._tasks = list(self._cfg.tasks or _DEFAULT_TASKS)
```
RdoubleA, line 138: can we just add the hellaswag task to the config? agree with Evan that users should be aware of what task they are running, that's pretty fundamental to eval and it will also show users how to configure the task through our config.

recipes/eleuther_eval.py
```diff
@@ -0,0 +1,205 @@
...
+        - llama2_eleuther_eval.yaml
+
+    Args:
+        cfg (DictConfig): OmegaConf object parsed from YAML file
+    """
+
+    def __init__(self, cfg: DictConfig) -> None:
+        self._cfg = cfg
+
+    def load_checkpoint(self, cfg: DictConfig) -> Dict[str, Any]:
```
RdoubleA, line 129: nit: name this cfg_checkpointer so it's not confused with the entire config (it confused me)

tests/recipes/test_eleuther_eval.py
```diff
@@ -0,0 +1,98 @@
...
+
+class TestEleutherEval:
+    tokenizer_pth = "/tmp/test-artifacts/tokenizer.model"
+    model_ckpt = "small_test_ckpt_tune"
+
+    import os
+
+    os.environ["CUDA_VISIBLE_DEVICES"] = ""
+
+    def test_torchune_checkpoint_eval_result(self, caplog, monkeypatch, tmpdir):
```
RdoubleA, line 36: bonk
```suggestion
    def test_torchtune_checkpoint_eval_result(self, caplog, monkeypatch, tmpdir):
```

torchtune/recipe_interfaces.py
```diff
@@ -58,3 +58,36 @@ def cleanup(self, **kwargs) -> None:
...
+
+
+class EvalRecipeInterface(Protocol):
+    """
+    This class provides a loose structure which every LLM evaluation recipe
+    should follow. Please note that the interface itself should not be a vehicle for
+    code reuse. TorchTune strictly prohibits implementation inheritance in the codebase.
+    """
+
+    def load_checkpoint(self, **kwargs) -> None:
```
RdoubleA, line 70: why do you need an explicit `load_checkpoint` if this is just going to call `checkpointer.load_checkpoint()` in most cases?

recipes/configs/llama2_eleuther_eval.yaml
joecummings: It depends on the epoch, so not sure what the nice solution is here.

recipes/eleuther_eval.py
joecummings: copy pasta from gpt-fast

torchtune/recipe_interfaces.py
joecummings: Following what I see in other recipes, good to be consistent. Do you think it's overkill?

recipes/configs/eleuther_eval.yaml
```diff
@@ -0,0 +1,30 @@
...
+#    tune eleuther_eval --config llama2_eleuther_eval tasks=["truthfulqa_mc2", "hellaswag"]
+
+# Model Arguments
+model:
+  _component_: torchtune.models.llama2.llama2_7b
+
+checkpointer:
+  _component_: torchtune.utils.FullModelTorchTuneCheckpointer
+  checkpoint_dir: /tmp/llama/
+  checkpoint_files: [finetuned_model.pt]
```
kartikayk, line 13: Actually the TorchTune checkpointer outputs a model with a specific name format. Maybe we should update the config to that name so folks can use OOTB

recipes/eleuther_eval.py
```diff
@@ -0,0 +1,202 @@
...
+    from lm_eval.tasks import get_task_dict
+except ImportError:
+    logger.error(
+        "Recipe requires EleutherAI Eval Harness v0.4. Please install with `pip install lm_eval==0.4.*`"
+    )
+    sys.exit(1)
+
+
+class _EvalWrapper(HFLM):
+    """An EvalWrapper for EleutherAI's eval harness based on fast-gpt's
```
kartikayk, line 37:
```suggestion
    """An EvalWrapper for EleutherAI's eval harness based on gpt-fast's
```

recipes/configs/eleuther_eval.yaml
joecummings: I commented this somewhere else, but the specific name format requires an epoch number so there's no way to know exactly what the checkpoint file will be. I'm fine changing this name to *resemble* the outputted file, but it will not match OOTB.

recipes/eleuther_eval.py
```diff
@@ -0,0 +1,202 @@
...
+        batch_size: int = 32,
+    ):
+        super().__init__(device=str(device))
+        self._model = model
+        self._tokenizer = tokenizer
+        self._max_seq_length = max_seq_length
+        self._batch_size = batch_size
+
+    @property
+    def eot_token_id(self):
```
kartikayk, line 64: I think you mean this, but not fully sure
```suggestion
    def eos_token_id(self):
```

recipes/eleuther_eval.py
```diff
@@ -0,0 +1,202 @@
...
+    def device(self):
+        return self._device
+
+    def tok_encode(self, text: str, **kwargs) -> List[int]:
+        # Note on add_bos flag: setting to False as this gives better results, for example
+        # +1% on truthfulqa_mc2 with a LoRA finetune. lit-gpt also sets this to False,
+        # see https://github.com/Lightning-AI/lit-gpt/blob/main/eval/lm_eval_harness.py#L66,
+        # though notably fast-gpt does the opposite
+        # https://github.com/pytorch-labs/gpt-fast/blob/main/eval.py#L123.
+        return self._tokenizer.encode(text=text, add_bos=False, add_eos=False)
```
kartikayk, line 89: Very noob question, do all of these evals expect just tokenization? Do we need to make sure we align this with the special tokens used during training? Or how does that work? Sorry atm I'm a bit confused about the use of special tokens. cc: @RdoubleA

recipes/eleuther_eval.py
kartikayk: What's the reason to fail on this?

recipes/eleuther_eval.py
joecummings: Nope! This is an lm-eval specific function name we have to match exactly.

torchtune/recipe_interfaces.py
kartikayk: The recipe ```load_checkpoint``` contains any logic that is needed after the state_dict is returned from the checkpoint. So for example if you want to extract the adapter weights separately and do stuff with it.

recipes/eleuther_eval.py
joecummings: A good question and one that looks like is still being discussed on Eleuther's Github: https://github.com/EleutherAI/lm-evaluation-harness/issues/1017. They typically handle it all through HF's tokenization, which does special token processing under the hood.

If we use special tokens during training, we will need to update our tokenizer with those special tokens. As we're not doing that at all, I think we tackle that as part of the changes required to do that? Otherwise, this PR will have to implement those changes for training, as well.

kartikayk: Yeh definitely not something we should address here, but something we should call out (if needed), esp if this means our eval results will be sub-optimal

recipes/eleuther_eval.py
ebsmothers: I know, I think it's also coming from lm_eval tbh

tests/recipes/test_eleuther_eval.py
RdoubleA: wait yall don't use rebase?

recipes/eleuther_eval.py
kartikayk: The second one - got you!

recipes/eleuther_eval.py
RdoubleA: all special tokens should be coupled with the tokenizer, and ideally we need to keep those special tokens that were used for training when we do eval. This should be the same tokenizer object that is used in evals, so as long as we're calling the tokenizer the same way as we do in training we should be good? I don't fully understand why people are removing these, because it gave a slight boost to performance?

recipes/eleuther_eval.py
```diff
@@ -0,0 +1,202 @@
...
+    """This recipe runs evaluation on a trained model using EleutherAI's eval harness.
+
+    This recipe supports:
+        - Single GPU evaluation
+        - Loading model in fp32 or bf16
+        - Any task from the EleutherAI eval harness that is not free generation
+
+    Assumptions:
+        - Evaluation is launched with the Tune CLI (recommended)
+        - User has the EleutherAI eval harness installed, see https://github.com/EleutherAI/lm-evaluation-harness
```
ebsmothers, line 114: nit: I would move this point up to the top of the docstring (right after the first sentence). Just say: "To run this recipe, make sure you have installed ..."

recipes/eleuther_eval.py
ebsmothers: Can we give a bit more detail here? And maybe an explicit type of exception?

torchtune/recipe_interfaces.py
```diff
@@ -58,3 +58,36 @@ def cleanup(self, **kwargs) -> None:
...
+        """
+        ...
+
+    def evaluate(self, **kwargs) -> None:
+        """
+        All of the evaluation logic, including reporting.
+        """
+        ...
+
+    def cleanup(self, **kwargs) -> None:
```
ebsmothers, line 89: So we define this in the interface but don't actually use it? Makes me wonder if we need it in the interface at all

ebsmothers (approved):

torchtune/recipe_interfaces.py
joecummings: Removed

recipes/eleuther_eval.py
joecummings: Added a comment.

PR #563, opened 2024-03-22
Add .nojekyll, versions.html, symlink

svekars:


stable
```diff
@@ -0,0 +1 @@
+main
```
NicolasHug, line 1: we'll add the `stable` symlink when the time comes but for now it shouldn't point to main - do you mind removing this?

svekars: I can remove but the index.html won't work then.

index.html
```diff
@@ -0,0 +1,5 @@
+<html>
+  <head>
+    <meta http-equiv="refresh" content="0; url=/torchtune/stable/index.html">
```
NicolasHug, line 3: Re https://github.com/pytorch/torchtune/pull/563/files#r1536130428, would that work? (with a TODO comment saying we'll need to update it when stable is available?)

```suggestion
    <meta http-equiv="refresh" content="0; url=/torchtune/main/index.html">
```

PR #566, opened 2024-03-22
Distributed Recipe Cleanups

kartikayk:
#### Context

Our distributed recipes (especially LoRA) see some skew in the memory stats during training. This is mainly due to stragglers which need to be cleaned up before training begins. Adding this here.

Comparisons with LoRA distributed Recipe:

```
tune --nnodes 1 --nproc_per_node 4 recipes/lora_finetune_distributed.py \
--config recipes/configs/lora_finetune_distributed.yaml \
epochs=2 enable_activation_checkpointing=False

tune --nnodes 1 --nproc_per_node 4 recipes/lora_finetune_distributed.py \
--config recipes/configs/lora_finetune_distributed.yaml \
epochs=2 enable_activation_checkpointing=True
```

Baseline without AC

Screenshot 2024-03-22 at 1 15 30 PM

Baseline with AC
Screenshot 2024-03-22 at 1 13 55 PM

This PR without AC

Screenshot 2024-03-22 at 1 07 29 PM

This PR with AC

Screenshot 2024-03-22 at 1 10 34 PM

#### Changelog
-  Add the cleanup utility
- Sync before training begins
- Move validation for LoRA state dict to before loading the state dict (unrelated change, but putting in here)

#### Test plan
- Unit Tests

```
pytest tests/torchtune
```

- Recipe Tests

```
pytest tests/recipes
```

### DIFF
--- a/torchtune/utils/memory.py
+++ b/torchtune/utils/memory.py
@@ -4,6 +4,8 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
+import gc
+
 from typing import Optional, Set
 
 import torch
@@ -27,6 +29,12 @@ def set_activation_checkpointing(
     """
     wrap_policy = ModuleWrapPolicy(auto_wrap_policy or set())
     apply_activation_checkpointing(model, auto_wrap_policy=wrap_policy, **kwargs)
+
+
+def cleanup_before_training() -> None:
+    gc.collect()
+    torch.cuda.empty_cache()
+    torch.cuda.reset_peak_memory_stats()
 
 
 def memory_stats_log(
--- a/recipes/full_finetune_distributed.py
+++ b/recipes/full_finetune_distributed.py
@@ -4,7 +4,6 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
-import gc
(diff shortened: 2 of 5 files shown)


torchtune/utils/memory.py
```diff
@@ -29,6 +31,12 @@ def set_activation_checkpointing(
     apply_activation_checkpointing(model, auto_wrap_policy=wrap_policy, **kwargs)
 
 
+def cleanup_before_training() -> None:
+    gc.collect()
+    torch.cuda.empty_cache()
+    torch.cuda.reset_peak_memory_stats()
```
rohan-varma, line 37: This makes sense, though do note now that during training, we might not get an accurate picture from the training logs on what is our true peak memory  across the entire job. I think this is fine, because if peak memory reserved occurs during model init, we'll still log that at the beginning, and then proceed to log memory peaks observed during training (that might be less).

rohan-varma (approved):
Makes sense overall. One thing I did not follow was the use of "stragglers", in my world those are normally ranks that are taking too long to join the collective. Could you explain what the specific stragglers that need to be cleaned up are in this context?

kartikayk:
Ah might just be unfortunate terminology on my part but I'm referring to stuff that we're asking GC to explicitly collect.

PR #569, opened 2024-03-23
[RFC] Default pytest run mode to unit tests only

ebsmothers:
It was pointed out by @kartikayk that the current UX for our testing is a bit clunky. E.g. executing a single unit test without downloading from S3 necessitates running `pytest tests/path/to/my_unit_test.py --without-integration --without-slow-integration`. This is a bit annoying, especially given this is a very common way to test individual changes.

This PR makes the default mode for our test runs to be unit-test only. This means that `pytest tests` will now run **only** unit tests, and will not run recipe tests or regression tests. It will also not download any artifacts for those tests.

My hope is that this is more intuitive given that most people are only running unit tests. One gotcha is that e.g. `pytest tests/recipes` and `pytest tests/regression_tests` will now be essentially a no-op, so we will need to instead specify e.g. `pytest tests -m integration_test` or `pytest tests -m slow_integration_test` instead. Assuming this isn't too much cognitive load, I claim this change should be a net positive.

### Test plan

Run a bunch of different commands and check (a) which assets (if any) are expected, and (b) which tests runs.

**All unit tests**
Should run only unit tests, so no artifacts are needed.
```
pytest tests
(no expected artifacts)
...
187 passed, 19 skipped, 5 warnings in 136.89s (0:02:16)
```

**tests/recipes unit tests only**
Runs all unit tests under `tests/recipes`. Since only unit tests are run, no artifacts are needed.
```
pytest tests/recipes
(no expected artifacts)
...
1 passed, 11 skipped, 1 warning in 2.11s (since recipe tests are skipped by default)
```

**tests/recipes, recipe tests only**
Runs all recipe tests under `tests/recipes`. Fetches tokenizer and small model checkpoints for recipe tests.
```
pytest tests/recipes -m integration_test
small-ckpt-01242024
small-ckpt-tune-03082024.pt
small-ckpt-meta-03082024.pt
small-ckpt-hf-03082024.pt
...
11 passed, 1 deselected, 3 warnings in 125.10s (0:02:05)
```

**Single file, unit tests only**
Runs unit tests from a single file. No artifacts are needed.
```
pytest tests/torchtune/modules/peft/test_peft_utils.py 
(no expected artifacts)
...
19 passed, 1 warning in 0.12s
```

**Single (unit test) file, regression tests only**
(Not an intended usage) This command runs only regression tests from a single file containing only unit tests.
As a result, the full 7B model and tokenizer are fetched, but no tests actually run.
```
pytest tests/torchtune/modules/peft/test_peft_utils.py -m slow_integration_test 
Expected artifacts for this test run are:
llama2-7b-torchtune.pt
tokenizer.model
...
19 deselected, 1 warning in 0.02s (none run because the file contains no regression tests)
```

**Single (unit test) file, with regression tests**
(Also not an intended usage) By using `--with-slow-integration-test` instead of `-m slow-integration-test`,
this test will run unit tests + regression tests instead of just regression tests (as in the last command). 
Again the full 7B model and tokenizer will be downloaded, but in this case all the unit tests do run.
```
pytest tests/torchtune/modules/peft/test_peft_utils.py --with-slow-integration-test 
Expected artifacts for this test run are:
...
19 passed, 1 warning in 0.12s (unit tests run by default)
```

**Recipe tests only with -k flag**
Finally, this command runs recipe tests only using a `-k` filter.
This will fetch recipe test models and tokenizers and run only recipe tests matching the wildcard string.
```
pytest tests -m integration_test -k 'test_training_state_on_resume'
small-ckpt-01242024
small-ckpt-tune-03082024.pt
small-ckpt-meta-03082024.pt
small-ckpt-hf-03082024.pt
...
3 passed, 203 deselected, 1 warning in 55.25s
```

CI should be green as well


joecummings (approved):

kartikayk:
@ebsmothers generally the concept makes sense to me. One noob question I would have is whether it would make sense to have the following mapping with no flags?

```
pytest tests/torchtune -> Unit tests
pytest tests/recipes -> Integration tests
pytest tests/regression_tests -> regression tests
pytest tests/ -> doesn't do anything
```

This makes it so that we'll have to manually run each sort of tests, but I think the RFC does that anyways? But I dont have to worry about remembering which flags to use. Not sure if this is possible though, so disregard if it isn't.

ebsmothers:
> @ebsmothers generally the concept makes sense to me. One noob question I would have is whether it would make sense to have the following mapping with no flags?
> 
> ```
> pytest tests/torchtune -> Unit tests
> pytest tests/recipes -> Integration tests
> pytest tests/regression_tests -> regression tests
> pytest tests/ -> doesn't do anything
> ```
> 
> This makes it so that we'll have to manually run each sort of tests, but I think the RFC does that anyways? But I dont have to worry about remembering which flags to use. Not sure if this is possible though, so disregard if it isn't.

@kartikayk yeah good point. I think the reason I opted not to do this is because of the last case. 

For a bit more info on how this all works: if we want to download assets on-demand for the test run, we need to define that at the directory level, and any test run we execute will move up the directory tree until it hits root, with configs being successively overridden the higher you go. So to do what you're suggesting, we could have separate test configs in `tests/recipes` and `tests/regression_tests` to download just those assets (as opposed to right now where I have a single test config at root that parses all the flags). One other minor thing is that there is not yet a 1:1 mapping between test type and directory (e.g. we have config tests that I would claim are unit tests that are in `tests/recipes`). So this also solves for that case, though if you say "we should just move the test" I certainly wouldn't argue with you.

As you point out, the no-flag approach means we can't just run `pytest/tests`, which I think is something that people tend to do (e.g. #484 was based on a bug manifesting when @RdoubleA was running `pytest/tests`). Also this precludes stuff like `pytest tests -k 'some_unit_test'`. But ultimately it depends on what people find most intuitive and easiest to remember. My assumption based on #296 was that adding these flags would be a net positive to the testing experience. But I wanna make sure this holds in practice, if folks on the team think they are actually adding too much cognitive load then we should scrap them.

kartikayk:
@ebsmothers Oh no, the flags make a lot of sense to me. I definitely want to keep them. My main point was around optimizing the use of flags for the use cases we plan to run the most. In terms of what people generally do, I think we will need to educate them anyways, so I'd just recommend taking the approach which you think is best and not worry about general patterns there. Overall, I agree with this direction.

ebsmothers:
Seems that there have been some changes in how S3 permissions get handled in CI on PRs created from forks, causing signals here to be red. To unblock, I created a copy of this PR using the main repo in #573 so gonna close this one.

PR #571, opened 2024-03-24
Generalize configs and add Llama2 13B + Mistral 7B

kartikayk:
## Context

All relevant TorchTune components (recipes, checkpointers and CLI) can easily be generalized to models beyond Llama2 7B. In this PR, I show that adding Llama2 13B and Mistral7B is as simple as adding some new builder functions. The Llama2 13B currently assumes HF-format checkpoints (see configs) since I need to add support for being able to deal with Meta's sharded checkpoints. I'll do this in a follow-up PR. 

Accompanying this addition re some cosmetic changes which makes the repo more user friendly for multiple models. This includes better organizing our configs and shortening their names so its not cumbersome to type them out, and better organizing our models.

I manually tested out all of the commands in the different READMEs and docstrings (except QLoRA which might be currently impacted by the torchao-nightly change).

Training speed for 13B is quite competitive. Quick comparisons showed us to be 2.5x faster than some competitors without any change to recipe code. The next section shows the correctness checks.

**Note**: Mistral 7B requires some data preprocessing changes for Alpaca finetuning. I'll follow up with those changes in a separate PR. In this PR, I add support for the model and show numerical parity with HF.

## Changelog
- Add ```llama2_13b``` and ```lora_llama2_13b``` builder functions to support the Llama2 13B model
- Move all existing configs (except ```alpaca_generate.yaml```) from ```configs/``` to ```configs/llama2/```.
- Shorten the names of the config
- Add some documentation to each config so users understand when to use which config
- Make all associated changes to tests, CLI, docs and READMEs
- Take ```_convert_weights``` out of ```models/llama2``` and make this a public file which I expect users to use.
- Add ```_component_builders.py``` and ```_model_builders.py``` under ```models/mistral``` to support the ```mistral_7b``` model. 

## Correctness Checks for Llama 13B

Numeric Parity of Llama2 13B with HF's implementation

image

&nbsp;

Eval using Eleuther's Harness on ```truthfulqa_mc2```

Baseline: 36.9%  vs Finetuned: 47.1%

image

image
 
Loss Curve (loss is comparable to what some [forums shared](https://github.com/tloen/alpaca-lora/issues/28))

image

## Correctness Checks for Mistral 7B

image

## Test plan
- All tests pass

```
pytest tests
```

- 13B full-finetune on 4 devices

image

- 13B LoRA on 4 devices

```
tune --nnodes 1 --nproc_per_node 4 lora_finetune_distributed --config llama2/13B_lora.yaml
```

image

### DIFF
--- a/docs/source/examples/configs.rst
+++ b/docs/source/examples/configs.rst
@@ -161,7 +161,7 @@ will list out all the locations where an error was found.
 
 .. code-block:: bash
 
-  tune validate --config recipes/configs/full_finetune_single_device.yaml batch_size=4
+  tune validate --config recipes/configs/llama2/7B_full.yaml batch_size=4
 
 Best practices for writing configs
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
--- a/recipes/README.md
+++ b/recipes/README.md
@@ -6,7 +6,7 @@
 
 Recipes are the primary entry points for TorchTune users. These can be thought of as end-to-end pipelines for training and optionally evaluating LLMs. Each recipe consists of three components:
 
-- **Configurable parameters**, specified through yaml configs [example](https://github.com/pytorch/torchtune/blob/main/recipes/configs/7B_full.yaml) and command-line overrides
+- **Configurable parameters**, specified through yaml configs [example](https://github.com/pytorch/torchtune/blob/main/recipes/configs/llama2/7B_full.yaml) and command-line overrides
 - **Recipe class**, core logic needed for training, exposed to users through a set of APIs [interface](https://github.com/pytorch/torchtune/blob/main/recipes/interfaces.py)
 - **Recipe script**, puts everything together including parsing and validating configs, setting up the environment, and correctly using the recipe class
 
--- a/recipes/README.md
+++ b/recipes/README.md
@@ -6,7 +6,7 @@
 
 Recipes are the primary entry points for TorchTune users. These can be thought of as end-to-end pipelines for training and optionally evaluating LLMs. Each recipe consists of three components:
 
-- **Configurable parameters**, specified through yaml configs [example](https://github.com/pytorch/torchtune/blob/main/recipes/configs/full_finetune_distributed.yaml) and command-line overrides
+- **Configurable parameters**, specified through yaml configs [example](https://github.com/pytorch/torchtune/blob/main/recipes/configs/7B_full.yaml) and command-line overrides
 - **Recipe class**, core logic needed for training, exposed to users through a set of APIs [interface](https://github.com/pytorch/torchtune/blob/main/recipes/interfaces.py)
 - **Recipe script**, puts everything together including parsing and validating configs, setting up the environment, and correctly using the recipe class
 
--- a/recipes/configs/llama2/13B_full.yaml
+++ b/recipes/configs/llama2/13B_full.yaml
@@ -1,0 +1,85 @@
+# Config for multi-device full finetuning in full_finetune_distributed.py
+# using a Llama2 13B model
+#
+# This config assumes that you've run the following command before launching
+# this run:
+#   tune download --repo-id meta-llama/Llama-2-13b-hf \
+#   --hf-token <HF_TOKEN> \
+#   --output-dir /tmp/llama2-13b-hf
+#
+# To launch on 4 devices, run the following command from root:
+#   tune --nnodes 1 --nproc_per_node 4 full_finetune_distributed \
+#   --config llama2/13B_full \
+#
+# You can add specific overrides through the command line. For example
+# to override the checkpointer directory while launching training
+# you can run:
+#   tune --nnodes 1 --nproc_per_node 4 full_finetune_distributed \
+#   --config llama2/13B_full \
+#   checkpointer.checkpoint_dir=<YOUR_CHECKPOINT_DIR>
+#
+# This config should be used with 2+ GPUs. Single device full fine-tuning
+# requires several memory optimizations which are exposed through
+# 7B_full_single_device.yaml. Please update the model and checkpoints to 13B
+# in that config.
+
+
+# Tokenizer
+tokenizer:
+  _component_: torchtune.models.llama2.llama2_tokenizer
+  path: /tmp/llama2/tokenizer.model
+
+# Dataset
+dataset:
+  _component_: torchtune.datasets.alpaca_dataset
+  train_on_input: True
+seed: null
+shuffle: True
+
+# Model Arguments
+model:
+  _component_: torchtune.models.llama2.llama2_13b
+
+checkpointer:
+  _component_: torchtune.utils.FullModelHFCheckpointer
+  checkpoint_dir: /tmp/llama2-13b-hf/
+  checkpoint_files: [
+    pytorch_model-00001-of-00003.bin,
+    pytorch_model-00002-of-00003.bin,
+    pytorch_model-00003-of-00003.bin
+  ]
+  recipe_checkpoint: null
+  output_dir: /tmp/llama2-13b-hf/
+  model_type: LLAMA2
+resume_from_checkpoint: False
+
+# Fine-tuning arguments
+batch_size: 2
+epochs: 3
+optimizer:
+  _component_: torch.optim.AdamW
+  lr: 2e-5
+loss:
+  _component_: torch.nn.CrossEntropyLoss
+max_steps_per_epoch: null
+gradient_accumulation_steps: 1
+
+
+# Training env
+device: cuda
+
+# Distributed
+cpu_offload: False
+
+# Memory management
+enable_activation_checkpointing: True
+
+# Reduced precision
+dtype: bf16
+
+# Logging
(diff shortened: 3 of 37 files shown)


torchtune/_cli/cp.py
```diff
@@ -66,7 +66,7 @@ def main(parser):
         epilog=textwrap.dedent(
             """\
         examples:
-            $ tune cp lora_finetune_distributed.yaml ./my_custom_llama2_lora.yaml
+            $ tune cp llama2/7B_lora.yaml ./my_custom_llama2_lora.yaml
```
kartikayk, line 69: @joecummings I think appending the model name (eg: llama2) is fine and in fact will be critical as we add more models. I tested and this should work with our pkg structure. Let me know what you think.

recipes/README.md
```diff
@@ -6,7 +6,7 @@
 
 Recipes are the primary entry points for TorchTune users. These can be thought of as end-to-end pipelines for training and optionally evaluating LLMs. Each recipe consists of three components:
 
-- **Configurable parameters**, specified through yaml configs [example](https://github.com/pytorch/torchtune/blob/main/recipes/configs/full_finetune_distributed.yaml) and command-line overrides
+- **Configurable parameters**, specified through yaml configs [example](https://github.com/pytorch/torchtune/blob/main/recipes/configs/7B_full.yaml) and command-line overrides
```
joecummings, line 9: I think this path is wrong?

recipes/configs/llama2/7B_full.yaml
```diff
@@ -1,9 +1,27 @@
-# Config for FullFinetuneRecipe in full_finetune_distributed.py
+# Config for multi-device full finetuning in full_finetune_distributed.py
+# using a Llama2 7B model
 #
-# To launch, run the following command from root:
-#   tune --nnodes 1 --nproc_per_node 1 full_finetune_distributed \
-#   --config full_finetune_distributed \
-#   checkpointer.checkpoint_dir=<YOUR_CHECKPOINT_DIR> ...
+# This config assumes that you've run the following command before launching
```
joecummings, line 4: Why add this huge block in the config?

recipes/README.md
kartikayk: Oh yeh sorry, let me fix this

recipes/configs/llama2/7B_full.yaml
kartikayk: Currently there's no documentation on the configs at all. Once we have live docs available, we can add these to the docs. But for now, I'd like to give some understanding to users about when and how to use each config.

docs/source/examples/configs.rst
```diff
@@ -161,7 +161,7 @@ will list out all the locations where an error was found.
 
 .. code-block:: bash
 
-  tune validate --config recipes/configs/full_finetune_single_device.yaml batch_size=4
+  tune validate --config recipes/configs/llama2/7B_full.yaml batch_size=4
```
ebsmothers, line 164: One thing to watch out for: there can be lots of issues when filenames start with a number. I've definitely seen it as a problem with Python imports, maybe it will be ok with YAML files? But something to keep in mind

recipes/configs/llama2/13B_full.yaml
```diff
@@ -0,0 +1,85 @@
...
+# to override the checkpointer directory while launching training
+# you can run:
+#   tune --nnodes 1 --nproc_per_node 4 full_finetune_distributed \
+#   --config llama2/13B_full \
+#   checkpointer.checkpoint_dir=<YOUR_CHECKPOINT_DIR>
+#
+# This config should be used with 2+ GPUs. Single device full fine-tuning
+# requires several memory optimizations which are exposed through
+# 7B_full_single_device.yaml. Please update the model and checkpoints to 13B
+# in that config.
```
ebsmothers, line 24: I guess implicit in our choice of naming here is that >1 device is kind of now the "default", right? While I understand that we are doing more memory optimizations in the single device recipes now, we've obviously seen that FSDP comes with its own nuances too. So I do wonder if it's now hard for someone to just come in and say "give me a simple single-device recipe to get started on" 

This is also a bit weird for QLoRA imo where we currently only support single device

torchtune/_cli/ls.py
```diff
@@ -44,12 +44,12 @@ def main():
         examples:
             $ tune ls
             RECIPE                           CONFIG
-            full_finetune_distributed.py     full_finetune_distributed.yaml
-            lora_finetune_distributed.py     lora_finetune_distributed.yaml
+            full_finetune_distributed.py     llama2/7B_full, llama2/13B_full
+            lora_finetune_distributed.py     llama2/7B_lora, llama2/13B_lora
```
ebsmothers, line 48: Happy to finally see more than one config per recipe. Nit: maybe we can split the configs over separate lines? E.g.

```
RECIPE                                           CONFIG
full_finetune_distributed.py                     llama2/7B_full
                                                 llama2/13B_full
lora_finetune_distributed                        llama2/7B_lora
                                                 llama2/13B_lora
```

recipes/configs/llama2/13B_full.yaml
kartikayk: yeh this is a good question. I did this primarily for two reasons:
- With the distributed CI testing sorted out, I removed the constraint on the distributed recipes. We can now run those on single device but without the memory optimizations.
- As we go to larger models, the single device setting will be less frequent. So when I thought about the default, distributed seem like the more natural one.

Does this make sense?

docs/source/examples/configs.rst
kartikayk: This is interesting - what sort of issues? But yeh I don't expect us to be importing the configs anymore?

recipes/configs/llama2/7B_full.yaml
joecummings: I think there's still some information in the README on configs and we can make that more clear. I think cluttering up the configs can be overwhelming.

torchtune/_cli/ls.py
joecummings: Yep, my vision was to be over multiple lines and I think that's actually how it works right now?

recipes/configs/llama2/7B_full.yaml
kartikayk: Why would it be overwhelming? Isn't it just documentation? I don't think we would be able to add config-level info to the README?

rohan-varma (commented):
Awesome to see this! Wonder if we have a parity check for 13b model, comparing to HF forward outputs and/or Meta model outputs?

kartikayk:
@rohan-varma yup! I don't think I'll compare with the Meta implementation since I don't have the code for sharded checkpointing in a decent state (will do that comparison in a follow up PR). But I'll add a comparison with HF implementation.

kartikayk:
Added numerical parity checks and e2e eval comparisons for Llama2 13B to the context section. Thanks @joecummings for the help on this!

torchtune/_cli/ls.py
kartikayk: Yup, sorry I just need to update this example.

recipes/lora_finetune_distributed.py
```diff
@@ -60,7 +60,7 @@ class LoRAFinetuneRecipeDistributed(FTRecipeInterface):
     The following configs can be used to run this recipe:
         >>> tune ls
         RECIPE                         CONFIG
-        lora_finetune_distributed      lora_finetune_distributed
```
joecummings, line 63: Have you actually ran this command? I think it goes on multiple lines?

joecummings (approved):
Awesome change! LGTM

recipes/lora_finetune_distributed.py
kartikayk: yup!
image

recipes/lora_finetune_distributed.py
```diff
@@ -80,20 +80,7 @@ def __init__(self, cfg: DictConfig) -> None:
...
 
-        # For CUDA devices, check if the HW supports bf16 if bf16 is specified.
-        if (
-            self._dtype == torch.bfloat16
-            and self._device != torch.device("cpu")
-            and not torch.cuda.is_bf16_supported()
-        ):
-            raise RuntimeError("Full bf16 training is not supported on this hardware.")
-
-        world_size, rank = utils.get_world_size_and_rank()
```
kartikayk, line 91: Now that we have our distributed tests working as expected, I'm removing this constraint.

recipes/configs/mistral/7B_full.yaml
```diff
@@ -19,22 +15,25 @@ shuffle: True
...
+  ]
   recipe_checkpoint: null
-  output_dir: /tmp/llama2
+  output_dir: /tmp/Mistral-7B-v0.1
   model_type: LLAMA2
 resume_from_checkpoint: False
 
 # Fine-tuning arguments
-batch_size: 2
+batch_size: 32
```
rohan-varma, line 33: These config changes will make for high consumption memory for mistral, right?

Issue #570, opened 2024-03-23
Lora single device recipe failing for qlora


chauhang:
The single device lora recipe with qlora configs is failing with below error on new machine with fresh install:

### Command: 
```
tune lora_finetune_single_device --config recipes/configs/qlora_finetune_single_device.yaml 
```

### Error: 
```
  File "/home/gchauhan/torchtune/recipes/lora_finetune_single_device.py", line 455, in <module>
    sys.exit(recipe_main())
             ^^^^^^^^^^^^^
  File "/home/gchauhan/torchtune/torchtune/config/_parse.py", line 54, in wrapper
    sys.exit(recipe_main(conf))
             ^^^^^^^^^^^^^^^^^
  File "/home/gchauhan/torchtune/recipes/lora_finetune_single_device.py", line 449, in recipe_main
    recipe.setup(cfg=cfg)
  File "/home/gchauhan/torchtune/recipes/lora_finetune_single_device.py", line 152, in setup
    self._model = self._setup_model(
                  ^^^^^^^^^^^^^^^^^^
  File "/home/gchauhan/torchtune/recipes/lora_finetune_single_device.py", line 248, in _setup_model
    utils.validate_expected_param_dtype(model, dtype=self._dtype)
  File "/home/gchauhan/torchtune/torchtune/utils/precision.py", line 161, in validate_expected_param_dtype
    raise ValueError(f"Parameter {name} has dtype {param.dtype}")
ValueError: Parameter layers.0._checkpoint_wrapped_module.attn.q_proj.weight has dtype torch.float8_e5m2fnuz
```

### Environment:
Created by the default install (uses PT 2.2.1)
```
python -m torch.utils.collect_env

Collecting environment information...
PyTorch version: 2.2.1+cu121
Is debug build: False
CUDA used to build PyTorch: 12.1
ROCM used to build PyTorch: N/A

OS: CentOS Stream 9 (x86_64)
GCC version: (GCC) 11.4.1 20231218 (Red Hat 11.4.1-3)
Clang version: Could not collect
CMake version: Could not collect
Libc version: glibc-2.34

Python version: 3.11.8 (main, Feb 26 2024, 21:39:34) [GCC 11.2.0] (64-bit runtime)
Python platform: Linux-5.19.0-0_fbk11_zion_11538_g61e82533119f-x86_64-with-glibc2.34
Is CUDA available: True
CUDA runtime version: 12.0.140
CUDA_MODULE_LOADING set to: LAZY
[...]

PR #574, opened 2024-03-24
Fix QLoRA recipe, add CI

rohan-varma:
#### Context
- As reported by https://github.com/pytorch/torchtune/issues/570 and by @ebsmothers, QLoRA recently broke due to torchao nightly update. In this PR, I pin torchao-nightly to the last known working version which contains all features we need, re-enable the unittests, and add a recipe test for QLoRA.

#### Changelog
- See above
- Also a couple of changes to `get_precision`. In particular, we accept a `device` argument which is mostly used for bf16 support checking. bf16 can run on CPU (there might be some unsupported ops versus CUDA in general), so I don't think we should run in fp32 for CPU if bf16 is specified.
- Basically, previously if bf16 was specified, but the GPU did not support, we'd return torch.float32, even if the user was training on CPUs. Now, if the GPU does not support but the user is using CPU, we return bf16.

#### Test plan
- Unit:  pytest tests/recipes/test_lora_finetune_single_device.py -v -k test_loss
- Fresh install, verify expected ao nightly. Run QLoRA: `tune lora_finetune_single_device --config recipes/configs/qlora_finetune_single_device.yaml`

#### Follow up
- We need to establish a process for regularly upgrading deps for libraries which we use nightlies for. The simplest thing to do is what I've done now - pin the version and bump it whenever a new feature/bugfix we need is released. Though if there's a large gap in time between these, the upgrade process could be time consuming and cumbersome to resolve all issues. I could see a process where we upgrade ~weekly to avoid this and more regularly deal with the issues. Unfortunately since we're using nightlies of a library that doesn't provide BC guarantee at the moment, its tricky to think of a more robust upgrade process.

### DIFF
--- a/tests/recipes/test_lora_finetune_single_device.py
+++ b/tests/recipes/test_lora_finetune_single_device.py
@@ -57,7 +57,7 @@ class TestLoRAFinetuneSingleDeviceRecipe:
         ckpt_dir = ckpt_path.parent
         log_file = gen_log_file_name(tmpdir)
 
-        prefix = f"qlora" if run_qlora else "lora"
+        prefix = "qlora" if run_qlora else "lora"
         cmd = f"""
         tune lora_finetune_single_device
             --config {prefix}_finetune_single_device \
--- a/tests/recipes/test_lora_finetune_single_device.py
+++ b/tests/recipes/test_lora_finetune_single_device.py
@@ -79,9 +79,11 @@ class TestLoRAFinetuneSingleDeviceRecipe:
         )
 
         # TODO (rohan-varma): QLoRA only supported with bf16 for now
-        cmd = cmd + self._get_test_config_overrides(
-            dtype_str="bf16" if run_qlora else "fp32"
-        ) + model_config
+        cmd = (
+            cmd
+            + self._get_test_config_overrides(dtype_str="bf16" if run_qlora else "fp32")
+            + model_config
+        )
         monkeypatch.setattr(sys, "argv", cmd)
         with pytest.raises(SystemExit):
             runpy.run_path(TUNE_PATH, run_name="__main__")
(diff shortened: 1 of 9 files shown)


pbontrager:
> We need to establish a process for regularly upgrading deps for libraries which we use nightlies for. The simplest thing to do is what I've done now - pin the version and bump it whenever a new feature/bugfix we need is released. Though if there's a large gap in time between these, the upgrade process could be time consuming and cumbersome to resolve all issues. I could see a process where we upgrade ~weekly to avoid this and more regularly deal with the issues. Unfortunately since we're using nightlies of a library that doesn't provide BC guarantee at the moment, its tricky to think of a more robust upgrade process.

For MMF we use a CI bot that automatically creates a PR to update the pinned version whenever the dependency is updated. This allows it to be tested before it's merged into main while keeping the dependencies up to date at the same time.

tests/recipes/test_lora_finetune_single_device.py
```diff
@@ -43,19 +43,24 @@ def _get_test_config_overrides(self):
...
+    def _fetch_expected_loss_values(self, run_qlora: bool = False):
+        if run_qlora:
+            return [10.5057, 10.5575, 10.5179, 10.4898]
+        else:
+            return [10.5074, 10.5614, 10.5205, 10.4918]
 
     @pytest.mark.integration_test
-    def test_loss(self, tmpdir, monkeypatch):
+    @pytest.mark.parametrize("run_qlora", [True, False])
+    def test_loss(self, run_qlora, tmpdir, monkeypatch):
```
ebsmothers, line 54: Given that the loss, config file, and overrides are all different, maybe this just merits a separate `test_qlora_loss` test case? I know it adds copypasta but I think it'd be clearer

joecummings (approved):
go go go

PR #576, opened 2024-03-24
Chat dataset + SlimOrca refactor + more templates

RdoubleA:
## Context
Chat and conversational data is one of the most common datasets that OSS users want to fine-tune on. Including tools and abstractions that empower users to quickly configure their own chat dataset without the overhead of data preprocessing can be immensely valuable. 

The challenge here is designing an API that is general enough to apply to many chat datasets but not too rigid that it adds friction to the developer workflow. This is what I would primarily like early feedback on. You can see an example of how it generalizes with the `slimorca_dataset` builder.

**Challenge:** Conversational data can take many different formats and it's difficult to anticipate most or all of them

This is the biggest hurdle, but if we engineer a well-designed solution it would make users' lives significantly easier, or at least provide strong guidelines for how to customize to their own dataset. The approach we take here is to define a few lightweight abstractions:
```
Role = Literal["system", "user", "assistant"]

class Message(TypedDict):
    role: Role
    content: str

Dialogue = List[Message]
```
These are not new ideas, these were taken straight from Meta's llama inference [repo](https://github.com/meta-llama/llama/blob/main/llama/generation.py). Axolotl also does something [similar](https://github.com/OpenAccess-AI-Collective/axolotl/blob/34ba634b8c114ece6ee5d5a38b4d2dc4b8067aa2/src/axolotl/prompt_strategies/orpo/chat_template.py#L19). We need to enforce a particular format so that other components can be easily designed around this assumption, and it's not entirely unreasonable to place the burden on users to format their data in this way. This tradeoff is preferable to designing for ANY type of conversation format, or multiple branching if-else statements.

The user will need to do this via `convert_to_dialogue`, a mandatory Callable parameter. The contract is pretty clear: process a `Sample` and return a `Dialogue`. You can see an example in the `sharegpt_to_llama_dialogue` transform. Users may typically want to transform their data anyway as a preprocessing step before templating and tokenization; this parameter simply takes the place of that.

**Challenge:** Multi-turn conversations

Handling multiple turns requires template each turn individually, and simultaneously respecting max sequence length, which can easily lead to a convoluted for loop. I think the approach here ended up being relatively straight-forward, but I need feedback here to see if I missed any edge cases.

**Challenge:** Interop with sample packing

This is still something I'm working through, so it is TBD.

## Changelog
- Added `ChatDataset` abstraction and unit tests
- Refactored `SlimOrcaDataset` -> `slimorca_dataset` builder
- Added new chat templates: `Llama2ChatTemplate`, `MistralChatTemplate`, `ChatMLTemplate`
- Refactored some data utilities: `tokenize_prompt_and_response`, `truncate_if_necessary`
- Moved all non-dataset related files under `torchtune/data/`

## Test plan
All unit tests and integration tests pass:

E2E test with a recipe: TODO


torchtune/data/_templates.py
```diff
@@ -5,7 +5,9 @@
 # LICENSE file in the root directory of this source tree.
 
 from abc import ABC, abstractmethod
-from typing import Any, Dict, Mapping, Optional
+from typing import Dict, Optional
+
+from torchtune.data._types import Sample
```
joecummings, line 10: Nit: What does creating a new datatype add for us? 

`Mapping[str, Any]` is common and makes sense, whereas the notion of a sample is something new entirely. If the value add is really tremendous, this makes sense, but in the spirit of our design principles to keep things as simple as possible, I want us to be conscious about this decision.

tests/torchtune/config/test_utils.py
```diff
@@ -9,9 +9,11 @@
 import pytest
 from torchtune.config._utils import (
     _get_component_from_path,
+    _get_template,
```
ebsmothers, line 12: Not the most pressing for this particular PR, but I find it quite confusing that we have separate `test_utils.py` files in our repo that either provide utilities for tests or test our utilities. We should clean this up

tests/torchtune/data/test_templates.py
```diff
@@ -4,12 +4,23 @@
...
 from torchtune.data import (
     AlpacaInstructTemplate,
+    ChatMLTemplate,
     GrammarErrorCorrectionTemplate,
+    Llama2ChatTemplate,
+    MistralChatTemplate,
     SummarizeTemplate,
 )
 
+# Taken from Open-Orca/SlimOrca-Dedup on HuggingFace
```
ebsmothers, line 17: add URL

tests/torchtune/data/test_templates.py
```diff
@@ -144,3 +155,98 @@ def test_format_with_column_map(self):
...
+    def test_format(self):
+        no_system_sample = CHAT_SAMPLE.copy()
+        del no_system_sample["system"]
+        actual = self.template.format(no_system_sample)
+        assert actual == self.expected_prompt
+
+        with pytest.raises(
+            ValueError, match="System prompts are not supported in MistralChatTemplate"
+        ):
+            _ = self.template.format(CHAT_SAMPLE)
```
ebsmothers, line 211: nit: should be a separate test case

torchtune/config/_utils.py
```diff
@@ -148,3 +150,37 @@ def _merge_yaml_and_cli_args(yaml_args: Namespace, cli_args: List[str]) -> DictC
...
     # CLI takes precedence over yaml args
     return OmegaConf.merge(yaml_conf, cli_conf)
+
+
+def _get_template(template: str) -> PromptTemplate:
+    """
+    Get the prompt template class from the template string.
+
+    String should either be the PromptTemplate class name directly, or a raw
+    string with 1 or more placeholders. If none of these apply, then raise an
```
ebsmothers, line 160: Dumb q: how often do people actually want to configure prompt template by directly passing a format string? E.g. 

```
tune full_finetune_single_device --config llama2/7B_full dataset.template=f"Below is an instruction that describes a task, paired with an input that provides further context. Write a response that appropriately completes the request.\n\n### Instruction:\n{instruction}\n\n### Input:\n{input}\n\n### Response:\n"
```

doesn't exactly roll off the keyboard

torchtune/data/_templates.py
RdoubleA: I can concede on Sample since it's more of a cosmetic change, but the other types are needed to enforce a standardized conversation format and roles.

torchtune/data/_utils.py
```diff
@@ -0,0 +1,73 @@
...
+
+    if not train_on_input:
+        labels[: len(encoded_prompt)] = [CROSS_ENTROPY_IGNORE_IDX] * len(encoded_prompt)
+
+    assert len(encoded_prompt_with_response) == len(labels)
+
+    return encoded_prompt_with_response, labels
+
+
+def truncate_if_necessary(
```
ebsmothers, line 48: nit: can probably just call this truncate

torchtune/config/_utils.py
```diff
@@ -148,3 +150,37 @@ def _merge_yaml_and_cli_args(yaml_args: Namespace, cli_args: List[str]) -> DictC
...
+            template string
+    """
+    path = "torchtune.data." + template
+    try:
+        template_class = _get_component_from_path(path)
+        return template_class()
+    except InstantiationError:
+        # Verify that string can be used as a template, should have variable
+        # placeholders
+        pattern = r"\{.+?\}"
```
ebsmothers, line 180: Is this the most robust validation? E.g. I think `\{hello\}` will pass but is not a valid template. Not to mention that we are not validating # of args or anything like that. Not a huge deal cause I know config validation is hard, but just wanna be realistic about how much we can accomplish with this.

torchtune/config/_utils.py
```diff
@@ -148,3 +150,37 @@ def _merge_yaml_and_cli_args(yaml_args: Namespace, cli_args: List[str]) -> DictC
...
+        template (str): class name of template, or string with placeholders
+
+    Returns:
+        PromptTemplate: the prompt template class or the same verified string
+
+    Raises:
+        ValueError: if the template is not a PromptTemplate class or a proper
+            template string
+    """
+    path = "torchtune.data." + template
```
ebsmothers, line 173: I'm confused, isn't this different from our usual instantiate logic? Why the change here?

tests/torchtune/config/test_utils.py
```diff
@@ -107,3 +109,33 @@ def test_merge_yaml_and_cli_args(self, mock_load):
...
+
+        # Test invalid template class
+        with pytest.raises(
+            ValueError,
+            match="Must be a PromptTemplate class or a string with placeholders.",
+        ):
+            _ = _get_template("InvalidTemplate")
+
+        # Test valid template strings
+        s = [
```
ebsmothers, line 126: nit: better variable names

torchtune/config/_utils.py
ebsmothers: Also I guess a higher-level point here: is the support for configurable f strings worth the complexity that we are adding? This is a bunch of stuff happening on the config backend basically just to have our config support a Union type. Wondering if there's a simpler way we can expose the f-string functionality (e.g. provide a builder taking in f-string as its only arg)

tests/torchtune/config/test_utils.py
RdoubleA: maybe rename this to `test_config_utils`?

torchtune/config/_utils.py
RdoubleA: > doesn't exactly roll off the keyboard

This could be specified in the config file itself, which I can see being a common use case

> Wondering if there's a simpler way we can expose the f-string functionality (e.g. provide a builder taking in f-string as its only arg)

yes I was planning to have a CustomTemplate class that does exactly this, but you'd still need to check to see if a string was passed instead of a template name

torchtune/config/_utils.py
RdoubleA: you know what I could just use instantiate

torchtune/config/_utils.py
ebsmothers: > yes I was planning to have a CustomTemplate class that does exactly this, but you'd still need to check to see if a string was passed instead of a template name

Just have 

```
CustomTemplate(PromptTemplate):
	def __init__(self, f_string):
		self.template = f_string
	
	def format()...
```
and either 

```
template:
	_component_: AlpacaInstructTemplate
```
or
```
template:
	_component_: CustomTemplate
	f_string: "{user}: a, {system}: b"
```

Then `instantiate(cfg.template)` will work either way, no? (I might be making a stupid config mistake here, lmk if so)

tests/torchtune/data/test_utils.py
```diff
@@ -0,0 +1,95 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
```
ebsmothers, line 6: Same comment here about filename

torchtune/config/_utils.py
RdoubleA: the template goes in the builder function, so it would be a nested component :) this was the reason I needed a str mapping here unfortunately... I don't like it but I haven't figured out a workaround

torchtune/data/_utils.py
RdoubleA: if it's not past max_seq_len then it won't truncate. I didn't want to convey to the user that I'm always truncating. WDYT?

torchtune/config/_utils.py
RdoubleA: why is `{hello}` not a valid template? It technically is with the variable placeholder `hello`

torchtune/data/_utils.py
kartikayk: My general understanding with truncate is that it's a no op below a certain length. So should be ok to shorten the name to truncate

torchtune/datasets/_chat.py
```diff
@@ -0,0 +1,182 @@
...
+    Sample,
+    tokenize_prompt_and_response,
+    truncate_if_necessary,
+)
+from torchtune.modules import Tokenizer
+
+
+class ChatDataset(Dataset):
+    """
+    Class that supports any custom dataset with multiturn conversations.
```
kartikayk, line 25: Is this true? This only supports custom datasets from the datasets repo?

RdoubleA: I mean that users can use this to configure any local or huggingface dataset, just like `InstructDataset`

torchtune/datasets/_chat.py
```diff
@@ -0,0 +1,181 @@
...
+        prompt_tokens = []
+        label_tokens = []
+        for prompt, label in self._get_turns(dialogue):
+            formatted_prompt = self.template.format(prompt)
+            encoded_prompt_with_response, labels = tokenize_prompt_and_response(
+                tokenizer=self._tokenizer,
+                prompt=formatted_prompt,
+                response=label,
+                train_on_input=self.train_on_input,
+            )
```
ebsmothers, line 94: We need to think about whether we can make this more general

tests/torchtune/data/test_data_utils.py
```diff
@@ -0,0 +1,95 @@
...
+    assert tokenized_label == expected_tokenized_label
+
+    tokenized_prompt, tokenized_label = tokenize_prompt_and_response(
+        tokenizer, prompt, response, train_on_input=True
+    )
+    assert tokenized_prompt == expected_tokenized_prompt
+    assert tokenized_label == expected_tokenized_prompt
+
+
+def test_truncate_if_necessary():
```
ebsmothers, line 71: nit: rename this too now

tests/torchtune/datasets/test_chat_dataset.py
```diff
@@ -0,0 +1,197 @@
...
+
+
+class DummyTokenizer:
+    def encode(self, text, **kwargs):
+        words = text.split()
+        return [len(word) for word in words]
+
+    @property
+    def eos_id(self):
+        return -1
```
ebsmothers, line 23: Can consider putting this in test_utils if we're defining it in multiple places

tests/torchtune/datasets/test_chat_dataset.py
```diff
@@ -0,0 +1,197 @@
...
+            template=template,
+            max_seq_len=100,
+            train_on_input=False,
+        )
+
+        # Test a normal multiturn dialogue
+        prompts, responses = zip(
+            *[(p, l) for p, l in ds._get_turns(dialogue[0]["dialogue"])]
+        )
+        print(prompts, responses)
```
ebsmothers, line 83: remove

tests/torchtune/datasets/test_chat_dataset.py
```diff
@@ -0,0 +1,197 @@
...
+        return [
+            {
+                "dialogue": [
+                    Message(role="system", content="You are an AI assistant."),
+                    Message(role="user", content="What is the meaning of life?"),
+                    Message(role="assistant", content="The meaning of life is 42."),
+                    Message(role="user", content="That's ridiculous."),
+                    Message(role="assistant", content="I agree."),
+                ],
+            },
```
ebsmothers, line 65: I'm a bit confused.. why isn't dialogue just a list of messages? Why the extra wrapping in dict with just "dialogue" key?

RdoubleA: made it this way to represent real data, which would be a row in a data table with various fields, and the conversations field points to the actual list of messages.

torchtune/datasets/_chat.py
RdoubleA: yeah - let's tokenize an individual prompt or response. let me rethink this

tests/torchtune/datasets/test_chat_dataset.py
```diff
@@ -0,0 +1,197 @@
...
+            source="iam/agoofy/goober",
+            convert_to_dialogue=lambda x: x["dialogue"],
+            template=template,
+            max_seq_len=100,
+            train_on_input=False,
+        )
+        assert len(ds) == 1
+        mock_load_dataset.assert_called_once()
+
+        for i in range(len(ds)):
```
ebsmothers, line 193: nit: prob don't need the for loop after the len assert

tests/torchtune/datasets/test_chat_dataset.py
```diff
@@ -0,0 +1,197 @@
...
+            prompt, label = ds[i]
+            print(prompt, label)
```
ebsmothers, line 195: remove

torchtune/config/_utils.py
ebsmothers: I thought we can get around nested components by building separately and piecing together in the recipe? Anyways if you think that hurts UX too much (or I am still missing your point) feel free to ignore

torchtune/data/_utils.py
```diff
@@ -0,0 +1,73 @@
...
+    if not train_on_input:
+        labels[: len(encoded_prompt)] = [CROSS_ENTROPY_IGNORE_IDX] * len(encoded_prompt)
+
+    assert len(encoded_prompt_with_response) == len(labels)
+
+    return encoded_prompt_with_response, labels
+
+
+def truncate(
+    tokenizer: Tokenizer,
```
ebsmothers, line 49: nit: do we really need to pass the full tokenizer? Maybe just eos_id would be sufficient here

ebsmothers (approved):
Great to see the improved chat dataset support! Left a bunch of comments but no major concerns from my side.

torchtune/datasets/_chat.py
```diff
@@ -0,0 +1,181 @@
...
+    Build a configurable dataset with conversations. This method should be
+    used to configure a custom chat dataset from the yaml config instead of
+    using `ChatDataset` directly, as it is made to be config friendly.
+
+    Args:
+        tokenizer (Tokenizer): Tokenizer used to encode data. Tokenize must implement an `encode` and `decode` method.
+        source (str): path string of dataset, anything supported by HuggingFace's `load_dataset`
+            (https://huggingface.co/docs/datasets/en/package_reference/loading_methods#datasets.load_dataset.path)
+        conversation_format (str): string specifying expected format of conversations in the dataset
+            for automatic conversion to the llama format. Supported formats are: "sharegpt"
```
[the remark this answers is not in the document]
RdoubleA: Format might be overloaded - here I mean what fields/roles to expect in the data. I purposely enforce users to convert to what our abstractions expect (calling it llama format) so that the subsequent steps can operate on these assumptions (templating, tokenizing, etc assume system role is called "system", user role is called "user" and not "human", etc). This is separate from the prompt template that a model expects, for example using the [INST] tags

torchtune/data/_utils.py
RdoubleA: I've been passing the whole tokenizer to be consistent with other methods - but no strong opinion here

torchtune/datasets/_chat.py
```diff
@@ -0,0 +1,181 @@
...
+        **load_dataset_kwargs (Dict[str, Any]): additional keyword arguments to pass to `load_dataset`.
+
+    Returns:
+        ChatDataset: the configured ChatDataset
+
+    Raises:
+        ValueError: if the conversation format is not supported
+    """
+    if conversation_format == "sharegpt":
+        convert_to_dialogue = sharegpt_to_llama2_dialogue
```
ebsmothers, line 169: Is this imported anywhere?

RdoubleA: huh, all my tests ran successfully on locally. maybe i need to push

torchtune/config/_utils.py
RdoubleA: ok I think I just needed to tighten the regex:
image

torchtune/config/_utils.py
RdoubleA: no - this is strictly for the dataset builders. this method was originally in datasets/utils but I moved it to config since it was more akin to config functionality

PR #587, opened 2024-03-25
Update contributing with latest testing guidelines

ebsmothers:
After #537, we need to update the contributing guide to reflect the recommended way to run tests in the repo.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -19,7 +19,7 @@ pip install -e ".[dev]"
 &nbsp;
 
 ## Testing
-TorchTune contains three different types of tests: unit tests, recipe test, and regression tests. These tests are distinguished by their complexity and the resources they require to run. Recipe tests and regression tests are explicitly marked via pytest.mark decorators.
+TorchTune contains three different types of tests: unit tests, recipe test, and regression tests. These tests are distinguished by their complexity and the resources they require to run. Recipe tests and regression tests are explicitly marked via pytest.mark decorators and both require S3 access to download the requisite assets.
 
 - **Unit tests**
   - These should be minimal tests runnable without remote access. (No large models, no downloading weights). Unit tests should be under [tests/torchtune](https://github.com/pytorch/torchtune/tree/main/tests/torchtune).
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -18,12 +18,26 @@ pip install -e ".[dev]"
 
 &nbsp;
 
-## Unit Tests and Recipe Tests
```
joecummings, line 21: Is there a note anywhere in here that says you need AWS access to see certain checkpoints?

joecummings (approved):
lgtm, one q

CONTRIBUTING.md
ebsmothers: I mentioned it under the recipe tests section, do you think it should be more visible?

CONTRIBUTING.md
```diff
@@ -18,12 +18,26 @@ pip install -e ".[dev]"
...
+## Testing
+TorchTune contains three different types of tests: unit tests, recipe test, and regression tests. These tests are distinguished by their complexity and the resources they require to run. Recipe tests and regression tests are explicitly marked via pytest.mark decorators.
+
+- **Unit tests**
+  - These should be minimal tests runnable without remote access. (No large models, no downloading weights). Unit tests should be under [tests/torchtune](https://github.com/pytorch/torchtune/tree/main/tests/torchtune).
+  - All unit tests can be run via ```pytest tests```.
+- **Recipe tests**
+  - These are relatively small-scale integration tests for running our recipes. These include
+  both single-device recipes and distributed recipes. In the latter case, tests should be marked with the `@gpu_test` decorator to indicate how many GPUs they need to run.
+  - Recipe tests require remote access as (small) model weights will be downloaded from S3 to run them.
```
msaroufim, line 30: nit: but would change s3 to hf hub cause that's likely what most of our users will be using with the exception of the core torchtune team

CONTRIBUTING.md
```diff
@@ -18,12 +18,26 @@ pip install -e ".[dev]"
...
+  - Recipe tests are found under [tests/recipes](https://github.com/pytorch/torchtune/tree/main/tests/recipes) and should be marked with the `@pytest.mark.integration_test` decorator.
+  - To run only recipe tests, you can run `pytest tests -m integration_test`.
+- **Regression tests**
```
msaroufim, line 33: can we be more explicit it seems to me like regression tests really means accuracy tests?

ebsmothers: Tbh it doesn't exactly mean that. Like one of the regression tests is just a test of the losses on the full 7B model over a few iterations. So I am deliberately being a little vague here, but lmk if this is coming at the cost of clarity.

CONTRIBUTING.md
ebsmothers: Oh interesting. My assumption here was that we would always use S3 for testing assets. So do you envision us just pushing our toy models to the hub and having some kind of fallback S3 -> HF hub based on permissions then?

CONTRIBUTING.md
ebsmothers: Updated

PR #589, opened 2024-03-26
Update torchao to latest working nightly, QLoRA fixes

rohan-varma:
#### Context
- Note: This PR should be a one-off (maybe n-off, for small n) to get us to a working state with a relatively recent version of torchao to bring in required changes for upcoming features. We should work towards an automated mechanism such as a bot to automatically upgrade packages when new ones are available and auto-run the appropriate CI tests.
- Upgrades torchao to 3/25 nightly, which contains features needed to run QLoRA in fp32
- Make appropriate changes to tests / recipes.
- In particular, get rid of torch.arange calls for QLoRA in tests. These are in `fixed_init_model`, but fixed init is not needed for the current QLoRA tests we are running.
- Limit bf16 validation to only adapter params. This is admittedly incomplete and we need to validate base model params are nf4 or bf16 based on quantization config. I'd like to punt this to a future PR as it requires more thought on how to get the param names we expect to be nf4, and I'd like to prioritze landing this ASAP.

#### Changelog
- See above

#### Test plan
- `pytest tests/torchtune/models/test_lora_llama2.py`
- `tune lora_finetune_single_device --config llama2/7B_qlora_single_device`

### DIFF
--- a/recipes/lora_finetune_single_device.py
+++ b/recipes/lora_finetune_single_device.py
@@ -23,7 +23,6 @@ from torchtune.modules.peft.peft_utils import (
     set_trainable_params,
     validate_state_dict_for_lora,
 )
-
 from torchtune.recipe_interfaces import FTRecipeInterface
 from tqdm import tqdm
 
@@ -245,8 +244,14 @@ class LoRAFinetuneRecipeSingleDevice(FTRecipeInterface):
         if lora_weights_state_dict:
             model.load_state_dict(lora_weights_state_dict, strict=False)
 
-        # Validate model was loaded in with the expected dtype.
-        utils.validate_expected_param_dtype(model, dtype=self._dtype)
+        # Validate model was loaded in with the expected dtype, allowing for separate
+        # dtypes for base and adapter params.
+        # Validate model adapter params were loaded in with the expected dtype
+        # TODO (rohan-varma): Further validation to ensure the appropriate base params
+        # are NF4 vs bf16 based on the quantization config.
+        utils.validate_expected_param_dtype(
+            self.adapter_params.items(), dtype=self._dtype
+        )
 
         log.info(f"Model is initialized with precision {self._dtype}.")
         log.info(
--- a/recipes/lora_finetune_single_device.py
+++ b/recipes/lora_finetune_single_device.py
@@ -244,8 +244,6 @@ class LoRAFinetuneRecipeSingleDevice(FTRecipeInterface):
         if lora_weights_state_dict:
             model.load_state_dict(lora_weights_state_dict, strict=False)
 
-        # Validate model was loaded in with the expected dtype, allowing for separate
-        # dtypes for base and adapter params.
         # Validate model adapter params were loaded in with the expected dtype
         # TODO (rohan-varma): Further validation to ensure the appropriate base params
         # are NF4 vs bf16 based on the quantization config.
--- a/requirements.txt
+++ b/requirements.txt
@@ -8,4 +8,4 @@ tqdm
 omegaconf
 
 # Quantization
-torchao-nightly==2024.3.13
+torchao-nightly==2024.3.25
--- a/tests/torchtune/models/test_lora_llama2.py
+++ b/tests/torchtune/models/test_lora_llama2.py
@@ -222,14 +222,14 @@ class TestLoRALlama2:
                 embed_dim=512,
                 run_fixed_init=False,
             )
-            qlora_sd = qlora.state_dict()
-            model_ref.load_state_dict(qlora_sd)
-            # Forward pass of model_ref and qlora should be the same, as QLoRA linear layers should use
-            # a special linear operator that runs the compute in bf16, but only saves the 4 bit tensors
-            # for backward.
-            ref_output = model_ref(inputs)
-            output = qlora(inputs)
-            torch.testing.assert_close(ref_output, output)
+        qlora_sd = qlora.state_dict()
+        model_ref.load_state_dict(qlora_sd)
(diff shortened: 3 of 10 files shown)


tests/torchtune/models/test_lora_llama2.py
```diff
@@ -276,4 +289,6 @@ def test_qlora_llama2_merged_state_dict(self):
 
         # Ensure checkpoint can be loaded into non-LoRA model
         with utils.set_default_dtype(torch.bfloat16):
-            llama2 = self.get_ref_llama2(vocab_size=50)
+            llama2 = self.get_ref_llama2(vocab_size=50, embed_dim=512)
+
+        llama2.load_state_dict(merged_ckpt)
```
rohan-varma, line 294: Missed this in the PR that added the test

requirements.txt
```diff
@@ -8,4 +8,4 @@ tqdm
 omegaconf
 
 # Quantization
-torchao-nightly==2024.3.13
+torchao-nightly==2024.3.25
```
ebsmothers, line 11: So we are still gonna pin to a specific version instead of using latest nightly?

torchtune/utils/precision.py
```diff
@@ -161,18 +171,20 @@ def set_default_dtype(dtype: torch.dtype) -> Generator[None, None, None]:
...
+def validate_expected_param_dtype(
+    named_params: Iterable[Tuple[str, nn.Parameter]], dtype: torch.dtype
+) -> None:
     """
-    Validates that all parameters in the model have the expected dtype.
+    Validates that all input parameters have the expected dtype.
 
     Args:
-        model (torch.nn.Module): Model to validate.
+        named_params (Iterable[Tuple[str, nn.Parameter]]): Iterable of named parameters.
```
ebsmothers, line 181: Any particular reason for the change from model to its named parameters? If not, I personally slightly preferred the model API

recipes/lora_finetune_single_device.py
```diff
@@ -245,8 +244,12 @@ def _setup_model(
...
             model.load_state_dict(lora_weights_state_dict, strict=False)
 
-        # Validate model was loaded in with the expected dtype.
-        utils.validate_expected_param_dtype(model, dtype=self._dtype)
+        # Validate model adapter params were loaded in with the expected dtype
+        # TODO (rohan-varma): Further validation to ensure the appropriate base params
+        # are NF4 vs bf16 based on the quantization config.
+        utils.validate_expected_param_dtype(
+            self.adapter_params.items(), dtype=self._dtype
+        )
```
ebsmothers, line 252: Why was `validate_expected_param_dtype` passing before? Shouldn't it have contained NF4 dtypes?

torchtune/utils/precision.py
```diff
@@ -161,18 +171,20 @@ def set_default_dtype(dtype: torch.dtype) -> Generator[None, None, None]:
         torch.set_default_dtype(old_dtype)
 
 
-def validate_expected_param_dtype(model: torch.nn.Module, dtype: torch.dtype) -> None:
+def validate_expected_param_dtype(
+    named_params: Iterable[Tuple[str, nn.Parameter]], dtype: torch.dtype
```
ebsmothers, line 175: nit: technically the get_adapter_params method in peft_utils is typed as returning a `Dict[str, Any]` that we pass along to this, so if we're doing it this way we may want to change `Any` -> `nn.Parameter`

tests/torchtune/models/test_lora_llama2.py
```diff
@@ -109,17 +110,22 @@ def get_lora_llama2(
         # To make final outputs less trivial
         if reset_norm:
             model.norm = nn.Identity()
-        fixed_init_model(model)
+
+        # Some tests don't rely on fixed_init, NF4Tensor does not have op
+        # coverage to cover ops done in fixed_init (i.e. arange). So only
+        # run fixed_init if needed.
```
ebsmothers, line 116: Couple questions: (a) is the op needed to support arange a lot of work to add an override for? (b) can we deterministically initialize tensors in another way without using an unsupported op? E.g. just use regular Python range and populate in tensor. I know the QLoRA tests don't need it at this moment but it's kind of a weird rough edge for us to leave laying around.

requirements.txt
rohan-varma: Yeah, too much risk for things to break out of our control especially now that we're public and need QLoRA consistently working/

recipes/lora_finetune_single_device.py
rohan-varma: Before in torchao, nf4.dtype would give torch.bf16 so this passed. now nf4.dtype gives `float8e5m2`

tests/torchtune/models/test_lora_llama2.py
rohan-varma: So I'm chatting with the ao team about what to do in the longer term here. Adding support for `arange` won't be similar to the work we did for torch_dispatch-based op registrations, since `arange` is a factory function (produces torch tensors) rather than an op on existing tensors. 

> t's kind of a weird rough edge for us to leave laying around

Totally agree with this - this is just to unblock and I'll have some follow up thoughts asap after chatting with ao folks.

rohan-varma: >  (b) can we deterministically initialize tensors in another way without using an unsupported op? E.g. just use regular Python range and populate in tensor

Sure, let me try to do this

rohan-varma: Ok, worked around this by forcing the fixed init when initializing QLoRA models to be bf16.

ebsmothers (approved):
Looks good!

PR #593, opened 2024-03-26
Move doc-preview to a separate job

svekars:
#### Context
The doc-build job needs to run on ubuntu-latest but the upload-preview needs to run on a self-hosted worker to be able to upload to S3. Therefore, moving doc-preview step to a separate job so that it can run on a self-hosted worker independently of other jobs.


NicolasHug (approved):
stamping

PR #595, opened 2024-03-26
Add fp32 support for QLoRA

rohan-varma:
#### Context
- QLoRA is currently coupled to bf16, but some older HW types don't support bf16, and we'd ideally like to enable at least one memory efficient finetuning solution for these HW arches. For example, T4s are 16GB and don't support bf16. This PR enables QLoRA to run compute + checkpoint in fp32 instead of bf16, to eliminate the hard coupling of bf16 to QLoRA  training.

#### Changelog
- Remove bf16 assumptions from nf4
- Remove bf16 assumptions from LoRALinear and upstream
- Generalize a few functions to be less coupled to bf16
- Fix tests

#### NOTE
- We've currently forked over the LinearNF4 from torchao while changes to LinearNF4 land in ao are in progress. Will revert back to using ao's implementation asap, but we need the changes in this forked version to decouple support from fp32. This is a temporary (~days) mitigation. 

#### Test plan
- Modified unittests - computation in fp32, checkpointing, parity is covered. fp32 coverage for qlora is at the same coverage as bf16.
- Verified manually that gradients are computed in fp32 (via inspecting .grad field)
- Checkpoints are saved in fp32:
image

- Run recipe: `tune lora_finetune_single_device --config llama2/7B_qlora_single_device dtype=fp32 epochs=1`
- Loss is comparable to QLoRA and LoRA-bf16 (see https://github.com/pytorch/torchtune/pull/478 for those curves): 
image

- Memory - increase over QLoRA bf16 is +20% peak memory allocated, +3% reserved memory

```
Memory Stats::
 GPU peak memory allocation: 6.98 GB
 GPU peak memory reserved: 9.57 GB
 GPU peak memory active: 6.98 GB
```
- Eval result:


torchtune/models/llama2/_component_builders.py
```diff
@@ -253,10 +258,16 @@ def lora_llama2(
...
 
     if quantize_base:
-        # For QLoRA, we reparametrize 4-bit tensors to bf16, and offload to CPU on the fly
+        # For QLoRA, we reparametrize 4-bit tensors to higher precision, and offload to CPU on the fly
         # so as to not increase peak memory
         model._register_state_dict_hook(
-            partial(reparametrize_as_bf16_state_dict_post_hook, offload_to_cpu=True)
+            partial(
+                reparametrize_as_dtype_state_dict_post_hook,
+                # TODO this is clowny, figure out a better way to get what precision the rest
```
ebsmothers, line 266: Honestly I don't really see a better way to do this

torchtune/modules/peft/lora.py
```diff
@@ -9,6 +9,8 @@
 import torch.nn.functional as F
 
 from torch import nn, Tensor
+
+# from torchtune.modules.low_precision.nf4_linear import _linear_nf4
```
ebsmothers, line 13: remove

ebsmothers (approved):
Looks great!

PR #596, opened 2024-03-26
torch.compile support for single device recipes

rohan-varma:
#### Context
- Adds torch.compile support for speed ups for single device recipes (lora, qlora, full finetunes) _only_ tested for llama-7b currently.
- `compile` support is added via a utility `wrap_compile`. `wrap_compile` is very simple, it simply calls into `torch.compile`, and also registers a `state_dict` post hook on the model to strip off a prefix "orig_mod" that is added by the compile function. This enables compile model produced state_dicts to behave exactly as an eager mode model, and centralizes this logic into a single utility instead of various different checkpointers having to check for this prefix and strip it out. 
- compile is configured via a `cfg.compile` flag in recipes. 
- This PR is limited to compile support on _single device_ for llama7b. The interaction with distributed training, such as FSDP, and torch.compile is very nuanced and fast-evolving - discussions around this will be punted to a future PR.

#### Changelog
- Adds a `compile` flag to config to control torch.compile
- If `compile` is true, runs `torch.compile`.
- Checkpointing changes are appropriately made.

#### Speed ups
- All reported speed ups are measured for 100 iterations and averaged over, beginning at iteration 300 (allowed a couple hundred iters to warm up). 
- Full finetune no compile: 3.022900000000003, Full finetune w/torch compile: 4.889200000000003 **(+61%)**
- LoRA finetune no compile: 3.0531000000000006, LoRA finetune w/compile: 4.123499999999996 (**+35%**) 
- QLoRA finetune no compile: 1.1966666666666668, QLoRA finetune w/compile: 3.6063636363636378 (+201%)
- QLoRA finetune only supported w/latest pytorch nightly. In CI, only runs with the nightlies. 
- Averages computed with this script: https://gist.github.com/rohan-varma/a33e75471faccdad7f0b6dfeb82a678c

#### Compile Overhead

- Full finetune's first iteration is 108.48s, reflecting the overhead in compile time. Similarly, lora finetune's is 122.85s. QLoRA compile time overhead is ~300s.

#### Accuracy
- LoRA w/compile: ```truthfulqa_mc2: {'acc,none': 0.48572842802124694, 'acc_stderr,none': 0.014656380523451883, 'alias': 'truthfulqa_mc2'}```
- Full w/compile: ```truthfulqa_mc2: {'acc,none': 0.41981779682199993, 'acc_stderr,none': 0.014149664935251433, 'alias': 'truthfulqa_mc2'}```
- These are at par with what we've seen with non-compile flows.

#### Checkpoint save/load
- Save a compiled checkpoint: ```tune run full_finetune_single_device --config recipes/configs/llama2/7B_full_single_device.yaml device=cuda:4 compile=True max_steps_per_epoch=0 epochs=1```
- Load compiled checkpoint back: ```tune run full_finetune_single_device --config recipes/configs/llama2/7B_full_single_device.yaml device=cuda:4 compile=True``` (after updating appropriate paths in the config)


torchtune/utils/_compile_utils.py
```diff
@@ -0,0 +1,24 @@
...
+
+	Args:
+		model (nn.Module): model to wrap with compile.
+    """
+    model = torch.compile(model)
+    model._register_state_dict_hook(_consume_torch_compile_prefix)
+    return model
+
+
+def _consume_torch_compile_prefix(model, state_dict, *args, prefix="_orig_mod.", **kwargs):
```
RdoubleA, line 19: don't need args or kwargs?

torchtune/utils/_compile_utils.py
```diff
@@ -0,0 +1,24 @@
...
+	register a state_dict post hook that allows state_dicts produced
+	with torch.compile training to behave as regular eager mode models.
+	In particular, it strips away a torch.compile specific prefix
+	added to the state_dict by torch.compile.
+
+	Args:
+		model (nn.Module): model to wrap with compile.
+    """
+    model = torch.compile(model)
+    model._register_state_dict_hook(_consume_torch_compile_prefix)
```
RdoubleA, line 15: nit: maybe rename to `_remove_torch_compile_prefix`

torchtune/utils/_compile_utils.py
rohan-varma: Kind of unfortunate but don't know how to work around this. The reason I have it is because this runs as a state_dict hook, so nn.Module implementation invokes these with additional arguments (such as the model, maybe additional metadata) that we don't need or use. So I just hide these in args, kwargs. Maybe could call them (unused_args, unused_kwargs) to make it more explicit?

skcoirz:
looks good. Any testing work I can help with? I can follow up on the loss curve comparison if it works for you?

recipes/configs/llama2/7B_lora_single_device.yaml
```diff
@@ -70,6 +70,7 @@ loss:
 epochs: 1
 max_steps_per_epoch: null
 gradient_accumulation_steps: 1
+compile: False
```
ebsmothers, line 73: Why default to false here? (Same q for QLoRA)

torchtune/utils/_compile_utils.py
ebsmothers: A couple questions here: 

(1) does prefix ever vary? If this is specific to compile do we actually need to parametrize it? (And if we do, are we able to move it out from in between args and kwargs)
(2) Anything else we have to watch out for here in terms of composability with other functionality (specifically other state dict hooks I guess)? I know you mentioned there are still some gaps with distributed, is this one of the rough edges there?

recipes/full_finetune_single_device.py
```diff
@@ -310,6 +317,11 @@ def train(self) -> None:
         The core training loop. Supports training on subsets of the dataset using the
         ``max_steps_per_epoch``.
         """
+
+        if self._model_compile:
+            log.info(
+                "NOTE: torch.compile is enabled and model is compiled in first forward. Expect a relatively slow first iteration."
```
ebsmothers, line 323: Roughly how much is the upfront cost here? (I know you mentioned you'll give more details soon so fine to wait until then)

ebsmothers (commented):
This is awesome! Looking forward to seeing more details on the speedups

recipes/configs/llama2/7B_lora_single_device.yaml
rohan-varma: I know I have true for one of the configs but that's mostly for showcase atm. IMO, we should ship compile with False by default. I don't think compile is a fully baked in stable feature yet, i.e. the integration with distributed is not flushed out, there are some concerns around things like long compile time and the like. Even if we comprehensively verify accuracy, I'd be hesitant to ship it on by default as users will likely modify recipes etc and could use features that don't synergize well with compile.

recipes/full_finetune_single_device.py
rohan-varma: Seeing 100-300s depending on the recipe. For example, QLoRA seems to need longer to compile, possibly due to more compiler passes for fusing quantization kernels (unsure)

torchtune/utils/_compile_utils.py
rohan-varma: `prefix` should remain constant modulo BC breaking changes from pytorch.

Re: composability with other state_dict hooks, the main thing to watch out for is hook execution order. If there is a wrapper such as DDP, that also adds prefix to modules, and this hook runs before DDP's state_dict hook, it won't remove any of the prefixes since the name begins with DDP's prefix and not torch compile prefix.

If we hit this issue though, checkpointing issues will be raised, so we'd be aware that way.

recipes/full_finetune_single_device.py
```diff
@@ -224,6 +226,12 @@ def _setup_model(
         # Validate model was loaded in with the expected dtype.
         utils.validate_expected_param_dtype(model.named_parameters(), dtype=self._dtype)
         log.info(f"Model is initialized with precision {self._dtype}.")
+
+        # Compile model, if enabled.
+        if compile_model:
+            log.info("Compiling model with torch.compile...")
+            print("RV: COMPILING", flush=True)
```
rohan-varma, line 233: rm

.github/workflows/recipe_test_multi_gpu.yaml
```diff
@@ -21,6 +21,10 @@ defaults:
 jobs:
   recipe_test_multi_gpu:
     runs-on: linux.8xlarge.nvidia.gpu
+    include:
+      - runs-on: linux.8xlarge.nvidia.gpu
+        compiler: gcc
+        gcc: 11
```
rohan-varma, line 27: Needs this for torch.compile backend with inductor to work.

.github/workflows/recipe_test_multi_gpu.yaml
```diff
@@ -24,6 +24,9 @@ jobs:
     strategy:
       matrix:
         python-version: ['3.8', '3.9', '3.10', '3.11']
+        include:
+          - compiler: gcc
+            gcc: 11
```
ebsmothers, line 29: Just curious, why do we only need this for the multi-GPU workflow?

README.md
```diff
@@ -176,6 +176,13 @@ Check out `tune --help` for all possible CLI commands and options.
 
 ---
 
+#### Support for torch.compile
```
ebsmothers, line 179: So I think this is a useful callout, but also idk if putting it as a subheader under "Getting Started" 100% makes sense to me. I almost wonder if we should have a separate section for supported features. Then we can put compile, distributed, quantization, other low-memory features in there, document each and give caveats. Not a blocker for this particular PR anyways, but lmk if that makes sense to you or not.

recipes/configs/llama2/7B_full_single_device.yaml
```diff
@@ -56,6 +56,7 @@ loss:
   _component_: torch.nn.CrossEntropyLoss
 max_steps_per_epoch: null
 gradient_accumulation_steps: 1
+compile: False
```
ebsmothers, line 59: Probably need to add this to any other configs that will run on single device recipes, right? (I guess that's just Mistral.) Otherwise they will error out due to missing config field

torchtune/utils/_compile_utils.py
```diff
@@ -0,0 +1,38 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
```
ebsmothers, line 1: Can we add a quick unit test for this file?

torchtune/utils/_compile_utils.py
```diff
@@ -0,0 +1,38 @@
...
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+import torch
+import torch.nn as nn
```
ebsmothers, line 8: Why not `from torch import nn`?

ebsmothers (approved):
These speedups are awesome! Couple comments around configs, testing and documentation, but no major concerns from my side. Modulo those comments and green CI this looks good to me!

recipes/configs/llama2/7B_full_single_device.yaml
rohan-varma: Yeah, can check mistral. Though IMO "a config that we support on single device" --> implies that it runs in CI today (can of course add additional ones to CI).

README.md
rohan-varma: Yeah agreed - we need to figure this out. README isn't in the best shape right now.

recipes/configs/llama2/7B_lora_single_device.yaml
drisspg: Do you think it would be to okay to ship local regions with compile turned on by default? In particular QLoRa as is shown in your benchmark very much benefits from torch.compile

rohan-varma: So unfortunately compile + QLoRA only works on latest nightly. The tradeoff here is QLoRA being runnable OOTB with stable version of torch (cannot be done with compile), versus requiring some extra lift to set up to run because compile is on by default. IMO due to our principles around being easy to use and removing all barriers to usage where possible, I'd vote for the former and document the changes required to enable QLoRA (i.e. upgrading to nightly). cc @kartikayk @ebsmothers for their thoughts though

drisspg: Totally make sense! I think that is likely the correct trade off

PR #604, opened 2024-03-28
[Perf Tools][1/n] Support pytorch profiler

SLR722:
#### Context
We plan to provide a centralized and flexible perf logging utility for torchtune users. Pytorch profller is an useful tool to help user understand the perf bottleneck and optimize training performance.  User manual: https://pytorch.org/docs/stable/profiler.html

Expected UX: 
- User can config to turn on pytorch profiler and the tracing file output path in yaml config 
- Hardcoded other profiler params to avoid bloating the config 

#### Changelog
- Add pytorch profiler support to all available recipes 
- Add pytoch profilers configs to all available yaml configs

#### Test plan
e2e test 
- tune --nnodes 1 --nproc_per_node 1 full_finetune_single_device --config llama2/7B_full_single_device
- tune --nnodes 1 --nproc_per_node 8 full_finetune_distributed --config llama2/7B_full
- tune --nnodes 1 --nproc_per_node 1 lora_finetune_single_device --config llama2/7B_lora_single_device 
- - tune --nnodes 1 --nproc_per_node 1 lora_finetune_single_device --config llama2/7B_qlora_single_device 
- tune --nnodes 1 --nproc_per_node 8 lora_finetune_distributed --config llama2/7B_lora 
= tune --nnodes 1 --nproc_per_node 8 full_finetune_distributed --config llama2/13B_full
- tune --nnodes 1 --nproc_per_node 8 lora_finetune_distributed --config llama2/13B_lora
all finished successfully 

example lgging tracing
Screenshot 2024-03-27 at 5 10 45 PM


recipes/configs/mistral/7B_lora.yaml
```diff
@@ -71,3 +71,7 @@ metric_logger:
   log_dir: ${output_dir}
 output_dir: /tmp/Mistral-7B-v0.1
 log_every_n_steps: null
+
+# Pytorch Profiling
+enable_torch_profiler: False
```
rohan-varma, line 76: Not sure on thoughts of how scalable adding flags to every single recipe is... @RdoubleA any thoughts?

recipes/full_finetune_distributed.py
```diff
@@ -402,69 +409,77 @@ def train(self) -> None:
         # zero out the gradients before starting training
         self._optimizer.zero_grad()
 
+        torch_profiler = self._torch_profiler
```
rohan-varma, line 412: not needed, just use self._torch_profiler?

recipes/lora_finetune_distributed.py
```diff
@@ -304,11 +311,11 @@ def _setup_model(
             sync_module_states=True,
             # Initialize empty modules on all non-zero ranks
             param_init_fn=(
-                lambda module: module.to_empty(
-                    device=torch.device("cuda"), recurse=False
+                lambda module: (
```
rohan-varma, line 314: These lint changes are unfortunate...we should really sync up the linting with what VSCode does.

torchtune/utils/perf_tracker.py
```diff
@@ -0,0 +1,52 @@
...
+        enabled (bool): Whether enable pytorch profiler or not.
+        output_file_path (str): Tracing file output path.
+        is_rank_zero (Optional[bool]): Whether the current rank is zero.
+
+    Returns:
+        ContextManager: pytorch profiler context manager
+    """
+
+    def trace_handler(prof) -> None:
+        prof.export_chrome_trace(output_file_path)
```
rohan-varma, line 36: Let's not crash the training job in case there are issues emitting the trace IMO. Let's catch errors here and simply log them.

torchtune/utils/perf_tracker.py
```diff
@@ -0,0 +1,52 @@
...
+    """
+    Utility to trace the code with pytorch profiler.
+    check the user manual: https://pytorch.org/docs/stable/profiler.html for more details.
+    The schedule for this profiler is wait 5 steps, warmup 5 steps, trace 5 steps
+    Note: Enable pytorch profiler may casue performance overhead.
+
+    Args:
+        enabled (bool): Whether enable pytorch profiler or not.
+        output_file_path (str): Tracing file output path.
+        is_rank_zero (Optional[bool]): Whether the current rank is zero.
```
rohan-varma, line 29: Clarify that this argument would disable the profiler if false.

torchtune/utils/perf_tracker.py
```diff
@@ -0,0 +1,52 @@
...
+        profile(
+            activities=[
+                torch.profiler.ProfilerActivity.CPU,
+                torch.profiler.ProfilerActivity.CUDA,
+            ],
+            schedule=torch.profiler.schedule(wait=5, warmup=5, active=5, repeat=1),
+            on_trace_ready=trace_handler,
+            record_shapes=True,
+            profile_memory=True,
+            with_stack=True,
```
rohan-varma, line 48: Let's set this to False as it causes increased overhead.

torchtune/utils/perf_tracker.py
```diff
@@ -0,0 +1,52 @@
...
+    return (
+        profile(
+            activities=[
+                torch.profiler.ProfilerActivity.CPU,
+                torch.profiler.ProfilerActivity.CUDA,
+            ],
+            schedule=torch.profiler.schedule(wait=5, warmup=5, active=5, repeat=1),
+            on_trace_ready=trace_handler,
+            record_shapes=True,
+            profile_memory=True,
```
rohan-varma, line 47: Can remove this for now as well.

torchtune/utils/perf_tracker.py
```diff
@@ -0,0 +1,52 @@
...
+    def trace_handler(prof) -> None:
+        prof.export_chrome_trace(output_file_path)
+
+    return (
+        profile(
+            activities=[
+                torch.profiler.ProfilerActivity.CPU,
+                torch.profiler.ProfilerActivity.CUDA,
+            ],
+            schedule=torch.profiler.schedule(wait=5, warmup=5, active=5, repeat=1),
```
rohan-varma, line 44: Would wait ~100 iterations for the training speed to stabilize.

torchtune/utils/__init__.py
```diff
@@ -44,6 +44,7 @@
     memory_stats_log,
     set_activation_checkpointing,
 )
+from .perf_tracker import pytorch_profiler_or_nullcontext
```
rohan-varma, line 47: Let's prefix filename with _ which is a convention we're starting to push.

torchtune/utils/perf_tracker.py
```diff
@@ -0,0 +1,52 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
```
rohan-varma, line 1: Would call this file `_profiler_utils` for now and we can evolve as needed.

rohan-varma (commented):
Thank youfor working on this! Just a couple of comments after which we should be good to go.

kartikayk (changes requested):
Thanks for putting this together @SLR722! Glad to see progress on this front.

Generally, I'm not a fan of this design. It's too intrusive for something which should be a composable utility. We should design this as one of the existing utilities (eg: Metric Logger or Checkpointer) and then show case how to add this as a component to one of the recipes. Unlike metric logger and checkpointer, not every recipe needs to support this OOTB. Anyone who needs this can clone the recipe locally and add it for their experiments.

Concretely speaking, I would like to see something like this:
- Profiling class that provides a bunch of capabilities (eg: profiling memory, tools for tracking QPS, maybe memory stats that we have currently). All of the features can then be cleanly abstracted into this one class.
- For recipes that want to integrate this, just add it as a component with the yaml pointing to this class and the component being initialized in the init. This also means that instead of adding a new flag everytime, these are organized within this class.
- Add it to one of the recipes to show an example of how this is done.

SLR722:
> Thanks for putting this together @SLR722! Glad to see progress on this front.
> 
> Generally, I'm not a fan of this design. It's too intrusive for something which should be a composable utility. We should design this as one of the existing utilities (eg: Metric Logger or Checkpointer) and then show case how to add this as a component to one of the recipes. Unlike metric logger and checkpointer, not every recipe needs to support this OOTB. Anyone who needs this can clone the recipe locally and add it for their experiments.
> 
> Concretely speaking, I would like to see something like this:
> 
> * Profiling class that provides a bunch of capabilities (eg: profiling memory, tools for tracking QPS, maybe memory stats that we have currently). All of the features can then be cleanly abstracted into this one class.
> * For recipes that want to integrate this, just add it as a component with the yaml pointing to this class and the component being initialized in the init. This also means that instead of adding a new flag everytime, these are organized within this class.
> * Add it to one of the recipes to show an example of how this is done.

Thanks for the suggestions @kartikayk! It makes a lot of sense to me! There are 3 things I'd like to clarify with you before proceeding. 
- For the profiling class, shall we make a single class for all the capabilities as functions or separate classes shared a single abstracted interface? 
- it's inevitable to embed some profiling code into the recipe. Like sandwich the code block with time.perf_counter() for QPS etc. Is that acceptable?
- Shall we create a dedicated perf tools showcase recipe without interact with the existing recipes?

recipes/lora_finetune_distributed.py
ebsmothers: I still don't really understand this, if you've installed pre-commit hooks then lint when you commit locally should be the same as the linter on CI, no?

torchtune/utils/perf_tracker.py
ebsmothers: What about the case of e.g. some quick and dirty profiling? Like I don't wanna run a whole 100 steps, just wanna run 10-20 to get a quick trace. Or is your point that we should discourage that

recipes/configs/llama2/13B_full.yaml
```diff
@@ -83,3 +83,7 @@ metric_logger:
   log_dir: ${output_dir}
 output_dir: /tmp/alpaca-llama2-finetune
 log_every_n_steps: null
+
+# Pytorch Profiling
+enable_torch_profiler: False
+output_file_path: ./torchtune_perf_tracing.json
```
ebsmothers, line 89: Not unique to this change but I feel we are starting to accumulate too many potentially disparate file paths here (logger, checkpoint output files, now profiling file path). As of now none of these are aware of each other, but I don't see why we shouldn't just have a single output directory for all this stuff.

recipes/lora_finetune_single_device.py
```diff
@@ -384,55 +390,65 @@ def train(self) -> None:
...
+                for idx, batch in enumerate(pbar := tqdm(self._dataloader)):
+                    if (
+                        self.max_steps_per_epoch is not None
+                        and (idx // self._gradient_accumulation_steps)
+                        == self.max_steps_per_epoch
+                    ):
+                        break
+
+                    if self._enable_torch_profiler:
+                        torch_profiler.step()
```
ebsmothers, line 410: One thing to watch out for here is the interaction of this with grad accumulation. E.g. if I set gradient_accumulation_steps=128 I will not even catch a single optimizer step with the profiler.

rohan-varma: This is a great point, good catch!

rohan-varma:
Agree with most of @kartikay s points, though I do see reasoning for having the profiler as a standalone utility over the class that provides a large set of individual capabilities. The reason is that profiling is almost always the first step for users to debug their code and often the first thing that's enabled for perf-related work. Having the torch.profiler easily enabled / integrated is useful as the user doesn't have to learn a torchtune specific component if they only want torch.profiler profiling, and the user doesn't have to use a larger abstraction that covers features that are not needed if they only want profiling.

cc @kartikayk @SLR722

kartikayk:
@SLR722 fair questions - I'd like to understand how some of the other solutions in this space approach this problem? I definitely don't want to have recipes with a bunch of perf_counters all over the place. I think our logging (both for the logger and metric logger) are already quite intrusive. For example we have a bunch of ```Is_rank_zero``` flags all over the place. I'm not saying this PR needs to address this, but I think these are reasonably related so I'd love to figure out a better solution. I want reasability of these recipes to still be high post these changes.

@rohan-varma what you say makes sense, but that seems more like a tutorial than a utility in its own? For example, I think we would just showcase how to use https://pytorch.org/tutorials/recipes/recipes/profiler_recipe.html with torchtune and this can be through a tutorial? or do you think we need additional code here?

PR #608, opened 2024-03-28
[WIP] Add a Perf Monitor for metric tracking.

rohan-varma:
#### Context
- In this PR, we introduce `TunePerfMonitor`, a utility class for tracking metrics across training. This class is meant to be flexible in the actual metrics that users can track, metrics will be defined and tracked by user (see example in recipe).
- Please see "LIMITATIONS" section in the code for limitations of this tracker in its current state.
-
#### Changelog
- ...

#### Test plan
- Unittests
- In full finetune single device, I've integrated a few basic metrics - average seconds per iter and max memory allocated post backward pass, and surface them to wandB. Charts look as follows:
image


SLR722:
Thanks for making this RFC PR @rohan-varma! Share my 2c
- Seems there are many functions support different tracking use cases. Shall we begin from less generic class design and focus on the priority tracking need we have? (such as QPS, memory stats etc)
- Based on the discussion in https://github.com/pytorch/torchtune/pull/604. I'm consider 2 tracking cases here
  - Necessary in almost every runs with minimum performance overhead (such as training QPS, memory stats etc). For this case, we can use the perfMonitor class as you proposed in this PR 
  - Not necessary in every run and only needed when debugging and the tracing will have perf overhead such as torch profiler, memory snapshot (https://pytorch.org/blog/understanding-gpu-memory-1/). We can consider make them as component and bring them into recipe when necessary. I think a better state of this it to enable them by 'debug' flag or calling them during OOM

cc: @kartikayk

torchtune/utils/perf_utils.py
```diff
@@ -0,0 +1,117 @@
...
+        End tracking metric given by ``str`` and compute and store the metric value.
+        """
+        # TODO: only MEAN reduction supported right now, and end_record is not the best place to specify it.
+        # Don't want to specify it in emit_metric either, as then we'll have to store metric for every
+        # iteration, possibly too much space used. Could have a solution where user a priori defines the metric
+        # they want to track and the reduction type for it.
+        assert (
+            metric_name in self._inflight
+        ), f"Not tracking metric {metric_name} for this iteration!"
+        cpu_elapsed_s = time.perf_counter() - self._inflight.pop(metric_name)
```
SLR722, line 96: I think the start_record, end_record design here only supports  iteration time related logging? How it fit into pytorch profiler, memory snapshot etc as well?

rohan-varma: @SLR722 , `log_metric` is intended to support the use case of emitting a metric that's only computed with a single data point / at a single point in time. 

> fit into pytorch profiler, memory snapsho

So IMO `TunePerfMonitor` is a way to track and retrieve metrics. Profiler and snapshot are related but seperate tools. I currentl don't see the value in integrating all of these into a single class, though I'm open to feedback and understand the motivation for that design. Could also be open to a solution of designing something like a `snapshot_memory` API within this perf monitor to take the memory snapshot, but I do feel like that it takes away from the simplicity of this class.

rohan-varma:
@SLR722 It makes sense to me to make the profiler and memory snapshot as individual standalone components, though I don't see why these should be together within the same component when they are quite different, have different APIs and are meant to debug different issues. What issues come up if we build these as two separate components that don't interfere with each other and can be enabled one, the other, or both?

SLR722:
> @SLR722 It makes sense to me to make the profiler and memory snapshot as individual standalone components, though I don't see why these should be together within the same component when they are quite different, have different APIs and are meant to debug different issues. What issues come up if we build these as two separate components that don't interfere with each other and can be enabled one, the other, or both?

Agree that we should build 2 separate components for profiler and memory snapshot

rohan-varma:
@SLR722 Seems reasonable to me. Again, I don't fully see the need to add all of these into a singular component, but do see value in terms of having a single entrypoint to manage all performance related things. Would like thoughts from @kartikayk and @ebsmothers

kartikayk:
Sure, we don't need to add all of these into a single component if it doesn't make sense to. But I'd like to preserve the design principle around having self-contained components which can be pulled into any recipe. Make as many components as we need (within reason).

recipes/full_finetune_single_device.py
```diff
@@ -341,24 +343,31 @@ def train(self) -> None:
                 # Check if this is the norm or not
                 pbar.set_description(f"{curr_epoch+1}|{idx+1}|Loss: {loss.item()}")
                 if self.total_training_steps % self._log_every_n_steps == 0:
+                    log_dict = {
+                        "loss": loss.item(),
+                        "lr": self._optimizer.param_groups[0]["lr"],
+                        "gpu_resources": torch.cuda.memory_allocated(),
+                        "avg_it_s": perf_monitor.get_metric_val("avg_it_s", 0),
```
skcoirz, line 350: where is perf_monitor defined?

skcoirz (commented):
hey Rohan, sorry about the stupid questions. I'm new here. What doest avg_it_s mean here? Could you share a few other metrics we can use with this monitor? Thanks!

recipes/full_finetune_single_device.py
RdoubleA: probably means this?
```suggestion
                        "avg_it_s": monitor.get_metric_val("avg_it_s", 0),
```

recipes/full_finetune_single_device.py
```diff
@@ -341,24 +343,31 @@ def train(self) -> None:
...
-                            "gpu_resources": torch.cuda.memory_allocated(),
-                        },
+                        log_dict,
                         step=self.total_training_steps,
                     )
 
                 loss = loss / self._gradient_accumulation_steps
                 loss.backward()
+                # Snapshot the max memory allocated after backward.
+                perf_monitor.log_metric("max_cuda_mem_allocated_gb", torch.cuda.max_memory_allocated() / 1e9)
```
RdoubleA, line 361:
```suggestion
                monitor.log_metric("max_cuda_mem_allocated_gb", torch.cuda.max_memory_allocated() / 1e9)
```

recipes/full_finetune_single_device.py
```diff
@@ -313,13 +313,15 @@ def train(self) -> None:
         # zero out the gradients before starting training
         self._optimizer.zero_grad()
 
+        monitor = utils.perf_utils.TunePerfMonitor()
```
RdoubleA, line 316: nit: I would just import this directly instead of using the full filepath here
```suggestion
        monitor = TunePerfMonitor()
```

rohan-varma:
Thanks @skcoirz @RdoubleA for your comments on the PR, sorry that the integration into the recipe is currently in a messy state. Currently mostly looking for feedback on the API and class itself, but realize this will be easier to contextualize with a clearer example of it in the recipe. Adding that now

SLR722:
LGTM! Please fix the tests and clean up the code and then we are good to go!

PR #610, opened 2024-03-28
[test] fix download test when os env is set

skcoirz:
#### Context
- when there is local env-var set up, the test will fail.
- instead of letting it fail, let's skip it for local developers when env-var is set. 
- CI will have this covered.

#### Changelog
- updated hf download test

#### Test plan
-  pytest tests/torchtune/_cli/test_download.py  and confirm it passed

### DIFF
--- a/tests/torchtune/_cli/test_download.py
+++ b/tests/torchtune/_cli/test_download.py
@@ -11,6 +11,7 @@ import tempfile
 from unittest.mock import patch
 
 import pytest
+import os
 
 from tests.common import TUNE_PATH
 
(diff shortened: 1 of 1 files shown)


joecummings:
Awesome catch, thanks!

This issue should be solved in #586 as the test changes to have better error handling for these exceptions and no longer actually executes a download call.

skcoirz:
@joecummings , wow that's a big refactoring. Like it! It feels #586 would need some time for review? PR here is quite trivial. Maybe we can land it to silent my local tests? (now it's a bit annoying.. every time when I run all pytests, the download test will fail...) Would like to hear your preference. thanks! :)

joecummings:
Fixed in #586 - thanks for your help on this @skcoirz !

PR #619, opened 2024-03-29
Inference

kartikayk:
## Context

Our current story for inference is sub-optimal. This PR makes an attempt at fixing this. The code for inference is heavily inspired by [gpt-fast](https://github.com/pytorch-labs/gpt-fast) though it's missing some of the functionality around compile. I'll add this as a follow up PR. This is basically setting up the inference to actually work.

## Changelog

- Update the way we setup KV Cache, including removing ```curr_pos``` which was really confusing and never made any sense to me. I replace this with ```input_pos``` (naming consistent with gpt_fast) which does exactly what you expect it to i.e. its a tensor with the current position. When we first start inference, this includes positions for all tokens in the prompt since the K and V tensors for these need to be correctly computed.
- Update all components and tests
- Remove the current GenerationUtils class and logit_transforms file. Replace these with standa-alone utilities for generation.

## Test plan
- All unit tests and recipe tests pass
- Output generation is sensible for a 13B model

image

Comparison with gpt-fast: Without compile, our generation speed is on par with gpt-fast. Adding compile support is beyond the scope of this PR

gpt-fast:

image

torchtune:

image


torchtune/modules/attention.py
```diff
@@ -149,10 +151,12 @@ def forward(
             - Return the attention weights
             - Make application of positional embeddings optional
         """
-
         # input has shape [b, s, d]
         bsz, seq_len, _ = x.shape
 
+        # self.wqkv.weight.data = torch.cat([self.q_proj.weight, self.k_proj.weight, self.v_proj.weight ])
```
joecummings, line 157: Old code?

torchtune/modules/attention.py
```diff
@@ -166,6 +170,8 @@ def forward(
         k = self.k_proj(x)
         v = self.v_proj(x)
 
+        # pdb.set_trace()
```
joecummings, line 173: 🙃

torchtune/modules/kv_cache.py
```diff
@@ -17,50 +17,38 @@ class KVCache(nn.Module):
...
-        Updates the kv-cache at curr_pos with the given k_val and v_val.
-
-        Args:
-            bsz (int): Batch size.
-            seq_len (int): Sequence length.
-            curr_pos (int): Current position in sequence.
-            k_val (Tensor): New k value.
-            v_val (Tensor): New v value.
+    def update(self, input_pos, k_val, v_val) -> Tuple[Tensor, Tensor]:
+        # input_pos: [S], k_val: [B, H, S, D]
```
joecummings, line 46: Love this comment

recipes/alpaca_generate.py
joecummings, line 1: WOOOOOOOOOOO

recipes/configs/generate.yaml
```diff
@@ -0,0 +1,39 @@
+# Config alpaca_generate.py recipe
```
joecummings, line 1:
```suggestion
# Config generate.py recipe
```

recipes/configs/generate.yaml
```diff
@@ -0,0 +1,39 @@
...
+#
+# To launch, run the following command from root:
+#    tune alpaca_generate \
```
joecummings, line 4: Update here, as well.

recipes/configs/generate.yaml
```diff
@@ -0,0 +1,39 @@
...
+#    model_checkpoint=<your_checkpoint_dir> \
+#    tokenizer_checkpoint=<your_tokenizer_dir>
+
+# Model arguments
+model:
+  _component_: torchtune.models.llama2.llama2_13b
+
+checkpointer:
+  _component_: torchtune.utils.FullModelHFCheckpointer
+  checkpoint_dir: /data/users/kartikayk/cpts/Llama-2-13b-hf/
```
joecummings, line 15: Remove reference to your name

recipes/generate.py
```diff
@@ -0,0 +1,176 @@
...
+        # Validate model was loaded in with the expected dtype.
+        utils.validate_expected_param_dtype(model.named_parameters(), dtype=self._dtype)
+        logger.info(f"Model is initialized with precision {self._dtype}.")
+
+        # Ensure the cache is setup on the right device
+        with self._device:
+            model.setup_caches(max_batch_size=1, dtype=self._dtype)
+        return model
+
+    def _multinomial_sample_one_no_sync(self, probs_sort):
```
joecummings, line 59: Can you leave a comment here explaining this?

recipes/generate.py
```diff
@@ -0,0 +1,176 @@
...
+        new_tokens, new_probs = [], []
+        for i in range(num_new_tokens):
+            with torch.backends.cuda.sdp_kernel(
+                enable_flash=False, enable_mem_efficient=False, enable_math=True
+            ):
+                next_token, next_prob = self._decode_one_token(
+                    model, cur_token, input_pos, temperature, top_k
+                )
+                input_pos += 1
+                new_tokens.append(next_token.clone())
```
joecummings, line 121: Do we care about performance here? How expensive is the double clone operation?

joecummings:
Should we also add a test ensuring that this does in fact speed up inference? Something like running the same inference twice, once with cache, and once without and checking the time difference?

recipes/generate.py
```diff
@@ -0,0 +1,176 @@
...
+        # Ensure the cache is setup on the right device
+        with self._device:
+            model.setup_caches(max_batch_size=1, dtype=self._dtype)
+        return model
+
+    def _multinomial_sample_one_no_sync(self, probs_sort):
+        q = torch.empty_like(probs_sort).exponential_(1)
+        return torch.argmax(probs_sort / q, dim=-1, keepdim=True).to(dtype=torch.int)
+
+    def _logits_to_probs(self, logits: torch.Tensor, temperature: float, top_k: int):
```
joecummings, line 63: We're pretty inconsistent with where our generation code lives. I'd be in favor of putting it all in one file.

recipes/configs/generate.yaml
```diff
@@ -0,0 +1,39 @@
...
+  ]
+  recipe_checkpoint: null
+  output_dir: /tmp/llama2
+  model_type: LLAMA2
+
+device: cuda
+dtype: bf16
+seed: 1234
+
+temperature: 0.8
```
joecummings, line 29: Can you leave a comment on this PR explaining how you chose these hyperparams?

recipes/configs/generate.yaml
```diff
@@ -0,0 +1,39 @@
...
+top_k: 300
+
+# Tokenizer arguments
+tokenizer:
+  _component_: torchtune.models.llama2.llama2_tokenizer
+  path: /tmp/llama2/tokenizer.model
+
+# Generation arguments
+prompt: "Hello, my name is"
```
joecummings, line 38: Maybe it would make sense to have this be a common default prompt. (Coming back to update this comment with examples)

recipes/generate.py
kartikayk: I don't want to generalize this code prematurely. I think for now I expect this to be limited to this recipe, but we can generalize as a follow up if that makes sense

recipes/generate.py
kartikayk: The memory impact of this is quite minimal. These are like 300 or so ints

recipes/configs/generate.yaml
```diff
@@ -0,0 +1,32 @@
+
+# Model arguments
+model:
+  _component_: torchtune.models.llama2.llama2_13b
```
rohan-varma, line 4: Should we call it generate_13b?

recipes/configs/generate.yaml
```diff
@@ -0,0 +1,32 @@
...
+
+checkpointer:
+  _component_: torchtune.utils.FullModelHFCheckpointer
+  checkpoint_dir: /tmp/Llama-2-13b-hf/
+  checkpoint_files: [
+    pytorch_model-00001-of-00003.bin,
+    pytorch_model-00002-of-00003.bin,
+    pytorch_model-00003-of-00003.bin
+  ]
+  recipe_checkpoint: null
```
rohan-varma, line 14: Is this pretty much always going to be null? We don't really need to pass in recipe state into generate, IIUC?

torchtune/modules/attention.py
```diff
@@ -190,20 +191,18 @@ def forward(
...
+        q = self.pos_embeddings(q, input_pos)
+        k = self.pos_embeddings(k, input_pos)
 
         # [b, n_h, s, h_d]
         q = q.transpose(1, 2)
         k = k.transpose(1, 2)
         v = v.transpose(1, 2)
 
+        # Update key-value cache
+        if self.kv_cache is not None:
```
rohan-varma, line 203: Just for my understanding, any benefit in updating kv-cache post transpose? Does it enable better access patterns or the like?

recipes/configs/generate.yaml
kartikayk: Oh yeh, good point. I should remove this

recipes/configs/generate.yaml
kartikayk: Not really - the model itself doesn't change during inference. The inference script is responsible for setting up the caches and the calling ```eval()``` on the model to disable all of the stochastic operations.

torchtune/modules/attention.py
kartikayk: It's a good question - let me think a bit about it

torchtune/utils/generation.py
```diff
@@ -163,8 +163,9 @@ def generate(
         eos_reached = torch.zeros(batch_size, dtype=torch.bool, device=device)
         for cur_pos in range(min_prompt_len, total_gen_len):
             input_ids = tokens[:, prev_pos:cur_pos]
+            input_pos = torch.arange(prev_pos, cur_pos)
             if incremental_decode:
-                outputs = self.decoder_lm(input_ids, curr_pos=prev_pos)
+                outputs = self.decoder_lm(input_ids, input_pos=input_pos)
             else:
                 outputs = self.decoder_lm(input_ids)
             if logits_accessor:
```
ebsmothers, line 171: I know it's not part of these changes, but I find the name "accessor" a bit confusing tbh. Is the idea here to do some sort of slicing along the vocab dim or something? What is the difference between this and logits_transforms? Is it just that one is applied before softmax and the other is applied after?

recipes/generate.py
ebsmothers: A couple questions here: 

1) Does this obviate the need for `logits_transforms.py`? Seems we are now handling all of that in the recipe. (As a follow-up, if we are gonna keep it, can we at least drop the `LogitsTransform` ABC? It's literally just a `Callable[FloatTensor, FloatTensor]`)
2) I notice we are missing top p sampling, which we otherwise have support for. Any particular reason for omitting it?

recipes/generate.py
```diff
@@ -0,0 +1,176 @@
...
+
+    def _prefill(
+        self,
+        model: nn.Module,
+        x: torch.Tensor,
+        input_pos: torch.Tensor,
+        temperature: float,
+        top_k: int,
+    ) -> torch.Tensor:
+        # input_pos: [B, S]
```
ebsmothers, line 86: One thing that's confusing to me.. do we support batch generation? The way we pass the prompt seems to indicate that we don't, but then comments like this would indicate that we do.

torchtune/modules/transformer.py
```diff
@@ -113,50 +125,83 @@ def __init__(
...
         self.output = output
+        self.max_seq_len = max_seq_len
+        self.num_heads = num_heads
+        self.head_dim = head_dim
+        self.causal_mask = None
 
-    def forward(
-        self, tokens: Tensor, mask: Optional[Tensor] = None, curr_pos: int = 0
-    ) -> Tensor:
+    def setup_caches(self, max_batch_size: int, dtype: torch.dtype) -> None:
```
ebsmothers, line 145: I'm curious about the choice to define this as a method on the `TransformerDecoder` class (especially with the addition of mandatory `max_seq_len`, `num_heads`, `head_dim` which are only needed for inference). Why not just define a utility method, e.g. 

```
def setup_caches(
	model: TransformerDecoder,
	max_batch_size: int,
	max_seq_len: int,
	num_heads: int,
	head_dim: int,
):
	# same body as setup_caches but with self -> model
```

Then optionally call this on the builder gated behind a flag. The slight drawback is that we do add one extra param to builders so it's not a 1:1 swap on the config side as it is now (maybe there's a way to do that though?). But honestly I think that's OK? Like if we are changing the model arch a bit for inference (which we are) there's no harm in being explicit about that.

I think the benefit is that anyone using `TransformerDecoder` directly (which should be a decent percentage of the people adding new models) does not have to worry about these extra params that are really only relevant for inference.

torchtune/utils/generation.py
ebsmothers: OK coming back to this now I'm a bit confused (prob should have read the full recipe code first). Why isn't this file deleted entirely? Seems you've covered all the generation logic in the recipe already. Or am I missing something?

recipes/generate.py
```diff
@@ -0,0 +1,176 @@
...
+        with self._device:
+            model.setup_caches(max_batch_size=1, dtype=self._dtype)
+        return model
+
+    def _multinomial_sample_one_no_sync(self, probs_sort):
+        q = torch.empty_like(probs_sort).exponential_(1)
+        return torch.argmax(probs_sort / q, dim=-1, keepdim=True).to(dtype=torch.int)
+
+    def _logits_to_probs(self, logits: torch.Tensor, temperature: float, top_k: int):
+        logits = logits / max(temperature, 1e-5)
```
ebsmothers, line 64: nit: maybe just add a value check on temperature somewhere?

torchtune/models/llama2/_component_builders.py
```diff
@@ -103,7 +94,7 @@ def llama2(
         v_proj=nn.Linear(embed_dim, num_kv_heads * head_dim, bias=False),
         output_proj=nn.Linear(embed_dim, embed_dim, bias=False),
         pos_embeddings=rope,
-        kv_cache=kv_cache,
+        kv_cache=None,
```
ebsmothers, line 97: Is this a useless arg now? Seems if we are building KV caches at the decoder level it will always be none?

torchtune/modules/kv_cache.py
```diff
@@ -17,50 +17,38 @@ class KVCache(nn.Module):
...
+        assert input_pos.shape[0] == k_val.shape[2]
 
-        Raises:
-            ValueError: if bsz is greater than the ``max_batch_size`` supported by the model
+        k_out = self.k_cache
+        v_out = self.v_cache
+        k_out[:, :, input_pos] = k_val
+        v_out[:, :, input_pos] = v_val
```
ebsmothers, line 52: If I understand correctly, is it now the case that `max_batch_size` really just means batch size? If so, maybe we should update its name

torchtune/modules/kv_cache.py
```diff
@@ -17,50 +17,38 @@ class KVCache(nn.Module):
     Args:
         max_batch_size (int): maximum batch size model will be run with
         max_seq_len (int): maximum sequence length model will be run with
-        n_kv_heads (int): number of kv heads
+        num_heads (int): number of heads. We take num_heads instead of num_kv_heads because
+            the cache is created after we've expanded the key and value tensors to have the
+            same shape as the query tensor. See attention.py for more details
```
ebsmothers, line 22: So this makes sense to me, but given that you didn't really change what we were doing on the attention side of things, how was this working properly before?

recipes/generate.py
```diff
@@ -0,0 +1,176 @@
...
+            temperature=cfg.temperature,
+            top_k=cfg.top_k,
+        )
+        seq[prompt_length] = next_token
+        input_pos = torch.tensor([prompt_length], device=self._device, dtype=torch.int)
+
+        t0 = time.perf_counter()
+        generated_tokens, _ = self._decode_n_tokens(
+            self._model,
+            next_token.view(1, -1),
```
ebsmothers, line 152: What is the view actually doing here? Isn't this just a single token?

recipes/generate.py
```diff
@@ -0,0 +1,176 @@
...
+        input_pos: torch.Tensor,
+        num_new_tokens: int,
+        temperature: float,
+        top_k: int,
+    ):
+        new_tokens, new_probs = [], []
+        for i in range(num_new_tokens):
+            with torch.backends.cuda.sdp_kernel(
+                enable_flash=False, enable_mem_efficient=False, enable_math=True
+            ):
```
ebsmothers, line 116: I assume this is for performance reasons? Might be good to add a bit of detail on why we're doing this though (is it because the math implementation is faster when seq len and batch size are small?)

recipes/generate.py
```diff
@@ -0,0 +1,176 @@
...
+        self,
+        model: nn.Module,
+        cur_token: torch.Tensor,
+        input_pos: torch.Tensor,
+        num_new_tokens: int,
+        temperature: float,
+        top_k: int,
+    ):
+        new_tokens, new_probs = [], []
+        for i in range(num_new_tokens):
```
ebsmothers, line 113: Do we always generate a fixed # of tokens? What if e.g. we get EOS before then?

recipes/generate.py
ebsmothers: Follow-up.. just a general comment on a bunch of these methods: it'd be nice to add docstrings with examples of each one (e.g. given a prompt of length n, prefill will fill the KV cache's first n elements and return token n+1, something similar for `_decode_one_token` and/or `_decode_n_tokens`)

recipes/generate.py
```diff
@@ -0,0 +1,176 @@
...
+from omegaconf import DictConfig
+
+from torch import nn
+
+from torchtune import config, utils
+
+logger = utils.get_logger("DEBUG")
+
+
+class InferenceRecipe:
```
ebsmothers, line 20: Can probably add documentation here around what's supported and what's not (e.g. explicitly call out we don't have speculative decoding, we support temperature and top-k sampling, etc.)

torchtune/utils/generation.py
kartikayk: No particular reason, was leaving this as a follow up. Let me think about this - maybe I should combine the change here

recipes/generate.py
kartikayk: There's nothing fundamentally stopping multi-sample generation. The inference recipe doesn;t currently handle this though

torchtune/modules/transformer.py
kartikayk: This is a really interesting point. 

I think you touch upon this a bit. The reason I think having a method in the class makes sense is because the behavior of forward changes depending on whether you're calling it during training or during inference and I'd rather the model be able to handle this based on its internal state (eg: causal_mask). For example, if this is a utility, then you also need to take care of passing around the causal_mask explicitly in every call when this can be easily handled by the class itself? I also think, that having the forward of the decoder just take the input and the corresponding positions is nicer than explicitly passing the mask (though that might be a bit of a personal preference).

I'm actually not too concerned about num_heads, head_dim and max_seq_len because the model already has this information as part of all of its components - it's related to the model isn't it? In fact I think this is strictly better than passing something like max_batch_size which was only used during inference :) Let me know if that makes sense.

torchtune/models/llama2/_component_builders.py
kartikayk: Hmm, yeh its default to None so I don't think it needs to exist in this call.

torchtune/modules/kv_cache.py
kartikayk: I don't know why it was called max_batch_size ever, but this seems to be the convention. Let me do some more research on this

torchtune/modules/kv_cache.py
kartikayk: I think this is the case where we haven't really run inference for the case where num_kv_heads != num_heads. If we did, this would break

recipes/generate.py
kartikayk: Currently we don't respect the EOS token. Let me add that functionality

recipes/generate.py
kartikayk: I removed the logits_transform.py file completely. For the support for top_p, I didn't really find a good reference. Maybe we can add back if its needed?

torchtune/utils/__init__.py
```diff
@@ -22,6 +22,7 @@
     validate_no_params_on_meta_device,
     wrap_fsdp,
 )
+from ._generation import generate  # noqa
```
joecummings, line 25: nit: why do you need the #noqa?

recipes/generate.py
```diff
@@ -0,0 +1,101 @@
...
+        t = time.perf_counter() - t0
+
+        logger.info(self._tokenizer.decode(generated_tokens))
+
+        tokens_generated = len(generated_tokens) - prompt.size(0)
+        tokens_sec = tokens_generated / t
+        logger.info(
+            f"Time for inference: {t:.02f} sec total, {tokens_sec:.02f} tokens/sec"
+        )
+        logger.info(f"Memory used: {torch.cuda.max_memory_allocated() / 1e9:.02f} GB")
```
joecummings, line 90: For my own understanding, why are you including this information?

joecummings (approved):
kewl!

torchtune/utils/__init__.py
ebsmothers: Pls someone add some analogous version of [this](https://github.com/pytorch/vision/blob/2c4665ffbb64f03f5d18016d3398af4ac4da5f03/setup.cfg#L14-L15) to our .flake8, then we can delete all these godforsaken noqas in our init files

kartikayk: I know you've been asking for this @ebsmothers! I'm going to be a pain and punt this to a follow up!

recipes/generate.py
kartikayk: This seems to be a standard for every generation tool. I added it for my own debugging, but then realized it might be useful for folks running this recipe as well. Anything look out of place?

PR #620, opened 2024-03-29
Configure max_seq_len in InstructDataset

RdoubleA:
## Context
ChatDataset already truncates samples based on max_seq_len - this simply adds the same functionality to InstructDataset.

## Test plan

### DIFF
--- a/tests/torchtune/data/test_data_utils.py
+++ b/tests/torchtune/data/test_data_utils.py
@@ -15,7 +15,7 @@ def test_tokenize_prompt_and_response():
     response = "I always know what I'm doing, do you?"
     prompt_length = 12
     expected_tokenized_prompt = [
-        -1,
+        0,
         12,
         4,
         2,
--- a/tests/torchtune/data/test_data_utils.py
+++ b/tests/torchtune/data/test_data_utils.py
@@ -13,8 +13,9 @@ def test_tokenize_prompt_and_response():
     tokenizer = DummyTokenizer()
     prompt = "Instruction:\nThis is an instruction.\n\nInput:\nThis is an input.\n\nResponse: "
     response = "I always know what I'm doing, do you?"
-    prompt_length = 11
+    prompt_length = 12
     expected_tokenized_prompt = [
+        -1,
         12,
         4,
         2,
@@ -44,6 +45,7 @@ def test_tokenize_prompt_and_response():
         6,
         2,
         4,
+        -1,
     ]
 
     tokenized_prompt, tokenized_label = tokenize_prompt_and_response(
--- a/tests/torchtune/data/test_data_utils.py
+++ b/tests/torchtune/data/test_data_utils.py
@@ -35,6 +35,7 @@ def test_tokenize_prompt_and_response():
         6,
         2,
         4,
+        -1,
     ]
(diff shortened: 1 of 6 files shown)


tests/torchtune/data/test_data_utils.py
```diff
@@ -13,8 +13,9 @@ def test_tokenize_prompt_and_response():
     tokenizer = DummyTokenizer()
     prompt = "Instruction:\nThis is an instruction.\n\nInput:\nThis is an input.\n\nResponse: "
     response = "I always know what I'm doing, do you?"
-    prompt_length = 11
+    prompt_length = 12
     expected_tokenized_prompt = [
+        -1,
```
ebsmothers, line 18: Maybe I'm being dumb here, but why is the first token the dummy tokenizer's eos ID?

ebsmothers (commented):
A couple questions on this change:

(1) Do we now always truncate to a fixed length? Why not make max_seq_len optional? 
(2) Kind of a follow-up to (1): do we wanna make any changes to the pad to max length collate fn in the case we are setting a fixed max length? (Strictly speaking we don't need to, it'll just become a no-op, but could cause some potential confusion.)
(3) Is Alpaca the only instruct dataset we wanna configure max seq len for?

tests/torchtune/data/test_data_utils.py
RdoubleA: huh why did this pass

RdoubleA:
> (1) Do we now always truncate to a fixed length? Why not make max_seq_len optional?

Yeah good point, thinking to make optional for InstructDataset and a default value for alpaca (cc @pbontrager)

> (2) Kind of a follow-up to (1): do we wanna make any changes to the pad to max length collate fn in the case we are setting a fixed max length? (Strictly speaking we don't need to, it'll just become a no-op, but could cause some potential confusion.)

mm I'm not inclined to, I think this logic would be unnecessarily complicated. It's generally understood that we won't pad if a sample is already at max_seq_len.

> (3) Is Alpaca the only instruct dataset we wanna configure max seq len for?

Yes - I have not seen any guidance for the other datasets so I will leave it optional.

tests/torchtune/datasets/test_chat_dataset.py
```diff
@@ -146,15 +147,18 @@ def test_get_item(self, mock_load_dataset, template, dialogue):
...
                 6,
                 11,
                 10,
                 1,
+                6,
                 -1,
             ]
         ]
-        prompt_lengths = (14, 4)
+        prompt_lengths = (15, 5)
```
ebsmothers, line 161: Why is stuff in chat dataset changing, just cause of dummy tokenizer update? Also why is it not incremented by two if you're adding both bos and eos? (I'm prob missing something obvious here)

ebsmothers (approved):
One more minor question, but looks good!

tests/torchtune/datasets/test_chat_dataset.py
RdoubleA: yes it's the dummy tokenizer update. the expected prompt and label are actually two turns, so going from (14, 4) -> (15, 5) is actually incrementing by 2

Issue #366, opened 2024-02-10
Potential future tokenization issue in Alpaca


laurencer:
In the [Alpaca dataset](https://github.com/pytorch-labs/torchtune/blob/48626d19d2108f92c749411fbd5f0ff140023a25/torchtune/datasets/alpaca.py#L112C1-L123C14) the prompt and input text is encoded in the following way:

```python
        encoded_prompt = self._tokenizer.encode(
            text=prompt, add_bos=True, add_eos=False
        )
        encoded_prompt_with_response = self._tokenizer.encode(
            text=prompt_with_response, add_bos=True, add_eos=True
        )
        labels = encoded_prompt_with_response.copy()


        if not self.train_on_input:
            labels[: len(encoded_prompt)] = [CROSS_ENTROPY_IGNORE_IDX] * len(
                encoded_prompt
            )
```

This may cause a subtle issue where `sentencepiece` tokenization is context-dependent and so the tokens generated for the `prompt` (i.e. `encoded_prompt`) may be different than the tokens generated for the same section of input in `encoded_prompt_with_response`. If this happens, the masking may be ever so slightly off (it also may effect generation in weird ways if encoding is done slightly differently).

Now this doesn't arise for the default prompt templates since they use newline delimiters (these always get encoded to `13` in the default llama2 tokenizer) but it would be a problem if someone where to copy this as the basis for their own 😄 

E.g. if someone wanted to use a custom prompt template like the following; then the tokenization could vary (FWIW I've run into this issue personally when using sentence-piece and trying to train only on inputs).

<pre>
# assume the provided instruction is "test" and the rest is the prompt template

>>> tokenizer.encode("&lt;INSTRUCTION&gt;test&lt;RESPONSE&gt;", add_eos=False)
[1, 529, 1177, 10810, 29965, 9838, 29958, 1688, 29966, 1525, 5550, 1164, 1660, <b>29958</b>]
</pre>

<pre>
# assume the actual response begins with "&lt;html&gt;"

>>> tokenizer.encode("&lt;INSTRUCTION&gt;test&lt;RESPONSE&gt;&lt;html&gt;", add_eos=False)
[1, 529, 1177, 10810, 29965, 9838, 29958, 1688, 29966, 1525, 5550, 1164, 1660, <b>5299, 1420, 29958</b>]
</pre>

You can see this directly when looking at what the difference when roundtripping the tokenization

```
# last token when just prompt encoded
>>> tokenizer.decode([29958])
'>'

[...]

PR #624, opened 2024-03-31
Refactor datasets and tokenizer

ebsmothers:
Note: a lot of the changes to templates here are primarily done by @RdoubleA 

#### Context
- Currently the way we handle tokenization and masking of prompts and responses across instruct and chat datasets is fragmented and hard to follow. Additionally, as pointed out in #366, there is potential for tokenization errors by tokenizing across message boundaries.
- The safest thing to do is to call the tokenizer's encode method at the level of individual messages, then stitch them together in our own tokenizer class with whatever custom logic we need. This PR allows us to do that.

#### Changelog
- Add `tokenize_messages` API to our tokenizer. This takes in a list of messages, tokenizes each one individually, then stitches the outputs together with any requisite special tokens, truncation, etc. For SentencePiece tokenizer (which is all we currently support), this is just BOS and EOS, but for tokenizers with more complicated sets of special tokens, this allows us to customize at the tokenizer level without the dataset class having to worry about it.
- Align on a unified API for both instruct and chat datasets calling `tokenize_messages`. This is more general than our current usage of `tokenize_prompt_and_response`.
- Split templates into chat formats (for chat datasets) and instruct templates (for instruct datasets). Chat formats are `List[Message] -> List[Message]` to better align with the natural format of chat conversations. Instruct templates still operate as string formatters to maintain the ability for simple prompt formatting on instruct datasets. 
- Other smaller stuff: simplify truncate API (shouldn't require tokens and labels and tokenizer), add handling of leading whitespaces when splitting across messages (sentencepiece defaults to prepending any tokenized string with a whitespace, which we don't always want).

#### Test plan

Added new tokenizer tests for (a) `tokenize_messages` API, and (b) encoding without leading whitespace
Refactored existing template tests into instruct template tests and chat format tests (again thanks @RdoubleA).

```
pytest tests
...
===== 199 passed, 17 skipped, 5 warnings in 90.91s (0:01:30) ===
```

```
...
====== 15 passed, 201 deselected, 3 warnings in 149.88s (0:02:29) =====
```

LoRA finetune on slimorca dataset (manually deleting `dataset.use_clean` from the config cause I can't override from CLI properly)
```
tune run --nproc_per_node 4 lora_finetune_distributed --config llama2/7B_lora checkpointer.checkpoint_dir=/data/users/ebs/checkpoints checkpointer.checkpoint_files=['llama2-7b-torchtune.pt'] checkpointer=torchtune.utils.FullModelTorchTuneCheckpointer tokenizer.path=/data/users/ebs/checkpoints/lora-debug/tokenizer.model dataset=torchtune.datasets.slimorca_dataset metric_logger=torchtune.utils.metric_lo
gging.WandBLogger metric_logger.project=lora-debug
```

Screenshot 2024-04-01 at 9 49 21 PM


tests/test_utils.py
```diff
@@ -50,6 +52,45 @@ def eos_id(self):
     def bos_id(self):
         return 0
 
+    def tokenize_messages(
```
RdoubleA, line 55: We should think of an alternative solution to this, because we'll have to update this every time the real tokenize messages is updated. Can we do a more stripped down approach for testing purposes or is this the most barebones?

tests/torchtune/datasets/test_chat_dataset.py
```diff
@@ -14,18 +14,37 @@
...
+
+    @classmethod
+    def format(
+        cls,
+        messages,
+    ):
+        formatted_dialogue = []
+        for message in messages:
+            content = ""
+            if message.role == "system":
```
RdoubleA, line 34: nit: could put system, user, assistant in a dictionary and just key on message.role

tests/torchtune/datasets/test_slimorca_dataset.py
```diff
@@ -8,12 +8,12 @@
...
 
 from tests.test_utils import get_assets_path
-from torchtune.data import Llama2ChatTemplate
+from torchtune.data import Llama2ChatFormat
 
 from torchtune.datasets import slimorca_dataset
 from torchtune.modules.tokenizer import Tokenizer
 
-LLAMA_TEMPLATE = Llama2ChatTemplate()
+LLAMA_TEMPLATE = Llama2ChatFormat()
```
RdoubleA, line 16: This should not be initialized here, just the class name

torchtune/data/_chat_formats.py
```diff
@@ -0,0 +1,210 @@
...
+
+    system = f"{B_SYS}{{content}}{E_SYS}"
+    user = f"{B_INST} {{system_message}}{{content}} {E_INST} "
+    assistant = ""
+
+    @classmethod
+    def format(
+        cls,
+        sample: List[Message],
+    ) -> str:
```
RdoubleA, line 68: Need to update return to List[Message]

torchtune/data/_templates.py
RdoubleA, line 1: This file should be deleted?

torchtune/data/_utils.py
```diff
@@ -4,70 +4,15 @@
...
-
-    return encoded_prompt_with_response, labels
+from typing import Any, List
 
 
 def truncate(
-    tokenizer: Tokenizer,
-    prompt_tokens: List[int],
-    label_tokens: List[int],
+    tokens: List[Any],
```
RdoubleA, line 11: Can tokens be anything other than int?

torchtune/datasets/_chat.py
```diff
@@ -81,65 +76,23 @@ def __getitem__(self, index: int) -> Tuple[List[int], List[int]]:
...
-                break
-
-        prompt_tokens, label_tokens = truncate(
-            self._tokenizer, prompt_tokens, label_tokens, self.max_seq_len
+        messages = self._convert_to_messages(sample)
+        messages = self.chat_format.format(messages)
+        tokens, mask = self._tokenizer.tokenize_messages(
+            messages, max_seq_len=self.max_seq_len
         )
+        labels = list(np.where(np.logical_not(mask), tokens, CROSS_ENTROPY_IGNORE_IDX))
```
RdoubleA, line 84: Very nice

torchtune/datasets/_chat.py
```diff
@@ -167,15 +120,15 @@ def chat_dataset(
...
     else:
         raise ValueError(f"Unsupported conversation format: {conversation_format}")
 
     return ChatDataset(
         tokenizer=tokenizer,
         source=source,
-        convert_to_dialogue=convert_to_dialogue,
-        template=_get_template(template),
+        convert_to_messages=convert_to_messages,
+        chat_format=chat_format,
```
RdoubleA, line 131: This needs a mapping from str to the actual class pointer, a simpler version of get_template

torchtune/datasets/_slimorca.py
```diff
@@ -61,8 +61,8 @@ def slimorca_dataset(
     return ChatDataset(
         tokenizer=tokenizer,
         source="Open-Orca/SlimOrca-Dedup",
-        convert_to_dialogue=sharegpt_to_llama2_dialogue,
-        template=Llama2ChatTemplate(),
+        convert_to_messages=sharegpt_to_llama2_messages,
+        chat_format=Llama2ChatFormat(),
```
RdoubleA, line 65: No () here because format is a class method

torchtune/modules/tokenizer.py
```diff
@@ -89,3 +91,47 @@ def decode(self, ids: List[int]) -> str:
...
+            tokens = self.spm_model.encode(
+                message.content,
+                add_bos=False,
+                add_eos=False,
+            )
+            tokenized_messages.extend(tokens)
+            mask.extend([message.masked] * len(tokens))
+
+            # Break out early if we reach max_seq_len
+            if max_seq_len and len(tokens) >= max_seq_len:
```
RdoubleA, line 122: len(tokenized_messages)?

torchtune/modules/tokenizer.py
```diff
@@ -89,3 +91,47 @@ def decode(self, ids: List[int]) -> str:
...
+                break
+
+            # If assistant message, append EOS at end
+            if end_of_turn:
```
RdoubleA, line 126: You should probably check for max seq len after the EOS id is added in case this puts it right at max seq len

RdoubleA (commented):
This is overall quite intuitive and lets Instruct and ChatDataset still play specialized roles while utilizing common APIs. I think we've struck a nice balance here.

torchtune/datasets/_chat.py
```diff
@@ -81,65 +76,23 @@ def __getitem__(self, index: int) -> Tuple[List[int], List[int]]:
...
-            label_tokens.extend(labels)
-
-            if len(prompt_tokens) >= self.max_seq_len:
-                break
-
-        prompt_tokens, label_tokens = truncate(
-            self._tokenizer, prompt_tokens, label_tokens, self.max_seq_len
+        messages = self._convert_to_messages(sample)
+        messages = self.chat_format.format(messages)
+        tokens, mask = self._tokenizer.tokenize_messages(
```
RdoubleA, line 81: Where is train_on_input used for ChatDataset?

tests/test_utils.py
ebsmothers: Yeah I agree. The main thing I wanted here was to guarantee I could replicate the performance in the existing unit test using the same logic. We can definitely use a simpler method but will have to change the expected values. (Really I should just add a test for `tokenize_messages` on the tokenizer, then we can use something simple here and still be confident it's working as expected.)

torchtune/data/_utils.py
ebsmothers: So rn I am also calling truncate on mask, which is a list of bools. I am still trying to figure out the cleanest way here.. I do like truncating one sequence at a time better than the existing logic, but the current application to masks is a bit awkward. Lmk if you have any suggestions here

torchtune/modules/tokenizer.py
ebsmothers: Personally I think it's not worth the extra code to check in the case we hit max_seq_len exactly given (a) it's a pretty rare edge case, (b) it adds complexity to the code, and (c) we would still break out on the next message anyways (and we still truncate everything to the right length at the end regardless). Lmk if this makes sense to you

tests/recipes/test_full_finetune_single_device.py
```diff
@@ -47,9 +47,10 @@ def _fetch_expected_loss_values(self):
         return [10.5074, 10.5563, 10.5152, 10.4851]
 
     @pytest.mark.integration_test
-    @pytest.mark.parametrize(
-        "config", ["full_single_device_low_memory", "full_single_device"]
-    )
+    # @pytest.mark.parametrize(
```
ebsmothers, line 50: just for some testing, will remove

torchtune/datasets/_chat.py
ebsmothers: Good point. Gonna pass it to _convert_to_messages, I think it makes most sense there tbh. Basically we want it wherever we are building the messages

torchtune/modules/tokenizer.py
```diff
@@ -60,24 +62,33 @@ def encode(
...
+        trim_leading_whitespace: bool = False,
     ) -> List[int]:
         """Encode text into token IDs.
 
         Args:
             text (str): The input text to be encoded, unbatched.
             add_bos (bool): Whether to prepend BOS to the input, defaults to True.
             add_eos (bool): Whether to append EOS to the input, defaults to True.
-
+            trim_leading_whitespace (bool): Whether to trim leading whitespace from
```
RdoubleA, line 73: wanna link the discussion you found here so people know why this is a thing?

torchtune/modules/tokenizer.py
RdoubleA: yeah I meant just move the if statement above to below this end_of_turn section, shouldn't be extra code. but yeah either way the result is the same, just a minor speed up. your call

torchtune/datasets/_chat.py
RdoubleA: mm `_convert_to_messages` is provided by the user right? so then we have to enforce that their Callables take in this parameter and the burden is on them to mask appropriately. maybe pass it to chat_format or tokenize_messages and let the user know that if they set it it will overwrite whatever custom masking they set up in `_convert_to_messages`?

tests/torchtune/data/test_chat_formats.py
```diff
@@ -0,0 +1,121 @@
...
+
+import pytest
+from torchtune.data import ChatMLFormat, Llama2ChatFormat, Message, MistralChatFormat
+
+# Taken from Open-Orca/SlimOrca-Dedup on HuggingFace:
+# https://huggingface.co/datasets/Open-Orca/SlimOrca-Dedup
+CHAT_SAMPLE = [
+    Message(
+        role="system",
+        content="You are an AI assistant. User will you give you a task. Your goal is to complete the task as faithfully as you can. While performing the task think step-by-step and justify your steps.",  # noqa: B950
```
RdoubleA, line 15: nit: maybe can just chunk the string like the `expected_dialogue`s below so we can remove the noqa?

tests/torchtune/datasets/test_chat_dataset.py
```diff
@@ -135,11 +126,10 @@ def test_get_item(self, mock_load_dataset, chat_format, dialogue):
             + [CROSS_ENTROPY_IGNORE_IDX] * prompt_lengths[1]
             + [1, 6, -1]
         ]
-
         ds = ChatDataset(
             tokenizer=DummyTokenizer(),
             source="iam/agoofy/goober",
-            convert_to_messages=lambda x: x["dialogue"],
+            convert_to_messages=lambda x, y: x["dialogue"],
```
RdoubleA, line 132: why x, y?

RdoubleA (commented):
`Llama2ChatFormat` and `MistralChatFormat` have a lot of whitespaces between messages - have you considered how that interacts with the trimming? should we remove those whitespaces pre-emptively in the ChatFormat classes?

tests/torchtune/datasets/test_chat_dataset.py
ebsmothers: This now takes in train_on_input

tests/torchtune/modules/test_tokenizer.py
```diff
@@ -59,3 +60,232 @@ def test_token_ids(self, tokenizer):
...
+            191,
+            332,
+            92,
+            92,
+            24,
+            24,
+            4,
+            2,
+        ]
+        expected_mask = [True] * 75 + [False] * 125
```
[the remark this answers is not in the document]
ebsmothers: Yikes, sorry. Added

torchtune/modules/tokenizer.py
```diff
@@ -72,16 +79,25 @@ def encode(
...
             trim_leading_whitespace (bool): Whether to trim leading whitespace from
                 underlying sentencepiece tokenization. Default: False
+            prefix (Optional[str]): Optional string to encode for trimming leading
+                whitespaces. Default: None
         Returns:
             List[int]: The encoded token IDs.
         """
         if trim_leading_whitespace:
+            # Can define our own custom prefix depending on vocab if needed
+            if not hasattr(self, "prefix"):
```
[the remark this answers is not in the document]
ebsmothers: This is basically just caching it the first time we use it, hence why it's set in the following line. If it's already been encoded, no need to keep re-encoding it since it will remain fixed for the life of the program.

torchtune/modules/tokenizer.py
ebsmothers: On second thought, I think it might be a bit misleading to just link that with no context. I am gonna add my own comment that imo is more relevant.

torchtune/modules/tokenizer.py
```diff
@@ -72,16 +79,25 @@ def encode(
...
+                self.prefix = prefix or "pre"
+                self.encoded_prefix = self.spm_model.encode(
```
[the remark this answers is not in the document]
ebsmothers: Similar to above, we do this one time and cache the result.

torchtune/modules/tokenizer.py
```diff
@@ -111,23 +127,36 @@ def tokenize_messages(
...
-                # Tokenize current message, append with masks
+
+            # We want to trim leading whitespace on the next message when
+            # (a) it is a continuation of the turn (i.e. not the first message)
+            # (b) the vocabulary explicitly encodes whitespace characters, and
+            # (c) the previous message did not end with a space
+            trim_leading_whitespace = (
+                (not start_of_turn)
+                and self.encodes_whitespace
+                and not prev_ends_with_space
```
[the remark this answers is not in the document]
ebsmothers: L153 as of this version. This is so that we can work with e.g. the grammar format

torchtune/modules/tokenizer.py
ebsmothers: Oh duh lol. Yeah I can do that

torchtune/modules/tokenizer.py
```diff
@@ -63,6 +69,7 @@ def encode(
         add_bos: bool = True,
         add_eos: bool = True,
         trim_leading_whitespace: bool = False,
+        prefix: Optional[str] = None,
```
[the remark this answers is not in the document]
ebsmothers: Idk about assert, but maybe warn? Cause we can set prefix and have it just be a no-op. In general I think it is quite rare to explicitly set prefix though, it should not be needed for the canonical Llama2 tokenizer vocab

torchtune/modules/tokenizer.py
```diff
@@ -42,6 +44,10 @@ def __init__(
         self.bos_id = bos_id
         self.eos_id = eos_id
         self.pad_id = pad_id
+        self.whitespace_encodings = {
+            c: self.spm_model.encode(c) for c in WHITESPACE_CHARS
+        }
+        self.encodes_whitespace = any(self.whitespace_encodings.values())
```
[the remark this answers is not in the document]
ebsmothers: Empty list. That's why I'm doing this -- to check if whitespaces are encoded as part of the model. In fact prob don't even need to save the encodings dict at all

torchtune/datasets/_chat.py
ebsmothers: I am just gonna pass the `ChatFormat` directly for now, lmk if that makes sense

torchtune/data/_transforms.py
```diff
@@ -9,7 +9,9 @@
 from torchtune.data._types import Message
 
 
-def sharegpt_to_llama2_messages(sample: Mapping[str, Any]) -> List[Message]:
+def sharegpt_to_llama2_messages(
+    sample: Mapping[str, Any], train_on_input: bool = False
```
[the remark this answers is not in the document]
ebsmothers: Yeah I agree. I'd like to tackle this as a follow-up tbh

torchtune/datasets/_chat.py
ebsmothers: Yeah I agree. Similar to my comment above, can we handle this in a follow-up? I think it's a relatively standalone change not too related to all this tokenizer business

torchtune/modules/tokenizer.py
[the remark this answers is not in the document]
ebsmothers: OK yeah I am gonna leave out the assertion cause tbh prefix is not really something that I expect people to be experimenting with. Lmk if that makes sense

tests/test_utils.py
RdoubleA: you added the `tokenize_messages` test - should we simplify here?

torchtune/data/_templates.py
RdoubleA, line 1: delete

torchtune/datasets/_chat.py
RdoubleA: hm, then this builder will require a nested component to use from the config. It won't work from the config then. I can update it in a follow up if needed

torchtune/modules/tokenizer.py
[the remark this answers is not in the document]
RdoubleA: this should throw an error because I don't see it defined on the very first message, it is only defined after. you should initialize it as False (or whichever is the default behavior) above the for loop

torchtune/data/_chat_formats.py
```diff
@@ -0,0 +1,215 @@
...
+    prompts with placeholders for the data inputs. There should be a template
+    for each role: system, user, assistant.
+    """
+
+    system = ""
+    user = ""
+    assistant = ""
+
+    @classmethod
+    @abstractmethod
```
kartikayk, line 25: TIL!

torchtune/data/_chat_formats.py
```diff
@@ -0,0 +1,215 @@
...
+        Raises:
+            ValueError: If system prompts are provided
+        """
+        formatted_dialogue = []
+        for message in sample:
+            content = ""
+            if message.role == "system":
+                raise ValueError(
+                    "System prompts are not supported in MistralChatFormat"
+                )
```
kartikayk, line 148: Sorry for the very noob question, but why is this not a pass through as we do for assistant? Basically, are we saying that this format doesnt support "system" as a role or that it doesn't support the system tags? Or are they the same thing? And if "system" isn't supported then should we be setting system above to None instead of empty string?

torchtune/data/_chat_formats.py
```diff
@@ -0,0 +1,215 @@
...
+            assert content != ""
+            formatted_dialogue.append(
+                Message(role=message.role, content=content, masked=message.masked),
+            )
+        return formatted_dialogue
+
+
+class ChatMLFormat(ChatFormat):
+    """
+    OpenAI's Chat Markup Language used by their chat models:
```
kartikayk, line 165: As far as I know, OpenAI models support only tiktoken. Are we adding support for that in this PR or in a follow up? Or not yet?

torchtune/data/_instruct_templates.py
```diff
@@ -0,0 +1,147 @@
...
+                If None, assume these are identical. Note: if the sample output is not named
+                as "output" in the dataset, you always need to map it to "output" in column_map.
+
+        Returns:
+            The formatted prompt
+        """
+        pass
+
+
+class AlpacaInstructTemplate(InstructTemplate):
```
kartikayk, line 40: Sorry if I misunderstood earlier, but I thought we're converting this to follow a similar structure to the chat formats i.e. tokenize instruction, input and response separately?

torchtune/data/_transforms.py
```diff
@@ -36,18 +38,17 @@ def sharegpt_to_llama2_dialogue(sample: Mapping[str, Any]) -> Dialogue:
...
-
-    dialogue = []
+    messages = []
     for message in conversations:
         role = role_map[message["from"]]
         content = message["value"]
-        dialogue.append(Message(role=role, content=content))
-
-    return dialogue
+        masked = (role != "assistant") and train_on_input
```
kartikayk, line 52: Hmm, maybe I'm misunderstanding this, but shouldn;t this be

```
masked = (role != "assistant") and not train_on_input
```

i.e. we don't mask if train_on_input is TRUE?

torchtune/datasets/_chat.py
kartikayk: Nice indeed, but please add comment for mere mortals like me who need to take more than a min to understand this :)

torchtune/datasets/_chat.py
kartikayk: will leave here and duck - is the one layer of component instantiation coming in the way of getting stuff done?

torchtune/modules/tokenizer.py
```diff
@@ -40,6 +44,9 @@ def __init__(
         self.bos_id = bos_id
         self.eos_id = eos_id
         self.pad_id = pad_id
+        self.encodes_whitespace = any(
+            [self.spm_model.encode(c) for c in WHITESPACE_CHARS]
+        )
```
kartikayk, line 49: no idea what's going on here...

torchtune/modules/tokenizer.py
```diff
@@ -60,24 +67,46 @@ def encode(
...
         Args:
             text (str): The input text to be encoded, unbatched.
             add_bos (bool): Whether to prepend BOS to the input, defaults to True.
             add_eos (bool): Whether to append EOS to the input, defaults to True.
-
+            trim_leading_whitespace (bool): Whether to trim leading whitespace from
+                underlying sentencepiece tokenization. Sentencepiece normally prepends
+                whitespace to any tokenized text, which can cause differences where
+                encode(s1) + encode(s2) != encode(s1 + s2) due to leading whitespace
+                added to s2. Default: False
```
kartikayk, line 83: So when would I set this to True?

torchtune/modules/tokenizer.py
```diff
@@ -89,3 +118,59 @@ def decode(self, ids: List[int]) -> str:
...
+    def tokenize_messages(
+        self, messages: List[Message], max_seq_len: Optional[int] = None
+    ) -> Tuple[List[int], List[bool]]:
+        start_of_turn = True
+        end_of_turn = False
+        tokenized_messages = []
+        mask = []
+        for message in messages:
+            # If assistant message, this is the end of a turn
+            end_of_turn = message.role == "assistant"
```
kartikayk, line 131: I'm grossly misunderstanding, but how are both start_of_turn and end_of_turn both True here?

ebsmothers: They shouldn't be. End of turn is just on an assistant message. We set that at the beginning of the iteration, then at the end of the iteration, we set end of turn to False and start of turn to True, since the next iteration will be the start of a turn. Lmk if I'm misunderstanding your question here though

tests/test_utils.py
ebsmothers: Yeah, it is nice to explicitly test the tokenizer on the chat dataset though. I have a better idea (spoiler alert: it involves inheritance 😃)

ebsmothers: Yeah, I do like explicitly testing fetching a sample from ChatDataset though. I have another idea to simplify the code here 😃

torchtune/data/_chat_formats.py
ebsmothers: @RdoubleA may have the best answer here tbh

torchtune/data/_chat_formats.py
ebsmothers: Let's leave it for a follow-up if you're good with that

torchtune/data/_instruct_templates.py
ebsmothers: No this is kinda our middle ground for now. This keeps the convenience of string formatting as an option for instruct datasets then casts to messages on the dataset side. Tbh this is one of the kinks I still want to iron out here, we could go all the way on the `Message` format but I'm hedging a bit for now cause I know people like the string-formatting of prompts. But yeah this may change in the future

torchtune/modules/tokenizer.py
ebsmothers: I should probably add a comment

torchtune/modules/tokenizer.py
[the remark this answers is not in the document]
ebsmothers: Oh god I think it's breaking out of the and early in the test cases so we never hit this. Good catch

torchtune/modules/tokenizer.py
ebsmothers: I tried to describe in the comment on L138-L141. First, we only run into mismatches due to leading whitespace on the nth split of a string, where n>1. This is because sentencepiece automatically prepends whitespace to a string prior to encoding it, so when we do encode(s1+s2) and encode(s1) + encode(s2), s1 has been prepended with whitespace in both cases, and so the behavior is the same (this is point (a)).

Second, we also only see a mismatch if the tokenizer actually explicitly encodes whitespace, which not all tokenizers do. E.g. if `spm_test` corresponds to our tokenizer in `tests/assets/m.model`, and `spm_llama` corresponds to the usual Llama2 tokenizer, we can actually see that

```
>>> spm_test.encode(" ", add_bos=False, add_eos=False)
[]
>>> spm_test.encode("\n", add_bos=False, add_eos=False)
[]
>>> spm_llama.encode(" ", add_bos=False, add_eos=False)
[259]
>>> spm_llama.encode("\n", add_bos=False, add_eos=False)
[29871, 13],
```

so our test tokenizer does not return any tokens when it only sees whitespace. Because of this, we get different behavior in both cases when splitting a single string.
```
# When the tokenizer doesn't tokenize whitespace, the results match
>>> spm_test.encode("hi\nthere", add_bos=False, add_eos=False)
[476, 70]
>>> spm_test.encode("hi\n", add_bos=False, add_eos=False) + spm_test.encode("there", add_bos=False, add_eos=False)
[476, 70]

# On the regular Llama2 tokenizer, the results do not match
>>> spm_llama.encode("hi\nthere", add_bos=False, add_eos=False)
[7251, 13, 12711]
>>> spm_llama.encode("hi\n", add_bos=False, add_eos=False) + spm_llama.encode("there", add_bos=False, add_eos=False)
[7251, 13, 727]
```

Finally, there are some prompts that end with " " (e.g. our `GrammarErrorCorrectionTemplate`). In this case, the concatenated string actually contains a space, so we do not want to strip it from s2. But we do want to remove it from the preceding string to ensure we don't double-count it. This is the reason for the `.rstrip(" ")` in L150.

So these are the three conditions we need to check to determine whether to do this special handling, and they are all checked in `tokenize_messages` below.

torchtune/modules/tokenizer.py
ebsmothers: But yeah, discussed the purpose of this variable in the second point [here](https://github.com/pytorch/torchtune/pull/624/files#r1547101264)

torchtune/datasets/_chat.py
ebsmothers: Yeah let's do in a follow-up. This will only affect custom chat datasets, right? Which we don't have any of yet anyways. Re nested instantiation, imo this is not a sufficient reason to add it.. I think we can find another way here

torchtune/datasets/_chat.py
ebsmothers: I made this needlessly complicated so that's my fault

torchtune/data/_transforms.py
ebsmothers: Thanks, good catch. I got thrown off by the `test_slimorca_dataset` test, which was always asserting that the final token of labels was equal to eos_id, even when max_seq_len was so short that we were not including the assistant message and train_on_input was False. So I think the test was doing the wrong thing in that case (if train_on_input=False and we only have inputs, then everything should be masked, even the EOS token). Lmk if this makes sense to you

tests/test_utils.py
```diff
@@ -31,8 +31,11 @@
     "llama2_7b": "/tmp/test-artifacts/llama2-7b-torchtune.pt",
 }
 
+# Inherit from tokenizer class to reuse its tokenize_messages method
```
RdoubleA, line 34: did you just say inherit 👀

torchtune/data/_chat_formats.py
RdoubleA: Mistral does not support the system role, so if a user passes in a message with a system role, we need to either error out or raise a warning that it will be ignored. See for context: https://github.com/vllm-project/vllm/issues/2080#issuecomment-1854436854

Agreed on setting system to None instead of an empty string.

torchtune/data/_chat_formats.py
RdoubleA: I believe they use this format with TikToken: https://community.openai.com/t/how-does-chatml-do-the-exact-formatting/80751

The main motivation to add this template is that it is default in HF if the model has no custom template: https://github.com/huggingface/transformers/blob/096f304695f7e7b169b031f7814352e900ad71c4/src/transformers/tokenization_utils_base.py#L1838

torchtune/datasets/_chat.py
RdoubleA: I think the string mapping is very simple and uses our existing tools without requiring nested instantiation or impacting UX. Will do a follow-up and we can discuss there.

torchtune/modules/tokenizer.py
RdoubleA: start of turn and end of turn can both be true here if the list of messages the user passes in consists of only a single assistant message. This would not be a valid dialogue, but the BOS and EOS should still be appended correctly. I wonder if we should validate that the dialogue is well-formed here or outside this method

RdoubleA (approved):
Couple of minor comments, otherwise looks good to go! This was a tough one, thanks for pushing this through. Let's make sure we keep track of the follow-up items that were discussed.

tests/test_utils.py
ebsmothers: Best way to reduce copy-pasted code 😉. But for this case I do think it's the right thing, since it still allows us to call `tokenize_messages` (e.g. when testing on datasets) with the stripped-down tokenizer but using equivalent logic. The only usage currently is in test_slimorca.py, but I do kinda like that test, so keeping it like this for now.

torchtune/modules/tokenizer.py
ebsmothers: Yeah good point. Imo it's not up to the tokenizer to decide this. Let's consider adding a separate utility e.g. `validate_messages` that can be called from the dataset class or elsewhere as a follow-up.

PR #625, opened 2024-03-31
Print out "Ignoring patterns" for download

### LINKED ISSUE #411: [Discussion] "tune download" behavior is unintuitive
As I walked through ```tune download``` as a user, I found the behavior unintuitive. Concretely:

- **Name**. ```download``` should be renamed to ```hf_download```. That's what this command does and without any information, I get the impression that I can use this to download a model from any source.

- **Restrictions**. ```download``` currently restricts my choices to what is defined in ```_SUPPORTED_MODEL_REPO_IDS```. This seems arbitrarily restrictive since I could download a model, write my own conversion script and use TorchTune to finetune the model. This was the use case I had in mind and i ended up having to modify code to do this. The behavior I would expect in this case is, let the user download the model they want but raise a warning if the model is not supported OOTB. Anyways ```convert-checkpoint``` will raise an error, so I'm not too concerned about the user shooting themselves in the foot by fine-tuning a model we don't support.

joecummings:
#### Context
- Increases visibility into which files WON'T be download. An ask from @kartikayk 

Closes #411 

#### Changelog
- Add print statement
- Update test

#### Test plan
- CI

### DIFF
--- a/tests/torchtune/_cli/test_download.py
+++ b/tests/torchtune/_cli/test_download.py
@@ -38,18 +38,29 @@ class TestTuneDownloadCommand:
         # Call the first time and get GatedRepoError
         with pytest.raises(SystemExit, match="2"):
             runpy.run_path(TUNE_PATH, run_name="__main__")
-        err = capsys.readouterr().err
-        assert "It looks like you are trying to access a gated repository." in err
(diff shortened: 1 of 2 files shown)


kartikayk (approved):
Thanks so much!

PR #627, opened 2024-04-01
[Perf Tools] Torch profiler component

SLR722:
#### Context
Based on the discussions in https://github.com/pytorch/torchtune/pull/604 and https://github.com/pytorch/torchtune/pull/608, we plan to make torch profiler (as well as memory snapshot in the future) as components to preserve the composable design principle. Uses can choose to bring in the provided perf tools when necessary

We also created torch profiler to full_finetune_single_device recipe to showcase how to use the perf tools. (Will add a section about how to use perf tools in readme tutorial)

#### Changelog
- add perf profiler component

#### Test plan
tune run lora_finetune_single_device --config llama2/7B_lora_single_device 

Screenshot 2024-04-01 at 12 05 32 AM


torchtune/utils/_perf_profiler.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+def perf_profiler(
+    output_dir: Optional[str] = "./torchtune_perf_tracing.json",
+) -> ContextManager:
+    """
+    Utility component to trace the code with pytorch profiler.
+    check the user manual: https://pytorch.org/docs/stable/profiler.html for more details.
+    The schedule for this profiler is wait 5 steps, warmup 5 steps, trace 5 steps
+    Note: Enable pytorch profiler may casue performance overhead.
```
RdoubleA, line 20:
```suggestion
    Note: Enabling pytorch profiler may add performance overhead.
```

Do you know if this is additional memory or a speed reduction?

torchtune/utils/_perf_profiler.py
```diff
@@ -0,0 +1,47 @@
...
+import torch
+from torch.profiler import profile
+
+
+def perf_profiler(
+    output_dir: Optional[str] = "./torchtune_perf_tracing.json",
+) -> ContextManager:
+    """
+    Utility component to trace the code with pytorch profiler.
+    check the user manual: https://pytorch.org/docs/stable/profiler.html for more details.
```
RdoubleA, line 18: Would be nice to add some details on what exactly the profiler tracks

RdoubleA (commented):
One nit comment: instead of perf_tools_showcase, maybe call it full_finetune_single_device_with_profiling? Much shorter and easier to type out

SLR722:
> One nit comment: instead of perf_tools_showcase, maybe call it full_finetune_single_device_with_profiling? Much shorter and easier to type out

Thanks for your feedback @RdoubleA! I made it more general instead of just profiling since I want to put the example usage of some other perf tools (such as memory snapshot) in this recipe as well. Not sure if all of them could fit in 'profling'

torchtune/utils/_perf_profiler.py
SLR722: Mainly speed reduction. Will update the comment to be more specific

recipes/full_finetune_single_device_perf_tools_showcase.py
```diff
@@ -0,0 +1,400 @@
...
+        - Overwritten by arguments from the command-line
+    """
+    recipe = FullFinetuneRecipeSingleDevicePerfToolsShowcase(cfg=cfg)
+    recipe.setup(cfg=cfg)
+    recipe.train()
+    recipe.cleanup()
+
+
+if __name__ == "__main__":
+    sys.exit(recipe_main())
```
rohan-varma, line 400: I don't think we need an entirely separate recipe + config just for the profiling component. I think one idea is to build the profiler as a component, and integrate it into a single _existing_ recipe as a showcase. Rest of the recipes can have it off.

tests/recipes/test_full_finetune_single_device_perf_tools_showcase.py
```diff
@@ -0,0 +1,145 @@
...
+        monkeypatch.setattr(sys, "argv", cmd_2)
+        with pytest.raises(SystemExit, match=""):
+            runpy.run_path(TUNE_PATH, run_name="__main__")
+
+        expected_loss_values = self._fetch_expected_loss_values()[2:]
+
+        loss_values = get_loss_values_from_metric_logger(log_file)
+        torch.testing.assert_close(
+            loss_values, expected_loss_values, rtol=1e-4, atol=1e-4
+        )
```
rohan-varma, line 145: Don't need a separate test file either IMO. Makes it hard to test more features across all recipes

recipes/full_finetune_single_device_perf_tools_showcase.py
SLR722: Thanks for raising this @rohan-varma! I actually thought hard about what is the best way to do this as well. My concern around integrating it with 1 existing recipe is that for hobbyists that won't look deep in the recipe, integrating will cause confusion to them as profiler is not necessary for every training and mainly for debugging purpose. Having a separate recipe is more clear although it will increase the complexity of our recipe system by introducing 1 more recipe. 

But I'm open to integrate the component into an existing recipe. Looking forward to other people's thought.

rohan-varma (commented):
Recipe component itself looks fine to me. Main question / ask is to remove the additional added recipe and simplify the testing.

recipes/configs/llama2/7B_lora_single_device.yaml
```diff
@@ -82,3 +82,9 @@ log_every_n_steps: null
 device: cuda
 dtype: bf16
 enable_activation_checkpointing: True
+
+# Show case the usage of pytorch profiler
```
rohan-varma, line 86: Would just comment this out and add this piece of documentation in the code itself.

recipes/lora_finetune_single_device.py
```diff
@@ -207,6 +207,8 @@ def setup(self, cfg: DictConfig) -> None:
             last_epoch=self.total_training_steps - 1,
         )
 
+        self._perf_profiler = config.instantiate(cfg.perf_profiler)
```
rohan-varma, line 210: nit: just call it `_profiler` which has better synergies with `torch.profiler`.

torchtune/utils/_perf_profiler.py
```diff
@@ -0,0 +1,47 @@
...
+
+import torch
+from torch.profiler import profile
+
+
+def perf_profiler(
+    output_dir: Optional[str] = "./torchtune_perf_tracing.json",
+) -> ContextManager:
+    """
+    Utility component to measure the time and memory consumption of the model’s operators.
```
rohan-varma, line 17: Let's beef up this documentation a bit IMO:

```
Utility component that wraps around `torch.profiler` to profile model's operators. See <docs> for more details. 
```

torchtune/utils/_perf_profiler.py
```diff
@@ -0,0 +1,47 @@
...
+    check the user manual: https://pytorch.org/docs/stable/profiler.html for more details.
+    The schedule for this profiler is wait 5 steps, warmup 5 steps, trace 5 steps
```
rohan-varma, line 19: IMO we need to be waiting for way more steps. Torch compile and FSDP iterations take a bit longer for this to warm up. 

Some initial defaults I'd put are wait 100 steps, warm up 5, trace 5.

torchtune/utils/_perf_profiler.py
```diff
@@ -0,0 +1,47 @@
...
+    Note: Enabling pytorch profiler may have training speed reduction.
+
+    Args:
+        output_dir (Optional[str]): Tracing file output path.
```
rohan-varma, line 23: Please specify the default if this is passed as None.

rohan-varma (commented):
Almost ready to go! thanks

recipes/configs/llama2/7B_lora_single_device.yaml
ebsmothers: I agree with commenting it out, probably don't want it turned on by default. But won't the `config.instantiate(cfg.perf_profiler)` error out in the recipe as written then?

recipes/lora_finetune_single_device.py
```diff
@@ -390,49 +392,59 @@ def train(self) -> None:
...
-                    self._lr_scheduler.step()
-                    # Update the number of steps when the weights are updated
-                    self.total_training_steps += 1
-                # Log peak memory for iteration
-                if self.total_training_steps % self._log_peak_memory_every_n_steps == 0:
-                    log.info(
-                        utils.memory_stats_log("Memory Stats:", device=self._device)
-                    )
+
+            # Showcase how to use pytorch profiler to trace the training loop.
```
ebsmothers, line 396: nit: 
```suggestion
            # Optionally profile the training loop
```

recipes/configs/llama2/7B_lora_single_device.yaml
SLR722: Yeah, as @ebsmothers mentioned, comment this out in yaml will cause recipe error. That's the reason I created a dedicate showcase recipe for all the non-essential perf tools stuff

rohan-varma: ah, this makes sense. 

How about an argument `enabled` that `perf_profiler` accepts? Then we can easily include these in configs / recipes, and if enabled is False, we just don't enable the profiler. cc @ebsmothers , does this make sense?

torchtune/utils/_profiler.py
```diff
@@ -0,0 +1,50 @@
...
+    output_dir: Optional[str] = "./torchtune_perf_tracing.json",
+) -> ContextManager:
+    """
+    Utility component that wraps around `torch.profiler` to profile model's operators.
+    See https://pytorch.org/docs/stable/profiler.html for more details.
+    The schedule for this profiler is wait 100 steps, warmup 5 steps, trace 5 steps
+    Note: Enabling pytorch profiler may have training speed reduction.
+
+    Args:
+        enabled (Optional[bool]): Enable pytorch profiler. Default is False.
```
rohan-varma, line 26: Incorrect default?

rohan-varma (approved):
Think this should be good to ship as a starting point. @ebsmothers had some comments about comment wording in the config, please ensure that's squared away before landing. Thanks so much!

Issue #616, opened 2024-03-29
Could it support Gemma?


solitude-alive:
The Google model have 2B model, it seems that we can use less than 4*24GB GPUs to fine-tune with full parameters. Do you plan to support it?

joecummings:
We are considering many new model additions and will keep you posted!

kartikayk:
@solitude-alive awesome! 

For a starting point, take a look at the Mistral 7B model builder in:
https://github.com/pytorch/torchtune/blob/main/torchtune/models/mistral/_model_builders.py

We expose specific models through model builders which basically stitich together components (eg: Attention, RoPE, RMS Norm etc). You can find some examples here:
https://github.com/pytorch/torchtune/blob/main/torchtune/models/mistral/_component_builders.py#L36

I think adding support for gemma_2b would be similar. You just need to make sure the components line up with what Gemma is doing.

solitude-alive:
@kartikayk Hi,

_model_builders.py and _component_builders.py have been mostly completed, except for a few details that need to be confirmed.

Is there any relevant documentation on load weight files? 
Gemma on hugging face seems support [model-00001-of-00002.safetensors, model-00001-of-00002.safetensors] rather than [consolidated.00.pth]

solitude-alive:
@kartikayk Hi,

_model_builders.py and _component_builders.py have been mostly completed, except for some components that need to be confirmed.

Is there documentation on how to load the weights file? It seems that Gemma only support [model-00001-of-00002.safetensors, model-00002-of-00002.safetensors] rather than .bin or .pth files.

joecummings:
Also, @solitude-alive - would love for you to join the Discord channel (see our README for invite link) so we can quickly answer any questions you may have as you work on this!

solitude-alive:
@joecummings Yeah, thanks.

kartikayk:
@solitude-alive Awesome! As @joecummings said, it would be awesome to add safetensor support to TorchTune's ```FullModelHfCheckpointer```.

 I verified that ```safetensors.safe_open``` produces the same state_dict with ```safetensor``` files as the TorchTune HF Checkpointer does with ```bin``` files for the llama-13B model. Here's a minimal validation:

&nbsp; 

```
# Examine safetensors

from safetensors import safe_open
from torchtune.models import convert_weights
from torchtune.utils import FullModelHFCheckpointer, ModelType

[...]

PR #630, opened 2024-04-01
Gemma

solitude-alive:
#### Context
- To support full fine-tune with Gemma-2B model 

#### Changelog
- Create gemma module, update for loading with .safetensors, support tied weight model.

#### Test plan
- ....


torchtune/models/gemma/_component_builders.py
```diff
@@ -0,0 +1,387 @@
...
+        hidden_dim (int): hidden dimension of the MLP
+    """
+    gate_proj = nn.Linear(dim, hidden_dim, bias=False)
+    down_proj = nn.Linear(hidden_dim, dim, bias=False)
+    up_proj = nn.Linear(dim, hidden_dim, bias=False)
+    activation = nn.GELU(approximate="tanh")
+    return FeedForward(gate_proj=gate_proj, down_proj=down_proj, up_proj=up_proj, activation=activation)
+
+
+def lora_gemma(
```
joecummings, line 127: Looks like you included the LoRA version of Gemma for this PR. Are you planning on including LoRA, as well, or just starting with the full fine-tuning version?

torchtune/modules/feed_forward.py
```diff
@@ -25,12 +26,13 @@ def __init__(
         gate_proj: nn.Module,
         down_proj: nn.Module,
         up_proj: nn.Module,
+        activation: nn.Module = F.silu,
     ):
         super().__init__()
         self.w1 = gate_proj
         self.w2 = down_proj
         self.w3 = up_proj
-        self.activation = F.silu
```
joecummings, line 33: Good abstraction!

torchtune/utils/_checkpointing/_checkpointer_utils.py
```diff
@@ -11,14 +11,17 @@
 import torch
 import torch.nn as nn
 import torch.optim as optim
+from safetensors import safe_open
```
joecummings, line 14: Can you add this to requirements.txt?

torchtune/models/gemma/_model_builders.py
```diff
@@ -0,0 +1,93 @@
...
+
+def gemma_2b() -> TransformerDecoder:
+    """
+    Builder for creating a Gemma 2B model initialized w/ the default 2b parameter values
+
+    Returns:
+        TransformerDecoder: Instantiation of Gemma 2B model
+    """
+    return gemma(
+        vocab_size=256_000,
```
joecummings, line 30: Still shocked by this vocab size - so large!

recipes/full_finetune_distributed.py
```diff
@@ -203,6 +210,7 @@ def _setup_model(
         cfg_model: DictConfig,
         enable_activation_checkpointing: bool,
         model_state_dict: Dict[str, Any],
+        mode_tie: bool = False,
```
joecummings, line 213:
```suggestion
        model_tie: bool = False,
```

recipes/full_finetune_distributed.py
```diff
@@ -259,6 +267,10 @@ def _setup_model(
             ),
         )
 
+        if mode_tie:  # Tie the weights of the model if required
```
joecummings, line 270:
```suggestion
        if model_tie:  # Tie the weights of the model if required
```

joecummings (commented):
This is great - thanks for the contribution! I left a couple comments, but generally looks good.

Can you add a screenshot of running a distributed full finetune with Gemma to confirm it works?

recipes/full_finetune_distributed.py
solitude-alive: Thank you for your suggestion, I have fixed it in the latest version.

recipes/full_finetune_distributed.py
solitude-alive: Thank you for your suggestion, I have fixed it in the latest version.

torchtune/models/gemma/_model_builders.py
solitude-alive: Yeah, :joy:

torchtune/utils/_checkpointing/_checkpointer_utils.py
solitude-alive: Yeah, thank you for your suggestion, I add it in the latest version.

torchtune/models/gemma/_component_builders.py
solitude-alive: May take a longer time to complete LoRA version of Gemma, could I PR full fine-tuning version first?

solitude-alive:
Yeah, this is a screenshot of running a distributed full finetune with Gemma. 
Screenshot 2024-04-02 at 8 16 34 AM

recipes/configs/gemma/2B_full.yaml
```diff
@@ -0,0 +1,83 @@
...
+#   --config gemma/2B_full \
+#
+# You can add specific overrides through the command line. For example
+# to override the checkpointer directory while launching training
+# you can run:
+#   tune --nnodes 1 --nproc_per_node 4 full_finetune_distributed \
+#   --config gemma/2B_full \
+#   checkpointer.checkpoint_dir=<YOUR_CHECKPOINT_DIR>
+#
+# This config works best when the model is being fine-tuned on 2+ GPUs.
```
ebsmothers, line 21: nit: I think if we are running with `full_finetune_distributed` recipe it will *only* work on 2+ GPUs

recipes/configs/gemma/2B_full.yaml
```diff
@@ -0,0 +1,83 @@
+# Config for multi-device full finetuning in full_finetune_distributed.py
+# using a gemma 2B model
+#
+# This config assumes that you've run the following command before launching
+# this run:
+#   tune download --repo-id meta-llama/gemma-2b \
```
ebsmothers, line 6:
```suggestion
#   tune download --repo-id google/gemma-2b \
```

torchtune/models/gemma/_model_builders.py
```diff
@@ -0,0 +1,93 @@
...
+
+"""
+Model builders build specific instantiations using component builders. For example
+the ``gemma_2b`` model builder uses the ``gemma`` component builder.
+"""
+
+
+def gemma_2b() -> TransformerDecoder:
+    """
+    Builder for creating a Gemma 2B model initialized w/ the default 2b parameter values
```
ebsmothers, line 24: nit: add pointer to the paper or blog post here

torchtune/models/gemma/_model_builders.py
ebsmothers: Does this mean the embedding(/output projection since they're tied) constitutes a full 25% of their params?!

torchtune/models/gemma/_component_builders.py
ebsmothers: I think it should be fine to start with just full fine-tune for now

torchtune/utils/_checkpointing/_checkpointer_utils.py
```diff
@@ -11,14 +11,17 @@
...
 from torch.distributed.fsdp import FullyShardedDataParallel as FSDP
 
 from torchtune.utils._distributed import contains_fsdp
+from transformers.utils import is_safetensors_available
```
ebsmothers, line 18: I think we shouldn't import from transformers here as it's not in our core dependencies. If you've added safetensors to our core dependencies (based on the above comment) probably don't need to do this check anyways.

torchtune/modules/feed_forward.py
```diff
@@ -25,12 +26,13 @@ def __init__(
         gate_proj: nn.Module,
         down_proj: nn.Module,
         up_proj: nn.Module,
+        activation: nn.Module = F.silu,
```
ebsmothers, line 29: nit: technically F.silu is a Callable, not an nn.Module

torchtune/models/gemma/_component_builders.py
```diff
@@ -0,0 +1,387 @@
...
+        dim (int): input dimension to the MLP
+        hidden_dim (int): hidden dimension of the MLP
+        lora_rank (int): rank of each low-rank approximation
+        lora_alpha (float): scaling factor for the low-rank approximation
+        lora_dropout (float): LoRA dropout probability. Default: 0.0
+        quantize_base (bool): whether to quantize the base of the LoRA approximation.
+            Default: False
+
+    Returns:
+        FeedForward: instantiation of the MLP module with LoRA applied to
```
ebsmothers, line 363: I think the second line of this docstring got lost somewhere along the way

recipes/full_finetune_distributed.py
```diff
@@ -151,13 +151,20 @@ def setup(self, cfg: DictConfig) -> None:
 
         ckpt_dict = self.load_checkpoint(cfg.checkpointer)
 
+        # TODO: This is a temporary fix to handle the model tie for GEMMA models.
+        if cfg.checkpointer.model_type == "GEMMA":
+            model_tie = True
+        else:
+            model_tie = False
```
ebsmothers, line 158: We could also consider associating a weight tying config explicitly with the model type and using that in the checkpointer. E.g. 

```
@dataclass
ModelType
	name: str
	weight_tying_config: Dict[str, str] = field(default_factory=dict)
```

Then Gemma would be `ModelType(name="GEMMA", weight_tying_config={"tok_embeddings.weight": "output.weight"}`

(Anyways, not a blocker for this PR as it's more of a design question)

recipes/full_finetune_distributed.py
```diff
@@ -259,6 +267,10 @@ def _setup_model(
             ),
         )
 
+        if model_tie:  # Tie the weights of the model if required
+            model.output.weight = model.tok_embeddings.weight
```
ebsmothers, line 271: It was pointed out by @rohan-varma that this may not actually do what we expect because FSDP has already sharded the params, so let's double-confirm via testing that the weights are tied correctly here.

ebsmothers (commented):
Thanks for this PR! Really excited to see how nicely this is shaping up. 

Re testing, aside from making sure training runs, let's try to get a sanity check that the model forward here lines up with the one from the original implementation on some dummy data (assuming you haven't done so already). We have a bunch of scripts we've used in the past for this with various components in the library, so you can use these as a reference if it helps. [For example](https://github.com/pytorch/torchtune/blob/81f67058af0c2868c3fb9c29c163e258e6642452/tests/torchtune/models/llama2/scripts/compare_decoder.py) (Note: you do not have to actually write a script like this and check it in, this is meant more as a reference if it helps you)

torchtune/models/gemma/_model_builders.py
```diff
@@ -0,0 +1,94 @@
...
+    """
+    return gemma(
+        vocab_size=256_000,
+        num_layers=18,
+        num_heads=8,
+        head_dim=256,
+        num_kv_heads=1,
+        embed_dim=2048,
+        intermediate_dim=16384,
+        max_seq_len=32768,
```
kartikayk, line 38: Is this right? I thought this was 8192 for Gemma 2B

torchtune/utils/_checkpointing/_checkpointer.py
```diff
@@ -383,6 +383,14 @@ def load_checkpoint(self) -> Dict[str, Any]:
             dim=self._config["hidden_size"],
         )
 
+        if (
+            self._model_type == "GEMMA"
```
kartikayk, line 387: Hmm so I have a question about this code.

```hf_to_tune``` makes an assumption that ```head_dim * num_heads = dim``` (see [here](https://github.com/pytorch/torchtune/blob/main/torchtune/models/convert_weights.py#L137)). 

But this isn't true for Gemma 7B where ```num_heads=16``` and ```head_dim= 256``` but ```dim=3072``` and not ```4096```. So we will need to differentiate between gemma 2b and 7b here

File: torchtune/utils/_checkpointing/_checkpointer.py [hunk 17]
```diff
@@ -431,6 +439,29 @@ def save_checkpoint(
             dim=self._config["hidden_size"],
         )
 
+        print(f"======={self._model_type}==========")
+        if (
+            self._model_type == "GEMMA"
```
kartikayk, line 444: Can we move this to a separate utility function in checkpointer_utils? we should keep the checkpointer as clean as possible.

recipes/configs/gemma/2B_full.yaml
solitude-alive: yeah, I update it in the latest version.

torchtune/models/gemma/_model_builders.py
solitude-alive: yeah, I add it in the latest version.

torchtune/utils/_checkpointing/_checkpointer_utils.py
solitude-alive: yeah, I remove it in the latest version.

torchtune/utils/_checkpointing/_checkpointer.py
kartikayk: Also, please move to a utility function in checkpointer_utils so we can keep this code clean.

torchtune/utils/_checkpointing/_checkpointer.py
```diff
@@ -431,6 +439,29 @@ def save_checkpoint(
...
+                    state_dict[utils.MODEL_KEY]["lm_head.weight"],
+                    state_dict[utils.MODEL_KEY]["model.embed_tokens.weight"],
+                ):
+                    del state_dict[utils.MODEL_KEY]["lm_head.weight"]
+                    logger.info(
+                        "Delete the model lm_head weight from the state dict "
+                        "because it is the same as the model embed_tokens weight"
+                    )
+                else:
+                    self._weight_map["lm_head.weight"] = "0002"
```
kartikayk, line 460: When will this else block hit? If we know the checkpoints don't contain this key, let's just work with that assumption? Anyways we're hard coding a bunch of stuff like the name of the key etc.

torchtune/models/gemma/_component_builders.py
solitude-alive: yeah, I fix it in the latest version.

torchtune/utils/_checkpointing/_checkpointer_utils.py
```diff
@@ -54,9 +57,16 @@ def safe_torch_load(checkpoint_path: Path) -> Dict[str, Any]:
     try:
         # convert the path into a string since pathlib Path and mmap don't work
         # well together
-        state_dict = torch.load(
-            str(checkpoint_path), map_location="cpu", mmap=True, weights_only=True
-        )
+        if str(checkpoint_path).endswith(".safetensors") and is_safetensors_available():
```
kartikayk: Not a fan of this approach. Can we just add a key to the config, something like ```is_safetensors_file``` and then based on the value determine if we use torch.load or not. Also please break this down into a sub function (eg: ```load_from_safetensor``` or something similar.

@joecummings WDYT?

torchtune/models/gemma/_model_builders.py
solitude-alive: I calculate it with `count_trainable_parameters `, the params of embed_tokens is 21%.

```
def count_trainable_parameters(model):
    return sum(p.numel() for p in model.parameters() if p.requires_grad)
```

kartikayk (commented):
Thanks for adding this PR @solitude-alive! This would be an awesome contribution to the repo!

Similar to @joecummings I have some questions in the code. My biggest question though is correctness. The loss from the screenshot seems to be much higher than what we've seen with Mistral/Llama2. have you compared this loss for gemma with the official implementation/some other implementation? Or have you seen some issues/blogs which show case the loss value during training that we can compare against?

Also, when adding models we provide some evidence of model numerical correctness - this is really important to build confidence with our users. Please see how we did this for llama2 13B and mistral 7B in the context section of this PR: https://github.com/pytorch/torchtune/pull/571. Would be great if you can add a similar check for Gemma2B. This check would look something like:

- Load official implementation of Gemma2B and take a random tensor, run forward and get output
- Load torchtune implementation, take same tensor, run forward and get output
- Compare outputs with torch.allclose and make sure this returns a True.

recipes/full_finetune_distributed.py
solitude-alive: Thank you for pointing out it, I check the model weight after training, they are not same. Is there any solution? I'm not familiar with that. This can cause some problems if the weights are tied before FSDP. [issue](https://github.com/pytorch/torchtune/issues/616#issuecomment-2028619062)

ebsmothers: OK sorry for the back and forth on this. Confirmed with @rohan-varma that we should not tie weights after FSDP wrapping after all. The main issue was not FSDP but the initialization on meta device. Unfortunately, weight tying + meta device is tricky because the usage of `to_empty` breaks existing references. 

Instead, for Gemma we can do everything on CPU without using meta device at all, basically initializing the model on CPU for every rank and then defining a more vanilla FSDP without the `param_init_fn` we currently have. This should work fine for smaller models (at least up to 7B). @kartikayk put together a snippet on what this can look like, you can find it [here](https://gist.github.com/ebsmothers/332b8432c19c0717778e553b133face3).

We need to decide what the best way to expose this is, but for now feel free to create a separate recipe for Gemma. It should look pretty much the same as the existing `full_finetune_distributed.py`, but with the changes needed to initialize everything on CPU and perform weight tying there before wrapping with FSDP.

Thanks also to @awgu for helping debug this.

torchtune/models/gemma/_component_builders.py
solitude-alive: Yeah, I remove them in the latest version.

recipes/full_finetune_distributed.py
solitude-alive: Thank you for your suggestion, I update in the `gemma_full_finetune.py`.

recipes/full_finetune_distributed.py
solitude-alive: Thank you for your suggestion, it works well.

recipes/full_finetune_distributed.py
```diff
@@ -259,6 +267,10 @@ def _setup_model(
             ),
         )
 
+        if model_tie:  # Tie the weights of the model if required
```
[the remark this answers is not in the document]
solitude-alive: Yeah, I update them in the latest version.

torchtune/models/gemma/_model_utils.py
```diff
@@ -0,0 +1,22 @@
...
+
+from torchtune.modules import TransformerDecoder
+
+
+def tie_weight(model: TransformerDecoder) -> None:
+    """
+    Tie the weights of the output embeddings and the token embeddings in the model.
+
+    Args:
+        model (TransformerDecoder): The to tie the weights of the output embeddings and the token embeddings.
```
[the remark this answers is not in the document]
solitude-alive: Sorry, I have modified it.

torchtune/models/gemma/_model_builders.py
solitude-alive: Sorry, this is my problem, I fixed it in the latest version.

torchtune/utils/_checkpointing/_checkpointer.py
solitude-alive: If I allow explicit parameter `num_heads ` passing in function `hf_to_tune `, is this allowed?

And I moved them to a utility function in checkpointer_utils.

torchtune/utils/_checkpointing/_checkpointer.py
solitude-alive: Yeah, I did it in the latest version.

torchtune/utils/_checkpointing/_checkpointer.py
solitude-alive: It was because the parameters were not really tied before. Now I removed the else block.

solitude-alive:
I fixed the some problems with Gemma2B and did the similar check for Gemma2B.
Screenshot 2024-04-03 at 1 25 49 PM
Screenshot 2024-04-03 at 1 27 24 PM

recipes/gemma_full_finetune_distributed.py
```diff
@@ -0,0 +1,520 @@
...
+
+from torchtune.recipe_interfaces import FTRecipeInterface
+
+from tqdm import tqdm
+
+
+log = utils.get_logger("DEBUG")
+
+
+@dataclass
```
kartikayk, line 40: You don't need to define this hear. This can just be added to the yaml file? Just add a section with:

```
share_weights: True
share_weights_mapping: {
   "output.weight" : "tok.weight"
}
```

I think this should work. And then you can read this directly from the yaml

recipes/gemma_full_finetune_distributed.py
```diff
@@ -0,0 +1,520 @@
...
+        Model initialization has some important considerations:
+            a. To minimize GPU peak memory, we load the model on CPU with the right
+               dtype. To ensure that we don't instantiate ``world_size`` number of models,
+               we initialize on meta_device for all ranks other than rank 0.
+            b. Rank 0 is also responsible for calling ``load_state_dict`` and loading the
+               model weights from checkpoint.
+            c. While wrapping the model with FSDP, we set ``sync_module_states``
+               to TRUE and broadcast module params and buffers from rank 0.
+            d. The ``device_id`` param ensures that the FSDP initialization happens on
+               the correct device.
```
kartikayk, line 236: You need to update this doc string since we no longer init on meta device. Also add info about sharing weights

kartikayk (commented):
Thanks for patiently making all of these changes! Some more comments, but this seems very close. I think once you address these comments, we should be able to accept and merge this.

joecummings:
One last thing before approval and merging, can you run `pre-commit run --all-files`? This will lint everything properly. @solitude-alive

recipes/gemma_full_finetune_distributed.py
```diff
@@ -0,0 +1,503 @@
...
+            sync_module_states=True,
+            # Initialize empty modules on all non-zero ranks
+            param_init_fn=(
+                lambda module: module.to_empty(
+                    device=torch.device("cuda"), recurse=False
+                )
+                if not self._is_rank_zero
+                else None
+            ),
+        )
```
kartikayk, line 262: Sorry I missed this, but now that we are initializing on CPU across all ranks, we should set ```sync_module_states``` to False and ```param_init_fn``` to ```None```.

recipes/gemma_full_finetune_distributed.py
```diff
@@ -0,0 +1,503 @@
...
+                ignore_idx=self._loss_fn.ignore_index,
+            ),
+        )
+
+        if self._is_rank_zero:
+            log.info("Dataset and Sampler are initialized.")
+
+        return sampler, dataloader
+
+    def save_checkpoint(self, epoch: int) -> None:
```
kartikayk, line 341: @solitude-alive just want to make sure you've run this recipe to completion and verified that we can correctly save checkpoint?

torchtune/models/gemma/_model_utils.py
```diff
@@ -0,0 +1,22 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+from torchtune.modules import TransformerDecoder
+
+
+def tie_weight(model: TransformerDecoder) -> None:
```
kartikayk, line 10: Where is this used?

torchtune/models/gemma/rms_norm.py
```diff
@@ -0,0 +1,26 @@
...
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+import torch
+from torch import nn
+
+
+class GemmaRMSNorm(nn.Module):
```
kartikayk, line 11: Thanks for separating this out!

torchtune/models/gemma/transformer.py
```diff
@@ -0,0 +1,111 @@
...
+
+        # shape: [b, s, d]
+        h = self.tok_embeddings(tokens)
+
+        # TODO: Fix the masking logic to not rely on checking kv_cache
+        if seq_len > 1 and self.layers[0].attn.kv_cache is not None:
+            mask = torch.full(
+                (1, 1, seq_len, seq_len), float("-inf"), device=tokens.device
+            )
+            mask = torch.triu(mask, diagonal=curr_pos + 1)
```
kartikayk, line 96: Once you rebase, this logic will need to be updated. See how we setup the mask on main

recipes/gemma_full_finetune_distributed.py
solitude-alive: Yeah, I update them in the latest version.

recipes/gemma_full_finetune_distributed.py
solitude-alive: Yeah, I tested it and was able to delete the last shared weight and save it.

torchtune/models/gemma/_model_utils.py
solitude-alive: This was used in a previous version and now I removed it.

torchtune/models/gemma/rms_norm.py
solitude-alive: Thank you.

kartikayk (approved):
A huge thank you @solitude-alive for adding this functionality and also patiently addressing the many review comments. This functionality makes TorchTune better and we really appreciate all of your hard work. I'll merge this into MAIN, make a few small changes to the core recipes based on some upcoming changes and then add this to our README and cite you as the author. Thanks so much for all of the hard work!

torchtune/utils/_checkpointing/_checkpointer_utils.py
```diff
@@ -54,12 +57,19 @@ def safe_torch_load(checkpoint_path: Path, weights_only: bool = True) -> Dict[st
...
         )
+        if is_safetensors_file:
+            result = {}
+            with safe_open(checkpoint_path, framework="pt", device="cpu") as f:
+                for k in f.keys():
+                    result[k] = f.get_tensor(k)
+            state_dict = result
+        else:
+            state_dict = torch.load(
+                str(checkpoint_path), map_location="cpu", mmap=True, weights_only=True
```
jerryzh168, line 71: looks like `weights_only` arg is not passed around here?

solitude-alive: Oh yes, but I just looked at the latest version and it has been updated.
```
state_dict = torch.load(
                str(checkpoint_path),
                map_location="cpu",
                mmap=True,
                weights_only=weights_only,
            )
```

torchtune/utils/_checkpointing/_checkpointer_utils.py
```diff
@@ -54,12 +57,19 @@ def safe_torch_load(checkpoint_path: Path, weights_only: bool = True) -> Dict[st
     try:
         # convert the path into a string since pathlib Path and mmap don't work
         # well together
-        state_dict = torch.load(
-            str(checkpoint_path),
-            map_location="cpu",
-            mmap=True,
-            weights_only=weights_only,
+        is_safetensors_file = (
+            True if str(checkpoint_path).endswith(".safetensors") else False
```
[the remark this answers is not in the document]
solitude-alive: Yeah.

PR #631, opened 2024-04-01
Add Mistral models to recipe registry

joecummings:
#### Context
- Added Mistral models to recipe registry

#### Changelog
- Add Mistral models to _recipe_registry.py

#### Test plan
- $ tune ls
Screenshot 2024-04-01 at 1 57 10 PM


kartikayk (approved):
Thanks for adding this!

PR #633, opened 2024-04-01
Remove _copy_tensor from usage

rohan-varma:
#### Context
- `_copy_tensor` was not needed and was added due to a misunderstanding on how pytorch internally handles tensor reference counting. The thinking was that we're creating a linear layer and want to save the bias, so we just deep clone the bias before returning it. This however is inaccurate, and its sufficient to just return a reference to the bias. This is because torch will internally bump the refcount to the bias's storage by 2, and when the function goes out of scope, we'll still have a reference count of 1 due to the returned `bias` variable.
- This change already exists for `weight`.

#### Changelog
- Get rid of `_copy_tensor`

#### Test plan
- CI


kartikayk (approved):
Thanks for revisiting this!

Issue #637, opened 2024-04-02
Datasets follow-ups mega-issue


ebsmothers:
We changed a lot of logic around our instruct and chat dataset APIs in #624, but there are still a number of follow-ups discussed on that PR that we should address. These are listed (roughly) in my own descending order of priority. Except for the last one, which is last cause it's more related to configs than anything. cc @RdoubleA @kartikayk 

- Consider moving InstructTemplate to match ChatFormat API (i.e. List[Message] -> List[Message] instead of str -> str). [Comment](https://github.com/pytorch/torchtune/pull/624/files#r1547059028)
- Improve handling of train_on_input. E.g. currently it’s expected to be passed to our chat conversation formatter, which makes it hard to write custom ones. [Comment](https://github.com/pytorch/torchtune/pull/624/files#r1546837891)
- String mapping of chat format to corresponding class. [Comment](https://github.com/pytorch/torchtune/pull/624/files#r1547164327)
- Add validate_messages utility to ensure conversations are well-formed. [Comment](https://github.com/pytorch/torchtune/pull/624/files#r1547168552)
- Rename _transforms.py as it’s really more of formatters and the naming is a bit confusing.
- Consider consolidating column map and transform for instruct datasets.
- (More of a config issue but) it is hard to override datasets from CLI because we currently pass use_clean=True to Alpaca in our configs which is not accepted by other datasets and we don’t have a way to delete the config from CLI

RdoubleA:
Thanks for collecting all the issues here @ebsmothers. I can take this issue and try to address most of these.

On the first point, I am inclined to leave InstructTemplate as is until we have more user feedback. I'm worried that merging InstructTemplate with ChatFormat will be generalizing too early and lead to rigid/hacky code just to fit the merged abstraction.

ebsmothers:
> On the first point, I am inclined to leave InstructTemplate as is until we have more user feedback.

Yeah I agree with this. Lemme move it down the list a bit

RdoubleA:
Ok I think everything should be covered except for:

>Consider consolidating column map and transform for instruct datasets.

which I think needs more discussion

ebsmothers:
@RdoubleA thanks for tackling all these! Given the only remaining item is kind of an open discussion, I am gonna close this issue.

PR #639, opened 2024-04-02
Split alpaca_dataset to alpaca + alpaca_cleaned

RdoubleA:
## Context
One of the follow ups mentioned in #637 - we should have reasonable defaults set in builder functions and configs should only have specific overrides. `alpaca_dataset` was a strange case because almost all the time we were overriding `use_clean=True` and the default was `False` (which is reasonable since the alpaca dataset by default should return the original dataset). It didn't make sense to make `use_clean=True` as the default because that is essentially a different dataset. This is also supported by the fact that we've noticed differences in memory usage and performance when using the cleaned version compared to the original, likely due to different sample length distributions (thanks to @SLR722 for pointing this out). At this point, it makes sense to just have a separate builder for the cleaned dataset.

## Changelog
Split alpaca into `alpaca_dataset` and `alpaca_cleaned_dataset` and refactor callsites accordingly

## Test plan


ebsmothers (approved):
Should we also add alpaca_cleaned_dataset to `api_ref_datasets.rst`? Otherwise looks good!

PR #641, opened 2024-04-02
Add string to InstructTemplate, ChatFormat getters

RdoubleA:
## Context
The builders `chat_dataset` and `instruct_dataset` need an easy way to specify from the config which `InstructTemplate` / `ChatFormat` to use without requiring using a nested component. Here, we add utility functions to do this easily using existing config utils. Addresses part of #637.

## Test plan
`pytest tests/torchtune/config/test_config_utils.py`


ebsmothers (approved):
Thanks for this!

PR #642, opened 2024-04-02
Add weights_only flag to torchtune checkpointer

kartikayk:
#### Context
- For loading checkpoints associated with quantized models, we need to expose the weights_only flag in the high level load_checkpoint API. Doing so in this PR.

#### Changelog
- Modified safe_torch_download and load_checkpoint to take a bool which is set to True by default

#### Test plan
- Added a unit test

```
pytest tests/torchtune/utils/test_checkpointer.py
```

image


ebsmothers (approved):

joecummings:
Can you update the qlora config??

Issue #623, opened 2024-03-31
[RFC] DPO Integration into Torchtune


yechenzhi:
## Intro
Direct Preference Optimization (DPO) offers a compelling approach to align Large Language Models (LLMs) with human or AI preferences, diverging from conventional reinforcement-based alignment methods. DPO streamlines the process into a straightforward loss function, optimized over a dataset of preference pairs {(x, y_w, y_l)}, where x is a prompt, and y_w and y_l represent the preferred and dispreferred responses, respectively.
DPO serves as an efficient and effective baseline for the alignment of LLMs.

## objective
Integrating DPO into Torchtune with minimal reliance on third-party libraries.

##  pseudo code
[DPO](https://github.com/eric-mitchell/direct-preference-optimization/) is open-sourced on GitHub, which allows us to refer to their code for reference and inspiration.

### data preparation
We could add a dataset like [InstructDataset](https://github.com/pytorch/torchtune/blob/73647e26eb327c1e7dff6d6d12e4060c16c11da9/torchtune/datasets/_instruct.py#L18)
named PairdDataset, return samples that include 'chosen', 'rejected', 'prompt' etc.

### model 
We need a Policy and Ref model respectively( in Lora, we only need a Policy, and Ref can be with adapters turned off ), this can be done [here](https://github.com/pytorch/torchtune/blob/73647e26eb327c1e7dff6d6d12e4060c16c11da9/recipes/lora_finetune_single_device.py#L210).
We also need to concatenate the chosen and rejected inputs into a single tensor, to avoid  doing two forward passes.

### loss 
With the outputs, we then can write a loss_fn.

kartikayk:
Thanks so much for putting this together @yechenzhi! The high level details look good.

A couple of questions and notes:
- I think this should be it's own recipe and so you should create a new file within recipes to add this functionality. Will this support multi-device training?
- What models will this support? It would be good to start with a single model and then increase coverage once the initial code is reviewed and merged.
- How do you plan on verifying correctness of your implementation? Are there benchmarks you plan to replicate?
- We don't have the concept of a trainer in TorchTune, so you'll need to do some refactoring of the code into existing utilities (torchtune/utils) or create new ones.

I think a few more details will become clearer once you have the code in a fork. Looking forward to it! Thanks so much for adding this!

RdoubleA:
[...]

PR #645, opened 2024-04-03
DPO

yechenzhi:
#### Context
integrating DPO into Torchtune, more details see [here](https://github.com/pytorch/torchtune/issues/623) 

#### Changelog
- ...

#### Test plan
- ....


torchtune/datasets/_preference.py
```diff
@@ -0,0 +1,85 @@
...
+        **load_dataset_kwargs (Dict[str, Any]): additional keyword arguments to pass to `load_dataset`.
+    """
+
+    def __init__(
+        self,
+        tokenizer: Tokenizer,
+        source: str,
+        template: InstructTemplate,
+        transform: Optional[Callable] = None,
+        column_map: Optional[Dict[str, str]] = None,
```
RdoubleA, line 56: I am inclined to merge column_map and transform as they almost serve the same purpose (get data into standardized fields). Based on your dataset (lvwerra/stack-exchange-paired) it seems column_map is sufficient to map to "prompt, chosen, rejected" fields. @ebsmothers do you think we should drop one of these here?

torchtune/data/_tokenize.py
```diff
@@ -0,0 +1,151 @@
...
+    # add EOS token to end of answer
+    chosen_tokens["input_ids"].append(tokenizer.eos_id)
+    rejected_tokens["input_ids"].append(tokenizer.eos_id)
+
+    longer_response_length = max(len(chosen_tokens["input_ids"]), len(rejected_tokens["input_ids"]))
+
+    # if combined sequence is too long, truncate the prompt
+    for answer_tokens in [chosen_tokens, rejected_tokens, prompt_tokens]:
+        if len(answer_tokens["prompt_input_ids"]) + longer_response_length > max_length:
+            if truncation_mode == "keep_start":
```
RdoubleA, line 115: you should use the `torchtune.data._utils.truncate` utility and add this functionality if needed

torchtune/utils/collate.py
```diff
@@ -71,3 +71,20 @@ def padded_collate(
             value=padding_idx,
         )
     return input_ids, labels
+
+
+def padded_collate_dpo(
+    batch: List[TokenPair],
+    padding_idx: int = 0,
+    ignore_idx: int = -100,
+) -> Tuple[torch.Tensor, torch.Tensor]:
```
RdoubleA, line 80: Could you add docstrings and also explain why this is needed for DPO instead of the normal `padded_collate`?

RdoubleA (commented):
Thanks for creating this PR, exciting to see all the components to enable DPO come together! I took an initial pass and had some high level comments.

torchtune/data/_instruct_templates.py
```diff
@@ -145,3 +145,40 @@ def format(
 
         prompt = cls.template.format(dialogue=sample[key_dialogue])
         return prompt
+
+
+class StackExchangedPairedTemplate(InstructTemplate):
```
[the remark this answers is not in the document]
yechenzhi: Thanks for your great advice! I'll use 'format' to handle prompt, and '_column_map' to get responses.

torchtune/utils/collate.py
yechenzhi: I've documented the padded_collate_dpo function.
And, it seems that the input data formats for padded_collate and padded_collate_dpo are different – one expects a list of token pairs, while the other expects a list of dictionaries. Therefore, using padded_collate for the data formatted as dictionaries might not be suitable.

torchtune/modules/peft/lora.py
```diff
@@ -59,6 +59,7 @@ def __init__(
         self.use_bias = use_bias
         self._quantize_base = quantize_base
         weight, bias = self._create_weight_and_bias()
+        self.disabled = False # if to turn off LoRA
```
RdoubleA, line 62: Do you mind expanding on this comment to explain that this can be used in DPO for treating the lora adapters as the policy model and disabling it to treat the base model as the reference model?

torchtune/modules/peft/peft_utils.py
```diff
@@ -256,3 +256,11 @@ def get_merged_lora_ckpt(
         del state_dict[f"{module}.lora_a.weight"]
         del state_dict[f"{module}.lora_b.weight"]
     return state_dict
+
+
+def disable_adapter(model: nn.Module, disabled: bool) -> None:
```
RdoubleA, line 261: would be great to add a unit test for this and make sure lora weights are indeed not being used

torchtune/utils/collate.py
```diff
@@ -71,3 +71,57 @@ def padded_collate(
             value=padding_idx,
         )
     return input_ids, labels
+
+
+def padded_collate_dpo(
+    batch: List[TokenPair],
```
RdoubleA, line 77: hm, why haven't we been using this type across the codebase? @joecummings any thoughts on keeping this abstraction or getting rid of it?

torchtune/utils/collate.py
```diff
@@ -71,3 +71,57 @@ def padded_collate(
...
+) -> Dict[str, torch.Tensor]:
+    """Pad a batch of sequences for Direct Preference Optimization (DPO).
+
+    This function takes a batch of sequences, where each sequence is represented
+    as a dictionary with multiple key-value pairs. Each key corresponds to a different
+    sequence component, such as input_ids or labels. The function pads each sequence
+    component independently to the longest sequence length in the batch.
+
+    Args:
+        batch (List[TokenPair]): A list of dictionaries, where each dictionary
```
RdoubleA, line 89: this should be `List[Dict[str, Any]]`?

torchtune/utils/collate.py
```diff
@@ -71,3 +71,57 @@ def padded_collate(
...
+            represents a sequence with multiple components including 'chosen_input_ids',
```
RdoubleA, line 90: nit: I would make the language stronger here saying that these fields are required in the batch

torchtune/utils/collate.py
```diff
@@ -71,3 +71,57 @@ def padded_collate(
             value=padding_idx,
         )
     return input_ids, labels
+
+
+def padded_collate_dpo(
+    batch: List[TokenPair],
+    padding_idx: int = 0,
+    ignore_idx: int = -100,
```
RdoubleA, line 79: nit: could you update this and the normal `padded_collate` to use `torchtune.data._common.CROSS_ENTROPY_IGNORE_IDX`? (this might need to be exposed as public API in `torchtune/data/__init__.py`)

recipes/lora_dpo_single_device.py
```diff
@@ -0,0 +1,629 @@
...
+    validate_state_dict_for_lora,
+    disable_adapter,
+)
+from torchtune.recipe_interfaces import FTRecipeInterface
+from tqdm import tqdm
+
+log = utils.get_logger("DEBUG")
+
+
+class LoRAFinetuneRecipeSingleDevice(FTRecipeInterface):
```
RdoubleA, line 34: let's rename this to `LoRADPORecipeSingleDevice`?

recipes/lora_dpo_single_device.py
```diff
@@ -0,0 +1,629 @@
...
+        self.seed = utils.set_seed(seed=cfg.seed)
+        self.epochs_run = 0
+        self.total_epochs = cfg.epochs
+        self.max_steps_per_epoch = cfg.max_steps_per_epoch
+        self.total_training_steps = 0
+
+        self._resume_from_checkpoint = cfg.resume_from_checkpoint
+        self._gradient_accumulation_steps = cfg.gradient_accumulation_steps
+
+    def load_checkpoint(self, cfg: DictConfig) -> Dict[str, Any]:
```
RdoubleA, line 103: nit: this should be updated in our other recipes, but let's rename this so we can distinguish between the checkpointer `DictConfig` object and the whole config `DictConfig` object
```suggestion
    def load_checkpoint(self, cfg_checkpointer: DictConfig) -> Dict[str, Any]:
```

recipes/lora_dpo_single_device.py
```diff
@@ -0,0 +1,629 @@
...
+            resume_from_checkpoint=self._resume_from_checkpoint,
+        )
+        checkpoint_dict = self._checkpointer.load_checkpoint()
+
+        if self._resume_from_checkpoint:
+            if utils.ADAPTER_KEY not in checkpoint_dict:
+                raise ValueError(
+                    "Adapter weights not found. Please ensure a valid adapter checkpoint is provided."
+                )
+            # _update_recipe_state will throw an exception if the recipe state is not corrctly loaded
```
RdoubleA, line 120:
```suggestion
            # _update_recipe_state will throw an exception if the recipe state is not correctly loaded
```

recipes/configs/llama2/7B_qlora_dpo_single_device.yaml
```diff
@@ -0,0 +1,82 @@
...
+optimizer:
+  _component_: torch.optim.AdamW
+  weight_decay: 0.01
+  lr: 3e-4
+lr_scheduler:
+  _component_: torchtune.modules.get_cosine_schedule_with_warmup
+  num_warmup_steps: 100
+
+loss:
+  _component_: torch.nn.CrossEntropyLoss
```
RdoubleA, line 65: You need to make a preference loss function, right?

recipes/lora_dpo_single_device.py
```diff
@@ -0,0 +1,629 @@
...
+
+        for k in batch:
+            if k.startswith("chosen") and isinstance(batch[k], torch.Tensor):
+                if "labels" in k:
+                    pad_value = label_pad_token_id
+                elif k.endswith("_input_ids"):
+                    pad_value = padding_value
+
+                concatenated_key = k.replace("chosen", "concatenated")
+                concatenated_batch[concatenated_key] = utils.pad_to_length(batch[k], max_length, pad_value=pad_value)
```
RdoubleA, line 431: why are we padding again here? if the collater already has access to both `chosen_input_ids` and `rejected_input_ids`, can't this be handled in `padded_collate_dpo`?

recipes/lora_dpo_single_device.py
```diff
@@ -0,0 +1,629 @@
...
+                    ),
+                    dim=0,
+                ).to(device=device)
+        return concatenated_batch 
+
+    @staticmethod
+    def get_batch_logps(
+        logits: torch.FloatTensor,
+        labels: torch.LongTensor,
+        label_pad_token_id: int = -100,
```
RdoubleA, line 453: same comment about using `CROSS_ENTROPY_IGNORE_IDX`

recipes/lora_dpo_single_device.py
```diff
@@ -0,0 +1,629 @@
...
+        logits = logits[:, :-1, :]
+        loss_mask = labels != label_pad_token_id
+
+        labels[labels == label_pad_token_id] = 0
+        per_token_logps = torch.gather(logits.log_softmax(-1), dim=2, index=labels.unsqueeze(2)).squeeze(2)
+
+        return (per_token_logps * loss_mask).sum(-1)
+         
+
+    def dpo_loss(
```
RdoubleA, line 468: let's move this to `torchtune/modules/loss/dpo.py` (might need to create this)

recipes/lora_dpo_single_device.py
```diff
@@ -0,0 +1,629 @@
...
+            A tuple of three tensors: (losses, chosen_rewards, rejected_rewards).
+            The losses tensor contains the DPO loss for each example in the batch.
+            The chosen_rewards and rejected_rewards tensors contain the rewards for the chosen and rejected responses, respectively.
+        """
+        pi_logratios = policy_chosen_logps - policy_rejected_logps
+        
+        ref_logratios = reference_chosen_logps - reference_rejected_logps
+
+        pi_logratios = pi_logratios.to(self._device)
+        ref_logratios = ref_logratios.to(self._device)
```
ebsmothers, line 493: Do we need all these .to(device) calls? (I see there are a bunch further down in this method as well)

recipes/lora_dpo_single_device.py
```diff
@@ -0,0 +1,629 @@
...
+            losses = (
+                -F.logsigmoid(self.beta * logits) * (1 - self.label_smoothing)
+                - F.logsigmoid(-self.beta * logits) * self.label_smoothing
+            )
+        elif self.loss_type == "hinge":
+            losses = torch.relu(1 - self.beta * logits)
+        elif self.loss_type == "ipo":
+            # eqn (17) of the paper where beta is the regularization parameter for the IPO loss, denoted by tau in the paper.
+            losses = (logits - 1 / (2 * self.beta)) ** 2
+        elif self.loss_type == "kto_pair":
```
ebsmothers, line 509: Once we move this to a standalone module, we can consider loss_type as a callable instead of a string, then just define quick functions for each of these that we can pass in (not super important though). We can also consider whether we actually need to support all of these loss types, if one is overwhelmingly prevalent we could just start there.

recipes/lora_dpo_single_device.py
ebsmothers: Agreed, this seems like it'd benefit from being a standalone module initialized with e.g. beta, loss_type, and label_smoothing params.

ebsmothers (commented):
This is awesome progress! I left a bunch of comments, please let me know if any of them are unclear. I need to take a second closer look at the loss a bit later on as well, but otherwise things are shaping up nicely here.

torchtune/datasets/_preference.py
```diff
@@ -0,0 +1,116 @@
...
+
+    def _prepare_sample(self, sample: Mapping[str, Any]) -> Dict[str, List[int]]:
+        transformed_sample = self._transform(sample) if self._transform else sample
+        prompt = self.template.format(transformed_sample, self._column_map)
+
+        key_chosen = self._column_map["chosen"]
+        key_rejected = self._column_map['rejected']
+
+        chosen_message = [
+            Message(role="user", content=prompt, masked=True),
```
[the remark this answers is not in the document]
yechenzhi: To the best of my knowledge, DPO does not train on inputs, so I simply did not include the 'train_on_input' parameter.

recipes/lora_dpo_single_device.py
yechenzhi: Sure! Actually, I have the same question before, I'll move this to 'padded_collate_dpo'.

torchtune/utils/collate.py
RdoubleA: ok with leaving it in this file only or dropping it completely

recipes/lora_dpo_single_device.py
```diff
@@ -0,0 +1,629 @@
...
+
+                with torch.no_grad():
+                    disable_adapter(self._model, True)
+                    (
+                        reference_chosen_logps,
+                        reference_rejected_logps,
+                        _,
+                        _,
+                    ) = self.concatenated_forward(self._model, batch)
+                    disable_adapter(self._model, False)
```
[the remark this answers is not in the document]
RdoubleA: +1 to using it as a context manager

torchtune/datasets/_stack_exchanged_paired.py
```diff
@@ -0,0 +1,65 @@
...
+        >>> for batch in Dataloader(alpaca_ds, batch_size=8):
+        >>>     print(f"Batch size: {len(batch)}")
+        >>> Batch size: 8
+    """
+
+    return PreferenceDataset(
+        tokenizer=tokenizer,
+        source="lvwerra/stack-exchange-paired",
+        template=StackExchangedPairedTemplate(),
+        column_map={'prompt':'question','chosen':'response_j','rejected':'response_k'},
```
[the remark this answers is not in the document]
RdoubleA: nope, the template only knows what keys to expect and not what the actual column names are. so it's needs a mapping to tell the template that prompt is actually question, chosen is actually response_j, etc. It only works in this direction.

torchtune/modules/peft/peft_utils.py
yechenzhi: > would be great to add a unit test for this and make sure lora weights are indeed not being used

I conducted the test manually by loading the original base model (without LoRA configuration) and subsequently loaded a LoRA model. Then, I enabled disable_adapter to the LoRA model and verified that the outputs matched. Should I include this logic in the unit test?

RdoubleA: You can replicate this with a single LoRA linear layer [initialized with constants](https://github.com/pytorch/torchtune/blob/2ab2721e6a176c4a698da9eec1be253d14f908d3/tests/test_utils.py#L61) so it's easier to write up, but yes that sounds like a good unit test

recipes/lora_dpo_single_device.py
```diff
@@ -0,0 +1,513 @@
...
+            current_iteration + 1
+        ) % self._gradient_accumulation_steps == 0
+        return should_update_weights
+
+    def concatenated_forward(self, model, batch):
+        concatenated_input_ids, concatenated_labels = batch
+        concatenated_input_ids = concatenated_input_ids.to(self._device)
+        concatenated_labels = concatenated_labels.to(self._device)
+
+        len_chosen = concatenated_input_ids.shape[0] // 2
```
RdoubleA, line 389: This assumes that the dataloader is using the padded_collate_dpo collator and concatenates chosen + rejected batchwise. What if we swap out or modify this collator, will this break? Or maybe this scenario is very unlikely?

recipes/lora_dpo_single_device.py
```diff
@@ -0,0 +1,513 @@
...
+        """
+        Determines whether the weights should be updated on the current iteration or not.
+        True is returned either if we've accumulated gradients for enough steps.
+        """
+        should_update_weights = (
+            current_iteration + 1
+        ) % self._gradient_accumulation_steps == 0
+        return should_update_weights
+
+    def concatenated_forward(self, model, batch):
```
RdoubleA, line 384: please add type annotations

recipes/lora_dpo_single_device.py
```diff
@@ -0,0 +1,629 @@
...
+                    and (idx // self._gradient_accumulation_steps)
+                    == self.max_steps_per_epoch
+                ):
+                    break
+            
+                (
+                    policy_chosen_logps,
+                    policy_rejected_logps,
+                    policy_chosen_logits,
+                    policy_rejected_logits,
```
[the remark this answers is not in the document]
RdoubleA: maybe don't need to return the logits?

recipes/lora_dpo_single_device.py
```diff
@@ -0,0 +1,513 @@
...
+                    disable_adapter(self._model, True)
+                    (
+                        reference_chosen_logps,
+                        reference_rejected_logps,
+                        _,
+                        _,
+                    ) = self.concatenated_forward(self._model, batch)
+                    disable_adapter(self._model, False)
+
+                loss, chosen_rewards, rejected_rewards = self._loss_fn(
```
RdoubleA, line 460: do you need to return the rewards? what is it used for?

recipes/configs/llama2/7B_qlora_dpo_single_device.yaml
RdoubleA, line 1: you may also need to update the recipe registry here: https://github.com/pytorch/torchtune/blob/main/torchtune/_recipe_registry.py

torchtune/datasets/_preference.py
```diff
@@ -0,0 +1,116 @@
...
+        # in DPO: first trunctate prompts, then responses
+        chosen_input_ids, c_masks = self._tokenizer.tokenize_messages(
+            chosen_message, self.max_seq_len
+        )
+        chosen_labels = list(np.where(c_masks, CROSS_ENTROPY_IGNORE_IDX, chosen_input_ids))
+
+        rejected_input_ids, r_masks = self._tokenizer.tokenize_messages(
+            rejected_message, self.max_seq_len
+        )
+        rejected_labels = list(np.where(r_masks, CROSS_ENTROPY_IGNORE_IDX, rejected_input_ids))
```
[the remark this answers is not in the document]
yechenzhi: Yes, when both the chosen and rejected messages are lengthy, it might inadvertently truncate most of the rejected parts, which isn't ideal. Do you happen to have any alternative suggestions for this?

recipes/configs/llama2/7B_qlora_dpo_single_device.yaml
```diff
@@ -0,0 +1,82 @@
...
+# To launch on a single device, run the following command from root:
+#   tune --nnodes 1 --nproc_per_node 1 lora_finetune_single_device \
+#   --config 7B_qlora_single_device \
+#
+# You can add specific overrides through the command line. For example
+# to override the checkpointer directory while launching training
+# you can run:
+#   tune --nnodes 1 --nproc_per_node 1 lora_finetune_single_device \
+#   --config 7B_qlora_single_device \
+#   checkpointer.checkpoint_dir=<YOUR_CHECKPOINT_DIR>
```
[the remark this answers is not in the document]
yechenzhi: Thank you for bringing that to my attention. I wasn't aware of that limitation. There isn't a specific reason for choosing QLoRA; I simply selected it randomly from a recipe to get started. In light of wanting multi-device support, I'll make the adjustment to LoRA.

recipes/lora_dpo_single_device.py
```diff
@@ -0,0 +1,629 @@
...
+                    disable_adapter(self._model, True)
+                    (
+                        reference_chosen_logps,
+                        reference_rejected_logps,
+                        _,
+                        _,
+                    ) = self.concatenated_forward(self._model, batch)
+                    disable_adapter(self._model, False)
+                
+                loss, chosen_rewards, rejected_rewards = self.dpo_loss(policy_chosen_logps, policy_rejected_logps,
```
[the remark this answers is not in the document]
yechenzhi: Certainly. I envision them being utilized as metrics.I apologize for forgetting to include them earlier.

recipes/lora_dpo_single_device.py
yechenzhi: > This assumes that the dataloader is using the padded_collate_dpo collator and concatenates chosen + rejected batchwise. 

Yes, I also added an assertion [here](https://github.com/yechenzhi/torchtune/blob/7d08813f8af0d10847ea28ddb591c9b2fb3d3b80/torchtune/utils/collate.py#L121).

> What if we swap out or modify this collator, will this break? Or maybe this scenario is very unlikely?
From my perspective, the likelihood of this scenario transpiring is minimal. The function's designation as 'concatenated_forward' suggests an assumption that the inputs are concatenated. Should I add a comment to this piece of code?

recipes/lora_dpo_single_device.py
yechenzhi: same as metrics.

torchtune/modules/peft/peft_utils.py
yechenzhi: Could you please review the addition I made [here](https://github.com/yechenzhi/torchtune/blob/e84729793cef772dc33eab25032144912fbf5441/tests/torchtune/modules/peft/test_peft_utils.py#L408)? I don't often engage in writing unit tests.

recipes/configs/llama2/7B_qlora_dpo_single_device.yaml
```diff
@@ -0,0 +1,82 @@
...
+batch_size: 2
+
+# Optimizer and Scheduler
+optimizer:
+  _component_: torch.optim.AdamW
+  weight_decay: 0.01
+  lr: 3e-4
+lr_scheduler:
+  _component_: torchtune.modules.get_cosine_schedule_with_warmup
+  num_warmup_steps: 100
```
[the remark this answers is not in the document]
yechenzhi: I have adjusted the hyperparameters based on [TRL](https://github.com/huggingface/trl/blob/main/examples/research_projects/stack_llama_2/scripts/dpo_llama2.py), but the optimizer used in TRL is paged_adamw_32bit. Should I also switch to this optimizer?

recipes/lora_dpo_single_device.py
RdoubleA: Yes I think a comment explaining that it assumes inputs are chosen + rejected concatenated would be great

torchtune/datasets/_preference.py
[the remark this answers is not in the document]
ebsmothers: No this makes sense to me, mainly wanted to clarify my own understanding here.

torchtune/modules/loss/dpo.py
```diff
@@ -0,0 +1,89 @@
...
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+import torch
+import torch.nn as nn
+import torch.nn.functional as F
+
+
+class DPOLoss(nn.Module):
```
ebsmothers, line 11: I think it would also be nice to add a unit test for this class.

recipes/lora_dpo_single_device.py
```diff
@@ -0,0 +1,512 @@
...
+                            "lr": self._optimizer.param_groups[0]["lr"],
+                            "rewards/chosen": chosen_rewards.mean().cpu(),
+                            "rewards/rejected": rejected_rewards.mean().cpu(),
+                            "rewards/accuracies": reward_accuracies.mean().cpu(),
+                            "rewards/margins": (chosen_rewards - rejected_rewards).mean().cpu(),
+                            "log_probs/rejected": policy_rejected_log_probs.detach().mean().cpu(),
+                            "log_probs/chosen": policy_chosen_log_probs.detach().mean().cpu(),
+                            "logits/rejected": policy_rejected_logits.detach().mean().cpu(),
+                            "logits/chosen": policy_chosen_logits.detach().mean().cpu(),
+                            "gpu_resources": torch.cuda.memory_allocated(),
```
ebsmothers, line 472: I'm curious, do we need all of these in general? E.g. mean of logits feels like it may be covered better by log_probs anyways. (Btw I think we should think about a better way to make our metric logging more configurable in general, so it's not necessarily specific to your DPO recipe)

ebsmothers (commented):
Overall this is looking pretty good! I left a few more comments but no major concerns from my side. The main ask I have is around testing: are you able to kick off a run of this recipe and document the results in the PR's test plan? Please let me know if you need any assistance with this and I'll be happy to help out.

torchtune/modules/loss/dpo.py
```diff
@@ -0,0 +1,89 @@
...
+        reference_rejected_logps: torch.Tensor,
+    ) -> tuple[torch.Tensor, torch.Tensor, torch.Tensor]:
+        """
+        Compute the DPO loss for a batch of policy and reference model log probabilities.
+
+        Args:
+            policy_chosen_logps (torch.Tensor): Log probabilities of the policy model for the chosen responses. Shape: (batch_size,)
+            policy_rejected_logps (torch.Tensor): Log probabilities of the policy model for the rejected responses. Shape: (batch_size,)
+            reference_chosen_logps (torch.Tensor): Log probabilities of the reference model for the chosen responses. Shape: (batch_size,)
+            reference_rejected_logps (torch.Tensor): Log probabilities of the reference model for the rejected responses. Shape: (batch_size,)
```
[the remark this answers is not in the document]
yechenzhi: Actually, the shape of each tensor is just the batch size, should I change it to (batch_size)?

torchtune/modules/loss/dpo.py
yechenzhi: Thank you for your suggestion. I appreciate your advice on adding a unit test for this class. However, I must admit that I don't frequently involve myself in writing unit tests, and I have a few inquiries regarding this matter. Specifically concerning the DPOLoss unit test, when the inputs are randomized, how should one compute the expected output? Conversely, if the input remains constant, wouldn't the utility of the unit test diminish?

yechenzhi:
> are you able to kick off a run of this recipe and document the results in the PR's test plan? Please let me know if you need any assistance with this and I'll be happy to help out.

I run this recipe on a 24g 4090, here is the log:
loss:
image
rewards margin:
image
avg log probs:
image
avg logits:
image

And I also run DPO by using [TRL](https://github.com/huggingface/trl/blob/main/examples/research_projects/stack_llama_2/scripts/dpo_llama2.py), differences include:
1. in TRL, I use batch size 2, gradient_accumulation = 32; in Torchtune, batch size is 4, and gradient_accumulation is 16.
2. in TRL, fine-tuned weights inlcudes "q_proj", "v_proj", "k_proj", "out_proj", "fc_in", "fc_out", "wte", while in torchtune, we only fine-tune "q_proj", "v_proj".
3. lr scheduler is different. 

the training log closely resembles ours, with the only difference being a slightly lower loss, I'll update the training log once the training is completed.

An intriguing observation is that the log probabilities of the selected (preferred) responses decreased, while the margin of rewards increased. This phenomenon has also been noted in this [paper](https://arxiv.org/abs/2402.13228).  

Should any other results be provided?

recipes/lora_dpo_single_device.py
yechenzhi: > I'm curious, do we need all of these in general? E.g. mean of logits feels like it may be covered better by log_probs anyways.

I have the same doubts, too, you can check my [training log](https://github.com/pytorch/torchtune/pull/645#issuecomment-2041861215) to see the difference. (by the way I'm still confused)

torchtune/modules/loss/dpo.py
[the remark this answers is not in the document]
ebsmothers: Oh yeah sorry that's just a misunderstanding on my part. I think as written it's fine then, thanks!

ebsmothers:
> 
> 1. In TRL, I utilize a batch size of 2 with a gradient accumulation of 32. Conversely, in Torchtune, the batch size is 4, and the gradient accumulation is 16.
> 2. In TRL, the fine-tuned weights include "q_proj," "v_proj," "k_proj," "out_proj," "fc_in," "fc_out," and "wte," whereas in Torchtune, we only fine-tune "q_proj" and "v_proj."
> 3. The learning rate scheduler differs between the two approaches.
> 4. Prior to training DPO in TRL, I also execute SFT, which is not performed in Torchtune.

> 
> Comparing the training results in TRL to those in Torchtune, it appears that the second and fourth differences may have influenced the outcome. Do you have any suggestions or insights regarding this?

@yechenzhi thanks for this analysis! I am curious about the difference in the loss curves between trl and torchtune, it seems like the torchtune loss does not decrease as much as the trl one does. 

Regarding the differences in training you point out, for  (2) you can enable LoRA some of the missing modules to get closer to an apples-to-apples comparison. E.g. set `lora_attn_modules=['q_proj', 'k_proj', 'v_proj', 'output_proj']` and `apply_lora_to_mlp=True` in your config. This is still not perfectly 1:1, but should at least provide a better comparison.

For (3): what is the lr scheduler used in the trl run? 

For (4), is it possible to perform the trl run without doing SFT first? (Apologies if this last point doesn't make sense, personally I am not so familiar with the trl setup.)

yechenzhi:
> I am curious about the difference in the loss curves between trl and torchtune, it seems like the torchtune loss does not decrease as much as the trl one does.

@ebsmothers Yes, I found two reasons for this. Firstly, in TRL, it defaults to shuffling the data, but in TorchTune's training, I set shuffle to false. Perhaps due to the continuous appearance of similar prompts in the data, such as consecutive Russian questions and answers, it makes training more challenging. Secondly, there is a [typo](https://github.com/yechenzhi/torchtune/blob/c6a6a3cfca62eca6a0bba96b1b370da153f12719/recipes/lora_dpo_single_device.py#L207) in the code which causes the learning rate to decrease extremely slow after warming up.

> set lora_attn_modules=['q_proj', 'k_proj', 'v_proj', 'output_proj'] and apply_lora_to_mlp=True in your config.

I've done that, but if I don't shuffle the data, the loss goes up.

>For (3): what is the lr scheduler used in the trl run?

The typo mentioned earlier caused this issue.

>For (4), is it possible to perform the trl run without doing SFT first?
 
I tried that as well, and it turns out it didn't really matter.

Here is the updated training log after addressing the two issues:
loss:
image
reward accuracies:
image
rewards margins:
image

yechenzhi:
To note, for a fair comparison, I selected approximately 70,000 samples for the 1000-step training in both TRL and Torchtune. Initially, I chose the first 64,000 * 5 samples from the dataset, and then filtered out those exceeding a length of 1024, resulting in approximately 70,000 remaining samples. I did not include this code in the PR. If you would like to reproduce the loss curve changes provided by me, you can add the line 'self._data = self._data.select(range(64000 * 5))' before [here](https://github.com/yechenzhi/torchtune/blob/a6ab0a6735943aa7455c0fdcdf9ae2c242c97e5a/torchtune/datasets/_preference.py#L61).

ebsmothers:
> > I am curious about the difference in the loss curves between trl and torchtune, it seems like the torchtune loss does not decrease as much as the trl one does.
> 
> @ebsmothers Yes, I found two reasons for this. Firstly, in TRL, it defaults to shuffling the data, but in TorchTune's training, I set shuffle to false. Perhaps due to the continuous appearance of similar prompts in the data, such as consecutive Russian questions and answers, it makes training more challenging. Secondly, there is a [typo](https://github.com/yechenzhi/torchtune/blob/c6a6a3cfca62eca6a0bba96b1b370da153f12719/recipes/lora_dpo_single_device.py#L207) in the code which causes the learning rate to decrease extremely slow after warming up.
> 
> > set lora_attn_modules=['q_proj', 'k_proj', 'v_proj', 'output_proj'] and apply_lora_to_mlp=True in your config.
> 
> I've done that, but if I don't shuffle the data, the loss goes up.
> 
> > For (3): what is the lr scheduler used in the trl run?
> 
> The typo mentioned earlier caused this issue.
> 
> > For (4), is it possible to perform the trl run without doing SFT first?
> 
> I tried that as well, and it turns out it didn't really matter.
> 

@yechenzhi thank you for the thorough analysis here! The loss curves look much better now, and thank you for catching the bug on how we were counting steps in the LR scheduler when gradient accumulation was turned on.

ebsmothers (approved):
Thank you @yechenzhi for your diligence throughout the code review and testing process! This is a major feature and we are very appreciative of you for contributing it to our library. I will do a couple checks on my end then merge this, we can take care of any smaller follow-ups on our end as needed. We are also planning to update our README soon, once we do we will cite you as the author of this recipe.

PR #647, opened 2024-04-03
Validate messages

RdoubleA:
## Context
Since the user is responsible for formatting their dataset into the `Messages` class, we need to add some quick validation to ensure that it is well-formed. For example, system prompt is at the beginning, assistant is always after a user message, no repeated user messages, etc. See #637 for context.

## Changelog
- Add `validate_messages` and test
- Change `_transforms.py` to `_converters.py`

## Test plan
`pytest tests/torchtune/data/test_data_utils.py`


rohan-varma (approved):
LG

PR #656, opened 2024-04-04
Fix packaging for WHL format

joecummings:
### Context

Join me, on my descent into madness.

We have a problem: **we want users to be able to copy and run our built-in recipes with our built-in configs without requiring users to clone the repository**. That means that these recipe files and config files need to be available in the [wheel package](https://packaging.python.org/en/latest/overview/#python-binary-distributions) (the preferred and default distribution format for Python packages). Alright, so how do we make sure these files are included in the end wheel package? This is where things get more complicated. 

We've made the conscious decision to formulate the relationship between recipes and torchtune as follows: **recipes build upon the reusable components in torchtune**. Therefore, recipes should not be an importable Python package. So how do we ensure that recipes are not an importable package? **We remove them from discovery via our packaging tools**. Now, we have solved one problem and created two others: we can no longer import anything from recipes for testing **AND** the recipe and config files are no longer included in our package, which means we cannot copy or run them.

New Problem A (Testing): This was actually solved via some awesome work by @ebsmothers in #537. The solution was essentially to utilize a subprocess to call the recipes via the command line, similar to how a user would do it. Therefore, we had no reason to import tests anymore. 

New Problem B (Missing files in distribution): The core of this question is actually **how do we include non-package data in a package?** Surely there must be some command or specification that you can turn on to include information. Introducing, the [MANIFEST.in file](https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#manifest-in)! We can simply add any files we want to include in the distribution with a command like `include recipe_a.py` and it will be added. For awhile, this seemed to solve all our problems and all was good and beautiful in the magical kingdom of TorchTune. But **this was all a lie**, for while this method worked while creating a TAR-format distribution, it failed when tested with a WHL-format. Two seconds of digging revealed an immediate problem: `MANIFEST.in does not affect binary distributions such as wheels.`. Okay, okay, so we just needed to find another way to package non-package data, right? Turns out, PyPA really really reaaaaaaally [does not want us to do this](https://setuptools.pypa.io/en/latest/userguide/datafiles.html#non-package-data-files) and therefore deprecated this ability. I spent some time trying to hack around this limitation, to no concrete avail.

> I was actually able to find a way to use deprecated commands to include non-package data in the final WHL, but there was no way to easy *find* these files using the distributed package. Also, relying on deprecated commands is ... not ideal.

New Problem B-revised (Missing files in **WHL**): When your back is against the wall, I always recommend giving up. In this case, that meant stopping the attempt to fight PyPA and do what they want: **recipes must be included as a package.** But author - won't users then be able to import from recipes and run amuck of our beautiful code? Yes, dear user, that is the implication. But, when all seems lost, there emerges a hero: @NicolasHug. A little while ago, he worked on   #594. Although this PR's primary purpose was to prevent tests from being packaged, he incorporated a "hack", which allowed recipes to be "packaged" by a local install, but raised a deliberate `ModuleNotFound` error if any errant user attempted to import something from recipes. The principle remains the same for packaging the recipes in a wheel. 

**To summarize**: This PR adds the recipes folder to the list of folders packaged into Torchtune.

#### FAQs

**What is the PyPA?**

Python packaging standards are maintained by [PyPA](https://www.pypa.io/en/latest/) - the Python Packaging Authority. Packaging in Python has gone through a number of changes in the last years. First, users could use [distutils](https://docs.python.org/3.10/library/distutils.html) to build and package their Python application. Then distutils was deprecated, and [setuptools](https://setuptools.pypa.io/en/latest/) became dominant. Now, there are several open-source build-backends like [poetry](https://python-poetry.org/) and [hatch](https://hatch.pypa.io/latest/). The point I'm trying to convey here is that **Python packaging is constantly evolving**. My goal is attempt to follow the PyPA's [advice](https://packaging.python.org/en/latest/) on packaging to set us up for success in the future. I aim to make the most sustainable and robust changes for the repo, rather than try to fit to a fad or newest standard. 

**Why are there other changes in this PR? Those does seem related to the problem above**

While going deep on all the proper Python packaging standards, I discovered that the recommended way to configure the packaging of an application is actually through the TOML file, rather than the setup.py file. Therefore, I made the necessary changes, which also allowed us to delete some extra top-level files. 

### Changelog
- Move all information in setup.py to pyproject.toml
- Added recipes to included directory in package
- Added version to torchtune __init__

### Test plan

- [x] Confirm WHL looks correct

1. `pip install build` # install build tool
2. `python -m build` # create WHL
3. Create a fresh conda env
4. `pip install torchtune/dist/torchtune-0.0.1-py3-none-any.whl` # install the built package
5. `tune cp full_finetune_single_device .` # confirm the files are there and we can locate them
6. `python -c "import recipes"` # confirm this throws an error
7. `python -c "import tests"` # confirm this throws an error

- [x] CI Green
- [x] Can be packaged and uploaded via test-infra workflows (see #663)


ebsmothers (approved):
incredible

NicolasHug:
Thanks Joe for all this work.

So with this PR, the `recipes` package gets installed when doing `pip install torchtune`, but `import recipes` throws an error with a decent error message.

At this point, and to avoid polluting the users' importable space with an unusable `recipes` package, should we just put `recipes` within `torchtune/recipes`? We can still raise the same error message, and we can still have the `recipes` folder be a top-level folder by just having a symlink? This way `recipes` isn't a package anymore, only `torchtune.recipes` exists.

----

Alternatively, I think we'd be fine by just providing sdists instead of wheels. Looking at the [Advantages of wheels](https://realpython.com/python-wheels/#advantages-of-python-wheels) that you shared offline, I don't think that any of these points are of importance for torchtune (at least at this stage). It wouldn't be too hard for us to not rely on the `test-infra` setup and just build/ship our own sdists.

----

> I discovered that the recommended way to configure the packaging of an application is actually through the TOML file, rather than the setup.py file. Therefore, I made the necessary changes, which also allowed us to delete some extra top-level files.

Sort of... But the majority of other pytorch projects still rely on `setupy.py` and you might get better support from our teams sticking to it ([internal link](https://fb.workplace.com/groups/pytorch.dev/posts/1397505970827879) for more context). Maybe @atalman can chime in on that aspect?

(Regardless, if it's not too much work/pain, it might be best to do the transition to pyproject.toml on a separate PR to really understand what's going on with atomic changes)

joecummings:
> At this point, and to avoid polluting the users' importable space with an unusable `recipes` package, should we just put `recipes` within `torchtune/recipes`? We can still raise the same error message, and we can still have the `recipes` folder be a top-level folder by just having a symlink? This way `recipes` isn't a package anymore, only `torchtune.recipes` exists.

Forgive my ignorance, but would a symlink work the same way from both a `git clone` installation and a `pip` installation? I want to make sure the user experience and how someone references the recipes is the same in both cases.

> Alternatively, I think we'd be fine by just providing sdists instead of wheels. Looking at the [Advantages of wheels](https://realpython.com/python-wheels/#advantages-of-python-wheels) that you shared offline, I don't think that any of these points are of importance for torchtune (at least at this stage). It wouldn't be too hard for us to not rely on the `test-infra` setup and just build/ship our own sdists.

This is a good point, and although we would eventually want a WHL, I can see that it's not necessary right now. The sticking point from my perspective is that I have a lot more confidence in building and uploading entirely via test-infra's setup, which has a plethora of testing, rather than my own custom workflow. (You can see my attempt at this in the history of commits on #663) 

> Sort of... But the majority of other pytorch projects still rely on `setupy.py` and you might get better support from our teams sticking to it ([internal link](https://fb.workplace.com/groups/pytorch.dev/posts/1397505970827879) for more context). Maybe @atalman can chime in on that aspect? (Regardless, if it's not too much work/pain, it might be best to do the transition to pyproject.toml on a separate PR to really understand what's going on with atomic changes)

Thanks for sharing the link! Regardless of what we go with, you're absolutely right, I should move this change to a separate PR.

NicolasHug:
> Forgive my ignorance, but would a symlink work the same way from both a git clone installation and a pip installation? I want to make sure the user experience and how someone references the recipes is the same in both cases.

Not sure I understand the concern completely but from my current understanding, this is a non-issue. When/if we move the recipes in `torchtune/recipes` (or even `torchtune/_recipes`), all the CLI tools and stuff would be adapted to rely on that. Then we can add a `recipes` symlink from the project root that points to `torhtune/[_]recipes/`, but the goal of that symlink is *only* for the recipes to be visible for users just browsing the repo on the web. It doesn't/shouldn't change anything with how recipes are used by torchtune in general. Or am I missing something?

> But the majority of other pytorch projects still rely on setupy.py and you might get better support from our teams sticking to it ([internal link](https://fb.workplace.com/groups/pytorch.dev/posts/1397505970827879) for more context). Maybe @atalman can chime in on that aspect?

BTW Andrey confirmed offline that relying on `pyproject.roml` this is OK with releng, so no issue here.

joecummings:
> Not sure I understand the concern completely but from my current understanding, this is a non-issue. When/if we move the recipes in `torchtune/recipes` (or even `torchtune/_recipes`), all the CLI tools and stuff would be adapted to rely on that. Then we can add a `recipes` symlink from the project root that points to `torhtune/[_]recipes/`, but the goal of that symlink is _only_ for the recipes to be visible for users just browsing the repo on the web. It doesn't/shouldn't change anything with how recipes are used by torchtune in general. Or am I missing something?

Nope, that makes sense! I filed #667 to track this.

PR #657, opened 2024-04-04
Render profiler component in docs

rohan-varma:
#### Context
- 

#### Changelog
- ...

#### Test plan
- 
image
image


RdoubleA (approved):
All the other utils are not private, but `_profiler` is? Why is that? And is there any way to list the public API version in the docs?

rohan-varma:
@RdoubleA good point, let me look into that

PR #658, opened 2024-04-04
Remove nf4linear

rohan-varma:
#### Context
- We didn't end up using this meaningfully for the QLoRA work. It's only used to create a quantized tensor within a linear module, but in `lora.py`, we just ended up creating this class and reading the quantized weight. Instead we can just build the quantized weight directly from the unquantized linear layer, via `to_nf4`. 

#### Changelog
- ...

#### Test plan
- Test components: `pytest tests/torchtune/modules/peft`


RdoubleA (approved):
LGTM, could you add the lora tests to the test plan?

PR #660, opened 2024-04-05
Improve Wandb experience

tcapelle:
The idea is create a better W&B experience for the end user
- Log the config used, taking into account the param override at runtime. We dump the OmegaConf inside the `checkpoint_dir` and save the file. This could also be added to other `metric_loggers`. We also keep track of the config in the Overview tab by updating the underlying `wandb.config`. We also add the run-id to the config filename, so they don't automatically overwrite. The naming is `f"torchtune_config_{self._wandb.run.id}.yaml"` at this time.
- I also updated the `memory_stats_log` function to return a dict so we can also log that to W&B. Also only log memory stats when training on GPU (some recipe tests use CPU)
- Haven't added artifact logging, this is useful for corporate users to keep track of their models and hav lineage.


msaroufim:
@tcapelle holler at me when you need a review

By QPS I mean how many elements and batches we're processing per second, right now tqdm provides some signal for that

tcapelle:
RIght now I am doing the following:
- When the `WandBLogger` get's created I inspect the `sys.argv` and grab the input to `--config` (don't make this a positional arg 🤣)
- Once I have the file, I load it using `yaml.safe_load` so I can update the `wandb.config`
- Then I copy the config to a temporary file and save it besides the stuff that W&B stores under Files (requirements, the actual script, etc..), and example [here](https://wandb.ai/capecape/torchtune/runs/gp74q2b4/files?nw=nwusercapecape)

tcapelle:
> @tcapelle holler at me when you need a review
> 
> By QPS I mean how many elements and batches we're processing per second, right now tqdm provides some signal for that

Regarding this, I don't think it is at the callback level, but I can try to figure that out on the library. I will raise a new PR for that.

tcapelle:
I would like to know @msaroufim if the Stateless system for the Metric Logging is fixed. This makes hard to connect the training loop logics with the logging, for instance, to save the model checkpoints to W&B I would love to have access to the checkpointer class instance, so I don't need to re-invent the wheel. 
Any guidance on this would helpful.

torchtune/utils/metric_logging.py
```diff
@@ -166,9 +176,29 @@ def __init__(
             group=group,
             reinit=True,
             resume="allow",
-            config=kwargs,
+            **kwargs,
         )
 
+    def log_config(self, config: OmegaConf) -> None:
```
RdoubleA, line 182: the config should be typed as `DictConfig`

torchtune/utils/metric_logging.py
```diff
@@ -36,7 +40,13 @@ def log(
             step (int): step value to record
         """
         pass
-
+    def log_config(self, config: OmegaConf) -> None:
```
RdoubleA, line 43: same as comment below

docs/source/examples/wandb_logging.rst
```diff
@@ -0,0 +1,137 @@
...
+Logging Model Checkpoints to W&B
+-------------------------------
+
+You can also log the model checkpoints to W&B by modifying the desired script `save_checkpoint` method. 
+
+A suggested approach would be something like this:
+
+.. code-block:: python
+
+      def save_checkpoint(self, epoch: int) -> None:
```
RdoubleA, line 97: I would trim the code here to only include the necessary information needed to convey how to log the checkpoints to wandb, i.e., I would remove the optimizer in bwd logic, maybe just assume we are already checkpointing the optimizer as well, etc.

recipes/configs/llama2/7B_lora_single_device.yaml
```diff
@@ -21,6 +21,8 @@
 # This config works only for training on single device.
 
 
+# export PATH="/home/ubuntu/miniforge3/envs/pt/bin:$PATH"
```
RdoubleA, line 24: do you need to include this?

recipes/configs/llama2/7B_lora_single_device.yaml
```diff
@@ -76,9 +78,11 @@ compile: False
 # Logging
 output_dir: /tmp/lora_finetune_output
 metric_logger:
-  _component_: torchtune.utils.metric_logging.DiskLogger
-  log_dir: ${output_dir}
-log_every_n_steps: null
+  _component_: torchtune.utils.metric_logging.WandBLogger
+  project: torchtune
+  kwarg1: 1
```
RdoubleA, line 83: what are these for :)

torchtune/models/llama2/_model_builders.py
```diff
@@ -21,6 +21,27 @@
 llama2 7B model.
 """
 
+def llama2_tiny() -> TransformerDecoder:
+    """
+    Creates a very small Llama2 model for testing purposes.
```
RdoubleA, line 26: Thank you for adding this! Mind adding what the parameter count for this is?

torchtune/models/llama2/_model_builders.py
```diff
@@ -21,6 +21,27 @@
...
+
+    Args:
+        max_batch_size (Optional[int]): Maximum batch size to be passed to KVCache.
+
+    Returns:
+        TransformerDecoder: Instantiation of Llama2 7B model
```
RdoubleA, line 32: This is no longer the 7B model

torchtune/models/llama2/_model_builders.py
```diff
@@ -21,6 +21,27 @@
 llama2 7B model.
 """
 
+def llama2_tiny() -> TransformerDecoder:
+    """
+    Creates a very small Llama2 model for testing purposes.
+
+    Args:
```
RdoubleA, line 28: there are no args?

torchtune/utils/memory.py
```diff
@@ -160,37 +160,33 @@ def optim_step(param) -> None:
...
-
     if reset_stats:
         torch.cuda.reset_peak_memory_stats(device)
 
-    return ret
+    memory_stats = {
+        "peak_memory_active": peak_memory_active,
+        "peak_memory_alloc": peak_mem_alloc,
+        "peak_memory_reserved": peak_mem_reserved,
+    }
```
RdoubleA, line 191: 👌

torchtune/utils/metric_logging.py
```diff
@@ -166,9 +176,29 @@ def __init__(
...
         )
 
+    def log_config(self, config: OmegaConf) -> None:
+        "Logs the config to W&B. Also updates config on overview tab."
+
+        resolved = OmegaConf.to_container(config, resolve=True)
+        self._wandb.config.update(resolved)
+        
+        output_config_fname = Path(os.path.join(
+            config.checkpointer.checkpoint_dir,
```
RdoubleA, line 189: Can we assume that this field will always exist in the config? I think it's fair to say this is required for any fine-tuning recipe, but I wonder if you'd ever want to log the config to wandb in non-finetuning recipes? Or maybe we should just pass the checkpoint dir as a parameter? cc @kartikayk

torchtune/utils/metric_logging.py
```diff
@@ -166,9 +176,29 @@ def __init__(
...
+            f"torchtune_config_{self._wandb.run.id}.yaml"
+            )
+        )
+        OmegaConf.save(config, output_config_fname)
+        try:
+            
+            print(f"Logging {output_config_fname} to W&B under Files")
+            self._wandb.save(output_config_fname, base_path=output_config_fname.parent)
+
+        except Exception as e:
```
RdoubleA, line 199: Is there a more specific error(s) we can catch?

torchtune/utils/metric_logging.py
```diff
@@ -166,9 +176,29 @@ def __init__(
...
+        
+        output_config_fname = Path(os.path.join(
+            config.checkpointer.checkpoint_dir, 
+            f"torchtune_config_{self._wandb.run.id}.yaml"
+            )
+        )
+        OmegaConf.save(config, output_config_fname)
+        try:
+            
+            print(f"Logging {output_config_fname} to W&B under Files")
```
RdoubleA, line 196: Use the debug logger instead of print:
```
# At the top right after imports and before class definitions
log = utils.get_logger("DEBUG")
...
# In methods
log.info(...)
```

torchtune/utils/metric_logging.py
```diff
@@ -166,9 +176,29 @@ def __init__(
...
+            self._wandb.save(output_config_fname, base_path=output_config_fname.parent)
+
+        except Exception as e:
+            print(f"Error saving {output_config_fname} to W&B.\nError: \n{e}")
```
RdoubleA, line 200: If you use the debug logger mentioned above, you can make this log a warning if you don't want to raise an error: `log.warning("Error saving to W&B")`

torchtune/utils/metric_logging.py
```diff
@@ -166,9 +176,29 @@ def __init__(
             group=group,
             reinit=True,
             resume="allow",
-            config=kwargs,
+            **kwargs,
         )
 
+    def log_config(self, config: OmegaConf) -> None:
+        "Logs the config to W&B. Also updates config on overview tab."
```
RdoubleA, line 183: This method is a lot more straightforward, thanks for iterating on this 👌

recipes/full_finetune_single_device.py
```diff
@@ -231,11 +234,9 @@ def _setup_model(
...
             log.info("Compiling model with torch.compile...")
             model = utils.wrap_compile(model)
-        log.info(
-            utils.memory_stats_log(
-                "Memory Stats after model init:", device=self._device
-            )
-        )
+        memory_stats = utils.memory_stats_log(device=self._device)
+        log.info(f"Memory Stats:\n{memory_stats}")
+        log.info(f"Model trainable parameters: {sum(p.numel() for p in model.parameters() if p.requires_grad)/1e6:,.2f}M")
```
ebsmothers, line 239: Can we remove this line?

recipes/full_finetune_single_device.py
```diff
@@ -231,11 +234,9 @@ def _setup_model(
         if compile_model:
             log.info("Compiling model with torch.compile...")
             model = utils.wrap_compile(model)
-        log.info(
-            utils.memory_stats_log(
-                "Memory Stats after model init:", device=self._device
-            )
-        )
+        memory_stats = utils.memory_stats_log(device=self._device)
+        log.info(f"Memory Stats:\n{memory_stats}")
```
ebsmothers, line 238: nit: change back to "Memory Stats after model init:" just to be explicit

recipes/full_finetune_single_device.py
```diff
@@ -415,9 +416,9 @@ def train(self) -> None:
 
                 # Log peak memory for iteration
                 if self.total_training_steps % self._log_peak_memory_every_n_steps == 0:
-                    log.info(
-                        utils.memory_stats_log("Memory Stats:", device=self._device)
-                    )
+                    memory_stats = utils.memory_stats_log(device=self._device)
+                    self._metric_logger.log_dict(memory_stats, step=self.total_training_steps)
+                    log.info(f"Memory Stats:\n{memory_stats}")
```
ebsmothers, line 421: So right now we log to stdout only when WandB is not enabled, but log to both WandB and stdout when it is enabled? Don't all our metric loggers support log_dict? If so, can we just call `log_dict` only here?

torchtune/utils/metric_logging.py
```diff
@@ -147,9 +160,10 @@ class WandBLogger(MetricLoggerInterface):
 
     def __init__(
         self,
-        project: str,
+        project: str = "torchtune",
```
ebsmothers, line 163: Ah nice, I've been setting this manually in all my runs for so long 😅

torchtune/utils/metric_logging.py
```diff
@@ -147,9 +160,10 @@ class WandBLogger(MetricLoggerInterface):
...
         entity: Optional[str] = None,
         group: Optional[str] = None,
+        log_strategy: Optional[str] = "main",
```
ebsmothers, line 166: nit: could type as a literal, like we do e.g. [here](https://github.com/pytorch/torchtune/blob/e7f1a6d8e7b7611a2528e97e594242453d8ffd86/torchtune/data/_types.py#L10)

recipes/configs/llama2/tiny_llama.yaml
```diff
@@ -0,0 +1,88 @@
+# Config for single-device full finetuning
```
ebsmothers, line 1: Can remove this file now?

torchtune/utils/memory.py
ebsmothers: Yes, I've had to hack this locally a few times for my own purposes. Really glad to see that we're supporting this properly now!

torchtune/utils/memory.py
```diff
@@ -160,37 +160,33 @@ def optim_step(param) -> None:
         p.register_post_accumulate_grad_hook(optim_step)
 
 
-def memory_stats_log(
-    prefix: str, device: torch.device, reset_stats: bool = True
-) -> None:
+def memory_stats_log(device: torch.device, reset_stats: bool = True) -> None:
```
ebsmothers, line 163: update return type

recipes/full_finetune_single_device.py
```diff
@@ -150,6 +150,9 @@ def setup(self, cfg: DictConfig) -> None:
         """
         self._metric_logger = config.instantiate(cfg.metric_logger)
 
+        # log config with parameter override
+        self._metric_logger.log_config(cfg)
```
ebsmothers, line 154: Do we plan to add this integration to our other recipes as well or just this one for now?

docs/source/examples/wandb_logging.rst
```diff
@@ -0,0 +1,69 @@
...
+      _component_: torchtune.utils.metric_logging.WandBLogger
+      # the W&B project to log to
+      project: torchtune
+
+We automatically grab the config from the recipe you are running and log it to W&B. You can find it in the W&B overview tab and the actual file in the `Files` tab.
+
+.. note::
+
+  Click on this sample [project to see the W&B workspace](https://wandb.ai/capecape/torchtune)
+  The config used to train the models can be found [here](https://wandb.ai/capecape/torchtune/runs/6053ofw0/files/torchtune_config_j67sb73v.yaml)
```
ebsmothers, line 36: If you want, you can even think about including an image in here just to show off some of the nice loss curves, memory profiling, etc. people can see.

docs/source/examples/wandb_logging.rst
```diff
@@ -0,0 +1,69 @@
...
+
+Logging Model Checkpoints to W&B
+-------------------------------
+
+You can also log the model checkpoints to W&B by modifying the desired script `save_checkpoint` method.
```
ebsmothers, line 41: nit
```suggestion
You can also log the model checkpoints to W&B by modifying the desired recipe's `save_checkpoint` method. 
```

docs/source/examples/wandb_logging.rst
```diff
@@ -0,0 +1,69 @@
...
+
+A suggested approach would be something like this:
+
+.. code-block:: python
+
+      def save_checkpoint(self, epoch: int) -> None:
+        ...
+        ## Let's save the checkpoint to W&B
+        ## depending on the Checkpointer Class the file will be named differently
+        ## Here it is an example for the full_finetune case
```
ebsmothers, line 51: nit
```suggestion
        ## Here is an example for the full_finetune case
```

recipes/full_finetune_single_device.py
ebsmothers: (Separate q: @RdoubleA can we easily support this for our other metric loggers as well?)

torchtune/utils/metric_logging.py
```diff
@@ -160,27 +174,63 @@ def __init__(
...
+                
+                log.info(f"Logging {output_config_fname} to W&B under Files")
+                self._wandb.save(output_config_fname, base_path=output_config_fname.parent)
+
+            except Exception as e:
+                log.warning(f"Error saving {output_config_fname} to W&B.\nError: \n{e}")
 
     def log(self, name: str, data: Scalar, step: int) -> None:
-        self._wandb.log({name: data}, step=step)
+        if self._wandb.run:
```
ebsmothers, line 220: Dumb q: what is this doing?

torchtune/utils/metric_logging.py
```diff
@@ -160,27 +174,63 @@ def __init__(
...
+        self.world_size, self.rank = get_world_size_and_rank()
+        self.local_rank = _get_local_rank()
+        self.local_rank = 0 if self.local_rank is None else self.local_rank
+        
+        if (
+                (self.log_strategy == "main" and self.rank == 0) 
+                or (self.log_strategy == "node" and self.local_rank == 0)
+                or self.log_strategy == "all"
+            ): 
+            self._wandb.init(
```
ebsmothers, line 189: missing indent?

torchtune/utils/metric_logging.py
```diff
@@ -160,27 +174,63 @@ def __init__(
...
+
+        # logging strategy options are "main", "node", "all"
+        self.log_strategy = log_strategy
+        self.world_size, self.rank = get_world_size_and_rank()
+        self.local_rank = _get_local_rank()
+        self.local_rank = 0 if self.local_rank is None else self.local_rank
+        
+        if (
+                (self.log_strategy == "main" and self.rank == 0) 
+                or (self.log_strategy == "node" and self.local_rank == 0)
```
ebsmothers, line 186: Can we disable "node" strategy for the time being? Main reason being that we are not testing our recipes very much on multi-node yet and I don't want people to see this and get the wrong impression that we officially support it at this time

torchtune/utils/metric_logging.py
```diff
@@ -160,27 +174,63 @@ def __init__(
...
-            group=group,
-            reinit=True,
-            resume="allow",
-            config=kwargs,
-        )
+
+        # logging strategy options are "main", "node", "all"
+        self.log_strategy = log_strategy
+        self.world_size, self.rank = get_world_size_and_rank()
+        self.local_rank = _get_local_rank()
```
ebsmothers, line 181: Probably worth integrating these changes into e.g. `full_finetune_distributed.py` as a test of this logic. Cause on single-device recipe these will all be equivalent, right?

torchtune/utils/metric_logging.py
```diff
@@ -126,6 +137,8 @@ class WandBLogger(MetricLoggerInterface):
         project (str): WandB project name
         entity (Optional[str]): WandB entity name
         group (Optional[str]): WandB group name
+        log_strategy (Optional[str]): Strategy to use for logging. Options are "main", "node", "all".
+            Default: "main"
```
ebsmothers, line 141: Would add more detail here explaining what each of these mean

ebsmothers (commented):
Thanks for working on this! Left a bunch of comments, please let me know if any of them are unclear. A couple other general things:

1) I think you need to run pre-commit hooks, you can follow the instructions [here](https://github.com/pytorch/torchtune/blob/main/CONTRIBUTING.md#coding-style) for setting that up.
2) It looks like the doc build job is failing. If it helps, you should be able to serve the docs locally for debugging by following the instructions [here](https://github.com/pytorch/torchtune/blob/main/CONTRIBUTING.md#build-docs)

torchtune/utils/metric_logging.py
RdoubleA: wandb won't be initialized if the main/node/rank 0 logic doesn't pass

recipes/full_finetune_single_device.py
RdoubleA: yes - we can leave that as a follow up

torchtune/utils/metric_logging.py
```diff
@@ -160,27 +174,63 @@ def __init__(
...
+        # logging strategy options are "main", "node", "all"
+        self.log_strategy = log_strategy
+        self.world_size, self.rank = get_world_size_and_rank()
+        self.local_rank = _get_local_rank()
+        self.local_rank = 0 if self.local_rank is None else self.local_rank
+        
+        if (
+                (self.log_strategy == "main" and self.rank == 0) 
+                or (self.log_strategy == "node" and self.local_rank == 0)
+                or self.log_strategy == "all"
```
RdoubleA, line 187: I'm inclined to make this a quick private method (e.g., `_should_log()`) since the logic here is more complex, that way you can test this logic in an isolated way

recipes/full_finetune_distributed.py
```diff
@@ -147,7 +147,11 @@ def setup(self, cfg: DictConfig) -> None:
         Sets up the recipe state correctly. This includes setting recipe attributes based
         on the ``resume_from_checkpoint`` flag.
         """
-        self._metric_logger = config.instantiate(cfg.metric_logger)
+        if self._is_rank_zero:
```
RdoubleA, line 150: you need to also update the cleanup method to check for rank zero before closing the logger

recipes/lora_finetune_single_device.py
```diff
@@ -447,9 +448,9 @@ def train(self) -> None:
                         self.total_training_steps % self._log_peak_memory_every_n_steps
                         == 0
                     ):
-                        log.info(
-                            utils.memory_stats_log("Memory Stats:", device=self._device)
-                        )
+                    # Log peak memory for iteration
```
RdoubleA, line 451: tab this code block?

recipes/full_finetune_distributed.py
```diff
@@ -451,9 +452,9 @@ def train(self) -> None:
                     self.total_training_steps % self._log_peak_memory_every_n_steps == 0
                     and self._is_rank_zero
                 ):
-                    log.info(
-                        utils.memory_stats_log("Memory Stats", device=self._device)
-                    )
+                # Log peak memory for iteration
```
RdoubleA, line 455: missing tab

RdoubleA (commented):
I think you'll need to run `pre-commit run --all-files` to fix some of the spacing and indent issues

recipes/lora_finetune_single_device.py
```diff
@@ -263,11 +267,9 @@ def _setup_model(
         if compile_model:
             log.info("Compiling model with torch.compile...")
             model = utils.wrap_compile(model)
-        log.info(
-            utils.memory_stats_log(
-                "Memory Stats after model init:", device=self._device
-            )
-        )
+        if self._device == torch.device("cuda"):
```
ebsmothers, line 270: This is for the CPU recipe tests?

docs/source/examples/wandb_logging.rst
```diff
@@ -0,0 +1,79 @@
...
+
+.. note::
+
+  You will need to install the `wandb`` package to use this feature.
+  You can install it via pip:
+
+  .. code-block:: bash
+
+    pip install wandb
+
```
joecummings, line 25: Can you add a "tip" to run `wandb login` before running?

recipes/lora_finetune_distributed.py
```diff
@@ -541,9 +541,12 @@ def train(self) -> None:
                 if (
                     self.total_training_steps % self._log_peak_memory_every_n_steps == 0
                     and self._is_rank_zero
+                    and self._device == torch.device("cuda")
```
ebsmothers, line 544: Do we really need this one too? For distributed tests they should only run on GPU

ebsmothers (approved):
Couple questions on the device == cuda checks (especially for distributed recipes). Otherwise looks good though

docs/source/examples/wandb_logging.rst
RdoubleA: good catch

recipes/lora_finetune_single_device.py
RdoubleA: yes, this actually wont throw the log just doesn't print anything

recipes/lora_finetune_distributed.py
RdoubleA: mm yeah I can remove this check for distributed recipes

PR #663, opened 2024-04-05
Add trigger for building Linux wheels

joecummings:
This uses the NOVA workflows from test-infra to build a Linux wheel.

Currently, I've set the ref to `pkg-for-wheel` for testing. This will need to be changed back before merging.

### DIFF
--- a/.github/workflows/build_linux_wheels.yaml
+++ b/.github/workflows/build_linux_wheels.yaml
@@ -63,14 +63,14 @@ jobs:
           repository: pytorch/test-infra
           ref: main
           path: test-infra
-      - uses: pytorch/test-infra/.github/actions/set-channel
+      - uses: pytorch/test-infra/.github/actions/set-channel/action.yml@main
       - name: Set PYTORCH_VERSION
         if: ${{ env.CHANNEL == 'test' }}
         run: |
           # When building RC, set the version to be the current candidate version,
           # otherwise, leave it alone so nightly will pick up the latest
           echo "PYTORCH_VERSION=${{ matrix.stable_version }}" >> "${GITHUB_ENV}"
-      - uses: pytorch/test-infra/.github/actions/setup-binary-builds
+      - uses: pytorch/test-infra/.github/actions/setup-binary-builds/action.yml@main
         env:
           PLATFORM: ${{ env.ARCH == 'aarch64'  && 'linux-aarch64' || ''}}
         with:
@@ -139,7 +139,7 @@ jobs:
 
   upload:
     needs: build
-    uses: pytorch/test-infra/.github/workflows/_binary_upload.yml
+    uses: pytorch/test-infra/.github/workflows/_binary_upload.yml@main
     if: always()
     with:
       repository: pytorch.torchtune
--- a/.github/workflows/build_linux_wheels.yaml
+++ b/.github/workflows/build_linux_wheels.yaml
@@ -142,7 +142,7 @@ jobs:
     if: always()
     with:
       repository: pytorch/torchtune
-      ref: ""
+      ref: pkg-for-wheel
       test-infra-repository: pytorch/test-infra
       test-infra-ref: main
       build-matrix: ${{ needs.build.needs.generate-matrix.outputs.matrix }}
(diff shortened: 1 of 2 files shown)


.github/workflows/build_linux_wheels.yaml
```diff
@@ -0,0 +1,155 @@
...
+          mkdir -p "${GITHUB_WORKSPACE}"
+
+          if [[ "${{ env.ARCH }}" = "aarch64" ]]; then
+            rm -rf "${RUNNER_TEMP}/*"
+          fi
+
+          echo "::endgroup::"
+      - uses: actions/checkout@v3
+        with:
+          # Support the use case where we need to checkout someone's fork
```
kartikayk, line 62: what does this mean?

.github/workflows/build_linux_wheels.yaml
```diff
@@ -0,0 +1,175 @@
...
+on:
+  pull_request:
+  push:
+    branches:
+      - nightly
+      - main
+      - release/*
+    tags:
+        # NOTE: Binary build pipelines should only get triggered on release candidate builds
+        # Release candidate tags look like: v1.11.0-rc1
```
ebsmothers, line 16: I don't understand this, aren't we also building nightlies here? (Sorry maybe this is a dumb question)

.github/workflows/build_linux_wheels.yaml
```diff
@@ -0,0 +1,175 @@
...
+    runs-on: ubuntu-latest
+    outputs:
+      TEST_INFRA_REPOSITORY: pytorch/test-infra
+      TEST_INFRA_REF: main
+      REPOSITORY: pytorch/torchtune
+      REF: pkg-for-wheel
+      PACKAGE_NAME: torchtune
+      PACKAGE_TYPE: wheel
+      ARCH: x86_64
+      SMOKE_TEST_SCRIPT: ""
```
ebsmothers, line 37: I think we had talked about potentially adding sanity checks of importing from the library or something like that here. Any plans to do that in this PR, or should we just do it later on?

.github/workflows/build_linux_wheels.yaml
```diff
@@ -0,0 +1,175 @@
...
+    steps:
+      - run: echo "Exposing env vars to all jobs"
+  generate-matrix:
```
ebsmothers, line 40: Any value to just adding a comment at the top of this job about which versions we should be building so folks don't have to go to the test-infra yaml to find this stuff?

.github/workflows/build_linux_wheels.yaml
```diff
@@ -0,0 +1,175 @@
...
+          echo "PYTORCH_VERSION=${{ matrix.stable_version }}" >> "${GITHUB_ENV}"
+      - uses: pytorch/test-infra/.github/actions/setup-binary-builds@main
+        env:
+          PLATFORM: ${{ env.ARCH == 'aarch64'  && 'linux-aarch64' || ''}}
+        with:
+          repository: ${{ needs.setup-variables.outputs.REPOSITORY }}
+          ref: ${{ needs.setup-variables.outputs.REF }}
+          setup-miniconda: true
+          python-version: ${{ env.PYTHON_VERSION }}
+          cuda-version: ${{ env.CU_VERSION }}
```
ebsmothers, line 99: Do we need this if we're not enabling CUDA builds?

.github/workflows/build_linux_wheels.yaml
```diff
@@ -0,0 +1,175 @@
...
+          source "${BUILD_ENV_FILE}"
+          # shellcheck disable=SC2086
+          ${CONDA_RUN} ${PIP_INSTALL_TORCH}
+      - name: Install python build dependency
+        run: |
+          set -euxo pipefail
+          # shellcheck disable=SC1090
+          source "${BUILD_ENV_FILE}"
+          # shellcheck disable=SC2086
+          ${CONDA_RUN} python -m pip install build
```
ebsmothers, line 114: So this will be supplanted by https://github.com/pytorch/test-infra/pull/5071/files? Will we be able to remove some of these other steps as well to rely directly on the test-infra version?

ebsmothers (approved):
No major concerns, accepting to unblock

PR #672, opened 2024-04-09
[EZ] cfg -> cfg_checkpointer in load_checkpoint

RdoubleA:
## Context
Be consistent with other setup methods that take in a DictConfig (`cfg_model`, `cfg_data`, etc) and update `cfg` -> `cfg_checkpointer` in `load_checkpoint` in all recipes

## Test plan
eyes and CI


joecummings (approved):
🫡

PR #674, opened 2024-04-09
Add Checkpointer Tutorial + Docs

kartikayk:
#### Context

We've gotten several round of questions on our checkpointer. In this PR I add a tutorial and render the APIs on our docs

#### Changelog
- Checkpointer Tutorial
- Checkpointer Docs
- Make HFCheckpointer the default in all llama2 configs

#### Test plan
- Ran all of the modified configs

Tutorial rendered correctly

Screenshot 2024-04-09 at 3 50 18 PM

Docs rendered correctly

Screenshot 2024-04-09 at 3 50 10 PM


docs/source/api_ref_utilities.rst
```diff
@@ -4,6 +4,19 @@ torchtune.utils
...
+
+Checkpointing
+-------------
+
+.. autosummary::
+    :toctree: generated/
+    :nosignatures:
+
+    FullModelHFCheckpointer
+    FullModelMetaCheckpointer
```
ebsmothers, line 18: Wait I thought we were doing away with `FullModel`?

docs/source/examples/checkpointer.rst
```diff
@@ -0,0 +1,319 @@
+.. _understand_checkpointer:
+
+==============================
+Understanding the Checkpointer
```
ebsmothers, line 4: Do we wanna add one of those "What you will learn"/"Prerequisites" headers (similar to e.g. our "Finetune your First LLM" tutorial)? Might be helpful to (a) standardize our format a bit, and (b) make sure readers have the appropriate context going in

docs/source/examples/checkpointer.rst
```diff
@@ -0,0 +1,319 @@
...
+--------
+
+TorchTune checkpointers are designed to be composable components which can be plugged
+into any recipe - training, evaluation or generation. Each checkpointer supports a
+set of models and scenarios making these easy to understand, debug and extend.
+
+TorchTune is designed to be "state-dict invariant".
+
+- At the input, TorchTune accepts checkpoints from multiple sources in multiple formats.
+  For Llama2 this includes both the HF Hub and the Meta Llama website. Model users don't
```
ebsmothers, line 21: nit: prefer to write to the reader (i.e. "Model users don't have to worry about" -> "You don't have to worry about")

docs/source/examples/checkpointer.rst
```diff
@@ -0,0 +1,319 @@
...
+Overview
+--------
+
+TorchTune checkpointers are designed to be composable components which can be plugged
+into any recipe - training, evaluation or generation. Each checkpointer supports a
+set of models and scenarios making these easy to understand, debug and extend.
+
+TorchTune is designed to be "state-dict invariant".
+
+- At the input, TorchTune accepts checkpoints from multiple sources in multiple formats.
```
ebsmothers, line 20: So one possible suggestion: do we wanna explicitly state the problem that this is solving up front? Like give an example of the usual `.load_state_dict()` flow and how it doesn't work when the model and weights are coming from different places, then show how the checkpointer allows us to just call `load_checkpoint()` without having to worry about it. (Btw I think the way you've laid it out is good too, this'd just be a slightly different exposition.)

docs/source/examples/checkpointer.rst
```diff
@@ -0,0 +1,319 @@
...
+
+
+Overview
+--------
+
+TorchTune checkpointers are designed to be composable components which can be plugged
+into any recipe - training, evaluation or generation. Each checkpointer supports a
+set of models and scenarios making these easy to understand, debug and extend.
+
+TorchTune is designed to be "state-dict invariant".
```
ebsmothers, line 18: Idk if it's too elementary, but could just give a sentence or two on state dicts and how they can differ (different keys, reshaped tensors, etc)

docs/source/examples/checkpointer.rst
ebsmothers: Also maybe I am nitpicking too much, but we could also just call this tutorial "Checkpoints in TorchTune". To me "Understanding X" indicates that X is inherently difficult to understand

docs/source/examples/checkpointer.rst
```diff
@@ -0,0 +1,319 @@
...
+
+- At the input, TorchTune accepts checkpoints from multiple sources in multiple formats.
+  For Llama2 this includes both the HF Hub and the Meta Llama website. Model users don't
+  have to worry about explicitly converting checkpoints every time they run a recipe.
+
+- At the output, TorchTune produces checkpoints in the same format as the source. This includes
+  converting the state_dict back into the original form and splitting the keys and weights
+  across the same number of files. As a result, users should be able to use these fine-tuned
+  checkpoints with any post-training tool (quantization, eval, inference) which supports the
+  source format.
```
ebsmothers, line 28: Personally I would emphasize this sentence more as it's a pretty big benefit.

docs/source/examples/checkpointer.rst
```diff
@@ -0,0 +1,319 @@
...
+
+
+|
+
+Checkpoint Formats
+------------------
+
+TorchTune supports three different
+`checkpointers <https://github.com/pytorch/torchtune/blob/main/torchtune/utils/_checkpointing/_checkpointer.py>`_,
+each of which supports a different checkpoint format.
```
ebsmothers, line 42: nit: support overload in this sentence

docs/source/examples/checkpointer.rst
```diff
@@ -0,0 +1,319 @@
...
+
+
+**HFCheckpointer**
+
+This checkpointer reads and writes checkpoints in a format which is compatible with the transformers
+framwork from Hugging Face. This is the most popular format within the Hugging Face Model Hub and is
```
ebsmothers, line 48:
```suggestion
framework from Hugging Face. This is the most popular format within the Hugging Face Model Hub and is
```

docs/source/examples/checkpointer.rst
```diff
@@ -0,0 +1,319 @@
...
+    # set to True if restarting training
+    resume_from_checkpoint: False
+
+|
+
+**TorchTuneCheckpointer**
+
+This checkpointer reads and writes checkpoints in a format that is compatible with TorchTune's
+model definition. This does not perform any state_dict conversions and is currently used either
+for testing or for loading quantized models for generation.
```
ebsmothers, line 179: or by half of torchtune devs 😃

docs/source/examples/checkpointer.rst
```diff
@@ -0,0 +1,319 @@
...
+---------------------------------
+
+TorchTune Checkpointers support two checkpointing scenarios:
+
+**End-of-training Checkpointing**
+
+The model weights at the end of a completed training
+run are written out to file. The checkpointer ensures that the output checkpoint
+files have the same keys as the input checkpoint file used to begin training. The
+checkpointer also ensures that the keys are paritioned across the same number of
```
ebsmothers, line 194:
```suggestion
checkpointer also ensures that the keys are partitioned across the same number of
```

docs/source/examples/checkpointer.rst
```diff
@@ -0,0 +1,319 @@
...
+        # dir for saving the output checkpoints. Usually set
+        # to be the same as checkpoint_dir
+        output_dir: <checkpoint_dir>
+
+        # model_type which specifies how to convert the state_dict
+        # into a format which TorchTune understands
+        model_type: LLAMA2
+
+    # set to True if restarting training
+    resume_from_checkpoint: True
```
ebsmothers, line 319: I feel like this ends kinda abruptly. Would either add some bit about offramps or even just a generic conclusion to wrap things up. Given that you're ending with LoRA, an example of generation with the merged checkpoint in e.g. HF format could be a nice way to tie everything together.

recipes/configs/llama2/7B_full.yaml
```diff
@@ -26,7 +26,7 @@
 # Tokenizer
 tokenizer:
   _component_: torchtune.models.llama2.llama2_tokenizer
-  path: /tmp/llama2/tokenizer.model
+  path: /tmp/Llama-2-7b-hf/tokenizer.model
```
ebsmothers, line 29: 😍

torchtune/utils/_checkpointing/_checkpointer.py
```diff
@@ -260,24 +260,27 @@ class FullModelHFCheckpointer(_CheckpointerInterface):
...
-        2. Use the directory created in 1. as the checkpoint_dir
-
     A few notes about the checkpoint reading logic:
-        - HF checkpoint names usually oredered by ID (eg: 0001_of_0003, 0002_of_0003, etc.) To ensure
-            we read the files in the right order, we sort the checkpoint file names before reading
-        - Checkpoint conversion to and from HF's format requires access to model params which are
-            read directly from the "config.json" file. This helps ensure we either load the weights
-            correctly or error out in case of discrepancy between the HF checkpoint file and TorchTune's
-            model implementations.
+    - HF checkpoint names usually oredered by ID (eg: 0001_of_0003, 0002_of_0003, etc.) To ensure
```
ebsmothers, line 264:
```suggestion
    - HF checkpoint names usually ordered by ID (eg: 0001_of_0003, 0002_of_0003, etc.) To ensure
```

torchtune/utils/_checkpointing/_checkpointer.py
```diff
@@ -600,25 +576,11 @@ def save_checkpoint(
         checkpoint file ``recipe_state.pt`` is created in ``_output_dir`` which contains the recipe
         state. The output state dicts have the following formats:
```
ebsmothers, line 577: Prob should remove this last bit too

docs/source/api_ref_utilities.rst
kartikayk: Oh sorry, decided to punt the change to a separate PR since that will be a fairly widespread change.

docs/source/examples/checkpointer.rst
kartikayk: Both good points. I had the "What you will learn" and then removed it. Will bring it back. Good point on the name

docs/source/examples/checkpointer.rst
kartikayk: Great idea! Let me try to put an example of this

docs/source/examples/checkpointer.rst
kartikayk: I was this close without a typo?

docs/source/examples/checkpointer.rst
kartikayk: Good point!

docs/source/api_ref_utilities.rst
rohan-varma, line 18: The tutorial discusses 3 checkpointers, but the torchtune one is left out here. Is the thinking that users don't really have to use the torchtune one? If so, then shall we just remove it from tutorial, as we're presenting user with info that they're not really going to use?

docs/source/examples/checkpointer.rst
```diff
@@ -0,0 +1,451 @@
...
+        # next section
+        recipe_checkpoint: null
+
+        # dir for saving the output checkpoints. Usually set
+        # to be the same as checkpoint_dir
+        output_dir: <checkpoint_dir>
+
+        # model_type which specifies how to convert the state_dict
+        # into a format which TorchTune understands
+        model_type: LLAMA2
```
rohan-varma, line 185: model_type is pretty confusing to me and not sure if its going to be super clear for users. For example, what's the supported list of `model_type`? If I want to work with mistral models, how do I find the appropriate `model_type`? Do I ever need to define my own?

docs/source/examples/checkpointer.rst
```diff
@@ -0,0 +1,451 @@
...
+
+.. code-block:: yaml
+
+    checkpointer:
+
+        # checkpointer to use
+        _component_: torchtune.utils.FullModelHFCheckpointer
+
+        # directory with the checkpoint files
+        # this should match the output_dir above
```
rohan-varma, line 331: nit: there is no `output_dir` above?

docs/source/examples/checkpointer.rst
```diff
@@ -0,0 +1,451 @@
...
+        checkpoint_dir: <checkpoint_dir>
+
+        # checkpoint files. These refer to intermediate checkpoints
```
rohan-varma, line 334: Noted that these are the intermediate checkpoint files generated, though might be clearer to explicitly mention what users need to change when loading in from a mid-training checkpoint. i.e. explicitly mention "replace the original checkpoint files with the ones saved during mid training". Might be obvious but feel like changing config stuff is pretty easy to forget.

docs/source/examples/checkpointer.rst
```diff
@@ -0,0 +1,451 @@
...
+    [_checkpointer.py:456] Model checkpoint of size 3.50 GB saved to /tmp/Llama-2-7b-hf/hf_model_0002_0.pt
+    [_checkpointer.py:467] Adapter checkpoint of size 0.01 GB saved to /tmp/Llama-2-7b-hf/adapter_0.pt
+
+
+2. Update the generation recipe with these checkpoints. The new config has the following changes to
+``generate.yaml``
+
+.. code-block:: yaml
+
+    checkpointer:
```
rohan-varma, line 430: Don't I need `adapter_checkpoint` somewhere? where does `adapter_0` get read? Or is it a default setting? If so, would be vaulable to specify that since its not clear where the actual adapter weights learnt during training are getting read here.

rohan-varma (commented):
Thanks for this! Couple of comments, but LG overall

docs/source/examples/checkpointer.rst
kartikayk: good catch!

docs/source/examples/checkpointer.rst
kartikayk: So we dont use adapter weights during generation since we always use merged checkpoints for post-training ops. Probably good to clarify this somewhere, not sure if I'll do this here though. Maybe in the post-training tutorial which I'll work on enxt.

docs/source/examples/checkpointer.rst
```diff
@@ -0,0 +1,451 @@
...
+**HFCheckpointer**
+
+This checkpointer reads and writes checkpoints in a format which is compatible with the transformers
+framework from Hugging Face. As mentioned above, this is the most popular format within the Hugging Face
+Model Hub and is the default format in every TorchTune config.
+
+For this checkpointer to work correctly, we assume that checkpoint_dir contains the necessary checkpoint
+and json files. The easiest way to make sure everything works correctly is to use the following flow:
+
+- Download the model from the HF repo using tune download. By default, this will ignore the "safetensors"
```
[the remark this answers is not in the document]
kartikayk: Yeh this is a good point, I was thinking about this too. TBH, I'll wait for user feedback on this. On first look it doesnt seem like there are a LOT of popular models with just safetensor files (I might be wrong). And the default is there to make sure people can actually download models without running out of memory (I was hitting this issue on runpod and had to hack through this). Right now I think we have decent enough logs that users can go in and modify. if we find this isnt enough or get a lot of requests, we can revist?

docs/source/examples/checkpointer.rst
```diff
@@ -0,0 +1,451 @@
...
+    Checkpoint conversion to and from HF's format requires access to model params which are
+    read directly from the "config.json" file. This helps ensure we either load the weights
+    correctly or error out in case of discrepancy between the HF checkpoint file and TorchTune's
+    model implementations. This json file is downloaded from the hub along with the model checkpoints.
+    More details on how these are used during conversion can be found
+    `here <https://github.com/pytorch/torchtune/blob/main/torchtune/models/convert_weights.py>`_.
+
+|
+
+**MetaCheckpointer**
```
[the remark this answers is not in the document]
kartikayk: Mostly self-contained info for this checkpointer?

docs/source/examples/checkpointer.rst
```diff
@@ -0,0 +1,451 @@
...
+                "key_2": weight_2,
+                ...
+            }
+
+
+**Mid-training Chekpointing**.
+
+If checkpointing in the middle of training, the output checkpoint needs to store additional
+information to ensure that subsequent training runs can be correctly restarted. In addition to
+the model checkpoint files, we output a ``recipe_state.pt`` file for intermediate
```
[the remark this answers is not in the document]
kartikayk: Its a good use case, but seems like an edge case for now? Happy to revisit this as well based on user info. The problem is that recipe_state can be quite a big cpt since we capture opt state as well. So we end up with ~13.5GB checkpoint/epoch for weights and then ~27GB checkpoint/epoch for opt state. This can be quite a lot esp when users dont have much disk space

docs/source/examples/checkpointer.rst
```diff
@@ -0,0 +1,451 @@
...
+        # checkpoint files. These refer to intermediate checkpoints
+        # and will always have a ".pt" extensions
+        checkpoint_files: [
+            hf_model_0001_0.pt,
+            hf_model_0002_0.pt,
+        ]
+
+        # if we're restarting a previous run, we need to specify
+        # the file with the checkpoint state
+        recipe_checkpoint: recipe_state.pt
```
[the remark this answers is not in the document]
kartikayk: I'm not a huge fan of changing behavior based on files tbh. Locks us into specific file names or formats. I think it's fine if the user forgets and we just have meaningful errors?

docs/source/examples/checkpointer.rst
```diff
@@ -0,0 +1,451 @@
...
+state dict will raise an exception) or worse - silent errors (loading will succeed but training or
+inference will not work as expected). In addition to the keys lining up, you also need the shapes
+of the weights (values in the state_dict) to match up exactly with those expected by the model
+definition.
+
+Let's look at the two popular formats for Llama2.
+
+**Meta Format**
+
+This is the format supported by the official Llama2 implementation. When you download the llama2 7B model
```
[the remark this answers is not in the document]
RdoubleA: Ah yes, yes it does

docs/source/examples/checkpointer.rst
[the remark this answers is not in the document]
RdoubleA: As long as it's clear how to not ignore safetensors, this makes sense

docs/source/examples/checkpointer.rst
```diff
@@ -0,0 +1,453 @@
...
+
+3. Run the generation recipe with the new config
+
+.. code-block:: bash
+
+    tune run generate --config generate
+
+    # output from the generation
+    [generate.py:68] Model is initialized with precision torch.bfloat16.
+    [generate.py:92] Welcome to the 'Alternative' Treatments and Therapies Forum
```
ebsmothers, line 450: nit: without understanding the generation recipe I have no idea what's going on here (even with understanding it I kinda have no idea what's going on here 😅)

ebsmothers (approved):
A couple nits but overall this looks good to me!

docs/source/examples/checkpointer.rst
kartikayk: oh you mean in the logs - hmm, let me take a closer look

tcapelle:
Hello!

I am wondering on best practice to enable saving checkpoints to Weights & Biases using the checkpointers. I was thinking on subclassing the Checkpointer classes but it seems that I would need to subclass every one of them.
Can you suggest me a good way to approach this? 

My current solution seems very convoluted and brittle:

```python

def save_checkpoint(self, epoch: int) -> None:
    ckpt_dict = {utils.MODEL_KEY: self._model.state_dict()}
    # if training is in-progress, checkpoint the optimizer state as well
    if epoch + 1 < self.total_epochs:
        ckpt_dict.update(
            {
                utils.SEED_KEY: self.seed,
                utils.EPOCHS_KEY: self.epochs_run,
                utils.TOTAL_EPOCHS_KEY: self.total_epochs,
                utils.MAX_STEPS_KEY: self.max_steps_per_epoch,
            }
        )
        if not self._optimizer_in_bwd:
            ckpt_dict[utils.OPT_KEY] = self._optimizer.state_dict()
        else:
            ckpt_dict[utils.OPT_KEY] = self._optim_ckpt_wrapper.state_dict()
    self._checkpointer.save_checkpoint(
        ckpt_dict,
        epoch=epoch,
        intermediate_checkpoint=(epoch + 1 < self.total_epochs),
    )
    ## Let's save the checkpoint to W&B
    ## depending on the Checkpointer Class the file will be named differently
    ## Here it is an example for the full_finetune case
    checkpoint_file = Path.joinpath(
        self._checkpointer._output_dir, f"torchtune_model_{epoch}"
    ).with_suffix(".pt")
    wandb_at = wandb.Artifact(
      name=f"torchtune_model_{epoch}",
      type="model",
      description="Model checkpoint",
      metadata={
        utils.SEED_KEY: self.seed,
        utils.EPOCHS_KEY: self.epochs_run,
        utils.TOTAL_EPOCHS_KEY: self.total_epochs,
        utils.MAX_STEPS_KEY: self.max_steps_per_epoch,
        }
    )
    wandb_at.add_file(checkpoint_file)
    wandb.log_artifact(wandb_at)
```

kartikayk:
@tcapelle thanks for taking a look!

Agreed, subclassing these is not a great idea like you pointed out. We also try to minimize implementation inheritance in this code base so that would not work out from a design principles perspective.

I'm trying to understand the use case a bit better. Do you want to add functionality to save the final checkpoint to WANDB? What would be the benefit of doing that? Do users get some visualization tools for the checkpoint? Or would the meta data be enough? 

If it's just the metadata, then we can add this to the recipe but will need to think a bit about how we expose this since not every user would have WANDB installed. If it's the checkpoint, then probably needs to happen after the conversion is done in which case this should be a utility which runs at the recipe and picks this up from the output folder. 

But yeh if you can describe the use case in more detail, that would be helpful. Maybe do this on a issue where we can have a directed discussion?

tcapelle:
Thanks for the detailed answer!

I am mostly interested on saving the output model checkpoint to W&B, mostly for our professional users that will want to keep everything integrated. This of course should be disabled by default.

I imagine users would want to save to S3, GCP and other file storage systems when integrating torchtune on their pipelines.

PR #677, opened 2024-04-10
Log memory usage during checkpoint save

rohan-varma:
#### Context
- To help debug peak memory usage during checkpoint save, adding logs at the end of `save_checkpoint`.

#### Changelog
- ...

#### Test plan
- ....


ebsmothers (approved):

kartikayk (changes requested):
This seems mostly for debugging right? Why does the user care if this works? I think we should think a bit about logs that are for debugging (mainly for us) and for actual users. Even iof the peak memory is high, I think the % of users who'll understand this is because of the _load_state_dict call is really small. I think we should think about the intent of these logs a bit and see how to help the user on this journey. Going to request changes for now and lets discuss.

PR #681, opened 2024-04-10
.github: Add build-python-only, cleanup vars

seemethere:
These vars have defaults so we don't need to explicitly name them, adds the build-python-only variable so we only build for one python version


seemethere:
Contingent on https://github.com/pytorch/test-infra/pull/5084

joecummings (approved):
lgtm

PR #680, opened 2024-04-10
Remove checkpointable dataloader

kartikayk:
#### Context

We don't have plans for adding this functionality for the launch and longer term we should reuse the work being done in torchdata/dataloading

#### Changelog
- Remove all of the relevant files and tests

#### Test plan
- All unit tests succed

PR #682, opened 2024-04-10
Remove checkpointable dataloader

kartikayk:
#### Context

We don't have plans for adding this functionality for the launch and longer term we should reuse the work being done in torchdata/dataloading.

Copy of #680  since that PR is stuck in some weird state

#### Changelog
- Remove all of the relevant files and tests

#### Test plan
- All units tests pass


ebsmothers (approved):
Second times a charm

PR #685, opened 2024-04-10
remove extra dev dependency

ebsmothers:
This was only explicitly used in some random comparison script which was removed in #537


joecummings (approved):
go go go

PR #687, opened 2024-04-10
drop 3.8 and 3.9 distributed tests from every PR

ebsmothers:
Our queue times for GPU runners are long. We doubled the number of distributed CI jobs that run on every PR yesterday. This gets us back to the original number, but keeps us testing on both nightly and stable versions of PyTorch


joecummings (approved):
I think we can go even further and test on just 3.11

PR #692, opened 2024-04-11
test checkpoints via curl

ebsmothers:
Moved our test checkpoints around so they can be fetched directly via curl with no requirement of S3 setup for recipe tests.

This PR is from my fork so if CI is green then we're a-ok


joecummings:
Can you promise to block creation of new branches for everyone but pytorch org after this?

joecummings (approved):

kartikayk (approved):
Awesome sauce!

PR #694, opened 2024-04-12
Local Alpaca for recipe tests

ebsmothers:
Our CI jobs keep failing on calls to HF APIs. We probably shouldn't be hitting the Alpaca dataset API with every CI job anyways, so this PR creates a stripped-down local sample of 100 elements to be used by our recipe tests (script for creating the sample is [here](https://gist.github.com/ebsmothers/9bb1c7b468cdaba71da45355fca8bd41)).

Similar to our toy models, we expose the dummy dataset via config, defining `dummy_alpaca_dataset_config` in our recipe test utils.

Finally, we point all recipe tests to use this local dataset instead of the remote Alpaca datasets on the hub, and update the corresponding expected loss values.

```
...
===== 18 passed, 183 deselected, 3 warnings in 246.24s (0:04:06) ======
```

CI should be green (hopefully consistently now)


tests/recipes/utils.py
```diff
@@ -38,6 +39,22 @@ def __len__(self):
...
 
 
+def get_assets_path():
+    return Path(__file__).parent.parent / "assets"
+
+
+def dummy_alpaca_dataset_config():
+    data_files = os.path.join(get_assets_path(), "alpaca_tiny.json")
+    out = [
+        "dataset._component_=torchtune.datasets.instruct_dataset",
```
RdoubleA, line 49: 😍

RdoubleA (approved):
Looks good to me, thanks for making the CI more robust!

PR #697, opened 2024-04-12
[Easy] Fix mistral LoRA single device and QLoRA config

SLR722:
#### Context
- mistral LoRA single device config and QLoRA config miss essential config key 'compile' and break training

#### Changelog
- add compile key to mistral LoRA single device config and QLoRA config

#### Test plan
- tune run lora_finetune_single_device --config mistral/7B_lora_single_device
- tune run lora_finetune_single_device --config mistral/7B_qlora_single_device


RdoubleA (approved):
very nice

ebsmothers (approved):
Thanks for fixing this! I think I missed this on the merge with Rohan's changes (another reason we need Mistral recipe tests asap)

PR #701, opened 2024-04-12
Remove low_precision directory and move helpers to utils

rohan-varma:
#### Context
- `modules.low_precision` originally housed `nf4linear`, but since we removed this, there are mostly utilities left here. I moved our state_dict hook to `torchtune.models.common_utils`, and moved `_register_nf4_dispatch_ops` to utils (this file is still not exposed).
- Cannot move the state_dict hook to utils, as we need to use it in `torchtune.models` and housing it in utils causes circular import issues. So made a `common_utils` folder in `torchtune.models` to house utilities common to all models, such as the state_dict hook.
- Also rendering the `state_dict post` hook API in docs as it is public.

#### Changelog
- ...

#### Test plan
- 
image


ebsmothers (approved):
Overall this looks good to me. One q: what do you think about putting common_utils under modules instead of models? The way I see it, we could in theory apply the state dict hook to any module, and putting it in models could cause the same circular import issue if we ever try to use in modules/

rohan-varma:
@ebsmothers That makes sense to me. We can definitely use this in any module and there's nothing preventing user from doing so. However using it at the very top level model is the most helpful, as if we use them in submodules, user will have to take care to apply them to all submodules in their model that contain nf4s. But will just go ahead with that suggestion for now and we can revisit

PR #702, opened 2024-04-12
Update lora tutorial

ebsmothers:
General cleanup, replacing any stale commands, and adding a section at the end on trading off memory and eval performance.


rohan-varma (approved):
Might have to update run command too.

ebsmothers:
> Might have to update run command too.

Oh yeah sorry maybe should've put *very* wip. I just pushed my local changes as-is, gonna do quite a bit more here.

docs/source/examples/lora_finetune.rst
```diff
@@ -204,7 +206,7 @@ model without any wrappers or custom checkpoint conversion logic.
 
 .. note::
     Whenever loading weights with :code:`strict=False`, you should verify that any missing or extra keys in
-    the loaded :code:`state_dict` are as expected. torchtune's LoRA recipe does this by default via
+    the loaded :code:`state_dict` are as expected. torchtune's LoRA recipes do this by default via e.g.
```
kartikayk, line 209: Is "e.g." needed here?

docs/source/examples/lora_finetune.rst
```diff
@@ -288,8 +290,8 @@ Let's run this experiment. We can also increase alpha (in general it is good pra
 
 .. code-block:: bash
 
-    tune run --nnodes 1 --nproc_per_node 2 lora_finetune_distributed --config lora_finetune_distributed \
-    lora_attn_modules='[q_proj, k_proj, v_proj, output_proj]' \
+    tune run --nnodes 1 --nproc_per_node 2 lora_finetune_distributed --config llama2/7B_lora \
+    lora_attn_modules=['q_proj','k_proj','v_proj','output_proj'] \
```
kartikayk, line 294: Oh interesting, does this work for you? Whenever I do this override for checkpointer.checkpoint_files I need to set the list as '[]' i.e. a string for this to be correctly parsed

docs/source/examples/lora_finetune.rst
```diff
@@ -300,7 +302,80 @@ A comparison of the (smoothed) loss curves between this run and our baseline ove
...
-As an exercise, you can also try running some evaluation tasks or manually inspecting generations
-output by your saved checkpoints (which can be found in :code:`output_dir`).
-You may want to train the model for longer first, as here we only looked at 500 steps
-(which corresponds to about 2% of one epoch of the Alpaca dataset).
+
+Trading off memory and model performance with LoRA
+--------------------------------------------------
+
+In the preceding example, we ran LoRA on two devices. But with LoRA's memory efficiency it is possible
+to run on a single device as well. This can be done via the command:
```
kartikayk, line 310:
```suggestion
In the preceding example, we ran LoRA on two devices. But given LoRA's low memory footprint, we can run fine-tuning on a single device using most commodity GPUs which support bfloat16 as a dtype. This can be done via the command:
```

docs/source/examples/lora_finetune.rst
```diff
@@ -300,7 +302,80 @@ A comparison of the (smoothed) loss curves between this run and our baseline ove
...
+
+.. code-block:: bash
+
+    tune run lora_finetune_single_device --config llama2/7B_lora_single_device
+
+On a single device, we may need to be more cognizant of our peak memory. Let's run a couple experiments
```
kartikayk, line 316:
```suggestion
On a single device, we may need to be more cognizant of our peak memory. Let's run a few experiments
```

docs/source/examples/lora_finetune.rst
```diff
@@ -300,7 +302,80 @@ A comparison of the (smoothed) loss curves between this run and our baseline ove
...
+to see our peak memory during a finetune on the Alpaca dataset. We will experiment along two axes:
+first, which model layers have LoRA applied, and second, the rank of each LoRA layer. (We will scale
+alpha in parallel to LoRA rank, as discussed above.)
+
+To compare the results of our experiments, we can evaluate our models on :code:`truthfulqa_mc2`, a task from
+the `TruthfulQA <https://arxiv.org/abs/2109.07958>`_ benchmark for language models. For more details on how to run this and other evaluation tasks
+with torchtune's EleutherAI evaluation harness integration, see our :ref:`End-to-End Workflow Tutorial <eval_harness_label>`.
+
+Previously, we only enabled LoRA for the linear layers in each self-attention, but in fact there are other linear
```
kartikayk, line 325:
```suggestion
Previously, we only enabled LoRA for the linear layers in each self-attention module, but in fact there are other linear
```

docs/source/examples/lora_finetune.rst
```diff
@@ -300,7 +302,80 @@ A comparison of the (smoothed) loss curves between this run and our baseline ove
...
+layers we can apply LoRA to: MLP layers and our model's final output projection. Note that for Llama-2-7B the final output
+projection maps to dimension 32000, so enabling LoRA for this layer will increase our peak memory a bit more
```
kartikayk, line 327:
```suggestion
projection maps to dimension 32000 instead of 4096, so enabling LoRA for this layer will increase our peak memory a bit more
```

docs/source/examples/lora_finetune.rst
```diff
@@ -300,7 +302,80 @@ A comparison of the (smoothed) loss curves between this run and our baseline ove
...
+  # Model Arguments
+  model:
+    _component_: lora_llama2_7b
+    lora_attn_modules: ['q_proj', 'k_proj', 'v_proj', 'output_proj']
+    apply_lora_to_mlp: True
+    apply_lora_to_output: True
+  ...
+
+.. note::
+    All the below finetunes use the `llama2/7B_lora_single_device <https://github.com/pytorch/torchtune/blob/ada52240514bb9fa07f91a50ad0a31063f13834c/recipes/configs/llama2/7B_lora_single_device.yaml>`_
```
kartikayk, line 341:
```suggestion
    All the finetuning runs below use the `llama2/7B_lora_single_device <https://github.com/pytorch/torchtune/blob/ada52240514bb9fa07f91a50ad0a31063f13834c/recipes/configs/llama2/7B_lora_single_device.yaml>`_
```

kartikayk: nit: can we just use the hyperlink with main instead of hardcoding to a particular commit. So in this case it would be: https://github.com/pytorch/torchtune/blob/main/recipes/configs/llama2/7B_lora_single_device.yaml

docs/source/examples/lora_finetune.rst
```diff
@@ -300,7 +302,80 @@ A comparison of the (smoothed) loss curves between this run and our baseline ove
...
+    config, which has a default batch size of 2. Modifying the batch size (or other hyperparameters, e.g. the optimizer) will impact both peak memory
+    and final evaluation results.
+
+.. list-table::
```
kartikayk, line 345: Love this table!

kartikayk (approved):
Looks great!

docs/source/examples/lora_finetune.rst
ebsmothers: I thought using a specific commit was safer tbh. I think there are pros and cons to each though, no strong preference as long as we are consistent

kartikayk: Since we usually say something like "look here for details", my concern is that the details will be outdated. The downside of linking to main is that the file may not exist at the same location and so the link is broken, but maybe thats better than stale info?

docs/source/examples/lora_finetune.rst
ebsmothers: Technically yes because we now have different validation checks in single device and distributed LoRA recipes cause of that FSDP state dict business, and I definitely do not wanna get into that here. So I could leave it out, but this "e.g." is actually doing a bit of heavy lifting tbh

docs/source/examples/lora_finetune.rst
ebsmothers: Yeah the secret is the lack of spaces

docs/source/examples/lora_finetune.rst
ebsmothers: Yeah fair point. I'm good to update it to main

PR #704, opened 2024-04-13
[QoL] Keep launch commands in config to one line

RdoubleA:
The launch commands in the configs are broken up into multiple lines - unfortunately since they are comments, I can't copy-paste them and run on command-line without having to manually delete all the comment #'s. So this PR puts them all in one line so it's easy to copy. I don't think the slightly longer line is offensive enough to offset the value of quickly copy-pasting the command.

## Test plan
eyes


rohan-varma (approved):
Makes sense!

PR #709, opened 2024-04-13
Speed up model loading for generate

albanD:
**This has not been extensively tested  (only mistral 7b) and more of a proposal!**

This change does the follow:
- Create the model on the meta device
- Load the state dict with assign=True which preserve the properties of the checkpoint (mmap-ed cpu Tensor in this case)
- Initialize non-persistent buffers remaining on the meta device
- Move the finalized model to the requested device/dtype

This makes the model loading almost instant on my machine.

### DIFF
--- a/recipes/generate.py
+++ b/recipes/generate.py
@@ -54,14 +54,22 @@ class InferenceRecipe:
         model_cfg: DictConfig,
         model_state_dict: Dict[str, Any],
     ) -> nn.Module:
-        with utils.set_default_dtype(self._dtype), self._device:
+        with torch.device("meta"):
             model = config.instantiate(model_cfg)
+
+        model.load_state_dict(model_state_dict, assign=True)
+
+        # Initialize non-persistent buffers that are still on meta
+        # device at this point
+        def init_non_persistent_buffers(m):
+            if hasattr(m, "reset_non_persistent_buffers"):
+                m.reset_non_persistent_buffers()
+        model.apply(init_non_persistent_buffers)
 
         if self._quantization_mode is not None:
             model = self._quantizer.quantize(model)
-            model = model.to(device=self._device, dtype=self._dtype)
 
-        model.load_state_dict(model_state_dict)
+        model = model.to(device=self._device, dtype=self._dtype)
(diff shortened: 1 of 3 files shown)


albanD:
Any way I can know from the CI logs what is my lint mistake so I can fix it?

kartikayk:
@albanD thanks so much for putting this up! I'll take a more detailed look tomorrow, but to answer your lint question - you can do the following:

```
pre-commit install
pre-commit run --all-files
```

This will fix all of the issues for you.

ebsmothers:
Thanks for the PR @albanD! Tbh we have already had a fraught relationship with meta device initialization 😅 (see e.g. #317, #418, #514). Our latest status is that we deliberately sacrifice a bit on time-to-first-batch for the sake of keeping code in the model components agnostic to meta device. But generation is an interesting case since the total runtime is much lower than on a finetune with FSDP (which is what we were focusing on previously). Out of curiosity, what is the speedup of meta device vs just initializing directly on GPU in this case?

albanD:
I would need to check once I go back on the machine in question. 
The more important bit tbh is that the CPU model was fully using the mmap-ed loaded Tensors and so was not blowing up my scarse RAM :D 

@kartikayk I saw that but I don't have pre-commit in my environment :p

recipes/generate.py
```diff
@@ -54,14 +54,22 @@ def _setup_model(
         model_cfg: DictConfig,
         model_state_dict: Dict[str, Any],
     ) -> nn.Module:
-        with utils.set_default_dtype(self._dtype), self._device:
+        with torch.device("meta"):
             model = config.instantiate(model_cfg)
 
+        model.load_state_dict(model_state_dict, assign=True)
```
jerryzh168, line 60: for quantized models we'd need to load after we do quantization I think

albanD: We should consider doing this https://docs-preview.pytorch.org/pytorch/tutorials/2824/recipes/recipes/swap_tensors.html?highlight=swap_tensor once we can use 2.3+ in AO/here.

PR #711, opened 2024-04-14
Update some config default values

ebsmothers:
Our defaults are not tuned super well in some of our configs. E.g. for our Llama-2-7B LoRA single device the model doesn't learn with the default LR cause our bsz is only 2. This PR makes some changes to our effective batch sizes so that out-of-the-box learning should be a bit better. 

Note: I have not run all of these myself yet, just the Llama-2-7B LoRA ones. For QLoRA I chatted with @rohan-varma, for Mistral I am going off of @kartikayk's tuning for the distributed recipes.

Also for Llama-2-7B LoRA I modified the default distributed config to be set up for 2 devices, can consider doing the same for other distributed recipe configs too since I think this'll be the most common starting point.

**Note**: I also need to update some recipe tests as a result of this, planning to do that shortly.

### DIFF
--- a/recipes/configs/llama2/13B_lora.yaml
+++ b/recipes/configs/llama2/13B_lora.yaml
@@ -70,6 +70,7 @@ loss:
 # Training
 epochs: 1
 max_steps_per_epoch: null
+gradient_accumulation_steps: 1
 
 # Logging
 output_dir: /tmp/lora_finetune_output
--- a/recipes/configs/llama2/7B_lora.yaml
+++ b/recipes/configs/llama2/7B_lora.yaml
@@ -5,13 +5,13 @@
 # this run:
 #   tune download meta-llama/Llama-2-7b --hf-token <HF_TOKEN> --output-dir /tmp/llama2
 #
-# To launch on 4 devices, run the following command from root:
-#   tune run --nnodes 1 --nproc_per_node 4 lora_finetune_distributed --config llama2/7B_lora
+# To launch on 2 devices, run the following command from root:
+#   tune run --nnodes 1 --nproc_per_node 2 lora_finetune_distributed --config llama2/7B_lora
 #
 # You can add specific overrides through the command line. For example
 # to override the checkpointer directory while launching training
 # you can run:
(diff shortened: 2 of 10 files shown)


RdoubleA:
How are you determining the number of gradient accumulation steps? And won't this impact all of our expected loss curves and evals when running these configs?

ebsmothers:
@RdoubleA 

> How are you determining the number of gradient accumulation steps?

**Llama-2-7B, LoRA single device**: goal is effective batch size of 128 as that's what I've seen good results with (this is the diff between the crappy loss curve I shared a few days ago over chat and the one that looks as we expect). So since our batch size is 2, I set grad accumulation steps to 64.
**Llama-2-7B, LoRA distributed**: same as single device, but assume two devices (and document as such in the config). So in this case, set grad accumulation steps to 32.
**Llama-2-7B, QLoRA single device**: talked to Rohan 😃 (he's been running more of these). I also have some local experiments of my own, so can check on those too.
**Mistral 7B, FFT single device**: Set based on discussions with Kartikay as he had tuned the distributed recipe on 4 devices with batch size 2. So for batch size 2 in the single-device recipe, I set grad accumulation steps to 4 to match the effective batch size of the distributed recipe.
**Mistral 7B, LoRA single device**: Same as the Mistral 7B FFT single device (but local batch size is 4 instead).
**Mistral 7B, QLoRA single device**: Basically just set to match LoRA. This one is probably the one with the fewest datapoints tbh.

> And won't this impact all of our expected loss curves and evals when running these configs?

Yes, but the idea is that the results will be strictly better after these changes. Currently our learning rate is tuned for larger effective batch sizes than in our default configs (at least for LoRA).

RdoubleA:
Thanks for those details, that all makes sense to me. If you don't mind updating all the distributed configs for 2 devices, I can go ahead and approve (and you're able to make CI green)

RdoubleA (approved):

rohan-varma (approved):
Makes sense. If any of these have implication toward memory, please update the table. The table right now assumes you can get that memory out of the box via defaults.

ebsmothers:
> If any of these have implication toward memory, please update the table.

Batch size is not actually changing anywhere so peak memory shouldn't be impacted

Issue #597, opened 2024-03-26
BF16 and MemoryEffient attention not working on AMD MI250


chauhang:
On running torchtume on AMD MI250, the fine-tune completes with ~1.58s/it for LoRA distributed recipe. However, getting following warnings for BF16 and MemoryEfficent attention support:

```
BF16 not supported on this hardware. Setting dtype to float32
...

/torchtune/torchtune/modules/attention.py:208: UserWarning: 1Torch was not compiled with memory efficient attention. (Triggered internally at ../aten/src/ATen/native/transformers/hip/sdp_utils.cpp:515.)
  output = nn.functional.scaled_dot_product_attention(
```

Full trace [here](https://gist.github.com/chauhang/7d633b427f5b3a9c35bcd077a4b412ff)

### Steps
1. Git clone and install torch tune
2. Download the model
3. Run lora recipe using: 
`tune --nnodes 1 --nproc_per_node 8 lora_finetune_distributed --config lora_finetune_distributed`

### Environment

```
python -m torch.utils.collect_env

Collecting environment information...
PyTorch version: 2.4.0.dev20240323+rocm6.0
Is debug build: False
CUDA used to build PyTorch: N/A
ROCM used to build PyTorch: 6.0.32830-d62f6a171

OS: CentOS Stream 9 (x86_64)
GCC version: (GCC) 11.4.1 20231218 (Red Hat 11.4.1-3)
Clang version: Could not collect
CMake version: version 3.26.5
Libc version: glibc-2.34

Python version: 3.11.5 (main, Sep 11 2023, 13:54:46) [GCC 11.2.0] (64-bit runtime)
Python platform: Linux-6.4.3-0_fbk1_zion_755_ga25447393a1d-x86_64-with-glibc2.34
Is CUDA available: True
CUDA runtime version: Could not collect
CUDA_MODULE_LOADING set to: LAZY
GPU models and configuration: AMD Instinct MI250X / MI250 (gfx90a:sramecc+:xnack-)
Nvidia driver version: Could not collect
cuDNN version: Could not collect
HIP runtime version: 6.0.32830
MIOpen runtime version: 3.0.0
Is XNNPACK available: True

CPU:
Architecture:                    x86_64
CPU op-mode(s):                  32-bit, 64-bit
Address sizes:                   48 bits physical, 48 bits virtual
Byte Order:                      Little Endian
CPU(s):                          256
On-line CPU(s) list:             0-255
Vendor ID:                       AuthenticAMD
[...]

PR #715, opened 2024-04-15
Raise when bf16 is requested but not supported.

rohan-varma:
#### Context
- It's a bit confusing to continue training in fp32 when bf16 is requested and the HW doesn't support it as the user will experience significantly different memory and performance characteristics as they would see with bf16. If user requests bf16 it's likely that they want to train w/bf16 and not fall back to fp32, and upon seeing this error user can readjust config to fp32 dtype if needed. 
- @kartikayk also mentioned a hard error makes sense in https://github.com/pytorch/torchtune/issues/597

#### Changelog
- Raise runtime error in get_dtype when asking for bf16 and bf16 is not avail.

#### Test plan
- pytest tests/torchtune/utils/test_precision.py -v -k test_error_bf16_unsupported


joecummings (approved):
This makes sense to me - feel free to merge once all signals are green.

PR #722, opened 2024-04-15
Update QLoRA card

joecummings:


rohan-varma (approved):
Thanks for catching it

PR #724, opened 2024-04-15
Add CLI to API ref

joecummings:


joecummings:
We will want to document this in a different manner.

PR #726, opened 2024-04-15
Updates to first finetune tutorial

joecummings:


docs/source/tutorials/first_finetune_tutorial.rst
```diff
@@ -38,118 +44,120 @@ Then, it's as simple as:
...
+
+For the purposes of this tutorial, you'll will be using the recipe for finetuning a Llama2 model using `LoRA <https://arxiv.org/abs/2106.09685>`_ on
+a single device. For a more in-depth discussion on LoRA in torchtune, you can see the complete :ref:`lora_finetune_label` tutorial.
+
+.. note::
+
+  **Why have a separate recipe for single device vs. distributed?** This is discussed in
+  :ref:`recipe_deepdive` but one of our core principles in torchtune is minimal abstraction and boilerplate code.
+  If you don't want to train in a distributed environment, you shouldn't need to wade through hundreds of lines of complicated
+  distributed-specific code in order to understand how your model is training.
```
ebsmothers, line 97: Imo this is worded too strongly and should be framed positively rather than negatively, e.g. "If you only want to train on a single GPU, our single-device recipe ensures you don't have to worry about additional features like FSDP that are only required for distributed training."

docs/source/tutorials/first_finetune_tutorial.rst
```diff
@@ -38,118 +44,120 @@ Then, it's as simple as:
...
-  You can also download the model directly through the Llama2 repository.
-  See `this page <https://llama.meta.com/get-started#getting-the-models>`_ for more details.
-
+|
 
 Selecting a recipe
 ------------------
 Recipes are the primary entry points for torchtune users.
-These can be thought of as end-to-end pipelines for training and optionally evaluating LLMs.
+These can be thought of as **singularly-focused scripts for interacting with LLMs**, including training
```
ebsmothers, line 58:
```suggestion
These can be thought of as **singularly-focused scripts for interacting with LLMs**, including training,
```

ebsmothers: Do we want to put "hackable" or "extensible" or something like that as well? Basically just emphasize that they are intended to be experimented with

docs/source/tutorials/first_finetune_tutorial.rst
```diff
@@ -38,118 +44,120 @@ Then, it's as simple as:
...
+inference, evaluation, and quantization.
 
 Each recipe consists of three components:
 
-* **Configurable parameters**, specified through yaml configs, command-line overrides and dataclasses
+* **Configurable parameters**, specified through yaml configs and command-line overrides
+* **Recipe script**, entry-point which puts everything together including parsing and validating configs, setting up the environment, and correctly using the recipe class
 * **Recipe class**, core logic needed for training, exposed to users through a set of APIs
```
ebsmothers, line 65: remove "users" while you're in here

ebsmothers, line 65: Sorry I'm a bit confused about the delineation between recipe script and recipe class here

docs/source/tutorials/first_finetune_tutorial.rst
```diff
@@ -38,118 +44,120 @@ Then, it's as simple as:
...
+  If you don't want to train in a distributed environment, you shouldn't need to wade through hundreds of lines of complicated
+  distributed-specific code in order to understand how your model is training.
+
+|
 
 Modifying a config
 ------------------
 YAML configs hold most of the important information needed for running your recipe.
 You can set hyperparameters, specify metric loggers like `WandB <wandb.ai>`_, select a new dataset, and more.
 For a list of all currently supported datasets, see :ref:`datasets`.
```
ebsmothers, line 105: Personally I feel like this sentence is a bit of a non sequitur. We are talking about a bunch of different configurable fields, but then for some reason specifically call out a reference to datasets only.

docs/source/tutorials/first_finetune_tutorial.rst
```diff
@@ -24,12 +24,18 @@ job using torchtune.
 
 Downloading a model
 -------------------
-First, you need to download a model. torchtune's supports an integration
+The first step in any finetuning job is to download a pretrained base model. torchtune supports an integration
 with the `Hugging Face Hub <https://huggingface.co/docs/hub/en/index>`_ - a collection of the latest and greatest model weights.
 
-For this tutorial, you're going to use the `Llama2 model from Meta <https://llama.meta.com/>`_. Llama2 is a "gated model",
+For this tutorial, you're going to use the **`Llama2 7B model** from Meta <https://llama.meta.com/>`_. Llama2 is a "gated model",
```
ebsmothers, line 30: This link is not rendering correctly for me

docs/source/tutorials/first_finetune_tutorial.rst
joecummings: This is pulled from recipe deep-dive, so if we change here we need to change there.

docs/source/tutorials/first_finetune_tutorial.rst
joecummings: Which sentence??

docs/source/tutorials/first_finetune_tutorial.rst
```diff
@@ -1,7 +1,7 @@
 .. _finetune_llama_label:
 
 =======================
-Finetune your First LLM
+Finetune your first LLM
```
kartikayk, line 4: Need to make this change in the overview file as well to be consistent.

ebsmothers (approved):

PR #746, opened 2024-04-15
capitalize titles

ebsmothers:
#### Context
- ...

#### Changelog
- ...

#### Test plan
- ....


rohan-varma:
Doc only change. So free up multi GPU queue and skip those

PR #752, opened 2024-04-15
Generalize dataset builders as families

RdoubleA:
## Context
By moving `source` to a parameter in the builder, we can genrealize dataset builders as "families" (i.e., alpaca family datasets, etc). Also update docstrings.

## Test plan
CI

image

### DIFF
--- a/torchtune/data/_instruct_templates.py
+++ b/torchtune/data/_instruct_templates.py
@@ -39,7 +39,7 @@ class InstructTemplate(ABC):
 
 class AlpacaInstructTemplate(InstructTemplate):
     """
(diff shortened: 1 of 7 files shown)


kartikayk (approved):

torchtune/data/_instruct_templates.py
```diff
@@ -149,7 +149,7 @@ def format(
 
 class StackExchangedPairedTemplate(InstructTemplate):
     """
-    Prompt template for the StackExchangedPaired dataset.
+    Prompt template for preference datasets similar to StackExchanged paired.
```
ebsmothers, line 152: This doesn't look right

PR #759, opened 2024-04-15
Update tune download command line for llama 7B model

SLR722:
#### Context
- Fix tune download command line 

#### Changelog
- Update tune download command line for llama 7B model

#### Test plan
- run tune download meta-llama/Llama-2-7b-hf --output-dir /tmp/Llama-2-7b-hf --hf-token <TOKEN> and kick off training

### DIFF
--- a/recipes/configs/llama2/7B_full.yaml
+++ b/recipes/configs/llama2/7B_full.yaml
@@ -3,7 +3,7 @@
 #
 # This config assumes that you've run the following command before launching
 # this run:
-#   tune download meta-llama/Llama-2-7b --hf-token <HF_TOKEN> --output-dir /tmp/llama2
+#   tune download meta-llama/Llama-2-7b-hf --output-dir /tmp/Llama-2-7b-hf --hf-token <TOKEN>
 #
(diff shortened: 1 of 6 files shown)


RdoubleA:
Thanks for the update - was just about to do this as well :) only comment I have is, all our other configs have --hf-token before output-dir, it would be good to be consistent with that

recipes/configs/llama2/7B_full.yaml
```diff
@@ -3,7 +3,7 @@
 #
 # This config assumes that you've run the following command before launching
 # this run:
-#   tune download meta-llama/Llama-2-7b --hf-token <HF_TOKEN> --output-dir /tmp/llama2
+#   tune download meta-llama/Llama-2-7b-hf --output-dir /tmp/Llama-2-7b-hf --hf-token <TOKEN>
```
ebsmothers, line 6: Unless there's a good reason to change it can we keep <HF_TOKEN> instead of <TOKEN>? That's how we have it everywhere else

ebsmothers (approved):
One nit, otherwise looks good

PR #763, opened 2024-04-15
Emphasize that torchtune can use PyTorch 2.2 and above

joecummings:


kartikayk (approved):
You're a hero!

README.md
```diff
@@ -71,7 +71,7 @@ Single-GPU recipes expose a number of memory optimizations that aren't available
 
 ## Installation
 
-**Step 1:** Install Pytorch. torchtune is currently tested with the latest stable PyTorch release (2.2.2). We recommend following the instructions [here](https://pytorch.org/get-started/locally/) and installing either version 2.2.2 or the preview nightly version.
+**Step 1:** Install Pytorch. torchtune is tested on PyTorch 2.2 and above. We recommend following the instructions [here](https://pytorch.org/get-started/locally/) and installing either version 2.2.2 or the preview nightly version.
```
rohan-varma, line 74: So the current stable is 2.2.2 so we should only make the claim that we test on 2.2.2 and nightlies. Technically we don't test on 2.2 for example. 

I actually like the original wording, would phrase it like - 

> torchtune is currently tested with the latest stable PyTorch release (2.2.2) as well as the preview nightly version.

joecummings: Updated

PR #771, opened 2024-04-16
DPO supports multi-device training

yechenzhi:
As previously discussed, we will support the parallel training of DPO. However, it seems that the content of the parallel training config file is almost the same as that of the single device, with only the difference of enabling activation checkpointing. Is it necessary to have two separate files?

### DIFF
--- a/recipes/lora_dpo_distributed.py
+++ b/recipes/lora_dpo_distributed.py
@@ -1,0 +1,690 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+import sys
+import time
+
+from functools import partial
+from typing import Any, Dict, Optional, Tuple
+from warnings import warn
+
+import torch
+from omegaconf import DictConfig
+
+from torch import nn
+from torch.distributed import destroy_process_group, init_process_group
+from torch.distributed.fsdp import (
+    FullOptimStateDictConfig,
+    FullStateDictConfig,
+    FullyShardedDataParallel as FSDP,
+    StateDictType,
+)
+from torch.optim import Optimizer
+from torch.utils.data import DataLoader, DistributedSampler
+from torchtune import config, modules, utils
+from torchtune.data import CROSS_ENTROPY_IGNORE_IDX
+from torchtune.modules.peft.peft_utils import (
+    disable_adapter,
+    get_adapter_params,
+    get_merged_lora_ckpt,
+    set_trainable_params,
+    validate_state_dict_for_lora,
+)
+from torchtune.recipe_interfaces import FTRecipeInterface
+
+from tqdm import tqdm
+
+log = utils.get_logger("DEBUG")
+
+
+class LoRADPORecipeDistributed(FTRecipeInterface):
+    """
+    Distributed LoRA finetuning recipe for dense transformer-based LLMs such as Llama2. This recipe supports
+    distributed training and can be run on a single node (1 to 8 GPUs).
+
+    Features:
+        - FSDP. Supported using PyTorch's FSDP APIs. DDP is currently not supported. Traning on CPU is not
+            supported.
+
+        - Activation Checkpointing. This can be controlled using the ``activation_checkpointing``
+            flag. Activation checkpointing helps reduce the memory footprint since we no longer keep
+            activations in memory and instead recompute them during the backward pass. This is especially
+            helpful for larger batch sizes when you're memory constrained. But these savings in memory
+            come at the cost of training performance. In most cases training can slow-down quite a bit as
+            a result of this activation recomputation.
+
+        - Precision. Full fp32 and bf16 training are supported. Precision is controlled using the ``dtype``
+            flag. When ``dtype=bf16``, all activations, gradients and optimizer states are in bfloat16. In
+            most cases this should halve the memory footprint of full precision (fp32) training, without
+            loss in model quality (will depend on the model, training data and other settings). For
+            GPUs which do not support bfloat16, we fall back to fp32. Mixed precision training and fp16
+            precision are currently not supported.
+
+        - Gradient Accumulation. You can simulate larger batch sizes by accumulating gradients. This is
+            controlled using the ``gradient_accumulation_steps`` flag.
+
+                Total Batch Size = batch_size * number of GPUs * gradient accumulation steps.
+
+            For example: with batch_size=1, nproc_per_node=2 and gradient_accumulation_steps=32 we get a
+            total batch size of 64.
+
+            Gradient accumulation is especially useful when you are memory constrained. In this case,
+            accumulating gradients might give you better training speed than enabling activation
+            checkpointing.
+
+        - Checkpointing. Model weights are checkpointed both at the end of each epoch and at the end of
+            training. Currently we checkpoint both the adapter weights (trainable params only) and the
+            complete merged weights (adapter weights added back to the base model). For more details
+            please take a look at our LoRA tutorial
+            (https://pytorch.org/torchtune/main/tutorials/lora_finetune.html).
+
+            Optimizer State and recipe state (seed, total_epochs, number of epochs run etc) are
+            only saved at the end of a given epoch and used in case of resuming training. Resuming
+            training is controlled by the ``resume_from_checkpoint`` flag. Mid-epoch checkpointing is
+            currently not supported.
+
+            For more details on the checkpointer, please take a look at
+            our checkpointer deepdive (https://pytorch.org/torchtune/main/tutorials/checkpointer.html).
+
+        - Logging. Terminal, Disk, WandB and TensorBoard are all supported.
+
+    For a full list of example configs for this recipe, run ``tune ls`` on the command line. Each config
+    has example commands for how to kick-off training.
+
+    Args:
+        cfg (DictConfig): OmegaConf object parsed from yaml file
+
+    Raises:
+        ValueError: If ``dtype`` is set to fp16.
+        ValueError: If world_size is 1
+        RuntimeError: If ``dtype`` is set to bf16 and the hardware does not support bf16.
+    """
+
+    def __init__(self, cfg: DictConfig) -> None:
+        self._device = utils.get_device(device=cfg.device)
+        self._dtype = utils.get_dtype(cfg.dtype, device=self._device)
+
+        if self._dtype == torch.float16:
+            raise ValueError(
+                "full fp16 training is not supported with this recipe. Please use bf16 or fp32 instead."
+            )
+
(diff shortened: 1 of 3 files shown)


recipes/lora_dpo_distributed.py
```diff
@@ -0,0 +1,690 @@
...
+
+from tqdm import tqdm
+
+log = utils.get_logger("DEBUG")
+
+
+class LoRADPORecipeDistributed(FTRecipeInterface):
+    """
+    Distributed LoRA finetuning recipe for dense transformer-based LLMs such as Llama2. This recipe supports
+    distributed training and can be run on a single node (1 to 8 GPUs).
```
ebsmothers, line 46: Need to update this

ebsmothers: Can you also add a citation similar to the one [here](https://github.com/pytorch/torchtune/blob/e6fc832097aa08f499d11f1d97e37a06c0263db4/recipes/lora_dpo_single_device.py#L36-L38)?

recipes/lora_dpo_distributed.py
```diff
@@ -0,0 +1,690 @@
...
+                    }
+                )
+
+            self._checkpointer.save_checkpoint(
+                checkpoint_dict,
+                epoch=epoch,
+                intermediate_checkpoint=intermediate_checkpoint,
+            )
+
+    def concatenated_forward(
```
ebsmothers, line 507: Please add the same docstring you had in the single-device recipe

recipes/lora_dpo_distributed.py
```diff
@@ -0,0 +1,690 @@
...
+        chosen_log_probs = all_log_probs[:len_chosen]
+        rejected_log_probs = all_log_probs[len_chosen:]
+
+        chosen_logits = all_logits[:len_chosen]
+        rejected_logits = all_logits[len_chosen:]
+
+        return (chosen_log_probs, rejected_log_probs, chosen_logits, rejected_logits)
+
+    @staticmethod
+    def get_batch_log_probs(
```
ebsmothers, line 530: Similar comment on docstring here

ebsmothers (commented):
So overall no major concerns on the code itself as I feel it's pretty similar to the single-device recipe. 

Main question is around testing: are you able to kick off a proper distributed run and see decreasing losses here? Specifically one concern I have is around the interaction with FSDP.. since we are doing multiple forward passes of the model (one with no_grad) and modifying the internals in nontrivial ways in between, there are potentially some gotchas here. So let's definitely make sure it trains properly. Happy to help out on this front if needed, just let me know.

yechenzhi:
> Main question is around testing: are you able to kick off a proper distributed run and see decreasing losses here? Specifically one concern I have is around the interaction with FSDP.. since we are doing multiple forward passes of the model (one with no_grad) and modifying the internals in nontrivial ways in between, there are potentially some gotchas here. So let's definitely make sure it trains properly. Happy to help out on this front if needed, just let me know.

Affirmative, I conducted distributed training using two Nvidia RTX 4090 GPUs. The loss appeared normal; however, I encountered an issue where utilizing two GPUs did not expedite the training process. I also attempted the 'tune run --nnodes 1 --nproc_per_node 2 lora_finetune_distributed --config llama2/7B_lora' command, leading me to suspect that the bottleneck may be due to the RTX 4090 GPUs.
loss:
image
similar to [here](https://github.com/pytorch/torchtune/pull/645#issuecomment-2047853377)

ebsmothers:
> The loss appeared normal; however, I encountered an issue where utilizing two GPUs did not expedite the training process. I also attempted the 'tune run --nnodes 1 --nproc_per_node 2 lora_finetune_distributed --config llama2/7B_lora' command, leading me to suspect that the bottleneck may be due to the RTX 4090 GPUs.

@yechenzhi thanks for the info. The loss curve looks good so no major concerns from my side. I am curious about this comment though. In general we expect slow per-iteration speed due to extra communication overhead of using FSDP. However, this should be balanced by the ability to fit larger batch sizes than you otherwise would. Out of curiosity, are you able to increase the per-device batch size beyond 4 on your RTX 4090s? And if so, does this improve the training time?

I will also patch your changes and run some experiments on my end.

yechenzhi:
> Out of curiosity, are you able to increase the per-device batch size beyond 4 on your RTX 4090s? 

Yes, I set per-device batch size to 6 now.

> And if so, does this improve the training time?

When compared to single-device training, there was no improvement in training speed. I suspect this is because the 4090 model lacks NVLink technology (though I'm not certain, as I'm not familiar with FSDP or NVLink).

> I will also patch your changes and run some experiments on my end.

Okay, you can take a look at [this](https://github.com/pytorch/torchtune/pull/645#issuecomment-2047915895).

ebsmothers:
@yechenzhi my loss (admittedly over only a few hundred iterations) using the commands from the new config file looks more like this: 
Screenshot 2024-04-21 at 5 14 53 PM


I may be missing something obvious here though, if so please let me know. I know you mentioned filtering out longer examples from the training set which I did not do. But does this really make such a substantial difference in the loss curve?

yechenzhi:
> I may be missing something obvious here though, if so please let me know. I know you mentioned filtering out longer examples from the training set which I did not do. But does this really make such a substantial difference in the loss curve?

Filtering out longer examples is implemented [here](https://github.com/pytorch/torchtune/blob/cd779783f9acecccbebc3c50265f6caf97fa99aa/torchtune/datasets/_preference.py#L61) and [here](https://github.com/pytorch/torchtune/blob/cd779783f9acecccbebc3c50265f6caf97fa99aa/recipes/configs/llama2/7B_lora_dpo_single_device.yaml#L49). I've checked the 'stack-exchange-paired' dataset, and many of the longer examples involve writing codes, which could make them more challenging to learn from. Additionally, I've been averaging the loss every 10 steps. Could you implement these changes and observe the new loss curve?

yechenzhi:
I'll rerun my recipes to see if there are any issues with them.

yechenzhi:
> I'll rerun my recipes to see if there are any issues with them.

Here is the code I have written for visualizing loss curves:
```
import numpy as np
import matplotlib.pyplot as plt
import re

filename = '/tmp/lora_dpo_output/log_1713751433.txt' # multi-device

steps = []
loss = []
lr = []
rewards_chosen = []
rewards_rejected = []
rewards_accuracies = []
rewards_margins = []
log_probs_rejected = []
log_probs_chosen = []
logits_rejected = []
logits_chosen = []
gpu_resources = []

with open(filename, 'r') as file:
    for line in file:
        match = re.match(r'Step (\d+) \| loss:(\S+) lr:(\S+) rewards/chosen:(\S+) rewards/rejected:(\S+) rewards/accuracies:(\S+) rewards/margins:(\S+) log_probs/rejected:(\S+) log_probs/chosen:(\S+) logits/rejected:(\S+) logits/chosen:(\S+) gpu_resources:(\S+)', line)
        if match:
            steps.append(int(match.group(1)))
            loss.append(float(match.group(2)))
            lr.append(float(match.group(3)))
            rewards_chosen.append(float(match.group(4)))
            rewards_rejected.append(float(match.group(5)))
            rewards_accuracies.append(float(match.group(6)))
            rewards_margins.append(float(match.group(7)))
            log_probs_rejected.append(float(match.group(8)))
            log_probs_chosen.append(float(match.group(9)))
            logits_rejected.append(float(match.group(10)))
            logits_chosen.append(float(match.group(11)))
            gpu_resources.append(float(match.group(12)))

steps = np.array(steps)
loss = np.array(loss)
lr = np.array(lr)
rewards_chosen = np.array(rewards_chosen)
rewards_rejected = np.array(rewards_rejected)
rewards_accuracies = np.array(rewards_accuracies)
rewards_margins = np.array(rewards_margins)
log_probs_rejected = np.array(log_probs_rejected)
log_probs_chosen = np.array(log_probs_chosen)
logits_rejected = np.array(logits_rejected)
logits_chosen = np.array(logits_chosen)
gpu_resources = np.array(gpu_resources)

window_size = 80

def get_avg(arr, window_size):
    avg_arr = []
    for i in range(0, len(arr), window_size):
        avg = np.mean(arr[i:i+window_size])
        avg_arr.append(avg)

    avg_arr = np.array(avg_arr)
    return avg_arr

avg_steps = get_avg(steps,window_size)
avg_lr = get_avg(lr, window_size)
avg_loss = get_avg(loss, window_size)
avg_rewards_accuracies = get_avg(rewards_accuracies, window_size)
avg_rewards_rejected = get_avg(rewards_rejected, window_size)
avg_rewards_chosen = get_avg(rewards_chosen, window_size)
avg_rewards_margins = get_avg(rewards_margins, window_size)
avg_log_probs_rejected = get_avg(log_probs_rejected, window_size)
avg_log_probs_chosen = get_avg(log_probs_chosen, window_size)
avg_logits_rejected = get_avg(logits_rejected, window_size)
avg_logits_chosen = get_avg(logits_chosen, window_size)
avg_gpu_resources = get_avg(gpu_resources, window_size)

plt.figure(figsize=(10, 6))
plt.plot(avg_steps, avg_loss, label='Average Loss')
plt.xlabel('Step')
plt.ylabel('Average')
plt.title('Average Metrics over Steps')
plt.legend()
plt.grid(True)
plt.show()
```

ebsmothers:
@yechenzhi apologies for the delayed response, I am just now getting to re-running the DPO recipe now. The loss curve I shared previously is run directly off of this PR via the command

```
tune run --nnodes 1 --nproc_per_node 2 lora_dpo_distributed --config llama2/7B_lora_dpo
```

I know you mentioned some data filtering up front in [this comment](https://github.com/pytorch/torchtune/pull/645#issuecomment-2047915895), are you doing that here as well? I'll let this one run for a bit longer, plot the smoothed loss curve, and get back to you once I have the results.

yechenzhi:
> I know you mentioned some data filtering up front in https://github.com/pytorch/torchtune/pull/645#issuecomment-2047915895, are you doing that here as well?

yechenzhi:
> I know you mentioned some data filtering up front in [this comment](https://github.com/pytorch/torchtune/pull/645#issuecomment-2047915895), are you doing that here as well? 

Yes, I applied data filtering in both TRL and Torchtune.

> I don't smooth exactly the same as in the code you provided above, but still it looks generally similar. This is with the data filtering from your other comment, just curious if you find this impacts the results substantially?

Indeed, I've observed a similar impact. I attribute it to the substantial size of the full dataset. It seems there's a discernible orderliness within the dataset, as previously [mentioned](https://github.com/pytorch/torchtune/pull/645#issuecomment-2047853377). However, onsidering our training duration of only 1000 steps and the shuffling of the dataset, focusing on the initial 64000 * 5 (approximately 70000 samples after filtering) makes the training process eaiser.

ebsmothers:
Thanks @yechenzhi for patiently bearing with all my questions here. I believe this PR should be good to go.

ebsmothers (approved):

PR #772, opened 2024-04-16
Some Quickfixes on W&B

tcapelle:
- Englobe the `OmegaConf` saving on the try block, to avoid crashing at any cost.
- Add repo telemetry

### DIFF
--- a/torchtune/utils/metric_logging.py
+++ b/torchtune/utils/metric_logging.py
@@ -176,7 +176,7 @@ class WandBLogger(MetricLoggerInterface):
         _, self.rank = get_world_size_and_rank()
 
         if self.rank == 0:
-            self._wandb.init(
+            run = self._wandb.init(
                 project=project,
                 entity=entity,
                 group=group,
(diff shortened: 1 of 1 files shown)


torchtune/utils/metric_logging.py
```diff
@@ -176,14 +176,15 @@ def __init__(
...
-            self._wandb.init(
+            run = self._wandb.init(
                 project=project,
                 entity=entity,
                 group=group,
                 reinit=True,
                 resume="allow",
                 **kwargs,
             )
+            run._label(repo="torchtune")
```
joecummings, line 187: Can you explain this a little more? What does the `repo` kwarg enable?

tcapelle: We get telemetry on the repo usage. We already do this by grabbing the git repo used to train but if people rename it or fork it can mess up the identification. We can share this data with you guys to see adoption.

tcapelle: In practice, it is just one extra key in: https://wandb.ai/capecape/torchtune/runs/ynahtfyw/files/config.yaml
the key `9`

joecummings: Makes sense, thanks for the explanation! (We would love to see adoption numbers soon :) )

joecummings (approved):
LGTM, thanks for the quickfixes

PR #774, opened 2024-04-16
Update build_docs.yaml

svekars:
Get torchtune version during the build and add so that it appear in the resulting HTML dropdown


docs/source/conf.py
```diff
@@ -78,6 +78,35 @@
...
+        version = ".".join(
+            torchtune_version_docs.split("/")[-1]
+            .split("-")[0]
+            .lstrip("v")
+            .split(".")[:2]
+        )
+        print(f"Version: {version}")
+        release = version
+        html_title = " ".join((project, version, "documentation"))
+    elif et_version_docs.startswith("refs/heads/"):
```
clee2000, line 99: What is et_version_docs?

clee2000 (approved):

docs/source/conf.py
svekars: @clee2000 - good catch, fixed the name

PR #783, opened 2024-04-17
Some extra W&B improvements

tcapelle:
- Overwrite the resolved config so we don't end up with a checkpoints folder full of configs
- Check if a run is active, use that run instead of creating a new one
- Create a unified X-axis following the `total_training_steps`

### DIFF
--- a/tests/torchtune/utils/test_metric_logging.py
+++ b/tests/torchtune/utils/test_metric_logging.py
@@ -10,6 +10,7 @@ from io import StringIO
 from typing import cast
 from unittest.mock import patch
 
+from omegaconf import OmegaConf
 from tensorboard.backend.event_processing.event_accumulator import EventAccumulator
 
 from tests.test_utils import assert_expected, captured_output
@@ -145,3 +146,20 @@ class WandBLoggerTest:
             logger.close()
 
             mock_log.assert_called_with(metric_dict, step=1)
+
+    def test_save_config(self) -> None:
+        with patch("wandb.init") as mock_init, patch(
+            "wandb.run", create=True
+        ) as mock_run, patch("OmegaConf.save") as mock_save, patch(
+            "wandb.save"
+        ) as mock_wandb_save:
+
+            logger = WandBLogger(project="test_project")
+            cfg = OmegaConf.create({"a": 1, "b": 2})
+
+            with patch.object(logger, "_wandb", mock_run):
+                logger.save_config(cfg)
+
+            expected_config_path = "torchtune_config.yaml"
+            mock_save.assert_called_once_with(cfg, expected_config_path)
+            mock_wandb_save.assert_called_once_with(expected_config_path)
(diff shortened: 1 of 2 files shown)


RdoubleA (commented):
While you're here, do you mind adding a test for `log_config` here? Just realized that this is missing: https://github.com/pytorch/torchtune/blob/4ef3be38fd33307ae0fd5ce71a2ffe92185ec076/tests/torchtune/utils/test_metric_logging.py#L128

Of course let me know if you'd like guidance on this

torchtune/utils/metric_logging.py
```diff
@@ -183,17 +183,16 @@ def __init__(
...
-                project=project,
-                entity=entity,
-                group=group,
-                reinit=True,
-                resume="allow",
-                dir=self.log_dir,
-                **kwargs,
-            )
-            run._label(repo="torchtune")
+        if self._wandb.run is None:
```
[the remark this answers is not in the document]
tcapelle: lol I had an else at some point

tcapelle:
what would you expect on this test, should I read a config.yaml, save it and check if it got saved?

torchtune/utils/metric_logging.py
```diff
@@ -205,13 +210,14 @@ def log_config(self, config: DictConfig) -> None:
             config (DictConfig): config to log
         """
         if self._wandb.run:
```
RdoubleA, line 212: does this also prevent ranks other than rank 0 from logging the config?

RdoubleA:
> what would you expect on this test, should I read a config.yaml, save it and check if it got saved?

You can pass in a dummy config: `cfg = OmegaConf.create({"a": 1, "b": 2 })` and mock a lot of the methods so you don't have to actually upload a config. You'd probably want to mock `wandb.run`, `OmegaConf.save`, `wandb.save` and make sure it returns the expected config. You can see the other tests for examples on some of the mocking.

torchtune/utils/metric_logging.py
RdoubleA: but is the fact that the run is active accessible from all ranks? will `if self._wandb.run` pass on rank 1 even though it was initialized on rank 0?

torchtune/utils/metric_logging.py
```diff
@@ -183,17 +183,25 @@ def __init__(
...
                 group=group,
-                reinit=True,
-                resume="allow",
                 dir=self.log_dir,
                 **kwargs,
             )
-            run._label(repo="torchtune")
+
+        if self._wandb.run:
+            self._wandb.run._label(repo="torchtune")
```
RdoubleA, line 197: nit: is this different from just keeping the `run._label(repo="torchtune")` line in the if statement above?

tests/torchtune/utils/test_metric_logging.py
```diff
@@ -145,3 +146,20 @@ def test_log_dict(self) -> None:
             logger.close()
 
             mock_log.assert_called_with(metric_dict, step=1)
+
+    def test_save_config(self) -> None:
```
RdoubleA, line 150: nit: rename to `test_log_config`

PR #785, opened 2024-04-17
Add Selective Activation Checkpointing 

lessw2020:
#### Context
This PR updates activation checkpointing (ac) to support selective layer and selective op activation checkpointing. 
It preserves the previous options enabled of full or None.  
This is controlled in the yaml file via:
enable_activation_checkpointing: bool
ac_mode:  ['full', 'selective']
ac_option:  [int, 'op'] 

if ac_mode is selective then the type of selective is determined based on ac_option.  
An integer represents checkpoint every x'th layer (i.e. 2 = checkpoint every other layer, 3 = every third, etc). 
'op' means to run selective op ac, where the ac is filtered by the op policy. 

Generically on llama-13B, selective AC 2 (every other layer) improved throughput +10% over No AC.

I updated the testing for llama3-8B where I tried to adjust the batch size under each setting to hit around 91GB.  I used 8 gpus with the idea of having less impact from model params and more finer grained tuning of the bs size and thus activations. This is not always perfectly possible as activations are chunky, but the net was selective AC 3 was the highest throughput followed by No AC.  Sel AC3 was +9% better throughput vs the original Full only option.  
For A100, 4090s etc. the actual best combo will vary but the point here is that selective AC provides generally better throughput options over the simple binary of Full (True/False). 

Screenshot 2024-04-19 at 1 01 53 PM

#### Changelog
- ...

#### Test plan
This code is largely a port from original source in torchtitan where it has already been tested.  However, I ran all 4 styles (none, full, sel ac op, sel ac 2) as shown above.
I also verified that the new impl of full matched the memory savings of the previous impl of full.


torchtune/utils/memory.py
```diff
@@ -6,29 +6,131 @@
 
 import gc
 
-from typing import Any, Dict, Optional, Set
+from collections import defaultdict
```
kartikayk, line 9: Changes in this file are a bit dense and I'm parsing them right now, but one thought I had was can we keep the current functionality and just add this as an additive change for now while we thoroughly stress test this? We can just update our default configs once we have tested this across fine-tunes and hardware types. WDYT?

torchtune/utils/memory.py
```diff
@@ -6,29 +6,131 @@
...
+        return ptd_checkpoint_wrapper(
+            module,
+            checkpoint_impl=CheckpointImpl.NO_REENTRANT,
+            checkpoint_fn=checkpoint,
+            use_reentrant=False,
+            preserve_rng_state=False,
+        )
+
+    # selective layer checkpointing...some checks in case we receive '2' or 2...
+    elif ac_mode == "selective" and (
```
kartikayk, line 71: Noob question: is "selective" and "1" the same as full? If so, should we just check for stle to be great than 1? Sorry if I misunderstand?

torchtune/utils/memory.py
```diff
@@ -6,29 +6,131 @@
...
+        """enables selective checkpointing of candidate layers.
+        Usage:
+        'selective_ac_option' with a positive 'int' value in config controls which layers to checkpoint.
+        1 == checkpointing every one (all).
+        2 == checkpoint every 2nd one
+        """
+        every_x_layer = int(ac_style)
+        assert (
+            every_x_layer >= 0
+        ), f"selective layer AC policy (every_x_layer) expects a positive integer, received {every_x_layer}"
```
kartikayk, line 84: nit: Let's replace ```asserts``` with if-Raise blocks. Those are more robust

torchtune/utils/memory.py
```diff
@@ -6,29 +6,131 @@
...
+        def _get_custom_policy(meta):
+            def _custom_policy(mode, func, *args, **kwargs):
+                mm_count_key = f"{mode}_mm_count"
+                if func == torch.ops.aten.mm.default:
+                    meta[mm_count_key] += 1
+                # Saves output of all compute ops, except every second mm
+
+                return func in no_recompute_list and not (
+                    func == torch.ops.aten.mm.default and meta[mm_count_key] % 2 == 0
+                )
```
kartikayk, line 45: This block could use more comments for noobs like me :)

torchtune/utils/memory.py
```diff
@@ -6,29 +6,131 @@
...
+
+            return _custom_policy
+
+        def selective_checkpointing_context_fn():
+            meta = defaultdict(int)
+            return _pt2_selective_checkpoint_context_fn_gen(_get_custom_policy(meta))
```
kartikayk, line 51: Same here, more comments on whats going on would be really helpful!

torchtune/utils/memory.py
```diff
@@ -6,29 +6,131 @@
...
+            checkpoint_impl=CheckpointImpl.NO_REENTRANT,
+            checkpoint_fn=checkpoint,
+            use_reentrant=False,
+            preserve_rng_state=False,
+        )
+
+    # selective layer checkpointing...some checks in case we receive '2' or 2...
+    elif ac_mode == "selective" and (
+        isinstance(ac_style, (int, float))
+        or (isinstance(ac_style, str) and ac_style.isdigit())
```
kartikayk, line 73: Should we just support int and if we get anything else we can fail? We can add more validation at the config layer to catch these type mis-matches.

torchtune/utils/memory.py
lessw2020: sure we can do that.  Do you have a way you want to toggle between old and new in the config or else I can just add a flag for use old or new as a temp solution.

torchtune/utils/memory.py
lessw2020: this is correct.  The reason I kept 'full' around was more for user friendliness in the config as 'full' is simpler.

torchtune/utils/memory.py
lessw2020: will do!

torchtune/utils/memory.py
lessw2020: np, will add!

torchtune/utils/memory.py
lessw2020: yes that sounds good...was just trying to be more 'user friendly' in terms of handling a variety of inputs. If you have config level validation that's a better way to enforce inputs.

torchtune/utils/memory.py
kartikayk: Ah yes that makes sense

torchtune/utils/memory.py
kartikayk: Good question, maybe before I answer that another noob question. I noticed the current ac changes go before the FSDP wrapping and the previous one after - does that matter?

kartikayk: Actually, sorry for the confusion on this. This particular function is used across many recipes. So this is my concrete proposal:

- can you move this functionality to maybe activation.py or something similar? I'd mention in the docstrings that this is under development or something like that and so we dont need to worry about BC or anything like that
- add a new config under configs/dev/7B_full.yaml with these changes

over time we can move this dev config to be the default and then also deprecate the stuff from memory.py. WDYT?

lessw2020: sure, activation.py sounds like a very modular way to implement this.
will work on that now and make a new config of 7B_full_experimental.yaml.

lessw2020: fyi, I've moved things into a new file activations.py and also added a new 7B_full_experimental.yaml file.
I then registered that yaml file in the tune registry as distributed recipe and everything seems to be working.
Internally, I've modified the yaml and code to be forward and back compat.
Specifically:
yaml file keeps the enable_activation_checkpointing: bool for old and new. 
True = do ac
False = do not do ac for all.
From there we check if ac_mode is present in the yaml - if not, then it just goes to the old AC assuming enable is true.
Otherwise it goes to the new activations.py and based on ac_mode ('full', 'selective') and option(int or 'op') does the right selective or full ac impl.

torchtune/utils/memory.py
lessw2020: updated

torchtune/utils/activations.py
```diff
@@ -0,0 +1,132 @@
...
+            return _pt2_selective_checkpoint_context_fn_gen(_get_custom_policy(meta))
+
+        return ptd_checkpoint_wrapper(
+            module,
+            checkpoint_impl=CheckpointImpl.NO_REENTRANT,
+            checkpoint_fn=checkpoint,
+            context_fn=selective_checkpointing_context_fn,
+            use_reentrant=False,
+            preserve_rng_state=False,
+        )
```
kartikayk, line 60: Mentioned this offline, but based on the numbers it doesn't seem like op-based AC is adding much. it's also really dense code which is hard to understand (at least for me). Given thiss observation (and if your agree), can we start with layer-based and drop op? WDYT/

kartikayk (commented):
Some noob questions, but starting to look good!

torchtune/utils/memory.py
lessw2020: actually we are removing selective op ac so will resolve this.

logs-Llama-2-7b-hf/config.json
```diff
@@ -0,0 +1 @@
+{"_name_or_path": "meta-llama/Llama-2-7b-hf", "architectures": ["LlamaForCausalLM"], "bos_token_id": 1, "eos_token_id": 2, "hidden_act": "silu", "hidden_size": 4096, "initializer_range": 0.02, "intermediate_size": 11008, "max_position_embeddings": 4096, "model_type": "llama", "num_attention_heads": 32, "num_hidden_layers": 32, "num_key_value_heads": 32, "pretraining_tp": 1, "rms_norm_eps": 1e-05, "rope_scaling": null, "tie_word_embeddings": false, "torch_dtype": "float16", "transformers_version": "4.31.0.dev0", "use_cache": true, "vocab_size": 32000}
```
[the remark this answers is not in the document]
lessw2020: ah sorry, these snuck in...yes removing.

recipes/full_finetune_distributed.py
```diff
@@ -266,6 +267,22 @@ def _setup_model(
...
+        ac_enabled = cfg_file.get("enable_activation_checkpointing", "False")
+
+        # we can't guarantee selective ac is present in the config...
+        ac_mode = cfg_file.get("ac_mode", None)
+        ac_option = cfg_file.get("ac_option", None)
+
+        print(f"ac_enabled: {ac_enabled}, ac_mode: {ac_mode}, ac_option: {ac_option}")
+
+        if ac_enabled and ac_mode:
+            apply_selective_activation_checkpointing(
```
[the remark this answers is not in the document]
lessw2020: when we first did AC for FSDP, it used to require being done after but in theory it should not have mattered.  Regardless, there was an error at the time with autograd engine unless done after...anyway, the old way was to do it after but now it can be done up front.  
Rohan might have more info on this as I don't know the more detailed underlying reasons from back then re: autograd error.

recipes/full_finetune_distributed.py
```diff
@@ -456,7 +476,8 @@ def train(self) -> None:
                         {
                             "loss": loss.item(),
                             "lr": self._optimizer.param_groups[0]["lr"],
-                            "gpu_resources": torch.cuda.memory_allocated(),
+                            "gpu_mem_alloc": torch.cuda.memory_allocated() / _gb,
```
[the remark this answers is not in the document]
lessw2020: updated.

torchtune/utils/activations.py
lessw2020: sounds good - removed.

recipes/configs/llama2/7B_full.yaml
```diff
@@ -42,12 +42,12 @@ checkpointer:
     pytorch_model-00002-of-00002.bin
   ]
   recipe_checkpoint: null
-  output_dir: /tmp/Llama-2-7b-hf
+  output_dir: ./logs-Llama-2-7b-hf
```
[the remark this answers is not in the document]
lessw2020: thanks! have reverted.

torchtune/utils/memory.py
lessw2020: removed

recipes/configs/dev/8B_full_experimental.yaml
```diff
@@ -0,0 +1,80 @@
+# Config for multi-device full finetuning in full_finetune_distributed.py
+# using a Llama3 8B model
+#
+# This config assumes that you've run the following command before launching
+# this run:
+#   tune download meta-llama/Meta-Llama-3-8B --output-dir /tmp/Meta-Llama-3-8B --hf-token <HF_TOKEN>
+#
+# To launch on 4 devices, run the following command from root:
+#   tune run --nproc_per_node 4 full_finetune_distributed --config llama3/8B_full
```
rohan-varma, line 9: Launch command does not seem correct

recipes/configs/dev/8B_full_experimental.yaml
```diff
@@ -0,0 +1,80 @@
...
+#
+# You can add specific overrides through the command line. For example
+# to override the checkpointer directory while launching training
+# you can run:
+#   tune run --nproc_per_node 4 full_finetune_distributed --config llama3/8B_full checkpointer.checkpoint_dir=<YOUR_CHECKPOINT_DIR>
```
rohan-varma, line 14: same comment as (9)

recipes/configs/dev/8B_full_experimental.yaml
```diff
@@ -0,0 +1,80 @@
...
+gradient_accumulation_steps: 1
+
+
+# Training env
+device: cuda
+
+# Memory management
+enable_activation_checkpointing: False
+ac_mode: 'selective'  # ['selective', 'full']
+ac_option: 2 # [int] = ac every positive int layer
```
rohan-varma, line 69: This is an interesting way to configure and IMO seems a little brittle. Is there any guidance around choosing this setting for users, and what if user selects an integer that exceeds the number of layers in the model?

recipes/full_finetune_distributed.py
```diff
@@ -266,6 +271,22 @@ def _setup_model(
         if self._dtype == torch.bfloat16:
             model = model.to(torch.bfloat16)
 
+        # We currently have two versions of activation checkpointing in this recipe
+        # for testing and BC purposes. ``enable_activation_checkpointing`` controls
+        # the older version of AC and this behavior is unchanged
+        # ac_mode and ac_option together control selective AC. This is only enabled
+        # when these are set AND ``enable_activation_checkpointing`` is set to False
+        # We'll clean this up as soon as testing of AC is complete
```
rohan-varma, line 279: Is there an issue and an owner?

recipes/full_finetune_distributed.py
```diff
@@ -266,6 +271,22 @@ def _setup_model(
...
+        ac_mode = ac_mode
+        ac_option = ac_option
+
+        if (not enable_activation_checkpointing) and (ac_mode is not None):
```
rohan-varma, line 283: I understand why its this way but still reads a little weird. Like we have a check for `not enable_activation_checkpointing`, and underneath the check we apply AC (granted it is selective).

recipes/full_finetune_distributed.py
```diff
@@ -290,10 +311,12 @@ def _setup_model(
         # Ensure no params and buffers are on meta device
         utils.validate_no_params_on_meta_device(model)
 
-        if enable_activation_checkpointing:
+        # original activation checkpointing (full) - flip the condition above
+        if enable_activation_checkpointing and ac_mode is None:
```
rohan-varma, line 315: What happens if user configures `enable_activation_checkpointing` and ac_mode is not None? Silently no use of AC/

torchtune/utils/activations.py
```diff
@@ -0,0 +1,91 @@
...
+from torch.distributed.algorithms._checkpoint.checkpoint_wrapper import (
+    checkpoint_wrapper as ptd_checkpoint_wrapper,
+    CheckpointImpl,
+)
+from torch.utils.checkpoint import checkpoint
+
+
+# Uses PTD FSDP AC wrapper
+# currently selective per layer checkpointing are supported
+def checkpoint_wrapper(module, ac_mode, ac_style):
```
rohan-varma, line 19: We generally have agreed to type our public APIs (cc @ebsmothers) and add docstrings and unittests where applicable. 

If these are private / prototype APIs, let's instead hide them with the appropriate `_`. We should avoid exposing new public APIs that are untested and undocumented where possible. Thanks!

torchtune/utils/activations.py
```diff
@@ -0,0 +1,91 @@
...
+
+    if ac_mode == "full":
+        return ptd_checkpoint_wrapper(
+            module,
+            checkpoint_impl=CheckpointImpl.NO_REENTRANT,
+            checkpoint_fn=checkpoint,
+            use_reentrant=False,
```
rohan-varma, line 26: why pass both checkpoint_impl and use_reentrant flags? Feels redundant.

torchtune/utils/activations.py
```diff
@@ -0,0 +1,91 @@
...
+
+        if not (every_x_layer >= 0):
+            raise ValueError(
+                f"Selective layer AC policy (every_x_layer) expects a positive integer, received {every_x_layer}"
+            )
+
+        checkpoint_wrapper.__dict__.setdefault("_count", 0)
+
+        checkpoint_wrapper._count += 1
+        if not every_x_layer or checkpoint_wrapper._count % every_x_layer == 0:
```
rohan-varma, line 48: So if every_x_layer = 2, then it seems it would get applied at layer 2, 4, 6, etc. What about layer 0 though?

torchtune/utils/activations.py
```diff
@@ -0,0 +1,91 @@
...
+
+    else:
+        raise NotImplementedError(
+            "Unknown AC type or AC config. Only selective op and selective layer ac implemented currently."
+        )
+
+
+def apply_selective_activation_checkpointing(
+    model: nn.Module,
+    ac_mode: str,
```
rohan-varma, line 68: What are the allowed options here? How do we prevent against errors?

torchtune/utils/activations.py
```diff
@@ -0,0 +1,91 @@
...
+    ac_option: Optional[Union[int, str]],
```
rohan-varma, line 69: Why int or str?

torchtune/utils/activations.py
```diff
@@ -0,0 +1,91 @@
...
+) -> None:
+    """Utility to setup activation checkpointing and wrap the model for checkpointing.
+
+    Args:
+        model (nn.Module): Model to setup activation checkpointing.
+        ac_mode (str): Activation checkpointing mode. ['none', 'full', 'selective']
+        ac_option (Optional[Union[int, str]]): Activation checkpointing option.
+            - If ac_mode is 'selective', ac_option can be an integer or a string
```
rohan-varma, line 77: Why do we need both int and string?

torchtune/utils/activations.py
```diff
@@ -0,0 +1,91 @@
...
+              representing the number of layers to checkpoint.
```
rohan-varma, line 78: Is it number of layers to checkpoint or skip # of layers in between checkpointed layers?

torchtune/utils/activations.py
```diff
@@ -0,0 +1,91 @@
...
+            - If ac_mode is 'selective' and ac_option is 'op', then selective op ac is run.
+            - If ac_mode is 'none' or 'full, ac_option is ignored.
+    """
+
+    for layer_id, transformer_block in enumerate(model.layers):
```
rohan-varma, line 83: We should document if we're assuming the passed in module requires a `layers` attribute.

torchtune/utils/activations.py
```diff
@@ -0,0 +1,91 @@
...
+        if ac_mode in ("full", "selective"):
+
+            transformer_block = checkpoint_wrapper(
```
rohan-varma, line 86: This in place modification is quite risky and brittle in general - for example, will have to carefully use it with FSDP and other wrapping APIs. Ideally we should do this via hooks, but of course this will require a lot of refactoring around our activation checkpointing infra.

torchtune/utils/activations.py
```diff
@@ -0,0 +1,91 @@
...
+        ac_mode (str): Activation checkpointing mode. ['none', 'full', 'selective']
+        ac_option (Optional[Union[int, str]]): Activation checkpointing option.
+            - If ac_mode is 'selective', ac_option can be an integer or a string
+              representing the number of layers to checkpoint.
+            - If ac_mode is 'selective' and ac_option is 'op', then selective op ac is run.
+            - If ac_mode is 'none' or 'full, ac_option is ignored.
+    """
+
+    for layer_id, transformer_block in enumerate(model.layers):
+        if ac_mode in ("full", "selective"):
```
rohan-varma, line 84: Silently not applying AC if `ac_mode` is not within this tuple seems bad, especially if its undocumented?

torchtune/utils/activations.py
```diff
@@ -0,0 +1,91 @@
...
+# LICENSE file in the root directory of this source tree.
+
+from typing import Optional, Union
+
+from torch import nn
+from torch.distributed.algorithms._checkpoint.checkpoint_wrapper import (
+    checkpoint_wrapper as ptd_checkpoint_wrapper,
+    CheckpointImpl,
+)
+from torch.utils.checkpoint import checkpoint
```
rohan-varma, line 14: Unused I think? Not sure why linter did not pick this up

torchtune/utils/activations.py
```diff
@@ -0,0 +1,91 @@
...
+    elif ac_mode == "selective":
+        """enables selective checkpointing of candidate layers.
+        Usage:
+        'selective_ac_option' with a positive 'int' value in config controls which layers to checkpoint.
+        1 == checkpointing every one (all).
+        2 == checkpoint every 2nd one
+        """
+        every_x_layer = int(ac_style)
+
+        if not (every_x_layer >= 0):
```
rohan-varma, line 40: every_x_layer < 0 is simpler?

torchtune/utils/activations.py
```diff
@@ -0,0 +1,91 @@
...
+            raise ValueError(
+                f"Selective layer AC policy (every_x_layer) expects a positive integer, received {every_x_layer}"
+            )
+
+        checkpoint_wrapper.__dict__.setdefault("_count", 0)
+
+        checkpoint_wrapper._count += 1
```
rohan-varma, line 47: Think the bump should be done after the check

torchtune/utils/activations.py
rohan-varma, line 48: `every_x_layer == 0` seems better than `if not every_x_layer` to guard against `every_x_layer` being None which is an unexpected state for this variable

PR #786, opened 2024-04-17
Fix LoRA finetune single device link in README

alexrs:
#### Context
- The link to the LoRA recipe in the Single device case in the README pointed to `recipes/lora_finetune_distributed.py` instead of `recipes/lora_finetune_single_device.py`. 

#### Changelog
- Fix LoRA finetune single device link in README

#### Test plan
- When I click on the link, it opens `recipes/lora_finetune_single_device.py`.


RdoubleA (approved):
Appreciate you for catching and fixing this :) just need to sign the CLA

PR #789, opened 2024-04-18
Gemma lora

solitude-alive:
#### Context
- Create lora fine-tune for Gemma model.

#### Changelog
- ...

#### Test plan
- ....

Screenshot 2024-04-18 at 7 37 25 PM

It can work with `apply_lora_to_mlp = True, apply_lora_to_output = False`, but not `apply_lora_to_output = True`.
The current Gemma model does not have output_layer, the `output` is calculated by `output = F.linear(h, self.tok_embeddings.weight).float()`, so I don't know if need to set `apply_lora_to_output = True `.


torchtune/models/gemma/_component_builders.py
```diff
@@ -121,3 +124,239 @@ def gemma_mlp(dim: int, hidden_dim: int) -> FeedForward:
...
+    layer = TransformerDecoderLayer(
+        attn=self_attn,
+        mlp=mlp,
+        sa_norm=GemmaRMSNorm(embed_dim, eps=norm_eps),
+        mlp_norm=GemmaRMSNorm(embed_dim, eps=norm_eps),
+    )
+    tok_embeddings = nn.Embedding(vocab_size, embed_dim)
+
+    # TODO: Fix the apply to output_proj.
+    assert apply_lora_to_output is False, "LoRA is not currently supported for the output projection."
```
ebsmothers, line 225: Personally I think this is the right way to handle this (assuming we don't want to support LoRA on the tied embedding weights). Maybe we can also consider not exposing the arg `apply_lora_to_output` at all, since it is basically trivial at this point? 

Re LoRA state dict validation, we could consider just doing e.g. `getattr(cfg_model, "apply_lora_to_output", False)`  instead of our current usage of `cfg_model.apply_lora_to_output`. But cc @RdoubleA for thoughts on the config experience here

torchtune/models/gemma/_component_builders.py
```diff
@@ -121,3 +124,239 @@ def gemma_mlp(dim: int, hidden_dim: int) -> FeedForward:
...
+    max_seq_len: int,
+    attn_dropout: float = 0.0,
+    norm_eps: float = 1e-6,
+    rope_base: int = 10_000,
+    norm_embeddings: bool = True,
+    # LoRA args
+    lora_rank: int,
+    lora_alpha: float,
+    lora_dropout: float = 0.0,
+    quantize_base: bool = False,
```
ebsmothers, line 151: Do you plan to enable QLoRA too? If not I would remove this arg

recipes/configs/gemma/2B_lora.yaml
```diff
@@ -0,0 +1,83 @@
...
+# this run:
+#   tune download google/gemma-2b --hf-token <HF_TOKEN> --output-dir /tmp/gemma --ignore-patterns ""
+#
+# To launch on 4 devices, run the following command from root:
+#   tune run --nnodes 1 --nproc_per_node 4 lora_finetune_distributed --config gemma/2B_lora
+#
+# You can add specific overrides through the command line. For example
+# to override the checkpointer directory while launching training
+# you can run:
+#   tune run --nnodes 1 --nproc_per_node 4 lora_finetune_distributed --config gemma/2B_lora checkpointer.checkpoint_dir=<YOUR_CHECKPOINT_DIR>
```
ebsmothers, line 14: I think this was mentioned on Discord, but given that Gemma is such a small model already, any reason to not also add a single-device LoRA config here?

ebsmothers (commented):
Thanks for this PR! Left a few comments but no major concerns on a first pass. I think the high-level approach of simply not exposing the output projection to LoRA makes sense (and we can always revisit proper LoRA embedding/tied weight support)

torchtune/models/gemma/_component_builders.py
solitude-alive: Yes, I update the QLoRA code.

recipes/configs/gemma/2B_lora.yaml
solitude-alive: Yes, I overlooked that. Now, it support lora and qlora on single device.

solitude-alive:
This is the lora single device test:
single


This is the qlora single device test:
Screenshot 2024-04-19 at 1 20 34 PM

recipes/configs/gemma/2B_lora.yaml
```diff
@@ -0,0 +1,82 @@
...
+
+lr_scheduler:
+  _component_: torchtune.modules.get_cosine_schedule_with_warmup
+  num_warmup_steps: 100
+
+loss:
+  _component_: torch.nn.CrossEntropyLoss
+
+# Fine-tuning arguments
+batch_size: 2
```
RdoubleA, line 63: wonder if we should just increase batch size here, given that we have a bit of breathing room with memory? what's the peak memory you observe with this config?

torchtune/models/gemma/_component_builders.py
RdoubleA: let's just remove this field entirely, and make a note in the docstring that output projection lora is not supported because it is tied to token embeddings? seems like the simplest approach

torchtune/models/gemma/_model_builders.py
```diff
@@ -3,24 +3,29 @@
...
+from functools import partial
 
 """
 Model builders build specific instantiations using component builders. For example
 the ``gemma_2b`` model builder uses the ``gemma`` component builder.
 """
 
 
-def gemma_2b() -> TransformerDecoder:
+def gemma_2b() -> GemmaTransformerDecoder:
```
RdoubleA, line 22: good catch!

RdoubleA:
do you also mind sharing what your peak memory allocated was for both the lora and qlora runs? and to be extra sure, do you mind making sure the checkpoints save correctly :) Left some small comments, otherwise it looks nearly ready to go

torchtune/models/gemma/_component_builders.py
solitude-alive: Thanks, I add `cfg_model.apply_lora_to_output = getattr(cfg_model, "apply_lora_to_output", False)` in `lora_finetune_single_device.py` and `lora_finetune_distributed.py`, and remove the arg `apply_lora_to_output ` in LoRA and QLoRA yaml file.

RdoubleA: could we also remove it from this builder to prevent confusion?

recipes/configs/gemma/2B_lora.yaml
solitude-alive:
```
INFO:torchtune.utils.logging:Memory Stats after model init:
{'peak_memory_active': 4.917779456, 'peak_memory_alloc': 4.917779456, 'peak_memory_reserved': 5.65182464}
```

I increase it to 4.

torchtune/models/gemma/_component_builders.py
solitude-alive: Yes, I remove it.

torchtune/models/gemma/_component_builders.py
```diff
@@ -121,3 +124,237 @@ def gemma_mlp(dim: int, hidden_dim: int) -> FeedForward:
     up_proj = nn.Linear(dim, hidden_dim, bias=False)
     activation = nn.GELU(approximate="tanh")
     return FeedForward(gate_proj=gate_proj, down_proj=down_proj, up_proj=up_proj, activation=activation)
+
+
+def lora_gemma(
+    lora_attn_modules: List[LORA_ATTN_MODULES],
+    apply_lora_to_mlp: bool = False,
+    apply_lora_to_output: bool = False,
```
RdoubleA, line 132: remove this as well?

torchtune/models/gemma/_component_builders.py
```diff
@@ -121,3 +124,237 @@ def gemma_mlp(dim: int, hidden_dim: int) -> FeedForward:
...
+    """
+    Return a version of Gemma with LoRA applied to some of the linear layers in its self-attention modules.
+
+    Args:
+        lora_attn_modules (List[LORA_ATTN_MODULES]): list of which linear layers
+            LoRA should be applied to in each self-attention block. Options are
+            ``{"q_proj", "k_proj", "v_proj", "output_proj"}``.
+        apply_lora_to_mlp (bool): whether to apply LoRA to the MLP in each transformer layer.
+            Default: False
+        apply_lora_to_output (bool): whether to apply LoRA to the model's final output projection.
```
RdoubleA, line 162: remove and keep the note with apply_lora_to_mlp

torchtune/models/gemma/_model_builders.py
```diff
@@ -40,3 +45,40 @@ def gemma_tokenizer(path: str) -> SentencePieceTokenizer:
     tokenizer = SentencePieceTokenizer(path)
     tokenizer.pad_id = 0
     return tokenizer
+
+
+def lora_gemma_2b(
+    lora_attn_modules: List[LORA_ATTN_MODULES],
+    apply_lora_to_mlp: bool = False,
+    apply_lora_to_output: bool = False,
```
RdoubleA, line 53: remove

RdoubleA (approved):
few more nits, but looks good otherwise!

solitude-alive:
LoRA single device:
```
INFO:torchtune.utils.logging:Memory Stats after model init:
{'peak_memory_active': 5.547595264, 'peak_memory_alloc': 5.547595264, 'peak_memory_reserved': 5.559549952}

Step 100 | peak_memory_active:14.325987328 peak_memory_alloc:14.325987328 peak_memory_reserved:24.664604672 
```

QLoRA single device:
```
INFO:torchtune.utils.logging:Memory Stats after model init:
{'peak_memory_active': 4.887322112, 'peak_memory_alloc': 4.887322112, 'peak_memory_reserved': 5.802819584}

Step 100 | peak_memory_active:11.555596288 peak_memory_alloc:11.555596288 peak_memory_reserved:24.301797376 
```


LoRA checkpoints save:
lora

QLoRA checkpoints save:
Screenshot 2024-04-20 at 12 11 00 AM

torchtune/models/gemma/_component_builders.py
solitude-alive: Yes, I remove them.

recipes/lora_finetune_distributed.py
```diff
@@ -287,6 +287,10 @@ def _setup_model(
                 f"Model instantiation took {time.perf_counter() - init_start:.2f} secs"
             )
 
+            cfg_model.apply_lora_to_output = getattr(
+                cfg_model, "apply_lora_to_output", False
+            )
```
ebsmothers, line 292: Technically you can probably just set this to `apply_lora_to_output` and not modify `cfg_model`, but not a huge deal either way

ebsmothers (approved):
One small comment, but not a blocker for merging. Thanks so much for adding this!

recipes/lora_finetune_distributed.py
solitude-alive: Thank you for your suggestion, I update them in `recipes/lora_finetune_distributed.py` and `recipes/lora_finetune_single_device.py`.

PR #790, opened 2024-04-18
MPS support

maximegmd:
#### Context
- For testing purposes it can be useful to run directly on a local Mac computer.

#### Changelog
- Checks support for BF16 on MPS device.
- Added a configuration targeting MPS, changes to path were required due to the way Mac resolves symlinks in /tmp as /private/ActualPath.
- Set optimizer to Adam instead of AdamW to fit in memory on 64GB devices.

#### Test plan
- Ran a training job on Mistral 7b full finetune.
- The current test jobs are very CUDA specific, maybe this could be changed as well?
- We may need to integrate Mac runners in the pipeline.
- Some dependencies such as bitsandbytes are not yet Mac compatible.


joecummings:
@maximegmd This is awesome! Can you post some loss curves for the finetune you ran?

maximegmd:
> @maximegmd This is awesome! Can you post some loss curves for the finetune you ran?

I will complete a run during the weekend, losses looked fine but the Llama3 release changed my priorities ^^

kartikayk (commented):
Thanks so much for making this change, supporting MPS has been on our TODO!

I'm a bit confused about how this is working because the device param is used to fetch the device using [this utility function](https://github.com/pytorch/torchtune/blob/main/torchtune/utils/_device.py#L101), which in turn depends on [this function](https://github.com/pytorch/torchtune/blob/main/torchtune/utils/_device.py#L52). We seemingly are never actually returning mps as the device. So how is this working? This'll just default to CPU I think

maximegmd:
device is not None when this function is called so it just passes 'mps' to torch.device() which is the expected pytorch name.

But you are correct that there is room for improvement to automatically return mps when device is not manually specified in the config.

kartikayk:
Oh good point, totally glossed over the fact the ```device = torch.device(device)``` outside the if block. Yup sounds good. What's the iter/sec you're getting with this on a mac?

maximegmd:
If I recall correctly it was around 20s/it but I suspect I was swapping a bit so I can probably improve the speed. The main issue is bitsandbytes not supporting MPS so it uses quite a bit of memory for the optimizer state. 

I will try to push a llama3 config tomorrow with some numbers now that my llama3 finetune is running :)

recipes/configs/mistral/7B_full_mps.yaml
```diff
@@ -0,0 +1,70 @@
...
+# This config assumes that you've run the following command before launching
+# this run:
+#   tune download mistralai/Mistral-7B-v0.1 --output-dir ./tmp/Mistral-7B-v0.1
+#
+# You can then run the recipe with:
+#   tune run full_finetune_single_device --config ./recipes/configs/mistral/7B_full_mps.yaml
+# Tokenizer
+tokenizer:
+  _component_: torchtune.models.mistral.mistral_tokenizer
+  path: ./tmp/Mistral-7B-v0.1/tokenizer.model
```
joecummings, line 16:
```suggestion
  path: /tmp/Mistral-7B-v0.1/tokenizer.model
```



nit: we actually want to download to the real `/tmp` directory in our examples so as to not persist a bunch of models on disk. Currently, this will write to `$CWD/tmp/Mistral-7B-v0.1`. This should be changed in a few places in the config.

maximegmd: on mac /tmp resolves to /private/tmp and it actually breaks the symlinks that are created, that's why I ended up doing this :(

maximegmd:
Here is a train run on Gemma 2B, sadly laptop went to sleep right before the end but this is a 14 hours run, should be representative enough. 
[log_1715964526.txt](https://github.com/pytorch/torchtune/files/15361933/log_1715964526.txt)

recipes/configs/mistral/7B_full_mps.yaml
joecummings: Gotcha, then I would actually prefer something like `./models/` directory so we don't confuse this with an actual tmp dir.

maximegmd:
The goal is to merge indeed, but considering the issue raised by @SalmanMohammadi maybe we should off and lock to pytorch 2.4+ ?

SalmanMohammadi:
bump bump.

Is this still planned @maximegmd? If not, happy to pick it up. It's been invaluable to me so far, so would be happy to other users be able to get up and running without needing big (medium?) GPUs.

byjlw:
This has been open without changes for 6 weeks. Can we merge this in @joecummings?

maximegmd:
This just needs to be updated to use the latest utiles functions, I will do it tomorrow, then it should be ready to merge unless I am not seeing something

SalmanMohammadi:
> This just needs to be updated to use the latest util functions, I will do it tomorrow, then it should be ready to merge unless I am not seeing something

Nice! Thanks so much. One small thing I've experienced with the recent torchao dependency version upgrade, is that `pip install torchao==0.3.1` doesn't find any suitable builds. I've gotten around this by installing from source (or `pip install --no-cache-dir git+https://github.com/pytorch/ao.git@v0.3.1-rc1`) - if you error out here make sure you've install `setuptools` and `wheel`. 

If this sounds like a sensible way around this, perhaps it's worth adding to the README?

ebsmothers:
> Nice! Thanks so much. One small thing I've experienced with the recent torchao dependency version upgrade, is that pip install torchao==0.3.1 doesn't find any suitable builds. I've gotten around this by installing from source (or pip install --no-cache-dir git+https://github.com/pytorch/ao.git@v0.3.1-rc1) - if you error out here make sure you've installedsetuptools and wheel.

I assume this is specifically on MPS? cc @msaroufim for thoughts here

SalmanMohammadi:
> I assume this is specifically on MPS? cc @msaroufim for thoughts here

Yeah just on my machine, it's been working fine on Linux.

byjlw:
Can you update the readme with an example for MPS?

msaroufim:
Was on PTO but `torchao` as of `0.3.1` is still a linux only library, mac binaries are possible just not high priority right now

byjlw:
checking in again to see what the plan is on merging. Still haven't seen any movement. Would love to see mps make it to main.

SalmanMohammadi:
Closed due to #1706. For anyone following along, feel free to try out torchtune on MPS and let us know how you get along!

PR #796, opened 2024-04-18
fixed mixed precision in FSDP

denadai2:
#### Context
- FSDP initializes but does not use MixedPrecision

### DIFF
--- a/torchtune/utils/_distributed.py
+++ b/torchtune/utils/_distributed.py
@@ -211,7 +211,7 @@ def wrap_fsdp(
             model,
             auto_wrap_policy=wrap_policy,
             device_id=device,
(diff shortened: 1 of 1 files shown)


ebsmothers:
Hi @denadai2 thanks for the PR! Actually this utility is not used in any of our recipes ([code search](https://github.com/search?q=repo%3Apytorch%2Ftorchtune%20wrap_fsdp&type=code) shows it's only in the unit test and init file). So I think this is dead code anyways. In fact we don't really support mixed precision training and instead prefer to use true bf16 training, which requires less memory. Let me know if this makes sense

ebsmothers:
Closing this (see my previous comment for why)

PR #803, opened 2024-04-18
remove check for cuda version and package so the bf16 check passes on non Nvidia CUDA devices that support bf16

supernovae:
Remove checks that prohibit AMD cards that support pytorch as cuda devices from training.

#### Context
I have a 7900xtx running pytorch 2.2 w/ROCM 6.1 and wanted to train a mistral/llama model but ran into an error that my card didn't support bf16

#### Changelog
Remove cuda package and version check - returns none for rocm

#### Test plan
Training mistral 07b on my 7900xtx using lora.

### DIFF
--- a/torchtune/utils/precision.py
+++ b/torchtune/utils/precision.py
@@ -18,7 +18,6 @@ from typing import (
 
 import torch
 import torch.nn as nn
-from pkg_resources import packaging
 
 from torch.cuda.amp import GradScaler
 from torch.distributed.fsdp.sharded_grad_scaler import ShardedGradScaler
--- a/torchtune/utils/precision.py
+++ b/torchtune/utils/precision.py
@@ -67,9 +67,7 @@ def list_dtypes() -> List[str]:
 def verify_bf16_support():
     return (
         torch.cuda.is_available()
-        and torch.version.cuda
         and torch.cuda.is_bf16_supported()
-        and packaging.version.parse(torch.version.cuda).release >= (11, 0)
         and torch.distributed.is_nccl_available()
         and torch.cuda.nccl.version() >= (2, 10)
     )


kartikayk:
@supernovae thanks so much for the PR, the change has been on my TODO list for a while and I'm glad you got to it first. Mind signing the CLA and then I can accept and merge.

BTW I'd be curious on what sort of training speed you're seeing (just a simple iter/sec number from tqdm would be good enough).

supernovae:
I did sign the CLA, i'll double check to make sure i didn't fat finger the username :) 

I am getting 1.33 to 1.60it/s if i leave my computer alone

is there a list of it/sec for 3090/4090 cards?

kartikayk:
Hmm, I'm getting around 3it/sec on a 4090 so 2x faster. To confirm, you're training using the default config?

kartikayk (approved):
Thanks for making the change! I added one fix to fix the lint error.

supernovae:
> Hmm, I'm getting around 3it/sec on a 4090 so 2x faster. To confirm, you're training using the default config?

I am using default config.

I did notice this error:

`/home/me/pyenv/versions/3.11.8/lib/python3.11/site-packages/torchtune/modules/attention.py:207: UserWarning: 1Torch was not compiled with memory efficient attention. (Triggered internally at ../aten/src/ATen/native/transformers/hip/sdp_utils.cpp:515.)
  output = nn.functional.scaled_dot_product_attention(
`
a quick check in python `toch.backends.cuda.flash_sdp_enabled()` returns True but i found there is a specific ROCM flash attention patch pending merge into pytorch 2.3.0 nightly with a reference to some poor performing PR from before.

https://github.com/pytorch/pytorch/pull/121561

I'm not sure if that PR is the "memory efficient flash attention" but i'm hopefully i can see it merged into a nightly i can try :+1:

PR #807, opened 2024-04-19
fix max seq len

ebsmothers:
Test plan:

```
...
1|21|Loss: 2.320160388946533:   0%|▏ 
```


kartikayk (approved):
Thanks for the fix. I finally found the reference:

```
We trained the models on sequences of 8,192 tokens, using a mask to ensure self-attention does not cross document boundaries.
```

kartikayk:
One note, we do double are RoPE cache. But I dont think this will meaningfully impact training.

rohan-varma:
@kartikayk Is there a way to prevent this that's used in pretraining or other finetuning libs that we don't have? Seems our implementation makes RoPE directly proportional to max_seq_len, right? https://github.com/pytorch/torchtune/blob/main/torchtune/modules/position_embeddings.py

PR #816, opened 2024-04-19
Added --output-dir-use-symlinks for tune download with default as auto

### LINKED ISSUE #813: tune download has no option to set local_dir_use_symlinks
For `tune download`, It would be helpful to have `local_dir_use_symlinks` option for the user to choose `auto`,`True` or `False`.
Currently it is set to `auto` which stores the files in .cache directory and uses symlinks for big files.

Ref: https://huggingface.co/docs/huggingface_hub/en/guides/download

apthagowda97:
#### Context
- Fixes #813 
- Added --output-dir-use-symlinks  option to `tune download` with default as `auto`.

#### Changelog
- 

#### Test plan
-

### DIFF
--- a/torchtune/_cli/download.py
+++ b/torchtune/_cli/download.py
@@ -9,11 +9,11 @@ import os
 import textwrap
 
 from pathlib import Path
+from typing import Literal, Union
 
 from huggingface_hub import snapshot_download
 from huggingface_hub.utils import GatedRepoError, RepositoryNotFoundError
 from torchtune._cli.subcommand import Subcommand
-from typing import List, Literal, Optional, Union
 
 
 class Download(Subcommand):
@@ -72,7 +72,7 @@ class Download(Subcommand):
             "--output-dir-use-symlinks",
             type=str,
             required=False,
-            default="False",
+            default="auto",
             help=(
                 "To be used with `output-dir`. If set to 'auto', the cache directory will be used and the file will be"
                 " either duplicated or symlinked to the local directory depending on its size. It set to `True`, a"
@@ -106,7 +106,7 @@ class Download(Subcommand):
         if use_symlinks_lowercase == "true":
             output_dir_use_symlinks = True
         elif use_symlinks_lowercase == "false":
-            output_dir_use_symlinks= False
+            output_dir_use_symlinks = False
         elif use_symlinks_lowercase == "auto":
             output_dir_use_symlinks = "auto"
         else:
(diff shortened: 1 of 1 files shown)


torchtune/_cli/download.py
```diff
@@ -67,6 +68,18 @@ def _add_arguments(self) -> None:
             default="./model",
             help="Directory in which to save the model.",
         )
+        self._parser.add_argument(
+            "--output-dir-use-symlinks",
+            type=str,
+            required=False,
+            default="False",
```
joecummings, line 75: Can we default this to 'auto' to mimic the behavior from before?

torchtune/_cli/download.py
```diff
@@ -86,11 +99,28 @@ def _add_arguments(self) -> None:
     def _download_cmd(self, args: argparse.Namespace) -> None:
         """Downloads a model from the Hugging Face Hub."""
         # Download the tokenizer and PyTorch model files
+
+        # Raise if local_dir_use_symlinks is invalid
+        output_dir_use_symlinks: Union[Literal["auto"], bool]
+        use_symlinks_lowercase = args.output_dir_use_symlinks.lower()
```
joecummings, line 105: `argparse` can actually handle all of this logic within the `add_argument` method.

e.g.

```
self._parser.add_argument(
    ...,
    choices = ["true", "false", "auto"],
)
```

joecummings (commented):
Left a couple comments, but overall looks good! Can you also please sign the CLA?

torchtune/_cli/download.py
apthagowda97: Sure

joecummings (approved):
Can you install and run the linter?

torchtune/_cli/download.py
apthagowda97: This wont work because coz.. `true` and `false` should be of type Boolean.

apthagowda97: But here the they convert to Boolean, I just used there logic  https://github.com/huggingface/huggingface_hub/blob/54e4f76594d2d8338ddccab566a9750cfe4da1bd/src/huggingface_hub/commands/download.py#L137

joecummings: Ahh I missed that in the code, thanks for pointing it out. It's so gross, but seems like the best way to go forward. 

Thanks for all your work on this @apthagowda97, I'll go ahead and merge!

PR #817, opened 2024-04-19
Make memory logging great again

RdoubleA:
Better format the memory stats log output

Ran recipes to confirm logs

image


kartikayk (commented):
Why not just add this to the ```memory_stats_log``` utils so you dont need to copy past this everywhere?

RdoubleA:
@kartikayk Don't we want the recipe to control all logging? there could be use cases where you just want the peak memory stats without logging it

kartikayk:
We already have logging done in many of our utils (eg: seed.py), so I think its ok? Plus this function sets expectation that it'll log. In a previous version of this, I was actually using this same function in setup_model as well to log stats after model initi. So it would be good to just do it in the function otherwise we'd just be copying all of this many time?

ebsmothers (approved):

ebsmothers:
Sorry missed the convo on here before stamping and just looked at the code. I'm fine with moving this to a util in memory.py, the one thing I would ask is that we keep the current version of `memory_stats_log` that returns a dict as it's much more amenable to our metric_logger classes

kartikayk:
> returns a dict as it's much more amenable to our metric_logger classes

What does this mean? :)

PR #820, opened 2024-04-19
keyword-only args for chat dataset, optional chat format

ebsmothers:
Not all datasets need chat format, so we should make it optional. Also we should be using keyword-only args more. In this case the latter is beneficial for the former, so enable it for ChatDataset.

#### Test plan:

```
pytest tests
...
======= 191 passed, 19 skipped, 4 warnings in 60.28s (0:01:00) =========
```

```
...
======= 14 passed, 4 skipped, 192 deselected, 2 warnings in 190.73s (0:03:10) =========
```

Now we can run without applying chat format via e.g. 

```
tune run lora_finetune_single_device --config llama3/8B_lora_single_device dataset=torchtune.datasets.slimorca_dataset dataset.chat_format=null
...
1|32|Loss: 2.4104795455932617:   0%|                                    
```

### DIFF
--- a/torchtune/datasets/_chat.py
+++ b/torchtune/datasets/_chat.py
@@ -4,7 +4,7 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
-from typing import Any, Callable, Dict, List, Mapping, Tuple
+from typing import Any, Callable, Dict, List, Mapping, Optional, Tuple
 
 import numpy as np
 
@@ -50,8 +50,9 @@ class ChatDataset(Dataset):
             (https://huggingface.co/docs/datasets/en/package_reference/loading_methods#datasets.load_dataset.path)
         convert_to_messages (Callable[[Mapping[str, Any]], List[Message]]): function that keys into the desired field in the sample
             and converts to a list of `Messages` that follows the llama format with the expected keys
-        chat_format (ChatFormat): template used to format the chat. If the placeholder variable
+        chat_format (Optional[ChatFormat]): template used to format the chat. If the placeholder variable
             names in the template do not match the column/key names in the dataset, use `column_map` to map them.
+            Default: None
         max_seq_len (int): Maximum number of tokens in the returned input and label token id lists.
         train_on_input (bool): Whether the model is trained on the prompt or not. Default is False.
         **load_dataset_kwargs (Dict[str, Any]): additional keyword arguments to pass to `load_dataset`.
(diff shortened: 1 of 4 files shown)


joecummings (approved):

torchtune/datasets/_chat.py
```diff
@@ -50,19 +50,21 @@ class ChatDataset(Dataset):
             (https://huggingface.co/docs/datasets/en/package_reference/loading_methods#datasets.load_dataset.path)
         convert_to_messages (Callable[[Mapping[str, Any]], List[Message]]): function that keys into the desired field in the sample
             and converts to a list of `Messages` that follows the llama format with the expected keys
-        chat_format (ChatFormat): template used to format the chat. If the placeholder variable
+        chat_format (Optional[ChatFormat]): template used to format the chat. If the placeholder variable
```
RdoubleA, line 53: let me know if this is too much details to add
```suggestion
        chat_format (Optional[ChatFormat]): template used to format the system, user, and assistant messages in the chat. This is used to add structured text around the actual messages, such as the [INST] tags in LLaMA2 and in Mistral. The extra text will still get tokenized as normal text, not as special tokens with separate embedding vectors. In models like LLaMA3 where the tokenizer adds tags as special tokens, `chat_format` is not needed, unless you want to structure messages in a particular way for inference. If the placeholder variable
```

torchtune/datasets/_slimorca.py
```diff
@@ -4,16 +4,20 @@
...
 from torchtune.datasets._chat import ChatDataset
 
 from torchtune.modules.tokenizers import Tokenizer
 
 
 def slimorca_dataset(
+    *,
     tokenizer: Tokenizer,
     source: str = "Open-Orca/SlimOrca-Dedup",
+    chat_format: Optional[ChatFormat] = Llama2ChatFormat,
```
RdoubleA, line 20: should this still be a required parameter? this is not easy to configure from the config (without a nested component)

torchtune/datasets/_chat.py
ebsmothers: It is a bit long, but I think it's useful. Gonna mostly adopt but will leave out the bit about separate embedding vectors

torchtune/datasets/_slimorca.py
ebsmothers: Yeah it's not perfect but see my test command, now I can at least disable to run Llama3 without the chat format from CLI alone

Issue #809, opened 2024-04-19
Seeking guidance on continuing pretraining


calmitchell617:
Hello, thanks very much for the excellent work on this repo.

There are several examples showing how to create a question-response style dataset, but I can't immediately tell how to continue pretraining with, for example, a corpus of unstructured text.

Are there any examples showing how to pack text examples and continue pretraining?

Thank you

RdoubleA:
Hi @calmitchell617, welcome to the repo and thanks for opening this! Sample packing and unstructured datasets for CPT have been on top of my to-do list for sometime. Do you mind sharing some examples of text corpuses you might want to train with?

calmitchell617:
@RdoubleA, thanks for the fast response.

One good example might be [The Stack V1](https://huggingface.co/datasets/bigcode/the-stack), as it is a helpful starting point when training coding assistants. I know a few people personally who would appreciate seeing an example using that dataset in particular.

I would be happy to attempt contributing a PR. I have already looked at the code quite a lot today, and will continue to work on the issue on my own, anyways.

RdoubleA:
Yeah that looks like a great example, and would be good to have that in the repo. Since it's a massive dataset, you would need to add streaming support via `load_dataset(stream=True)`. If you're interested in opening up a PR, I'm more than happy to take a look at it and work with you on it.

calmitchell617:
Great, I will make a PR in the next few days.

felipemello1:
@RdoubleA , can we  close this issue?

PR #827, opened 2024-04-21
feat: added packed continued pretraining dataset functionality

calmitchell617:
#### Context
This PR seeks to resolve issue #809 by adding the ability to pack and tokenize input/label pairs from a Huggingface dataset and then continue a model's pretraining phase on that processed dataset. The dataset can be local, or streamed.

This is my first attempt contributing to this repo. I expect a few changes will be requested, but hope you will consider it a good effort!

Ccing @RdoubleA.

One limitation of this new feature is that you can only pack as many examples as will fit in your system's RAM. This does not mean you can't use large datasets - just that you will only be able to use as many examples as will fit in RAM. This is because when creating a dataset class, you need to implement `__get_item__`, which requires an index to get an item. However, you will not know what a given index will return without packing the examples ahead of time.

One other thing to note is the `train_on_input` argument in the `stack_dataset` function, which is only there for compatibility with existing config yaml files. I'm not convinced it's a good practice to include that kind of thing, but I without it, we may have to change many of the existing recipe configs.

#### Changelog
The PR implements a new class, `ConcatDataset`, which was inspired by the [ConcatDataset class in llama-recipes](https://github.com/meta-llama/llama-recipes/blob/79aa70442e97c3127e53c2d22c54438c32adcf5e/src/llama_recipes/data/concatenator.py#L10).

It also implements a new example dataset, `stack_dataset`, which makes use of `ConcatDataset` to stream and pack the the [Stack V1 Dedup](https://huggingface.co/datasets/bigcode/the-stack-dedup) dataset. This is a popular (and very large) dataset for training coding assistants.

#### Test plan
The new `ConcatDataset` class and `stack_dataset` function both get tests that attempt to mimic the structure and spirit of existing tests - specifically, I used the tests covering `InstructDataset` and `alpaca_dataset` as a guide.

Also, I tested the code by training on a custom dataset which is designed to increase a model's knowledge of a certain Python library, and after a few thousand optimizer steps, the model had clearly learned about that library.

You can kick off a training run with a command like this:

```
clear && tune run \
    --nproc_per_node=4 \
    full_finetune_distributed \
    --config llama3/8B_full \
    batch_size=1 \
    seed=29 \
    tokenizer.path=<checkpoint_dir> \
    checkpointer.checkpoint_dir=<checkpoint_dir> \
    checkpointer.output_dir=<checkpoint_dir> \
    dataset=torchtune.datasets.stack_dataset \
    gradient_accumulation_steps=5 \
    lr_scheduler.num_warmup_steps=100 \
    epochs=3 \
    dataset.max_rows=200 \
    dataset.streaming=True
    dataset.token='<HF_token_goes_here>'
```

### DIFF
--- a/tests/torchtune/datasets/test_concat_dataset.py
+++ b/tests/torchtune/datasets/test_concat_dataset.py
@@ -1,0 +1,46 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+from unittest import mock
+
+from datasets import Dataset
+from tests.test_utils import DummyTokenizer
+from torchtune.datasets import ConcatDataset
+
+
+class TestInstructDataset:
+    expected_tokenized_prompts = [
+        [0, 4, 2, 1, 7, 4, -1, 0, 1, 9],
+        [5, 2, 6, 4, 3, 8, -1, 0, 4, 3],
+    ]
+
+    def get_samples(self):
+        samples_list = [
+            "This is a packing test",
+            "A fantastic test. It should pack two samples.",
+            "This one will not be fully packed.",
+        ]
+
+        samples_dict = {"content": samples_list}
+
+        return Dataset.from_dict(samples_dict)
+
+    @mock.patch("torchtune.datasets._concat.load_dataset")
+    def test_get_item(self, mock_load_dataset):
+        mock_load_dataset.return_value = self.get_samples()
+        dataset = ConcatDataset(
+            tokenizer=DummyTokenizer(),
+            source="were/going/jellyfishing",
+            text_column="content",
+            max_seq_len=10,
+        )
+        assert len(dataset) == 2
+        mock_load_dataset.assert_called_once()
+
+        for i in range(len(dataset)):
+            prompt, label = dataset[i]
+            assert prompt == self.expected_tokenized_prompts[i]
+            assert label == self.expected_tokenized_prompts[i]
--- a/tests/torchtune/datasets/test_stack_dataset.py
+++ b/tests/torchtune/datasets/test_stack_dataset.py
@@ -1,0 +1,57 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+from unittest.mock import patch
+
+import pytest
+from datasets import Dataset
+
+from tests.test_utils import get_assets_path
+from torchtune.datasets import stack_dataset
+from torchtune.modules.tokenizers import SentencePieceTokenizer
+
+
+class TestAlpacaDataset:
+    @pytest.fixture
+    def tokenizer(self):
+        # m.model is a pretrained Sentencepiece model using the following command:
+        # spm.SentencePieceTrainer.train('--input=<TRAIN_FILE> --model_prefix=m --vocab_size=2000')
+        return SentencePieceTokenizer(str(get_assets_path() / "m.model"))
+
+    def get_samples(self):
+        samples_list = [
+            "This is a packing test",
+            "A fantastic test. It should pack two samples.",
+            "This one will not be fully packed.",
+        ]
+
+        samples_dict = {"content": samples_list}
+
+        return Dataset.from_dict(samples_dict)
+
+    @patch("torchtune.datasets._concat.load_dataset")
+    def test_label(self, load_dataset, tokenizer):
+        """
+        Test whether the input and the labels are correctly created.
+        """
+
+        # mock the call to HF datasets
+        load_dataset.return_value = self.get_samples()
+
+        stack_ds = stack_dataset(
+            tokenizer=tokenizer,
+            max_seq_len=10,
+            train_split_name=None,
+        )
+
+        for i in range(len(stack_ds)):
+            inputs, label = stack_ds[i]
+            assert len(inputs) == 10
+            assert len(inputs) == len(label)
+            assert inputs == label
+
+            if i == 0:
+                assert inputs[0] == tokenizer.bos_id
--- a/torchtune/datasets/_concat.py
+++ b/torchtune/datasets/_concat.py
@@ -1,0 +1,134 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+from typing import Any, Dict
+
+from datasets import load_dataset, load_from_disk
+from torch.utils.data import Dataset
+from torchtune.modules.tokenizers import Tokenizer
+from tqdm import tqdm
+
+
+class ConcatDataset(Dataset):
+    """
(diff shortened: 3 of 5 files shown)


torchtune/datasets/_concat.py
```diff
@@ -0,0 +1,134 @@
...
+
+        assert tokenizer is not None, "tokenizer must be provided"
+        assert source is not None, "source must be provided"
+        assert text_column is not None, "text_column must be provided"
+        assert max_seq_len is not None, "max_seq_len must be provided"
+
+        if is_local:
+            dataset = load_from_disk(source, **load_dataset_kwargs)
+        else:
+            dataset = load_dataset(source, **load_dataset_kwargs)
```
RdoubleA, line 72: can we just use `load_dataset`'s ability to load from a local file, or is there some missing functionality?

torchtune/datasets/_concat.py
```diff
@@ -0,0 +1,134 @@
...
+
+        if train_split_name is not None:
```
RdoubleA, line 74: this can also be passed into load_dataset

torchtune/datasets/_concat.py
```diff
@@ -0,0 +1,134 @@
...
+                self.samples.append(buffer[:max_seq_len])
+                buffer = buffer[max_seq_len:]
+
+                if max_rows is not None:
+                    pbar.update(1)
+
+            if max_rows is None:
+                pbar.update(1)
+
+    def __getitem__(self, idx):
```
RdoubleA, line 130: ideally we shouldn't have to use the `__getitem__` method, for streaming/packing an Iterable dataset would be more suitable than our current map-style dataset. Do you think that would make the implementation easier?

RdoubleA (commented):
Thanks for this awesome PR, sample packing has been a high priority TODO on my list for some time and I appreciate you taking an initial pass at this! I have a couple high level questions:
- How do you plan to extend packing functionality to all our other datasets (InstructDataset and ChatDataset)? Is there a way can keep the packing logic isolated in a way where we can just turn it on with a flag in any of our datasets?
- There's two significant features here - streaming and packing. What is most important for what you are trying to accomplish? Based on The Stack dataset you are trying to use, streaming is mandatory because of the size of the data. I ask because I think we should try to tackle both of these features separately. I imagine streaming would be best served by an iterable dataset, which we don't currently have. Or do you just need something like a unstructured text corpus dataset? What would you like to see added to get you unblocked :)

tests/torchtune/datasets/test_stack_dataset.py
```diff
@@ -0,0 +1,57 @@
...
+
+import pytest
+from datasets import Dataset
+
+from tests.test_utils import get_assets_path
+from torchtune.datasets import stack_dataset
+from torchtune.modules.tokenizers import SentencePieceTokenizer
+
+
+class TestAlpacaDataset:
```
SLR722, line 17:
```suggestion
class TestStackDataset:
```

tests/torchtune/datasets/test_concat_dataset.py
```diff
@@ -0,0 +1,46 @@
...
+# LICENSE file in the root directory of this source tree.
+
+from unittest import mock
+
+from datasets import Dataset
+from tests.test_utils import DummyTokenizer
+from torchtune.datasets import ConcatDataset
+
+
+class TestInstructDataset:
```
SLR722, line 14:
```suggestion
class TestConcatDataset:
```

calmitchell617:
@SLR722 thank you for the basic copy/paste error finds.

@RdoubleA, I'll address your questions and comments one by one:

> How do you plan to extend packing functionality to all our other datasets (InstructDataset and ChatDataset)? Is there a way can keep the packing logic isolated in a way where we can just turn it on with a flag in any of our datasets?

It's almost certainly possible to make the packing feature available to other dataset types. I will investigate what is required to do that.

> There's two significant features here - streaming and packing. What is most important for what you are trying to accomplish?

Packing a customized dataset is the more important than streaming for my use case.

> I imagine streaming would be best served by an iterable dataset, which we don't currently have.

I have not implemented an iterable dataset before, but maybe using one would alleviate the issue of having to fit the entire dataset in CPU RAM. I will look into that.

> I ask because I think we should try to tackle both of these features separately.

Yes, that may be a good idea. It will be trivial to switch to a smaller example dataset to finish developing the packing functionality, then we can address streaming a larger dataset afterwards.

> can we just use load_dataset's ability to load from a local file, or is there some missing functionality?

`load_dataset()` does not allow you to load a customized dataset from your local disk. This is a very important part of many people's workflow. Here is a minimal reproducible example showing what happens when you try to load a local dataset, saved with `save_to_disk()`, with `load_dataset()`:

```
from datasets import load_dataset, load_from_disk

dataset_id = 'hails/mmlu_no_train'
subset = 'astronomy'
local_path = 'my_dataset'

# Load the dataset
dataset = load_dataset(
    dataset_id,
    subset,
    trust_remote_code=True, # you can comment this out, but it will raise a warning
    )['dev']
print(f'Length of dataset before customizing: {len(dataset)}')

# Customize our dataset (select the first 3 rows)
dataset = dataset.select(range(3))
print(f'Length of dataset after customizing: {len(dataset)}')

# Save our customized dataset to disk
dataset.save_to_disk(local_path)

# Load our customized dataset from disk, will work
dataset = load_from_disk(local_path)
print(dataset[0])

# Attempt to load local dataset with load_dataset (will fail)
try:
    dataset = load_dataset(local_path)
except Exception as e:
    print(f"Got error: {e}")
```

calmitchell617:
@ebsmothers yes, it is most likely an issue of how I'm saving the dataset, but I believe the [`save_to_disk()`](https://huggingface.co/docs/datasets/en/package_reference/main_classes#datasets.Dataset.save_to_disk) function is quite popular in the HF ecosystem.

Whenever I try to load a dataset saved with `save_to_disk` with `load_dataset`, I get an error: 

> You are trying to load a dataset that was saved using `save_to_disk`. Please use `load_from_disk` instead.

You will see that error if you run the script pasted above.

calmitchell617:
@ebsmothers @RdoubleA @kartikayk, thank you all very much for taking my use case and problems into account! I really appreciate it.

Since you are all asking my opinion on various use cases and functions used in various, fragmented places, I would like to provide one central comment listing the functions that, in my experience, are the most commonly used entrypoints into the HF ecosystem:

### Datasets
The most common way to load a dataset is [`load_dataset`](https://huggingface.co/docs/datasets/v2.19.0/en/package_reference/loading_methods#datasets.load_dataset), followed by [`load_from_disk`](https://huggingface.co/docs/datasets/en/package_reference/main_classes#datasets.Dataset.load_from_disk). `load_from_disk` if used if you downloaded a dataset, did some customization, then saved it with [`save_to_disk`](https://huggingface.co/docs/datasets/en/package_reference/main_classes#datasets.Dataset.save_to_disk).

As far as I know, you cannot load a dataset saved with `save_to_disk` with `load_dataset` - you must use `load_from_disk`. See comment above.

### Models
The most common way I've seen to load a model is [`from_pretrained`](https://huggingface.co/docs/transformers/main_classes/model#transformers.PreTrainedModel.from_pretrained). Lots of people will be looking for the models they train with torchtune to work OOTB with that function. I [outlined my methodology for converting a model output by torchtune to one that can be loaded with that function here](https://github.com/pytorch/torchtune/issues/832#issuecomment-2072823848).

calmitchell617:
Closing this PR in favor of #875

PR #835, opened 2024-04-22
Add support for optional chat format to chat dataset builder

ebsmothers:
Made chat format optional in the `ChatDataset` class but still need to expose the same functionality in the builder

Update docs too per @joecummings request:

Screenshot 2024-04-22 at 10 14 57 AM

### DIFF
--- a/torchtune/datasets/_chat.py
+++ b/torchtune/datasets/_chat.py
@@ -87,7 +87,7 @@ class ChatDataset(Dataset):
 
     def _prepare_sample(self, sample: Mapping[str, Any]) -> Tuple[List[int], List[int]]:
         messages = self._convert_to_messages(sample, self.train_on_input)
-        if self.chat_format:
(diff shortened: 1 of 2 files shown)


RdoubleA (approved):

torchtune/datasets/_chat.py
```diff
@@ -141,7 +142,7 @@ def chat_dataset(
         tokenizer=tokenizer,
         source=source,
         convert_to_messages=convert_to_messages,
-        chat_format=_get_chat_format(chat_format),
+        chat_format=_get_chat_format(chat_format) if chat_format else None,
```
joecummings, line 145: This check and the check in the class are technically incorrect.

`if arg` checks for a boolean or a co-erced empty list. 

What you want is to check `if arg is None`

joecummings (approved):
Evan 4 president

Issue #812, opened 2024-04-19
[RFC] Proximal Policy Optimisation


SalmanMohammadi:
# Implementing Proximal Policy Optimisation

I've used some of the [PyTorch RFC](https://github.com/pytorch/rfcs/blob/master/README.md) template here for clarity.

**Authors:**
* @salmanmohammadi

## **Summary**
I'd like to add support for fine-tuning models using the [Proximal Policy Optimisation](https://arxiv.org/pdf/1506.02438.pdf) (PPO) reinforcement learning (RL) algorithm. Similar to Direct Policy Optimisation, PPO is a core component in Reinforcement Learning from Human Feedback (RLHF) for aligning language models. 

PPO optimises a language model which acts as a policy with an action space equal to the model's vocabulary, and where the observation space is the distribution over all possible prompts, and the reward is some scalar value indicating the "preference" of the model's completion for a given prompt (the reward is usually given by a reward model calibrated for human preferences).

## **Motivation**
This repository helps make a fascinating technology even more accessible. Supporting PPO will help users to understand and explore LLM alignment techniques in native PyTorch, which is already widely adopted and easy to get started with.

## **Proposed Implementation**

- The algorithm itself could be implemented as a recipe in `recipes/`. 
- I don't think datasets need to be in a specific format when using a reward model, so existing dataset functionality can be used.
- Integration of reward models into the codebase: Would this require reward model implementation, if the authors of this repo would like all models to be in native PyTorch? In practice, reward models are (sometimes smaller parameter) copies of the model being fine tuned, so it could be as simple as inheriting from current model implementations and adapting the last layer to output a scalar reward.

## **Prior art**
TRL implements a generalised [PPO trainer](https://github.com/huggingface/trl/blob/main/trl/trainer/ppo_trainer.py). A policy is defined using a thin wrapper around a pre-trained LLM and adds a value function head to be optimised during PPO training. A copy of the model being trained is also initialised and frozen as a reference model.

Feedback and thoughts are very much appreciated. I'm hoping to add value here and I'm grateful for any guidance to help me do so.

kartikayk:
@SalmanMohammadi thanks so much for the high quality RFC. PPO would be an amazing technique to add to torchtune!

Overall the plan looks good. A quick comment on the model itself:

>  Integration of reward models into the codebase

The description here would lend very well to the concepts we have in torchtune. 
[...]

PR #840, opened 2024-04-22
Transformer classifier

SalmanMohammadi:
Context/Changelog

See https://github.com/pytorch/torchtune/issues/812

This PR adds a TransformerClassifier layer which extends the TransformerDecoder functionality to classification tasks. Exemplar component and model builders have been implemented for the base mistral model.

Test plan

Testing this was tricky as there is currently no reference implementation for base mistral to test against. I performed some numerical testing against HuggingFace MistralModel and AutoModelForSequenceClassification with mistralai/Mistral-7B-v0.1, with the same parameters found in [test_llama3.py](https://github.com/pytorch/torchtune/blob/main/tests/torchtune/models/test_llama3.py). However, neither the classification model outputs, nor the base MistralModel and torchtune.models.mistral.mistral produced similar outputs to the HF models.
I've left a sort-of dummy test in which just asserts the output shapes are correct. I can probably test the sequence pooling and classification independently once we agree on how they integrate into the codebase.

Questions/Next steps

This is part of a broader plan to implement RLHF in Torchtune. The TransformerClassifier can hopefully be used against any sequence model we have. @kartikayk - we could implement a recipe for training a mistral reward model using this. I can start implementing a PPO recipe using this reward model, too.


RdoubleA:
Thanks @SalmanMohammadi for this PR! This looks mostly reasonable but I have a few high level suggestions.

The new functionality you're adding is mainly using the output layer of `TransformerDecoder` to project to number of classes, and then extracting the last predicted token. Why not just make a separate classifier head class that consists of this linear projection and post-processing step in the forward? Then you can compose these together directly in the `TransformerDecoder` constructor in `mistral_classifier`:
```
TransformerDecoder(
    tok_embeddings=tok_embeddings,
    layer=layer,
    num_layers=num_layers,
    max_seq_len=max_seq_len,
    num_heads=num_heads,
    head_dim=head_dim,
    norm=norm,
    output=ClassificationHead(embed_dim, num_classes),
)
```
This still keeps the flexibility of swapping out the classification head with some other head, and you keep all the added logic contained in a new class. Your testing will be simpler too because you only need to test the head and not the entire transformer classifier.

Also, please make sure you've run the linters with `pre-commit run --all-files` :)

SalmanMohammadi:
Thanks @RdoubleA :) I've fixed the linting. 

I tried setting it up this way, but in the `forward` for `TransformerClassifier` I also use the input token ids to grab the last non-padding token in each sequence of the batch. I thought I'd have to modify the function call in `TransformerDecoder`'s `forward`:
```
        h = self.norm(h)

        output = self.output(h).float() # pass input tokens in here
        return output
```
to achieve this - so I wrapped around `TransformerDecoder` instead. 

I agree your suggestion is cleaner. Do you have any thoughts on how I'd still be able to use the input `tokens` in `output`?

torchtune/modules/transformer.py
```diff
@@ -206,6 +206,92 @@ def forward(self, tokens: Tensor, input_pos: Optional[Tensor] = None) -> Tensor:
         # shape: [b, s, d]
         h = self.norm(h)
 
-        # shape: [b, s, v]
         output = self.output(h).float()
         return output
+
+
+class TransformerClassifier(nn.Module):
```
ebsmothers, line 213: OK so if we are writing a standalone module I think this is the way to do it. The one question I would ask here is if we need a standalone module at all. To explain my thought process here a bit:

The only real difference between this class and an appropriately-parametrized builder of `TransformerDecoder` is the slicing you do to get just the EOS embedding (i.e. what you currently have in L282-296). But this is basically just "find the last non-padded index in the tokens and slice the logits based on that". In that case it may actually be conceptually simpler to do this in the training script. There you already know the padding index and it is not so different than our existing logic for e.g. ignore_index in our cross-entropy loss calculation. This way you have to pass the padding_index down into the model (in this case it is hardcoded, but in general I imagine we'd want it to be parametrized). If we handle this in the recipe, then you don't need to worry about this. You can also see in our PPO recipe where we do something similar in a [utility method](https://github.com/pytorch/torchtune/blob/bec7babec9c924a0ee7ad27e3f6582bc5bd1fef5/recipes/lora_dpo_single_device.py#L434-L440) with the cross entropy ignore index.

So I think the question comes down to what the full training script will look like. If I understand correctly, the outputs of this model will be used for training the reward model on a paired dataset, is that right? Do you think slicing to the EOS token makes more sense in the reward model training loop or in the model implementation? Admittedly it's kind of a philosophical question without a correct answer, but I bias towards doing it in the recipe because pad ID is a tokenizer property and the model should not have to be aware of it. But please let me know your thoughts here!

tests/torchtune/models/mistral/test_mistral_classifier.py
```diff
@@ -0,0 +1,51 @@
...
+NUM_CLASSES = 10
+INTERMEDIATE_DIM = 512
+
+
+@pytest.fixture(autouse=True)
+def random():
+    set_seed(16)
+
+
+class TestMistralClassifier:
```
ebsmothers, line 30: This test is looking great!

SalmanMohammadi:
Note, the type hints are wrong for the component and model builders. I think you're right that my pre-commits still aren't working right. I'll fix those and address any comments in a bit.

torchtune/modules/transformer.py
SalmanMohammadi: Thanks for raising this. I think you make a good point - you could easily do this directly in the recipe. I think my main consideration would be (also philosophical) how explicit we want to be about that step. Generally speaking, any Seq2Label model will follow the step of classifying the last non-padding token, and produce a `[b, n_classes]` Tensor. Without the additional step of sequence pooling, a `mistral_classifier()` will return a `[b, s, n_classes]` Tensor.

You're right about what the training script looks like. Each recipe will need to replicate this step if we didn't use a dedicated module. If I was relatively new to all this, with good documentation I could understand by reading the recipe that you need to do some kind of pooling to convert `[b, s, n_classes]`, to `[b, n_classes]`, though it may not be clear that this is a common step that will occur after _any_ sequence classification recipe.

Looking forward, we'll want to pull the token padding ID directly from the tokenizer. We have a reference to this in the recipe. I think this is the most significant reason for me to agree that we should lose the separate module and just construct the classifier in the component builder, if we're okay with a `_classifier` model not _technically_ doing the complete classification (sans activation function).

Let me know your thoughts : )

ebsmothers: OK actually I am gonna raise an entirely orthogonal point here (sorry 😅). But during your description of dimensions it dawned on me that we are not necessarily doing things in the most efficient way here. We are applying the projection to the full sequence, when in reality we only need to do the projection on the EOS position. So rather than project and then slice we could slice and then project. This would decrease the size of the intermediate by a factor of `s`. I think the usefulness of this depends on how large the overall tensor is, if e.g. n_classes is really small then the intermediate may be small relative to other tensors in the model and the memory savings would not be as substantial. It also makes fitting into the existing class even thornier, since we would then be adding another operation before `TransformerDecoder`'s output projection.

Assuming that we don't follow the new wrench I threw in things, I think I am OK with either of the two approaches. Ultimately we won't know if we bump up into a wall until we integrate into a recipe, so feel free to make a call; then as we get closer to the actual training loop we can see how it works. How does that sound?

ebsmothers: Completely agree with all of this. Let’s not hack in changes for the memory savings now, I feel it makes most sense to start from a place of clean code and then see whether optimizations can/should be made after that. The utility method sounds good too, will take another pass once you update. Thanks for the patience on this discussion!

SalmanMohammadi:
I've updated my PR @ebsmothers with the changes we discussed :)

torchtune/utils/pooling.py
```diff
@@ -0,0 +1,49 @@
...
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+import torch
+from torch import Tensor
+
+
+def pool_sequence_logits(
+    tokens: Tensor, logits: Tensor, padding_token_idx: int = None
```
ebsmothers, line 12: nit: type padding_token_idx as `Optional[int]`

tests/torchtune/models/mistral/test_mistral_classifier.py
```diff
@@ -0,0 +1,51 @@
...
+            embed_dim=EMBED_DIM,
+            intermediate_dim=INTERMEDIATE_DIM,
+            max_seq_len=MAX_SEQ_LEN,
+        )
+        fixed_init_model(model, min_val=-0.25, max_val=0.5)
+        actual = model(inputs)
+        # TODO find the actual value
+        # expected = torch.tensor(3.9763)
+        assert actual.shape == (BSZ, SEQ_LEN, NUM_CLASSES)
+        # torch.testing.assert_close(actual.mean(), expected, atol=1e-4, rtol=1e-4)
```
ebsmothers, line 51: Need to update this prior to merging. Let me know if you need any assistance with the parity checks here

torchtune/models/mistral/_component_builders.py
```diff
@@ -404,3 +404,212 @@ def lora_mistral_mlp(
         down_proj=down_proj,
         up_proj=up_proj,
     )
+
+
+def mistral_classifier(
```
ebsmothers, line 409: Another path we could consider here is providing an optional output dim argument in the existing Mistral builders and if it's not passed just set to vocab_size. I hope it's not too unintuitive, but could help get around your previous point regarding "classifier" now being a bit of a misnomer.

torchtune/models/mistral/_model_builders.py
```diff
@@ -94,10 +99,92 @@ def lora_mistral_7b(
...
 qlora_mistral_7b.__doc__ = """
 Builder for creating a Mistral model with QLoRA enabled. Base model weights in linear layers
 that LoRA is applied to are quantized per the QLoRA paper: https://arxiv.org/abs/2305.14314.
 Please see `lora_mistral_7b` for full API arguments.
 """
+
+
+def mistral_classifier_7b() -> TransformerDecoder:
+    """
+    Builder for creating a Mistral 7B lassifier model initialized w/ the default 7b
```
ebsmothers, line 114: super nit
```suggestion
    Builder for creating a Mistral 7B classifier model initialized w/ the default 7b
```

torchtune/models/mistral/_component_builders.py
```diff
@@ -404,3 +404,212 @@ def lora_mistral_mlp(
...
+    )
+
+    tok_embeddings = nn.Embedding(vocab_size, embed_dim)
+
+    # TODO: quantize_base is not applied to final output_proj currently.
+    output_proj = (
+        LoRALinear(embed_dim, num_classes, rank=lora_rank, alpha=lora_alpha)
+        if apply_lora_to_output
+        else nn.Linear(embed_dim, num_classes, bias=False)
+    )
```
ebsmothers, line 590: Is it standard to apply LoRA to the `nn.Linear(embed_dim, num_classes)` in PPO or no? (No code changes needed here, I am mainly just curious)

torchtune/utils/pooling.py
```diff
@@ -0,0 +1,49 @@
...
+    Raises:
+        ValueError: If padding_token_idx is not provided and batch size is greater than 1.
+    """
+    batch_size = tokens.shape[0]
+
+    # padding token needs to be defined to pool multiple sequences
+    if padding_token_idx is None and batch_size != 1:
+        raise ValueError("padding_token_idx must be set if batch_size > 1")
+    if padding_token_idx is None:
+        sequence_lengths = -1
```
ebsmothers, line 37: This part I don't fully understand. Is the assumption here that for a batch size of 1 there will be no padding? This may hold in the current state of the world since we pad to the max length in the batch, but for general padding schemes (e.g. pad to fixed length) this may not hold. Personally I would just err on the side of caution and make padding_token_idx a required arg, then you don't have to branch based on batch_size (unless there's an important use case you're solving for here).

tests/torchtune/utils/test_pooling.py
```diff
@@ -0,0 +1,70 @@
...
+                [[0.3, 1.5, 1.6], [0.1, 1.8, 0.2], [1.1, 1.3, 1.4], [0.5, 1.7, 0.1]],
+                [[0.4, 1.6, 1.7], [0.8, 0.9, 1.0], [1.2, 1.4, 1.5], [0.6, 1.8, 0.2]],
+            ]
+        )
+        expected_output = torch.tensor(
+            [
+                [1.3, 0.5, 1.6],
+                [1.0, 1.2, 1.3],
+                [0.3, 1.5, 1.6],
+                [0.4, 1.6, 1.7],
```
ebsmothers, line 32: I'm curious about this case: if all tokens are padding we return the first element? Any particular reason for this? (Hopefully something that won't happen very often in practice anyways)

ebsmothers (commented):
A few more comments but overall this is looking pretty good! Special thanks for the great work hitting a high bar on testing and documentation here.

tests/torchtune/models/mistral/test_mistral_classifier.py
SalmanMohammadi: I could use some guidance here :) Could you point to an existing test in the repo you think is a good example? I'm still working on the base `mistral` tests in another branch.

torchtune/models/mistral/_component_builders.py
SalmanMohammadi: I see what you're meaning. My only point would be that the implementation I have now minimises changes to the base mistral models. I think keeping it modular for now while we're still experimenting with how this model is going to be used downstream would be easiest, and we can try out potentially cleaner paths once things are more concrete.

torchtune/models/mistral/_component_builders.py
SalmanMohammadi: So the `mistral_classifier`'s role in PPO is just to provide preference outputs during optimisation of the model being aligned. Infact, [The `trl` implementation of RLHF using PPO is agnostic to where the preference label comes from](https://github.com/huggingface/trl/blob/main/trl/trainer/ppo_trainer.py). In the context of RLHF using PPO, the reward model is trained separately, and PPO shouldn't really care how it was trained.

torchtune/utils/pooling.py
SalmanMohammadi: The assumption is the other way round - if there's no padding, there _must be a batch size of 1_. Otherwise, you wouldn't have a way to grab the last token from multiple sequences in a batch with variable lengths.

If you do have a batch size of 1, the function is agnostic to this, it'll just give you the last token if you don't pass in a padding token, but it'll give you the last non-padding token otherwsie (the default behaviour). This is [how HF implements it](https://github.com/huggingface/transformers/blob/928331381ef6ce0622c0b1ac704299046b3afa21/src/transformers/models/mistral/modeling_mistral.py#L1339), I'll attribute it in the function for clarity. My intuition here is that the use case is for inference, potentially using a tokenizer where the padding token isn't defined, or you just generally know you have a full sequence/remove padding on your own side. 

Removing the branch for batch size wouldn't actually change how the sequence(s) are pooled, but just prevents the failure case. Allowing the function to be used without a padding token just gives a little flexibility.

Let me know if this makes sense. Open to your thoughts :)

tests/torchtune/utils/test_pooling.py
SalmanMohammadi: Yeah, this wasn't something I was super sure about. Agreed it'll be a weird edge case in practice. I think this is the implementation that makes the most sense to me - you always want the earliest token in the sequence pre-padding. If it's all padding, it feels consistent to just return the first token.

torchtune/models/mistral/_component_builders.py
ebsmothers: Yep this sounds good to me!

tests/torchtune/utils/test_pooling.py
ebsmothers: Yeah it's true, the expected behavior is not really well-defined in this case. We definitely want to return something (don't want to get into the business of throwing out samples in the pooler), so I think your approach to return the first token makes sense.

tests/torchtune/models/mistral/test_mistral_classifier.py
ebsmothers: (Note: this is a long comment, but for the actual practical suggestion for this particular PR just skip to the end)

The parity checks can be tricky, it also depends on the model and how readily available a solid reference implementation is. I will use our [test_lora.py](https://github.com/pytorch/torchtune/blob/main/tests/torchtune/modules/peft/test_lora.py) as a simple example here. We typically:

1) Find a reference implementation, and either copy-paste it or import it into a runnable Python script. We have a bunch of these under tests/torchtune/models/llama2/scripts (admittedly not the best location, but they are not really used and more meant as references). Here's the [copy-pasted LoRA layer](https://github.com/pytorch/torchtune/blob/main/tests/torchtune/models/llama2/scripts/compare_lora.py#L15-L111) from the original Microsoft repo.
2) In the same script, import our implementation. Then 
  a) set the random seed for reproducibility, create some dummy input tensor for the model, and initialize the weights of our implementation in a deterministic way. This is done [here](https://github.com/pytorch/torchtune/blob/main/tests/torchtune/models/llama2/scripts/compare_lora.py#L127-L141) using our test utility [fixed_init_model](https://github.com/pytorch/torchtune/blob/ea3d4eaca82d17fc83a979f81479c69c79dc9b92/tests/test_utils.py#L97)
  b) Load the weights we initialized for our model into the reference implementation. This usually involves remapping of state dicts (sometimes just a key mapping, sometimes some tensor reshaping). This part is usually fairly annoying, a bit of guess-and-check, and can actually get quite hairy for full models with nontrivial differences in how parameters are represented. [Example for the LoRA comparison](https://github.com/pytorch/torchtune/blob/main/tests/torchtune/models/llama2/scripts/compare_lora.py#L155-L162)
  c) Finally call forward on both models and check that the outputs line up. Then take some summary stat (like mean) to use as the expected value in the test. [LoRA example](https://github.com/pytorch/torchtune/blob/main/tests/torchtune/models/llama2/scripts/compare_lora.py#L166-L167)
3) Then in the unit test we use the value from (2) as our expected value ([in the LoRA test](https://github.com/pytorch/torchtune/blob/main/tests/torchtune/modules/peft/test_lora.py#L24)). We [set the same seed](https://github.com/pytorch/torchtune/blob/main/tests/torchtune/modules/peft/test_lora.py#L29) and do the [same initialization](https://github.com/pytorch/torchtune/blob/main/tests/torchtune/modules/peft/test_lora.py#L60) so that we get the same result from the comparison script.


If you've made it this far, you can see it's actually a fair amount of work to do things this way. We try to hit a really high bar on testing for correctness and have found this is the best way to do it. That being said, going through all these steps is definitely not a prerequisite for landing this change. We still need a proper comparison script for Mistral (though it was tested locally prior to landing) and shouldn't block your PR on that. 

Instead for testing this change, I would propose: assume the existing Mistral implementation in the repo is correct. Then the delta for your changes in terms of the Mistral builder is really just the change in output layer. Then the much-abbreviated version of the above is: (a) import an existing Mistral builder and hardcode the output projection to `nn.Linear(embed_dim, n_classes)`. Do the same initialization and run forward on that model to get the expected value. Then use this value in your unit test (no state dict remapping needed).

We can tackle the full testing for Mistral I outlined above in a separate PR, but definitely don't want to block this particular change on that. Let me know if this makes sense to you, and if you are still interested in adding the Mistral test following this approach in a separate PR I am happy to provide (even) more detailed guidance when needed.

torchtune/models/mistral/_component_builders.py
ebsmothers: Ah OK this helps my understanding a lot, thanks for the clarification!

torchtune/utils/pooling.py
ebsmothers: OK this generally makes sense. My main question is whether we will actually ever hit this function during inference; if its purpose is just to extract the EOS from an existing sequence it seems like we shouldn't, right? And yes attribution to the HF implementation would be great here

tests/torchtune/models/mistral/test_mistral_classifier.py
SalmanMohammadi: Thanks so much for this. I think some of it I grokked from the code but 2.b would've been tricky and sounds like it can be a pain in the ass. I've started this process with just the attention layer for mistral so far, I'll put a draft PR up to get your thoughts early on in this soon.

I think your suggestion for testing the classifier sounds sensible. It could be a pretty general test to make sure `output_layer` does what we want it to do.

torchtune/utils/pooling.py
SalmanMohammadi: Yeah, I thought about this for a bit and I think I'm with you on making it required. I think having to handle this error wherever its used is more annoying than just supporting a very nice usecase. 

> if its purpose is just to extract the EOS from an existing sequence

My usecase for performing inference was: run your trained `_classifier` model to obtain `[b, s, n]`, then having to call `pool_sequence_logits` to get `[b, n]`.

tests/torchtune/models/mistral/test_mistral_classifier.py
SalmanMohammadi: I've taken a go at writing a test following `test_lora.py`. Thoughts appreciated.

tests/torchtune/models/mistral/scripts/compare_mistral_classifier.py
```diff
@@ -0,0 +1,188 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
```
ebsmothers, line 1: Thanks so much for adding this! You really went above and beyond here. Actually one more change request here (this one should be much easier, I promise). Can you add back the previous unit test you deleted in the latest commit? I think my previous wall of text could've been a bit clearer, but the main point I was getting at was that the values from this script can then be used to set the expected values in the unit test.

The idea being that then we will actually run the unit test in our CI, whereas this script is more of a nice-to-have so we can understand the initial testing that went into setting the unit test values. In case it helps, I ran your script locally myself: [here's the output](https://gist.github.com/ebsmothers/ba220949626911788282f4b338f354a6) I got so that you don't have to run it again. And [here](https://gist.github.com/ebsmothers/031a88763285454d342cf197d4dad900) is roughly what I think the unit test will look like now (hopefully not too many bugs in there 😅). Again, sorry for the back and forth on this!

torchtune/modules/transformer.py
```diff
@@ -111,7 +111,7 @@ class TransformerDecoder(nn.Module):
             to setup the :func:`~torchtune.modules.KVCache`
         norm (nn.Module): Callable that applies normalization to the output of the decoder,
             before final MLP.
-        output (nn.Linear): Callable that applies a linear transformation to the output of
+        output (nn.Module): Callable that applies a linear transformation to the output of
```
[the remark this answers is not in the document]
ebsmothers: Bumping this comment, let's switch back to nn.Linear here unless there's a reason not to

tests/torchtune/models/mistral/scripts/compare_mistral_classifier.py
SalmanMohammadi: I get it now :)

ebsmothers:
OK the changes look good, but the unit test is failing on CI. Does it pass for you locally? I can also help a bit here with debugging if you need

SalmanMohammadi:
Yeah it's passing locally. Are there more detailed logs?

ebsmothers:
OK I think I cracked the case here. The lack of detailed logs can often be indication that the runner actually just crashed (can be due to out of memory or something like that). It crashed on the 2nd test case, which is the largest one. You can either reduce the batch size or even just remove that test case, since it's not actually testing fundamentally different logic than the other cases. I tested on my fork with a batch size of 4 and confirmed that the CI succeeds (but tbh I'd just scrap the test case for the reason I mentioned).

SalmanMohammadi:
Thanks so much for your help debugging :)

torchtune/modules/transformer.py
```diff
@@ -129,7 +129,7 @@ def __init__(
         num_heads: int,
         head_dim: int,
         norm: nn.Module,
-        output: nn.Linear,
+        output: nn.Module,
```
ebsmothers, line 132: Can we change this back to nn.Linear? I don't think we need the general nn.Module in your latest version (lmk if I'm missing something here though)

torchtune/modules/transformer.py
```diff
@@ -206,6 +206,5 @@ def forward(self, tokens: Tensor, input_pos: Optional[Tensor] = None) -> Tensor:
         # shape: [b, s, d]
         h = self.norm(h)
 
-        # shape: [b, s, v]
```
ebsmothers, line 209: nit: I would still keep some version of shape comment in just cause it's generally useful for people. could just be `# shape: [b, s, out_dim]` or something like that

SalmanMohammadi:
Good catch!

torchtune/modules/transformer.py
ebsmothers: @SalmanMohammadi sorry one more thing on this.. can you also change the corresponding docstring back to nn.Linear too? Then I promise we are good to merge 😄

SalmanMohammadi: my bad! I had a thought that the pre-commits sometimes catch things like this?

ebsmothers: Yeah I'm also confused why it wasn't caught tbh..

ebsmothers (approved):
Looks good! Thanks for your patience on the review!

SalmanMohammadi:
Thank you for *your* patience!!

PR #841, opened 2024-04-22
Better dataset and data documentation

joecummings:
#### Context
- This addresses some concerns about documentation for our datasets.

#### Changelog
- Expose dataset and data classes
- Update docstrings

#### Test plan
- Eyes

### DIFF
--- a/docs/source/api_ref_data.rst
+++ b/docs/source/api_ref_data.rst
@@ -49,4 +49,3 @@ Helper funcs
     :nosignatures:
 
     validate_messages
-    truncate
--- a/docs/source/api_ref_data.rst
+++ b/docs/source/api_ref_data.rst
@@ -32,6 +32,15 @@ Chat formats
     Llama2ChatFormat
     MistralChatFormat
 
+Types
+~~~~~
+
+.. autosummary::
+    :toctree: generated/
+    :nosignatures:
+
+    Message
+
 Helper funcs
 ------------
 
--- a/docs/source/api_ref_data.rst
+++ b/docs/source/api_ref_data.rst
@@ -33,7 +33,7 @@ Chat formats
     MistralChatFormat
 
 Types
-~~~~~
+-----
 
 .. autosummary::
     :toctree: generated/
@@ -49,3 +49,4 @@ Helper funcs
     :nosignatures:
 
     validate_messages
+    truncate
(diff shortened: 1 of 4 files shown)


torchtune/datasets/_chat.py
```diff
@@ -28,11 +28,12 @@ class ChatDataset(Dataset):
...
+    If the column/key names differ from the expected names in the ``ChatFormat``,
+    then the ``column_map`` argument can be used to provide this mapping.
+
+    Use ``convert_to_messages`` to prepare your dataset into the llama conversation format
+    and roles::
 
-    Use `convert_to_messages` to prepare your dataset into the llama conversation format
-    and roles:
         [
             {
```
RdoubleA, line 38: while you're here, can you also update this format example? It should actually be something like:
```
[
    Message(
        role=<system|user|assistant>,
        content=<message>,
    ),
    ...
]
```

docs/source/api_ref_data.rst
```diff
@@ -14,8 +17,36 @@ torchtune.data
...
     MistralChatFormat
+
+Types
+-----
+
+.. autosummary::
+    :toctree: generated/
+    :nosignatures:
+
+    Message
```
RdoubleA, line 42: I think sphinx breaks for Message because it does not have docstrings

RdoubleA:
I'm betting you'll have to fix a lot of sphinx errors in the docstrings, so I will stamp once that passes

docs/source/api_ref_data.rst
joecummings: Message is fine, it broke for ``truncate``

torchtune/datasets/_chat.py
joecummings: Wait so the user has to create a `Message` class? They can't just have a JSONL file that parses this into a Message format?

RdoubleA: hm yeah, that would be more ideal... maybe we should add this step in the dataset class to parse it into Message format for you if needed

joecummings: Well we *could* enforce it here on the class, but in the builder, it will format for you?

RdoubleA: yeah if you specify a converter... need to think more on this, let's revisit this.

RdoubleA (approved):

torchtune/datasets/_chat.py
joecummings: Definitely, but for now, what should the docstring read?

PR #855, opened 2024-04-24
enable LoRA + FSDP2

weifengpy:
how to run it
* 7B: `tune download meta-llama/Llama-2-7b-hf --output-dir /tmp/Llama-2-7b-hf --hf-token <HF_TOKEN>` && `tune run --nproc_per_node 8 lora_finetune_fsdp2 --config llama2/7B_lora`
* 70B: `tune download meta-llama/Llama-2-70b-hf --output-dir /tmp/Llama-2-70b-hf --hf-token <HF_TOKEN>` && `tune run --nproc_per_node 8 lora_finetune_fsdp2 --config llama2/70B_lora`

recipe tests: `pytest tests/recipes/test_lora_finetune_fsdp2.py -m integration_test`
unit test: `pytest tests/torchtune/utils/test_distributed.py -k test_state_dict`

Highlights
* +12% tokens per second (see wandb snapshot) with +0.8% memory reserved
* 3.2 x speedup in model init: from 250s to 77s

Screenshot 2024-05-16 at 2 09 59 PM

FSDP2 changes
- [x] init models on device=`meta` instead of `cpu`
- [x] activation checkpointing happens before register FSDP hooks

checkpoint changes
- [x] loading: load HF checkpoint with `mmap=True`. convert plain tensor into `DTensor`
- [x] saving:  gather full state dict on rank0 cpu. convert `DTensor` into plain tensor
- [x] resume finetuning from previous checkpoints

optional
- [ ] foreach=True for adam + DTensor through `optimizer.foreach=True` since pytorch PR have not land yet https://github.com/pytorch/pytorch/pull/123394


recipes/lora_finetune_distributed.py
```diff
@@ -451,22 +468,22 @@ def save_checkpoint(
...
         # we consolidate the full model and optim state dicts on CPU for rank 0
-        with FSDP.state_dict_type(
-            self._model,
-            StateDictType.FULL_STATE_DICT,
-            FullStateDictConfig(offload_to_cpu=True, rank0_only=True),
-            FullOptimStateDictConfig(offload_to_cpu=True, rank0_only=True),
-        ):
-            cpu_state_dict = self._model.state_dict()
-            if intermediate_checkpoint:
-                opt_state_dict = FSDP.optim_state_dict(self._model, self._optimizer)
```
weifengpy, line 462: @awgu do we have examples for FSDP2 optim static dict ?

torchtune/modules/position_embeddings.py
```diff
@@ -42,20 +42,21 @@ def __init__(
...
         self.max_seq_len = max_seq_len
-        self._rope_init()
+        theta = self.get_theta()
+        self.register_buffer("theta", theta, persistent=False)
+        self.build_rope_cache(self.max_seq_len)
 
     # We need to explicitly define reset_parameters for FSDP initialization, see
     # https://github.com/pytorch/pytorch/blob/797d4fbdf423dd9320ebe383fb57ffb1135c4a99/torch/distributed/fsdp/_init_utils.py#L885
     def reset_parameters(self):
-        self._rope_init()
```
weifengpy, line 50: prefer `copy_` over registering buffer again since this is resetting

recipes/lora_finetune_distributed.py
```diff
@@ -4,6 +4,7 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
```
weifengpy, line 5: I am updating the file so it's easy to review the UX difference. Do we have preference on adding `if-else` for config `--fsdp2` or adding a new file lora_finetune_distributed_fsdp2.py @karthikprasad  @rohan-varma

recipes/lora_finetune_distributed.py
awgu: FSDP2 optimizer state dict will also give a sharded optimizer state dict (tensors should be `DTensor`s). If we want to save/load full optimizer state dicts, then similar conversion is required as model state dict.

recipes/lora_finetune_distributed.py
```diff
@@ -277,45 +274,98 @@ def _setup_model(
...
 
             # Load both the base model weights and (if available) the adapter weights. Both
             # of this should happen only on Rank 0
-            model.load_state_dict(base_model_state_dict, strict=False)
-            if lora_weights_state_dict:
-                model.load_state_dict(lora_weights_state_dict, strict=False)
-
+            log.info("FSDP is enabled. Loading checkpoints for Rank 0 ...")
+            init_start = time.perf_counter()
+            for param_name, full_param in base_model_state_dict.items():
```
awgu, line 321: In the future, I think refactoring the conversion of CPU full state dict on rank 0 to sharded state dict on all ranks can be moved to a utility. (Having it here initially in the draft PR to show the difference sounds good though.)

recipes/lora_finetune_distributed.py
```diff
@@ -277,45 +274,98 @@ def _setup_model(
...
+        for m in model.modules():
+            if isinstance(m, FSDP):
+                continue
+            param_is_meta = [
+                x.is_meta
+                for x in itertools.chain(
+                    m.parameters(recurse=False), m.buffers(recurse=False)
+                )
+            ]
+            if len(param_is_meta) > 0 and all(param_is_meta):
```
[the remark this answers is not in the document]
awgu: For the LoRA weights, I would expect calling `lora_linear.initialize_parameters()` here. For the buffer, it is more debatable. For torchtitan, we include re-initializing the buffer in `init_weights`:
https://github.com/pytorch/torchtitan/blob/be432e1185f9da24dcebc0b1f6c587cad539da9e/torchtitan/models/llama/model.py#L398-L405

IMHO, this is the current cost of changing from CPU init to meta-device init. We need an initializer for the buffer that we can call after.

torchtune/modules/position_embeddings.py
awgu: Why not make `get_theta()` just register the buffer again?

I did not think that you can `copy_` into meta-device tensor 🤔 curious how this works

awgu (commented):
I think one main question for discussion is what the torchtune folks feel about defining an explicit initializer for the RoPE theta buffer, which would be required to do meta-device init (which should speed up initialization!).

recipes/lora_finetune_distributed.py
kartikayk: I think you tagged Karthik by mistake (sorry for the notifications Kartik!)

Yeh I'd vote for a different recipe with its own config instead of if-else blocks. I think BC is important and we shouldn't move off of FSDP until we fully tested this out. Let me know if this makes sense?

torchtune/modules/peft/lora.py
```diff
@@ -84,6 +84,9 @@ def initialize_parameters(self):
         _lora_a_init_params(self.lora_a)
         _lora_b_init_params(self.lora_b)
 
+    def reset_parameters(self):
```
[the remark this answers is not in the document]
rohan-varma: I thought this is due to FSDP and how FSDP materializes modules on meta device - if there is a `reset_parameters` attribute, it calls it?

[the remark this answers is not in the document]
awgu: For FSDP2, we have a new meta-device initialization flow, and we do not have any explicit depending on `reset_parameters()` anymore.

The difference with FSDP2 is that we can shard on meta-device, so the user just needs to initialize a sharded meta-device module. For example, in torchtitan, we do the following:
```
# Construct the unsharded model on meta
with torch.device("meta"):
    model = Transformer(model_args)

# Shard with FSDP on meta
for layer in model.layers:
    fully_shard(layer, **fsdp_kwargs)
for tensor in itertools.chain(model.parameters(), model.buffers()):
    assert tensor.device == torch.device("meta")

# Allocate the sharded model on GPU
model.to_empty("cuda")

# Initialize parameters/buffers
model.init_weights()
```

The last step of `model.init_weights()` is flexible. The user just needs to make sure that every tensor got initialized somehow. There is no longer of requirement that the user must call `reset_parameters()` or what `reset_parameters()` needs to do.

recipes/lora_finetune_distributed.py
weifengpy: got you. will try to do similar conversion for optimizer state dict

torchtune/modules/position_embeddings.py
weifengpy: > just register the buffer again?

registering again is fine. I called `module.to_empy(self._deivce)` before calling `module.reset_parameters()`. By default `get_theta()`  creates tensors on cpu. It's more like a stylish thing where I followed fsdp unit test

recipes/lora_finetune_distributed.py
weifengpy: good catch. I just corrected my tagging

I am good with creating a new recipe

recipes/lora_finetune_distributed.py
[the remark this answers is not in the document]
weifengpy: do we want to maintain the convention of calling `reset_parameters()` for meta modules? otherwise I am fine with calling `initialize_parameters()`.

recipes/lora_finetune_distributed.py
weifengpy: yep. will move this to utility since the same logic applies to lora state dict

recipes/lora_finetune_distributed.py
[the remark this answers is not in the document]
weifengpy: got you. will call `initialize_parameters` directly

torchtune/modules/position_embeddings.py
awgu: Hmm. I am not sure we need to call `module.to_empty()`. That would create an extra sharded copy of the parameters on GPU when there is also one in the sharded state dict?

awgu: The point of `load_state_dict(assign=True)` is that the `module` can still have meta-device parameters when we are loading, as long as the state dict itself has GPU parameters.

weifengpy: oh, I only call `module.to_empty()` if module is on meta device. but again, it's stylish thing. another way is simply call `_rope_init()` with `self._device` ctx

weifengpy: got you. will call `_rope_init()` instead

torchtune/utils/_distributed.py
```diff
@@ -297,3 +299,31 @@ def lora_wrap_fsdp(module: nn.Module, recurse: bool, **kwargs):
         return isinstance(module, tuple(modules_to_wrap))
 
     return lora_wrap_fsdp
+
+
```
weifengpy, line 303: @awgu for FSDP checkpoint util, should we put it here in torchtune or move it into pytorch? it seems to be needed in general beyond torchtune

awgu: My current personal opinion: We should not land it in `pytorch` repo yet.
1. We do not know the problem space well enough to know if the generality of this utility and how users may want to customize it (e.g. add coalescing of broadcasts, add quantization, etc.).
2. The utility does not have a high degree of complexity. The building block doing the heavy lifting is `DTensor`. This function can be copy-and-pasted and tailored to specific needs.

We can see how the torchtune folks feel about having it in their utils folder for now.

torchtune/utils/_distributed.py
```diff
@@ -297,3 +299,31 @@ def lora_wrap_fsdp(module: nn.Module, recurse: bool, **kwargs):
...
+    is_rank_zero: bool,
+):
+    meta_sharded_sd = model.state_dict()
+    sharded_sd = {}
+    for param_name, full_tensor in full_sd.items():
+        sharded_meta_param = meta_sharded_sd.get(param_name)
+        mesh = sharded_meta_param.device_mesh
+        if is_rank_zero:
+            full_tensor = full_tensor.detach().to(device)
+            torch.distributed.broadcast(full_tensor, src=0, group=mesh.get_group(0))
```
awgu, line 317: I realized that `mesh.get_group(0)` is not right if we want this to generalize to 2D (just for future-proofing).

I might be missing something, but it looks like `DeviceMesh` does not have the API for what we want 🤔 
In the 2D case, we want to get the global PG over both TP and DP ranks, but it looks like `get_group()` would return a list of the DP PG and TP PG, meaning that we would need to make one `broadcast` call per mesh dim.

cc: @wanchaol @wz337 Is there any way to get the single global PG from a (parent) `DeviceMesh`?

recipes/lora_finetune_distributed.py
```diff
@@ -277,86 +272,62 @@ def _setup_model(
...
+                model, lora_weights_state_dict, self._device, self._is_rank_zero
+            )
 
-        if self._dtype == torch.bfloat16:
-            model = model.to(torch.bfloat16)
+        with utils.set_default_dtype(self._dtype), self._device:
+            for m in model.modules():
+                if isinstance(m, LoRALinear) and not lora_weights_state_dict:
+                    # to_empty is needed since kaiming_uniform_ is inplace
+                    m.to_empty(device=self._device)
```
weifengpy, line 313: @awgu I ended up using `m.to_empty` becuase `nn.kaiming_uniform_` needs to apply to a GPU tensor. let me know if this make sense

torchtune/utils/_distributed.py
wanchaol: hmmm why we want the global pg for FSDP broadcasting? I think FSDP broadcast states would only require dp_pg to make sure every rank in the DP group starting off the same

awgu: For this particular case, we are broadcasting the full state dict (source of truth) from rank 0 to all other ranks. We then shard the replicated tensor via `distribute_tensor`.

recipes/lora_finetune_distributed.py
awgu: This makes sense to me!

torchtune/utils/_distributed.py
awgu: The broadcasting is not specific to FSDP but rather to full state dict -> sharded state dict conversion. If we wanted to convert a full state dict on rank 0 to a sharded state dict on all ranks for a pure TP model, we can use this same utility.

wanchaol: Aha I see, so only rank0 got the full_state_dict initially instead of every rank, that's why we need to do a whole world broadcasting?

wanchaol: Ok not yet sure why we don't load full state dict to every rank but if rank0 only, we can just use the `get_default_group` API from c10d to get a world_pg?

Even if DeviceMesh offers such a thing, it would be just calling into this function I feel. Given that we are not sure yet if we want to expose a default_pg API in DeviceMesh (maybe we can have a `get_world_group` API, but I feel it's a bit over kill if c10d already provide such a thing)

awgu: > Ok not yet sure why we don't load full state dict to every rank

1. We do not want to 8x the CPU RAM usage. If we `torch.load(mmap=True)` can allow processes to share the CPU memory, then this is not a problem.
2. We may not want to 8x the H2D copy volume, which may be slower than broadcast (depends on network bandwidth).
    - Each rank needs to copy each replicated tensor from CPU to GPU. Today, `distribute_tensor()` moves the tensor to device before sharding, but in theory, the sharding could be done on CPU too, meaning that this is not a fundamental problem.

For (1), we can explore if `torch.load(mmap=True)` can allow processes to share the CPU memory, but for (2), we need to benchmark to understand the performance implication for different setups.

We can discuss the option to load full state dict on all ranks more offline. I have some more thoughts.

---

Using the global PG seems okay for now. @weifengpy we can just not pass `group` then, and `dist.broadcast` will use the global PG by default.

wanchaol: I see, yeah this make sense to me!

weifengpy: removed `group` argument from `dist.broadcast` to use global PG

torchtune/utils/_distributed.py
```diff
@@ -297,3 +299,69 @@ def lora_wrap_fsdp(module: nn.Module, recurse: bool, **kwargs):
...
+    cpu_state_dict = {}
+    for param_name, sharded_param in sharded_sd.items():
+        full_param = sharded_param.full_tensor()
+        if is_rank_zero:
+            cpu_state_dict[param_name] = full_param.cpu()
+        else:
+            del full_param
+    return cpu_state_dict
+
+
```
weifengpy, line 344: @awgu I made 2 assumptions about `opt.state_dict()` when converting DTensor into full tensor. Let me know if it is too aggresive
* `sharded_sd["param_groups"]` are the same across ranks. no need to communicate
* `sharded_sd["state"]`: for AdamW, `key=step` and `value` is plain Tensor. `key=exp_avg/exp_avg_sq`, value is `DTensor`. We only need to communicate on DTensor. It might not work for Apex optimizers

awgu: This seems reasonable to me since these assumptions follow the SPMD programming model that `DTensor`/FSDP generally assume. If the `"param_groups"` were different across ranks, I am not sure how to coalesce them.

cc: @fegin @wz337  who have more experience with this

Separately, I think we should discuss with the torchtune folks why they prefer full optimizer state dict. For model, I can buy that we want to export the model state dict for downstream use, but I would be curious if there are cases where we want to export the optimizer state dict and resume training e.g. from a single-GPU setup.

recipes/lora_finetune_distributed.py
```diff
@@ -451,22 +429,21 @@ def save_checkpoint(
...
-                opt_state_dict = None
+
+        cpu_state_dict = utils.get_full_model_state_dict(
+            self._model, self._is_rank_zero
+        )
+
+        if intermediate_checkpoint:
+            opt_state_dict = utils.get_full_optimizer_state_dict(
+                self._optimizer, self._is_rank_zero
+            )
```
weifengpy, line 440: @kartikayk `get_full_optimizer_state_dict` is converting sharded opotimizer state dict to full state dict on rank0. let me know if the UX looks smooth enough

torchtune/utils/_distributed.py
fegin: @weifengpy Why not just use `torch.distributed.checkpoint.state_dict.get_state_dict`? That will save all the implementations here.

recipes/lora_finetune_distributed.py
fegin: I think we should just use `torch.distributed.checkpoint.state_dict` instead of reinventing another API.

weifengpy: talked with @fegin. will update PR soon to use `torch.distributed.checkpoint.state_dict` instead of `get_full_optimizer_state_dict`

weifengpy: @fegin DCP to use `key=fqn` instead of `key=int` in `opt_state_dict['param_groups'][0].keys()`. would `key=fqn` fails `torch.optim.Optimizer.load_state_dict`? Trying to see if implementing rank0 load + boradcast in `torch.distributed.checkpoint.state_dict` become a hard blocker for this PR

weifengpy: @fegin perf-wise, seems DCP is doing N gather if there are N params? `torch/distributed/_state_dict_utils.py`? want to double confirm since we were thinking of bucketing/coalescing to do fewer collectives on larger tensors (not an immeidate priority)

awgu: @weifengpy For your first question, I think there should be a complementary API `set_state_dict` to call to modify the optimizer state dict before `torch.optim.Optimizer.load_state_dict`.

recipes/lora_finetune_distributed.py
```diff
@@ -307,16 +306,36 @@ def _setup_model(
...
 
-        else:
-            # For non-zero ranks, load the model on meta device
-            with utils.set_default_dtype(self._dtype), torch.device("meta"):
-                model = config.instantiate(cfg_model)
+        if enable_activation_checkpointing:
+            utils.set_activation_checkpointing(
+                model, auto_wrap_policy={modules.TransformerDecoderLayer}
+            )
+
```
weifengpy, line 316: `if isinstance(m, modules.TransformerDecoderLayer):` is equivalent of `auto_wrap_policy` in FSDP1

recipes/lora_finetune_distributed.py
```diff
@@ -307,16 +306,36 @@ def _setup_model(
...
+        fully_shard(model)
+
+        utils.load_from_full_model_state_dict(
+            model, base_model_state_dict, self._device, self._is_rank_zero
+        )
+        if lora_weights_state_dict:
+            utils.load_from_full_model_state_dict(
+                model, lora_weights_state_dict, self._device, self._is_rank_zero
+            )
+
```
weifengpy, line 329: pros and cons of meta init. pros is 4.5x speed up during model init and thus shorter TTFB. cons is user need to call `initialize_parameters` on `LoRALinear` explicitly to move them from `meta` to `gpu`

tests/recipes/test_lora_finetune_distributed.py
```diff
@@ -87,6 +92,7 @@ def test_loss(self, tmpdir, monkeypatch):
 
     @pytest.mark.integration_test
     @gpu_test(gpu_count=2)
+    @pytest.mark.skipif(True, reason="resolve FSDP2 optimizer state dict and enable")
```
weifengpy, line 95: will follow up on resuming training. the gap is on FSDP2 + optimizer state dict

weifengpy:
I see meaningful numeric difference between this PR and torchtune trunk. will debug it before publishing

pyproject.toml
```diff
@@ -41,12 +41,13 @@ tune = "torchtune._cli.tune:main"
 dev = [
     "bitsandbytes>=0.43.0",
     "pre-commit",
-    "pytest",
+    "pytest==7.4.0",
```
weifengpy, line 44: `from torch.testing._internal.common_utils import run_tests` has a depdency on `pytest==7.4.0` and `expecttest`, borrowed from pytorch repo

torchtune/utils/_distributed.py
```diff
@@ -8,11 +8,17 @@
...
 import os
 from itertools import chain
-from typing import Callable, Dict, Set, Tuple, Type
+from typing import Any, Callable, Dict, Set, Tuple, Type
 
 import torch
 import torch.distributed as dist
 from torch import nn
+
+try:
```
awgu, line 17: Sorry for the inconvenience on this rename 😢

kartikayk (commented):
Sorry reviewing this piece meal - will likely have more questions as I dig deeper

recipes/lora_finetune_distributed.py
```diff
@@ -279,16 +281,13 @@ def _setup_model(
...
+            log.info("FSDP is enabled. Model init and checkpoint loading on Rank 0 ...")
             init_start = time.perf_counter()
 
-            with utils.set_default_dtype(self._dtype):
-                model = config.instantiate(cfg_model)
-
-            log.info(
-                f"Model instantiation took {time.perf_counter() - init_start:.2f} secs"
-            )
+        with utils.set_default_dtype(self._dtype), torch.device("meta"):
```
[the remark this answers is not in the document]
weifengpy: the docstring used to be `Instantiating Model on CPU` (left) and I removed the mention of CPU. I did not mention meta device because it measures meta init + checkpoing loading now. Happy to improve if you are referring to this docstring

recipes/lora_finetune_distributed.py
```diff
@@ -307,16 +306,36 @@ def _setup_model(
...
-                model = config.instantiate(cfg_model)
+        if enable_activation_checkpointing:
+            utils.set_activation_checkpointing(
+                model, auto_wrap_policy={modules.TransformerDecoderLayer}
+            )
+
+        for m in model.modules():
+            if isinstance(m, modules.TransformerDecoderLayer):
+                fully_shard(m)
+        fully_shard(model)
```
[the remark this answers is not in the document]
weifengpy: In FSDP1, we wrap each `TransformerDecoderLayer` and then root `model` as well. It's blackboxed in `auto_wrap_policy=utils.lora_fsdp_wrap_policy(modules_to_wrap={modules.TransformerDecoderLayer})`

In FSDP2, we un-blackboxed it to this for-loop. It you perfer, this can be factored into a util function in torchtune so user call `util.fully_shard(model, modules_to_wrap)` 

Personally I have bias towards un-blackboxed approach since people can modify the for-loop to achieve different wrapping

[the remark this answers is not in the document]
weifengpy: the equivalence SHARD_GRAD_OP in FSDP2 is `reshard_after_forward=False` . Do you want it as a config in .yaml?

`fully_shard(model, reshard_after_forward=False)`

recipes/lora_finetune_distributed.py
weifengpy: you are right. when finetuning from a original HF checkpoint, `lora_weights_state_dict = None`

for resuming training, `lora_weights_state_dict is not None` and we avoided calling `m.initialize_parameters()` again

recipes/lora_finetune_distributed.py
```diff
@@ -371,12 +364,14 @@ def _setup_optimizer(
...
-            # just the state dict of the adapter cfg, but should be equivalent
-            opt_state_dict = utils.transform_opt_state_dict(
-                opt_state_dict, self._model, optimizer
+            set_optimizer_state_dict(
+                self._model,
+                optimizer,
+                optim_state_dict=opt_state_dict,
+                options=StateDictOptions(
+                    broadcast_from_rank0=True, full_state_dict=True
+                ),
```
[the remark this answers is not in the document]
weifengpy: for both FSDP1/FSDP2, parameters are shared into 1/GPUs. Each optimizer only see 1/GPUs portion of the parameters. We start with full optimizer state dict on rank0, and each rank should only load 1/GPUs of it. 

In FSDP2, `set_optimizer_state_dict` serves the purpose. In FSDP1, `utils.transform_opt_state_dict` serves the purpose

torchtune/utils/_distributed.py
```diff
@@ -242,6 +248,38 @@ def lora_wrap_fsdp(module: nn.Module, recurse: bool, **kwargs):
...
+):
+    meta_sharded_sd = model.state_dict()
+    sharded_sd = {}
+    for param_name, full_tensor in full_sd.items():
+        sharded_meta_param = meta_sharded_sd.get(param_name)
+        sharded_tensor = distribute_tensor(
+            full_tensor, sharded_meta_param.device_mesh, sharded_meta_param.placements
+        )
+        sharded_sd[param_name] = nn.Parameter(sharded_tensor)
+    model.load_state_dict(sharded_sd, strict=False, assign=True)
```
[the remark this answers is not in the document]
weifengpy: happy to discuss offline. at high-level, both FSDP1 and FSDP2 shards model parameters into 1/GPUs. For each rank, model.parameters() should only see 1/GPUs of the full parameters (aka `full_tensor`)

In FSDP1, the logic is blackboxed in the left
```
with FSDP.state_dict_type(
      self._model,
      StateDictType.FULL_STATE_DICT,
      FullStateDictConfig(offload_to_cpu=True, rank0_only=True),
      FullOptimStateDictConfig(offload_to_cpu=True, rank0_only=True),
  ):
```

in FSDP2, the logic is unblack boxed as a util function in torchtune

torchtune/utils/_distributed.py
weifengpy: it's all good :D

recipes/lora_finetune_distributed.py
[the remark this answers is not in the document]
weifengpy: oh. Just got you point. Updated docstring for `_setup_model`

torchtune/utils/_distributed.py
awgu: Since we depend on new code changes anyway for DCP, should we just enforce at least a certain pytorch version and avoid this `try`/`except`?

weifengpy: I removed the depdency on newer changes in DCP since there is still gaps on sd["state"], numerics, "initial_lr". FSDP is now the only one. I am open to pytorch version though if it feels more statisfied than `try/except`

weifengpy: that’s a good one. will try to find a version from pytorch nightly

recipes/lora_finetune_distributed.py
```diff
@@ -307,16 +303,40 @@ def _setup_model(
...
-        else:
-            # For non-zero ranks, load the model on meta device
-            with utils.set_default_dtype(self._dtype), torch.device("meta"):
-                model = config.instantiate(cfg_model)
+        if enable_activation_checkpointing:
+            utils.set_activation_checkpointing(
+                model, auto_wrap_policy={modules.TransformerDecoderLayer}
+            )
+
+        for m in reversed(list(model.modules())):
```
kartikayk, line 314: My noob understanding here is that you want to account for the Linear module before the TransformerDecoderLayer - is that right? If so, is there a better way to do this? It wasn't immediately obvious to me that reversing this list gets me that outcome - but maybe that's because I haven't played around with the modules function? Disregard if this is a pretty standard way to achieve this

recipes/lora_finetune_distributed.py
kartikayk: Got you, thanks so much for the explanation! I think something that would be super helpful would be document here in the form of comments the relationship between:
- the modules on which we call ```fully_shard```
- init on meta device
- calling ```initialize_parameters``` and ```reset_parameters```

Also I think there was a technical reason with FSDP1 to call the function ```reset_parameters```. Is that still true? Or can we standardize this with ```initialize_parameters``` in the  modules code? Happy to chat about this offline!

recipes/lora_finetune_distributed.py
```diff
@@ -307,16 +303,40 @@ def _setup_model(
...
+            if (
+                isinstance(m, nn.Linear)
```
kartikayk, line 316: Does this need to be LoRALinear? Or am I misunderstanding?

recipes/lora_finetune_distributed.py
```diff
@@ -307,16 +303,40 @@ def _setup_model(
...
+                model, lora_weights_state_dict, self._device, self._is_rank_zero
+            )
+
+        with utils.set_default_dtype(self._dtype), self._device:
+            for m in model.modules():
+                if isinstance(m, LoRALinear) and not lora_weights_state_dict:
+                    m.lora_a.to_empty(device=self._device)
+                    m.lora_b.to_empty(device=self._device)
+                    m.initialize_parameters()
+                if isinstance(m, modules.RotaryPositionalEmbeddings):
```
kartikayk, line 334: Just to clarify, we special handle RoPE because the buffer is not being loaded from a state dict, right?

torchtune/utils/_distributed.py
```diff
@@ -242,6 +250,79 @@ def lora_wrap_fsdp(module: nn.Module, recurse: bool, **kwargs):
     return lora_wrap_fsdp
 
 
+def load_from_full_model_state_dict(
```
kartikayk, line 253: This is a bit unfortunate but we have two different meanings of "full model" IIUC:
- One is related to FSDP i.e. full vs sharded - is that right?
- Other is LoRA i.e. full model vs LoRA adapters

The current function is a bit confusing since we pass in the adapter state_dict and this is referred to as ```full_sd``` in the function itself.

torchtune/utils/_distributed.py
```diff
@@ -242,6 +250,79 @@ def lora_wrap_fsdp(module: nn.Module, recurse: bool, **kwargs):
...
+    full_sd: Dict[str, Any],
+    device: torch.device,
+    is_rank_zero: bool,
+):
+    meta_sharded_sd = model.state_dict()
+    sharded_sd = {}
+    for param_name, full_tensor in full_sd.items():
+        sharded_meta_param = meta_sharded_sd.get(param_name)
+        sharded_tensor = distribute_tensor(
+            full_tensor, sharded_meta_param.device_mesh, sharded_meta_param.placements
```
kartikayk, line 264: Where is device_mesh and placements information coming from?

recipes/lora_finetune_distributed.py
awgu: I wanted to quickly mention my opinion:

What we want is a post-order traversal of the modules so that we visit lower modules before higher ones. `nn.Module.modules()` always gives _reverse_ post-order, so reversing it gives post-order.

FSDP1 hides all of this under the `auto_wrap_policy` argument, which does the post-order traversal for the user. Personally, I did not want to have an `auto_wrap_policy` argument for FSDP2 because the name is a misnomer. Auto wrapping is not automatically choosing which modules to wrap for the user -- rather, it just performing a post-order traversal and applying a user-defined policy to determine if a module is wrapped or not. What I am open to though is having some other higher-level utility (say, `apply_fsdp()`) that does the same thing as `auto_wrap_policy` but exists outside of the function `fully_shard`.

torchtune/utils/_distributed.py
weifengpy: it's from `fully_shard(model, mesh)`. we are using default `mesh` where every rank serves FSDP, since there is no 2D/3D parallasim involved

after `fully_shard(model, mesh)`, `model.parameters()` is converted from plain tensor to `DTensor` with `mesh`

torchtune/utils/_distributed.py
weifengpy: good catch. you are right "full" is opposing to "sharded" here. What do you think if rename to "local"?  `load_from_local_model_state_dict` ?

recipes/lora_finetune_distributed.py
weifengpy: that's correct

recipes/lora_finetune_distributed.py
weifengpy: `nn.Linear` with `required_grad=True` are referring to `lora_a` and `lora_b`

Trunk FSDP1 is wrapping `lora_a` and `lora_b` separately thus we are do the same wrapping for parity. But as you mentioned, we can wrap `LoRALinear` instead so `lora_a` and `lora_b` are communicated together

tests/recipes/test_lora_finetune_distributed.py
weifengpy: FSDP2 + optimizer state dict and resuming training are supported

recipes/lora_finetune_distributed.py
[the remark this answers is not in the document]
kartikayk: thanks for the explanation! I love the un-blackboxed approach here - just needs more documentation and explanation :) After reading the FSDP2 RFC, this became a lot clearer.

recipes/lora_finetune_distributed.py
weifengpy: good point! will add comment to explain `fully_shard`, meta init, and `reset/initialize_parameters`

FSDP1 calls `reset_parameters` for the exact same reason FSDP2 call `reset/initialize_parameters`: RoPE are not covered in checkpoints, `lora_a` and `lora_b` are not covered in checkpoints for `resume_training=False`

It's just FSDP1 have a contract to call overrided `nn.Module.reset_parameter` through `FSDP(model, param_init=)`, but FSDP2 does not impose overriding `reset_parameter`. now use can name it `reset_parameter` or `initialize_parameters`

torchtune/utils/_distributed.py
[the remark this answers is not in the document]
weifengpy: add docstring, what the function is doing (full tensor into sharded), and `mesh` and `assign=True`

weifengpy:
synced with @kartikayk on action items 
* add docstring for meta init, state dict, migration to DCP
* place FSDP2 in receipes/dev/
* attach wandb loss curve
* pending feedbacks on unit tests

torchtune/utils/_distributed.py
```diff
@@ -8,13 +8,21 @@
...
+from typing import Any, Callable, Dict, Set, Tuple, Type
 
 import torch
 import torch.distributed as dist
 from torch import nn
+
+try:
+    from torch.distributed._composable.fsdp import FSDPModule
+except ImportError:
+    from torch.distributed._composable.fsdp import FSDP as FSDPModule  # noqa: N811
```
ebsmothers, line 20: (In reference to the previous resolved conversation) To make sure I'm understanding correctly, is the try/except related to FSDP1 vs FSDP2? Is there a clear branch cut between the two versions? One other thing that could be tricky about an explicit version check is that I guess these imports will be executed anytime we import (i.e. even for recipes that have less restrictive version requirements). And I don't see a clear way to e.g. gate inside a function since the module is used as a type-hint in the function signature. So for that reason I feel like try/except might be the way to go here

pyproject.toml
ebsmothers: Is `run_tests` strictly required for the usage of `FSDPTest`, or is it more used for convenience? (Either way not a huge issue)

torchtune/utils/_distributed.py
ebsmothers: Can we add docstrings for the functions added in this file?

torchtune/utils/_distributed.py
```diff
@@ -242,6 +250,83 @@ def lora_wrap_fsdp(module: nn.Module, recurse: bool, **kwargs):
...
+def get_full_model_state_dict(
+    model: FSDPModule,
+    is_rank_zero: bool,
+) -> Dict[str, Any]:
+    sharded_sd = model.state_dict()
+    cpu_state_dict = {}
+    for param_name, sharded_param in sharded_sd.items():
+        full_param = sharded_param.full_tensor()
+        if is_rank_zero:
+            cpu_state_dict[param_name] = full_param.cpu()
```
ebsmothers, line 283: When saving QLoRA state dict, will we also upcast from NF4 here? (Obv not a concern for this PR)

torchtune/utils/_distributed.py
```diff
@@ -242,6 +250,83 @@ def lora_wrap_fsdp(module: nn.Module, recurse: bool, **kwargs):
...
+        sharded_meta_param = meta_sharded_sd.get(param_name)
+        # `.to(dtype)` ensures same dtype when `assign=True`
+        sharded_tensor = distribute_tensor(
+            full_tensor.to(sharded_meta_param.dtype),
+            sharded_meta_param.device_mesh,
+            sharded_meta_param.placements,
+        )
+        sharded_sd[param_name] = nn.Parameter(sharded_tensor)
+    # choose `assign=True` since we cannot call `copy_` on meta tensor
+    model.load_state_dict(sharded_sd, strict=False, assign=True)
```
ebsmothers, line 271: If we catch missing and unexpected keys from `load_state_dict` with `strict=False` what format will the keys be in? Previously with FSDP1 the keys contained all the info about FSDP wrapping. E.g.`model.layers.0._fsdp_flat_param.attn.q_proj.weight` (probably not exactly right but something like that). Will that still be the case here?

torchtune/utils/_distributed.py
```diff
@@ -242,6 +250,83 @@ def lora_wrap_fsdp(module: nn.Module, recurse: bool, **kwargs):
...
+) -> Dict[str, Any]:
+    PARAMS = "params"  # noqa: N806
+    _init_optim_state(opt)
+    param_groups = opt.state_dict()["param_groups"]
+    state = opt.state_dict()["state"]
+
+    full_param_groups = full_sd["param_groups"]
+    full_state = full_sd["state"]
+
+    for param_group, full_param_group in zip(param_groups, full_param_groups):
```
ebsmothers, line 302: Do we ever need to worry about proper ordering for param groups (similar q for the zip in L307)?

torchtune/utils/_distributed.py
```diff
@@ -242,6 +250,83 @@ def lora_wrap_fsdp(module: nn.Module, recurse: bool, **kwargs):
...
+            for attr, full_tensor in full_param_state.items():
+                sharded_tensor = param_state[attr]
+                if isinstance(sharded_tensor, DTensor):
+                    param_state[attr] = distribute_tensor(
+                        full_tensor,
+                        sharded_tensor.device_mesh,
+                        sharded_tensor.placements,
+                    )
+                else:
+                    param_state[attr] = full_tensor
```
ebsmothers, line 321: Might be useful to add some comments for this code block

ebsmothers (commented):
This is looking great! Most of my comments are minor and around documentation. 

I saw there was also some discussion around the location for this recipe. I am inclined to agree with the point you and @kartikayk discussed around putting it in `recipes/dev`. I don't want that to be a long-term home for it, but also want to make sure we don't break folks who are using the current recipe and aren't on the latest version of PyTorch yet.

pyproject.toml
weifengpy: it's strictly required for the usage of `FSDPTest`

tests/recipes/test_lora_finetune_distributed.py
```diff
@@ -50,6 +51,9 @@ def _fetch_expected_loss_values(self):
         return [10.5136, 10.4857, 10.5292, 10.5345]
 
     @pytest.mark.integration_test
+    @pytest.mark.skipif(
+        version.parse(torch.__version__).base_version < "2.4.0", reason=""
+    )
```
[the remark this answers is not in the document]
weifengpy: that's correct. FSDP2 will be released in 2.4.0, although it has been in nightly for a while

torchtune/utils/_distributed.py
weifengpy: > is the try/except related to FSDP1 vs FSDP2

try/except is for FSDP2, where we renamed `FSDP` to `FSDPModule` in `torch.distributed._composable.fsdp`

torchtune/utils/_distributed.py
weifengpy: for FSDP2, it's clean FQNs without FSDP prefix. For example, `layers.0.attn.q_proj.lora_a.weight`

FSDP2 is clean because 1) `fully_shard` register hooks instead wrap `nn.Module`,  2) `fully_shard` set `module.__class__ =  type(f"FSDP{cls.__name__}", (FSDPModule, cls), dct)`  https://fburl.com/i20yr3s2

torchtune/utils/_distributed.py
weifengpy: good question! I am open to discuss on 2 options in case you have preference in QLoRA PR
* torch.save(NF4) directly
* upcast and torch.save(full precision)

torchtune/utils/_distributed.py
weifengpy: good question! I assume `load_from_full_optimizer_state_dict` is called when resume training. that means as long as the order matches with `get_optimizer_state_dict`, it should be fine.  both functions are under our control

a longer-term solution might be using FQNs to align on order

torchtune/utils/_distributed.py
awgu: I thought we could remove this `try: ... except: ...` since we need newer PyTorch version than when we renamed it anyway.

tests/recipes/test_lora_finetune_distributed.py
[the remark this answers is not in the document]
ebsmothers: Thanks for clarifying! In that case I think it makes sense to create `dev/recipes/lora_finetune_fsdp2.py` (or something like that) in the short-term, then migrate to replace `recipes/lora_finetune_distributed.py` once we've socialized and gotten enough users onto >= 2.4.0. Let me know how that sounds to you.

torchtune/utils/_distributed.py
ebsmothers: > I thought we could remove this try: ... except: ... since we need newer PyTorch version than when we renamed it anyway.

Oh yeah but this file may still be imported in other recipes with less strict version requirements (e.g. our full finetune distributed recipe). So in that case we could run into trouble. (Ultimately we can eventually restrict to >= 2.4 across the board, but idk that we're there yet)

torchtune/utils/_distributed.py
ebsmothers: I think the latter is preferable if possible, since NF4 is not really a universally-recognized dtype iiuc. This is what we do currently (see [here](https://github.com/pytorch/torchtune/blob/0c4056a6a55bfffd3cec7a99a13dc978cfe85963/torchtune/modules/common_utils.py#L15))

torchtune/utils/_distributed.py
ebsmothers: This is great! I think this means we can actually do validation of LoRA state dict load more cleanly (note that we actually have two separate utilities for this for the [single-device](https://github.com/pytorch/torchtune/blob/0c4056a6a55bfffd3cec7a99a13dc978cfe85963/recipes/lora_finetune_single_device.py#L282) vs [distributed](https://github.com/pytorch/torchtune/blob/0c4056a6a55bfffd3cec7a99a13dc978cfe85963/recipes/lora_finetune_distributed.py#L304) case because of this FSDP prefix issue). Not a concern for this PR but this will allow us to clean up our code a bit

torchtune/utils/_distributed.py
ebsmothers: Thanks for the clarification, this sounds like a reasonable assumption to me (we make kinda similar assumptions for other intermediate state saved for checkpoint resume anyways)

torchtune/utils/_distributed.py
weifengpy: the pointer is very helpful. now i know the torchtune way. will upcast then

tests/recipes/test_lora_finetune_distributed.py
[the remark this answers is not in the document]
weifengpy: proposing a middle ground: type hinting with "FSDPModule" without try...except
```
def load_from_full_model_state_dict(
    model: "FSDPModule",
```

ebsmothers (approved):
A couple small comments on the configs but otherwise no major concerns. Really excited to see this in our library!

recipes/configs/dev/llama2/13B_lora_fsdp2.yaml
```diff
@@ -0,0 +1,86 @@
+# Config for multi-device LoRA in lora_finetune_distributed.py
+# using a Llama2 13B model
```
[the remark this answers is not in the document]
weifengpy: good catch! updated to lora_finetune_fsdp2 and mentioned FSDP2

recipes/dev/lora_finetune_fsdp2.py
```diff
@@ -0,0 +1,671 @@
...
+        set_trainable_params(model, self.adapter_params)
+
+        if enable_activation_checkpointing:
+            utils.set_activation_checkpointing(
+                model, auto_wrap_policy={modules.TransformerDecoderLayer}
+            )
+
+        # iterating from lowerer modules to higher
+        # eg grouping lora adapters before transformer block
+        for m in reversed(list(model.modules())):
```
awgu, line 322: By the way, another option here is to just make two passes if that is clearer.
```
for module in model.modules():
    if <LoRA adapter>:
        fully_shard(module)
for module in model.modules():
    if <transformer block>:
        fully_shard(module)
```

PR #865, opened 2024-04-25
FSDP Llama3 wrapping improvements for full finetune

rohan-varma:
#### Context
- [x] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

#### Changelog

This PR primarily seeks to improve memory efficiency specifically for llama3 full distributed training and enable a distributed finetune in 4x 24GB of memory. We do this with a new FSDP wrapping policy that wraps the token embedding and output projections. These are much larger for llama3 due to the increased vocab size, so sharding them across GPUs has more of an effect.

- Added new API to retrieve memory efficient FSDP wrapping policy. To control whether the memory efficient wrapping policy is retrieved, we introduce a new flag `memory_efficient_wrapping` in our configs. Currently, this is only set to True for llama3 distributed full finetuning. As follow up work, we'll investigate other workloads with this wrapping and enable where beneficial.
- Added appropriate unittests
- Integrated in `full_finetune_distributed`. Did a bit of study on potential integration into LoRA, but memory savings were less pronounced there - this needs further investigations.

#### Test plan

- Added unittests.

This PR only seeks to ship improvements to llama3 training. 

#### Docs
image
image
image

### Full finetune

Run command for full finetune: `CUDA_VISIBLE_DEVICES=0,3,6,7 tune run --nproc_per_node 4 full_finetune_distributed --config llama3/8B_full batch_size=1`
- With this PR: `peak_memory_active:20.830272 peak_memory_alloc:19.085376 peak_memory_reserved:23.699914752`, 1.06it/s
- Without this PR: `peak_memory_active:24.170446336 peak_memory_alloc:21.988057088 peak_memory_reserved:27.908898816`, 1.08it/s
- About 13% savings in allocated memory, 15% in memory resereved. This allows us to get a 4x 24GB finetune.
- NOTE: A previous version of this PR also wrapped the token embedding and output projection in their own activation checkpointing units, but this is not needed. Vocabulary size is increased, but activations generated are proportional to sequences, not vocab size, so checkpointing these won't help more for llama3 compared to llama2. A quick study checkpointing these versus not shows roughly the same memory efficiency. In particular, with checkpointing the token embedding and output proj, we achieve `peak_memory_active:20.880037376 peak_memory_alloc:19.135141376 peak_memory_reserved:24.13821952`, while without it, we achieve the numbers reported above: they are very comparable.


torchtune/utils/memory.py
```diff
@@ -23,6 +23,40 @@
...
+        # ensure torch.nn.Embedding activations are checkpointed.
+        modules_to_wrap.add(torch.nn.Embedding)
+        return partial(_llama3_ac_policy, modules_to_wrap=modules_to_wrap)
+    else:
+        return ModuleWrapPolicy(modules_to_wrap)
+
+def _llama3_ac_policy(module: nn.Module, recurse: bool, modules_to_wrap, **kwargs):
+    # Label that output_proj should be wrapped individually.
+    if isinstance(module, modules.TransformerDecoder):
+        targ = module.output.module
```
rohan-varma, line 47: this should probably use a helper function called `_get_fsdp_wrapped_module` or something that is aware of whether `module.output` is wrapped in FSDP or not and intelligently unwraps it.

ebsmothers (commented):
Overall this makes sense to me and the memory savings for full finetune are great. My main question is around whether model type is the most natural way to expose this feature.

There's nothing about this functionality that is unique to Llama3, it's just that it proves most beneficial there. By doing things this way we are kinda making the decision for people that only Llama3 should have this feature, and supporting other models with large vocab sizes will then require updating the wrapping internals instead of just flipping a config. I know we had discussed Gemma as one specific model where this is a challenge, but I wonder if we can do an assert on the backend to raise an error if the model class is not `TransformerDecoder` as we would expect.

ebsmothers:
>  CUDA_VISIBLE_DEVICES=0,3,6,7

👀

rohan-varma:
> Overall this makes sense to me and the memory savings for full finetune are great. My main question is around whether model type is the most natural way to expose this feature.
> 
> There's nothing about this functionality that is unique to Llama3, it's just that it proves most beneficial there. By doing things this way we are kinda making the decision for people that only Llama3 should have this feature, and supporting other models with large vocab sizes will then require updating the wrapping internals instead of just flipping a config. I know we had discussed Gemma as one specific model where this is a challenge, but I wonder if we can do an assert on the backend to raise an error if the model class is not `TransformerDecoder` as we would expect.

@ebsmothers I definitely agree here. My proposal on a way forward would be to decouple the model type from the checkpointer and offer it as a general accessor to determine which model is being trained - there's currently no easy way to go about this. And I'd like for this change to be especially focused on llama3 (so the initial rollout of these policies will only be done for llama3). As follow up work we should enable for llama2 and investigate other models, although verifying these improvements are currently a long-running process and should ideally be done iteratively and/or by multiple folks, IMO

@joecummings could you chime in on ModelType for this sort of use case and if you happen to have any, alternative ways to achieve this sort of gating based on specific models here?

joecummings:
@rohan-varma I could totally be missing something here, but why can't we include `embedding` in the modules to wrap within the config for Llama3, rather than tie this directly to `ModelType`? That way, you can expand this to any new models that have this large embedding space, which is starting to become more popular. 

cc: @ebsmothers

rohan-varma:
> @rohan-varma I could totally be missing something here, but why can't we include `embedding` in the modules to wrap within the config for Llama3, rather than tie this directly to `ModelType`? That way, you can expand this to any new models that have this large embedding space, which is starting to become more popular.
> 
> cc: @ebsmothers

@joecummings This is because `modules_to_wrap` is not configurable right now, and configuring it would be a little tricky (i.e. we'd have to parse the string like `torch.nn.Embedding` and make it a class)

rohan-varma:
> I haven't tested this but will this allow 70B on 8x80GB? I was only able to full-fine tune 70B with cpu offloading

This unfortunately won't allow 70B on 8x80GB from my experiments without CPU offloading, but can do a bit more testing. our current thinking is to enable full finetune for 70B models with CPU offload.

torchtune/utils/memory.py
```diff
@@ -15,14 +16,67 @@
...
+    """
+    if isinstance(module, FSDP):
+        return module.module
+    return module
+
+
+def _llama3_ac_policy(module: nn.Module, recurse: bool, modules_to_wrap, **kwargs):
+    # Label that output_proj should be wrapped individually.
+    if isinstance(module, modules.TransformerDecoder):
+        targ = _maybe_fsdp_unwrap(module.output)
```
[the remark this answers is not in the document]
rohan-varma: So this isn't really because we want to support FSDP and non FSDP models, but we do happen to have this support with this function.

The reason we need this unwrap call is if the model, when we wrap in AC, has already been wrapped in FSDP. Then we want to AC wrap `module.output`, but just accessing `module.output` when the model is FSDP wrapped may give us back the FSDP class. So we further unwrap to retrieve the local nn.module.

recipes/full_finetune_distributed.py
```diff
@@ -291,7 +292,10 @@ def _setup_model(
         # across all available GPUs.
         model = FSDP(
             module=model,
-            auto_wrap_policy=ModuleWrapPolicy({modules.TransformerDecoderLayer}),
+            auto_wrap_policy=utils.get_full_finetune_fsdp_wrap_policy(
+                memory_efficient_wrapping=cfg.memory_efficient_wrapping,
```
ebsmothers, line 296:
```suggestion
                memory_efficient_wrapping=memory_efficient_wrapping,
```

recipes/full_finetune_distributed.py
```diff
@@ -314,7 +318,11 @@ def _setup_model(
         # original activation checkpointing (full) - flip the condition above
         if enable_activation_checkpointing and ac_mode is None:
             utils.set_activation_checkpointing(
-                model, auto_wrap_policy={modules.TransformerDecoderLayer}
+                model,
+                auto_wrap_policy=utils.get_ac_policy(
+                    memory_efficient_wrapping=cfg.memory_efficient_wrapping,
```
ebsmothers, line 323: same comment here

torchtune/utils/_distributed.py
```diff
@@ -212,3 +214,59 @@ def lora_wrap_fsdp(module: nn.Module, recurse: bool, **kwargs):
...
+    Returns:
+        FSDPPolicyType: Wrapping policy that can be passed into ``FullyShardedDataParallel``.
+    """
+    if memory_efficient_wrapping:
+        return _llama3_full_fsdp_wrap_policy(modules_to_wrap=modules_to_wrap)
+    else:
+        return ModuleWrapPolicy(modules_to_wrap)
+
+
+def _llama3_full_fsdp_wrap_policy(modules_to_wrap: Set[Type]) -> FSDPPolicyType:
```
ebsmothers, line 245: Maybe rename from llama3 based on latest changes?

torchtune/utils/memory.py
```diff
@@ -15,14 +16,77 @@
...
 from torchtune.utils.logging import get_logger
 
 _log: logging.Logger = get_logger()
 
 ACWrapPolicyType: Type = Union[Set[Type], Callable[[nn.Module, bool, int], bool]]
 
 
+def get_ac_policy(
+    memory_efficient_wrapping: bool, modules_to_wrap: Set[Type]
+) -> ACWrapPolicyType:
```
kartikayk, line 31: I know some of this isn't introduced by this PR, but this is a really dense function which requires way more documentation, comments and explanation.

I have no idea what this means:
```
Union[Set[Type], Callable[[nn.Module, bool, int], bool]]
```

torchtune/utils/_distributed.py
kartikayk: We had a bit of this discussion on the chat thread, but at some point Nicolas called this out and I tend to agree, the return type and associated doc string is not very helpful here - I dont know what ```FSDPPolicyType``` means. Can we add more documentation on this? I don't think most of the users will be at the level of sophistication where this type or how it maps to the policy will be immediately clear.

torchtune/utils/_distributed.py
```diff
@@ -212,3 +214,59 @@ def lora_wrap_fsdp(module: nn.Module, recurse: bool, **kwargs):
...
+        modules_to_wrap (Set[Type]): nn.Module types to recursively wrap
+    Returns:
+        FSDPPolicyType: Wrapping policy that can be passed into ``FullyShardedDataParallel``.
+    """
+    modules_to_wrap.add(torch.nn.Embedding)
+
+    def llama3_wrap(module: nn.Module, recurse: bool, **kwargs):
+        # Label that output_proj should be wrapped individually.
+        if isinstance(module, modules.TransformerDecoder):
+            module.output._wrap = True
```
kartikayk, line 262: @ebsmothers how does this impact some of the transformer decoder changes that we're planning to do?

torchtune/utils/memory.py
```diff
@@ -15,14 +16,77 @@
...
+        # ensure torch.nn.Embedding activations are checkpointed.
+        modules_to_wrap.add(torch.nn.Embedding)
+        return partial(_mem_efficient_ac_policy, modules_to_wrap=modules_to_wrap)
+    else:
+        return ModuleWrapPolicy(modules_to_wrap)
+
+
+def _maybe_fsdp_unwrap(module: nn.Module) -> nn.Module:
+    """
+    Unwrap a module if it is wrapped by FSDP.
```
kartikayk, line 58: Add more details on the scenarios where something like this is needed

torchtune/utils/_distributed.py
kartikayk: What does ```_wrap``` refer to here? I tried to look at the FSDP docs and the nn.Module docs and it's not immediately clear to me what this is doing? Sorry if I'm being dense on this.

torchtune/utils/memory.py
```diff
@@ -15,14 +16,77 @@
...
+    # Label that output_proj should be wrapped individually.
+    if isinstance(module, modules.TransformerDecoder):
+        targ = _maybe_fsdp_unwrap(module.output)
+        targ._ac_wrap = True
+    if recurse:
+        return True
+
+    # Wrap output_proj individually. Manual check for module not being FSDP,
+    # since when attributes are accessed on FSDP module, FSDP forwards the attribute
+    # access, and we don't want
```
kartikayk, line 83: I think this is trailing off?

kartikayk (commented):
Thanks for adding this!

I don't think I fully understand this:

> New AC wrapping policy that checkpoints the token embedding and output projections as well. Similar reason to above - they generate larger activations so it would be useful to not store those in memory.

Irrespective of the size of the vocab, the output of the embedding table would just depend on the sequence length? So why does this have anything to do with the vocab size? Or am I misunderstanding?

recipes/full_finetune_distributed.py
rohan-varma: yep, thanks for catching these! Accidentally committed a wrong version of the code.

torchtune/utils/_distributed.py
```diff
@@ -212,3 +237,60 @@ def lora_wrap_fsdp(module: nn.Module, recurse: bool, **kwargs):
...
+        with FSDP.
+        modules_to_wrap (Set[Type]): Set of module types to wrap.
+
+    Note:
+        ``memory_efficient_fsdp_wrap`` memory improvements have currently only been verified on llama3 workloads,
+            to provide ~15% memory improvement (when used alongside AC memory efficient wrapping). Other workloads
+            have not been verified and may not see the same improvements.
+    Returns:
+        FSDPPolicyType: Wrapping policy that can be passed into ``FullyShardedDataParallel`` as the ``auto_wrap_policy``
+            argument. Please see documentation for `torchtune.utils.FSDPPolicyType` for additional details.
```
rohan-varma, line 261: Is there any way to link directly to this docstr? @ebsmothers or @NicolasHug  maybe?

rohan-varma:
> Thanks for adding this!
> 
> I don't think I fully understand this:
> 
> > New AC wrapping policy that checkpoints the token embedding and output projections as well. Similar reason to above - they generate larger activations so it would be useful to not store those in memory.
> 
> Irrespective of the size of the vocab, the output of the embedding table would just depend on the sequence length? So why does this have anything to do with the vocab size? Or am I misunderstanding?

@kartikayk Thanks for the feedback and the review! You're totally right that this doesn' t have anything to do with the vocab size and this was an oversight on my end. I verified that if we remove the modified AC wrapping, we don't change anything about the memory improvements we're shippping here. So this PR is now only limited to FSDP wrapping changes. 

Also added a bunch more documentation to FSDPPolicyType to clearly explain it to the user  and link back to FSDP wrapping docs where useful. thanks!

torchtune/utils/_distributed.py
ebsmothers: Re the transformer decoder changes, I think the main thing is that the `if isinstance` check may need to be generalized (since now e.g. our `TransformerDecoderLM` or `TransformerDecoderClassifier` classes will both have output layers). But realistically I think the main use case will still be for when we're projecting to vocab_size, so maybe just directly replacing with `TransformerDecoderLM` (or whatever we're calling it) will be sufficient here.

torchtune/utils/_distributed.py
ebsmothers: Also bump: are we still naming this based on llama3?

rohan-varma: Renamed

torchtune/utils/_distributed.py
```diff
@@ -212,3 +240,61 @@ def lora_wrap_fsdp(module: nn.Module, recurse: bool, **kwargs):
...
+        return _llama3_full_fsdp_wrap_policy(modules_to_wrap=modules_to_wrap)
+    else:
+        return ModuleWrapPolicy(modules_to_wrap)
+
+
+def _llama3_full_fsdp_wrap_policy(modules_to_wrap: Set[Type]) -> FSDPPolicyType:
+    """
+    A default policy for wrapping Llama-3 style models for full finetuning using FSDP. Specifically,
+    this will wrap the model's token embedding and output projection into their own FSDP units to
+    maximize memory savings. After this is done, model will also be hierarchically wrapped
```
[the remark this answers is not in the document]
rohan-varma: Added

torchtune/utils/_distributed.py
rohan-varma: @ebsmothers IIUC, it doesn't seem that we need any particular changes right now since the classifier support is only there for mistral at the moment? Although yeah we should keep an eye in the future - if we send a classifier model to this wrapping policy, it might not end up wrapping things correctly. Probably can consider it in a future PR?

ebsmothers (approved):
OK a few more comments but overall no major concerns from my side

recipes/configs/llama2/7B_full.yaml
```diff
@@ -63,6 +63,7 @@ device: cuda
 
 # Memory management
 enable_activation_checkpointing: True
+memory_efficient_fsdp_wrap: False
```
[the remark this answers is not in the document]
rohan-varma: Oh yeah, unfortunate that it doesn't fail. We should really run all our configs in CI.

torchtune/utils/_distributed.py
```diff
@@ -212,3 +240,63 @@ def lora_wrap_fsdp(module: nn.Module, recurse: bool, **kwargs):
...
+) -> FSDPPolicyType:
+    """
+    Retrieves an FSDP wrapping policy based on the specified flags ``memory_efficient_fsdp_wrap`` and
+    ``modules_to_wrap``. Specifically, if ``memory_efficient_fsdp_wrap`` is set to ``True``, the returned
+    policy will wrap the model's token embedding and output projection in addition to the modules specified
+    to maximize memory savings.
+
+    Args:
+        memory_efficient_fsdp_wrap (bool): If ``True``, will also wrap embedding and output projection layers
+        with FSDP.
```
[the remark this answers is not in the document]
rohan-varma: Oh oops, I changed this for LoRA on L226 but didn't change it here, thanks!

Issue #781, opened 2024-04-17
utils.set_activation_checkpointing is unnecessarily restrictive


rohan-varma:
The API enforces that the wrapping policy just be a set of modules, which is sufficient for a few use cases but the underlying API offers more generality in terms of wrapping policies. We should expose this to our users or have them directly use apply_activation_checkpointing.

felipemello1:
Closing it since PR was merged.

PR #866, opened 2024-04-25
Revise the implementation of set_activation_checkpointing

rohan-varma:
#### Context
- [x] add a new feature
- [x] fix a bug
- [ ] other (please add here)

`set_activation_checkpointing` is a bit restrictive in that it only allows policies to be taken in via a defined `set` instead of allowing users to also author policies. We'll soon be improving our activation checkpointing to make llama3 training more memory efficient, which will require a more custom policy this PR will enable. More context in https://github.com/pytorch/torchtune/issues/781. Also improving some of the documentation around the API and removing allowing passing of `None` into the policy since this does not make sense

#### Changelog
- Improve documentation
- Stop allowing `None` passed as the policy
- Allow custom policy to be passed in, in addition to sets
- Add unittests 

#### Test plan

`pytest test_memory.py -v`

### DIFF
--- a/tests/torchtune/utils/test_memory.py
+++ b/tests/torchtune/utils/test_memory.py
@@ -1,0 +1,36 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+import pytest
+import torch.nn as nn
+from torch.distributed.algorithms._checkpoint.checkpoint_wrapper import (
+    CheckpointWrapper,
+)
+from torchtune.utils import set_activation_checkpointing
+
+
+class TestSetActivationCheckpointing:
+    @pytest.fixture
+    def model(self) -> int:
+        return nn.Sequential(nn.Linear(10, 10), nn.Linear(10, 10), nn.Linear(10, 10))
+
+    def _verify(self, model):
+        for submodule in model.modules():
+            if isinstance(submodule, CheckpointWrapper):
+                assert isinstance(submodule._checkpoint_wrapped_module, nn.Linear)
+
+    def test_activation_checkpoint_set_policy(self, model):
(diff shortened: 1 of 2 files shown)


torchtune/utils/memory.py
```diff
@@ -20,19 +20,27 @@
...
-    """Utility to setup activation checkpointing and wrap the model for checkpointing.
+    """Utility to apply activation checkpointing to the passed in model.
 
     Args:
         model (nn.Module): Model to setup activation checkpointing.
-        auto_wrap_policy (Optional[Set[nn.Module]]): Policy to wrap module.
-        **kwargs: additional arguments to pass to torch.distributed activation checkpointing.
+        auto_wrap_policy (ACWrapPolicyType): Policy to wrap module.
+            This can either be a set of ``nn.Module`` types, in which case, modules of the specified type(s)
+            will be wrapped individually with activation checkpointing or a ``callable`` policy describing
```
ebsmothers, line 35: nit: commas
```suggestion
            This can either be a set of ``nn.Module`` types, in which case modules of the specified type(s)
            will be wrapped individually with activation checkpointing, or a ``callable`` policy describing
```

tests/torchtune/utils/test_memory.py
```diff
@@ -0,0 +1,36 @@
...
+from torch.distributed.algorithms._checkpoint.checkpoint_wrapper import (
+    CheckpointWrapper,
+)
+from torchtune.utils import set_activation_checkpointing
+
+
+class TestSetActivationCheckpointing:
+    @pytest.fixture
+    def model(self) -> int:
+        return nn.Sequential(nn.Linear(10, 10), nn.Linear(10, 10), nn.Linear(10, 10))
```
ebsmothers, line 18: nit: maybe something slightly more nontrivial here (i.e. not all nn.Linears)

ebsmothers (approved):
Couple small nits but looks good!

PR #875, opened 2024-04-26
Sample packing for map datasets with correct RoPE encoding and no cross-contamination

RdoubleA:
### The Problem
Packing multiple samples within a single context window means the model may accidentally attend to other samples it should not attend to. If there are sequences that are completely orthogonal in topic/content, cross-contamination is adding a lot of noise for that sample. Additionally, if position ids are not correctly adjusted for each individual sequence with a packed sample, then tokens that are later in the pack get an unwarranted later position bias. So, we need to add two things:
- Lower triangular causal block mask for attention to prevent cross contamination
- Adjusted rope embeddings to handle positions ids for each sequence in a packed sample

This is a highly requested feature by the community, as seen in many discussions in TRL, Hugging Face, and llama recipes [here](https://github.com/huggingface/trl/issues/805) and [here](https://github.com/huggingface/transformers/issues/17726) and [here](https://github.com/meta-llama/llama-recipes/issues/341) and [here](https://github.com/huggingface/transformers/issues/25452)

However, PyTorch Core's SDPA does not support flash attention with a non-causal mask (you can see several discussions on this [here](https://discuss.pytorch.org/t/scaled-dot-product-attention-bf16-with-my-attn-mask/182658), [here](https://github.com/pytorch/pytorch/issues/96099#issuecomment-1498388194), and [here](https://github.com/Lightning-AI/litgpt/issues/620#issuecomment-1899350319)). This means to do sample packing properly, we need to either turn to other implementations or use PyTorch's memory-efficient SDPA. 

### The Approach
Mem-eff SDPA is not as fast as flash v2 (2x slower), so we _could_ turn to Dao's original implementation of flash attention that supports varied sample lengths. Since torchtune does not use fused QKV to support MQA and GQA, we should use the [`flash_attn_varlen_func`](https://github.com/Dao-AILab/flash-attention/blob/83aef842beec1037eb8c1d9c3ef3ed8aae80b091/flash_attn/flash_attn_interface.py#L843) as an alternative to SDPA. This allows our between sample masking via cumulative sequence lengths in the batch. 

However, adding a third party dependency especially in such a core module should not be taken lightly and invites risk for breakages, lack of support, lack of control, etc. We need to evaluate whether the improved performance significantly outweighs this tradeoff.

### Changelog
In this update, to unblock sample packing I decided to stick with mem eff SDPA + proper packing mask. More benchmarking of Dao's flash attention compared to mem eff attention needs to be done to understand the tradeoffs, which I will do as a follow-up. Here is a summary of the updates:
- For proper masking and position encoding, I took advantage of the `mask` and `input_pos` kwargs in the transformer layers. This was enough to cover both without requiring any additional conversion or mask calculations, while still maintaining minimal changes to model forward signature.
- Leaned on `PackedDataset` to create the lower triangular block mask and positions ids for each subsample within a pack, since it's much more efficient to create it while packing alongside the tokens and labels. Having to infer the mask and position ids batchwise is expensive and requires iterating through the whole batch again. 
- This means `PackedDataset` now returns tokens, labels, mask, and input_pos. To generalize this for all datasets, all dataset classes now return a dictionary. All of them will still just have "tokens" and "labels" keys, but `PackedDataset` will additionally return "mask" and "input_pos". This also gives flexibility for new datasets to also return mask and input_pos if needed
- Because the dataset return signature changes, all recipes needed to be updated to key into the batch accordingly (just a few lines)
- `TransformerDecoder` forward now can take in an optional mask. This cannot be used on inference when causal mask is used. The mask is propagated down to the attention module. Now packing mask can be used in SDPA.
- Rope positional embeddings can use `input_pos` as a 2D tensor that has position ids for each sample in each pack in a batch, with a very tiny change.
- Added a positional embedding test for the 2D input_pos case
- Collating happens at the sample level for packed datasets with `_padded_collate_packed`, so we do not use a collator in the dataloader if dataset is packed

## Test plan
`tune run full_finetune_single_device --config llama3/8B_full_single_device epochs=1 dataset.packed=True dataset.max_seq_len=4096`

Sample packing improves performance proportional to max sequence length - at 4096, tokens/sec were at about 1800, which is nearly 6x the unpacked version at a higher max sequence length of 8192. This effect may be exaggerated the more long-tailed the sequence length distribution of the dataset is. Longer sequence length (4096) also brought down run time to about 1 hour per epoch, compared to 1 hr 45 min at 512 when packed and nearly 5-6 hours at 512 when not packed. Packing for 52k samples for Alpaca only takes 40 seconds.

Loss curves across all packed and unpacked runs align.

image
image

Evals are nearly identical for packed runs vs non-packed runs.

<google-sheets-html-origin><style type="text/css"></style>
run | truthfulqa | stderr | hellaswag | stderr
-- | -- | -- | -- | --
llama3_alpaca_unpacked | 0.4589353557 | 0.01488350256 | 0.5846444931 | 0.004917761182
llama3_alpaca_packed | 0.490953071 | 0.01462141724 | 0.5734913364 | 0.00493558773
llama3_alpaca_packed_no_mask | 0.4783502665 | 0.01494382987 | 0.5749850627 | 0.004933349622
llama3_alpaca_packed_no_pos | 0.4909587427 | 0.01462954265 | 0.5723959371 | 0.00493719976
llama3_alpaca_packed_no_mask_no_pos | 0.4798960754 | 0.014649153 | 0.5780720972 | 0.004928578106

--------------------------

Huge shoutout to @calmitchell617 for initiating this in #827, I've incorporated some of your logic here for a basic sample packing feature.

### What is sample packing?
Sample packing maximizes the sequence length / context window by jamming as many samples as can fit so we don't waste compute on padding tokens. This leads to much faster training because the model can process more data with fewer forward passes. However, there's usually a slower start up if you perform packing on your dataset prior to training. 

There are ways to do this offline, with a bin packing algorithm, or on-the-fly, but for now I do the naive approach of greedily packing samples as a preprocessing step when initializing the dataset. This will need to evolve further once we support IterableDatasets and streaming.

### How is shuffling handled?
The DistributedSampler will still handle shuffling in the recipe layer, so for a packed dataset only the packs are shuffled and not within a pack. We may need to add within-pack shuffling later, although it will be simpler with IterableDatasets.

### How can I configure packing?
Simply set `packed=True` in the config. This gets routed to the builder, so all dataset builders are updated with this flag. This is preferable to having a recipe parameter, because that would require updating all configs.
```
dataset:
  _component_: torchtune.dataset.alpaca_dataset
  packed: True
```


torchtune/datasets/_packed.py
```diff
@@ -0,0 +1,78 @@
...
+        buffer = {
+            "input_ids": [],
+            "labels": [],
+        }
+
+        for input_ids, labels in tqdm(
+            self.ds, desc="Packing dataset", dynamic_ncols=True
+        ):
+            buffer["input_ids"].extend(input_ids)
+            buffer["labels"].extend(labels)
```
xingyaoww, line 62: I wonder if we want to optionally allow the user to add a separator (e.g., `<eod>`) for packing (e.g., via a argument?)?

torchtune/datasets/_packed.py
```diff
@@ -0,0 +1,78 @@
...
+
+            # If buffer has reached max_seq_len, append packed sample
+            while len(buffer["input_ids"]) > self.max_seq_len:
+                self.samples.append(
+                    {k: v[: self.max_seq_len] for k, v in buffer.items()}
```
xingyaoww, line 67: Maybe it will be good to support only packing examples together only when all of them can fit into ONE window.
If adding an example will exceed the context window, maybe we should NOT add that example to the current pack, set the rest of the current pack with `<PAD>`, and move that example to the next pack? This will be very helpful for people doing SFT (fewer examples, but hope to get the benefit of packing without the need to truncate any sentences).

I did something like I described above here - feel free to re-use some part of the logic if interested: https://github.com/xingyaoww/Megatron-LLM/blob/main/tools/preprocess_instruct_data.py#L148-L194

torchtune/datasets/_packed.py
```diff
@@ -0,0 +1,78 @@
...
+    def _pack(self) -> None:
+        """
+        Iterate through the dataset. Use a buffer to hold samples until max_seq_len,
+        then append the buffer to self.samples as a single "packed" sample. Continue
+        until max_rows or end of dataset.
+        """
+        # buffer to hold samples until they are long enough to be added to self.samples
+        buffer = {
+            "input_ids": [],
+            "labels": [],
```
xingyaoww, line 55: I am actually thinking about whether we should also build attention/sentence masks for packed sequence (something like flash-attn did: https://github.com/Dao-AILab/flash-attention/issues/432#issuecomment-1698610752). So that the model would be able to know which part belongs to sentence 1, which part belongs to sentence 2, etc, this can be used to adjust the self-attention so that different sequences cannot attend to each other.

torchtune/datasets/_packed.py
```diff
@@ -0,0 +1,78 @@
...
+        max_seq_len (int): Maximum number of tokens to pack
+        max_rows (Optional[int]): maximum number of samples to pack. Default is None, which will pack as many samples as possible.
+    """
+
+    def __init__(
+        self,
+        ds: Dataset,
+        max_seq_len: int,
+        max_rows: Optional[int] = None,
+    ) -> None:
```
xingyaoww, line 38: Ideally, if this is an `IterableDataset` it will be much more efficient (e.g., when doing pre-training - dataset is too large to fit in memory for packing). But i guess that can be pretty challenging to do, especially considering shuffling.

xingyaoww (commented):
Super happy to see this project :) I wanted to find a codebase that's simple, usable, with reasonable performance.
I think `torchtune` pretty much satisfies most of my needs except sentence packing (which you already did in this PR!) & integrating more models (e.g., `Mixtral`). Would be excited to see this PR go through!

torchtune/datasets/_packed.py
RdoubleA: The EOS tokens should serve as inherent separators, no? And your other comment about creating a sentence mask is a preferable approach imo

torchtune/datasets/_packed.py
RdoubleA: Excellent point. I agree, I think we should do this by default for finetune. Although, for pretraining datasets / unstructured text data maybe we don't need to perform padding and it's ok to split samples?

torchtune/datasets/_packed.py
RdoubleA: Yes, absolutely agree. Having an integer mask that assigns an integer for each sample in the pack to prevent cross-attending should be supported. Although, I need to think about how to propagate this mask properly to the model forward, right now our datasets have only returned the input tokens and labels. But this should probably change...

torchtune/datasets/_packed.py
RdoubleA: Yeah the IterableDataset would be a more efficient approach, but the design of this in a distributed setting with shuffling and packing support with modular, extensible design that can generalize to all our datasets is challenging. This is currently in the works (cc @SLR722) but may take some time so I wanted to add packing for simple map style datasets that can load in memory for now.

torchtune/datasets/_packed.py
xingyaoww: `sentence_mask` is the ultimate solution i agree! with that we probably don't need this separator at all

torchtune/datasets/_packed.py
xingyaoww: Agree! For pre-training, we can just go with the current implementation!

torchtune/datasets/_packed.py
xingyaoww: I agree! This will a long-term thing IMO - i think load in memory is pretty sufficient more 99% SFT workload :)

torchtune/datasets/_packed.py
RdoubleA: I'll add a flag then to control this!

torchtune/datasets/_packed.py
```diff
@@ -0,0 +1,78 @@
...
+        }
+
+        for input_ids, labels in tqdm(
+            self.ds, desc="Packing dataset", dynamic_ncols=True
+        ):
+            buffer["input_ids"].extend(input_ids)
+            buffer["labels"].extend(labels)
+
+            # If buffer has reached max_seq_len, append packed sample
+            while len(buffer["input_ids"]) > self.max_seq_len:
```
calmitchell617, line 65: A relatively unimportant point, but I believe this `>` could be `>=`, which may end up creating an extra sample.

calmitchell617:
@RdoubleA this looks great. Your implementation of the packing logic is much cleaner than mine :-)

One thing you might want to add is a test that validates how individual tokens are packed. This was easy for me to do because my code had complete control over initialization. It might be more complex now because `PackedDataset` is integrated with other dataset classes, but should still be manageable.

This kind of test might be overkill now, but further down the line when someone else is working on this class (or other dataset classes that integrate with it) it might be a helpful check.

Here is how I did it:

```
class TestInstructDataset:
    expected_tokenized_prompts = [
        [0, 4, 2, 1, 7, 4, -1, 0, 1, 9],
        [5, 2, 6, 4, 3, 8, -1, 0, 4, 3],
    ]

    def get_samples(self):
        samples_list = [
            "This is a packing test",
            "A fantastic test. It should pack two samples.",
            "This one will not be fully packed.",
        ]

        samples_dict = {"content": samples_list}

        return Dataset.from_dict(samples_dict)

    @mock.patch("torchtune.datasets._concat.load_dataset")
    def test_get_item(self, mock_load_dataset):
        mock_load_dataset.return_value = self.get_samples()
        dataset = ConcatDataset(
            tokenizer=DummyTokenizer(),
            source="were/going/jellyfishing",
            text_column="content",
            max_seq_len=10,
        )
        assert len(dataset) == 2
        mock_load_dataset.assert_called_once()

        for i in range(len(dataset)):
            prompt, label = dataset[i]
            assert prompt == self.expected_tokenized_prompts[i]
            assert label == self.expected_tokenized_prompts[i]
```

torchtune/datasets/_packed.py
RdoubleA: Hmm let me double check this

RdoubleA:
@calmitchell617 great suggestion, I was actually planning to borrow some of your testing logic so I'll try to include this. Also I'm glad someone else caught my hidden SpongeBob reference 👌

calmitchell617:
I do not need streaming, this unblocks my use case.

torchtune/datasets/_slimorca.py
```diff
@@ -17,17 +15,17 @@ def slimorca_dataset(
     *,
     tokenizer: Tokenizer,
     source: str = "Open-Orca/SlimOrca-Dedup",
-    chat_format: Optional[ChatFormat] = Llama2ChatFormat,
+    chat_format: Optional[str] = None,
```
ebsmothers, line 18: Just curious: why do we default to None here now? For better Llama3 support out of the box?

tests/torchtune/datasets/test_packed_dataset.py
```diff
@@ -0,0 +1,86 @@
...
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+import pytest
+from tests.test_utils import DummyTokenizer
+from torch.utils.data import Dataset
+
+from torchtune.datasets import PackedDataset
```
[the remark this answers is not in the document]
ebsmothers: And a second testing q: do we want to test packed versions of existing datasets as well?

tests/torchtune/datasets/test_packed_dataset.py
```diff
@@ -0,0 +1,86 @@
...
+        # Check we get right number of packs
+        assert len(packed) == max_rows
+        # Check input ids and labels are same length
+        assert len(packed[0][0]) == len(packed[0][1])
+        # Check that samples are packed correctly - very last individual sample
+        # should have index value of the number of times dataset was iterated over
+        if split_samples:
+            # If we split samples, we'll know how many samples by taking the
+            # full length and dividing by sample size
+            last_index, remainder = divmod(max_rows * max_seq_len, sample_size)
```
ebsmothers, line 60: Did not even know divmod was a thing. Nice

torchtune/datasets/_packed.py
```diff
@@ -0,0 +1,113 @@
...
+            self.ds, desc="Packing dataset", dynamic_ncols=True
+        ):
+            # If the dataset outputs samples that are larger than the specified
+            # max_seq_len and we're unable to split it, user needs to modify
+            # one of the two parameters
+            if len(input_ids) > self.max_seq_len and not self.split_samples:
+                raise ValueError(
+                    f"Dataset sample is too long ({len(input_ids)} > {self.max_seq_len}). "
+                    "Please set `split_samples=True` or increase `max_seq_len`."
+                )
```
ebsmothers, line 78: We should think about whether this is the right thing to do here. Without sample packing we wouldn't error out here, right? We would just truncate. I wonder if it makes sense to do the same here.

tests/torchtune/datasets/test_packed_dataset.py
ebsmothers: Actually could also use math.ceil(max_rows * max_seq_len / sample_size) here I think? Maybe that's clearer tbh

ebsmothers (approved):
This is awesome! Left a handful of comments but no huge concerns from my side

torchtune/datasets/_packed.py
```diff
@@ -0,0 +1,113 @@
...
+        when buffer is long enough, add to self.samples.
+
+    During training, returns self.samples[idx] as input and label.
+
+    Args:
+        ds (Dataset): dataset to sample pack. This should return a tuple of tokenized
+            inputs and labels.
+        max_seq_len (int): Maximum number of tokens to pack
+        max_rows (Optional[int]): maximum number of samples to pack. Default is None, which will pack as many samples as possible.
+        split_samples (bool): if the last sample in a pack does not fit in ``max_seq_len``,
```
kartikayk, line 31: I feel like ```split_across_instances``` or ```split_across_boundary``` would be more intuitive. If I just found this flag in a config, I'd interpret it as splitting all samples across instances i.e. no packing. Ignore this comment if this is an accepted convention in the community. If not, I'd suggest thinking a bit about this flag.

torchtune/datasets/_packed.py
```diff
@@ -0,0 +1,113 @@
...
+
+from typing import Dict, List, Optional, Tuple
+
+from torch.utils.data import Dataset
+from tqdm import tqdm
+
+
+class PackedDataset(Dataset):
+    """
+    Performs greedy sample packing on a provided dataset. This is done as a single
```
kartikayk, line 15: It would be nice to call out that this only supports in-memory datasets

torchtune/datasets/_packed.py
```diff
@@ -0,0 +1,113 @@
...
+            if len(input_ids) > self.max_seq_len and not self.split_samples:
+                raise ValueError(
+                    f"Dataset sample is too long ({len(input_ids)} > {self.max_seq_len}). "
+                    "Please set `split_samples=True` or increase `max_seq_len`."
+                )
+
+            current_pack["input_ids"].extend(input_ids)
+            current_pack["labels"].extend(labels)
+
+            if len(current_pack["input_ids"]) > self.max_seq_len:
```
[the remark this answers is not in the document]
kartikayk: Maybe this is just me, but I find the logic a bit round about. Why not just see if the current length of the pack + length of incoming sample is > max length or not? If it is, you right the current pack out, if not just add it. Maybe I'm missing something basic?

torchtune/datasets/_packed.py
```diff
@@ -0,0 +1,113 @@
...
+                current_pack = self._add_pack(
+                    current_pack=current_pack,
+                    boundary=self.max_seq_len
+                    if self.split_samples
+                    else previous_sample_boundary,
+                )
+
+            previous_sample_boundary = len(current_pack["input_ids"])
+            if self.max_rows is not None and len(self.samples) >= self.max_rows:
```
kartikayk, line 92: So max_rows is only respected if we have < max_seq_len tokens right? Then why have that option at all? User can just reduce the seq len to get fewer samples? Or why would they want fewer samples?

kartikayk (changes requested):
As discussed offline, let's figure out masking correctly before landing this PR.

tests/torchtune/datasets/test_packed_dataset.py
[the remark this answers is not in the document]
RdoubleA: I don't see the need to as long as the packing functionality is properly tested. That's why we have @calmitchell617 's test on example texts

torchtune/datasets/_slimorca.py
RdoubleA: yes, llama3 will be used more than llama2 certainly so that should be the default

torchtune/datasets/_packed.py
RdoubleA: hm yeah that makes sense to me

torchtune/datasets/_packed.py
RdoubleA: I'll change it to `split_across_pack`

torchtune/datasets/_packed.py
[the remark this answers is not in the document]
RdoubleA: The nuance is when you split the sample, if you check current length + length of incoming then you either write the current pack out or split the incoming sample upto max seq len and write the pack out. IMO that logic and the logic here are almost identical

torchtune/datasets/_packed.py
RdoubleA: not exactly - max rows lets users limit how many total packs in the dataset is returned. I imagine this will be a lot more relevant for streamed / iterable datasets. max_seq_len controls the size of each individual pack. reducing max_seq_len but keeping dataset size the same will actually result in more samples.

calmitchell617: Yes, I agree with @RdoubleA. Max rows is how many packed rows you end with. Most users will choose `None` for datasets that fit in RAM, but will choose some lesser value for OOM datasets. A very nice helper arg might be to calculate `max_rows` for you. Not needed for now.

torchtune/datasets/_packed.py
```diff
@@ -0,0 +1,113 @@
...
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+from typing import Dict, List, Optional, Tuple
+
+from torch.utils.data import Dataset
+from tqdm import tqdm
+
+
+class PackedDataset(Dataset):
```
joecummings, line 13: Also, maybe call this GreedyPackedDataset? For when we add different versions?

torchtune/datasets/_samsum.py
```diff
@@ -45,11 +47,12 @@ def samsum_dataset(
         >>> Batch size: 8
     """
 
-    return InstructDataset(
+    return instruct_dataset(
         tokenizer=tokenizer,
         source=source,
-        template=SummarizeTemplate,
+        template="SummarizeTemplate",
```
joecummings, line 53: why quotes here?

torchtune/datasets/_instruct.py
```diff
@@ -107,12 +108,14 @@ def _prepare_sample(self, sample: Mapping[str, Any]) -> Tuple[List[int], List[in
 
 
 def instruct_dataset(
+    *,
```
joecummings, line 111: 🫡

torchtune/datasets/_packed.py
RdoubleA: mm I'm inclined to leave it as is and rename it when we add different versions if needed. don't want to jump the gun

torchtune/datasets/_samsum.py
RdoubleA: we are going through the instruct dataset builder instead of the class which requires specifying the template as a string (the builder is meant to be more config friendly)

torchtune/datasets/_packed.py
kartikayk: Oh hmm, then I'm not sure this is clear in the doc string

> max_rows (Optional[int]): maximum number of samples to pack. Default is None, which will pack as many samples as possible.

This leads me to believe this param controls the number of samples read in not the number f packs returned? Or am I misunderstanding?

calmitchell617: I wrote that comment.

The intent is to signal that this arg controls the total amount of packed samples you will end up with at the end of the process. However, if you don't understand it, then others wont, either. So, the phrasing should probably be revised for clarity.

torchtune/datasets/_packed.py
tcapelle: This is indeed something we may want, but... the guys from trl have been doing packing for a while without this feature with really good fine-tunes.

kartikayk: @tcapelle you mean without masking or something else? If it's without masking, that's just.....wrong? Even if it gives decent performance on benchmarks, I dont think we want to enable that?

torchtune/modules/transformer.py
```diff
@@ -158,12 +160,22 @@ def setup_caches(self, max_batch_size: int, dtype: torch.dtype) -> None:
             torch.ones(self.max_seq_len, self.max_seq_len, dtype=torch.bool)
         )
 
-    def forward(self, tokens: Tensor, input_pos: Optional[Tensor] = None) -> Tensor:
+    def forward(
+        self,
+        tokens: Tensor,
```
joecummings, line 165: I would be in favor of having the optional params be keyword-only to avoid confusion in the future.

cc: @ebsmothers

torchtune/datasets/_alpaca.py
```diff
@@ -6,16 +6,17 @@
...
 from functools import partial
 
-from torchtune.data import AlpacaInstructTemplate
-from torchtune.datasets._instruct import InstructDataset
+from torchtune.datasets._instruct import instruct_dataset, InstructDataset
 from torchtune.modules.tokenizers import Tokenizer
 
 
 def alpaca_dataset(
+    *,
```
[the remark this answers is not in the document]
joecummings: nit: I like it, but technically since `tokenizer` is the only required param, you could put it after that param.

torchtune/datasets/_packed.py
```diff
@@ -0,0 +1,142 @@
...
+            "tokens": [],
+            "labels": [],
+            "mask": [],
+            "input_pos": [],
+        }
+        # Keep track of what index the previous sample ends in case we need
+        # to end a pack early
+        previous_sample_boundary = 0
+
+        for tokens, labels in tqdm(self.ds, desc="Packing dataset", dynamic_ncols=True):
```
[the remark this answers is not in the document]
RdoubleA: Not sure I understand, I would still need to retrieve the tokens and labels right, which would require iterating? I'm doing a naive greedy approach here

calmitchell617 (approved):
@RdoubleA, good job! I performed the following integration / sanity checks, and it all worked as expected:

I performed several training runs with:
```
```

For the first run, I enabled packing by altering `recipes/configs/llama3/8B_lora_single_device.yaml` like so:

```
# Dataset and Sampler
dataset:
  train_on_input: True
  packed: True # turn on packing
seed: null
shuffle: True
batch_size: 2
```
This halved the amount of iterations, so it is clearly doing it's job as intended!

I then decided to see if label masking is still working (for excluding user tokens from loss calculations) by adding a few checks and printlines in `recipes/lora_finetune_single_device.py` like so:

```
                    if (
                        self.max_steps_per_epoch is not None
                        and (idx // self._gradient_accumulation_steps)
                        == self.max_steps_per_epoch
                    ):
                        break

                    if self._profiler_enabled:
                        self._profiler.step()

                    tokens, labels = batch["tokens"], batch["labels"]

                    # see if some tokens are -100
                    print(tokens)
                    print(labels)

                    for i, sample in enumerate(labels):
                        for j, token in enumerate(sample):
                            if token not in [tokens[i][j], -100]:
                                raise Exception
```

Once again, everything looks good in this high level integration test.

My only comment is that the project's readme (and maybe a recipe config) should include a little info on enabling the feature. Other than that, LGTM!

tcapelle:
There is a tradeoff, pre-computing them is faster but if we serialize it would be very memory expensive. We can compute them with the position ids, maybe storing them as ids like `1, 1, 1, 2, 2, 2, 3]` and then constructing the block triu:
image
We are already touching the `TransformerDecoderLayer`, maybe there?

Also, maybe renaming `positon_ids` instead of `input_pos` ❤️ ?

RdoubleA:
> Or when creating the batch with a DataCollator?

I debated this exact thing when trying to figure out where to create the mask and position ids, I was mainly focused on simplicity and speed. If we have to recreate the mask in TransformerDecoderLayer or the collator we need to loop through the entire batch and the seqlens for each pack which is quite inefficient, though it might be worth benchmarking what the time difference is. But you're right that this would be much cheaper memory wise.

> Also, maybe renaming positon_ids instead of input_pos ❤️ ?

This was the existing name, so I'd rather not add a refactor on top of all the changes here :)

> Can we still train with packing and without the positional masking? I mean attending to other samples? 

My opinion is that this is incorrect and should not be supported. Samples in an SFT dataset are typically highly correlated and should not cross-attend. But others are welcome to add their thoughts on this.

torchtune/datasets/_chat.py
```diff
@@ -100,7 +101,7 @@ def _prepare_sample(self, sample: Mapping[str, Any]) -> Tuple[List[int], List[in
         labels = list(np.where(mask, CROSS_ENTROPY_IGNORE_IDX, tokens))
         assert len(tokens) == len(labels)
 
-        return tokens, labels
+        return {"tokens": tokens, "labels": labels}
```
kartikayk, line 104: Can't believe we haven't been doing this all along :)

torchtune/datasets/_packed.py
```diff
@@ -0,0 +1,163 @@
...
+                    current_pack=current_pack,
+                    boundary=self.max_seq_len
+                    if self.split_across_pack
+                    else previous_sample_boundary,
+                )
+
+            if rank == 0:
+                pbar.update()
+            previous_sample_boundary = len(current_pack["tokens"])
+            if self.max_packs is not None and len(self.packs) >= self.max_packs:
```
kartikayk, line 121: Isn't ```len(self.packs) >= self.max_packs``` violating the ```max_packs``` value? I thought the intent was to make sure you return less than ```max_packs```?

torchtune/datasets/_packed.py
```diff
@@ -0,0 +1,163 @@
...
+            )
+            assert len(current_pack["tokens"]) == 0
+
+    def _add_pack(
+        self, current_pack: Dict[str, List[int]], boundary: int
+    ) -> Dict[str, List[int]]:
+        """
+        Pad and add the current pack to self.packs and return what's remaining.
+        """
+        packing_mask = torch.block_diag(*current_pack["mask"])
```
kartikayk, line 139: Can use more comments about the structure of the mask and the expected shape

torchtune/modules/position_embeddings.py
```diff
@@ -107,8 +109,9 @@ def forward(self, x: Tensor, input_pos: Optional[Tensor] = None) -> Tensor:
         xshaped = x.float().reshape(*x.shape[:-1], -1, 2)
 
         # reshape the cache for broadcasting
-        # tensor has shape [1, s, 1, n_d // 2, 2]
-        rope_cache = rope_cache.view(1, xshaped.size(1), 1, xshaped.size(3), 2)
+        # tensor has shape [b, s, 1, n_d // 2, 2] if packed samples,
```
kartikayk, line 112: I dont think I understand this?

tcapelle:
This is more regarding the naming of the padding functions
> Collating happens at the sample level for packed datasets with _padded_collate_packed, so we do not use a collator in the dataloader if dataset is packed
When I read this, I got the idea that the `PackedDataset` is already returning batches of data that don't need collation. I feel that these functions (also the `padded_collate`) should be named `packed_padding`  or something like, as the collation for me at least is constructing a batch, `cat([sample1, sample2])`.

recipes/full_finetune_distributed.py
```diff
@@ -364,8 +364,10 @@ def _setup_data(
                 for single_cfg_dataset in cfg_dataset
```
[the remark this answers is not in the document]
RdoubleA: @ebsmothers and I discussed moving this logic to a config util that can instantiate ListConfigs, but longer term this might also be updated with iterable datasets

torchtune/modules/position_embeddings.py
RdoubleA: previously the rope embeddings are the same for each sample within a batch because they will all go to max seq len, but now the position ids may be different for each sample because each sample has multiple sequences of different lengths, so now the rope cache has an added batch dimension.

RdoubleA:
> When I read this, I got the idea that the PackedDataset is already returning batches of data that don't need collation

Hm I see what you mean, technically this is not a collator, `_padded_collate_packed` is just handling padding. Maybe this should be moved to a utils file

winglian:
> I debated this exact thing when trying to figure out where to create the mask and position ids, I was mainly focused on simplicity and speed. If we have to recreate the mask in TransformerDecoderLayer or the collator we need to loop through the entire batch and the seqlens for each pack which is quite inefficient, though it might be worth benchmarking what the time difference is. But you're right that this would be much cheaper memory wise.


Another option is building it in the model forward rather than each decoder layer. This way it only needs to be generated once.

torchtune/datasets/_packed.py
```diff
@@ -0,0 +1,163 @@
...
+
+    During training, returns self.packs[idx] as input, label, attention mask, and
+    position ids. The attention mask is a lower triangular block mask to prevent
+    samples from cross-attending within a pack. The position ids indicate the position
+    of each token relative to its sample within a pack. These are all padded to max
+    sequence length, so a batch-wise collator is not needed.
+
+    Args:
+        ds (Dataset): dataset to sample pack. This should return a tuple of tokenized
+            inputs and labels.
```
ebsmothers, line 37: Update this

torchtune/datasets/_packed.py
```diff
@@ -0,0 +1,163 @@
...
+    The class loads, tokenizes, and packs examples on initialization - no tokenization is done during training.
+
+    The general flow on initialization is: load tokenized sample -> add to buffer ->
+        when buffer is long enough, add to self.packs.
+
+    During training, returns self.packs[idx] as input, label, attention mask, and
+    position ids. The attention mask is a lower triangular block mask to prevent
+    samples from cross-attending within a pack. The position ids indicate the position
+    of each token relative to its sample within a pack. These are all padded to max
+    sequence length, so a batch-wise collator is not needed.
```
ebsmothers, line 33: Overall great docstring. There's a lot going on here, so wonder if it might make sense to just write out an explicit example here (especially what mask and input_pos look like when packing a couple samples). It'd be really nice to have something like that in the live docs for people to refer to.

torchtune/utils/collate.py
```diff
@@ -133,3 +129,68 @@ def padded_collate_dpo(
...
+    mask = sample["mask"]
+    input_pos = sample["input_pos"]
+
+    # Pad to max sequence length
+    tokens = F.pad(
+        torch.tensor(tokens), (0, max_seq_len - len(tokens)), value=padding_idx
+    )
+    labels = F.pad(
+        torch.tensor(labels), (0, max_seq_len - len(labels)), value=ignore_idx
+    )
```
ebsmothers, line 174: Thoughts on validating keys, shapes, anything like that here? Personally I think it's OK since it's a private API and we want it to be fast, but otherwise usually I would suggest to do it

torchtune/utils/collate.py
```diff
@@ -133,3 +129,68 @@ def padded_collate_dpo(
...
+    # For attention mask, simply use identity matrix for the pad tokens
```
ebsmothers, line 175: Identity matrix is not for any particular reason, right? Just for convenience?

torchtune/utils/collate.py
```diff
@@ -133,3 +129,68 @@ def padded_collate_dpo(
...
+    mask_pad = torch.eye(max_seq_len - mask.shape[0], dtype=torch.bool)
+    mask = torch.block_diag(mask, mask_pad)
+    # For position ids, continue to increment for pad tokens
+    input_pos_pad = torch.arange(
+        input_pos[-1] + 1, max_seq_len - len(input_pos) + input_pos[-1] + 1
```
ebsmothers, line 180: nit: a bit confusing, maybe just define `next_pos=input_pos[-1] + 1` in a separate line to make the arange clearer

torchtune/utils/collate.py
```diff
@@ -133,3 +129,68 @@ def padded_collate_dpo(
...
+    )
+    # Do not go beyond max_seq_len - 1
+    input_pos_pad = input_pos_pad.clamp(max=max_seq_len - 1)
```
ebsmothers, line 183: I don't fully follow this. Does this only happen when `split_across_pack=True` or something? Also in that case where are we truncating the tokens and labels? Is it in the base dataset? If so it's a bit confusing to have it spread out across three places (though admittedly I don't really see a better way since base dataset obviously shouldn't care about input_pos)

torchtune/modules/transformer.py
```diff
@@ -189,12 +202,19 @@ def forward(self, tokens: Tensor, input_pos: Optional[Tensor] = None) -> Tensor:
         # shape: [b, s, d]
         h = self.tok_embeddings(tokens)
 
-        mask = None
+        # shape: [b, 1, s, s]
+        if mask is not None:
+            mask = mask[:, None, :, :]
```
ebsmothers, line 207: Thoughts on doing this in attention.py? Cause (a) that's where we actually need it to be this shape, and (b) then we do not have separate contracts on mask shape in `TransformerDecoder` and `TransformerDecoderLayer`

torchtune/modules/position_embeddings.py
ebsmothers: Yeah maybe I'm being a little dense here, but I don't understand how we index into `self.cache` with the 2D input_pos tensor and it just works out of the box.

tests/torchtune/datasets/test_packed_dataset.py
```diff
@@ -0,0 +1,206 @@
...
+                torch.eye(remainder, dtype=torch.bool),
+            )
+            input_pos.extend(list(range(sample_size, sample_size + remainder)))
+
+        return mask[:max_seq_len, :max_seq_len], torch.tensor(input_pos[:max_seq_len])
+
+    @pytest.mark.parametrize("max_seq_len", [10, 25])
+    @pytest.mark.parametrize("sample_size", [2, 5])
+    @pytest.mark.parametrize("max_packs", [5, 10])
+    @pytest.mark.parametrize("split_across_pack", [True, False])
```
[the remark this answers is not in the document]
RdoubleA: at the very least need to test when a sample gets split or not split

torchtune/datasets/_chat.py
RdoubleA: dictionaries are simpler, more generalizable :)

tests/torchtune/modules/test_position_embeddings.py
```diff
@@ -85,6 +85,35 @@ def test_forward_with_curr_pos(
...
+        )
+
+        # these values should be exactly the same as test_forward
+        # AND test_forward_with_current_pos. In this case input_pos
+        # covers the entire batch dim and is defined for each sample separately.
+        # This tests that input_pos works as expected i.e.
+        # extracts the embeddings for the relevant positions for each sample
+        assert_expected(x_out.mean(), tensor(6.4543e-05), atol=1e-4)
+        assert_expected(x_out.sum(), tensor(2165.7053))
+        assert_expected(x_out.max(), tensor(5.4546))
```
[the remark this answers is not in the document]
RdoubleA: but doesn't this test that input_pos is indexing correctly even as a batch? and that's the only thing we need to test imo

recipes/full_finetune_distributed.py
```diff
@@ -364,8 +364,10 @@ def _setup_data(
...
             ]
             ds = ConcatDataset(datasets=datasets)
+            packed = False
```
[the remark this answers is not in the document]
RdoubleA: maybe not, but frankly I don't want to inflate this PR more than it already is

torchtune/modules/position_embeddings.py
RdoubleA: pytorch magic
image

tests/torchtune/modules/test_position_embeddings.py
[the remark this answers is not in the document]
ebsmothers: Yeah I guess, I am still just wary about cases where we have something other than pure torch.arange for each row. But anyways your other comment convinced me so no blocker here

torchtune/modules/transformer.py
```diff
@@ -158,12 +161,22 @@ def setup_caches(self, max_batch_size: int, dtype: torch.dtype) -> None:
...
         Args:
             tokens (Tensor): input tensor with shape [b x s]
-            input_pos (Optional[Tensor]): Optional tensor which contains the position
-                of the current token. This is only used during inference. Default is None
+            mask (Optional[Tensor]): Optional tensor which contains mask to indicate
+                which tokens will participate in the attention calculation, shape [b x s x s].
+            input_pos (Optional[Tensor]): Optional tensor which contains the position ids
+                of each token. During training, this is used to indicate the positions
+                of each token relative to its sample when packed, shape [b x s].
+                During inference, this indicates the position of the current token.
```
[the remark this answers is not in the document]
RdoubleA: yes but it is being used similarly in both cases no? or we can rename this to `position_ids` as @tcapelle suggested

torchtune/utils/collate.py
```diff
@@ -133,3 +129,68 @@ def padded_collate_dpo(
...
+
+    This is a sample-wise collator and should not be used with the dataloader.
+
+    Sample should look like::
+        {
+            "tokens": List[int],
+            "labels": List[int],
+            "mask": Tensor,
+            "input_pos": List[int],
+        }
```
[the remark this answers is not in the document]
RdoubleA: I'm using `torch.tril` to create the mask so it's easier to keep as a tensor... we could make all the other types tensors, don't know if there's a specific reason "tokens" and "labels" were kept as lists

torchtune/utils/collate.py
```diff
@@ -133,3 +129,68 @@ def padded_collate_dpo(
...
+    sample: Dict[str, Any],
+    max_seq_len: int,
+    padding_idx: int = 0,
+    ignore_idx: int = CROSS_ENTROPY_IGNORE_IDX,
+) -> Dict[str, torch.Tensor]:
+    """Pad a batch of packed sequences to max sequence length in the batch, and
+    convert integer lists to tensors. Account for attention mask and position
+    ids.
+
+    This is a sample-wise collator and should not be used with the dataloader.
```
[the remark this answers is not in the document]
RdoubleA: Yeah agreed as @tcapelle mentioned, this is technically not a collator so I should rename this for clarity

torchtune/utils/collate.py
RdoubleA: no reason, tried to pick the most neutral thing

torchtune/utils/collate.py
RdoubleA: we are not truncating - the creation of the packs are "truncating" to a set max_seq_len. If we split across packs there is an edge case for the last sample if it bleeds over into the next pack where the pos ids continue to the end of that sample (say sample had length 5 and max_seq_len is 8, and [..., 0, 1, 2] ended up in the previous pack, and the sample got split into the last pack [3, 4]. We still have 6 slots left, how do we pad position ids here without going beyond max_seq_len?). So I just enforced this, these position ids should be ignored anyway by the loss. But the better solution is a padding mask

torchtune/utils/collate.py
RdoubleA: yeah, shapes might be good to validate

torchtune/modules/transformer.py
RdoubleA: yes, I think that makes sense

torchtune/utils/collate.py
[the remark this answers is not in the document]
RdoubleA: gonna skip on this because I need the list append functionality as we continue packing

torchtune/datasets/_alpaca.py
```diff
@@ -6,16 +6,17 @@
...
 
-from torchtune.data import AlpacaInstructTemplate
-from torchtune.datasets._instruct import InstructDataset
+from torchtune.datasets._instruct import instruct_dataset, InstructDataset
 from torchtune.modules.tokenizers import Tokenizer
 
 
 def alpaca_dataset(
     tokenizer: Tokenizer,
+    *,
```
joecummings, line 15: nit: be consistent here. In other things you put the kw delineator as the first arg, and in others you put it after tokenizer. Just pick one, either works.

RdoubleA: Hm for instruct and chat dataset builders there are many more required positional arguments which is why I put the asterisk in the beginning. I can just put it in the beginning for all of them

torchtune/datasets/_alpaca.py
```diff
@@ -37,6 +38,7 @@ def alpaca_dataset(
         max_seq_len (int): Maximum number of tokens in the returned input and label token id lists.
             Default is 512, but we recommend setting this to the highest you can fit in memory and
             is supported by the model. For example, llama2-7B supports up to 4096 for sequence length.
+        packed (bool): Whether or not to pack the dataset prior to training. Default is False.
```
rohan-varma, line 41: Guessing it'll be packed to `max_seq_len`? Might be worth clarifying this?

torchtune/datasets/_packed.py
```diff
@@ -0,0 +1,274 @@
...
+        max_packs: Optional[int] = None,
+        split_across_pack: bool = False,
+    ) -> None:
+        self.ds = ds
+        self.max_seq_len = max_seq_len
+        self.max_packs = max_packs
+        self.split_across_pack = split_across_pack
+        # where final samples will be held
+        self.packs: List[Dict[str, List[int]]] = []
+        self._pack()
```
rohan-varma, line 96: why not just return it? `self.packs = self._pack()`?

torchtune/datasets/_packed.py
```diff
@@ -0,0 +1,274 @@
...
+    sample normally.
+
+    Args:
+        ds (Dataset): dataset to sample pack. This should return a dictionary with field
+            "tokens" and "labels" containing the tokenized and label samples.
+        max_seq_len (int): Maximum number of tokens to pack
+        max_packs (Optional[int]): maximum number of packs. Default is None, which will create as many
+            packs as possible.
+        split_across_pack (bool): if the last sample in a pack does not fit in ``max_seq_len``,
+            split the sample into the next pack, or move it to the beginning of the next pack.
```
rohan-varma, line 77: "or move it" -> "or move it entirely" maybe for some clarity

torchtune/datasets/_packed.py
```diff
@@ -0,0 +1,274 @@
...
+    def _add_pack(
+        self, current_pack: Dict[str, List[int]], boundary: int
+    ) -> Dict[str, List[int]]:
+        """
+        Pad and add the current pack to self.packs and return what's remaining.
+        """
+        # So far we've kept a list of causal masks, one for each sample in the pack.
+        # Now we need to combine them into a single mask for the entire pack.
+        packing_mask = torch.block_diag(*current_pack["mask"])
+        pack = {
```
rohan-varma, line 177: Not sure if it'll be simpler to manage this `pack` as a dataclass but might be worth considering.

torchtune/datasets/_packed.py
```diff
@@ -0,0 +1,274 @@
...
+                )
+
+            # Create integer mask and position ids for current sample and extend
+            # current pack
+            current_sample = {
+                "tokens": tokens,
+                "labels": labels,
+                # Mask is simply a causal mask within this sample length
+                "mask": [torch.tril(torch.ones(seq_len, seq_len, dtype=torch.bool))],
+                "input_pos": list(range(seq_len)),
```
rohan-varma, line 139: curious, why do we store an entire list instead of just the sequence length offset?

torchtune/datasets/_packed.py
```diff
@@ -0,0 +1,274 @@
...
+            }
+            current_pack = {k: v + current_sample[k] for k, v in current_pack.items()}
+
+            # If the current pack is long enough, add it to self.packs and retain
+            # any truncated samples for next pack, if splitting samples
+            if len(current_pack["tokens"]) > self.max_seq_len:
```
rohan-varma, line 145: Curious, why not `>=`?

torchtune/datasets/_packed.py
```diff
@@ -0,0 +1,274 @@
...
+
+    The identity matrix is used in the mask for pad tokens instead of a causal mask.
+    For position ids for pad tokens, we simply continue to increment from the previous
+    sample normally.
+
+    Args:
+        ds (Dataset): dataset to sample pack. This should return a dictionary with field
+            "tokens" and "labels" containing the tokenized and label samples.
+        max_seq_len (int): Maximum number of tokens to pack
+        max_packs (Optional[int]): maximum number of packs. Default is None, which will create as many
```
rohan-varma, line 74: Do we have the appropriate error checking here and do we really need to expose this? What if I specify max_packs as something silly like 1 or 2, but the dataset packed into 2 samples isn't possible because it exceeds max_seq_len?

torchtune/datasets/_packed.py
```diff
@@ -0,0 +1,274 @@
...
+
+            if rank == 0:
+                pbar.update()
+            previous_sample_boundary = len(current_pack["tokens"])
+            if self.max_packs is not None and len(self.packs) >= self.max_packs:
+                break
+
+        # Add the last pack with remaining samples that did not fit in previous
+        if len(current_pack["tokens"]) > 0 and (
+            self.max_packs is None or len(self.packs) < self.max_packs
```
rohan-varma, line 161: Don't fully understand why we need `max_packs` and if we don't need it, this code might become simpler.

torchtune/datasets/_packed.py
rohan-varma: Oh, this is related to RoPE, awesome!

rohan-varma (commented):
Amazing work, this is probably the best PR I've seen in torchtune!

torchtune/datasets/_packed.py
```diff
@@ -0,0 +1,274 @@
...
+        # Keep track of what index the previous sample ends in case we need
+        # to end a pack early
+        previous_sample_boundary = 0
+
+        # Only show progress bar on rank 0
+        _, rank = get_world_size_and_rank()
+        if rank == 0:
+            pbar = tqdm(total=len(self.ds), desc="Packing dataset", dynamic_ncols=True)
+
+        for batch in self.ds:
```
rohan-varma, line 120: Is this iteration rank / distributed aware? i.e. will each data parallel rank read different data to pack?

rohan-varma (commented):
General concern about data parallelism -

1) When running sample pack, will each rank read different data, to ensure the generated packs are still correctly sharded across ranks and ranks don't see duplicated data (and no data is dropped)?
2) Assuming each rank is indeed packing different data as expected. There seems that there could be a potential for `self.packs` to become different size across ranks in some situations? For example, if rank 0 encounters 2 max_seq_len samples, it might pack this into 2, while rank 1 encounters a bunch of smaller samples and packs this into 1. In general I don't see any guarantees that the # of packs are the same across ranks.

torchtune/datasets/_packed.py
RdoubleA: it won't pack the entire dataset into 2 packs, it will just create 2 packs with whatever samples possible and then drop the rest.

`max_packs` may become more relevant for iterable datasets, but maybe it's not as useful right now. Will revisit this.

torchtune/datasets/_packed.py
RdoubleA: mm maybe but I like letting all the methods have access to `self.packs`

torchtune/datasets/_packed.py
RdoubleA: as discussed offline, you're right that it may be worth using `DistributedSampler` here so that packing is partitioned across ranks (cc @tcapelle who also suggested this)

torchtune/datasets/_packed.py
RdoubleA: the code will still be ugly, won't it? since I have to assign to each field. and I want to return them as dicts in the batch

RdoubleA:
> There seems that there could be a potential for self.packs to become different size across ranks in some situations? For example, if rank 0 encounters 2 max_seq_len samples, it might pack this into 2, while rank 1 encounters a bunch of smaller samples and packs this into 1. In general I don't see any guarantees that the # of packs are the same across ranks.

This is a great point and something we'll need to consider anyway when we move to iterable datasets. I need to think about this more, @gokulavasan had some suggestions on this so let me connect with him offline

kartikayk (approved):
Thanks for pushing this through!

noforit:
Thank you very much for your contribution. I have a question: So the current implementation still uses PyTorch's Mem-eff SDPA instead of flash_attn_varlen_func, right?
@RdoubleA

noforit:
Thank you very much for your contribution。Another question, I would like to understand whether making correct pos_id has a significant impact, especially with regard to Rotary Position Embedding. After applying correct attention mask between samples, they no longer influence each other. Within a single sample, the positioning under Rotary Position Embedding is relative. For instance, the relative position between token positions 1 and 10 is 9, or between seq_len + 1 and seq_len + 10 is also 9.


image
And I have observed that the results are almost identical.

RdoubleA:
@noforit Yes based on those benchmarks I didn't observed a significant difference between packing with relative position encoding vs absolute. Actually, I realized some of the numbers in the PR summary are stale. I've just updated them, but also pasting them here:

> Packing with in-sample masking improved fine-tuned models on hellaswag (+2%) compared to packing without in-sample masking, but showed no statistically significant difference on truthfulqa. Relative position encoding showed no statistically significant difference on either benchmark compared to absolute position encoding.

<meta charset="utf-8"><b style="font-weight:normal;" id="docs-internal-guid-1e6883e3-7fff-3297-f266-703a5be3ee0d"><div dir="ltr" style="margin-left:0pt;" align="center">
run | truthfulqa_mc2 (+/- stderr) | hellaswag (+/- stderr)
-- | -- | --
llama3_alpaca_unpacked | 0.452 +/- 0.016 | 0.613 +/- 0.005
llama3_alpaca_packed | 0.463 +/- 0.015 | 0.605 +/- 0.005
llama3_alpaca_packed_no_mask | 0.454 +/- 0.015 | 0.587 +/- 0.005
llama3_alpaca_packed_no_pos | 0.461 +/- 0.015 | 0.605 +/- 0.005
llama3_alpaca_packed_no_mask_no_pos | 0.462 +/- 0.015 | 0.588 +/- 0.005

</div></b>

<meta charset="utf-8"><b style="font-weight:normal;" id="docs-internal-guid-96d52aef-7fff-0158-5348-fa39866b69fc"><div dir="ltr" style="margin-left:0pt;" align="center">
one tailed t-test | truthfulqa_mc2 | hellaswag
-- | -- | --
packed > packed no mask | p = 0.3465 | p = 0.00425
packed > packed no mask no pos | p = 0.4766 | p = 0.0057
packed > packed no pos | p = 0.46645 | p = 0.477
packed > unpacked | p = 0.3113 | p = 0.2846 (packed < unpacked)

</div></b>

So in-sample masking did improve on hellaswag by 2%, but relative vs absolute position encoding did not show any difference. I will point out though that it's difficult to make general claims to prove the negative because there's some limitations to this analysis, namely: 1) fine-tuned models were only trained for three epochs only on Alpaca, 2) only evaluated on truthfulqa and hellaswag. My personal take is that if we can prevent cross-contamination of samples and achieve at least equivalent eval/model quality or possibly better, then it's a no-brainer

datalee:
mark

dyang67:
Need more experiments. I am doing this now.

laogonggong847:
Hello，I have a question, when I use packing, if my random seed is not fixed, will the results of two packings on the same machine be inconsistent? If so, what should I do to make sure that the results of my two packings are consistent?

PR #881, opened 2024-04-26
refacto: expose output layer for gemma models

Optimox:
#### Context
- [ ] add a new feature
- [ ] fix a bug
- [ x] minor re-factorization proposal

#### Changelog
This PR proposes a minor change to `GemmaTransformerDecoder` module.
The goal is to make the Gemma architecture closer to the LLama architecture, the idea is simply to expose an `output` layer by creating a `ReversibleEmbedding` module which performs the linear transformation from hidden representation to final vocabulary using the embeddings weights.

This does not change anything for the current proposed recipes which aim to finetune a model to predict the next words. However, being able to access (and bypass) this ReversibleEmbedding layer is very useful when trying to finetune a model on other tasks (token classification, sentence classification etc...), otherwise the hidden states are inaccessible with the current implementation.

As this is a very minor change, I don't know what should be the test plan. I am happy to discuss whether this change is worth being merged or not.

#### Test plan

### DIFF
--- a/torchtune/models/gemma/transformer.py
+++ b/torchtune/models/gemma/transformer.py
@@ -15,6 +15,14 @@ from torchtune.modules import KVCache
 from torchtune.modules.transformer import _get_clones, TransformerDecoderLayer
 
 
+class ReversibleEmbedding(nn.Module):
+    """
+    This is just a very simple module which allows to call independently this module for final output
+    """
+
+    def forward(self, x, emb_weights):
+        return F.linear(x, emb_weights)
+
 class GemmaTransformerDecoder(nn.Module):
     """
     Transformer Decoder derived from the Gemma architecture. A key difference between
@@ -66,7 +74,8 @@ class GemmaTransformerDecoder(nn.Module):
         self.head_dim = head_dim
         self.causal_mask = None
         self.norm_embeddings = norm_embeddings
-
+        self.output = ReversibleEmbedding()
+        
     def setup_caches(self, max_batch_size: int, dtype: torch.dtype) -> None:
         for layer in self.layers:
             layer.attn.kv_cache = KVCache(
@@ -130,5 +139,5 @@ class GemmaTransformerDecoder(nn.Module):
         h = self.norm(h)
 
         # shape: [b, s, v]
-        output = F.linear(h, self.tok_embeddings.weight).float()
+        output = self.output(h, self.tok_embeddings.weight).float()
         return output


RdoubleA:
Thanks for this PR! You are right that Gemma does not expose the output projection in the transformer decoder, which makes it hards to replace it with a linear classifier (as in the [mistral classifier in this PR](https://github.com/pytorch/torchtune/pull/840/files#diff-11f33a8220b4c18ab09dd5a7bb4f6f6c71ad27c0a08821c2928b91395c043a6cR475)).

I think you may need to go a step further and see how you can use the TransformerDecoder class for Gemma and scrap GemmaTransformerDecoder altogether. So, instead of using a normal nn.Linear in the output proj like llama3 does in the [builder](https://github.com/pytorch/torchtune/blob/9a9a3964f03423ea43868131108bd89b9e94fdda/torchtune/models/llama3/_component_builders.py#L107), you could use your ReversibleEmbedding which just does F.linear. Since it doesn't contain any parameters, it should not affect checkpoint loading, but would be good to check...

For testing, I would suggest just running the gemma recipe before your change and after your change and make the sure the loss values look reasonably close. Since you're changing the output layer we should confirm that forward, backward, checkpoint loading, etc all still work the same. In general, we still need to add more testing for Gemma in the CI to make these changes easier to test.

ebsmothers:
So I think this is a tricky change. To add on to what @RdoubleA said, Gemma ties the output projection weight to the token embedding weight. Normally the thing to do here would be build all the usual components and set e.g. `model.output.weight = model.tok_embeddings.weight`. However, for distributed recipes this won't work, because `to_empty` (as used [here](https://github.com/pytorch/torchtune/blob/9a9a3964f03423ea43868131108bd89b9e94fdda/recipes/full_finetune_distributed.py#L303) in our full finetune recipe) will create a new tensor object and break the weight tie.

So a couple questions on this change:

1) Can you confirm that adding the ReversibleEmbeddingLayer will not result in the same problem where we break the weight tying? I know it is not adding any new parameters, but given that `to_empty` operates on `nn.Module`s and you are adding a new `nn.Module` to the model I want to be sure here.
2) I don't fully understand the goal of parametrizing the output projection like this. If Gemma ties the embedding weights to the output projection weights and we break that tie then we don't really have a Gemma architecture anymore. If we instead want to initialize a model with the pretrained Gemma weights but learn a new output projection, why not just create a new checkpoint with randomly-initialized output projection and load it into e.g. a llama2-style builder?

Also I don't think we should try to integrate into the TransformerDecoder class (at least in the form presented here). The reason is that it will require changing the signature of the output layer [here](https://github.com/pytorch/torchtune/blob/9a9a3964f03423ea43868131108bd89b9e94fdda/torchtune/modules/transformer.py#L210) for something that is pretty specific to Gemma (although please let me know if there is a way around this that I'm missing here).

Optimox:
Thank you for your feedback!

This is probably not the best approach then!

RdoubleA:
@Optimox let us know how we can help out with what you're trying to achieve!

Optimox:
I am just trying to access hidden states more easily, to finetune any model with different heads (for classification, regression etc...). My quick fix is enough for me at the moment, I just wanted to share this approach as I'm sure I will not be the only one willing to finetune a Gemma model for a different task than next token generation.

Issue #848, opened 2024-04-23
Testing for Mistral models


SalmanMohammadi:
I would have found it helpful to have tests for mistral while I was working on https://github.com/pytorch/torchtune/pull/840, and I noticed it was mentioned in https://github.com/pytorch/torchtune/issues/691, so I thought I'd work on it :)

Creating this issue to avoid overlap, and to give an opportunity to review my plan. The llama2 tests involve testing each component with reference code pulled from https://github.com/meta-llama/llama/blob/main. There's a similarly lightweight and minimal implementation of Mistral at https://github.com/mistralai/mistral-src/tree/main. I'm going to follow pretty much exactly the same process as each of the `tests/torchtune/models/llama2` and create corresponding scripts for testing each Mistral component.

PR #888, opened 2024-04-26
Mistral testing

SalmanMohammadi:
#### Context
- [ ] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

https://github.com/pytorch/torchtune/issues/848

#### Changelog

I've started adding scripts to verify the implementation of `mistral`. I'm using the reference implementation from the [official repo](https://github.com/mistralai/mistral-src/blob/main/one_file_ref.py). There's [another implementation](https://github.com/mistralai/mistral-src/blob/main/mistral/model.py) in the repo which uses `xformers` for the attention mechanism, but it's [not straightforward to replicate](https://github.com/mistralai/mistral-src/issues/55). I ended up running into lots of issues when I initially tried.

So far, I've added a script to compare the attention implementation. I've verified the attention implementation produces consistent ouputs using `python -m tests.torchtune.models.mistral.scripts.compare_attention`. I'll be keeping the reference implementation in `tests/torchtune/models/mistral/scripts/mistral_reference.py`.

### Next steps

I'm generally following [this process](https://github.com/pytorch/torchtune/pull/840#discussion_r1579801172) - the plan is to continue copying and testing the components of the mistral implementation, and then testing models as a whole and implementing mapping `torchune.models.mistral` into the reference implementation. Finally, I'll add unit tests to integrate into CI.

Good to make sure I'm not too far off the mark : )


tests/torchtune/models/test_mistral.py
SalmanMohammadi, line 1: Ignore this - I was just drafting the unit test to work back from :)

SalmanMohammadi:
I've updated scripts for the rest of the mistral components. I need to write the comparison involving mapping state dicts, update the unit test, and (potentially) add LoRA comparisons.

SalmanMohammadi:
Okay, all seems good. We now have a unit test for the base `mistral` model using the copied implementation from [the mistral repo](https://github.com/mistralai/mistral-src/blob/147c4e68279b90eb61b19bdea44e16f5539d5a5d/one_file_ref.py).

For the unfortunate reviewer seeing my +1160 line PR (I hope you read this first!):
I'm hoping the `mistral/scripts/compare_{component}.py` for the individual components weren't unnecessary - I'm realising that we already test each component in `llama2/scripts/compare_{component}.py`. Maybe it's good that they've been verified with two implementations? If it's not useful I can take them out - `mistral/scripts/compare_mistral.py` is the main one.

tests/torchtune/models/mistral/scripts/compare_mistral.py
```diff
@@ -0,0 +1,186 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
```
SalmanMohammadi, line 1: This is the main file used for comparing implementations.

ebsmothers:
Thanks for all this extensive testing! 

> I'm hoping the mistral/scripts/compare_{component}.py for the individual components weren't unnecessary - I'm realising that we already test each component in llama2/scripts/compare_{component}.py. The only component that I'm testing that wasn't compared in llama2/scripts/ is mistral_mlp.

I think we wanna find the right balance of rigorous testing and maintenance here. So while I don't want your work to be in vain, I wonder if we should just add those comparison scripts that differ nontrivially from the Llama2 ones, and for other components point to the Llama2 ones. So in this case that would mean keep compare_mistral and compare_feedforward (since you mentioned it's not tested under llama2). Then you can add a readme to `tests/torchtune/models/mistral/scripts` (similar to [this one](https://github.com/pytorch/torchtune/blob/aa650129bcfa08771ba76ad9bdfe7dafc48daffe/tests/torchtune/models/llama2/scripts/README.md) in the llama2 scripts directory) and state that components X, Y, and Z are identical to the llama2 ones and their comparison scripts can be found in that directory. (If you want you can even move the MLP comparison under llama2 so that everything is colocated, but tbh I have no strong preference here.)

SalmanMohammadi:
not in vain at all - I learnt lots! I've updated and added a README.

tests/torchtune/models/mistral/scripts/README.md
```diff
@@ -0,0 +1,13 @@
+## Verifying correctness
+This directory compares the current implementation of `mistral` to the reference implementation at https://github.com/mistralai/mistral-src/blob/main/one_file_ref.py. Additionally, `torchtune.models.mistral._component_builders import mistral_mlp` is compared in `tests.torchtune.models.mistral.scripts.compare_feed_forward.py`
```
ebsmothers, line 2: nit
```suggestion
This directory compares the current implementation of `mistral` to the reference implementation at https://github.com/mistralai/mistral-src/blob/main/one_file_ref.py. Additionally, `torchtune.models.mistral._component_builders.mistral_mlp` is compared in `tests.torchtune.models.mistral.scripts.compare_feed_forward.py`
```

tests/torchtune/models/mistral/scripts/compare_mistral.py
```diff
@@ -0,0 +1,186 @@
...
+        dim=embed_dim,
+        n_kv_heads=num_kv_heads,
+        hidden_dim=intermediate_dim,
+        max_seq_len=max_seq_len,
+        rope_base=rope_base,
+        norm_eps=norm_eps,
+    )
+    # fixed_init_model(ref_mistral_model)
+    # for k, v in mistral_model.state_dict().items():
+    #     print(k, v.shape)
```
ebsmothers, line 71: nit: can remove

ebsmothers (commented):
Looks good! Two small nits, with green CI this is good to merge

SalmanMohammadi:
Thanks again for your review @ebsmothers  :)

ebsmothers (approved):

PR #889, opened 2024-04-27
Feat: Add support of multiple datasets in config

EvilFreelancer:
#### Context

- [x] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

#### Changelog

I've added the ability to use multiple sources for any types of datasets.

After the merger of this PR, users of TorchTune will be able to pass multiple datasets in different formats. For example, it will be possible to mix chat and instruct datasets using different templates, splits, etc.

Example of the new version of the config:

```yml
dataset:
  - _component_: torchtune.datasets.instruct_dataset
    source: tatsu-lab/alpaca
    template: AlpacaInstructTemplate
    split: train
    train_on_input: True
  - _component_: torchtune.datasets.chat_dataset
    source: Open-Orca/SlimOrca-Dedup
    conversation_style: sharegpt
    chat_format: Llama2ChatFormat
    max_seq_len: 1024
    split: train
seed: null
shuffle: True
```

For backward compatibility, users can continue using the previous format of the `dataset` field if they do not wish to use multiple datasets:

```yaml
dataset:
  _component_: torchtune.datasets.instruct_dataset
  source: tatsu-lab/alpaca
  template: AlpacaInstructTemplate
  split: train
  train_on_input: True
```

To run unit tests:

```shell
pytest ./tests/torchtune/datasets/test_multi_dataset.py
```

#### Test plan


EvilFreelancer:
Hi! I've signed CLA, how to rerun failed check?

RdoubleA:
Thanks for this awesome PR @EvilFreelancer! Handling multiple datasets has been a north star for us, as most mature data pipelines for fine-tuning models typically incorporate multiple data sources, so I appreciate you adding this.

> The only one problem is: dataset must have simmilar formats.

This is my main concern - the concatenated datasets must have the same columns AND the same instruct template / chat format. Also, all the keyword arguments will need to be shared (max_seq_len, train_on_input). This is quite restrictive and will require users to do a lot of offline preprocessing work. Ideally, using multiple datasets should be flexible enough that each dataset can have different columns and different template yet we're able to coalesce these together. In the end, all the data gets tokenized as `Message`s, so I do think this is possible.

What are your thoughts on letting `InstructDataset` and `ChatDataset` handle single data sources only, and creating some container class that can hold multiple `InstructDatasets` or `ChatDatasets`? Since the dataset classes return the token IDs ready to be used by the model, the container class can focus on handling the logic of sampling from the list of datasets. 
```
class MultiDataset(torch.utils.data.Dataset):
    def __init__(self, datasets: List[Dataset]):
        self.datasets = datasets

    def __getitem__(self, index: int) -> Tuple[List[int], List[int]]:
        # Figure out how to sample/interleave multiple datasets here
```
Then, in the config, you could specify multiple datasets like this (or something similar):
```
dataset:
  - _component_: torchtune.datasets.instruct_dataset
    source: tatsu-lab/alpaca
    ...
  - _component_: torchtune.datasets.instruct_dataset
    source: vicgalle/alpaca-gpt4
    ...
```
And the recipe can instantiate a MultiDataset if the dataset param is a list. This way, each dataset can keep their individual parameters.

Ideally, we can make use of `concatenate_datasets` and `interleave_datasets` from HF but since we have preprocessing logic in our dataset classes that are specific to each data source, we might have to create similar logic ourselves.

EvilFreelancer:
Hi @RdoubleA, thank you for your response!

Hm, the `MultiDataset` class and the possibility to pass an array of datasets through the `dataset` parameter - it's a great idea, and it's much simpler for end-users to understand how to use this feature than dealing with datasets in similar formats.

I have a couple of spare days, so I may start working on implementing this solution, as it is critical for the project I am working on to have the ability to train on various combinations of datasets.

EvilFreelancer:
Hi! I've implemented logic of [MultiDataset](https://github.com/pytorch/torchtune/pull/889/files#diff-404ae02c70619a365d896837164b187a3b5e89ca5b58a5eb73d3de0c1445559b) and enabled it in traning recipes.

I've also removed original multi-source logic and cleaned tests.

How to use:
```yml
dataset:
  - _component_: torchtune.datasets.instruct_dataset
    source: tatsu-lab/alpaca
    template: AlpacaInstructTemplate
    split: train
    train_on_input: True
  - _component_: torchtune.datasets.instruct_dataset
    source: vicgalle/alpaca-gpt4
    template: AlpacaInstructTemplate
    split: train
    train_on_input: True
  - _component_: torchtune.datasets.chat_dataset
    source: Open-Orca/SlimOrca-Dedup
    conversation_style: sharegpt
    chat_format: Llama2ChatFormat
    max_seq_len: 1024
    split: train
  - _component_: torchtune.datasets.chat_dataset
    source: ajibawa-2023/Code-290k-ShareGPT
    conversation_style: sharegpt
    chat_format: Llama2ChatFormat
    max_seq_len: 1024
    split: train
seed: null
shuffle: True
```

By the way, this gave an interesting side effect: using this logic, you can freely mix datasets of any format.

In the example that I gave, dataset formats `instrcut` and `chat` are mixed, it seems to me that this will be a very interesting and convenient feature of TorchTune.

recipes/full_finetune_distributed.py
```diff
@@ -357,10 +357,12 @@ def _setup_data(
         iterable datasets and streaming datasets are not supported.
         """
         world_size, rank = utils.get_world_size_and_rank()
-        ds = config.instantiate(
-            cfg_dataset,
-            tokenizer=self._tokenizer,
-        )
+
+        if isinstance(cfg_dataset.get(0), DictConfig):
```
RdoubleA, line 361: you can also just directly check if it's a ListConfig. If it's a single dataset then this might fail
```suggestion
        if isinstance(cfg_dataset, ListConfig):
```

I also wonder if there's a better way to handle this so we don't have to repeat this if-else check across all recipes...

tests/torchtune/datasets/test_alpaca_dataset.py
RdoubleA, line 1: Do we still need to make these changes to the test samples to use Dataset across all the individual dataset test files?

torchtune/utils/multi_dataset.py
```diff
@@ -0,0 +1,46 @@
...
+from omegaconf import DictConfig
+from torch.utils.data import Dataset
+from torchtune import config, utils
+from tqdm import tqdm
+
+log = utils.get_logger("DEBUG")
+
+
+class MultiDataset(Dataset):
+    def __init__(self, datasets: DictConfig[DictConfig], tokenizer):
```
RdoubleA, line 18: I'm not sure about passing in the datasets as config objects, because then this class is hard to use without having to create a config. And if you ever want to play around with `MultiDataset` in a notebook environment, this might get tedious. Config objects should be separate from the actual code as much as possible, except in the recipe where components need to be instantiated. But I see that it's hard to create a `MultiDataset` without having to instantiate the individual datasets first...

What do you think about an alternate approach where we instantiate the datasets outside of this class and then pass in the dataset objects?
```
# in recipe
datasets = [config.instantiate(single_cfg_dataset, tokenizer) for single_cfg_dataset in cfg_dataset]
ds = MultiDataset(datasets=datasets)

# Modified MultiDataset constructor - won't need tokenizer anymore
def __init__(self, datasets: List[Dataset]):
```

torchtune/utils/multi_dataset.py
```diff
@@ -0,0 +1,46 @@
...
+        self._datasets = datasets
+        self._tokenizer = tokenizer
+        self._data = []
+
+        # Load all datasets one by one
+        for dataset in datasets:
+            self._data.extend(self._loading_dataset(dataset))
```
RdoubleA, line 25: This means that all the data will iterate sequentially, right? What if I wanted to interleave the datasets and jump around during training instead of exhausting one dataset at time? Would the `DistributedSampler` handle this?

torchtune/utils/multi_dataset.py
```diff
@@ -0,0 +1,46 @@
...
+        """
+        Loads the dataset from provided omegaconf object, and returns it as a list.
+        Each element in the list is a tuple of tokens and labels.
+        """
+        log.info(f"Loading dataset {dataset.source}")
+        items = config.instantiate(dataset, tokenizer=self._tokenizer)
+
+        output = []
+        for tokens, labels in tqdm(items):
+            output.append((tokens, labels))
```
RdoubleA, line 45: this is nice because it will do all the templating and tokenization for all your datasets before we even start training so training might be a bit faster, but I imagine users might want to have control over this, in case they want to do the sample preparation / templating on the fly instead (like when you're just debugging things or you have very large datasets, you don't want to process all the datasets in their entirety every time)

RdoubleA (commented):
I like the simplicity of the `MultiDataset` class you implemented, very easy to understand! There's some considerations around how we instantiate each individual dataset we should figure out.

recipes/full_finetune_distributed.py
EvilFreelancer: Good point, i've replaced it with `ListConfig` check.

torchtune/utils/multi_dataset.py
EvilFreelancer: Clever idea, patch added.

torchtune/utils/multi_dataset.py
EvilFreelancer: Yeah, you right, i've changed this logic a new one, please refer MultiDataset class.

torchtune/utils/multi_dataset.py
EvilFreelancer: Indeed, replaced to kind of `sample preparation`.

tests/torchtune/datasets/test_alpaca_dataset.py
EvilFreelancer: I think we need to keep the `datasets.Dataset` objects, since this is a common format and anyone can see that the data should come in this format, plus there is no need to perform any extra transformations, as was the case, for example in `_chat` tests.

EvilFreelancer:
@RdoubleA hi! Thanks for your review, requested fixes added.

By the way I've noticed one small thing, for example in `vicgalle/alpaca-gpt4` dataset have 52k rows, why TorchTune shows me 26k items on training stage? Maybe it's because train/val splits?

recipes/full_finetune_distributed.py
EvilFreelancer: > I also wonder if there's a better way to handle this so we don't have to repeat this if-else check across all recipes...

Yeah, this logic can be moved to `config.instantiate` method, but I guess it will break single responsibility principle.

torchtune/utils/multi_dataset.py
```diff
@@ -0,0 +1,34 @@
...
+        for dataset in self._datasets:
+            # Let's check if the index is in the dataset
+            if cumulative_index + len(dataset) > index:
+                # If yes, then prepare the sample and return it
+                return dataset[index - cumulative_index]  # noqa
+            # Otherwise, go to the next dataset
+            cumulative_index += len(dataset)
+
+    def __len__(self) -> int:
+        return sum(len(sublist) for sublist in self._datasets)
```
RdoubleA, line 34: let's just store this on init so we don't have to calculate it every time

torchtune/utils/multi_dataset.py
```diff
@@ -0,0 +1,34 @@
...
+log = utils.get_logger("DEBUG")
+
+
+class MultiDataset(Dataset):
+    def __init__(self, datasets: List[List[Dataset]]):
+        self._datasets = datasets
+
+    def __getitem__(self, index: int) -> Tuple[List[int], List[int]]:
+        if index < 0 or index >= self.__len__():
+            raise IndexError("Index out of range")
```
RdoubleA, line 22: hm, will this cause negative indexing into MultiDataset to fail? but if you support negative indexing then it might complicate the logic below. Maybe it's fine to restrict it.

torchtune/utils/multi_dataset.py
```diff
@@ -0,0 +1,34 @@
...
+
+log = utils.get_logger("DEBUG")
+
+
+class MultiDataset(Dataset):
+    def __init__(self, datasets: List[List[Dataset]]):
+        self._datasets = datasets
+
+    def __getitem__(self, index: int) -> Tuple[List[int], List[int]]:
+        if index < 0 or index >= self.__len__():
```
RdoubleA, line 21: nit: reads a bit better
```suggestion
        if index < 0 or index >= len(self):
```

RdoubleA (commented):
This is turning into something really nice! Thanks for your patience working through this. Could you add a unit test for `MultiDataset` on some toy data?

Also tagging @ebsmothers, @kartikayk to get their thoughts on how we should handle instantiating this in the recipes. I think we could just make every dataset a `MultiDataset`, since you could just pass a list with a single dataset and it should work the same (correct me if I'm wrong here). That should simplify the logic in the recipes.

torchtune/utils/multi_dataset.py
EvilFreelancer: Thanks, fixed

torchtune/utils/multi_dataset.py
EvilFreelancer: I guess you right, it's probably not possible to out of range, i've removed this check.

EvilFreelancer:
@RdoubleA hi! I've added fixes you mentioned, and couple simple test on `MultiDataset` class.

What do you think about:

```python
if not isinstance(cfg_dataset, ListConfig):
    cfg_dataset = [cfg_dataset]
datasets = [config.instantiate(cfg_item, tokenizer=self._tokenizer) for cfg_item in cfg_dataset]
ds = utils.MultiDataset(datasets=datasets)
```

instead of:

```
    datasets = [
        config.instantiate(single_cfg_dataset, tokenzier=self._tokenizer)
        for single_cfg_dataset in cfg_dataset
    ]
    ds = utils.MultiDataset(datasets=datasets)
else:
    ds = config.instantiate(cfg_dataset, tokenizer=self._tokenizer)
```

RdoubleA:
@EvilFreelancer Thanks for the updates! I'll take another pass soon.

In the meantime, since this is a significant feature we are adding, I'd like to make sure this is rigorously tested. Do you mind confirming the following:
- The unit test for multidataset passes (please add the test command to the PR summary)
- Run one of the distributed recipes and confirm that loss curves and tokens/sec look reasonable (you can use WandBLogger for easy visualization). Main thing I want to confirm here is that we can still sample from multiple datasets correctly in a distributed environment
- Run one of the single device recipes for a similar reason above

Also, I forgot to mention this earlier, but I think a better location for `MultiDataset` would be in `torchtune/datasets` instead of `torchtune/utils`, what do you think?

torchtune/utils/multi_dataset.py
```diff
@@ -0,0 +1,39 @@
...
+
+from torch.utils.data import Dataset
+
+from torchtune import utils
+
+log = utils.get_logger("DEBUG")
+
+
+class MultiDataset(Dataset):
+    def __init__(self, datasets: List[List[Dataset]]):
```
ebsmothers, line 17: Why List[List[ here?

torchtune/utils/multi_dataset.py
```diff
@@ -0,0 +1,39 @@
...
+        self._datasets = datasets
+        self._len = sum(len(dataset) for dataset in datasets)
+        self._indexes = []
+
+        # Calculate distribution of indexes in all datasets
+        cumulative_index = 0
+        for idx, dataset in enumerate(datasets):
+            next_cumulative_index = cumulative_index + len(dataset)
```
ebsmothers, line 25: Not a blocker for this PR, but worth thinking about how we're gonna do this for iterable datasets @RdoubleA @SLR722

torchtune/utils/multi_dataset.py
```diff
@@ -0,0 +1,39 @@
...
+            cumulative_index = next_cumulative_index
+
+        log.debug(f"Datasets summary length: {self._len}")
+        log.debug(f"Datasets indexes: {self._indexes}")
+
+    def __getitem__(self, index: int) -> Tuple[List[int], List[int]]:
+        for start, stop, dataset_index in self._indexes:
+            if start <= index < stop:
+                dataset = self._datasets[dataset_index]
+                return dataset[index - start]  # noqa
```
ebsmothers, line 36: why is noqa needed here?

torchtune/utils/multi_dataset.py
EvilFreelancer: Hi, oops, it's a typo from the first iterations of the class, changed to `List[Dataset]` as it should be, thanks!

torchtune/utils/multi_dataset.py
EvilFreelancer: In first iteration of MultiDataset on this line was a type check error, but now noqa is not required. Thanks for noticing.

EvilFreelancer:
@joecummings hi!

> Can you update the README with examples on how this will now look in the YAML config?

I've [added](https://github.com/pytorch/torchtune/pull/889/files#diff-6f46eaee6237147c156aa798061cbdc81ed1f598387e5d059eb9306132e7ba8c) small section to dataset page in tutorial.

EvilFreelancer:
The tests for Multi GPU Gemma 2B training are available [here](https://api.wandb.ai/links/miteigi-nemoto/ewedo1mb).

RdoubleA:
@EvilFreelancer Thanks for sharing the test runs! Sorry for the delay in following up on this. An update from our side is that we've been having a lot of discussions around iterable datasets, which is something that we want to support more and design around moving forward. This would enable more flexibility around interleaving, weighted sampling, etc. for multiple datasets that won't fit in memory.

That being said, I still think the `MultiDataset` here is valuable for datasets that fit in memory and can still leverage map-style functionality. This will unblock users that want to quickly concatenate multiple data sources until we add iterable datasets and a more powerful `MultiDataset`. So to accurately reflect the scope of this class, my suggestion is to rename this to `ConcatDataset` (you might've started with this name to begin with, I apologize) and make it clear in the docstrings that this is for map datasets that all fit in memory.

Let me know if this makes sense!

torchtune/datasets/_multi.py
```diff
@@ -0,0 +1,39 @@
...
+from typing import List, Tuple
+
+from torch.utils.data import Dataset
+
+from torchtune import utils
+
+log = utils.get_logger("DEBUG")
+
+
+class MultiDataset(Dataset):
```
RdoubleA, line 16: Let's add a comprehensive docstring here

EvilFreelancer: A comprehensive docstring added.

docs/source/tutorials/datasets.rst
```diff
@@ -47,6 +47,30 @@ You could tweak :code:`max_seq_len` to achieve that directly from the config.
...
+      train_on_input: True
+    - _component_: torchtune.datasets.instruct_dataset
+      source: samsum
+      template: SummarizeTemplate
+      column_map: {"output": "summary"}
+      split: train
+      train_on_input: False
+
+This configuration demonstrates how to seamlessly integrate two datasets from different
+formats for comprehensive training.
```
ebsmothers, line 72: Can we remove this from the live docs? I feel like there are too many subtleties here that are not immediately clear. E.g. a very natural next question based on this is "how can I sample from different datasets with different probabilities?", which is not supported. So I think it's better to leave it out until we have a full feature set, then we can showcase the functionality properly and in more detail.

recipes/full_finetune_distributed.py
```diff
@@ -357,10 +357,16 @@ def _setup_data(
...
-        ds = config.instantiate(
-            cfg_dataset,
-            tokenizer=self._tokenizer,
-        )
+
+        if isinstance(cfg_dataset, ListConfig):
+            datasets = [
+                config.instantiate(single_cfg_dataset, tokenizer=self._tokenizer)
+                for single_cfg_dataset in cfg_dataset
+            ]
```
ebsmothers, line 365: Not necessary for this PR, but it'd be nice to have a utility like 

```
def instantiate_list_config(cfg: ListConfig, **kwargs) -> List[Any]:
    return [
        instantiate(element, kwargs) for element in cfg
    ]
```

to reduce copypasta across the recipes a bit. cc @RdoubleA to tackle in a follow-up

tests/torchtune/datasets/test_alpaca_dataset.py
```diff
@@ -28,17 +29,19 @@ def test_label_no_masking(self, load_dataset, tokenizer):
...
-                "instruction": "Give three tips for staying healthy.",
-                "input": "",
-                "output": (
-                    "1.Eat a balanced diet and make sure to include plenty of fruits and vegetables."
-                    "2. Exercise regularly to keep your body active and strong."
-                    "3. Get enough sleep and maintain a consistent sleep schedule."
-                ),
-            }
-        ]
+        load_dataset.return_value = Dataset.from_list(
```
ebsmothers, line 32: What's the story with these `Dataset.from_list` changes? I know it will give us the right return type (`Dataset` instead of raw `List`), anything besides that motivating the change? (I am fine with keeping them in, mainly asking out of curiosity)

RdoubleA: Technically it's a more accurate return type of load dataset, I think it's ok to just leave these out and stick to primitives for simplicity, but I don't have a strong opinion here.

docs/source/tutorials/datasets.rst
EvilFreelancer: Maybe better just to change it to something like this?

```yml
  # Dataset
  dataset:
      max_seq_len: 256
    - _component_: torchtune.datasets.samsum_dataset
```

tests/torchtune/datasets/test_alpaca_dataset.py
EvilFreelancer: I've invested considerable time into understanding how to test my new dataset class. It was initially unclear regarding the required format and content of its elements. Hence, I suggest providing clear guidelines to save fellow programmers time, elucidating the expected format for dataset elements.

ebsmothers: Thanks, this is very good feedback. I agree that the contracts of various dataset components are not always obvious and take some time to sort through. Aside from improving live docs and better code comments, I'm open to any suggestions you have on how to make this clearer based on your experience.

docs/source/tutorials/datasets.rst
ebsmothers: So actually my comment is less on the config itself, but more on the way that we frame our multi-dataset support. Basically I want to make sure we are explicit about what's supported and what's not. So while the ability to configure each individual dataset separately does imply some flexibility, I would prefer not to focus on flexibility as much as you have in this particular snippet, as it can easily imply sampling strategies etc. that we do not (yet) support. Instead it'd be good to clearly state that we are concatenating the datasets when built in this way. How about something like this?

```suggestion
It is also possible to train on multiple datasets by combining them into a single :class:`~torchtune.datasets.ConcatDataset`. For example:
.. code-block:: yaml

  dataset:
    - _component_: torchtune.datasets.instruct_dataset
      source: vicgalle/alpaca-gpt4
      template: AlpacaInstructTemplate
      split: train
      train_on_input: True
    - _component_: torchtune.datasets.instruct_dataset
      source: samsum
      template: SummarizeTemplate
      column_map: {"output": "summary"}
      split: train
      train_on_input: False

The preceding snippet demonstrates how you can configure each individual dataset's parameters, then combine them into a single concatenated dataset for training.
```

(Note: for the above suggestion to work you will also need to add `ConcatDataset` to [api_ref_datasets.rst](https://github.com/pytorch/torchtune/blob/9274c8989cfa75feb8804a8e15b5c8ba9d71ccdb/docs/source/api_ref_datasets.rst#L4))

EvilFreelancer: I like it!

Thank you for pointing out `api_ref_datasets.rst`, I will address it promptly.

EvilFreelancer: `ConcatDataset` added to `api_ref_datasets.rst` as you suggested.

ebsmothers:
Ah crap one more thing: my suggestion screwed up the rendering of the multi-dataset YAML. Sorry about that! I think you just need to add back a newline before the codeblock statement.

Screenshot 2024-05-03 at 2 35 18 PM

EvilFreelancer:
@ebsmothers typo fixed

ebsmothers (approved):
Looks great, thank you for adding this!

EvilFreelancer:
Thank you for the meticulous code-review and excellent advice, I thoroughly enjoyed working on this project with all of you!

PR #903, opened 2024-04-30
total_training_steps -> global_step

tcapelle:
That's it, renaming the step counter to `global_step`. Why?
- It makes more sense and reads better
- It is named `global_step` on the HF Trainer, make using torchtune familiar to folks coming from there
- if we add evals during training (at the end of an epoch for instance) this would make more sense that keeping a "training" naming


joecummings (approved):
Will land once CI is green

PR #909, opened 2024-05-01
[WIP] enable QLoRA + FSDP2

weifengpy:
this PR is stacked on LoRA + FSDP2: https://github.com/pytorch/torchtune/pull/855

they share the components
* loading HF checkpoint
* FSDP wrapping
* saving checkpoint
* load checkpoint to resume training

the major difference is in config
* `model._component_`: `torchtune.models.llama2.qlora_llama2_7b` instead of `lora_llama2_7b`. `LoRALinear(quantize_base=True/False)`

### DIFF
--- a/torchtune/utils/_distributed.py
+++ b/torchtune/utils/_distributed.py
@@ -14,14 +14,15 @@ import torch
 import torch.distributed as dist
 import torch.distributed._composable.fsdp
 from torch import nn
-from torch.distributed._tensor import distribute_tensor, DTensor
+from torch.distributed._tensor import distribute_tensor
 from torch.distributed.fsdp import (
     FullyShardedDataParallel as FSDP,
     MixedPrecision,
     ShardingStrategy,
 )
 from torch.distributed.fsdp.wrap import ModuleWrapPolicy
-from torch.optim import Optimizer
+
+# from torch.optim import Optimizer
 from torchtune.modules.peft.lora import (
     _lora_a_init_params,
     _lora_b_init_params,
@@ -301,6 +302,8 @@ def lora_fsdp_wrap_policy(modules_to_wrap: Set[Type]) -> FSDPPolicyType:
     return lora_wrap_fsdp
 
 
+# remove if torch.distributed.checkpoint
+# implements rank0 load and broadcoast
 def load_from_full_state_dict(
     model: torch.distributed._composable.fsdp.FSDP,
     full_sd: Dict[str, Any],
@@ -327,41 +330,28 @@ def load_from_full_state_dict(
     model.load_state_dict(sharded_sd, strict=False, assign=True)
 
 
-def get_full_model_state_dict(
-    model: torch.distributed._composable.fsdp.FSDP,
-    is_rank_zero: bool,
-) -> Dict[str, Any]:
-    sharded_sd = model.state_dict()
-    cpu_state_dict = {}
-    for param_name, sharded_param in sharded_sd.items():
-        full_param = sharded_param.full_tensor()
-        if is_rank_zero:
-            cpu_state_dict[param_name] = full_param.cpu()
-        else:
-            del full_param
-    return cpu_state_dict
-
-
-def get_full_optimizer_state_dict(
-    opt: Optimizer,
-    is_rank_zero: bool,
-) -> Dict[str, Any]:
-    sharded_sd = opt.state_dict()
-    sharded_state = sharded_sd["state"]
-    full_state = {}
-    for group_id, sharded_group in sharded_state.items():
-        group_state = {}
-        for attr, sharded_tensor in sharded_group.items():
-            if isinstance(sharded_tensor, DTensor):
-                full_tensor = sharded_tensor.full_tensor()
-            else:
-                full_tensor = sharded_tensor
-            if is_rank_zero:
-                group_state[attr] = full_tensor.cpu()
-            else:
-                del full_tensor
-        full_state[group_id] = group_state
-    return {
-        "param_groups": sharded_sd["param_groups"],
-        "state": full_state,
-    }
+# remove if torch.distributed.checkpoint
+# implements rank0 load and broadcoast
+# def get_full_optimizer_state_dict(
+#     opt: Optimizer,
+#     is_rank_zero: bool,
+# ) -> Dict[str, Any]:
+#     sharded_sd = opt.state_dict()
+#     sharded_state = sharded_sd["state"]
+#     full_state = {}
+#     for group_id, sharded_group in sharded_state.items():
+#         group_state = {}
+#         for attr, sharded_tensor in sharded_group.items():
+#             if isinstance(sharded_tensor, DTensor):
+#                 full_tensor = sharded_tensor.full_tensor()
+#             else:
+#                 full_tensor = sharded_tensor
+#             if is_rank_zero:
+#                 group_state[attr] = full_tensor.cpu()
+#             else:
+#                 del full_tensor
+#         full_state[group_id] = group_state
+#     return {
+#         "param_groups": sharded_sd["param_groups"],
+#         "state": full_state,
+#     }
--- a/torchtune/utils/_distributed.py
+++ b/torchtune/utils/_distributed.py
@@ -14,19 +14,19 @@ import torch
 import torch.distributed as dist
 import torch.distributed._composable.fsdp
 from torch import nn
-from torch.distributed._tensor import distribute_tensor
+from torch.distributed._tensor import distribute_tensor, DTensor
 from torch.distributed.fsdp import (
     FullyShardedDataParallel as FSDP,
     MixedPrecision,
     ShardingStrategy,
 )
 from torch.distributed.fsdp.wrap import ModuleWrapPolicy
+from torch.optim import Optimizer
 from torchtune.modules.peft.lora import (
     _lora_a_init_params,
     _lora_b_init_params,
     LoRALinear,
 )
-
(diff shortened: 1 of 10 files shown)


tests/torchtune/utils/test_distributed.py
```diff
@@ -245,3 +251,160 @@ def test_lora_meta_device_init_fsdp(self):
...
+
+        # inp is different for each rank
+        torch.manual_seed(42 + self.rank)
+
+        # test get full state dict
+        for _ in range(epochs):
+            inp = torch.randn((2, mlp_dim), device="cuda")
+            base_model(inp).sum().backward()
+            for param in base_model.parameters():
+                torch.distributed.all_reduce(param.grad)
```
awgu, line 294: nit: 😄 
```suggestion
                torch.distributed.all_reduce(param.grad, op=ReduceOp.AVG)
```

`from torch.distributed.distributed_c10d import ReduceOp`

weifengpy: neat!

weifengpy: updated LoRA PR to use `ReduceOp.AVG`. This PR will be stacked on it when landing

torchtune/modules/low_precision/_register_nf4_dispatch_ops.py
```diff
@@ -17,26 +17,3 @@ def clone(func, *args, **kwargs):
     in precision.
     """
     return to_nf4(args[0][0].get_original_weight())
-
-
```
weifengpy, line 21: starting from TorchAO==0.2.0, we implemented `NF.copy_`. It's the superset of `inplace_copy`. We cover both `NF4.copy_(bf16)` and `NF4.copy_(NF4)`

recipes/configs/dev/llama2/7B_qlora.yaml
```diff
@@ -0,0 +1,91 @@
...
+  lr: 3e-4
+lr_scheduler:
+  _component_: torchtune.modules.get_cosine_schedule_with_warmup
+  num_warmup_steps: 100
+
+loss:
+  _component_: torch.nn.CrossEntropyLoss
+
+fsdp:
+  cpu_offload: False
```
weifengpy, line 65: comparing with `7B_qlora_single_device.yaml`, `cpu_offload` is the new feature

recipes/configs/dev/llama2/7B_qlora.yaml
```diff
@@ -0,0 +1,91 @@
+# Config for single device QLoRA with lora_finetune_single_device.py
+# using a Llama2 7B model
+#
+# This config assumes that you've run the following command before launching
+# this run:
+#   tune download meta-llama/Llama-2-7b-hf --output-dir /tmp/Llama-2-7b-hf --hf-token <HF_TOKEN>
+#
+# To launch on a single device, run the following command from root:
+#   tune run lora_finetune_single_device --config llama2/7B_qlora_single_device
```
weifengpy, line 9: will replace `lora_finetune_single_device` to `lora_finetune_fsdp2` before landing. saving CI to publish PR earlier now

recipes/dev/lora_finetune_fsdp2.py
```diff
@@ -290,25 +291,6 @@ def _setup_model(
         with utils.set_default_dtype(self._dtype), torch.device("meta"):
             model = config.instantiate(cfg_model)
 
-        if self._is_rank_zero:
-            # The model contains LoRA params which won't have any matching keys in
-            # the state dict. As a result, we need to load with strict=False.
-            # Before loading the state dict, ensure the state dict keys for the base
-            # model and adapters (if available) match the keys in the full LoRA model
-            # This is a good sanity check to prevent silent errors
-            validate_state_dict_for_lora(
```
weifengpy, line 299: `validate_state_dict_for_lora` needs some refactor since FSDP2 use clean FQNs. 
FSDP2 also check param on meta and throw Runtime error https://fburl.com/wrhxykn3

```
"FSDP parameters should be materialized from meta device before training, "
f"but the following were still on meta device: {param_names_on_meta}\n"
"For example, call module.to_empty(device) to materialize to device and "
"call module.reset_parameters() on each module to initialize values."
```

recipes/dev/lora_finetune_fsdp2.py
```diff
@@ -590,6 +577,8 @@ def train(self) -> None:
                 logits = logits.transpose(1, 2)
                 # Compute loss
                 loss = self._loss_fn(logits, labels)
+                # free logits otherwise it peaks backward memory
+                del logits
```
weifengpy, line 581: `del logits` saves 6% peak memory (baseline has `cpu_offload=True`). see https://github.com/pytorch/torchtune/pull/1046

recipes/configs/dev/llama2/7B_qlora.yaml
ebsmothers: Strictly speaking you probably do not need to put `cpu_offload` config under `fsdp` (unless you think it helps for clarity, or anticipate other kwargs that we'd wanna pass through to `fully_shard`). No strong preference either way here though

torchtune/utils/_distributed.py
```diff
@@ -280,12 +281,39 @@ def load_from_full_model_state_dict(
...
+                sharded_meta_param.device_mesh,
+                sharded_meta_param.placements,
+                shape=sharded_meta_param.size(),
+                dtype=sharded_meta_param.dtype,
+                requires_grad=sharded_meta_param.requires_grad,
+                stride=sharded_meta_param.stride(),
+            )
+        else:
+            sharded_tensor = distribute_tensor(
+                full_tensor,
```
ebsmothers, line 313: No longer need to convert dtype as before?

recipes/configs/dev/llama2/70B_qlora.yaml
```diff
@@ -0,0 +1,90 @@
...
+# This config assumes that you've run the following command before launching
+# this run:
+#   tune download meta-llama/Llama-2-70b-hf --output-dir /tmp/Llama-2-70b-hf --hf-token <HF_TOKEN>
+#
+# This config needs 8 GPUs to run
+#   # tune run --nproc_per_node 8 lora_finetune_fsdp2 --config llama2/70B_qlora
+#
+
+# Model Arguments
+model:
```
weifengpy, line 13: comparing with `70B_lora.yaml`, added `output_proj`, `apply_lora_to_mlp: True`

torchtune/utils/_distributed.py
```diff
@@ -280,12 +281,39 @@ def load_from_full_model_state_dict(
...
+            sharded_param[: chunk.size(0)].copy_(chunk)
+            sharded_tensor = DTensor(
+                sharded_param,
+                sharded_meta_param.device_mesh,
+                sharded_meta_param.placements,
+                shape=sharded_meta_param.size(),
+                dtype=sharded_meta_param.dtype,
+                requires_grad=sharded_meta_param.requires_grad,
+                stride=sharded_meta_param.stride(),
+            )
```
ebsmothers, line 310: If it's not too much work, is it possible to extend the unit test you added in #855 to cover this case as well? As is I find it a bit hard to follow and want to make sure we have a reliable sanity check in case anything breaks in the future

weifengpy: you're right. will try to come up with a unit test to cover NF4

torchtune/utils/_distributed.py
weifengpy: it's moved to line 284 `full_tensor = full_tensor.to(sharded_meta_param.dtype).to(device)`. The extra `.to(device)` is for NF4 since quantization on cpu is prohibitively slow

recipes/configs/dev/llama2/7B_qlora.yaml
weifengpy: I know Less has been working on `cpu_offload` for activations, which might be a root config

another things is, other `fully_shard` kwargs like `reshard_after_forward=False` can boost QPS

Does it make sense to you?

recipes/configs/dev/llama2/7B_qlora.yaml
```diff
@@ -0,0 +1,91 @@
+# Config for single device QLoRA with lora_finetune_single_device.py
```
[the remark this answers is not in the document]
weifengpy: good point! updating now

torchtune/utils/_distributed.py
weifengpy: @ebsmothers see `pytest -s tests/torchtune/utils/test_distributed.py -k test_qlora_state_dict`

recipes/dev/lora_finetune_fsdp2.py
weifengpy: `validate_missing_and_unexpected_for_lora` works out of the box. added it in this PR. finger acrossed!

recipes/configs/dev/llama2/7B_qlora.yaml
ebsmothers: Ah yeah this is a good point, makes sense to me. Thanks for clarifying

ebsmothers (approved):
This is looking good! As discussed offline we can separate out the checkpoint save support into a follow-up PR to unblock other ongoing work. Once CI is green this one should be good to land. Thank you for enabling a huge huge feature for our users!

weifengpy:
merged after confirming CI's all green. Will follow up on saving state dict

PR #911, opened 2024-05-01
Loading mistral reward model checkpoints

SalmanMohammadi:
See https://github.com/pytorch/torchtune/issues/812#issuecomment-2067790583 for context.

#### Context
- [x] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

#### Changelog

I've added some lightweight logic in `torchtune/utils/_checkpointing/_checkpointer.py` to support loading a mistral reward model [from huggingface](https://github.com/huggingface/transformers/blob/1e05671d213642a67dd418c13a18a17a18c25117/src/transformers/models/mistral/modeling_mistral.py#L1282), and added a test case to ensure model weights are converted correctly.

#### Test plan

### DIFF
--- a/torchtune/models/convert_weights.py
+++ b/torchtune/models/convert_weights.py
@@ -176,7 +176,7 @@ def tune_to_hf(
         dim (int): Dimension of the model.
 
     Returns:
-        Dict[str, torch.Tensor]: State dict in Meta's format.
+        Dict[str, torch.Tensor]: State dict in HF's format.
     """
     converted_state_dict = {}
     inverted_mapping_dict = {v: k for k, v in _FROM_HF.items()}
--- a/torchtune/models/convert_weights.py
+++ b/torchtune/models/convert_weights.py
@@ -126,7 +126,7 @@ def hf_to_tune(
     repo in HF (https://huggingface.co/meta-llama/Llama-2-7b-hf).
 
     Args:
-        state_dict (Dict[str, torch.Tensor]): State dict in Meta's format.
+        state_dict (Dict[str, torch.Tensor]): State dict in HF's format.
         num_heads (int): Number of heads in the model.
         num_kv_heads (int): Number of heads in the key/value projection layers.
         dim (int): Dimension of the model.
--- a/torchtune/utils/_checkpointing/_checkpointer.py
+++ b/torchtune/utils/_checkpointing/_checkpointer.py
@@ -384,6 +384,14 @@ class FullModelHFCheckpointer(_CheckpointerInterface):
 
         if self._model_type == ModelType.PHI3_MINI:
             converted_state_dict[utils.MODEL_KEY] = phi3_hf_to_tune(merged_state_dict)
+        elif self._model_type == ModelType.MISTRAL_REWARD:
+            merged_state_dict["lm_head.weight"] = merged_state_dict.pop("score.weight")
+            converted_state_dict[utils.MODEL_KEY] = convert_weights.hf_to_tune(
+                merged_state_dict,
+                num_heads=self._config["num_attention_heads"],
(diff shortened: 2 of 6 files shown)


torchtune/utils/_checkpointing/_checkpointer_utils.py
```diff
@@ -24,6 +24,7 @@ class ModelType(Enum):
     LLAMA3 = "llama3"
     MISTRAL = "mistral"
     PHI3_MINI = "phi3_mini"
+    MISTRAL_REWARD = "mistral_reward"
```
SalmanMohammadi, line 27: maybe this could be mistral_classifier?

torchtune/utils/_checkpointing/_checkpointer.py
```diff
@@ -384,6 +384,14 @@ def load_checkpoint(self) -> Dict[str, Any]:
 
         if self._model_type == ModelType.PHI3_MINI:
             converted_state_dict[utils.MODEL_KEY] = phi3_hf_to_tune(merged_state_dict)
+        elif self._model_type == ModelType.MISTRAL_REWARD:
+            merged_state_dict["lm_head.weight"] = merged_state_dict.pop("score.weight")
```
kartikayk, line 388: It does seem very tempting to add this one line change here. But this will start to add up and make this checkpointer really hard to read.

I do need to redesign this component a bit. But in the mean time sn alternative here (granted it's more lines of code with some copy paste) would be to define a custom ```convert_weights.py``` within the mistral folder and define a new ```mistral_reward_hf_to_tune``` and ```mistral_reward_tune_to_hf``` functions which do this mapping (and any future changes you might need for reward modeling). An example is what we did for [```phi3```](https://github.com/pytorch/torchtune/blob/main/torchtune/models/convert_weights.py). As I mentioned, its more copy-paste code, but I can definitely see this function updating as you add more heads etc. And I'm afraid the checkpointer will bloat up because of it.

@ebsmothers let me know what you think about this.

torchtune/utils/_checkpointing/_checkpointer.py
```diff
@@ -426,6 +434,16 @@ def save_checkpoint(
         # convert the state_dict back to hf format; do this inplace
         if self._model_type == ModelType.PHI3_MINI:
             state_dict[utils.MODEL_KEY] = phi3_tune_to_hf(state_dict[utils.MODEL_KEY])
+        elif self._model_type == ModelType.MISTRAL_REWARD:
```
kartikayk, line 437: same comment as above

torchtune/utils/_checkpointing/_checkpointer_utils.py
kartikayk: I think ```mistral_reward``` makes sense unless you think there will be classifier checkpoints we'll need to load?

SalmanMohammadi: I'll leave as-is for now, we can always generalise later if there's use cases.

torchtune/utils/_checkpointing/_checkpointer.py
SalmanMohammadi: I think this makes sense to me.

torchtune/models/convert_weights.py
```diff
@@ -126,7 +126,7 @@ def hf_to_tune(
     repo in HF (https://huggingface.co/meta-llama/Llama-2-7b-hf).
 
     Args:
-        state_dict (Dict[str, torch.Tensor]): State dict in Meta's format.
+        state_dict (Dict[str, torch.Tensor]): State dict in HF's format.
```
kartikayk, line 129: Nice catch!

kartikayk (approved):
Generally looks good, thank you!

PR #933, opened 2024-05-03
Save adapter config and remapped adapter weights for loading into PEFT

ebsmothers:
### PEFT integration

This is a PR for integration with [PEFT](https://github.com/huggingface/peft). With this integration, you can take a fine-tuned checkpoint from torchtune and load it into a PEFT model using `from_pretrained` for continued fine-tuning or inference.

First, finetune a model in torchtune. Using the tune CLI:

```
tune run lora_finetune_single_device --config llama2/7B_lora_single_device \
checkpointer.output_dir=/my/output/dir
```

Then in Python:

```
from transformers import AutoModelForCausalLM
from peft import PeftModel

# hub ID of the base model from the above fine-tune
model_id = "meta-llama/Llama-2-7b-hf" 

# output_dir from tune command
checkpoint_dir = "/my/output/dir" 

model = AutoModelForCausalLM.from_pretrained(model_id)
peft_model = PeftModel.from_pretrained(model, checkpoint_dir)
```

And that's it! You can now use `peft_model` as you would any other PEFT model class.

We automatically output the necessary files/formats for PEFT integration whenever using torchtune's [HF checkpointer](https://github.com/pytorch/torchtune/blob/29ae975fc6d2f8e85ce33634116e0bda0472253c/torchtune/utils/_checkpointing/_checkpointer.py#L261), so make sure to use that in your fine-tuning config if you want to load your torchtune checkpoints into PEFT ([example config usage](https://github.com/pytorch/torchtune/blob/29ae975fc6d2f8e85ce33634116e0bda0472253c/recipes/configs/llama2/7B_lora.yaml#L35)).

### Implementation

We save a file `adapter_config.json`, along with `adapter_model.bin` to match the format expected by PEFT. We also remap the LoRA weights to match the HF format (due to differences in RoPE implementations).

The save logic differs depending on checkpointer and model. In summary:

- For Meta checkpointer and tune checkpointer we do not make any changes, and continue to save adapter weights in the tune format. This is to be consistent with the principle of same input format, same output format (since PEFT output format only matches HF format).
- For HF checkpointer we still output tune format adapter weights (to allow resumption from intermediate checkpoints), but we also output HF-mapped adapter weights **except** for in the case of phi-3 models. See [this comment](https://github.com/pytorch/torchtune/pull/933#discussion_r1605612076) for the rationale.

#### Unit tests
Added unit test in `test_checkpointer.py` to verify that adapter config and PEFT-compatible weights are saved as expected

```
pytest tests/torchtune/utils/test_checkpointer.py
...
======= 6 passed in 1.19s ==========
```

#### Recipe tests

```
pytest -m integration_test tests/recipes
...
==== 18 passed, 1 deselected, 3 warnings in 167.99s (0:02:47) =======
```

#### Manual E2E test

First create the file test_peft_integration.py as in [this gist](https://gist.github.com/ebsmothers/98cca7eb05d4ce29e8c5e745d27d087e).

(1) ✅ Permute of LoRA weights works as expected (i.e. `_permute_lora_matrix(B) * A = _permute(B*A)`, which I think is what we want).
(2) ✅ Uploaded adapter weights can be loaded into a transformers model via `from_pretrained`
(3) ✅  Model forwards match within a reasonable tolerance across PEFT-loaded and torchtune-loaded checkpoints

For (3):

### Test case 1: default config (Q and V only)
```
tune run lora_finetune_single_device --config llama2/7B_lora_single_device gradient_accumulation_steps=1 \
max_steps_per_epoch=500 dtype=fp32 checkpointer.output_dir=/data/users/ebs/test_peft_integration
```

to save a fine-tuned LoRA checkpoint with adapter config and adapter weights in PEFT format. Then to compare forward pass when loading our fine-tuned checkpoint into PEFT vs into torchtune:

```
python3 test_peft_integration.py --checkpoint-dir=/data/users/ebs/test_peft_integration
...
Maximum difference: 9.298324584960938e-05
```

### Test case 2: all layers, custom LoRA rank and alpha
```
tune run lora_finetune_single_device --config llama2/7B_lora_single_device \
model.lora_attn_modules=['q_proj','k_proj','v_proj','output_proj'] model.apply_lora_to_mlp=True \
model.apply_lora_to_output=True gradient_accumulation_steps=1 max_steps_per_epoch=100 model.lora_rank=16 \
model.lora_alpha=64 dtype=fp32 checkpointer.output_dir=/data/users/ebs/test_peft_integration_full
```
Then

```
python3 test_peft_integration.py --checkpoint-dir=/data/users/ebs/test_peft_integration_full
...
Maximum difference: 0.000152587890625
```


recipes/lora_finetune_single_device.py
```diff
@@ -256,6 +258,9 @@ def _setup_model(
 
         self._lora_rank = cfg_model.lora_rank
         self._lora_alpha = cfg_model.lora_alpha
+        self._lora_attn_modules = list(cfg_model.lora_attn_modules)
+        self._apply_lora_to_mlp = cfg_model.apply_lora_to_mlp
+        self._apply_lora_to_output = getattr(cfg_model, "apply_lora_to_output", False)
```
kartikayk, line 263: not related to this PR, but maybe at some point we should consider replacing the ```apply_lora_to_*``` flags with just adding ```mlp``` and ```output``` to the ```lora_modules```?

recipes/lora_finetune_single_device.py
```diff
@@ -405,10 +410,22 @@ def save_checkpoint(self, epoch: int) -> None:
...
         ckpt_dict.update({utils.ADAPTER_KEY: adapter_state_dict})
+        adapter_config = {
+            "r": self._lora_rank,
+            "lora_alpha": self._lora_alpha,
+            "target_modules": get_lora_module_names(
+                self._lora_attn_modules,
+                self._apply_lora_to_mlp,
+                self._apply_lora_to_output,
+            ),
+            "peft_type": "LORA",
```
kartikayk, line 421: Nice!

torchtune/models/convert_weights.py
```diff
@@ -198,3 +198,78 @@ def _permute(t, n_heads):
         converted_state_dict[new_key] = value
 
     return converted_state_dict
+
+
+_TO_PEFT_KEYS = {
```
kartikayk, line 203: Maybe some quick comments on what these dicts refer to?

torchtune/models/convert_weights.py
```diff
@@ -198,3 +198,78 @@ def _permute(t, n_heads):
...
+    for k, v in _TO_PEFT_KEYS.items():
+        full_mapping.update(
+            {
+                vv.replace(".weight", f".{k}.weight"): kk.replace(
+                    ".weight", f".{v}.weight"
+                )
+                for kk, vv in _FROM_HF.items()
+                if vv is not None
+            }
+        )
```
kartikayk, line 256: This block can use some comments explaining what's going on here

torchtune/models/convert_weights.py
```diff
@@ -198,3 +198,78 @@ def _permute(t, n_heads):
...
+
+    head_dim = dim // num_heads
+
+    def _permute_lora_matrix(t, n_heads):
```
kartikayk, line 260: So these are permuted as well - nice find!

kartikayk (commented):
Generally looks good - just needs some clean up and comments to make this easy to understand

recipes/lora_finetune_single_device.py
ebsmothers: Yeah agreed, I think this is likely where we'll head eventually. One thing is that we will probably want to make LoRA in MLP more configurable (i.e. use `w1`, `w2`, `w3` (or hopefully more descriptive names) instead of `mlp`). Otherwise the relationship between e.g. `q_proj` (nn.Linear) and `mlp` (FeedForward) being in the same config is a bit confusing. Anyways this shouldn't be a huge effort to change

torchtune/models/convert_weights.py
ebsmothers: Only B matrices though 😃

torchtune/models/convert_weights.py
```diff
@@ -198,3 +198,78 @@ def _permute(t, n_heads):
...
 
     return converted_state_dict
+
+
+_TO_PEFT_KEYS = {
+    "lora_a": "lora_A",
+    "lora_b": "lora_B",
+}
+
+_TO_PEFT_TARGET_MODULES = {
```
BenjaminBossan, line 208: I wonder if a single mapping can be maintained for all supported architectures. I haven't actually tried if it works, but just checked the key names for the supported models and Phi3 seems to use `gate_up_proj` (https://huggingface.co/microsoft/Phi-3-mini-4k-instruct/tree/main?show_file_info=model-00001-of-00002.safetensors). So I wonder if one mapping per architecture is required (with this being the default mapping).

recipes/lora_finetune_single_device.py
BenjaminBossan: I agree that a single list is more intuitive, since, AFAICT, this is just consolidated into a single list under the hood.

> or hopefully more descriptive names

Changing names later on can invalidate the saved checkpoints, so would require some versioning for backwards compatibility.

BenjaminBossan (commented):
Great work, this should hopefully be useful to many users. I don't have any critical comments, just a couple of smaller ones.

It might also be a good idea to add a test for each supported architecture, just to be sure that the re-mappings of the keys are the same for all of them.

test_peft_integration.py
```diff
@@ -0,0 +1,110 @@
...
+    tt_model.eval()
+    peft_model.eval()
+    with torch.no_grad():
+        peft_out = peft_model(inputs)
+        tt_out = tt_model(inputs)
+    print(f"Maximum difference: {torch.max(torch.abs(peft_out.logits - tt_out))}")
+
+
+if __name__ == "__main__":
+    # test_permute()
```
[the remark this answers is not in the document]
ebsmothers: Yeah actually this whole test is probably going to be scrapped. As you point out in another comment, it is pretty expensive to run. Really I think I am gonna adopt a version of your other suggestion and will just add a unit test to confirm that key conversions etc are done correctly.

test_peft_integration.py
```diff
@@ -0,0 +1,110 @@
...
+    )
+    checkpoint_dict = checkpointer.load_checkpoint()
+    tt_model.load_state_dict(checkpoint_dict["model"])
+
+    tt_model.eval()
+    peft_model.eval()
+    with torch.no_grad():
+        peft_out = peft_model(inputs)
+        tt_out = tt_model(inputs)
+    print(f"Maximum difference: {torch.max(torch.abs(peft_out.logits - tt_out))}")
```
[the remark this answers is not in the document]
ebsmothers: Yeah, will probably wind up removing this anyways. But in the forthcoming unit test I will use asserts

test_peft_integration.py
```diff
@@ -0,0 +1,110 @@
...
+    model_id = "meta-llama/Llama-2-7b-hf"
+    model = AutoModelForCausalLM.from_pretrained(model_id)
+    peft_model = PeftModel.from_pretrained(model, checkpoint_dir)
+
+    vocab_size, bsz, seq_len = 32000, 2, 128
+    inputs = torch.randint(0, vocab_size, (bsz, seq_len))
+
+    # Initialize Llama2 and load merged checkpoint
+    # (just testing that forward lines up)
+    tt_model = llama2_7b()
```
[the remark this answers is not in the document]
ebsmothers: Yeah this is prob the better way. I was taking advantage of the extra memory for debugging.. defining attributes on each model class for intermediate values then comparing each step along the way. But this is gonna get scrapped anyways

torchtune/models/convert_weights.py
ebsmothers: Good point. I've actually only tested for Llama2 so far, I think you're right that we'll need a separate mapping at least for Phi-3. We do have something [here](https://github.com/pytorch/torchtune/blob/dc2b9911cf5a793a8b4f9cd4df321c5271745b4a/torchtune/models/phi3/_convert_weights.py#L18) for the full checkpoint mapping already, will just need to adapt it for PEFT purposes.

recipes/lora_finetune_single_device.py
ebsmothers: Ah good point. I was trying to avoid this initially since it may necessitate some changes to our load_checkpoint [method](https://github.com/pytorch/torchtune/blob/dc2b9911cf5a793a8b4f9cd4df321c5271745b4a/torchtune/utils/_checkpointing/_checkpointer.py#L371), as right now we really only retrieve and remap model weights. If it's more of a nice-to-have, I may punt on it for this particular PR to keep things more isolated to save_checkpoint. Lmk if this makes sense. Also cc @kartikayk if you have any general thoughts on loading state/metadata through `load_checkpointer` and passing through our recipe. I imagine this is something we may want to start supporting more for various integrations anyways.

torchtune/models/convert_weights.py
ebsmothers: Update: there are other challenges with loading fine-tuned phi-3 checkpoints into PEFT from torchtune related to fused vs non-fused QKV. Namely, if someone fine-tunes in torchtune only on e.g. Q and K, they will not really be able to continue fine-tuning in PEFT in the way they would expect. In that case we can of course zero out the weights of the V chunk of the PEFT QKV LoRA matrix to get something that is in spirit correct, but (a) the user would probably expect only Q and K to remain trainable, which would not be the case, and (b) the learned LoRA weights from the torchtune finetune based on Q and K only may put any subsequent PEFT fine-tune using V as well in a suboptimal initial parameter space.

We could enforce up front that phi-3 LoRA is all-or-nothing on Q, K, and V for PEFT integration but I feel that's a bit messy. So for the time being I am opting to just raise a warning on checkpoint save that phi-3 adapter weights cannot be loaded into PEFT, and save just the usual torchtune adapter weights in that case.

recipes/lora_finetune_distributed.py
```diff
@@ -477,6 +477,19 @@ def save_checkpoint(
         # to be sent to the checkpointer and ultimately written to file
         if self._is_rank_zero:
 
+            # if training is in-progress, checkpoint the optimizer state and recipe state
```
joecummings, line 480: Dumb q: Why move this up?

torchtune/models/convert_weights.py
joecummings: Where's this warning?

recipes/lora_finetune_distributed.py
ebsmothers: Mainly to align with the ordering in the single-device recipe, but realistically it doesn't matter too much. Maybe I'll just leave as-is to not mix in extra complexity with the current set of changes. Also I realized somehow my changes to save PEFT config did not get pushed to this recipe, will update that now.

torchtune/models/convert_weights.py
BenjaminBossan: I see, yes I think giving a warning is the best solution in this situation.

The only issue I have with the warning is that it is only given during checkpointing. I would be afraid that a user starts an expensive training run only to find out the next day that the checkpoint was not saved as expected. Would it be possible to give the warning already at model initialization time?

recipes/lora_finetune_single_device.py
kartikayk: Can you expand a bit more on why we would need ```base_model_name_or_path```? Is this to make sure there are no bugs related to selecting the right base model for further training in HF land? If so, I wonder if this is something which is a "must have" rather than a "good to have"? or let me know if I misunderstand?

If it's a must have, then is this something we can read from one of the json files or do we need to pass this information along through the recipe?

torchtune/models/convert_weights.py
```diff
@@ -198,3 +198,83 @@ def _permute(t, n_heads):
...
+    "w2": "down_proj",
+    "w3": "up_proj",
+    "output": "lm_head",
+}
+
+
+def tune_to_peft_adapter_config(
+    adapter_config: Dict[str, Any],
+):
+    expected_keys = ["target_modules", "r", "lora_alpha"]
```
kartikayk, line 225: Does this need to be a constant like ```_TO_PEFT_TARGET_MODULES```?

torchtune/models/convert_weights.py
```diff
@@ -198,3 +198,83 @@ def _permute(t, n_heads):
...
+        if k not in _TO_PEFT_TARGET_MODULES:
+            raise ValueError(f"Unknown target module {k}")
+    adapter_config["target_modules"] = list(
+        map(_TO_PEFT_TARGET_MODULES.get, adapter_config["target_modules"])
+    )
+
+    return adapter_config
+
+
+def tune_to_peft_adapter_weights(
```
kartikayk, line 241: @BenjaminBossan I'm curious what your thoughts are on this function. It seems like this (along with other similar conversion functions) are fairly brittle and susceptible to breakages resulting from changes in PEFT/Transformers. A couple of questions:

- How brittle is this in practice? Do we expect changes in these keys or permutation logic often?
- Are the unit tests enough to capture this? Do we need to add similar tests on the PEFT side as well?

torchtune/utils/_checkpointing/_checkpointer.py
```diff
@@ -482,12 +484,57 @@ def save_checkpoint(
                 f"{os.path.getsize(output_path) / 1000**3:.2f} GB "
                 f"saved to {output_path}"
             )
+            # Phi-3-mini uses fused QKV in PEFT, this will not work as expected
+            # if only a proper subset of Q, K, V have been fine-tuned
+            if self._model_type == ModelType.PHI3_MINI:
```
kartikayk, line 489: Definitely getting very wieldy. As we discussed, probably there is opportunity to refactor checkpointers with a focus on adapters

kartikayk (approved):
Generally looks good - thanks for persevering through all of the unknowns! A bunch of future-facing questions which would be good to think about.

recipes/lora_finetune_single_device.py
BenjaminBossan: We don't strictly need `base_model_name_or_path`, but not having it means that the burden is on the user to figure out which base model this adapter belongs to. Of course, this can be solved with good documentation, but having it automatically in the `adapter_config.json` would be quite convenient.

Other points to consider:
- When shared on HF Hub, this metadata can be used for other things (I'm not an expert on this though)
- If `base_model_name_or_path` is present, users can load the adapter + base model in a single line of code (e.g. `AutoModelForCausalLM.from_pretrained(<path-to-adapter>)`).

torchtune/models/convert_weights.py
BenjaminBossan: > * How brittle is this in practice? Do we expect changes in these keys or permutation logic often?

No, there shouldn't be any frequent changes in this regard, as that would result in incompatibilities of old HF checkpoints as well. Generally, when something changes in the modeling code, we try to preserve the format of the checkpoint and re-map while loading the `state_dict`. I won't say it never happened in the past but I think it would generally be considered a bug and we'd fix it if notified.

> * Are the unit tests enough to capture this? Do we need to add similar tests on the PEFT side as well?

This probably wouldn't hurt. I could imagine that if you push a converted checkpoint to the HF Hub (ideally a small model), we can add a test to check if we can load it successfully.

PR #953, opened 2024-05-08
Utilize compile on model rather than via torch API

joecummings:
**Context**:

As part of the compile hackathon organized by @rohan-varma, this allows us to delete some unused code that is now handled by the API on ``nn.Module``.

**Changelog**:
* Change method to call .compile on the ``nn.Module`` class in the recipe
* Delete unused code + test

**Testing**:

1. First run with old compile and `tune run full_finetune_single_device --config phi3/mini_full_low_memory compile=True checkpointer.checkpoint_dir=./phi3 tokenizer.path=./phi3/tokenizer.model`. Loss looks the same and it/s is ~3
2. Then run with new compile and `tune run full_finetune_single_device --config phi3/mini_full_low_memory compile=True checkpointer.checkpoint_dir=./phi3 tokenizer.path=./phi3/tokenizer.model`. Loss looks the same is it/s is ~3.


torchtune/utils/_compile_utils.py
```diff
@@ -6,37 +6,25 @@
...
-
-def wrap_compile(model: nn.Module) -> None:
+def wrap_compile(model: nn.Module) -> nn.Module:
     """
     Wraps the model with torch.compile. This function will also
-        register a state_dict post hook that allows state_dicts produced
-        with torch.compile training to behave as regular eager mode models.
-        In particular, it strips away a torch.compile specific prefix
-        added to the state_dict by torch.compile.
+    register a state_dict post hook that allows state_dicts produced
```
rohan-varma, line 15: Remove this part of the docstring

rohan-varma (approved):
One comment, other than that LGTM pending CI.

rohan-varma (commented):
There might be some UX implications such as recipe readability by having the backend reading from the env in the recipe directly, this was a little hidden previously through the wrapper. The reason this is here is because we haven't figured out how to run inductor backend in unittests.

Deferring to @ebsmothers on the UX aspect

ebsmothers (approved):

PR #954, opened 2024-05-08
Update docs

joecummings:
#### Context
- [ ] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

#### Changelog

#### Test plan

### DIFF
--- a/docs/source/api_ref_models.rst
+++ b/docs/source/api_ref_models.rst
@@ -2,105 +2,292 @@
 torchtune.models
 ================
 
-.. currentmodule:: torchtune.models
 
-llama3
+This page contains information on all model architectures available in the core torchtune library.
+
+.. note::
+
+    If you don't want to explicitly provide a Hugging Face token to the model download command,
+    you can also run `huggingface-cli login` to persist your token in the environment.
+
+
+Llama3
 ------
 
-All models from the `Llama3 family <https://llama.meta.com/llama3/>`_.
+.. currentmodule:: torchtune.models.llama3
 
-.. code-block:: bash
+Model architectures from the `Llama3 family <https://llama.meta.com/llama3/>`_.
 
-    tune download meta-llama/Meta-Llama-3-8B-Instruct --hf-token <ACCESS_TOKEN>
-
-
-.. autosummary::
-    :toctree: generated/
-    :nosignatures:
-
-    llama3.llama3_8b
-    llama3.llama3_70b
-    llama3.lora_llama3_8b
-    llama3.qlora_llama3_8b
-    llama3.lora_llama3_70b
-
-
-llama2
-------
-
-All models from the `Llama2 family <https://llama.meta.com/llama2/>`_.
+Download the model
+^^^^^^^^^^^^^^^^^^
 
 Pre-trained models can be downloaded from the Hugging Face Hub with the following command:
 
 .. code-block:: bash
 
-    tune download meta-llama/Llama-2-7b-hf --hf-token <ACCESS_TOKEN>
+    tune download meta-llama/Meta-Llama-3-8B-Instruct --output-dir /tmp/Meta-Llama-3-8B-Instruct --hf-token <ACCESS_TOKEN>
+
+Model architectures
+^^^^^^^^^^^^^^^^^^^
 
 .. autosummary::
     :toctree: generated/
     :nosignatures:
 
-    llama2.llama2_7b
-    llama2.lora_llama2_7b
-    llama2.qlora_llama2_7b
-    llama2.llama2_13b
-    llama2.lora_llama2_13b
-    llama2.qlora_llama2_13b
-    llama2.llama2_70b
-    llama2.lora_llama2_70b
+    llama3
+    llama3_8b
+    lora_llama3_8b
+    qlora_llama3_8b
+    llama3_70b
+    lora_llama3_70b
 
-phi-3
+Tokenizer
+^^^^^^^^^
+
+.. autosummary::
+    :toctree: generated/
+    :nosignatures:
+
+    llama3_tokenizer
+
+Usage
+~~~~~
+
+.. code-block:: python
+
+    >>> from torchtune.models.llama3 import llama3_8b, llama3_tokenizer
+    >>> from torchtune.utils import FullModelHFCheckpointer
+    >>> checkpointer = FullModelHFCheckpointer(
+    ...     checkpoint_dir="/tmp/Meta-Llama-3-8B-Instruct",
+    ...     checkpoint_files=[""],
+    ...     model_type="llama3",
+    ...     output_dir="/tmp/finetuned-model",
+    ... )
+    >>> state_dict = checkpointer.load_checkpoint()
+    >>> model = llama3_8b()
+    >>> model.load_state_dict(state_dict)
+    >>> tokenizer = llama3_tokenizer(path="/tmp/Meta-Llama-3-8B-Instruct/tokenizer.model")
+
+
+Llama2
+------
+
+.. currentmodule:: torchtune.models.llama2
+
+Model architectures from the `Llama2 family <https://llama.meta.com/llama2/>`_.
+
+Download the model
+^^^^^^^^^^^^^^^^^^
+
+Pre-trained models can be downloaded from the Hugging Face Hub with the following command:
+
+.. code-block:: bash
+
+    tune download meta-llama/Llama-2-7b-hf --output-dir /tmp/Llama-2-7b-hf --hf-token <ACCESS_TOKEN>
+
+Model architectures
+^^^^^^^^^^^^^^^^^^^
+
+.. autosummary::
+    :toctree: generated/
+    :nosignatures:
+
+    llama2
+    llama2_7b
+    lora_llama2_7b
+    qlora_llama2_7b
+    llama2_13b
+    lora_llama2_13b
+    qlora_llama2_13b
+    llama2_70b
+    lora_llama2_70b
+
+Tokenizer
+^^^^^^^^^
+
+.. autosummary::
+    :toctree: generated/
+    :nosignatures:
+
+    llama2_tokenizer
+
+Usage
+^^^^^
+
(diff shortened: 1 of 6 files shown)


docs/source/api_ref_models.rst
```diff
@@ -2,105 +2,409 @@
...
-All models from the `Llama3 family <https://llama.meta.com/llama3/>`_.
+.. currentmodule:: torchtune.models.llama3
+
+`Llama 3 <https://llama.meta.com/llama3/>`_ - released in April of 2024 - is an auto-regressive language model that uses an optimized transformer architecture.
+The tuned versions use supervised fine-tuning (SFT) and reinforcement learning with human feedback (RLHF) to align with human preferences for helpfulness and safety.
+The model was pretrained on over **15 trillion tokens** of data from publicly available sources. The pretraining data has a cutoff of March 2023 for the 8B and December 2023
+for the 70B models respectively. The fine-tuning data includes publicly available instruction datasets, as well as over **10M human-annotated examples**.
+You can read more about the model from the official `model card <https://github.com/meta-llama/llama3/blob/main/MODEL_CARD.md#model-details>`_.
+
+Download the model
```
SalmanMohammadi, line 19: I think we've sufficiently covered downloading models in our other docs

docs/source/api_ref_models.rst
```diff
@@ -2,105 +2,409 @@
...
+^^^^^^^^^^^^^^^^^^
+
+Pre-trained models can be downloaded from the Hugging Face Hub with the following command:
 
 .. code-block:: bash
 
-    tune download meta-llama/Meta-Llama-3-8B-Instruct --hf-token <ACCESS_TOKEN>
+    tune download meta-llama/Meta-Llama-3-8B-Instruct --output-dir /tmp/Meta-Llama-3-8B-Instruct --hf-token <ACCESS_TOKEN>
+
+You can find all model checkpoints on the `official Meta Llama model page <https://huggingface.co/collections/meta-llama/meta-llama-3-66214712577ca38149ebb2b6>`_.
```
SalmanMohammadi, line 28: I like including this for every model family though

docs/source/api_ref_models.rst
```diff
@@ -2,105 +2,409 @@
...
-    llama3.llama3_70b
-    llama3.lora_llama3_8b
-    llama3.qlora_llama3_8b
-    llama3.lora_llama3_70b
+    llama3_tokenizer
+
+Usage
+^^^^^
+
+You can specify your Llama3 model through code ...
```
SalmanMohammadi, line 67: can probably just link to somewhere else in our docs for this

docs/source/api_ref_models.rst
```diff
@@ -2,105 +2,409 @@
...
 .. autosummary::
     :toctree: generated/
     :nosignatures:
 
-    llama3.llama3_8b
-    llama3.llama3_70b
-    llama3.lora_llama3_8b
-    llama3.qlora_llama3_8b
-    llama3.lora_llama3_70b
+    llama3_tokenizer
```
SalmanMohammadi, line 62: worth linking to something like https://github.com/meta-llama/llama3/blob/main/llama/tokenizer.py for a one-file implementation?

SalmanMohammadi:
Overall this is a huge improvement. I'm not married to the blurbs - we can link to model cards/papers which sufficiently explain the models IMO.

I'd love to see the following things (some of which are here already):

- A top-level note to let users know they can swap all the models they see here across all of our recipes. How much detail we provide on doing so vs. linking to other docs I'm open to discussing.
- **Memory usage of different models** - I think this is just a nice QOL add, even just to say the memory usage default for bf16.
- The explainer on generic builders has been sorely missing. I wouldn't mind a little example showing when to use one (e.g. for tinyllama or something)
- Would it be inappropriate to immortalize a contributor's contribution if they contributed a specific model, by linking to their contribution in the model section?
- I think the tokenizer sections could use a little love, particularly since we've fleshed these out a bit. For example, worth specifying llama tokenizer is a sentencepiece tokenizer, and not the fast BPE tokenizer?

More thoughts maybe to come. Thanks for reviving this.

docs/source/api_ref_models.rst
```diff
@@ -2,105 +2,409 @@
...
+    :toctree: generated/
+    :nosignatures:
+
+    llama3_8b
+    lora_llama3_8b
+    qlora_llama3_8b
+    llama3_70b
+    lora_llama3_70b
+
+If you want to experiment with different sizes of models or train a model from scratch, we also provide generic builder
```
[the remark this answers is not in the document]
SalmanMohammadi: and uh, omit the train from scratch?

docs/source/api_ref_models.rst
```diff
@@ -2,105 +2,409 @@
...
-llama3
+This page contains information on all model architectures available in the core torchtune library.
+
+Llama3
 ------
 
-All models from the `Llama3 family <https://llama.meta.com/llama3/>`_.
+.. currentmodule:: torchtune.models.llama3
+
+`Llama 3 <https://llama.meta.com/llama3/>`_ - released in April of 2024 - is an auto-regressive language model that uses an optimized transformer architecture.
```
SalmanMohammadi, line 13: I kind of see what you mean about maintaining the blurbs now, we could be writing a lot of these

SalmanMohammadi:
Also, maybe this is obvious to everyone, but it might be a good opportunity to add a little noobsplainer™  about what's up with all our `lora` model and component builders. We can defer to the tutorial for in-depth details here. For example, `lora_llama2_7b` isn't actually another Llama2 model you can download or find the weights for, it's just the LoRA version of the same model, and here's how to specify it, et cetera. Maybe my section in the memory glossary is a good link here?

docs/source/api_ref_models.rst
felipemello1: The description is interesting, but we dont follow the same standard for the other models. I think that giving general info, e.g. paper link, size, config link may be good enough.

docs/source/api_ref_models.rst
```diff
@@ -2,105 +2,409 @@
...
-    llama3.lora_llama3_70b
+    llama3_tokenizer
+
+Usage
+^^^^^
+
+You can specify your Llama3 model through code ...
+
+.. code-block:: python
+
```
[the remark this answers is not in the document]
felipemello1: Maybe this could be in one of the tutorials, e.g. all about configs: 
Or a new tutorial: "using torchtune with notebooks"

felipemello1 (commented):
I like it, but I think it needs a bit more of brainstorming. I worry a bit with noise/information ratio. I feel that maybe reorganizing some information/tutorials can yield a better result (but also more work).

For example: 
- should we have a single page with all models, or a page per models with dropdown? I remember that i couldnt find how to download gemma in torchtune page. API makes me think of code, and not information about models.
- Should we share how to download/instantiate a model, but not how to finetune? Is there a better approach?
- If/when we have model results memory/tokens per second, should we add it here too?
- Is this the right place to expose the tokenizer / builder? Or is this too low level?

docs/source/api_ref_models.rst
```diff
@@ -2,105 +2,409 @@
...
+Download the model
+^^^^^^^^^^^^^^^^^^
+
+Pre-trained models can be downloaded from the Hugging Face Hub with the following command:
 
 .. code-block:: bash
 
-    tune download meta-llama/Meta-Llama-3-8B-Instruct --hf-token <ACCESS_TOKEN>
+    tune download meta-llama/Meta-Llama-3-8B-Instruct --output-dir /tmp/Meta-Llama-3-8B-Instruct --hf-token <ACCESS_TOKEN>
```
felipemello1, line 26: what if we replace this "download" part, with a recommendation to read the config docstring? When I was trying torchtune, it was not obvious to me that the details would be there.

docs/source/api_ref_models.rst
SalmanMohammadi: Having slept on it, I actually like this, particularly if we make the move towards a single page per model family. I think it adds value for users who maybe haven't immediately clocked the usage of `ignore-patterns` or just generally want to get started with fine-tuning a given model without thinking too hard about it.
+1

docs/source/api_ref_models.rst
ebsmothers: Personally I would support adding a quick 1-2 sentences at the top for each model though. I think this helps give a quick idea of what the main points are, then can point to references (imo this version has about 25% too much info though, keep it short and sweet). Agree with @felipemello1's point.. whatever precedent we set we should follow it for every model.

docs/source/api_ref_models.rst
ebsmothers: > if we make the move towards a single page per model family

Where are we at on this? I think this would be a nice thing to do

ebsmothers:
Overall I think this is a huge improvement, we just need to ensure the updates are applied consistently across all the different models we support. A few responses to various comments I've seen so far:

(1) I see a few comments about noise/information ratio, especially around which APIs we do and don't document. Imo we shouldn't compromise on documenting our public APIs, even if it means there is extra stuff on our docs pages. Instead, we should structure our docs such that the important stuff is easy to find, and the less pressing stuff is still available. I think separating models out into separate pages will help, as will structuring each individual page from most essential info at the top to least essential info at the bottom. 

(2) Re stuff like commands to fine-tune a given model or a given model's memory, I don't think we should put it here. Instead, we should provide this information elsewhere and link to it. These are API docs and so first and foremost their job is to provide everything a user needs to know about interacting with said APIs. So the existing example usage is the right idea imo -- it shows how to use the provided APIs in Python in a minimal and self-contained way. (Though I would think about how we can make the example usage even a bit clearer, the checkpointer portion is unfortunately a bit of a black box in its current format.)

> Would it be inappropriate to immortalize a contributor's contribution if they contributed a specific model, by linking to their contribution in the model section?

This I wouldn't do personally. Unlike tutorials where we can reference a contributor in a single line at the top, our API docs shouldn't have any extra fluff

PR #956, opened 2024-05-09
Update tokens_per_sec to tokens_per_sec_per_gpu

kartikayk:
#### Context
We log ```tokens_per_sec_per_gpu``` but this is called ```tokens_per_sec``` which causes some confusion.

#### Changelog
Update the name of the param being logged

#### Test plan

Eyes:

Screenshot 2024-05-09 at 10 04 52 AM


RdoubleA:
would `tokens_per_sec_per_gpu` ever differ slightly among ranks?

rohan-varma (approved):

kartikayk:
> would tokens_per_sec_per_gpu ever differ slightly among ranks?

it's a good question - it will, but the current nomenclature doesn't really call this difference out either. We assume rank 0 is representative (reasonable for randomly shuffled data). This just makes what we log more explicit

PR #961, opened 2024-05-10
Update README table

rohan-varma:
#### Context

Can't reproduce 24.1GB on a single device full finetune and instead seeing 20.2 peak memory allocated (and similar for reserved), so updating this table.

To reproduce, I ensured all inputs are seq_len=2048 - 

```
input_ids = torch.zeros((input_ids.shape[0], 2048), dtype=torch.long)
labels = torch.zeros((labels.shape[0], 2048), dtype=torch.long)
```
And ran with batch_size=4 - 

```tune run full_finetune_single_device --config recipes/configs/llama2/7B_full_low_memory.yaml batch_size=4```


kartikayk (commented):
Thanks for the update! Are you running on a 4090? If not, can you test on the right hardware? Not sure why that will cause discrepancies in memory numbers. But I'd like to make sure the numbers are on the right hardware. On my A100, I'm getting 22.3, which is still higher than the number you're seeing. So seems like there's some discrepancy?

joecummings:
I think we need to get new numbers for this in general, probably can do this through an automated process. Closing this for now.

PR #963, opened 2024-05-10
[DONT MERGE - JUST RFC] Use hierarchical compilation

anijain2305:
Just to show how to use hierarchical compilation. On my machine, I see compilation finishing very quickly. Would love for someone from TorchTune to do two things

* Measure the compilation time using the same infra as before to compare against full model compilation
* Compare perf numbers with full model compilation


kartikayk (commented):
Thanks for the PR @anijain2305! @ebsmothers and I can help with this. We'll take a look early next week on this

PR #966, opened 2024-05-12
Updating generation to use custom_generate_next_token for initial token generation

SalmanMohammadi:
#### Context
- [ ] add a new feature
- [x] fix a bug
- [ ] other (please add here)

I'n using `generate` with a custom `custom_generate_next_token` defined for a model which outputs both the output of `TransformerDecoder` and an additional value head. `generate` fails when making the first call to `generate_next_token` as it encounters an unexpected return signature for `model(...)`.

Let me know if there's a better way for me to do this.

#### Changelog

Updated the initial call to `generate_next_token` to use `custom_generate_next_token`.

#### Test plan

### DIFF
--- a/torchtune/utils/_generation.py
+++ b/torchtune/utils/_generation.py
@@ -120,7 +120,7 @@ def generate(
         custom_generate_next_token = generate_next_token
 
     # generate the first tokens conditioned on the prompt
-    tokens = generate_next_token(
+    tokens = custom_generate_next_token(
         model,
         input_pos=torch.arange(0, prompt_length, device=prompt.device),
         x=prompt,


kartikayk (approved):
@SalmanMohammadi Thanks for opening this PR!

So my understanding of why the first token is generated using ```generate_next_token``` and not ```custom_generate_next_token``` is because thus far we were using the ```custom_generate_next_token``` to pass in the compiled version of the function (see [here](https://github.com/pytorch/torchtune/blob/main/recipes/generate.py#L95)). The way I understand this works is that the compiled version assumes that the input sequence has a length of 1 (since we're caching the KV tensors). The first token on the other hand has an input sequence of length equal to the length of the prompt. If we make this change, I think we'll likely trigger a recompile which will increase generation time for the first few generations, though I dont think there will be other side effects. Since we're anyways not using compile outside of the quantization use case, I think it should be ok to land this. But we'll need to figure this out when we add support for compile during generation.

PR #690, opened 2024-04-11
E2E Tutorial

kartikayk:
#### Context

Adding an e2e tutorial which captures the complete flow within TorchTune. This tutorial is a good overview of the capabilities of the library and so adding this to the Getting Started section instead of burying it in the tutorials section.

#### Changelog
- Add said tutorial
- Couple of other fixes including changing eval and quantize configs to default to HF models
- A short fix to the quantize API

#### Test plan
- Ran all of the recipes

Tutorial Rendered

Screenshot 2024-04-10 at 6 03 38 PM

Updates to the entry page rendered

Screenshot 2024-04-10 at 6 04 12 PM

PR #980, opened 2024-05-14
Add QAT support for distributed finetuning

andrewor14:
**Summary:** This commit adds the option to run quantization-aware training (QAT) during finetuning. QAT refers to "fake quantizing" the weights and activations during training, which performs the following transformation on the inputs but still keeps all intermediate values in floating point:

```
x_q = clamp((x_bf16 / scale) + zp)
x_fq = (x_q - zp) * scale
```

Currently only 8-bit per token dynamic activations + 4-bit grouped per channel weights (8da4w) is supported. Users can enable this by specifying a QAT quantizer in their config files:

```
tune run --nnodes 1 --nproc_per_node 8 qat_distributed --config llama3/8B_qat_full

# or add this to your config file
# quantizer:
#   _component_: torchtune.utils.quantization.Int8DynActInt4WeightQATQuantizer
#   groupsize: 256
```

**Test Plan:**

Initial results for Llama2 demonstrate that QAT is able to recover the loss in accuracy from quantization by about half for some tasks (last two rows):

|                 | hellaswag |          |     wikitext    |                 |               | arc_easy |          | arc_challenge |          |
|-----------------|:---------:|----------|:---------------:|-----------------|---------------|:--------:|----------|:-------------:|----------|
|                 |    acc    | acc_norm | word_perplexity | byte_perplexity | bits_per_byte |    acc   | acc_norm |      acc      | acc_norm |
| No quant |  59.659%  |  76.927% |      12.183     |      1.596      |     0.674     |  76.010% |  72.054% |    48.720%    |  47.867% |
| PTQ             |  57.150%  |  74.945% |      12.995     |      1.615      |     0.692     |  75.968% |  70.118% |    46.416%    |  45.904% ||
| QAT |  58.504%  |  76.170% |      12.199     |      1.596      |     0.675     |  76.431% |  71.928% |    47.184%    |  48.123% |
| PTQ degradation |  -2.509%  |  -1.982% |      +0.812      |      +0.019      |     +0.018     |  -0.042% |  -1.936% |    -2.304%    |  -1.963% |
| QAT degradation |  -1.155%  |  -0.757% |      +0.016      |      +0.000      |     +0.001     |  0.421%  |  -0.126% |    -1.536%    |  0.256%  |

### DIFF
--- a/recipes/qat_distributed.py
+++ b/recipes/qat_distributed.py
@@ -1,0 +1,656 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+import sys
+import time
+
+from functools import partial
+from typing import Any, Dict, Optional, Tuple
+from warnings import warn
+
+import torch
+from omegaconf import DictConfig, ListConfig
+
+from torch import nn
+from torch.distributed import init_process_group
+from torch.distributed.fsdp import (
+    CPUOffload,
+    FullOptimStateDictConfig,
+    FullStateDictConfig,
+    FullyShardedDataParallel as FSDP,
+    StateDictType,
+)
+from torch.optim import Optimizer
+from torch.utils.data import DataLoader, DistributedSampler
+
+from torchtune import config, modules, utils
+from torchtune.datasets import ConcatDataset
+from torchtune.recipe_interfaces import FTRecipeInterface
+from torchtune.utils.activations import apply_selective_activation_checkpointing
+
+from tqdm import tqdm
+
+
+log = utils.get_logger("DEBUG")
+
+
+class QATRecipeDistributed(FTRecipeInterface):
+    """
+    Quantization-aware training (QAT) recipe for dense transformer-based LLMs such as Llama2.
+    This recipe supports distributed training and can be run on a single node (1 to 8 GPUs).
+    Only compatible with PyTorch 2.4+.
+
+    Features:
+        - Quantization-aware training (QAT). Perform fake quantization on weights and/or activations
+            during finetuning, with the goal of ultimately producing a quantized model with minimal
+            accuracy degradation. This recipe produces an unquantized model in the original dtype,
+            which can then be quantized separately.
+
+        - Delayed fake quantization. Optionally specify the step after which fake quantization occurs.
+            Empirically, allowing the model to finetune without fake quantization initially allows the
+            weight and activation values to stabilize before fake quantizing them, potentially leading
+            to improved quantized accuracy. This can be specified through ``fake_quant_after_n_steps``.
+
+        - FSDP. Supported using PyTorch's FSDP APIs. DDP is currently not supported. Training on CPU
+            is not supported.
+
+        - Activation Checkpointing. This can be controlled using the ``activation_checkpointing``
+            flag. Activation checkpointing helps reduce the memory footprint since we no longer keep
+            activations in memory and instead recompute them during the backward pass. This is especially
+            helpful for larger batch sizes when you're memory constrained. But these savings in memory
+            come at the cost of training performance. In most cases training can slow-down quite a bit as
+            a result of this activation recomputation.
+
+        - Precision. Full fp32 and bf16 training are supported. Precision is controlled using the ``dtype``
+            flag. When ``dtype=bf16``, all activations, gradients and optimizer states are in bfloat16. In
+            most cases this should halve the memory footprint of full precision (fp32) training, without
+            loss in model quality (will depend on the model, training data and other settings). For
+            GPUs which do not support bfloat16, we fall back to fp32. Mixed precision training and fp16
+            precision are currently not supported.
+
+        - Gradient Accumulation. You can simulate larger batch sizes by accumulating gradients. This is
+            controlled using the ``gradient_accumulation_steps`` flag.
+
+                Total Batch Size = batch_size * number of GPUs * gradient accumulation steps.
+
+            For example: with batch_size=1, nproc_per_node=2 and gradient_accumulation_steps=32 we get a
+            total batch size of 64.
+
+            Gradient accumulation is especially useful when you are memory constrained. In this case,
+            accumulating gradients might give you better training speed than enabling activation
+            checkpointing.
+
+        - Checkpointing. Model weights are checkpointed both at the end of each epoch and at the end of
+            training. Optimizer state and recipe state (seed, total_epochs, number of epochs run etc) are
+            only saved at the end of a given epoch and used in case of resuming training.
+
+            Resuming training is controlled by the ``resume_from_checkpoint`` flag. Mid-epoch checkpointing is
+            currently not supported.
+
+            For more details on the checkpointer, please take a look at
+            our checkpointer deepdive (https://pytorch.org/torchtune/main/deep_dives/checkpointer.html).
+
+        - Logging. Terminal, Disk, WandB and TensorBoard are all supported.
+
+    For a full list of example configs for this recipe, run ``tune ls`` on the command line. Each config
+    has example commands for how to kick-off training.
+
+    Args:
+        cfg (DictConfig): OmegaConf object parsed from yaml file
+
+    Raises:
+        ValueError: If ``dtype`` is set to fp16.
+    """
+
+    def __init__(self, cfg: DictConfig) -> None:
+
+        self._device = utils.get_device(device=cfg.device)
+        self._dtype = utils.get_dtype(cfg.dtype, device=self._device)
+
+        if self._dtype == torch.float16:
+            raise ValueError(
+                "full fp16 training is not supported with this recipe. Please use bf16 or fp32 instead."
+            )
+
+        if (
+            cfg.get("fsdp_cpu_offload", False)
+            and cfg.get("fused", False)
+            and not utils.torch_version_ge("2.4.0")
+        ):
+            raise RuntimeError(
+                "Using fused optimizer on CPU is only supported in PyTorch nightly."
+            )
+
+        # logging attributes
+        self._output_dir = cfg.output_dir
+        self._log_every_n_steps = cfg.get("log_every_n_steps", 1)
+        self._log_peak_memory_stats = cfg.get("log_peak_memory_stats", False)
+
+        # _is_rank_zero is used primarily for logging. In the future, the logger
+        # should directly take care of this
+        _, rank = utils.get_world_size_and_rank()
+        self._is_rank_zero = rank == 0
+
+        # Training cfg
+        self._resume_from_checkpoint = cfg.resume_from_checkpoint
+        self._gradient_accumulation_steps = cfg.gradient_accumulation_steps
+        self._fake_quant_after_n_steps = cfg.get("fake_quant_after_n_steps", None)
+        self._quantizer_mode = None
+
+        # These are public properties which are updated by the checkpoint loader
+        # when ``resume_from_checkpoint`` is `True` or validated in tests
+        self.seed = utils.set_seed(seed=cfg.seed)
+        self.epochs_run = 0
+        self.total_epochs = cfg.epochs
+        self.max_steps_per_epoch = cfg.max_steps_per_epoch
+        self.global_step = 0
+
+    def load_checkpoint(self, cfg_checkpointer: DictConfig) -> Dict[str, Any]:
+        """
+        Extract the checkpoint state from file and validate. If resume_from_checkpoint
+        is True, this also includes the recipe state.
+        """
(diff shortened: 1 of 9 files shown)


andrewor14:
By the way, the tests are failing due to torchtune's dependency on an old version of torchao, which doesn't have QAT support yet. We're about to release 0.2.0 on the torchao side (aiming early next week), so the tests won't pass until then.

kartikayk (commented):
Thanks for adding this @andrewor14! A few high-level comments apart from the one on teasing this out into a separate recipe:

> Currently only 8-bit per token dynamic activations + 4-bit grouped per channel weights (8da4w) is supported

A couple of questions:
- Do we plan on adding other quantization methods? Or whats the long term support plan for this?
- We need a lot of documentation here to make sure users understand what this actually means. Is this reasonably well understood by the general audience? for example, I'm not sure about all of the details here.

We also need to consider how to lower the bar for adopotion here. I think this needs a tutorial or a deepdive added to the torchtune docs.

cc: @ebsmothers

recipes/full_finetune_distributed.py
```diff
@@ -116,6 +116,8 @@ def __init__(self, cfg: DictConfig) -> None:
         # Training cfg
         self._resume_from_checkpoint = cfg.resume_from_checkpoint
         self._gradient_accumulation_steps = cfg.gradient_accumulation_steps
+        self._qat_enable_fake_quant_step = cfg.get("qat_enable_fake_quant_step", None)
```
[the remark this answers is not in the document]
andrewor14: This config is about when to enable fake quant, not whether to enable QAT itself. E.g. setting this to 1000 means we will run regular finetuning for the first 1000 steps, and only enable fake quant after 1000 steps. I'll add better comments/docs about this. Do you think the name makes sense or do you have suggestions?

andrewor14:
Hi @kartikayk, thanks for the comments, responding inline:

> * Do we plan on adding other quantization methods? Or whats the long term support plan for this?
Yes, in the long term we do plan to support other QAT configurations (e.g. 2- or 3-bit weight only if we can get good results), that's why I kept the quantizer specification general.

> * We need a lot of documentation here to make sure users understand what this actually means. Is this reasonably well understood by the general audience? for example, I'm not sure about all of the details here. We also need to consider how to lower the bar for adopotion here. I think this needs a tutorial or a deepdive added to the torchtune docs.

For sure. Should I add the README in this PR and add the tutorial separately?

andrewor14:
Hi @kartikayk, thanks for the comments, responding inline:

> * Do we plan on adding other quantization methods? Or whats the long term support plan for this?

Yes, in the long term we do plan to support other QAT configurations (e.g. 2- or 3-bit weight only if we can get good results), that's why I kept the quantizer specification general.

> * We need a lot of documentation here to make sure users understand what this actually means. Is this reasonably well understood by the general audience? for example, I'm not sure about all of the details here. We also need to consider how to lower the bar for adopotion here. I think this needs a tutorial or a deepdive added to the torchtune docs.

For sure. Should I add the README in this PR and add the tutorial separately?

kartikayk:
> Should I add the README in this PR and add the tutorial separately?

I think a tutorial/deep-dive in the docs would be really helpful. You can add some details on what the quantization methods mean as well - I think this will be very useful and make the flow more noob friendly.

Here are some pointers:
- Example PR with a tutorial (checkout the e2e_flow.rst file): https://github.com/pytorch/torchtune/pull/690/files and this with the LoRA tutorial: https://github.com/pytorch/torchtune/pull/368
- Working with the live docs in torchtune: https://github.com/pytorch/torchtune/blob/main/CONTRIBUTING.md#building-docs

I don't mind this as a follow up. 

@ebsmothers let me know if you have differing thoughts on this.

recipes/full_finetune_distributed.py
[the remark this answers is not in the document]
kartikayk: Ah, then I grossly misunderstood :) Maybe something like ```quant_after_n_steps```?

recipes/full_finetune_distributed.py
```diff
@@ -288,6 +292,18 @@ def _setup_model(
                 ac_option,
             )
 
+        # Optionally apply quantization-aware training during finetuning
```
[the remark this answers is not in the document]
kartikayk: We've been pretty good at making sure we update all of the recipes with new features. I do think QAT is something we want to publicize heavily and so having its own recipe opens up avenues for future work as well

ebsmothers:
Thanks for the PR! I think I'm in agreement with most everything that's been said already: (1) separate recipe for QAT (I agree this will make it easier to scale once we add other quant techniques anyways) and (2) add some kind of tutorial but as a follow-up (also happy to provide any pointers or guidance you need in advance).

tests/recipes/test_qat_distributed.py
```diff
@@ -0,0 +1,102 @@
...
+            quantizer.groupsize={quantizer_group_size} \
+        """.split()
+        model_config = MODEL_TEST_CONFIGS[model_type]
+        cmd = cmd + self._get_test_config_overrides() + model_config
+
+        monkeypatch.setattr(sys, "argv", cmd)
+        runpy.run_path(TUNE_PATH, run_name="__main__")
+        loss_values = get_loss_values_from_metric_logger(log_file)
+        # TODO: delete
+        print("LOSS VALUES ARE ", loss_values)
```
andrewor14, line 98: note to self: delete

recipes/qat_distributed.py
```diff
@@ -0,0 +1,644 @@
...
+                        self._metric_logger.log_dict(
+                            log_dict,
+                            step=self.global_step,
+                        )
+
+                    # Reset running stats for the next step
+                    running_loss = 0
+                    num_tokens = 0
+                    t0 = time.perf_counter()
+                global_step += 1
```
ebsmothers, line 603: Are we double-incrementing global_step between here and L573?

recipes/quantization.md
```diff
@@ -1,13 +1,43 @@
...
+#### Quantization-Aware Training (QAT)
+
+(PyTorch 2.4+)
+
+QAT refers to applying fake quantization to weights and/or activations during finetuning,
+which means simulating only the quantization math without actually casting the original
+dtype to a lower precision.
+
+```
+tune run qat_distributed --config qat
```
ebsmothers, line 20: Need nproc_per_node here? Also nit: Add some segue before the code block, e.g. "QAT can be run via the command"

recipes/quantization.md
```diff
@@ -1,13 +1,43 @@
...
+```
+
+This produces an unquantized model in the original data type. To get an actual quantized model,
+follow this with `tune run quantize` while specifying the same quantizer in the config.
+
+The `qat_distributed` recipe is also compatible with any full finetuning config,
+as long as a QAT quantizer is specified in the config:
```
ebsmothers, line 27: Sort of. But strictly speaking we associate a given config with only a single recipe. In that case we technically don't support something like e.g.

```
tune run --nproc_per_node 4 qat_distributed --config mistral/7B_full \
quantizer=torchtune.utils.quantization.Int8DynActInt4WeightQATQuantizer \
quantizer.groupsize=256
```

(This is kinda related to my other comment about how we want to actually expose QAT configs)

recipes/quantization.md
```diff
@@ -1,13 +1,43 @@
...
+```yaml
+# QAT specific args
+quantizer:
+  _component_: torchtune.utils.quantization.Int8DynActInt4WeightQATQuantizer
+  groupsize: 256
+```
+
+Currently only `torchtune.utils.quantization.Int8DynActInt4WeightQATQuantizer`
+is supported. This refers to int8 dynamic per token activation quantization
+combined with int4 grouped per axis weight quantization.
```
ebsmothers, line 38: nit: can add a link to a reference or something here

recipes/quantize.py
```diff
@@ -54,6 +54,15 @@ class QuantizationRecipe:
...
             e.g. 32, 64, 128, 256, smaller numbers means more fine grained and higher accuracy
+
+    8da4w-qat (PyTorch 2.4+):
+        torchtune.utils.quantization.Int8DynActInt4WeightQATQuantizer
+        int8 per token dynamic activation with int4 weight only per axis group quantization
+        Same as "8da4w", but for quantizing QAT checkpoints
+        Args:
+            `groupsize` (int): a parameter of int4 weight only quantization,
+            it refers to the size of quantization groups which get independent quantization parameters
+            e.g. 32, 64, 128, 256, smaller numbers means more fine grained and higher accuracy
```
ebsmothers, line 65: Presumably also higher memory, right?

recipes/quantize.py
```diff
@@ -93,7 +104,10 @@ def _setup_model(
     @torch.no_grad()
     def quantize(self, cfg: DictConfig):
         t0 = time.perf_counter()
-        self._model = self._quantizer.quantize(self._model)
+        if "qat" in self._quantization_mode:
+            self._model = self._quantizer.convert(self._model)
+        else:
+            self._model = self._quantizer.quantize(self._model)
```
ebsmothers, line 110: So we need to gate on quantization mode because the QAT checkpoints are in bf16, right? Noob question but if we are not doing any subsequent training why can't we just call `.quantize` directly and infer all the quantizer params from the checkpoint?

recipes/qat_distributed.py
andrewor14: oh great, I didn't realize we already have a `self.global_step`. Will remove this

pyproject.toml
```diff
@@ -25,7 +25,8 @@ dependencies = [
     "omegaconf",
 
     # Quantization
-    "torchao==0.1",
+    # TODO: update to 0.3
+    "torchao==0.2",
```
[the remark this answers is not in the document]
andrewor14: I think we'll wait till 0.3 is available

recipes/configs/qat.yaml
```diff
@@ -0,0 +1,78 @@
+# Config for multi-device QAT finetuning in qat_distributed.py
```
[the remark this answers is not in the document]
andrewor14: Sounds good. I mostly just followed `quantization.yaml`, `eleuther_evaluation.yaml`, and `generation.yaml` so far in this PR. I think these configs have the same problem. Both (2) and (3) make sense to me. What do you think?

recipes/quantize.py
andrewor14: Technically we can, the numerics may be the the same, but officially the torchao QAT flow is:

```
quantizer = 8da4wQATQuantizer()
model = quantizer.prepare(model)
train(model)
model = quantizer.convert(model)
```

If we just call quantize here we would have to introduce a different quantizer

```
quantizer = 8da4wQATQuantizer()
model = quantizer.prepare(model)
train(model)
ptq_quantizer = 8da4wQuantizer()
ptq_quantizer.quantize(model)
```

I feel it's better to call the complete QAT flow rather than to switch quantizers in the middle

recipes/configs/qat.yaml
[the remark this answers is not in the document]
andrewor14: Ok, I did (2) for now (put it in respective llama2 and llama3 dirs). Let me know if this sounds reasonable to you

recipes/quantization.md
andrewor14: We don't have a README or proper docs in torchao yet, will just link to the implementation for now

recipes/configs/qat.yaml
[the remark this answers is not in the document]
ebsmothers: Yeah I think this is reasonable. It is a bit of a pain to override all the necessary config fields to change models from the command line, so I think it makes sense to separate out QAT configs by model (since for (3) it'd get really verbose to change the model from the default config anyways). The downside is the configs are slightly less visible, but given we have a standalone top-level recipe this is OK imo.

andrewor14:
Update: I think I've addressed all the comments and this PR is ready from my side. Do you have other comments @ebsmothers @kartikayk?

Note that landing is blocked right now on the torchao 0.3 release (currently scheduled for 6/26). This is because QAT was only added in torchao 0.2, but the following error was not fixed until torchao 0.3, so there's no other way to get the QAT feature unless we want to rely on nightlies, which we don't.

```
  File "/__w/_temp/conda_environment_9553124795/lib/python3.8/site-packages/torchtune/modules/common_utils.py", line 12, in <module>
    from torchao.dtypes.nf4tensor import NF4Tensor
  File "/__w/_temp/conda_environment_9553124795/lib/python3.8/site-packages/torchao/__init__.py", line 14, in <module>
    from . import _C
ImportError: /__w/_temp/conda_environment_9553124795/lib/python3.8/site-packages/torchao/_C.cpython-38-x86_64-linux-gnu.so: undefined symbol: _ZN5torch3jit11parseSchemaERKSs
```

ebsmothers (approved):

PR #983, opened 2024-05-14
Filter out special tokens and placeholder tokens

### LINKED ISSUE #976: Update Phi3 Decode method to filter out anything less than 32064, not just special tokens.

joecummings:
#### Context

The Phi3 team added 10 special tokens for instruction tuning. They round up to the nearest multiple of 64 so the total vocab size for the embedding is 32064. We should filter out ALL of these tokens during decode (however, unlikely it is that they are generated past 32010). 

If the number is greater than 32064, something is seriously wrong with the model and this should crap out completely.

Closes #976 

#### Changelog
* Change the filtering function to be a range from 32k to 32064

#### Test plan

```
from torchtune.models.phi3 import phi3_mini_tokenizer
tokenizer = phi3_mini_tokenizer(PATH_TO_TOKENIZER)
tokenizer.decode([32000, 32064])
""
```

### DIFF
--- a/torchtune/models/phi3/_sentencepiece.py
+++ b/torchtune/models/phi3/_sentencepiece.py
@@ -114,7 +114,7 @@ class Phi3MiniSentencePieceTokenizer:
         for token_id in ids:
             # Filter out special tokens and the placeholder tokens added
             # by the Phi3 team
-            if token_id >= 32_000 and token_id < 32_064:
+            if token_id >= 32_000 and token_id <= 32_064:
                 continue
             else:
                 ids_for_decode.append(token_id)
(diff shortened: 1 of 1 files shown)


kartikayk (commented):
I don't think I follow why we need this special filtering? For example, for Llama3 we have a handful of special tokens but the vocab size is 128256 - do we something similar there?

joecummings:
> I don't think I follow why we need this special filtering? For example, for Llama3 we have a handful of special tokens but the vocab size is 128256 - do we something similar there?

TikToken adds special tokens directly into the embedding table. See [here](https://github.com/pytorch/torchtune/blob/1e8d081ad86af2003e8d507191c852875947553a/torchtune/modules/tokenizers/_tiktoken.py#L115) where we're doing this. SPM is an older tokenizer and doesn't do this, so we have to handle special tokens ourselves for instruction-tuned models.

pbontrager (approved):
I'm a bit uncomfortable that we have to hardcode this range, but I can't think of a way around it given how the original phi code hardcodes the extra "special" tokens.

PR #802, opened 2024-04-18
Llama3-70B LoRA multi GPU

rohan-varma:
#### Context
- Adds LoRA support for distributed training of Llama3-70B model.
- Note that HF currently bundles both the meta and HF checkpoints into a single repo for llama3. This makes downloading only a single model a little tricky (and we should avoid user having to download 2 copies of model). For llama3-8B, the documented download commands work out of the box since safetensors are by default ignored by `tune download` and we use the meta checkpointer.
- However, using the meta checkpointer for 70B model is much more complicated and requires significant development work to get it to support sharded checkpoints and unsharding/resharding. We can get 70B support much quicker if we use the HF checkpoints and safe tensors. Hence, we're documenting (currently in configs only)how to run the download command to get the safetensors and NOT the meta checkpoint files (so users only have 1 copy of the model), and using the meta checkpointer as the default.

#### Changelog
- Render recipe in recipe registry
- Add config for 70B LoRA training
- NOTE: need to download with `--ignore-patterns "original/consolidated*"` to be able to get the safetensors
- Update README and docs as appropriate

#### Test plan
- ```tune download meta-llama/Meta-Llama-3-70b --hf-token <> --output-dir /tmp/Meta-Llama-3-70b --ignore-patterns "original/consolidated*"```
- ```tune run --nproc_per_node 8 lora_finetune_distributed --config recipes/configs/llama3/70B_lora.yaml```
- Loss curve: 
image
- Can't easily do eval due to lack of distributed support. Generation result is pending.
- Render in `tune ls` - 
image
- Some logs - 

```
2024-04-18:13:49:43,502 INFO     [lora_finetune_distributed.py:280] FSDP is enabled. Instantiating Model on CPU for Rank 0 ...
2024-04-18:13:51:59,203 INFO     [lora_finetune_distributed.py:286] Model instantiation took 135.70 secs
2024-04-18:13:53:10,922 INFO     [lora_finetune_distributed.py:360] Memory Stats after model init:
{'peak_memory_active': 28.329404416, 'peak_memory_alloc': 26.228041728, 'peak_memory_reserved': 32.377929728}
```
- Peak memory allocated is < 30 GB on GPU 0:
image

- Peak memory reserved on GPU 0:
image

PR #993, opened 2024-05-17
Llama3-70b: Full Finetune w/CPU offload + fused optimizer

rohan-varma:
#### Context
- [x] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

#### Changelog
- This PR enables 70B full finetune for Llama3 workload on 8xA100 set up with CPU offload. The configuration offered only works with PyTorch nightly which has recent changes to support `fused=True` for AdamW to utilize vectorized CPU instructions.
- Similar to https://github.com/pytorch/torchtune/pull/802, we use the HF checkpointer and load the model with HF format, with the meta format currently unsupported.
- QPS looks as follows: `1|95|Loss: 0.7184870839118958:   3%|▎         | 95/3251 [22:02<11:53:47, 13.57s/it]` - while slow, this actually seems competitive with 70B training workloads seen in other libraries.
- Appropriate warnings and error checking is added
- This PR also calls `torch.set_num_threads` after set up to ensure that the appropriate # of CPU cores are set for intra-op parallelism. On my box, this results in ~2x speed up compared to not setting it. I got the idea to do this after talking with @msaroufim @janeyx99 

#### Test plan

```
tune run --nproc_per_node 8 full_finetune_distributed --config recipes/configs/llama3/70B_full.yaml metric_logger=torchtune.utils.metric_logging.WandBLogger log_peak_memory_stats=True
```

Memory usage, sits at ~35GB active per GPU, 42 GB reserved - 

image

Loss looks comparable to 70B  model trained w/LoRA (see loss curves https://github.com/pytorch/torchtune/pull/802) - 
image


recipes/configs/llama3/70B_full.yaml
```diff
@@ -0,0 +1,108 @@
...
+seed: null
+shuffle: True
+
+# Model Arguments
+model:
+  _component_: torchtune.models.llama3.llama3_70b
+
+checkpointer:
+  _component_: torchtune.utils.FullModelHFCheckpointer
+  checkpoint_dir: /tmp/Meta-Llama-3-70b
```
rohan-varma, line 37: need to covert the paths to instruct

recipes/full_finetune_distributed.py
```diff
@@ -563,6 +584,12 @@ def recipe_main(cfg: DictConfig) -> None:
         )
 
     init_process_group(backend="gloo" if cfg.device == "cpu" else "nccl")
+    if cfg.get("fsdp_cpu_offload", False):
+        # Utilize all available CPU cores for intra-op parallelism. This provides ~2x
+        # speed up when benchmarking fused AdamW on CPU
+        threads = os.cpu_count() // torch.distributed.get_world_size()
```
janeyx99, line 590: How did you calculate this? I opened the draftiest PR https://github.com/pytorch/pytorch/pull/126199 to account for other environments where CPU affinity is set and cgroups could be used, and plan on talking with Intel to get this into better shape. But why `os.cpu_count() // torch.distributed.get_world_size()`?

janeyx99 (approved):
Woo! Nice to integrate fused CPU adam into torchtune :)

recipes/full_finetune_distributed.py
rohan-varma: @janeyx99 Unfortuantely I don't have the bets answer to this, @msaroufim suggested following this over chat and referred to answerAI doing similar (following his suggestions) - https://github.com/AnswerDotAI/fsdp_qlora/commit/1a9fddfb660b85e46bdf18c3fcf0d599628c9c6c

Intuitively, this makes sense since it's essentially assigning each process an equal number of vCPUs out of the vCPUs available. But as you mentioned, this doesn't take into account physical CPUs being < vCPUs, cgroups, and the like. Mostly meant to be a stopgap for now, and I'm sure perf will be very different on different HW set ups. In the medium term, we might want to use what you end up landing in torch core, or even expose this to the user so they can configure this.

msaroufim: Yeh I suspect comprehensive benchmarking with how diverse cpus are might take a while. So this is a fine stopgap

recipes/full_finetune_distributed.py
```diff
@@ -103,6 +105,21 @@ def __init__(self, cfg: DictConfig) -> None:
...
+                    offloading, otherwise optimizer will use non vectorized kernels and cause massive slowdown.
+                    """
+                )
+            else:
+                # We are fusing + CPU offloading - check to make sure we're in a nightly.
+                if not utils.torch_version_ge("2.4.0"):
+                    raise RuntimeError(
+                        "Using fused optimizer on CPU is only supported in PyTorch nightly."
+                    )
+
```
kartikayk, line 122: I'm not a fan of this if-else block being in the main recipe at all. This is starting to get back into the territory of the wild-wild-west recipe scripts that we spoke about a while back. Let's please brainstorm a bit about how to handle this and come up with the right design for this.

recipes/full_finetune_distributed.py
```diff
@@ -563,6 +584,12 @@ def recipe_main(cfg: DictConfig) -> None:
         )
 
     init_process_group(backend="gloo" if cfg.device == "cpu" else "nccl")
+    if cfg.get("fsdp_cpu_offload", False):
```
kartikayk, line 587: Again not a fan of this being in the recipe code, this should likely be in a utility or something which the user doesn't probably need to reason about.

cc: @ebsmothers to help brainstorm this a bit

kartikayk (changes requested):
Thanks for adding this change! I think there are some open recipe UX questions here which I'd like for us to think through before we land this.

recipes/full_finetune_distributed.py
ebsmothers: I gave a suggestion in _recipe_registry.py, at least related to the versioning side of things (but it's a bigger change than what we're looking at here). Separately we are kinda drawing a line (I don't think?) we've drawn before.. we are putting an explicit nightly check in one of our core 4 recipes. While I 

As for the actual validation of fsdp_cpu_offload + optimizer.fused, I guess it's a bit tricky to put in a single util since it's a more cross-cutting config pair. The only alternative I can think of offhand is to move into a config validation, which I think we also had legitimate concerns about doing previously. (I guess we can't validate directly on the optimizer either since it gets instantiated after the FSDP wrapping)

recipes/full_finetune_distributed.py
ebsmothers: Yeah +1 can we put this in a utility?

recipes/configs/llama3/70B_full.yaml
```diff
@@ -0,0 +1,108 @@
...
+
+# Fine-tuning arguments
+batch_size: 2
+epochs: 3
+
+optimizer:
+  _component_: torch.optim.AdamW
+  lr: 2e-5
+  foreach: False
+  fused: True
```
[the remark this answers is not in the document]
rohan-varma: This is a speedup only change (will clarify the amount of speed up we get once I re-run some benchmark)

recipes/full_finetune_distributed.py
rohan-varma: Versioning recipes make sense, though this portion has a bit more than versioning since I also want to warn if fused optimizer is not used with CPU offload. 

> we are putting an explicit nightly check in one of our core 4 recipes.

We are - but only under a very specific case - CPU offload with FSDP distributed training.

Can definitely add a helper function, though let me know if we should consider something else / alternative design. @ebsmothers @kartikayk

recipes/full_finetune_distributed.py
rohan-varma: Yeah, definitely!

ebsmothers:
Do we know the story behind the large delta between peak active and reserved? (Basically I am wondering if this is similar to other cases where we saw that we could get peak reserved down just by constraining the available memory a bit. And if so, could we actually run this on 4x A100? That'd be nice)

recipes/full_finetune_distributed.py
ebsmothers: Bumping this comment. Also I don't fully understand the code comment just below: do we only need to `set_num_threads` in the case of fused optimizer? Looking at the answer.ai code seems like the answer is no, but based on the code comment it's not clear to me

rohan-varma: We see the biggest gain setting this when fused optimizer is used as it controls threads used for intra-op parallelism, and CPU optimizer is the heaviest CPU op. We may have very slight speed ups when not using fused optimizer, but those can be investigated separately of this PR.

torchtune/utils/_distributed.py
```diff
@@ -117,6 +117,22 @@ def init_distributed(**kwargs: Dict) -> bool:  # noqa: DOC106, DOC109
...
+    to be the number of physical CPU cores divided by the number of GPUs as we
+    use one process per GPU, and this avoids CPU oversubscription. Note that this is
+    currently a rough approximation, and doesn't take into account environments where
+    things like CPU affinity is set.
+    """
+    num_threads = os.cpu_count() // (
+        torch.distributed.get_world_size() if torch.distributed.is_initialized() else 1
+    )
+    torch.set_num_threads(num_threads)
+    log.info(f"Set intra op parallelism no. of threads to {num_threads}")
```
ebsmothers, line 133: For some reason the logger in this file is `_log` and not `log`, so this errors out

ebsmothers (approved):
Thanks for working through all the comments here! This is awesome functionality to have and I like the way the UX turned out here. One more comment to address before merging but otherwise looks great!

torchtune/utils/_distributed.py
rohan-varma: Ouch, great catch

andyl98:
Hi @rohan-varma , I was trying some fine-tuning runs with 8x A100-80G setup & 1T memory. The training loop works fine but when entering the checkpoint saving stage, the memory utilization continues to go up until it exceeds my node's limit. Do you know if this is a bug or an intended behavior? Thanks a lot for helping out!
Screenshot 2024-06-05 at 9 13 07 PM

andyl98:
Update: I think I can successfully save the checkpoint after training if I don't checkpoint the optimizer states. Perhaps an option should be added to save memory by not saving optim states?

Screenshot 2024-06-06 at 2 13 20 PM

IvanLukianenko:
Hi @andyl98, your code changes fix the problem in your setup and you can now save checkpoints at the end of epochs? I ran into the same problem in almost the same setup, but your code changes didn't help me. Many thanks for the help!

musabgultekin:
@andyl98 Extra note: I was able to train&save with 4xA100 80GB with 1TB DRAM. It also worked with 8xA100 with 2TB DRAM. Haven't done 8xA100 with 1TB though like in your case, it might be that 1TB DRAM is too small for 8x.

andyl98:
> @andyl98 Extra note: I was able to train&save with 4xA100 80GB with 1TB DRAM. It also worked with 8xA100 with 2TB DRAM. Haven't done 8xA100 with 1TB though like in your case, it might be that 1TB DRAM is too small for 8x.

Right, seems like the model full shard checkpointing stage inevitably consumes quite some memory. Appreciate the note!

ebsmothers:
@andyl98 @musabgultekin sorry for the delay here, I agree it's suspicious that we'd require 1TB to save the 70B model and optimizer states. I created #1092 so we can track it there

PR #994, opened 2024-05-17
Datasets tutorial improvements

RdoubleA:
#### Context
- [ ] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

#### Changelog
- Add needed details around using local datasets, custom instruct/chat datasets, how to use built-in datasets, sample packing, and more all with python/config/CLI examples
- Clean up some config utilities and let users specify any path for custom instruct templates or chat formats. This also needed to update some test utils
- Make column optional in TextCompletionDataset for local datasets
- Remove redundant truncate in tokenizer utils

Sample packing will have a dedicated page, so details on that are limited for now.

#### Test plan
doc build, pytest tests


torchtune/datasets/_instruct.py
```diff
@@ -127,7 +127,7 @@ def instruct_dataset(
         tokenizer (Tokenizer): Tokenizer used to encode data. Tokenize must implement an `encode` and `decode` method.
         source (str): path string of dataset, anything supported by Hugging Face's `load_dataset`
             (https://huggingface.co/docs/datasets/en/package_reference/loading_methods#datasets.load_dataset.path)
-        template (str): class used to format the prompt. If the placeholder variable
+        template (str): full import path of class used to format the prompt. If the placeholder variable
```
joecummings, line 130: While you're here, can you add an example of using this from code and from YAML through the `Examples::` tag - similar to `chat_dataset`. This is helpful for anyone going through the documentation looking at the builder function.

docs/source/tutorials/datasets.rst
```diff
@@ -208,9 +378,6 @@ adjustments for chosen and rejected samples in preference data.
         np.where(r_masks, CROSS_ENTROPY_IGNORE_IDX, rejected_input_ids)
     )
 
-If any of the existing dataset classes do not serve your purposes, you can similarly
-use one of them as a starting point and add the functionality you need.
-
 To be able to use your custom dataset from the config, you will need to create
```
joecummings, line 381: This is incorrect, no? As long as they fit into `chat_dataset` or `instruct_dataset`, they should be able to specify from the config without adding their own class?

joecummings:
Screenshot 2024-05-17 at 9 35 51 AM

Long CLI commands are getting hard to understand. Might be good to break up over multiple lines with backslash operator.

joecummings:
Screenshot 2024-05-17 at 9 39 29 AM

PreferenceDataset is not added to our documentation, therefore the class isn't rendering here.

docs/source/tutorials/datasets.rst
```diff
@@ -108,33 +153,60 @@ Here is an example of sample that is formatted with :class:`~torchtune.data.Alpa
...
+            cls, sample: Mapping[str, Any], column_map: Optional[Dict[str, str]] = None
+        ) -> str:
+            ...
+
+    # Load in tokenizer
+    tokenizer = ...
+    dataset = instruct_dataset(
+        tokenizer=tokenizer,
+        source="my/dataset/path",
+        template="import.path.to.CustomTemplate",
```
pbontrager, line 180: Why is this an import path and not an instance of CustomTemplate?

docs/source/tutorials/datasets.rst
```diff
@@ -108,33 +153,60 @@ Here is an example of sample that is formatted with :class:`~torchtune.data.Alpa
...
+        # Implement this method
+        @classmethod
+        def format(
+            cls, sample: Mapping[str, Any], column_map: Optional[Dict[str, str]] = None
+        ) -> str:
+            ...
+
+    # Load in tokenizer
+    tokenizer = ...
+    dataset = instruct_dataset(
```
pbontrager, line 177: Aside: we have both instruct/chat datasets and templates. Can they only go together? Do we mention this or enforce this?

docs/source/tutorials/datasets.rst
```diff
@@ -108,33 +153,60 @@ Here is an example of sample that is formatted with :class:`~torchtune.data.Alpa
...
 
+    # YAML config
     dataset:
       _component_: torchtune.datasets.instruct_dataset
-      source: mydataset/onthehub
-      template: CustomTemplate
-      train_on_input: True
-      max_seq_len: 512
+      source: my/dataset/path
+      template: import.path.to.CustomTemplate
```
pbontrager, line 189: Same question

docs/source/tutorials/datasets.rst
```diff
@@ -108,33 +153,60 @@ Here is an example of sample that is formatted with :class:`~torchtune.data.Alpa
...
+
+.. code-block:: bash
+
+    # Command line
+    tune run full_finetune_single_device --config llama3/8B_full_single_device dataset=torchtune.datasets.instruct_dataset dataset.source=my/dataset/path dataset.template=import.path.to.CustomTemplate
```
pbontrager, line 194: Same template question

docs/source/tutorials/datasets.rst
```diff
@@ -108,33 +153,60 @@ Here is an example of sample that is formatted with :class:`~torchtune.data.Alpa
...
+
+
+Custom chat dataset and chat formats
```
pbontrager, line 197: I think the breakdown here between instruct and chat is good for now. But some of this information, in the future should be contained in the Chat/Instruct api reference and this tutorial should provide the extra high level information (like how to customize things) about all our dataset formats. This is assuming that the number of formats grows much more.

docs/source/tutorials/datasets.rst
```diff
@@ -172,6 +244,104 @@ Here is how messages would be formatted using the :class:`~torchtune.data.Llama2
...
+        ) -> List[Message]:
+            ...
+
+    # Load in tokenizer
+    tokenizer = ...
+    dataset = chat_dataset(
+        tokenizer=tokenizer,
+        source="my/dataset/path",
+        conversation_style="openai",
+        chat_format="import.path.to.CustomChatFormat",
```
pbontrager, line 273: Same question about templates for all chat_format inputs

docs/source/tutorials/datasets.rst
```diff
@@ -172,6 +244,104 @@ Here is how messages would be formatted using the :class:`~torchtune.data.Llama2
...
+.. code-block:: bash
+
+    # Command line
+    tune run full_finetune_single_device --config llama3/8B_full_single_device dataset=torchtune.datasets.chat_dataset dataset.source=my/dataset/path dataset.conversation_style=openai dataset.chat_format=import.path.to.CustomChatFormat
+
+
+Multiple in-memory datasets
+---------------------------
+
+It is also possible to train on multiple datasets by combining them into a single :class:`~torchtune.datasets.ConcatDataset`.
```
pbontrager, line 294: Is ConcatDataset inferred here somewhere? Does the dataset init know what to do when it sees a list? If so we should make that clear as doing a ConcatDataset in code would look different.

docs/source/tutorials/datasets.rst
```diff
@@ -172,6 +244,104 @@ Here is how messages would be formatted using the :class:`~torchtune.data.Llama2
...
+-------------------------
+
+To use a dataset saved on your local hard drive, simply specify the file type for
+``source`` and pass in the ``data_files`` argument using any of the dataset
+builder functions. We support all `file types <https://huggingface.co/docs/datasets/en/loading#local-and-remote-files>`_
+supported by Hugging Face's ``load_dataset``, including csv, json, txt, and more.
+
+.. code-block:: python
+
+    from torchtune.datasets import instruct_dataset
```
pbontrager, line 324: This should include a yaml example too right?

docs/source/tutorials/datasets.rst
```diff
@@ -172,6 +244,104 @@ Here is how messages would be formatted using the :class:`~torchtune.data.Llama2
...
+        source="csv",
+        template="import.path.to.CustomTemplate"
+        data_files="path/to/my/data.csv",
+    )
+    # Remote files
+    dataset = instruct_dataset(
+        tokenizer=tokenizer,
+        source="json",
+        template="import.path.to.CustomTemplate"
+        data_files="https://rajpurkar.github.io/SQuAD-explorer/dataset/train-v2.0.json",
```
pbontrager, line 340: Are we able to directly point to a HF hosted dataset here?

docs/source/tutorials/datasets.rst
pbontrager: Even if it's not an Instruct dataset, I'd believe you could directly instantiate your custom dataset class without a builder.

torchtune/datasets/_chat.py
```diff
@@ -10,7 +10,7 @@
 
 from datasets import load_dataset
 from torch.utils.data import Dataset
-from torchtune.config._utils import _get_chat_format
```
pbontrager, line 13: I guess this gets to the heart of half of the comments I left earlier. My concern with this is we're creating two ways to do the same things in the config. Is this meant to be a way to do nested instantiation from yaml without supporting recursive instantiation?

pbontrager (commented):
Really great tutorial updates. I've left some comments and questions, some of them just because this was my first time going through the tutorial. One concern I have with the way we're doing deepdives/tutorials is that we're putting a lot of important information here while our API references are pretty sparse. I'm afraid that people won't take the time to go through the tutorial and jump to what they're looking for in the api reference section.

docs/source/tutorials/datasets.rst
RdoubleA: good point. we only enforce it through typing. but it might be good to do an explicit check.

docs/source/tutorials/datasets.rst
RdoubleA: because then I have to do a nested target in the config, which requires recursive instantiation. something we currently don't support.

In the `InstructDataset` class itself you pass in an instance of CustomTemplate, but the builders were made to be config friendly, hence the import path

docs/source/tutorials/datasets.rst
RdoubleA: Yeah make sense, I would say the level of detail here is sufficient for tutorial / general usage but anything more should go in the API reference. lmk if I should generalize this a bit more

docs/source/tutorials/datasets.rst
RdoubleA: The recipes were updated to handle init-ing a list. I can add that code snippet here, but frankly we're in the process of migrating to IterableDatasets and `ConcatDataset` may change significantly

docs/source/tutorials/datasets.rst
RdoubleA: what do you mean, like a dataset that's not on the hub?

docs/source/tutorials/datasets.rst
RdoubleA: Let me clarify this. This really applies to the case of `InstructDataset` and `ChatDataset` which use nested components which is difficult to specify from the config. Hence, the builder functions. I made the assumption that any new dataset class would follow the same pattern.

torchtune/datasets/_chat.py
RdoubleA: > Is this meant to be a way to do nested instantiation from yaml without supporting recursive instantiation?

ding ding ding

RdoubleA: it's a fair concern. this is my workaround for recursive instantiation for now... may need to revisit this if lack of recursive instantiation becomes increasingly problematic, especially as datasets become more complex with multimodal and iterable. One of the reason we did not want to use hydra was that recursive instantiation is a bit too powerful and may lead to coding via config (cc @ebsmothers). Happy to chat more offline if you want to go into that rabbit hole :)

docs/source/tutorials/datasets.rst
pbontrager: I meant if it's possible to use this setup with remote hugging face datasets, just by providing the name/url. It would be a good example to show if it's supported.

docs/source/tutorials/datasets.rst
RdoubleA: makes sense, I'll remove mention of it but keep the yaml example

docs/source/tutorials/datasets.rst
pbontrager: I don't know how I feel about this, it feels like we're saying we don't do recursion but then we do this slight of hand to get around it. The alternative would be that we instantiate the template in the recipe and pass it into the dataset.

RdoubleA: Yeah I'm not a fan of doing this, but instantiating it in the recipe means any user that is writing their own script will have to figure this out. Until we find an alternative to this sleight of hand, I would much rather handle it for the user. Happy to chat more about a more long-term solution

pbontrager (approved):
Thanks for doing this and updating from all the comments.

PR #1005, opened 2024-05-19
RLHF with PPO

### LINKED ISSUE #812: [RFC] Proximal Policy Optimisation

SalmanMohammadi:
#### Context
- [x] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

https://github.com/pytorch/torchtune/issues/812

### Background reading:

[The N Implementation Details of RLHF with PPO, Huang et al.](https://iclr-blogposts.github.io/2024/blog/the-n-implementation-details-of-rlhf-with-ppo)
[The N+ Implementation Details of RLHF with PPO: A case study on TL;DR Summarization](https://arxiv.org/pdf/2403.17031)
The original RLHF paper - [Fine-Tuning Language Models from Human Preferences, Ziegler et al.](https://arxiv.org/pdf/1909.08593)
Anthropic's RLHF paper - [Training a Helpful and Harmless Assistant with Learning from Human Feedback](https://arxiv.org/pdf/2204.05862)
[Training language models to follow instructions with human feedback, Ouyang et al.](https://arxiv.org/pdf/2203.02155)
Shameless plug, but I would have genuinely found this post helpful when I started out with PPO, even for skimming through some of the references - [The theory of Proximal Policy Optimization implementations](https://salmanmohammadi.github.io/content/ppo/)

### Changelog:
- **Implemented LoRA PPO recipe**
	- *Changes*
		- I used https://github.com/huggingface/trl/blob/main/trl/trainer/ppov2_trainer.py#L373, https://github.com/huggingface/trl/blob/d1aa0b6b2c8dfd78c0f771759d1ff2469c0e5ed2/trl/trainer/ppo_trainer.py and https://github.com/vwxyzjn/lm-human-preference-details/blob/main/lm_human_preference_details/train_policy_accelerate.py as references.
	- *Tests:*
		- TODO
- **Refactored TransformerDecoder**
	- *Changes*
		- The following changes were added as a submodule in `torchtune.models.mistral`:
			- Changed `TransformerDecoder` to `TransformerDecoderWithHiddenLayer`
			- Added `TransformerLM` which wraps an output projection around `TransformerDecoderWithHiddenLayer`
				- Component and model builders now return `TransformerLM`
			- Added `TransformerLMWithValueHead`, with two linear projections: one for the LM head and one for the value head.
		- Support for checkpointing with models wrapping around `TransformerDecoderWithHiddenLayer` added
		- Updated checkpointing to correctly convert HF weights to refactored models
		- Added support for checkpointing value heads.
	- *Tests:*
		- TODO
- **Added mistral value head** models
	- *Changes:*
		- Added model and component builders  
	- *Tests*
		- The implementation is identical to `MistralClassifier`.
- **Added PPOLoss**
	- _Tests_:
		- `test_ppo_loss` tests for correct behaviour based on expected relative value and policy loss for different inputs.
- **Added `utils.ppo_utils` for various ppo utils, and tests for all files including**:
	- `_generation.py`
		- Added `custom_generate_next_token` functions for generating with value head models, and for generating with masks and input positions.
		- Added `get_causal_masks` for creating masks of shape `[bsz, seq_len, seq_len]` which correctly mask leading padding tokens, suitable for use with `scaled_dot_product_attention`.
		- Added a custom `generate` function which generates sequences using above functionality.
	- `collate.py`
		- Added support for collating input sequences by left-padding to a specified maximum sequence length.
	- `rewards.py`
		- Support for calculation of rewards, advantage estimation, adaptive and fixed KL controllers, and reward normalisation.
### TODO:

- [ ] Complete or remove all `TODO (@SalmanMohammadi)`
- [ ] Remove temporary changes for MPS support.
- [ ] Upload randomly initialised models and write recipe tests for:
     - [ ] Verifying recipe checkpointing works correctly w.r.t. saving and loading base model and value head weights.
     - [ ] Verifying expected loss values.
- [ ] Run full model training and verify loss curves
- [ ] Add support for reward models and base models using different tokenizers.
- [ ] Add docs to api reference.

Adding this to open up discussion and get some feedback (@kartikayk) while I train models and verify correctness. Maybe a good place to start would be the [TransformerDecoder](https://github.com/pytorch/torchtune/blob/466b6830c179d280362631f3789c42f22a052287/torchtune/models/mistral/modules/transformer.py) refactor?

closes #812 


tests/recipes/test_ppo_single_device.py
SalmanMohammadi, line 1: TODO

recipes/lora_ppo_single_device.py
```diff
@@ -30,17 +30,9 @@
     validate_state_dict_for_lora,
 )
 from torchtune.recipe_interfaces import FTRecipeInterface
+from torchtune.utils import ppo as ppo_utils
```
SalmanMohammadi, line 33: perhaps ppo_utils should be a standalone module under the `torchtune` namespace? or just under the `torchtune.utils.ppo_utils` namespace so we have either from `torchtune import ppo_utils`, or `from torchtune.utils import ppo_utils`.

torchtune/utils/ppo/_generation.py
```diff
@@ -0,0 +1,185 @@
...
+    if stop_tokens is not None:
+        # convert stop tokens to tensor for easy matching
+        stop_tokens = (
+            torch.tensor(stop_tokens, device=prompt.device) if stop_tokens else None
+        )
+        # keeps track at a high level if we've already hit a stop token in a sequence so we can early stop
+        stop_token_reached = torch.zeros(bsz, dtype=torch.bool, device=prompt.device)
+        # everything in stop_token_mask starts as 1s, and we'll set them to 0 for sequences
+        # that already hit a stop token
+        stop_token_mask = torch.ones(
```
SalmanMohammadi, line 134: I don't actually need to support stop tokens here atm.

torchtune/utils/ppo/_generation.py
```diff
@@ -0,0 +1,185 @@
...
+        # update stop_token_mask if we reached a stop token in a previous step
+        # by appending the logical not of stop_token_reached to the end of the mask
+        # reshaped to be bsz first
+        if stop_tokens is not None:
+            stop_token_mask = torch.cat(
+                [stop_token_mask, ~stop_token_reached.reshape(bsz, 1)], dim=-1
+            )
+
+        padding_mask = generated_tokens == pad_id
+        mask = get_causal_mask(generated_tokens, padding_id=pad_id, dtype=dtype)
```
SalmanMohammadi, line 151: I can probably be more efficient about this, we can update position IDs and attention masks by just adding the next position id per batch, and the next column+row to the attention mask, without recalculating from scratch.

recipes/lora_ppo_single_device.py
SalmanMohammadi: going to go with the latter.

torchtune/utils/ppo/_generation.py
SalmanMohammadi: Saying that, a more recent implementation of RLHF [from OpenAI](https://github.com/openai/summarize-from-feedback/blob/8af822a428c93432aa80ffbe5b065a8f93895669/summarize_from_feedback/utils/experiment_helpers.py#L19) does stop sampling when hitting EOS tokens. I'll need to re-work things slightly to make things work with truncation during generation, since I use the truncation masks for rejection sampling, and I use the attention masks and position IDs created when generating during PPO optimisation steps.

torchtune/utils/ppo/_generation.py
SalmanMohammadi: adding to this, I can also just return the calculated input ids and masks here rather than recalculate them in the recipe
 - or even pre-calculate them before generation, since we know what the masks and pos ids for the responses will be (sequential from the pos id of the last token in the sequence, and no tokens masked out for attention)

recipes/lora_ppo_single_device.py
```diff
@@ -0,0 +1,777 @@
...
+        # setup training params:
+        # num_steps - global number of optimisation steps (batches)
+        # batch_size - number of samples in a batch
+        # ppo_epochs - number of epochs to optimise the policy over a batch of episodes
+        # ppo_batch_size - number of minibatches (sampled from a single batch) to optimise the policy over
+        self.num_steps = cfg.num_steps
+        self.batch_size = cfg.batch_size
+        self.ppo_epochs = cfg.ppo_epochs
+        self.ppo_batch_size = cfg.ppo_batch_size
+        self._gradient_accumulation_steps = cfg.gradient_accumulation_steps
```
SalmanMohammadi, line 256: remove

torchtune/models/mistral/modules/transformer.py
```diff
@@ -0,0 +1,258 @@
...
+
+import torch
+from torch import nn, Tensor
+
+from torchtune.modules import TransformerDecoderLayer
+from torchtune.modules.kv_cache import KVCache
+from torchtune.modules.transformer import _get_clones
+
+
+class TransformerDecoderWithHiddenOutput(nn.Module):
```
ebsmothers, line 16: Personally I find the name a bit confusing, to me it implies that we should be able to access the hidden layers which we actually can't (I know there is other work ongoing with the transformer decoder refactor so if this is covered there feel free to ignore)

torchtune/models/mistral/modules/transformer.py
```diff
@@ -0,0 +1,258 @@
...
+            # shape: [b, s, d]
+            h = layer(h, mask=mask, input_pos=input_pos)
+
+        # shape: [b, s, d]
+        h = self.norm(h)
+
+        return h
+
+
+class TransformerLMWithValueHead(nn.Module):
```
ebsmothers, line 155: This class can just be a builder version of `TransformerLM` with appropriate choice of `output` (and flexibility of output return types), right? (Similar caveat to my above comment btw)

torchtune/modules/loss/ppo.py
```diff
@@ -0,0 +1,98 @@
...
+# LICENSE file in the root directory of this source tree.
+
+from typing import Optional, Tuple
+
+import torch
+import torch.nn as nn
+from torchtune.utils import ppo_utils
+
+
+class PPOLoss(nn.Module):
```
ebsmothers, line 14: If this loss does not have any trainable parameters I would be more inclined to keep it as a function (it's possible that I'm missing something obvious here though, if so please correct me)

torchtune/modules/loss/ppo.py
```diff
@@ -0,0 +1,98 @@
...
+                - policy_loss: The policy function loss.
+                - value_loss: The value function loss.
+        """
+        ratio = torch.exp(pi_logprobs - pi_old_logprobs)
+        clipped_ratios = torch.clamp(ratio, 1.0 - self.epsilon, 1.0 + self.epsilon)
+
+        policy_losses = -advantages * clipped_ratios
+
+        # taking max instead of min to minimise loss
+        policy_loss = torch.max(policy_losses, -advantages * ratio)
```
ebsmothers, line 78: Shouldn't the second arg to torch.max be an int? (Again apologies if I'm missing something obvious here)

torchtune/utils/ppo_utils/collate.py
```diff
@@ -0,0 +1,34 @@
...
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+from typing import Dict, List
+
+import torch
+from torch.nn.utils.rnn import pad_sequence
+
+
+def left_padded_collate(
```
ebsmothers, line 13: Personally I wonder whether this is a sufficiently general function to just add to `torchtune/utils/collate.py`. The one thing I could see being a bit confusing is that the return signatures differ from our existing `padded_collate`

torchtune/utils/pooling.py
```diff
@@ -4,38 +4,23 @@
...
-) -> Tensor:
-    """Pool sequence logits by selecting the predicted logits for the last non-padding token
-    for each sequence in the batch.
+def get_last_non_masked_token(mask: torch.Tensor, dtype=torch.long) -> torch.Tensor:
+    """Returns the index for the last unmasked entry for each row of a 2D boolean mask.
     Args:
-        tokens (Tensor): input tensor with shape [b x s]
-        logits (Tensor): predicted logits for input tokens with shape [b x s x n]
-        padding_token_idx (int): Padding token id used in the tokenizer.
+        mask (torch.Tensor): Boolean mask with shape [b x s]
```
ebsmothers, line 12: nit: explicitly say that true means masked and false means unmasked

torchtune/modules/loss/ppo.py
```diff
@@ -0,0 +1,98 @@
...
+        gamma (float): Discount factor.
+        lmbda (float): lambda parameter for GAE.
+        epsilon (float): clipping parameter for PPO.
+    """
+
+    def __init__(
+        self,
+        gamma: float = 1.0,
+        lmbda: float = 0.95,
+        epsilon: float = 1e-5,
```
ebsmothers, line 33: Why such a small default value for epsilon? In this case won't we always just clamp the ratio to almost identically 1?

torchtune/modules/loss/ppo.py
```diff
@@ -0,0 +1,98 @@
...
+
+    Using:
+        https://github.com/vwxyzjn/lm-human-preference-details/blob/ccc19538e817e98a60d3253242ac15e2a562cb49/lm_human_preference_details/train_policy_accelerate.py#L719
+        https://github.com/openai/baselines/blob/ea25b9e8b234e6ee1bca43083f8f3cf974143998/baselines/ppo2/model.py#L68-L75
+
+    as references.
+    Args:
+        gamma (float): Discount factor.
+        lmbda (float): lambda parameter for GAE.
+        epsilon (float): clipping parameter for PPO.
```
ebsmothers, line 26: missing a couple args in the docstring here

ebsmothers (commented):
Probably review 1/N here.. took a first pass and left a bunch of basic comments. Will probably take another look tomorrow at some point as well. For anyone else reviewing I have not gotten into detail on the training loop or the generation logic much yet, so that'd be a good place to start. Sorry it took so long to get to this, but hopefully a partial review is better than no review 😅

torchtune/modules/loss/ppo.py
SalmanMohammadi: great catch! i've always used torch.max for both (not sure if this is the right ref. to the internal code https://github.com/pytorch/pytorch/blob/217aac96d779841666527402fc113493b6bd6323/torch/onnx/symbolic_opset7.py#L38), but it looks like `torch.max` supports both. i'll change it to `torch.maximum` for clarity.

torchtune/modules/loss/ppo.py
SalmanMohammadi: yeah, eps is usually 0.1/0.2

recipes/lora_ppo_single_device.py
```diff
@@ -0,0 +1,813 @@
...
+            cfg.reward_checkpointer,
+            # reward model is never being trained
+            # base model checkpointer handles recipe state
+            resume_from_checkpoint=False,
+        )
+        # load base model checkpoint
+        model_checkpoint_dict = self._checkpointer.load_checkpoint()
+
+        # load reward model checkpoint
+        reward_model_checkpoint_dict = self._reward_checkpointer.load_checkpoint()
```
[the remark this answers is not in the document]
SalmanMohammadi: image

you should be able to use arbitrary reward + base language model to be optimized pairs

recipes/configs/mistral/7B_lora_ppo.yaml
```diff
@@ -0,0 +1,151 @@
...
+# Tokenizer
+tokenizer:
+  _component_: torchtune.models.mistral.mistral_tokenizer
+  path:  ./target/weights/mistral_base/tokenizer.model
+
+# Dataset
+dataset:
+  _component_: torchtune.datasets.text_completion_dataset
+  source: nvidia/HelpSteer
+  max_seq_len: 32
```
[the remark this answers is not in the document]
SalmanMohammadi: currently looking a bit more into this, i think more recent implementations sample much longer seq. lengths 
the original justification is in https://arxiv.org/pdf/1909.08593 sec. 3.1
 
> We sample excerpts with lengths of 32 64 tokens, and the policy generates 24 additional tokens. We set the temperature of the pretrained model to T = 0.7 as described in section 2.1.

recipes/lora_ppo_single_device.py
```diff
@@ -0,0 +1,813 @@
...
+
+        if enable_activation_checkpointing:
+            utils.set_activation_checkpointing(
+                model, auto_wrap_policy={TransformerLMWithValueHead}
+            )
+
+        # add placeholder value head key to base model state dict keys
+        # to ensure wrapped model is validated correctly
+        model_state_dict_keys = list(model_state_dict.keys())
+        model_state_dict_keys += ["value_head.weight", "value_head.bias"]
```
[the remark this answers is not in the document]
SalmanMohammadi: I've been reworking this whole bit since the lora finetuning recipe does things a little differently - I grabbed my lora validation from the lora dpo recipe. agree w your points, the updated implementation should be much cleaner

recipes/lora_ppo_single_device.py
```diff
@@ -0,0 +1,813 @@
...
+            reward_model = config.instantiate(cfg_reward_model)
+
+        self._lora_rank = cfg_model.lora_rank
+        self._lora_alpha = cfg_model.lora_alpha
+        self.adapter_params = get_adapter_params(model)
+        set_trainable_params(model, self.adapter_params)
+
+        if enable_activation_checkpointing:
+            utils.set_activation_checkpointing(
+                model, auto_wrap_policy={TransformerLMWithValueHead}
```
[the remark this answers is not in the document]
SalmanMohammadi: yep - also caught this, will be included in my setup_model rework

recipes/lora_ppo_single_device.py
```diff
@@ -0,0 +1,813 @@
...
+                query_responses,
+                masks,
+                padding_masks,
+                value_padding_masks,
+                values,
+            )
+            if self._device.type == "cuda":
+                torch.cuda.empty_cache()
+            elif self._device.type == "mps":
+                torch.mps.empty_cache()
```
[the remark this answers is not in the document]
SalmanMohammadi: agreed : )

recipes/lora_ppo_single_device.py
```diff
@@ -0,0 +1,813 @@
...
+        self.reward_penalty = cfg.reward_penalty
+
+        self.total_epochs = self.num_steps // self.batch_size
+        self._gradient_accumulation_steps = cfg.gradient_accumulation_steps
+        # one "step" is an update over a batch of trajectories
+        self._steps_per_epoch = len(self._dataloader) // self.batch_size
+        self.global_step = self.epochs_run * self._steps_per_epoch
+
+        # setup adaptive KL controller
+        self.kl_controller = config.instantiate(cfg.kl_controller)
```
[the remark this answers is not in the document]
SalmanMohammadi: agreed!

re: your second point, this is also something i've been thinking a lot about. implementing ppo for rlhf has been eye-opening for me, i totally see why open-source much prefers DPO, it's so simple. policy gradient algorithms can be a real pain in the ass to get working, but when they do they're super powerful. there's also evidence that ppo can perform better than dpo, but there just arent that many accurate, performant, and easy to use implementations out there (aside from TRL which has been a huge help for me). I'd like to include a cookbook/deep dive and include lots of resources to help people use this technique - your suggestion about minimal docs in the config is great for the interim

torchtune/models/mistral/modules/transformer.py
SalmanMohammadi: yep agreed - i think for the most part you can ignore my transformerdecoder work here since it'll be removed once I merge in the transformerdecoder refactor.

torchtune/models/mistral/modules/transformer.py
SalmanMohammadi: yeah! hopefully having worked examples here has helped inform how we can think about the transformerdecoder refactor

torchtune/modules/loss/ppo.py
SalmanMohammadi: there's no trainable parameters - this was mainly to ensure consistency with the other configs in that there's a callable loss component which is instantiated. in practice there's no other loss you're going to be using with this recipe though, so we can probably drop this format here

recipes/lora_ppo_single_device.py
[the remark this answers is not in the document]
SalmanMohammadi: ^ in practice you want to ensure the two models have some SFT so they're both on-distribution for the domain you're interested in fine-tuning for. the extreme might be e.g. using llama2 as the base model and a gemma reward mdoel, but you could also use a 13B base model and a smaller 7B reward model from the same family.

what has been shown to work is that you start with some SFT-trained model, train a reward model from this, and then fine-tune from the same baseline SFT using RLHF and the trained reward model, but it'd be cool to give people flexibility around this

recipes/lora_ppo_single_device.py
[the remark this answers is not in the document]
ebsmothers: Yeah this makes a ton of sense. Right now our alignment story is probably the most half-baked part of the library. Even our DPO recipe, which imo is super useful, is not really surfaced in any of our docs. Longer-term I would love something like what you're suggesting. If we really want to claim to support these methods properly they deserve an entire section in our live docs with explanations of the algorithms, how-to guides, etc. I read your PPO explainer (and realistically will probably read it a few more times while reviewing this code..) and found it really well-written. I think we would love to have your help on providing proper documentation for users who are interested in these algorithms.

torchtune/modules/loss/ppo.py
ebsmothers: Ah right, the instantiation logic.. in that case I am OK to leave it like this. Actually for this kind of stuff what I like to do is keep the loss as a function and do something like this in the config

```
loss:
  _component_: functools.partial
  func: torchtune.modules.loss.PPOLoss
  gamma: 1
  lmbda: 0.95
  epsilon: 0.2
  value_coeff: 0.1
  value_clip_range: 0.1
```

Caveat: I haven't actually tried this in the torchtune config system so it may not actually work. Also more importantly I think most people will find this unintuitive.

recipes/lora_ppo_single_device.py
```diff
@@ -0,0 +1,813 @@
...
+        cfg_lr_scheduler: DictConfig,
+        num_training_steps: int,
+        last_epoch: int,
+    ) -> Optimizer:
+        lr_scheduler = config.instantiate(
+            cfg_lr_scheduler,
+            self._optimizer,
+            num_training_steps=num_training_steps,
+            last_epoch=last_epoch,
+        )
```
ebsmothers, line 383: Seems you aren't actually using an LR scheduler, can this be deleted?

torchtune/utils/ppo_utils/_generation.py
```diff
@@ -0,0 +1,182 @@
...
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+from typing import Optional
+
+import torch
+from torchtune.utils._generation import sample, update_stop_tokens_tracker
+
+
+def generate_next_token(
```
ebsmothers, line 13: Is the only difference between this and the existing `generate_next_token` the inclusion of the `mask` parameter? If so I'd be inclined to just add it to the existing function

torchtune/utils/ppo_utils/_generation.py
```diff
@@ -0,0 +1,182 @@
...
+    top_k: int = None,
+) -> torch.Tensor:
+    """Generates the next tokens."""
+    # model produces logits in [bsz, seq_length, vocab_size]
+    # we want to take the last token's logits as the input to the next model call
+    logits = model(x, input_pos=input_pos, mask=mask)[:, -1]
+    return sample(logits, temperature, top_k)
+
+
+def generate_next_token_with_value_head_model(
```
ebsmothers, line 29: Similar comment here -- is the purpose of this just to unpack the tuple returned by the value head model? If so I wonder if it's more appropriate to unpack directly in the recipe with an appropriate code commment

recipes/lora_ppo_single_device.py
```diff
@@ -0,0 +1,813 @@
...
+                    )
+                    valid_response_lengths = valid_response_lengths.clip(
+                        0, self.max_generated_tokens - 1
+                    )
+                    value_padding_masks[
+                        torch.arange(input_ids.shape[0]), valid_response_lengths
+                    ] = False
+                    values = values[:, context_length - 1 : -1].squeeze(
+                        -1
+                    )  # [b, max_generated_tokens]
```
ebsmothers, line 613: Sorry I'm not sure I fully follow this block. What exactly are we aiming to mask here? Also, I see the `context_length - 1 : -1` slice coming up quite a bit below as well, is this just the last padding token plus all the generated tokens, stopping just short of the eos token?

recipes/lora_ppo_single_device.py
```diff
@@ -0,0 +1,813 @@
...
+
+                            pi_logits = pi_logits[:, context_length - 1 : -1]
+                            pi_logits /= self.temperature
+                            pi_logprobs = torch.gather(
+                                F.log_softmax(pi_logits, dim=-1),
+                                2,
+                                backward_query_responses[:, context_length:].unsqueeze(
+                                    -1
+                                ),
+                            ).squeeze(-1)
```
ebsmothers, line 702: I see this pattern in several places, worth factoring out as a utility?

ebsmothers (commented):
OK I've finally made it through all of the core code (admittedly haven't looked at the unit tests yet). Sorry it's taken so long, this is a serious PR! I probably will have more informed comments as I go through this a second time. 

In the meantime, one question about the structure of the recipe itself: do you think there is any possible logical separation of the training loop into N subcomponents? I notice in the lm-human-preference-details they enumerate six different pieces in their code comments (and this is before they even get to the actual PPO update logic). I think the algo is involved enough that if we can split out into a relatively small handful of clearly-delineated methods it'd make the core `train` method much easier to reason about. I think you've done some of this already with the generation utilities, but wonder if we could also separate out e.g. masking logic, trajectory generation, actual PPO updates, etc into their own methods inside the recipe which get called from the main train method. (Btw his may be a pipe dream and I may just be a bit naive about the complexity of the algo, tell me if so!)

torchtune/utils/ppo_utils/__init__.py
joecummings, line 1: A lot of these things are really just relevant for a variety of RLHF varations - maybe worth renaming to the more general and then we can add other things here too?

torchtune/utils/ppo_utils/_generation.py
```diff
@@ -0,0 +1,182 @@
...
+    mask = torch.tril(
+        torch.ones(seq_len, seq_len, device=padding_mask.device, dtype=bool), diagonal=0
+    )
+    mask = mask & (padding_mask[:, None, :] & padding_mask[:, :, None])
+    mask.diagonal(dim1=1, dim2=2)[:] = True
+    return mask
+
+
+@torch.inference_mode()
+def generate(
```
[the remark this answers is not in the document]
joecummings: + 1

torchtune/utils/ppo_utils/collate.py
joecummings: Yeah there's already a padded_collate_dpo, we should add this to our collator functions, but also make the signature match` e.g. Dict[str, Tensor]`

torchtune/utils/ppo_utils/kl_controller.py
```diff
@@ -0,0 +1,56 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+import numpy as np
+
+
+class AdaptiveKLController:
```
joecummings, line 10: @ebsmothers  Should this be an nn.Module?

torchtune/utils/ppo_utils/rewards.py
```diff
@@ -0,0 +1,205 @@
...
+
+    return total_reward, kl, kl_reward
+
+
+def whiten(x: torch.Tensor, shift_mean: bool = True) -> torch.Tensor:
+    """
+    Whitens (normalizes) the input tensor.
+
+    Args:
+        advantages (torch.Tensor): The advantages.
```
joecummings, line 59: Mismatch here with the args.

torchtune/utils/precision.py
joecummings, line 1: I think this snuck in from another PR?

torchtune/utils/constants.py
```diff
@@ -22,3 +22,5 @@
 # total number of epochs for training; resumed training runs for
 # (total_epochs - epochs_run) number of epochs
 TOTAL_EPOCHS_KEY = "total_epochs"
+# value head key
+VALUE_HEAD_KEY = "value_head"
```
joecummings, line 26: Can you add a little more context for where this value_head is relevant?

torchtune/models/mistral/modules/transformer.py
joecummings, line 1: Not looking too hard at this file b/c stated that author will rebase with TransformerDecoder refactor: #1017

torchtune/models/mistral/_component_builders.py
```diff
@@ -619,3 +625,413 @@ def lora_mistral_classifier(
...
+        head_dim=(embed_dim // num_heads),
+        norm=RMSNorm(embed_dim, eps=norm_eps),
+    )
+    if quantize_base:
+        # For QLoRA, we reparametrize 4-bit tensors to higher precision, and offload to CPU on the fly
+        # so as to not increase peak memory
+        decoder._register_state_dict_hook(
+            partial(
+                reparametrize_as_dtype_state_dict_post_hook,
+                # TODO this is clowny, figure out a better way to get what precision the rest
```
joecummings, line 825: Can't wait to still be seeing this comment in 3 years @ebsmothers

joecummings (commented):
Can't wait to try this out!!! 

Can you update the docs, too?

torchtune/utils/ppo_utils/_generation.py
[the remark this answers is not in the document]
SalmanMohammadi: Hmmm, I had a quick try and looks like you're right. Was this added recently at all? I feel like when I tried it out I couldn't get it to work because generation required KV caches to be setup, and the causal mask used when setting up cacheing didn't account for padding? Maybe I'm misremembering.

tests/torchtune/modules/loss/test_ppo_loss.py
```diff
@@ -0,0 +1,85 @@
...
+        )
+
+    def test_policy_loss_clipped_for_high_logprobs(self, loss_fn):
+        # fixed old policy logprobs, advantages, returns
+        pi_old_logprobs = torch.tensor([0.5, 0.8, 1.2])
+        advantages = torch.tensor([1.0, 1.0, 1.0])
+        values = torch.tensor([1.0, 1.0, 1.0])
+        returns = torch.tensor([1.0, 1.0, 1.0])
+
+        pi_logprobs_high = torch.tensor([0.5, 0.8, 1.2])
```
SalmanMohammadi, line 35: redo - not sure how this ended up here

torchtune/utils/precision.py
SalmanMohammadi: nah - I've rolled mps support in this branch to allow me to run small experiments locally. I think we're still waiting on the MPS PR. I'll merge it in if it's done by the time this is ready, or otherwise take the MPS changes out

recipes/lora_ppo_single_device.py
```diff
@@ -0,0 +1,813 @@
...
+
+                    del eos_mask, scores, valid_response_lengths
+
+                # trajectory generated! time to optimise
+                approx_policy_kls = []
+                losses = []
+                policy_losses = []
+                value_losses = []
+                entropies = []
+                for _ in range(self.ppo_epochs):
```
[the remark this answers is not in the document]
vwxyzjn: Yeah PPO epoch is related to the batch of data, not like an epoch of the entire dataset.

```
import numpy as np
batch_size = 8
nminibatches = 2
gradient_accumulation_steps = 2
mini_batch_size = batch_size // nminibatches
micro_batch_size = mini_batch_size // gradient_accumulation_steps
data = np.arange(batch_size).astype(np.float32)
print("data:", data)
print("batch_size:", batch_size)
print("mini_batch_size:", mini_batch_size)
print("micro_batch_size:", micro_batch_size)
for epoch in range(4):
    batch_inds = np.random.permutation(batch_size)
    print("epoch:", epoch, "batch_inds:", batch_inds)
    for mini_batch_start in range(0, batch_size, mini_batch_size):
        mini_batch_end = mini_batch_start + mini_batch_size
        mini_batch_inds = batch_inds[mini_batch_start:mini_batch_end]
        
        # `optimizer.zero_grad()` set optimizer to zero for gradient accumulation
        for micro_batch_start in range(0, mini_batch_size, micro_batch_size):
            micro_batch_end = micro_batch_start + micro_batch_size 
            micro_batch_inds = mini_batch_inds[micro_batch_start:micro_batch_end]
            print("____⏩ a forward pass on", data[micro_batch_inds])
        # `optimizer.step()`
        print("⏪ a backward pass on", data[mini_batch_inds])

# data: [0. 1. 2. 3. 4. 5. 6. 7.]
# batch_size: 8
# mini_batch_size: 4
# micro_batch_size: 2
# epoch: 0 batch_inds: [6 4 0 7 3 5 1 2]
# ____⏩ a forward pass on [6. 4.]
# ____⏩ a forward pass on [0. 7.]
# ⏪ a backward pass on [6. 4. 0. 7.]
# ____⏩ a forward pass on [3. 5.]
# ____⏩ a forward pass on [1. 2.]
# ⏪ a backward pass on [3. 5. 1. 2.]
# epoch: 1 batch_inds: [6 7 3 2 0 4 5 1]
# ____⏩ a forward pass on [6. 7.]
# ____⏩ a forward pass on [3. 2.]
# ⏪ a backward pass on [6. 7. 3. 2.]
# ____⏩ a forward pass on [0. 4.]
# ____⏩ a forward pass on [5. 1.]
# ⏪ a backward pass on [0. 4. 5. 1.]
# epoch: 2 batch_inds: [1 4 5 6 0 7 3 2]
# ____⏩ a forward pass on [1. 4.]
# ____⏩ a forward pass on [5. 6.]
# ⏪ a backward pass on [1. 4. 5. 6.]
# ____⏩ a forward pass on [0. 7.]
# ____⏩ a forward pass on [3. 2.]
# ⏪ a backward pass on [0. 7. 3. 2.]
# epoch: 3 batch_inds: [7 2 4 1 3 0 6 5]
# ____⏩ a forward pass on [7. 2.]
# ____⏩ a forward pass on [4. 1.]
# ⏪ a backward pass on [7. 2. 4. 1.]
# ____⏩ a forward pass on [3. 0.]
# ____⏩ a forward pass on [6. 5.]
# ⏪ a backward pass on [3. 0. 6. 5.]
```

torchtune/utils/ppo_utils/__init__.py
SalmanMohammadi: hows `utils/rlhf/`?

pbontrager: Just a fly by comment, but if you're changing the folder name anyway, this should really be in `modules/rlhf`. Utils are primarily around training (checkpoints, distributed, precision, etc) while modules are components/tools for algorithms (loss, peft, transforms, etc)

SalmanMohammadi:
Thanks so much for all the comments and feedback @joecummings @ebsmothers @pbontrager 

Just a quick update and q, I'll have a bit more time later this week to push some refactors to address your comments.

I'm seeing pretty similar behaviour to @vwxyzjn's PPOV2 implementation in TRL using Pythia1B. I'll do a little more testing before I run a full-length training run (my runpods have been getting interrupted D :), but here's an example of an [experiment on TinyLlama2-1B](https://wandb.ai/salman-mohammadi/torchtune_ppo/runs/0gtxfjj9/workspace?nw=nwusersalmanmohammadi): 

image


Reward goes up; loss goes down. The scale of my PPO losses are a bit higher than the TRL run here https://wandb.ai/costa-huang/huggingface/runs/fmof4oxq/workspace?nw=nwusercostahuang - not sure if you've experienced this before @vwxyzjn?

Thinking a bit ahead, I don't think I'll need the TransformerDecoder refactor to merge this in, as I've switched to training with separate value and policy models as per https://github.com/huggingface/trl/issues/1783. It's something I'll maybe try experimenting more in the future with as a memory saving technique. In terms of experimental results, what would you guys ideally like to see? I probably only have the funds to do training runs on the 1B models I've been working with atm. If there's a possibility of you guys spinning up a couple runs, it'd be cool to ensure the recipe scales correctly to Llama7B, or Mistral7B.

torchtune/modules/loss/ppo.py
SalmanMohammadi: I personally like this suggestion, but think it's easier to interpret to keep it as a class. If we don't like inheriting from `nn.Module`, I could just replace `forward` with `__call__`? Not sure if that would also be confusing to some people.

torchtune/utils/ppo_utils/collate.py
SalmanMohammadi: I'm happy to make this general if you think it's of use elsewhere. It's mainly like this because a) I was being lazy and didn't want to handle logic for padding label/input IDs of differing lengths, b) we only ever want prompt inputs for PPO where this was being used. 
I could move the core logic from this function to `left_pad` and add a `ppo_collate` which only applies left-padding to the input tokens. If someone wanted to use this kind of padding they could then re-use the underlying logic by defining their own collate function.

torchtune/utils/constants.py
SalmanMohammadi: Going to remove this for now.

torchtune/utils/ppo_utils/__init__.py
SalmanMohammadi: makes sense, thank you

recipes/lora_ppo_single_device.py
```diff
@@ -0,0 +1,813 @@
...
+                            approx_policy_kls.append(
+                                (0.5 * (pi_logprobs - backward_logprobs).pow(2)).mean()
+                            )
+                            losses.append(loss.detach().item())
+                            policy_losses.append(policy_loss.item())
+                            value_losses.append(value_loss.item())
+
+                            loss.backward()
+                            self._optimizer.step()
+                            self._optimizer.zero_grad(set_to_none=True)
```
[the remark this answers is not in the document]
SalmanMohammadi: yep! will be updated in the next commit : )

recipes/lora_ppo_single_device.py
```diff
@@ -0,0 +1,813 @@
...
+            if utils.VALUE_HEAD_KEY not in model_checkpoint_dict:
+                raise ValueError(
+                    "Value head weights not found. Please ensure a valid value head checkpoint is provided."
+                )
+            self._update_recipe_state(model_checkpoint_dict)
+
+        # ``_setup_model`` handles initialization and loading the state dict. This method
+        # should be called before ``_setup_optimizer`` since transforming the optimizer
+        # state dict requires the model
+        self._model, self._reward_model = self._setup_model(
```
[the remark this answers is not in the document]
SalmanMohammadi: I like this change. One snag I've had with it, is that if I want, for example, to be consistent and specify the model checkpointer as `policy_checkpointer`, the reward checkpointer as `reward_checkpointer`, and so on, then this breaks when using [wandb logging](https://github.com/pytorch/torchtune/blob/b317c8fba35af95cb3f82d190115f4c5e9c533e4/torchtune/utils/metric_logging.py#L225) as the wandb logger assumes `cfg.checkpoint` exists.

ebsmothers (commented):
Taking a second pass but got pulled away for a bit, will be coming back with more comments later today

SalmanMohammadi:
> Taking a second pass but got pulled away for a bit, will be coming back with more comments later today

Thanks for looking! Was planning a larger update sooner but I've been dealing with some absolutely unhinged behaviour when testing recipe resuming - model training is deterministic only for each instance of the recipe being run, and doesn't reproduce exact outputs when resuming. For the sake of my sanity I'll hopefully fix this soon.

Relevant info for you - I've scaled back this PR and I'm just contributing a non-LoRa recipe atm (here's a rough idea of what I've done https://github.com/SalmanMohammadi/torchtune/pull/2). I've significantly tidied the recipe up and things are almost as polished as I'd like them to be. If it's okay I'll re-request a review when I push these changes?

ebsmothers:
@SalmanMohammadi ah thanks for the info, in that case I'm glad I didn't make it all the way through 😅. I'll hold off on further review until you push the newest changes then.

> model training is deterministic only for each instance of the recipe being run, and doesn't reproduce exact outputs when resuming

What do you mean by "instance" here? Just a single continuous run? Is there an additional recipe state you may need to save at intermediate checkpoints (e.g. the KL controller state or something like that?)

torchtune/models/llama2/__init__.py
```diff
@@ -4,7 +4,7 @@
 # This source code is licensed under the BSD-style license found in the
```
SalmanMohammadi, line 4: All changes in models/ will be discarded following merge of https://github.com/pytorch/torchtune/pull/1160

SalmanMohammadi:
### Relatively Large Update

**Stepping back from LoRA**

This PR now only includes a recipe for full-finetuning. I'm going to revisit LoRA another time - my aim with this PR is to have a canonical, accurate, and highly performant RLHF PPO implementation. This work now follows the implementation in the original _Learning to summarize from human feedback, Steinnon et al., 2020_ paper, and replicates the results from the PPOV2 training in TRL (which, in turn, replicates the results from _Learning to summarize_).

I've also stepped away from a joint language-head-and-value-head model. The recipe now uses separate value and policy models (resulting in 4 models being loaded in overall, two of which are being optimised). I'll need to do a bit more testing but I might be able to fine-tune something like Gemma2B using a low-memory config in < 24GB to strengthen the value proposition for training on consumer GPUs - don't hold me to this.

Based on this, there are no dependencies on https://github.com/pytorch/torchtune/issues/1017 for this PR. The current dependencies are https://github.com/pytorch/torchtune/pull/1160, and uploading my reward model checkpoint to S3 so the tests run (@ebsmothers will update on this next week or so).

**Scaling to 7B**

There was some discussion on Discord on potentially running a 7B recipe on your end. I ran a successful experiment [here](https://wandb.ai/salman-mohammadi/torchtune_ppo/runs/rkqybabd/workspace?nw=nwusersalmanmohammadi) using Mistral 7B:
image
Training seemed stable and from an optimization perspective, things seem to be working. I ran the experiment on an A100 - the default config in the repo includes the memory optimization parameters needed to make this work. I used `optimizer_in_bwd` and `PagedAdamW`. Training was pretty slow at the start, but I was seeing >10x speedups once `torch.compile` kicked in. There's a small caveat here in that I'm using instruct policy and reward models for a pretty simple text completion task.

image

Performance and GPU utilisation is pretty good overall - there was a bit of tweaking to avoid bottlenecks, and I've left a couple notes in the recipe doc to highlight this.

I've done my best to clean the recipe up and refactor based on the comments. I've also included tests for pretty much every component I've added. There's lightweight docs in the config and recipes - I'd like to include some more detailled tutorials in the future.

recipes/lora_ppo_single_device.py
```diff
@@ -0,0 +1,813 @@
...
+            if torch.cuda.is_available():
+                if not torch.cuda.is_bf16_supported():
+                    raise RuntimeError(
+                        "Full bf16 training is not supported on this hardware."
+                    )
+            elif torch.backends.mps.is_available():
+                if packaging.version.parse(torch.__version__).release < (2, 3):
+                    raise RuntimeError(
+                        "Full bf16 training is not supported on this hardware."
+                    )
```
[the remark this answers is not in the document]
SalmanMohammadi: I've now removed mps support.

torchtune/utils/ppo_utils/_generation.py
SalmanMohammadi: Generation utils have now been updated to return logits too. This saves an additional forward pass to calculate logprobs over generated sequences.

torchtune/utils/ppo_utils/kl_controller.py
SalmanMohammadi: This has been dropped in more recent implementations (e.g. https://arxiv.org/abs/2204.05862), so I omitted it.

torchtune/utils/ppo_utils/_generation.py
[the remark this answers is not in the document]
SalmanMohammadi: Update: this function now also returns logits and accepts a `torch.Generator`.

recipes/configs/llama2/1B_full_ppo.yaml
```diff
@@ -0,0 +1,182 @@
+# Config for single device RLHF full finetuning using PPO in ppo_full_finetune_single_device.py
```
SalmanMohammadi, line 1: This is the only 1B recipe in the repo, but I think this recipe in particular benefits from starting out with smaller models. The reward model I've linked here has been trained using TRL and the HF page includes the relevant commands to reproduce the model, so hopefully users should be able to adapt the pipeline to their own tasks.

recipes/configs/mistral/7B_lora_ppo.yaml
```diff
@@ -0,0 +1,151 @@
...
+  value_clip_range: 0.1
+
+# Training env
+device: mps
+
+# Memory management
+enable_activation_checkpointing: True
+
+# Reduced precision
+dtype: bf16
```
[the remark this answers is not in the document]
SalmanMohammadi: I think so. I'm not sure I have the expertise to answer this based on experience, but intuitively I'd say that there's several factors in PPO optimisation which try to help stabilise training and reduce variance in gradient updates, so the impact of reduced precision may be smaller than in SFT scenarios. 

The PPO loss also isn't a "distance" like in SFT, so you may not have the same loss landscapes because gradient updates point in the direction of maximising the reward. This means the smoothness of the loss landscape is more related related to the amount of variance in your trajectories e.g. if your generations are significantly different to eachother (due to e.g. generation args), your reward model isn't well-calibrated, or something as simple as your batch size being too small.

Empirically, the reference TRL results I compared against below used fp32, and my results were in bf16.

recipes/configs/llama2/1B_full_ppo.yaml
```diff
@@ -0,0 +1,182 @@
...
+  max_seq_len: null
+  split: train
+  column: prompt
+  add_eos: False
+
+
+# manually constructing 1B models
+policy_model:
+  _component_: torchtune.models.llama2.llama2
+  vocab_size: 32000
```
joecummings, line 41: Is it possible to just use the builder function for this? Or is there a reason why we need to build from here?

recipes/configs/llama2/1B_full_ppo.yaml
```diff
@@ -0,0 +1,182 @@
...
+  num_heads: 32
+  num_kv_heads: 4
+  embed_dim: 2048
+  max_seq_len: 2048
+  intermediate_dim: 5632
+  attn_dropout: 0.0
+  norm_eps: 1e-5
+
+reward_and_value_model:
+  _component_: torchtune.models.llama2.llama2_classifier
```
joecummings, line 52: Same comment as above

recipes/configs/llama2/1B_full_ppo.yaml
```diff
@@ -0,0 +1,182 @@
...
+  lr: 3e-6
+
+optimizer_in_bwd: False
+log_peak_memory_stats: False
+enable_activation_checkpointing: True
+
+# Reduced precision
+dtype: bf16
+
+# Trajectory generation arguments
```
joecummings, line 136: What's this?

recipes/configs/llama2/1B_full_ppo.yaml
```diff
@@ -0,0 +1,182 @@
...
+metric_logger:
+  _component_: torchtune.utils.metric_logging.DiskLogger
+  log_dir: ${output_dir}
+
+log_every_n_steps: 1
+
+resume_from_checkpoint: False
+output_dir: /tmp/llama2-1b-ppo-finetune
+seed: null
+shuffle: True
```
joecummings, line 182: In general, I think we want to keep configs as simple as possible and leave any complexity not necessarily needed to code or maybe just by showing an example in a tutorial. 182 lines for a config feels uncomfortable. Are there any other simplifications we can make here? Orr is there anything torchtune can do to help make the config simpler for a use case like this?

cc @kartikayk  and @ebsmothers

torchtune/modules/rlhf/collate.py
```diff
@@ -0,0 +1,100 @@
...
+
+from typing import Dict, List, Tuple
+
+import torch
+from torch.nn.utils.rnn import pad_sequence
+
+from torchtune.data import CROSS_ENTROPY_IGNORE_IDX
+
+
+def left_padded_collate(
```
joecummings, line 15: I think this can go in generate `collate.py` - it should be general enough :)

torchtune/modules/rlhf/_generation.py
```diff
@@ -0,0 +1,172 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
```
joecummings, line 1: This looks as though it could be not in rlhf, but rather in the general `_generation.py`

joecummings (commented):
This is shaping up nicely - I very much like that we de-scoped this to focus on full finetune first with lora and qlora as follow-ups. A couple high level things to note:

1. I'm concerned about the configs becoming too bloated and would like to discuss how to minimize storing lost of logic there.

2. What are the largest size model you can fit in 80G A100? I see you include configs for both 7B and 1B?

recipes/configs/llama2/1B_full_ppo.yaml
SalmanMohammadi: Sorry, could be clear, should be a high-level header comment for the next block of args

recipes/configs/llama2/1B_full_ppo.yaml
SalmanMohammadi: Sorry, I'm missing something here - do we have a builder function for llama 1B? I opted for a longer yaml file vs. adding new model builders here.

SalmanMohammadi:
Thanks for another review.


>  I'm concerned about the configs becoming too bloated and would like to discuss how to minimize storing lots of logic there.

I feel this. One thing that stuck out to me when writing this - we currently need 4 checkpointers, two of which are solely used to point to the original weights for the policy and reward models, respectively. They're necessary because you need the reference to the original weights when resuming training, and the choice to me at the time was managing this state in the config vs the checkpoints.

The model definitions are also taking up a lot of space, but that's largely because I didn't see another obvious way to configure a 1B Llama2. The model definition in the Mistral config is annoying because that specific reward model uses a different vocab size. Please let me know if I can make this cleaner!

There's also ~30 lines for hyperparameters in the config. Hopefully this won't be overwhelming to the user once we include a cookbook. I could remove 5 or so of these from the config and set as defaults in the recipe.

>  What are the largest size model you can fit in 80G A100? I see you include configs for both 7B and 1B?

The training run in my [above comment](https://github.com/pytorch/torchtune/pull/1005#issuecomment-2230700024) trained Mistral 7B on an 80GB A100.

torchtune/modules/rlhf/_generation.py
SalmanMohammadi: I could merge the functionality into the existing generation code with sensible defaults, and a `return_logits` flag? 

Though, not sure how you guys feel about return types being `Union[Tuple[..], Tensor]`, but it always feels weird to me to have function calls needing to be unpacked in some places, but not others.

torchtune/modules/rlhf/collate.py
SalmanMohammadi: Would you like to see it match the signature of `padded_collate`, too? There's also a more convoluted suggestion here https://github.com/pytorch/torchtune/pull/1005#discussion_r1648391833

recipes/lora_ppo_single_device.py
SalmanMohammadi: The `context_length - 1 : -1` doesn't actually have anything to do with padding/eos tokens, it's because we're always predicting the next token, so the logprobs are a step behind. Consider that at each step in the response, the action taken is `a_t = response_t`, and the subsequent state produced by the "environment",  is `query_0, ... query_n, response_0, ...response_t` (i.e. the environment dynamics are just whichever sampling strategy you choose).

The `context_length - 1 : -1` happens because we're estimating logprobs of our policy for every action `pi_theta(a_t, s_t)`. So at the start of the trajectory, the logprobs for the first response token, `pi_theta(response_0 | query_0, ..., query_n)` are `F.log_softmax(self._policy_model([query_0, ..., query_n])`, hence why we do `context_length - 1`. Similarly, we slice up to `:-1` because the logprobs for the last action are at position -1, and there are no more actions taken (tokens generated) to estimate logprobs for.

SalmanMohammadi:
bump bump @joecummings @ebsmothers.

My dearest reviewers,

Sorry to ping you when you're busy. In my defense, @kartikayk did tell me to. What can I do to help move this along? I'm more than happy to help reduce the review overhead if I can.

recipes/configs/llama2/1B_full_ppo.yaml
SalmanMohammadi: Removed 1B yaml. Going to add later in a deepdive/recipe cookbook.

SalmanMohammadi:
Outstanding discussions/tasks:

@joecummings @ebsmothers 

- [ ] @ebsmothers can you please upload my reward model to S3 so my recipe tests pass? See the PPO channel on Discord. Let me know if you need a refresh on this.
- [ ] @joecummings pointed out that the config is a bit unweildy. I think I could address this by removing a bunch of parameters from the config and setting sensible defaults, and then correctly documenting them in a recipe doc (which might look like #1230). Can we leave this as a follow up, though? 🥺
- [ ] Are you happy with my replication results here? https://github.com/pytorch/torchtune/pull/1005/#issuecomment-2230700024
- [ ] Can we also leave generalising the generation utils to a follow up? Joe has started discussion in #1250.
- [ ] My use of rng generator checkpointing (https://github.com/pytorch/torchtune/pull/1005/#issuecomment-2225247139) is unprecedented in the codebase. Are you happy with this? I'm just [adding another key](https://github.com/pytorch/torchtune/blob/fe87dfbb662b9ed74b727ad1f81ee35eb4a775b7/torchtune/utils/constants.py#L28) to the checkpoint.
- [ ] Do you care about generalizing the RLHF collation utils to the `torchtune.utils.collate`? (Joe's comment [here](https://github.com/pytorch/torchtune/pull/1005/#discussion_r1648391833) and [here](https://github.com/pytorch/torchtune/pull/1005/#discussion_r1680178159) They currently aren't being used outside of the PPO recipe itself, and the DPO collation isn't being used outside the DPO recipe.

tests/test_utils.py
```diff
@@ -29,6 +29,7 @@
     "llama2_tune": "/tmp/test-artifacts/small-ckpt-tune-03082024.pt",
     "llama2_meta": "/tmp/test-artifacts/small-ckpt-meta-03082024.pt",
     "llama2_hf": "/tmp/test-artifacts/small-ckpt-hf-03082024.pt",
+    "llama2_reward_hf": "/tmp/test-artifacts/small-ckpt-hf-reward-12072024.pt",  # TODO (SalmanMohammadi)
```
ebsmothers, line 32: Sorry for being an American chauvinist but I changed the filename to `small-ckpt-hf-reward-07122024.pt` (really just want to make it consistent with the format of the other ones). Also I think you will need to update [cache_artifacts.sh](https://github.com/pytorch/torchtune/blob/78055fa7d1faa885df98d039c4d8059e3477cfc8/tests/cache_artifacts.sh#L15) correspondingly

recipes/ppo_full_finetune_single_device.py
```diff
@@ -0,0 +1,1139 @@
...
+                    utils.TOTAL_EPOCHS_KEY: self._total_epochs,
+                    utils.MAX_STEPS_KEY: self._total_steps,
+                    utils.STEPS_KEY: self._steps_run,
+                    utils.RNG_KEY: self._rng.get_state(),
+                }
+            )
+            if not self._optimizer_in_bwd:
+                policy_ckpt_dict[utils.OPT_KEY] = self._optimizer.state_dict()
+            else:
+                policy_ckpt_dict[utils.OPT_KEY] = self._optim_ckpt_wrapper.state_dict()
```
ebsmothers, line 641: Wait this is confusing to me.. so we run the optimizer on both the policy and the value model, right? But then we save the optimizer states for both in the policy model checkpoint only?

recipes/ppo_full_finetune_single_device.py
```diff
@@ -0,0 +1,1139 @@
...
+        except KeyError as e:
+            raise KeyError from e(
+                "Checkpoint does not contain the required keys needed for updating recipe state."
+                "Are you sure you passed in the right recipe checkpoint?"
+            )
+
+    def generate_trajectory(self, input_ids: torch.Tensor) -> Trajectory:
+        """
+        Generates a trajectory given the current policy and value models, the reference policy model, the reward model,
+        and batch of inputs.
```
ebsmothers, line 687: The enumeration of steps 1-6 in the code comments is useful, but might help to consolidate a more detailed description in the docstring of what this method is doing too

recipes/ppo_full_finetune_single_device.py
```diff
@@ -0,0 +1,1139 @@
...
+        )
+        ref_logprobs = rlhf.logits_to_logprobs(ref_logits, responses, self._temperature)
+
+        del ref_logits
+
+        # step 3. estimate values from the generated responses using the value function
+        values = self._value_model(query_responses, input_pos=position_ids, mask=masks)
+        values = rlhf.query_response_logits_to_response_logits(
+            values, context_length
+        ).squeeze(-1)
```
ebsmothers, line 759: Maybe this util is not the most aptly-named if we're using it here?

recipes/ppo_full_finetune_single_device.py
```diff
@@ -0,0 +1,1139 @@
...
+
+        # step 4. replace any tokens in the response after the first stop token (usually EOS token) with padding
+        response_padding_masks, responses = rlhf.truncate_sequence_at_first_stop_token(
+            responses, self._stop_token_ids, self._tokenizer.pad_id
+        )
+
+        # step 5. run the reward model on the query truncated-response pairs
```
ebsmothers, line 766: nit: all the places you say stuff like "query truncated-response pair" or "query-generated-response pair" I would say it like "(query, truncated response) pair" or "(query, generated response) pair" to make it explicit which words map to distinct entities

recipes/ppo_full_finetune_single_device.py
```diff
@@ -0,0 +1,1139 @@
...
+
+        # step 6. mask out all the invalid values in the trajectory due to padding tokens
+        logprobs[response_padding_masks] = 1.0
+        ref_logprobs[response_padding_masks] = 1.0
+
+        # step 6.1 values are masked out *after* the last valid token in the response
+        value_seq_idxs = torch.where(
+            (seq_lens > 0) & (seq_lens < self._max_generated_tokens - 1),
+            seq_lens + 1,
+            seq_lens,
```
ebsmothers, line 797: Sorry not sure I fully follow what the purpose of this is

ebsmothers:
General comment on the checklist you left earlier: all the points look good to me, let's just file tasks for some of the more important todos that don't have them already.

Also, leaving some miscellaneous remarks here in response to several of your previous comments:

> I ran the experiment on an A100 - the default config in the repo includes the memory optimization parameters needed to make this work. I used optimizer_in_bwd and PagedAdamW. Training was pretty slow at the start, but I was seeing >10x speedups once torch.compile kicked in. 

Looking at the figures seems this is necessary even for A100? Since you are still pretty close to 80GB allocated memory. I'm also curious whether the overall training speed is decent as these configs can slow things down quite a bit.

> Can we also leave generalising the generation utils to a follow up? Joe has started discussion in https://github.com/pytorch/torchtune/issues/1250.

Just want to confirm: will we actually be able to run this recipe without batched generation support?

recipes/ppo_full_finetune_single_device.py
```diff
@@ -0,0 +1,1139 @@
...
+        returns: torch.Tensor,
+        context_length: int,
+    ) -> Tuple[
+        torch.Tensor,
+        torch.Tensor,
+        torch.Tensor,
+        torch.Tensor,
+        torch.Tensor,
+        torch.Tensor,
+    ]:
```
ebsmothers, line 985: Similar comment here to elsewhere.. maybe time to introduce a dataclass (or namedtuple as you did with Trajectory). Also this'd be nice cause you can use the dataclass as an opportunity to clearly document the meaning of each of the fields in more detail

recipes/ppo_full_finetune_single_device.py
```diff
@@ -0,0 +1,1139 @@
...
+
+                # step 3. estimate the advantages using Generalized Advantage Estimation (GAE)
+                advantages, returns = rlhf.estimate_advantages(
+                    trajectory.values,
+                    rewards,
+                    self._gamma,
+                    self._lmbda,
+                    masks=~trajectory.response_padding_masks,
+                )
+                advantages[trajectory.response_padding_masks] = 0.0
```
ebsmothers, line 889: Any reason not to mask inside `estimate_advantages` if you're already passing the mask in there anyways?

recipes/ppo_full_finetune_single_device.py
```diff
@@ -0,0 +1,1139 @@
...
+
+                # step 4. optimise using the PPO objective over multiple epochs
+                ppo_stats = [
+                    [],
+                ] * 6
```
ebsmothers, line 894: nit: add a comment for the magic number (if you wind up using a dataclass could also just do like `len(MyDataclass.fields())` or something)

ebsmothers (commented):
OK a bunch more comments but after that there are no major concerns. Home stretch here -- thanks again for your immense patience on this one

torchtune/modules/loss/ppo.py
```diff
@@ -0,0 +1,120 @@
...
+            value_padding_masks (Optional[torch.Tensor]): Padding token masks of the same shape as ``pi_logprobs``,
+                where True indicates the corresponding loss values should participage in value loss calculation.
+
+        Returns:
+            Tuple[torch.Tensor, torch.Tensor, torch.Tensor, torch.Tensor, torch.Tensor]: A tuple containing:
+                - loss: The total PPO loss.
+                - policy_loss: The policy function loss.
+                - value_loss: The value function loss.
+                - ratios: The ratio between the current and old policy probabilities.
+                - clipfrac: The fraction of ratios that were clipped.
```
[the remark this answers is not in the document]
SalmanMohammadi: I hate to agree. You don't know how many different things I tried before I added the Trajectory NamedTuple. I don't know why they make me feel bad.

recipes/ppo_full_finetune_single_device.py
SalmanMohammadi: ...
...
...
Dare I say..... excalidraw?

tests/test_utils.py
SalmanMohammadi: Oh how the Empire has fallen from grace.

torchtune/modules/rlhf/rewards.py
```diff
@@ -0,0 +1,251 @@
...
+    # estimate advantage for every predicted token position
+    for t in reversed(range(response_length)):
+        # value of the next state
+        next_values = values[:, t + 1] if t < response_length - 1 else 0.0
+        # exponentially discounted temporal difference error:
+        # delta_t = r_t + gamma * V(s_{t+1}) - V(s_t)
+        delta = rewards[:, t] + gamma * next_values - values[:, t]
+        # GAE-Lambda advantage discouting saved for the next iteration
+        # as A_t = delta_t + gamma * lambda * A_{t+1} + ...
+        last_gae_lam = delta + gamma * lmbda * last_gae_lam
```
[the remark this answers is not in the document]
SalmanMohammadi: the latter is correct : )

torchtune/modules/loss/ppo.py
```diff
@@ -0,0 +1,120 @@
...
+
+
+        """
+        ratios = torch.exp(pi_logprobs - pi_old_logprobs)
+        clipped_ratios = torch.clamp(ratios, 1.0 - self.epsilon, 1.0 + self.epsilon)
+
+        policy_losses_clipped = -advantages * clipped_ratios
+        policy_losses_unclipped = -advantages * ratios
+
+        clipfrac = (policy_losses_clipped > policy_losses_unclipped).float()
```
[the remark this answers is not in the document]
SalmanMohammadi: since we take `torch.maximum(policy_losses_clipped, policy_losses_unclipped)` this checks out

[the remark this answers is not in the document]
SalmanMohammadi: right?

recipes/ppo_full_finetune_single_device.py
SalmanMohammadi: ...yes. This was a pretty hacky but simple solution that worked for me, and meant we only managed recipe state using one checkpointer

recipes/ppo_full_finetune_single_device.py
SalmanMohammadi: In all seriousness, I can send you a diagram from my notes on Discord. This took me a while to wrap my head around, and longer to explain coherently, so, thanks for the nerd snipe.

The TL;DR - the value function is estimating the return for the whole sequence at each step, which is the reward model score for the (query, truncated response), plus the KL per-token penalty. We want to use this for the advantage estimation, and the advantage for the last action taken (the last valid non-padding token generated by the model), is:

image

So, we need the value estimate (return) for the sequence up to now, plus one step ahead. For the last token, this means we need extend the padding mask out by one for the values - instead of masking everything after the last non-padding token, we mask everything one value after the last non-padding token.

These three lines do this, but just add some logic to say if we're already at the end of the sequence then we don't need to extend the mask.

torchtune/modules/rlhf/sequence_processing.py
```diff
@@ -0,0 +1,69 @@
...
+    ).squeeze(-1)
+
+
+def query_response_logits_to_response_logits(
+    query_response_logits: torch.Tensor, context_length: int
+) -> torch.Tensor:
+    """
+    Converts logits estimated over a query-generated-response pair logits to logits for the response.
+
+    See the excalidraw linked in TRL's PPOV2 for a visual explanation
```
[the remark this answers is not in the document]
SalmanMohammadi: What did we send you to that fancy school for, Dr. Smothers?

torchtune/modules/loss/ppo.py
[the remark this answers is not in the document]
SalmanMohammadi: I left this as-is and used NamedTuples in the recipe.

recipes/ppo_full_finetune_single_device.py
```diff
@@ -0,0 +1,1139 @@
...
+        - Optimizer in Backward. Fusing the optimizer step into the backward pass helps reduce the memory
+            footprint associated with gradients. This can be especially helpful when you are memory
+            constrained. Note that users can only use ONE of gradient accumulation or optimizer in backward.
+            These features currently do not work together. For more details on optimizer in backward, please
+            see this tutorial: https://pytorch.org/tutorials/intermediate/optimizer_step_in_backward_tutorial.html
+
+            This paramater can provide significant performance gains, since there the number of optimization steps
+            scales with ``ppo_epochs`` and ``batch_size``. Depending on the maximum sequence length sampled from the dataset,
+            we've found that setting ``ppo_batch_size`` to the highest you can fit in memory, and `optimizer_in_bwd=True` to
+            provide significant speedups and memory savings.
```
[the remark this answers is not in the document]
SalmanMohammadi: Probably more testing needed to make the stronger claim.

SalmanMohammadi:
O' what glorious reviews. 

> Looking at the figures seems this is necessary even for A100? Since you are still pretty close to 80GB allocated memory. I'm also curious whether the overall training speed is decent as these configs can slow things down quite a bit.

For 7B, since we're fitting 4x7B models in it took a little wrangling to fit it all. The run I posted took around ~3 hours. I haven't found any tests to benchmark against here on comparable hardware, to estimate appropriate speed/memory usage. DeepSpeed's [RLHF](https://github.com/microsoft/DeepSpeedExamples/blob/master/applications/DeepSpeed-Chat/training/step3_rlhf_finetuning/README.md#%F0%9F%99%8B-instablity-of-rlhf-training-and-others) states:

> Theoretically, the largest model you can train for this step is similar to the step-1 SFT finetuning if you enable

>    zero stage 3 (if you use multiple GPUs)
    gradient checkpoint
    LoRA
    reference model offloading.

> However, in practice, this is not always the case, and we are still investigating the reasons behind it. For now, we suggest that users use "Total-GPU-Memory-in-GB / 6" as the upper parameter bound in billions for the sum of the actor model and critical model, for safety. Nevertheless, users are welcome to try the real limit. 

Not 100% clear on whether the upper parameter bound calculation requires the specific config they listed, but going by that, their method offers ~13GB for the sum of the actor and critic models, max, on an 80GB A100. TRL [trained Pythia 6.9B](https://wandb.ai/costa-huang/tldr_summarize/runs/vou23bja?nw=nwusercostahuang) on their PPOV2 trainer with 8xH100. 


> Just want to confirm: will we actually be able to run this recipe without batched generation support?

The generation utils I include in this PR do provide batched generation support.

torchtune/modules/loss/ppo.py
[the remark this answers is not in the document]
ebsmothers: Stolen from a trusted resource on the topic 😃 

> If the advantage is positive, then the new policy becomes greedier relative to the old policy, and we have r(theta) > 1 - the inverse applies for negative advantage, where we have r(theta) < 1.

I was missing this detail, now it makes sense

recipes/ppo_full_finetune_single_device.py
ebsmothers: OK, I think this is fine for this recipe, but does raise an interesting point about the best way to train multiple models in a single recipe. I think right now our checkpointer is not really designed for this kinda thing -- I know you hacked around here, but wondering if some kinda MultiModelCheckpointer abstraction wrapping our existing checkpointers could make sense for something like this (then each model and optimizer state can be wrapped together with a global recipe state saved separately and not tied to an individual model).

recipes/ppo_full_finetune_single_device.py
ebsmothers: Honestly it's really just something like `truncate_first_n_tokens`, right? Idk though, naming is hard

torchtune/modules/rlhf/sequence_processing.py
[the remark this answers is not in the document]
ebsmothers: Kinda related to my comment above about renaming this. While (some may say) this is a helpful diagram, idk that it actually helps clarify a function that's really just dropping the query tokens. When I click into that my reaction is "whoa, all that's happening in this single line?"

recipes/ppo_full_finetune_single_device.py
SalmanMohammadi: Hmm yeah. Maybe `truncate_sequence_for_logprobs`?

recipes/ppo_full_finetune_single_device.py
SalmanMohammadi: Can you think of any other paradigms we could use this for?

ebsmothers (approved):
Thank you for your immense patience on this one. I left a couple of other follow-up comments, but none of them are blocking us from landing this. 🚀

Issue #1001, opened 2024-05-17
Sanity check: How to run inference after fine-tuning Llama-3 on chat data?


julianstastny:
Hi! I'm hopeful that I'm doing this correctly, but not completely sure, and I don't know how best to verify this -- I've fine-tuned Llama-3 8B (using qlora) following this tutorial: https://pytorch.org/torchtune/main/tutorials/chat.html

Now I wanted to run inference (actually I'd love it if there was a simple way for me to run inference on a validation dataset that is formatted like my training dataset, but in absence of that), by adapting the generation template: 

```yaml
# Model arguments
model:
  _component_: torchtune.models.llama3.llama3_8b

checkpointer:
  _component_: torchtune.utils.FullModelMetaCheckpointer
  checkpoint_dir: <directory path>/Meta-Llama-3-8B-Instruct/
  checkpoint_files: [
    meta_model_0.pt
  ]
  output_dir: <directory path>/Meta-Llama-3-8B-Instruct/
  model_type: LLAMA3

device: cuda
dtype: bf16

seed: 1234

# Tokenizer arguments
tokenizer:
  _component_: torchtune.models.llama3.llama3_tokenizer
  path: <directory path>/Meta-Llama-3-8B-Instruct/original/tokenizer.model

# Generation arguments; defaults taken from gpt-fast
prompt: '<|start_header_id|>system<|end_header_id|>\n\nYou are a helpful and harmless assistant.<|eot_id|><|start_header_id|>user<|end_header_id|>\n\nGive three tips for staying healthy.<|eot_id|><|start_header_id|>assistant<|end_header_id|>\n\n'
max_new_tokens: 300
temperature: 0.6 # 0.8 and 0.6 are popular values to try
top_k: 300

quantizer: null
```

Is it correct that I specify the prompt in that format, or am I making a mistake because the tokenizer takes care of the special tokens somehow? Thanks!

RdoubleA:
Hi @julianstastny, good question. The tokenizer should handle the template for you. You just need to specify the raw string in your config. You might get unexpected behavior by putting in the special tokens manually.

cc @joecummings on running inference on a validation dataset, I think that would be a good use case to support

musabgultekin:
Generate recipe doesn't handle chat format automatically. But its adding the BOS token when encoding: See: https://github.com/pytorch/torchtune/blob/eb66b627221a770f32f8517a91178bf1ccac4a3c/recipes/generate.py#L86

[...]

PR #1019, opened 2024-05-24
Generalize prompting in generate recipe

ebsmothers:
### Context

Based on #1001, it's clear that our generation recipe is not the most flexible when it comes to different tokenizers/formats/templates we support.

Right now this stuff is inherently a bit messy because there are two tokenizer APIs that could be called: `tokenizer.encode` or `tokenizer.tokenize_messages`. I claim the latter is the "better" API as it is more general. But there are still cases where someone just wants to pass a raw string with no special formatting, and have it get tokenized directly. At least as of right now our tiktoken tokenizer does not support that via `tokenize_messages`.

Pending a larger-scale redesign (which I do hope to see sooner rather than later), this is one way we could bridge the gap. 

### Changes

- Generalize `prompt` field in the generation config to support either a string (current version) or a dictionary mapping roles to message bodies.
- Add optional `chat_format` and `instruct_template` fields to the config for further customization of the prompt.
- For raw string, assume any formatting is passed via instruct_template, and directly use `tokenizer.encode`. 
- For dictionary, format into `Message` dataclass and call `tokenizer.tokenize_messages`

On the YAML side, the two options would be:

#### Existing flow
```
prompt: "Four score and seven years ago"
instruct_template: MyInstructTemplate (can also be null)
chat_format: null
```

#### New flow
```
prompt:
  system: "You are a kind and helpful assistant. Respond to the following request."
  user: "Write a generation recipe for torchtune."
instruct_template: null
chat_format: MyChatFormat (can also be null)
```

Note: I acknowledge that the utility as I've added it is a bit messy. The aim here is to demonstrate that this generally supports the cases that we want to support, and can rewrite to be a bit cleaner if folks are on board with the approach.

### Test plan

Four test cases:
- Raw string passed to Llama3-8B instruct model
- 

#### Pass raw string
```
tune run generate --config generation model=torchtune.models.llama3.llama3_8b \
checkpointer=torchtune.utils.FullModelMetaCheckpointer \
checkpointer.checkpoint_dir=/tmp/Meta-Llama-3-8B-Instruct/original/ \
checkpointer.checkpoint_files=['consolidated.00.pth'] checkpointer.model_type=LLAMA3 \
tokenizer=torchtune.models.llama3.llama3_tokenizer \
prompt='Give three tips for staying healthy.'
...
Give three tips for staying healthy. - Essay Example
The importance of staying healthy cannot be overstated. A healthy body and mind are essential for a happy and fulfilling life. Here are three tips for staying healthy: 1. Eat a Balanced Diet: A healthy diet is the foundation of good health. It is essential to eat a variety of foods, including fruits, vegetables, whole grains, lean proteins, and healthy fats...
Extract of sample "Give three tips for staying healthy"
Download file to see full Extract
The importance of staying healthy cannot be overstated. A healthy body and mind are essential for a happy and fulfilling life. Here are three tips for staying healthy: 1. Eat a Balanced Diet: A healthy diet is the foundation of good health. It is essential to eat a variety of foods, including fruits, vegetables, whole grains, lean proteins, and healthy fats. A balanced diet provides the necessary nutrients, vitamins, and minerals for the body to function properly. It is also important to limit the intake of processed and sugary foods, which can be detrimental to health.

2. Exercise Regularly: Exercise is another crucial aspect of staying healthy. It helps to maintain a healthy weight, boost energy levels, and improve overall physical and mental well-being. Regular exercise can also reduce the risk of chronic diseases, such as heart disease, diabetes, and some types of cancer. Aim for at least 30 minutes of moderate-intensity exercise per day, such as brisk walking, cycling, or swimming.

3.
```

#### Pass dict without any template or format
```
tune run generate --config generation model=torchtune.models.llama3.llama3_8b \
checkpointer=torchtune.utils.FullModelMetaCheckpointer \
checkpointer.checkpoint_dir=/tmp/Meta-Llama-3-8B-Instruct/original/ \
checkpointer.checkpoint_files=['consolidated.00.pth'] checkpointer.model_type=LLAMA3 \
tokenizer=torchtune.models.llama3.llama3_tokenizer \
prompt.system='You are a helpful and harmless assistant.' prompt.user='Give three tips for staying healthy.'
...
 <|start_header_id|>system<|end_header_id|>

You are a helpful and harmless assistant.<|eot_id|><|start_header_id|>user<|end_header_id|>

Give three tips for staying healthy.<|eot_id|><|start_header_id|>assistant<|end_header_id|>

<|eot_id|><|start_header_id|>assistant<|end_header_id|>

Here are three tips for staying healthy:

1. **Stay Hydrated**: Drinking plenty of water throughout the day is essential for flushing out toxins, regulating body temperature, and maintaining energy levels. Aim to drink at least eight glasses of water a day, and adjust according to your activity level and climate.

2. **Eat a Balanced Diet**: Focus on consuming a variety of whole, unprocessed foods like fruits, vegetables, whole grains, lean proteins, and healthy fats. Aim to include a rainbow of colors on your plate to ensure you're getting a range of essential vitamins and minerals.

3. **Get Moving Regularly**: Regular physical activity can help boost your mood, energy levels, and overall health. Aim for at least 30 minutes of moderate-intensity exercise, such as brisk walking, cycling, or swimming, most days of the week. You can also incorporate strength training, high-intensity interval training (HIIT), and flexibility exercises to keep things interesting and challenging.

Remember to always consult with a healthcare professional before starting any new exercise or diet program, especially if you have any underlying health conditions or concerns!<|eot_id|>

```

#### Pass dict with instruct template

Using @joecummings's [fine-tuned model for answering math questions](https://huggingface.co/jrc/phi3-mini-math). First: `tune download jrc/phi3-mini-math --output-dir /tmp/phi3-mini-math --hf-token <HF_TOKEN>`. Then

```
tune run generate --config generation model=torchtune.models.phi3.phi3_mini \
checkpointer=torchtune.utils.FullModelHFCheckpointer checkpointer.checkpoint_dir=/tmp/phi3-mini-math/ \
checkpointer.checkpoint_files=['model-0001-of-0002.pt','model-0002-of-0002.pt'] \
checkpointer.model_type=PHI3_MINI \
tokenizer=torchtune.models.phi3.phi3_mini_tokenizer tokenizer.path=/tmp/phi3-mini-math/tokenizer.model \
prompt.instruction='Prove that the square root of 2 is irrational.' \
instruct_template=torchtune.data.AlpacaInstructTemplate
...

Below is an instruction that describes a task. Write a response that appropriately completes the request.

### Instruction:
Prove that the square root of 2 is irrational.

### Response:

 
 
To prove that the square root of 2 is irrational, we will use a proof by contradiction.

Assume that the square root of 2 is rational. This means that it can be expressed as a fraction a/b, where a and b are integers with no common factors other than 1 (i.e., a and b are coprime).

(√2)^2 = 2
(a/b)^2 = 2
a^2 / b^2 = 2
a^2 = 2b^2

Now, we know that a^2 and 2b^2 are both perfect squares. Therefore, their prime factors must also be perfect squares.

Since 2 is a prime number, its square, 4, is the only prime factor of 2b^2. Therefore, 2b^2 must be a multiple of 4.

a^2 = 2b^2
a^2 / 2 = b^2
a^2 / 2 is a perfect square (since b^2 is a perfect square).

However, this contradicts our initial assumption that a and b are coprime, as a^2 / 2 implies that a^2 has a factor of 2.

Therefore, our initial assumption that the square root of 2 is rational must be

```

#### Pass dict with chat format

First download Mistral-7B-Instruct-v0.1 via `tune download mistralai/Mistral-7B-Instruct-v0.1 --hf-token <HF_TOKEN> /tmp/Mistral-7B-Instruct-v0.1`. Then run

```
tune run generate --config generation model=torchtune.models.mistral.mistral_7b checkpointer=torchtune.utils.FullModelHFCheckpointer checkpointer.checkpoint_dir=/tmp/Mistral-7B-Instruct-v0.1/ checkpointer.model_type=MISTRAL tokenizer=torchtune.models.mistral.mistral_tokenizer tokenizer.path=/tmp/Mistral-7B-Instruct-v0.1/tokenizer.model prompt.user='You are a helpful and harmless assistant. Give three tips for st
aying healthy.' chat_format=torchtune.data.MistralChatFormat
...
[INST] You are a helpful and harmless assistant. Give three tips for staying healthy. [/INST] 1. Eat a Balanced Diet: Make sure you consume a wide variety of nutrients from fresh, whole foods. This includes plenty of fruits and vegetables, lean proteins, whole grains, and healthy fats. A balanced diet provides your body with the nutrients and energy it needs to function properly and helps to maintain a healthy weight.

2. Exercise Regularly: Regular physical activity is essential for good health. Aim for at least 150 minutes of moderate-intensity aerobic exercise or 75 minutes of vigorous-intensity aerobic exercise per week, along with muscle-strengthening activities at least two days a week. Exercise helps to improve cardiovascular health, build and maintain muscle mass, and boost overall physical and mental well-being.

3. Get Enough Sleep: Quality sleep is essential for good health. Aim for 7-9 hours of sleep per night, and try to maintain a consistent sleep schedule. Adequate sleep helps to improve cognitive function, reduce stress and anxiety, and support physical health and growth.
```


torchtune/modules/tokenizers/_tiktoken.py
```diff
@@ -335,6 +337,7 @@ def tokenize_messages(
         messages: List[Message],
         max_seq_len: Optional[int] = None,
         tokenize_header: bool = True,
+        add_eos: bool = True,
```
joecummings, line 340: So this was defaulted to true before?

joecummings:
Can you update the default in the YAML as well?

torchtune/modules/tokenizers/_tiktoken.py
musab-mk: Since generate recipe wasn't using tokenize_messages, it wasn't a problem before. As this was only used for training before

musabgultekin: Since generate recipe wasn't using tokenize_messages, it wasn't a problem before. As this was only used for training before

recipes/generate.py
```diff
@@ -81,9 +81,40 @@ def _setup_model(
...
         return model
 
+    def convert_prompt_to_tokens(
+        self,
+        prompt: Union[DictConfig, str],
+        chat_format: ChatFormat,
+        instruct_template: InstructTemplate,
+    ) -> List[Message]:
+        # Should only be chat-style prompt or instruct-style prompt
+        if chat_format and instruct_template:
```
RdoubleA, line 91: we should make `chat_format` and `instruct_template` optional - if both are required then this is inevitably raised, and it doesn't match the examples in your summary? also having both be none is a possible use case

recipes/generate.py
```diff
@@ -81,9 +81,40 @@ def _setup_model(
...
+            raise ValueError(
+                "Cannot pass both chat format and instruct template for generation"
+            )
+
+        # A single string can be formatted with instruct_template
+        if isinstance(prompt, str):
```
RdoubleA, line 97: this highlights a reason why I want to move towards combining instruct_template / chat_format. You can just treat a single string as a user message, then apply instruct template only on the user message. I don't like the logic here but making this cleaner might require updating InstructTemplate / ChatFormat

recipes/generate.py
```diff
@@ -81,9 +81,40 @@ def _setup_model(
...
+                prompt = template.format(prompt)
+            return self._tokenizer.encode(prompt, add_bos=True, add_eos=False)
+
+        # Otherwise pass a DictConfig mapping role -> message, format with ChatFormat
+        # dict.items() will respect order for Python >= 3.7
+        else:
+            messages = [Message(role=k, content=v) for k, v in prompt.items()]
+            messages += [
+                Message(role="assistant", content="")
+            ]  # TODO: pass via config?
```
RdoubleA, line 108: why would you want to pass this via config?

recipes/generate.py
```diff
@@ -81,9 +81,40 @@ def _setup_model(
...
+            if chat_format:
+                messages = chat_format.format(messages)
```
RdoubleA, line 110: doing this after appending the assistant message is slightly problematic because the chat format may add tags at the end of the assistant message (like ChatML, [for example](https://github.com/pytorch/torchtune/blob/3b1bcf9f084ae7a03fd20251402e0292fa9a9495/torchtune/data/_chat_formats.py#L190)), but ideally we just want the model to output tokens starting with where the assistant message should start. I don't have a good solution to this tbh...

RdoubleA (commented):
This can be a good short-term fix, but frankly this highlights some of my gripes with InstructTemplate / ChatFormat that I think they should be revisited, especially since current models are moving away from requiring prompt templates that aren't special tokens and handling this all in the tokenizer

recipes/generate.py
ebsmothers: What if e.g. you don't want to explicitly add assistant in this format?

recipes/generate.py
ebsmothers: Yeah this is def a short-term fix. I am not getting into instruct template/chat format here cause that is a heavier refactor

recipes/generate.py
ebsmothers: Ah sorry yeah this is what I had in mind anyways. Will update

recipes/generate.py
ebsmothers: Yeah that's kinda annoying. Where are we actually even using `ChatMLFormat` though?

ebsmothers:
> This can be a good short-term fix, but frankly this highlights some of my gripes with InstructTemplate / ChatFormat that I think they should be revisited, especially since current models are moving away from requiring prompt templates that aren't special tokens and handling this all in the tokenizer

100% agree, let's account for this refactor somewhere (maybe open an issue to track it?)

recipes/generate.py
RdoubleA: nowhere really, I added it because at the time it was a widely used prompt template used by OpenAI. not sure if that is still true today.

recipes/generate.py
RdoubleA: but the prompt field in the config implies that it is system + user only message, it may be weird to include an assistant message even if it's just formatting

ebsmothers: I guess my comment is more around the assumption that we always want the tokenizer to tokenize a message with role="assistant" and empty body as the last part of the prompt. What if e.g. the user has a different role name, or wants to add "\n\n", or something like that?

joecummings (approved):

Issue #1014, opened 2024-05-22
Configurable sharding strategy for distributed finetune


tambulkar:
Can we add an option to the config to make the sharding strategy configurable for the distributed fine tune runs?

kartikayk:
Thank you for opening this issue! 

This has been on our list of TODOs since different sharding strategies expose different memory-perf trade-offs. Would you be open to making this change? We'd be happy to review and land this. Should be reasonably straightforward:
- Add a param to the config
- Read it while wrapping the model with FSDP

tambulkar:
Yup I'll go ahead and implement this

pbontrager:
There's potentially some extra considerations here. Specific models should have good default wrapping strategies, which we don't currently have a way of supporting. Secondary is there should be a direct way to overwrite that default. For overriding the FSDP wrapping, you might run into config nesting limits, which would require using _get_component_from_path as well.

@ebsmothers do you have any thoughts on this?

pbontrager:
Sorry, my previous comment was about making the wrap_policy configurable, I misread the question. Please ignore.

PR #1024, opened 2024-05-27
Enable Configurable FSDP Sharding

tambulkar:
#### Context
- [x] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

https://github.com/pytorch/torchtune/issues/1014

#### Changelog
Add FSDP sharding options to the config

#### Test plan

Example CLI Commands For sanity Checking

```bash
tune run --nproc_per_node 2 lora_finetune_distributed --config llama3/8B_lora fsdp_sharding_strategy=test_invalid (This breaks)
tune run --nproc_per_node 2 lora_finetune_distributed --config llama3/8B_lora fsdp_sharding_strategy=NO_SHARD
tune run --nproc_per_node 2 lora_dpo_distributed --config llama3/8B_lora fsdp_sharding_strategy=NO_SHARD
tune run --nproc_per_node 2 full_finetune_distributed --config llama3/8B_full fsdp_sharding_strategy=FULL_SHARD
```


rohan-varma (commented):
This is a great idea - zero2 is used in a bunch of use cases and in general configurable sharding enables better speed / memory tradeoff. cc @ebsmothers to sign off

Also, would you mind adding testing details / details about speedups seen on your workloads? Thanks!

tambulkar:
gpu_utilization_over_time
I have attached the gpu utilization from some small test runs. My set up was
1. Gemma 2b
2. 1% of alpaca training data
3. 1 epoch
4. batch size 4
5. LORA distributed

It feels a little weird to me that fully shard didn't really use less memory than DDP for example. Maybe someone else could double check with a test run on their end. The speed of the runs makes sense to me though 

NO_SHARD < _HYBRID_SHARD_ZERO2 < HYBRID_SHARD < SHARD_GRAD_OP < FULL_SHARD

tambulkar:
Also any advice on how to unit test this? `tests/torchtune/config/*` doesnt seem like the correct place to test it

ebsmothers:
Hi @tambulkar thanks for the PR! I agree with @rohan-varma, I think this is something we want to support. I have no major concerns with the changes themselves. 

Re testing: since this is really only exposed at the recipe level, I agree that it probably doesn't make sense to add a unit test under `config/`. One option is to update our existing recipe tests: e.g. for the full finetune recipe you could update the parametrization [here](https://github.com/pytorch/torchtune/blob/6bc150dd49888c9a2bd74326237c5dc79f7c515e/tests/recipes/test_full_finetune_distributed.py#L55-L57) to pass in sharding strategy and just add one other sharding strategy for one of the models (no need to test them all). In that case I think the loss should be the same as with the default config (since the data parallel portion should be unchanged). Btw you can run these locally via e.g. `pytest -m integration_test tests/recipes/test_full_finetune_distributed` (assuming you have >1 GPU in your dev environment. If not lmk and we can figure something out)

Re the GPU utilization you're seeing.. I agree it's a bit counterintuitive. I can take a look on my end as well. Couple quick questions: (1) are you just using the default `gemma/2B_lora` config here (with the only overrides being the ones you described)? Also how are you generating the figure? Is it from WandB's native logging, or something else?

tambulkar:
Hi @ebsmothers yeah I used the gemma 2b lora config with just the overrides mentioned above. Didn't see the WandB integration so I actually just generated the graph by logging the gpu utilization using pynvml and graphing it myself.

Unfortuantely I don't really have a great multigpu set up - I just used tensordock to spin up a multigpu machine to test this out

ebsmothers:
Hey @tambulkar sorry I am just getting back to this. I ran a quick test on my end via

```
tune run --nnodes 1 --nproc_per_node 4 lora_finetune_distributed --config gemma/2B_lora \
metric_logger=torchtune.utils.metric_logging.WandBLogger metric_logger.project=lora-debug \
log_peak_memory_stats=True epochs=1 max_steps_per_epoch=100 
```

which should be pretty similar to your setup and added `fsdp_shard_strategy=SHARD_GRAD_OP` etc for other sharding strategies. Using torchtune's peak memory allocated logging I see the below logged from rank 0:

Screenshot 2024-06-10 at 2 46 24 PM

So in my case it's 

FULL_SHARD = HYBRID_SHARD < NO_SHARD < SHARD_GRAD_OP = _HYBRID_SHARD_ZERO_2

This is closer to what I'd expect than the results you got, but still the whole `NO_SHARD < SHARD_GRAD_OP` seems a bit strange to me. Gemma is a bit of a weird case too since there are tied weights (unlike most of our other models). I just kicked off a run with Llama2-7B instead and it appears that `SHARD_GRAD_OP < NO_SHARD` does hold (over the few iterations I checked). 

So anyways this is a long way to say that I think this is working as expected. Can you update the PR summary with the commands used to run the three recipes (mainly as a sanity check that nothing will be obviously broken, we have CI but don't have coverage on the DPO recipe yet)? After that I think this is good to merge.

joecummings:
@tambulkar Are you interested in finishing up this PR?

tambulkar:
Hey @joecummings sorry bout that I forgot about this - will finish it up today

tambulkar:
the lora fine-tune recipe works fine but lora dpo and full fine-tune seem to have some issues on my end

```
tune run --nproc_per_node 2 lora_dpo_distributed --config llama3/8B_lora fsdp_sharding_strategy=NO_SHARD
```
fails with
```
    chosen_input_ids = [torch.tensor(ex["chosen_input_ids"]) for ex in batch]
KeyError: 'chosen_input_ids'
```

and 
```
tune run --nproc_per_node 2 full_finetune_distributed --config llama3/8B_lora fsdp_sharding_strategy=NO_SHARD
```

fails with

```
RuntimeError: Error(s) in loading state_dict for TransformerDecoder:
	Missing key(s) in state_dict: "layers.0.attn.q_proj.lora_a.weight", ...
```

I downloaded llama3 using 
```
tune download meta-llama/Meta-Llama-3-8B-Instruct --output-dir /tmp/Meta-Llama-3-8B-Instruct
``` not sure what the issue is exactly

**Comment by felipemello1** [comment]
> and
> 
> ```
> tune run --nproc_per_node 2 full_finetune_distributed --config llama3/8B_lora fsdp_sharding_strategy=NO_SHARD
> ```
> 
> fails with
> ```

You are using a lora config for full_finetune_distributed. Running the code below with "lora_finetune_distributed" should work:

```
tune run --nproc_per_node 2 lora_finetune_distributed --config llama3/8B_lora fsdp_sharding_strategy=NO_SHARD
```

**Comment by tambulkar** [comment]
@felipemello1 that worked thanks - the DPO one should use the lora config though right?

**Comment by felipemello1** [comment]
My guess is that it is a dataset issues. @SalmanMohammadi @RdoubleA , can you confirm?

**Comment by SalmanMohammadi** [comment]
Thanks for this PR @tambulkar - this looks super cool.

@tambulkar, could you share the dataset you're using?

**Comment by felipemello1** [comment]
> could you share the dataset you're using

I think it would be the default in the config

https://github.com/pytorch/torchtune/blob/main/recipes/configs/llama3/8B_lora.yaml#L47
```
dataset:
```

**Comment by SalmanMohammadi** [comment]
Hmm, that's a instruct dataset - it should be using [this](https://github.com/pytorch/torchtune/blob/288ff4435b0cf17325b5c3b112f6859a6cdf0ea2/torchtune/datasets/_stack_exchanged_paired.py#L12), no?

I see @tambulkar is using a config for 8B_lora, wheras they should copy [this config](https://github.com/pytorch/torchtune/blob/main/recipes/configs/llama2/7B_lora_dpo.yaml) and adapt it to Llama3 8B.

**Comment by tambulkar** [comment]
Good catch - it seems to get further when I use
```
# Dataset and Sampler
dataset:
  _component_: torchtune.datasets.stack_exchanged_paired_dataset
  max_seq_len: 1024
seed: null
shuffle: True
batch_size: 4
``` as the dataset with the 8B lora config
but I get some NCLL failures - could be my set up im using runpod

**Comment by SalmanMohammadi** [comment]
Very silly question, since I'm not familiar with distributed debugging.

Is the loss in your config is the same as in the DPO config?

**Comment by tambulkar** [comment]
Good call @SalmanMohammadi but even when I use
 ```
loss:
  _component_: torchtune.modules.loss.DPOLoss
  beta: 0.1
  label_smoothing: 0
```
in my config I still get the ncll failures

tambulkar:
@felipemello1 @SalmanMohammadi is there anything else to include here?

felipemello1:
I guess we just need to make sure that the DPO script runs ncll failures, is that right? You had errors in your machine, so its not clear if its the recipe or the machine. 

I can run your recipe in my machine, to see if its fine
If its not too much work for you, maybe you can run from main the dpo recipe, to confirm that you also have these failures?

tambulkar:
@felipemello1 NCLL errors went away on new machine I spun up - 
`tune run --nproc_per_node 2 lora_dpo_distributed --config llama3/8B_lora fsdp_sharding_strategy=NO_SHARD`
starts running now I just get OOM with 2 x RTX 4090 - might still be worth running on your end as well

SalmanMohammadi:
Can you please add a quick comment in the docs at the top of each recipe (e.g. [here](https://github.com/pytorch/torchtune/blob/c3851fd9aa6027d8036523fe614e209442d92534/recipes/lora_finetune_distributed.py#L51)) about this?

Just a minimal 1-2 sentences explaining what the config parameter you're adding does, and the different options we can use.

It'll help a lot with keeping track of which features we support as we start to document them more comprehensively.

tambulkar:
Thanks for the feedback @SalmanMohammadi updated the docstrings

felipemello1 (approved):

Issue #1018, opened 2024-05-24
"Bus error (core dumped)" when saving recipe state after restarting training


calvinpelletier:
(machine: Fedora Linux, 1xH100 gpu)

Run `tune run lora_finetune_single_device --config llama3/8B_lora_single_device seed=0 epochs=4 max_steps_per_epoch=2`

Then interrupt after one epoch has been successfully saved to disk

Then run `tune run lora_finetune_single_device --config llama3/8B_lora_single_device seed=0 epochs=4 max_steps_per_epoch=2 checkpointer.checkpoint_dir=<out_dir> checkpointer.adapter_checkpoint=adapter_0.pt checkpointer.recipe_checkpoint=recipe_state.pt resume_from_checkpoint=True`

Then when writing the next checkpoint, I get `Bus error (core dumped)` consistently after writing 804mb of the 11mb recipe state.

If I remove the opt state from the recipe state, then it saves successfully. Weirdly, if I call `torch.save` on just the opt state, it also saves successfully.

Setting `mmap=False` when loading the recipe state fixes the issue.

Writing the recipe state to a new file instead of overwriting the old one also fixes the issue.

joecummings:
Closed by #1027

Issue #988, opened 2024-05-16
No param to control save checkpoints every N steps ?


apachemycat:
I can't find param to control save checkpoints every N steps, too many checkpoint files in training porcess 
anothere suggestion , maybe whole model checkpoint (merged lora checkpiont model) should only output in last train step will better ?

apachemycat:
INFO:torchtune.utils.logging:Learning rate scheduler is initialized.
1|2|Loss: 3.4330687522888184: 100%|██████████████████████████████████████████████████████████████████████████| 2/2 [00:05<00:00,  2.53s/it]
INFO:torchtune.utils.logging:**Model checkpoint of size 16.06 GB** saved to /tmp/meta-Llama-3-8B-original/meta_model_0.pt
INFO:torchtune.utils.logging:Adapter checkpoint of size 0.04 GB saved to /tmp/meta-Llama-3-8B-original/adapter_0.pt
INFO:torchtune.utils.logging:Recipe checkpoint of size 0.00 GB saved to /tmp/meta-Llama-3-8B-original/recipe_state.pt
2|2|Loss: 4.366740703582764: 100%|███████████████████████████████████████████████████████████████████████████| 2/2 [00:03<00:00,  1.98s/it]

merged model checkpoint file too large

calvinpelletier:
Hey @apachemycat, an option to save only the trainable weights for intermediate checkpoints is a great idea! We will add support for this soon.

Regarding checkpointing every N steps, this will require more planning because we would need to save the state of the data loader (i.e. which samples have been iterated over) in order to cleanly resume training from one of these checkpoints.

Another option we could consider is making the number of checkpoints that will persist on disk configurable (independent of how frequently the latest checkpoint is saved). Is something you'd find useful?

Thanks for raising this issue!

apachemycat:
1 making the number of checkpoints that will persist on disk configurable  
2  auto compare current loss with prev trained loss ,if current loss is less than prev , then remove prev whole checkpoint ,save current whole checkpoint ,
maybe choise 2 is better ?

RdoubleA:
A potential solution is being discussed in #1107

PR #1027, opened 2024-05-28
Fix for "Bus error (core dumped)" when saving recipe state

calvinpelletier:
#### Context
Quick fix for this issue: https://github.com/pytorch/torchtune/issues/1018

An alternative solution is to save recipe states with the epoch appended to the filename. This might be a better solution but there are trade-offs. There is some overlap between this issue and https://github.com/pytorch/torchtune/issues/988 so I'll make an RFC discussing some potential changes to how we do checkpointing. 

#### Changelog
- disable mmap when loading recipe state

#### Test plan

Run finetune then interrupt after 1 epoch: 
`tune run lora_finetune_single_device --config llama3/8B_lora_single_device seed=0 epochs=4 max_steps_per_epoch=2 tokenizer.path=/home/cpelletier/model/llama3-8b/original/tokenizer.model checkpointer.checkpoint_dir=/home/cpelletier/model/llama3-8b/original/ checkpointer.output_dir=/home/cpelletier/out/bus_error/`

Resume finetune: `tune run lora_finetune_single_device --config llama3/8B_lora_single_device seed=0 epochs=4 max_steps_per_epoch=2 tokenizer.path=/home/cpelletier/model/llama3-8b/original/tokenizer.model checkpointer.checkpoint_dir=/home/cpelletier/out/bus_error/ checkpointer.adapter_checkpoint=adapter_0.pt checkpointer.recipe_checkpoint=recipe_state.pt resume_from_checkpoint=True checkpointer.output_dir=/home/cpelletier/out/bus_error/`

Before the change, this resulted in `Bus error (core dumped)` when saving `recipe_state.pt` after resuming.


joecummings:
Can you take a look at updating our test to catch this kind of failure?

torchtune/utils/_checkpointing/_checkpointer_utils.py
```diff
@@ -53,7 +53,9 @@ def get_path(input_dir: Path, filename: str, missing_ok: bool = False) -> Path:
     return file_path
 
 
-def safe_torch_load(checkpoint_path: Path, weights_only: bool = True) -> Dict[str, Any]:
+def safe_torch_load(
+    checkpoint_path: Path, weights_only: bool = True, mmap: bool = True
```
kartikayk, line 57: not related to this PR, but @ebsmothers we've come back to the API of torch.load with this change. At this point I don't think there's any use of safe_torch_load other than as a single wrapper for safetensors vs not. We probably should think about refactoring this in a clean way.

kartikayk (approved):
Thanks for a) catching this, b) debugging this and c) the fix!

ebsmothers (approved):

recipes/lora_finetune_single_device.py
```diff
@@ -156,20 +156,18 @@ def _update_recipe_state(self, ckpt_dict: Dict[str, Any]) -> None:
...
             or self.max_steps_per_epoch != ckpt_dict[utils.MAX_STEPS_KEY]
         ):
             warn(
-                message="""Configured value for seed, epochs or max_steps_per_epoch
+                message="""Configured value for seed or max_steps_per_epoch
                 does not match the value stored in checkpoint."""
             )
         self.seed = utils.set_seed(seed=ckpt_dict[utils.SEED_KEY])
         self.epochs_run = ckpt_dict[utils.EPOCHS_KEY]
-        self.total_epochs = ckpt_dict[utils.TOTAL_EPOCHS_KEY]
```
calvinpelletier, line 172: @joecummings I made a quick fix to the `test_training_state_on_resume` test so that we'll notice errors like this, but it required slightly different configurations for the pre- and post-resume runs (first `epochs=2` and then `epochs=3`), which was not allowed by this `_update_recipe_state` above, so i removed the overwrite for total epochs. 

Is there a reason for this overwrite? IMO we should allow users to resume a training job with a different number of total epochs, like for example if they finish a training job and then decide they want to continue training for a few more epochs. To support this, we should also save the recipe state during the final checkpoint.

Thoughts? Should I remove the epoch overwrite for the other recipes? Or should I find a different way to test for this error?

joecummings: I 100% agree with this approach, but maybe we should opt to be overly communicative here? Log a warning if the total epochs from the ckpt_dict differ from the total epochs requested?

Issue #1028, opened 2024-05-28
Bug: 'txt' instead of 'text' for `text_completion_dataset`


SimiPixel:
In the tutorial on datasets (https://pytorch.org/torchtune/main/tutorials/datasets.html#custom-unstructured-text-corpus) both in the config and the code example the `source` keyword should be `text` and not `txt`

```yaml
# YAML config
dataset:
  _component_: torchtune.datasets.text_completion_dataset
  source: text 
  data_files: path/to/my_data.txt
```

RdoubleA:
Ah good catch. Thank you @SimiPixel for flagging this. Feel free to open a PR to fix this and I can review it if you'd like :) Or I can go ahead and patch it, whichever you prefer.

SimiPixel:
Yes, i can do that, however, there is still a couple of other things that are not working correctly here. Let me elaborate:
Let's assume we start with newest Pytorch (v2.3.0) installed and with torchtune installed from git at the time of 28.05.24.

---
>  tune run full_finetune_single_device --config llama3/8B_full_single_device
```
  File "/data/vy43meti/miniforge3/envs/llama3/lib/python3.11/site-packages/torchtune/config/_parse.py", line 50, in wrapper
    sys.exit(recipe_main(conf))
             ^^^^^^^^^^^^^^^^^
  File "/data/vy43meti/miniforge3/envs/llama3/lib/python3.11/site-packages/recipes/full_finetune_single_device.py", line 504, in recipe_main
    recipe.setup(cfg=cfg)
  File "/data/vy43meti/miniforge3/envs/llama3/lib/python3.11/site-packages/recipes/full_finetune_single_device.py", line 201, in setup
    self._optimizer = self._setup_optimizer(
                      ^^^^^^^^^^^^^^^^^^^^^^
  File "/data/vy43meti/miniforge3/envs/llama3/lib/python3.11/site-packages/recipes/full_finetune_single_device.py", line 283, in _setup_optimizer
    optim_dict = {
                 ^
  File "/data/vy43meti/miniforge3/envs/llama3/lib/python3.11/site-packages/recipes/full_finetune_single_device.py", line 284, in <dictcomp>
    p: config.instantiate(cfg_optimizer, [p])
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #1031, opened 2024-05-29
Remove config fields with CLI overrides using ~

RdoubleA:
## Problem
We provide an easy way to update or add config fields from a CLI override using the `key=value` syntax. However, we have no way of removing an existing field from a yaml config through the CLI. This was recently highlighted in #1028 with `text_completion_dataset` CLI examples in our tutorials. `text_completion_dataset` does not use the parameter `train_on_input`, but this is a default parameter set in all our configs that's supported by most of our datasets and is exposed intentionally to highlight this commonly used feature. If a user wants to quickly override our built-in configs with `text_completion_dataset`, they need to remove this parameter.

## Possible approaches
There are several ways we can approach this problem:
- **Remove `train_on_input` from all our configs**. `alpaca_dataset` already has this set to True by default, and so it is not needed. This will address our issue with `text_completion_dataset` (and other datasets that do not use this parameter) but we lose the visibility of this feature
- **Overriding a component clears all its kwargs by default**. Currently, when we override a component, like dataset with `dataset=torchtune.datasets.text_completion_dataset`, it keeps the previous kwargs in the config file. In this case, `train_on_input` remains, causing the error. We could instead delete all kwargs, requiring the users to specify them again in the CLI overrides, but this can get very tedious for components like the checkpointer which has many kwargs. Additionally, if we decide to enable nested components, this approach is just a no-go.
- **Ask users to just clone and modify**. Removing a field from a built-in config might be considered advanced behavior that should be handled directly in the config file. CLI overrides are meant to be a quick way to change fields but not a replacement for complex changes to configs. We could draw the line here.
- **Adding a way to remove params from CLI override**. This has been discussed numerous times offline as a solution to pain points in quick experimentation. The main downside with this is that it can make CLI overrides a bit intrusive and awkward

## Changes in this PR
- enable removing config fields in the YAML with CLI overrides using the `~` flag. This is inspired by how [Hydra](https://hydra.cc/docs/advanced/override_grammar/basic/#modifying-the-config-object) handles this. 
- remove `train_on_input` from all configs as only alpaca uses this and it defaults to true in the builder anyway
- Update all docs on text_completion_dataset, bug fixes from #1028   

#### Context
- [x] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

#### Test plan

	- [ ] 	- [ ] 	- [ ] 	- [ ] 	- [ ] 	- [ ] 	- [ ] 	- [ ] 	- [ ] 	- [ ] 	- [ ] 	- [ ] 	- [ ] 	- [ ] 	- [ ] 	- [ ] 	- [ ] 	- [ ] 	- [ ] 	- [ ] 	- [ ] 	- [ ] 	- [ ] 	- [ ] 	- [ ] 	- [ ] 	- [ ] 	- [ ] 	- [ ] 	- [ ] 	- [ ] 	- [ ] 	- [ ] 	- [ ] 	- [ ] 	- [ ] 	- [ ] 	- [ ] 	- [ ] 	- [ ] 	- [ ] 	- [ ] 	- [ ] 	- [ ]


docs/source/deep_dives/configs.rst
```diff
@@ -236,6 +231,26 @@ If you would like to override a class or function in the config that is instanti
...
+    dataset=torchtune.datasets.slimorca_dataset dataset.train_on_input=True
+
+Removing config fields
+^^^^^^^^^^^^^^^^^^^^^^
+You may need to remove certain parameters from the config when changing components
+through overrides that require different keyword arguments. You can do so by using
+the `~` flag and specify the dotpath of the config field you would like to remove.
+For example, if you want to override a built-in config and use :func:`~torchtune.datasets.text_completion_dataset`
+which does not use the ``train_on_input`` parameter, you can delete it from the
+command-line.
```
ebsmothers, line 252: Wait this is kinda no longer a relevant example, right?

docs/source/tutorials/datasets.rst
```diff
@@ -184,24 +184,26 @@ files:
...
 
 .. code-block:: bash
 
     # Command line
-    tune run full_finetune_single_device --config llama3/8B_full_single_device \
-    dataset=torchtune.datasets.text_completion_dataset dataset.source=txt \
-    dataset.data_files=path/to/my_data.txt
+    tune run full_finetune_distributed --config llama3/8B_full \
+    dataset=torchtune.datasets.text_completion_dataset dataset.source=text \
+    dataset.data_files=path/to/my_data.txt dataset.split=train
```
ebsmothers, line 206: Why this change? Shouldn't we have number of devices in here if we're doing this?

ebsmothers (approved):
Couple small comments but no major concerns. Also you may need to merge with latest main, there are a couple more configs under dev/ that will need to be updated as well

docs/source/tutorials/datasets.rst
RdoubleA: single device recipe requires bitsandbytes, which needs to be installed separately by the user. I'd rather not use that as an example. But yes, I need to add number of devices..

Issue #1032, opened 2024-05-30
Circular import error when importing Llama3 model


parthsarthi03:
Hello! Thank you for the excellent work.

I'm trying to load the Llama3 model from the models directory for some experiments, but I'm encountering a circular import issue.

Steps to reproduce:
1. Import the Llama3 model directly using `from torchtune.models.llama3 import llama3`.

Behavior:
A circular import error is raised when attempting to import the Llama3 model.

Troubleshooting attempts:
I have tried the following alternative import statements in the `torchtune/torchtune/models/__init__.py` file, but I still encounter the same circular import error:
- `from torchtune.models import llama3`
- `from .llama3 import llama3`

 Any help would be appreciated! 
```

----> 1 from torchtune.models.llama3 import llama3

File ~/ps/torchtune/torchtune/models/llama3/__init__.py:7
      1 # Copyright (c) Meta Platforms, Inc. and affiliates.
      2 # All rights reserved.
      3 #
      4 # This source code is licensed under the BSD-style license found in the
      5 # LICENSE file in the root directory of this source tree.
----> 7 from ._component_builders import llama3, lora_llama3
      9 from ._model_builders import (  # noqa
     10     llama3_70b,
     11     llama3_8b,
   (...)
     15     qlora_llama3_8b,
     16 )
     17 from ._model_utils import scale_hidden_dim_for_mlp

File ~/ps/torchtune/torchtune/models/llama3/_component_builders.py:26
     14 from torchtune.modules import (
     15     CausalSelfAttention,
     16     FeedForward,
   (...)
     21     TransformerDecoderLayer,
     22 )
     24 from torchtune.modules.common_utils import reparametrize_as_dtype_state_dict_post_hook
---> 26 from torchtune.modules.peft import LORA_ATTN_MODULES, LoRALinear
     28 """
     29 Component builders for the Llama3 model and popular variants such as LoRA.
     30 
   (...)
     37 the building blocks simple.
     38 """
     41 # ------------------ Vanilla Llama3 ------------------

File ~/ps/torchtune/torchtune/modules/peft/__init__.py:7
      1 # Copyright (c) Meta Platforms, Inc. and affiliates.
      2 # All rights reserved.
      3 #
      4 # This source code is licensed under the BSD-style license found in the
      5 # LICENSE file in the root directory of this source tree.
----> 7 from .lora import LoRALinear
      8 from .peft_utils import (  # noqa
      9     AdapterModule,
     10     disable_adapter,
   (...)
[...]

PR #1035, opened 2024-05-30
Move nf4 op registration from utils to modules

ebsmothers:
Fixing a circular import error due to nf4 ops registration happening in torchtune/utils and being used in LoRALinear under torchtune/modules. We shouldn't have a modules -> utils dependency and this isn't used anywhere else, so move under modules (where it originally lived anyways) 

Addresses #1032


RdoubleA (approved):
Looks good to me!

PR #1036, opened 2024-05-30
SimPO : Simple Preference Optimization with a Reference-Free Reward

nivibilla:
#### Context
- [x] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

#1037

#### Changelog
New training method, simPO : https://arxiv.org/html/2405.14734v1

#### Test plan

- [ ] remove TODO
- [ ] possibly refactor to merge DPO and simPO

### DIFF
--- a/recipes/configs/llama2/7B_lora_simPO.yaml
+++ b/recipes/configs/llama2/7B_lora_simPO.yaml
@@ -1,0 +1,87 @@
+# Config for multi-device LoRA DPO alignment in lora_dpo_distributed.py
+# using a Llama2 7B model
+#
+# This config assumes that you've run the following command before launching
+# this run:
+#   tune download meta-llama/Llama-2-7b-hf --output-dir /tmp/Llama-2-7b-hf --hf-token <HF_TOKEN>
+#
+# To launch on 2 devices, run the following command from root:
+#   tune run --nnodes 1 --nproc_per_node 2 lora_dpo_distributed --config llama2/7B_lora_dpo
+#
+# You can add specific overrides through the command line. For example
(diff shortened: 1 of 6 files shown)


nivibilla:
Some open questions:

- The only real change is the loss function. And an omission of two lines in the trainer. So I feel like there is a lot of duplicated code, is it possible to combine dpo and simPO into one trainer and based on the config apply the loss?
- I wasn't sure but for simPO since there is no need for reference log probs this [chunk](https://github.com/pytorch/torchtune/blob/8c6e57746a87ab284162b37d8f5eed609058ca0a/recipes/lora_simPO_distributed.py#L631) can be omitted right?
- What is the best way to test correctness against the princeton nlp release version?

nivibilla:
@kartikayk I will take that previous offer to review and help 😆

pbontrager:
I wanted to followup here to help this from getting stuck.
- I believe we want to add this as its own loss and make this work with the DPO recipe. Maybe that could be given a more general name as you mentioned on the issue.
- For correctness, as a start, you should be able to basically test your unit test values on the simPO loss and the reference simPO loss and get the same outcomes. After you see that the individual components are the same, try and do a small training run on both code bases and you should see very similar loss curves. You can attach all of this verification to the PR description.

ebsmothers:
Hi @nivibilla are you still working on adding this? Do @pbontrager's suggestions make sense? Happy to provide any additional suggestions as needed here

nivibilla:
Hey @ebsmothers Im currently occupied with other stuff tbh. Not sure when I will get round to this. I can close this PR and let someone else do it if necessary

PR #1046, opened 2024-06-04
[LoRA] del logits to save peak memory

weifengpy:
`logits.shape=(bs, seq_len, vocab_size)`. In my snapshot, logits.shape=(2, 512, 32000) in float32, 6% of peak memory, see blue chunk

`del logits` could lower peak memory otherwise it won't be freed util the end of the inner for loop

**Snapshot** of memory traces showing logits
Screenshot 2024-06-04 at 11 24 33 AM

### DIFF
--- a/recipes/lora_finetune_distributed.py
+++ b/recipes/lora_finetune_distributed.py
@@ -590,6 +590,8 @@ class LoRAFinetuneRecipeDistributed(FTRecipeInterface):
                 logits = logits.transpose(1, 2)
                 # Compute loss
                 loss = self._loss_fn(logits, labels)
+                # free logits otherwise it peaks backward memory
+                del logits
 
                 loss = loss / self._gradient_accumulation_steps
                 running_loss += loss


joecummings:
Can we use this in the `full_finetune_distributed.py` recipe, too?

kartikayk (commented):
@weifengpy thanks for the analysis!

I'm having a little trouble parsing the snapshots you linked in the context. Are the red, yellow and blue bumps related to the logits overhead? Is this for the default config with BS=2 and Seq Len < 100? I'd imagine for the current models this isn't a huge hit? For example for seq_len=4096, bs=4 this would be around 4096 * 4 * 4 bytes? But as seq length gets longer this would become a concern? Let me know if I did the math wrong :)

weifengpy:
> Can we use this in the `full_finetune_distributed.py` recipe, too?

yes, this is applicable to `full_finetune_distributed.py` too

weifengpy:
> For example for seq_len=4096, bs=4 this would be around 4096 * 4 * 4 bytes

Good question. Just checked the formula. `logits.shape`=(bs, seq_len, vocab_size). In my snapshot, `logits.shape`=(2, 512, 32000) in float32, that's `120MB`, 6% of the peak memory

kartikayk:
Ah dang, I forgot the vocab dimension. This then becomes very restrictive esp for Llama3 where the vocab size is 128K. Awesome catch! Love this!

kartikayk (approved):

PR #1048, opened 2024-06-04
Add ``log_config`` to config API page

joecummings:
#### Context
- [ ] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

#### Changelog
* Add ``log_config`` to config API page
* Update docstring for ``log_config``

#### Test plan
EYES

### DIFF
--- a/torchtune/config/__init__.py
+++ b/torchtune/config/__init__.py
@@ -10,8 +10,8 @@ from ._utils import log_config
 from ._validate import validate
 
 __all__ = [
-    "parse",
     "instantiate",
-    "validate",
+    "parse",
     "log_config",
+    "validate",
 ]
--- a/torchtune/config/_utils.py
+++ b/torchtune/config/_utils.py
@@ -17,7 +17,7 @@ from torchtune.utils import get_logger, get_world_size_and_rank
 
 def log_config(recipe_name: str, cfg: DictConfig) -> None:
     """
-    Logs the parsed config to rank zero.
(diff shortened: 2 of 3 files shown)


torchtune/config/_utils.py
```diff
@@ -17,7 +17,7 @@
 
 def log_config(recipe_name: str, cfg: DictConfig) -> None:
     """
-    Logs the parsed config to rank zero.
+    Logs the resolved config to rank zero.
```
pbontrager, line 20: Can you say what resolved means here? e.g. A resolved config is the provided config updated with any command line overrides.

pbontrager (commented):
I looks good. I can't accept this PR though because the docs aren't rendering for me.

torchtune/config/__init__.py
```diff
@@ -10,8 +10,8 @@
 from ._validate import validate
 
 __all__ = [
-    "parse",
     "instantiate",
-    "validate",
+    "parse",
     "log_config",
+    "validate",
```
RdoubleA, line 16: I wonder if we should just put an assert at the bottom of all our `__init__.py` to ensure this remains alphabeticaly

RdoubleA (approved):

torchtune/config/__init__.py
joecummings: I'm sure this is a flake8 command somewhere - I'll investigate.

PR #1057, opened 2024-06-05
TODO: [multimodal][llava-next] image processor

felipemello1:
#### Context
- [x] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

As a step to bring multimodality to torchtune through llava-next, in this PR I added the **image processing** step used to prepare raw images for the image encoder model.

> Speed: The implemented processor is 6x faster than HF implementation for an input image of size (2560, 1920) and chunk_size 224.

I didn't measure exactly what caused the speed up, but I suspect its a combination of 3 things: i) vectorized functions, ii) pytorch modules + tensors instead of HF native modules + np.array, iii) removal of many redundant checks, like checking if channels are first/last before many operations.

proposed torchtune implementation:
```
from torchtune.models.llava_next.image_processor import ImageProcessor

image_transforms_tune = ImageProcessor(
    patch_size=336,
    possible_resolutions=[
        (224, 896),
        (448, 448),
    ]
    **params
)

processed_images = []
for image in images:

    #input: (PIL/np.array/torch.array), grayscale, ch first or last
    #output torch.tensor(num_patches, 3, patch_size, patch_size)

    output_tune = image_transforms_tune.preprocess(image) 
    processed_images.append(output_tune)

# data collator step
batch = your_collator(processed_images, pad_patches = True)
```

#### Changelog
1) image_processor: Added ```torchtune/models/llava_next/image_processor.py``` based on [HF's implementation](https://github.com/huggingface/transformers/blob/main/src/transformers/models/llava_next/image_processing_llava_next.py), which is what [llava-next uses for inference](https://github.com/LLaVA-VL/LLaVA-NeXT/blob/19ddd2e3d3093691cb9affe2bd506942685579b8/docs/LLaVA-NeXT.md).

2) image_transforms_utils: Added two util functions to estimate possible image resolutions based on max_num_chuncks in ```torchtune/utils/image_transforms_utils.py```. This is not part of llava or HF implementation, which require predefined possible_resolutions, but it felt like a nice/easy add in case a user wants to play with number of chunks + patch_size.

3) Simplification: It is simpler and, therefore, with limited functionality. For this implementation, I worked with the [default values here defined here](https://huggingface.co/lmms-lab/llama3-llava-next-8b/blob/main/preprocessor_config.json).

Assumptions made:
- Patches are always squares;
- Resizing and cropping always use the same patch_size;
- do_cropping and do_resizing always set to True, to guarantee patches of the same size. In HF, setting it to false would break the code if image_size was not a multiple of patch_size;
- Output is always of shape [num_patches, 3, H, W]

4) input: Its is important to notice that HF processor accepts List[image], and therefore, also has an option to pad all the images to the same number of chunks. In this implementation, we only accept one image at a time, and batching (padding for num_chunks) has to be done elsewhere.

#### Test plan

- Got results using HF functions and added unit tests confirming that torchtune functions yield the same value. 
- Also compared all patches generated, using multiple configurations, greyscale and RGB, confirming that it produced the same output. There are mild differences of some pixels (~0.2%) being +1/-1 from HF before normalization.

I am working on getting colab access to share a notebook with the results. Currently I only have intern bento.

```
pytest tests/torchtune/models/test_llava_image_processor.py
>17 passed in 0.31s 
```

```
pytest tests/torchtune/utils/test_image_transforms_utils.py
>3 passed in 0.02s
```
- [] run recipe tests via `pytest tests -m integration_test`

image


felipemello1:
Forgot to add torchvision to the requirements

ebsmothers (commented):
This is great! Will defer to others more familiar with Llava code on details of `GetImagePatches`, but really love the thoroughness in both your testing and PR summary.

ebsmothers:
> Speed: The implemented processor is 6x faster than HF implementation for an input image of size (2560, 1920) and chunk_size 224.

Yeah would definitely be interested to know more about why this is the case. I know v2 torchvision transforms are supposed to be quite a bit faster and looks like HF is not using those.

> I am working on getting colab access to share a notebook with the results

It should also be possible to set up a regular Jupyter notebook and tunnel from your server, lmk if you need any pointers on this.

torchtune/models/llava_next/image_processor.py
```diff
@@ -0,0 +1,344 @@
...
+            [self.patch_size, self.patch_size],
+            interpolation=self.resample,
+        ).float()
+
+        image_patches = torch.stack([patch_sized_image] + image_patches)
+
+        return image_patches
+
+
+class ImageProcessor(nn.Module):
```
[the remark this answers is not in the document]
felipemello1: Thats a good question! When I chatted with @pbontrager , we talked about having nn.modules and using nn.sequential to call the different processing steps. I am not sure if it helps with torch.compile either (but i dindt see any improvements in speed using it). So I think that using __call__ makes sense.

torchtune/models/llava_next/image_processor.py
```diff
@@ -0,0 +1,344 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
```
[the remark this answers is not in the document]
felipemello1: I personally liked the style that huggingface uses, where everything that is specific to the model, stays close to the model, like here: https://github.com/huggingface/transformers/tree/main/src/transformers/models/llava_next

And util functions that can be used by many models can go in more generic folders: https://github.com/huggingface/transformers/blob/v4.41.1/src/transformers/image_utils.py

happy to change if you guys want me to

torchtune/utils/image_transforms_utils.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+def factors(n: int):
+    """Return all factors of a number."""
+    return set(
+        reduce(
+            list.__add__,
+            ([i, n // i] for i in range(1, int(n**0.5) + 1) if n % i == 0),
+        )
+    )
```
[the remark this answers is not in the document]
felipemello1: personally, I am not a fan of functions like this! haha

But I copied it from somewhere else and didnt come back to refactor it. I will make it easier to read.

torchtune/models/llava_next/image_processor.py
```diff
@@ -0,0 +1,344 @@
...
+
+        self.patchfy = GetImagePatches(
+            possible_resolutions=possible_resolutions,
+            patch_size=patch_size,
+            resample=resample,
+        )
+
+        _preprocess = []
+        if do_rescale:
+            _preprocess.append(v2.Normalize(mean=[0] * 3, std=[1 / rescale_factor] * 3))
```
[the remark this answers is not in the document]
felipemello1: Good question! I think that people usually just do image = image / rescaling_factor, but since I am using nn.sequential, I wanted to leverage that to avoid creating a new function of adding more code to forward().

Let me think a bit about it.

torchtune/models/llava_next/image_processor.py
```diff
@@ -0,0 +1,344 @@
...
+        if isinstance(resample, torchvision.transforms.InterpolationMode):
+            self.resample = resample
+        if resample == "bilinear":
+            self.resample = torchvision.transforms.InterpolationMode.BILINEAR
+        elif resample == "bicubic":
+            self.resample = torchvision.transforms.InterpolationMode.BICUBIC
+        else:
+            raise ValueError(
+                "resample must be of type torchvision.transforms.InterpolationMode or ['bilinear', 'bicubic]."
+            )
```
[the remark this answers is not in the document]
felipemello1: I was thinking that it could make it more convenient for the user. Instead of having to another library just to define the interpolation, for example, it can just be a string. I am fine with removing it though.

torchtune/models/llava_next/image_processor.py
```diff
@@ -0,0 +1,344 @@
...
+        Class used to divide and image with variable resolutions into patches of equal sizes,
+        including the original image resized to [patch_size, patch_size].
+
+        Forward returns a tensor of shape [num_patches + 1, channels, patch_size, patch_size].
+
+        Args:
+            possible_resolutions (List[Tuple[int, int]]): List of possible resolutions as [height, width].
+            patch_size (int): Size of the patches to divide the image into.
+            resample (Union[str, torchvision.transforms.InterpolationMode]):
+            Resampling method, either 'bilinear', 'bicubic' or torchvision.transforms.InterpolationMode.
```
RdoubleA, line 47: nit: tab here

torchtune/models/llava_next/image_processor.py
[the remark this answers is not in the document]
RdoubleA: Another option is to keep it as string only, then access the enum with the string:
```
In [5]: InterpolationMode["BILINEAR"]
Out[5]: <InterpolationMode.BILINEAR: 'bilinear'>
```
that way you don't need to maintain these if statements

torchtune/models/llava_next/image_processor.py
```diff
@@ -0,0 +1,344 @@
...
+        self.center_crop = v2.CenterCrop(size=patch_size)
+
+        self.patch_size = patch_size
+
+    @staticmethod
+    def _get_new_size_without_distortion(
+        image_size: Tuple[int, int],
+        target_resolution: List[int],
+    ) -> Tuple[int, int]:
+
```
RdoubleA, line 72: nit: remove newline

torchtune/models/llava_next/image_processor.py
```diff
@@ -0,0 +1,344 @@
...
+        """
+        Determines the maximum dimensions to which an image can be resized without distorting its
+        aspect ratio, based on the target resolution.
+
+        Args:
+            image_size (Tuple[int, int]): The original dimensions of the image (height, width).
+            target_resolution (List[int]): The desired resolution to fit the image into (height, width).
+        Returns:
+            Tuple[int, int]: The optimal dimensions (height, width) to which the image should be resized.
+        Example:
```
RdoubleA, line 82: awesome example

torchtune/models/llava_next/image_processor.py
```diff
@@ -0,0 +1,344 @@
...
+            (450, 338)
+        """
+
+        original_height, original_width = image_size
+        target_height, target_width = target_resolution
+
+        scale_w = target_width / original_width
+        scale_h = target_height / original_height
+
+        if scale_w < scale_h:
```
RdoubleA, line 95: nit: one quick comment explaining this would be helpful, explaining that you only resize to the shortest edge

torchtune/models/llava_next/image_processor.py
```diff
@@ -0,0 +1,344 @@
...
+            patch_size (int): Size of each patch.
+        Returns:
+            List[torch.Tensor]: List of image patches.
+        Example:
+            >>> image = torch.rand(3, 200, 300)
+            >>> patches = _divide_to_patches(image, patch_size=50)
+            >>> len(patches)
+            24
+            >>> image = torch.rand(3, 400, 500)
+            >>> patches = _divide_to_patches(image, patch_size=450)
```
RdoubleA, line 121: why does this return 2 patches if 450 > 400? maybe it would be helpful to also print the patch dims? Also what happens if the image dim does not divide evenly by patch size, do we just drop the remainder? We don't resize it?

torchtune/models/llava_next/image_processor.py
```diff
@@ -0,0 +1,344 @@
...
+            2
+        """
+
+        channel_size, height, width = image.shape
+
+        patches = []
+        for i in range(0, height, patch_size):
+            for j in range(0, width, patch_size):
+                patch = image[:, i : i + patch_size, j : j + patch_size]
+                patches.append(patch)
```
RdoubleA, line 132: try looking into nn.Unfold: https://pytorch.org/docs/stable/generated/torch.nn.Unfold.html
although this code is very clear, I'm ok with keeping this for user clarity

torchtune/models/llava_next/image_processor.py
```diff
@@ -0,0 +1,344 @@
...
+    @staticmethod
+    def _select_best_resolution(
+        original_size: Tuple[int, int], possible_resolutions: torch.Tensor
+    ) -> List[int]:
+        """
+        Selects the best resolution from a list of possible resolutions based on the original size.
+
+        This is done by calculating the effective and wasted resolution for each possible resolution.
+
+        The best fit resolution is the one that maximizes the effective resolution and minimizes the wasted resolution.
```
RdoubleA, line 145: do you mind explaining "effective" and "wasted" resolution a bit more? Either here or in the code or both

torchtune/models/llava_next/image_processor.py
```diff
@@ -0,0 +1,344 @@
...
+            )
+
+        self.center_crop = v2.CenterCrop(size=patch_size)
+
+        self.patch_size = patch_size
+
+    @staticmethod
+    def _get_new_size_without_distortion(
+        image_size: Tuple[int, int],
+        target_resolution: List[int],
```
RdoubleA, line 70: why is this not a tuple of ints?

torchtune/models/llava_next/image_processor.py
```diff
@@ -0,0 +1,344 @@
...
+        for i in range(0, height, patch_size):
+            for j in range(0, width, patch_size):
+                patch = image[:, i : i + patch_size, j : j + patch_size]
+                patches.append(patch)
+
+        return patches
+
+    @staticmethod
+    def _select_best_resolution(
+        original_size: Tuple[int, int], possible_resolutions: torch.Tensor
```
RdoubleA, line 138: you've made possible_resolutions List[Tuple[int, int]] elsewhere, why tensor here?

torchtune/models/llava_next/image_processor.py
```diff
@@ -0,0 +1,344 @@
...
+        possible_resolutions: Optional[Tuple[Tuple[int, int]]] = None,
+        max_num_chunks: int = 4,
+        resample: str = "bicubic",
+        do_rescale: bool = True,
+        rescale_factor: Union[int, float] = 1 / 255,
+        do_normalize: bool = True,
+        image_mean: Optional[Union[float, Tuple[float]]] = None,
+        image_std: Optional[Union[float, Tuple[float]]] = None,
+    ) -> None:
+        """
```
RdoubleA, line 285: add a description of what this is doing to the docstring

torchtune/models/llava_next/image_processor.py
```diff
@@ -0,0 +1,344 @@
...
+            do_normalize (bool): Flag to determine whether to normalize the image.
+            image_mean (Optional[Union[float, List[float]]]): Mean values for normalization.
+            image_std (Optional[Union[float, List[float]]]): Standard deviation values for normalization.
+        """
+        super().__init__()
+
+        logger.info("Initializating ImageProcessor...")
+
+        if image_mean is None:
+            image_mean = [0.48145466, 0.4578275, 0.40821073]
```
RdoubleA, line 303: where are these magic numbers from and why do we use these?

torchtune/models/llava_next/image_processor.py
```diff
@@ -0,0 +1,344 @@
...
+            _preprocess.append(v2.Normalize(mean=[0] * 3, std=[1 / rescale_factor] * 3))
+
+        if do_normalize:
+            _preprocess.append(v2.Normalize(mean=image_mean, std=image_std))
+
+        self._preprocess = nn.Sequential(*_preprocess)
+
+    def preprocess(
+        self, image: ImageInput
+    ) -> Dict[str, torch.Tensor | Tuple[int, int]]:
```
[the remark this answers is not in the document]
RdoubleA: `|` is Gen Z, `Union` is millennial

torchtune/models/llava_next/image_processor.py
```diff
@@ -0,0 +1,344 @@
...
+
+        # Make image have dimension [3, H, W]. Input can be grayscale, RGB, channels-first or last.
+        image = F.grayscale_to_rgb_image(F.to_image(image))
```
RdoubleA, line 337: why is this class not just `nn.Sequential` of torchvision transforms? the image patch seems to be a custom transform, the `F` functions here are also transforms, and `preprocess` is just normalize.

I know this is coupled with llava next, but if this will be our primary example for image transforms that users will emulate, then we need to consider how they can look at these, copy, and easily extend for their own use case

RdoubleA (commented):
Enjoyed this PR, all your comments and examples were helpful in just learning about the patching and resizing, kudos to you :)

torchtune/models/llava_next/image_processor.py
felipemello1: I agree! I thought about using unfold and have it vectorized, but llava preprocessor patches first, and only then crops/resizes/etc.

So if you have  an image of 800x600, and your patch_size is 350x350, the output will only cover 750x350 of the image.

Let me know if that makes sense.

torchtune/models/llava_next/image_processor.py
felipemello1: good catch! There are 3 places where possible_resolutions is an input:

1. ImageProcessor
2. GetImagePatches
3. _select_best_resolution

The user enters this in ImageProcessor, which uses it to initialize GetImagePatches.

in __init__ of GetImagePatches, I make it a tensor. Then, for every call of _select_best_resolution, we can use it as a tensor, since this function is vectorized.

We could ask the user to input it a torch.Tensor in ImageProcessor, but I feel that it would make their lives harder.

Does it make sense?

torchtune/models/llava_next/image_processor.py
felipemello1: My confident guess is that this is the mean/std of the dataset used to pretrain the image encoder, and should be used for any subsequent training/inference. Kind like the learned batch normalization parameters.

The reason I dont put it as default is because its bad practice to have lists there, since they are mutable.

It comes from the model config in HF

torchtune/models/llava_next/image_processor.py
felipemello1: I was debating myself about it.

The reason I didnt add them to nn.Sequential is because they are not compatible with torch.compile / jit.script, but nn.Sequential is.

Alternatively, I can use v2.compose(*transforms) instead of nn.Sequential(*transforms), but then torch.compile / jit.script has no effect (i guess). Thas the difference between compose and sequential.

I tried compiling it, and didnt see any changes in performance, so maybe that doesnt matter? I am happy to discuss it more / change it.

here is the doc: https://pytorch.org/vision/stable/generated/torchvision.transforms.v2.ToImage.html#torchvision.transforms.v2.ToImage

torchtune/models/llava_next/image_processor.py
RdoubleA: I see, makes sense. Could we place it as a constant at the top of the file and make it a tuple?

torchtune/models/llava_next/image_processor.py
```diff
@@ -0,0 +1,344 @@
...
+class GetImagePatches(nn.Module):
+    def __init__(
+        self,
+        possible_resolutions: List[Tuple[int, int]],
+        patch_size: int,
+        resample: Union[str, torchvision.transforms.InterpolationMode] = "bilinear",
+    ) -> None:
+        super().__init__()
+        """
+        Class used to divide and image with variable resolutions into patches of equal sizes,
```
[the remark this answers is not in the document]
felipemello1: It should take a single image. I guess what I meant to say is that the image can have any size, eg 800x400, 600x1200, etc. I will reword it.

torchtune/models/llava_next/image_processor.py
```diff
@@ -0,0 +1,344 @@
...
+        including the original image resized to [patch_size, patch_size].
```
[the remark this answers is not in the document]
felipemello1: a image is divided into N patches, and the the image itself is resized to patch_size X patch_size and added to the patches. So if you image is 400x400, and your patch is 200x200, then you will have 4 patches + 1. Final tensor is of size (5, 3, 200, 200)

torchtune/models/llava_next/image_processor.py
felipemello1: we resize it in later steps. The patches have unequal sizes. I tried to express it in the docstring, but I agree that adding it to the examples would help.

torchtune/models/llava_next/image_processor.py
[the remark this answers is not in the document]
RdoubleA: I think keeping in models makes sense since they are model specific. We can move any general utilities out into data/ or modules/. It's the same idea as how Gemma needed its own decoder, Phi3 needed its own tokenizer, and they were kept in the models folder.

torchtune/models/llava_next/image_processor.py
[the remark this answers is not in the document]
RdoubleA: Makes sense, maybe make it more explicit? "including an additional patch of the original image resized"

torchtune/models/llava_next/image_processor.py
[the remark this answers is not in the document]
ebsmothers: Drive-by comment cause I just realized this on another PR: I don't think we can use `|` in type annotation yet cause we still run Python 3.8 and 3.9 in our CI. Millennials live to fight another day (for now)

felipemello1:
converting it to a draft until we come back to llava

Issue #1063, opened 2024-06-06
[Bug] Phi3 tokenizer adds a extra start token when `ignore_system_prompts=True`


hmosousa:
Hi there,

When running the `phi3_mini_tokenizer` I noticed that if the messages have a `"system"` message and the tokenizer parameter `ignore_system_prompts` is set to `True` the tokenizer produces an extra start of turn token. Code to replicate the bug:

```python
from torchtune.data import Message
from torchtune.models.phi3 import phi3_mini_tokenizer

messages = [
    Message(role="system", content="system content"),
    Message(role="user", content="users content")
]
tokenizer = phi3_mini_tokenizer("path/to/tokenizer.model")
input_ids, _ = tokenizer.tokenize_messages(messages, ignore_system_prompts=True)
print(input_ids)  # [1, 1, 32010, 29871, 13, 7193, 2793, 32007, 29871, 13]
```

One way to fix this would be to add the condition `if message.role == "system" and ignore_system_prompts: continue` at the beginning of the message loop:

https://github.com/pytorch/torchtune/blob/8f93a1c748d255f2b932e02f7de14a7f4522e64e/torchtune/models/phi3/_sentencepiece.py#L173

Should I create a PR with that solution?

joecummings:
Great catch! We'd love to accept a PR with this change.

PR #1065, opened 2024-06-06
Fix extra start token in phi3 tokenizer. Address #1063

hmosousa:
#### Context
- [ ] add a new feature
- [ x] fix a bug
- [ ] other (please add here)

Issue #1063

#### Changelog

#### Test plan

### DIFF
--- a/torchtune/models/phi3/_sentencepiece.py
+++ b/torchtune/models/phi3/_sentencepiece.py
@@ -171,6 +171,10 @@ class Phi3MiniSentencePieceTokenizer:
         new_line_token_id = self.encode("\n", add_bos=False, add_eos=False)
 
         for message in messages:
+            # Skip system prompt
+            if ignore_system_prompts and message.role == "system":
+                continue
+
             # Prepend BOS on start of new turns
             if start_of_turn:
                 tokenized_messages.append(self.bos_id)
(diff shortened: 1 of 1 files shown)


joecummings:
Can you include a gist with your test confirming this works?

Thanks!

hmosousa:
Sure. Gist follows:

```python 
from torchtune.data import Message
from torchtune.models.phi3 import phi3_mini_tokenizer

messages = [
    Message(role="system", content="system content"),
    Message(role="user", content="users content")
]
tokenizer = phi3_mini_tokenizer("tmp/tokenizer.model")
input_ids, _ = tokenizer.tokenize_messages(messages, ignore_system_prompts=False)
print(input_ids)  # [1, 32010, 29871, 13, 7193, 2793, 32007, 29871, 13]
```

RdoubleA:
> Sure. Gist follows:
> 
> ```python
> from torchtune.data import Message
> from torchtune.models.phi3 import phi3_mini_tokenizer
> 
> messages = [
>     Message(role="system", content="system content"),
>     Message(role="user", content="users content")
> ]
> tokenizer = phi3_mini_tokenizer("tmp/tokenizer.model")
> input_ids, _ = tokenizer.tokenize_messages(messages, ignore_system_prompts=False)
> print(input_ids)  # [1, 32010, 29871, 13, 7193, 2793, 32007, 29871, 13]
> ```

why not just add this as a quick test? you can use the sentencepiece test file we use here: https://github.com/pytorch/torchtune/blob/main/tests/torchtune/modules/tokenizers/test_sentencepiece.py

ebsmothers:
@hmosousa thanks for fixing this bug! I am actually gonna veto @RdoubleA's suggestion.. in this case you already did most of the important work by finding and fixing the bug with a reasonable repro. We should definitely have a unit test for the Phi3 tokenizer, and it should ideally test this behavior. But given no such unit test currently exists, I don't think it should be on you to add it in this PR. (If you want to add one separately we would of course love to have it, but we don't need to block this fix on it, and otherwise we can and should add it ourselves.)

joecummings (approved):

hmosousa:
Thanks @RdoubleA and @ebsmothers. I might be able to add the tests. Will create a new issue for that.

PR #1069, opened 2024-06-07
update checkpoint paths in llama3 tutorial

ebsmothers:
The checkpoint files provided in the Llama3 tutorial generation, eval, and quantization configs are incorrect. They currently point to `consolidated.00.pth` which is the raw checkpoint prior to any fine-tuning. Instead they should point to the filename after fine-tuning, which is `meta_model_0.pt`.


RdoubleA (approved):
Good catch, thanks for the fix. We need a way to continuously test our tutorials so they don't get stale. Not that this was the issue here, but something to consider...

PR #1070, opened 2024-06-07
Profiler

jeromeku:
## [FEATURE] Enable more configurable profiler

Add more configuration options for `torch.profiler.profile` in `config` files.

Currently `torchtune.utils.profiler` has hardcoded options.  

This PR adds the ability to fully configure the profiler by expanding the accepted args under the `profiler` key in `config` files.

Features:
- `torch.profiler.profile` settings such as `with_stack`, `profile_memory`, etc. can be specified under `profiler.profile`
- Custom profiling [schedule](https://pytorch.org/docs/stable/profiler.html#torch.profiler.schedule) under `profiler.schedule`
- Sensible defaults: the profiler is instantiated with sensible defaults such that user does not need to specify all sections
  - E.g., if only `profiler.enabled=true` but no other keys are included, will default to the current `torchtune.utils.profiler` 
- Works in both single device and distributed settings 
- API is identical to that of current `torchtune` profiler such that the only change required is to use the newly introduced `setup_torch_profiler` to instantiate the profiler within the `recipe`.
  
## Usage

Changes needed to profile a `recipe`:
- Add the a `profiler` section to `config` yaml (see below)
- Call `torchtune.utils._profiler.setup_torch_profiler` within the `setup` method of the `recipe` with the top-level `args`
- Wrap the training loop with the profiler and `step` the profiler after each `optimizer` step (per the current profiler `recipe`).

### Profiler Config
```
profiler:
    enabled: bool

    #Output directory of trace artifacts
    output_dir: str

    #`torch.profiler.ProfilerActivity` types to trace
    CPU: bool
    CUDA: bool

    #`torch.profiler.profile` options
    profile:
        # _component_ is optional as the setup method will handle
        _component_: torch.profiler.profile
        profile_memory: bool
        with_stack: bool
        record_shapes: bool
        with_flops: bool
    
    #`torch.profiler.schedule` options
    schedule:
        # _component_ is optional as the setup method will handle
        _component_: torch.profiler.schedule
        wait: int
        warmup: int
        active: int
        repeat: int
```
**Notes**
- All args are optional and if not specified, will default to minimal working set per `torch.profiler.profile` defaults (see the added `_DEFAULTS` in `torchtune.utils._profiler`)
- Certain options are overridden to match requirements of the torch profiler
  - E.g., `profile_memory` requires `with_stack` and `record_shapes` so if user sets `profile_memory: true`, then the latter two will automatically be set within the setup method
- Top-level `cfg` will be modified in-place by `setup_torch_profiler` to reflect updated `profiler` config for logging / debugging purposes

## Examples
- Run single device, same as current `torchtune.utils.profiler`:
    ```
    tune run lora_finetune_single_device \
    --config llama2/7B_lora_single_device \
    profiler.enabled=True
    ```
- Same as above but with profiler disabled:
    ```
    tune run lora_finetune_single_device \
    --config llama2/7B_lora_single_device \
    profiler.enabled=False
    ```

- Run single device, profiling on a 10-step cycle:  
    ```
    tune run lora_finetune_single_device \
    --config llama2/7B_lora_single_device \
    profiler.enabled=True \
    profiler.output_dir=./profiler_outputs \
    profiler.schedule.wait=5 \
    profiler.schedule.warmup=4 \
    profiler.schedule.active=1 \
    profiler.schedule.repeat=0
    ```
    At the end of each cycle, a `tensorboard` / `chrome` trace along with key averages table of operators ordered by `cuda` time will be dumped to `profiler_outputs`.

- Run distributed, profiling for single cycle:
    ```
    tune run --nnodes=1 --nproc_per_node=2 lora_finetune_distributed \
    --config llama2/7B_lora \
    profiler.enabled=True \
    profiler.schedule.wait=10 \
    profiler.schedule.warmup=10 \
    profiler.schedule.active=1 \
    profiler.schedule.repeat=1
    ```
    - In addition to the previous outputs, additional memory usage info will be recorded and a viewable [memory timeline](https://pytorch.org/docs/stable/profiler.html#torch.profiler._KinetoProfile.export_memory_timeline) will be exported as well.
    - Each trace will be marked with the `rank` of the process.
    - Note that profiling memory adds significant overhead (training time and large trace files) so the profiling schedule should be set accordingly. 

## Tests

See `tests/test_profiler.py` for unit testing of profiler instantiation with `torchtune.config` as well as the expected functionality of `torchtune.utils.setup_torch_profiler`.

## Next Steps

This PR edits `recipes/lora_finetune_single_device.py` and `recipes/dev/lora_finetune_distributed.py` and their corresponding configs to demonstrate usage.  

If there is interest, can additionally update other registered recipes and configs to reflect the expanded profiler API.


ebsmothers:
Hi @jeromeku thanks for opening this PR! The profiler experience is something we've been meaning to improve for a while so really appreciate you taking it on. A couple general comments on the `setup_torch_profiler` API: you probably noticed already but we don't really do nested instantiation of configs (hence why you have to hack around it a bit with `OmegaConf.create`).  We also tend not to pass `cfg` objects to utilities or other core library code (maybe we break that rule in a couple places, but we try our best not to). Do you think there's a world where you define a utility method in the recipe to unpack the config, then pass only primitive datatypes to `setup_torch_profiler` (basically all the leaf nodes in the `profiler` config) and build everything in there (e.g. just call `torch.profiler.schedule` directly without config-ifying it)? Or is that too cumbersome? We're trying very hard to keep things configurable without giving the config too much power, but would be interested to know if you find this to be something that slows down experimentation for you.

I will take a more thorough pass soon but wanted to hear your thoughts on this first.

jeromeku:
@ebsmothers 

Thanks for the response!

Regarding your questions, I'd initially considered just adding unpacking functions within the recipe per your suggestion.  However, I decided to abstract away the processing by passing the entire top-level `cfg` to a separate library function.  This reduces code duplication, as now all a recipe user / writer needs to do is to call the `setup_torch_profiler` function to handle entire the profiler setup.

The intent is to make it easy for users to add / tweak profiler configs with a few args, as the `setup` function implements a number of checks and defaults to ensure the profiler is properly set up regardless of how fully specified the `profiler` section is (see `tests/test_profiler` for validation of these conditions).

That said, happy to change to align more closely with the design philosophy of `torchtune`.

ebsmothers:
@jeromeku thanks for the response and sorry for the delay in getting back to you here. Regarding torchtune's philosophy, we are OK with a bit of code duplication (e.g. you can see we have a fair bit across our model and component builders under `torchtune/models`, we also have the `tune cp` CLI command to directly copy a recipe or config to a local path for modification). 

I think the configurability you've added here is really nice so don't want to sacrifice on that. I would propose the following division: first define something like a `_setup_profiler` method inside the recipe. This will take the config and parse it into fields that can then be passed directly to `setup_torch_profiler`. `setup_torch_profiler` can then continue to handle all the validation you described (and of course the actual setting up of the profiler). I can imagine a signature roughly like this:

```
def setup_torch_profiler(
	enabled: bool,
	cpu: bool,
	cuda: bool,
	profile_memory: bool,
	with_stack: bool,
	record_shapes: bool,
	with_flops: bool,
	wait: int,
	warmup: int,
	active: int,
	repeat: int,
) -> torch.profiler.profile:
```

(with appropriate defaults as you've already defined inside the existing function). One nice thing about this is that it can then be imported in a standalone fashion outside of our recipes without users having to figure out which config fields are required.

Btw the testing you added is quite comprehensive, really appreciate you doing that.

jeromeku:
@ebsmothers 

Thanks for the feedback.  Will refactor per the comments!

RdoubleA:
Really appreciate the thoroughness of this PR and the code and the exhaustiveness of the testing. Your docstrings are very descriptive, awesome to see. Have a few overall comments:
- Can we move `_setup_profiler` and all the config parsing to be within the recipe? One principle we maintain is that DictConfigs should be used only the recipe layer (and in `torchtune/config`). Once we get to the general `setup_profiler` utility all the params should be flattened and there should be no DictConfigs. It is totally ok if `_setup_profiler` is redundant across recipes
- The defaults are awesome, I would encourage you to remove any config related items from these (such as component fields) for the principle mentioned above.
- It's not clear to me why we need to create and return the profiler config again in `setup_profiler`?
- In `_setup_profiler`, just pass in `cfg["profiler"]` directly and name the parameter `cfg_profiler` similar to the other setup methods. That way you don't have to keep indexing with `PROFILER_KEY`

jeromeku:
Closing this and issuing a new [PR](https://github.com/pytorch/torchtune/pull/1089) as the commit history was corrupted after rebasing on main.

PR #1075, opened 2024-06-08
utils doc updates

ebsmothers:
#### Context
- [ ] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

#### Changelog
Updating a bunch of docstrings under `torchtune/utils` and exposing all public APIs in our live docs (for better or for worse).
Oh also while I was at it I got rid of a couple useless APIs (I wanted to remove more..)

#### Test plan
Inspected docs page for every public API under `torchtune/utils`. Since I updated some resume logic for recipes that don't have tests for resuming from checkpoint :( I will run them manually when I get a chance.

### DIFF
--- a/recipes/full_finetune_distributed.py
+++ b/recipes/full_finetune_distributed.py
@@ -367,8 +367,8 @@ class FullFinetuneRecipeDistributed(FTRecipeInterface):
         optimizer = config.instantiate(cfg_optimizer, self._model.parameters())
 
         if opt_state_dict:
-            opt_state_dict = utils.transform_opt_state_dict(
-                opt_state_dict, self._model, optimizer
+            opt_state_dict = FSDP.optim_state_dict_to_load(
+                self._model, optimizer, opt_state_dict
             )
             optimizer.load_state_dict(opt_state_dict)
 
--- a/tests/torchtune/utils/test_precision.py
+++ b/tests/torchtune/utils/test_precision.py
@@ -30,7 +30,7 @@ class TestPrecisionUtils:
         """
         Tests that the correct dtype is returned based on the input string.
         """
-        dtypes = [None, torch.half] + PRECISION_STR_TO_DTYPE.keys()
+        dtypes = [None, torch.half] + list(PRECISION_STR_TO_DTYPE.keys())
         expected_dtypes = [
             torch.float32,
             torch.float16,
--- a/tests/torchtune/utils/test_precision.py
+++ b/tests/torchtune/utils/test_precision.py
@@ -15,7 +15,7 @@ import torch
 from torchtune.utils.precision import (
     _set_float32_precision,
     get_dtype,
-    list_dtypes,
+    PRECISION_STR_TO_DTYPE,
     set_default_dtype,
     validate_expected_param_dtype,
     verify_bf16_support,
@@ -30,7 +30,7 @@ class TestPrecisionUtils:
         """
         Tests that the correct dtype is returned based on the input string.
         """
-        dtypes = [None, torch.half] + list_dtypes()
+        dtypes = [None, torch.half] + PRECISION_STR_TO_DTYPE.keys()
         expected_dtypes = [
             torch.float32,
             torch.float16,
@@ -55,9 +55,6 @@ class TestPrecisionUtils:
             RuntimeError, match="bf16 precision was requested but not available"
         ):
             get_dtype(torch.bfloat16)
-
-    def test_list_dtyes(self):
-        assert set(list_dtypes()) == {"fp16", "bf16", "fp32", "fp64"}
 
     @pytest.mark.skipif(not cuda_available, reason="The test requires GPUs to run.")
     def test_set_float32_precision(self) -> None:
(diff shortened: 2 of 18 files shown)


RdoubleA (approved):
Mainly minor comments, but thanks for vastly improving utils documentation!

recipes/full_finetune_distributed.py
```diff
@@ -367,8 +367,8 @@ def _setup_optimizer(
         optimizer = config.instantiate(cfg_optimizer, self._model.parameters())
 
         if opt_state_dict:
-            opt_state_dict = utils.transform_opt_state_dict(
-                opt_state_dict, self._model, optimizer
+            opt_state_dict = FSDP.optim_state_dict_to_load(
```
joecummings, line 370: ?

ebsmothers: This is needless indirection imo. Aren't we always using FSDP in our distributed recipes?

torchtune/utils/_checkpointing/_checkpointer_utils.py
```diff
@@ -57,7 +75,20 @@ def safe_torch_load(
     checkpoint_path: Path, weights_only: bool = True, mmap: bool = True
 ) -> Dict[str, Any]:
     """
-    Utility to load a checkpoint file in a safe manner.
+    Utility to load a checkpoint file onto CPU in a safe manner. Provides separate handling for
+    safetensors files.
+
+    Args:
+        checkpoint_path (Path): Path to the checkpoint file.
+        weights_only (bool): Whether to load only tensors, primitive types, and dictionaries
```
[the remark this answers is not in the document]
ebsmothers: When true, we will only load tensors, primitive types, and dictionaries. When false, we will load everything. I can make this clearer. 

But really both this arg and mmap are just passthroughs to torch.load. Honestly at this point this utility is pretty much just an if/else to load with safetensors vs torch.load API, I have half a mind to scrap it altogether (or at least be explicit in this and just provide e.g. `**torch_load_kwargs` or something).

tests/torchtune/utils/test_precision.py
```diff
@@ -30,7 +30,7 @@ def test_get_dtype(self):
         """
         Tests that the correct dtype is returned based on the input string.
         """
-        dtypes = [None, torch.half] + list_dtypes()
+        dtypes = [None, torch.half] + list(PRECISION_STR_TO_DTYPE.keys())
```
pbontrager, line 33: Global variables like PRECISION_STR_TO_DTYPES should not be public, which is the reason for a function like list_dtypes. This also allows us to add any logic on top of the list if we need to in the future without breaking backward compatibility.

recipes/full_finetune_distributed.py
pbontrager: What about FSDP2, does this still apply?

torchtune/utils/_checkpointing/_checkpointer_utils.py
```diff
@@ -57,7 +75,20 @@ def safe_torch_load(
...
+            (passthrough to torch.load). Default: True
+        mmap (bool): Whether to mmap from disk into CPU memory. Default: True
```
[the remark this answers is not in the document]
ebsmothers: This is also a torch.load passthrough, which I think is just using Linux [mmap](https://man7.org/linux/man-pages/man2/mmap.2.html) under the hood. Actually will just point to [this tutorial section](https://pytorch.org/tutorials/recipes/recipes/module_load_state_dict_tips.html#using-torch-load-mmap-true) which describes it in detail.

In general I think we want this to be true most of the time, but @calvinpelletier pointed out that when we modify and overwrite the same file during the training process we actually need to set `mmap=False`. Actually though I just noticed [this PR](https://github.com/pytorch/pytorch/pull/124889) from core, maybe we can consider `set_default_mmap_options(MAP_SHARED)` for this case instead? cc @mikaylagawarecki on if this is the correct understanding here

recipes/full_finetune_distributed.py
ebsmothers: No there is other mapping that has to occur there. But I am explicitly not aiming to tackle that here since it is still a dev recipe

tests/torchtune/utils/test_precision.py
ebsmothers: So (a) we are not really using `PRECISION_STR_TO_DTYPE` publicly or exposing it as a public API in any official capacity, right? We are already using it in `get_dtype` anyways, I just added one more usage to remove the `list_dtypes` API. And (b) it's similar to my comment about `transform_opt_state_dict` -- we are doing extra indirection for no real reason here. Also not sure I follow the BC-breaking argument.. like no matter what this is just a global mapping that we update with new(/deprecated) dtypes, we're not gonna change `list_dtypes` and so we break BC in the same way no matter what (unless I'm missing something obvious here)

pbontrager: To the BC comment, get_list is essentially a getter to the mapping. I believe it's best practices to use a getter instead of directly exposing the mapping because you'd run into problems if you ever needed to add any dynamic logic (like filtering on hardware, etc). Either way this is a nit, so it's up to you.

recipes/full_finetune_distributed.py
pbontrager: As long as you don't believe we'll need to add this solution back soon if FSDP2 use expands a bunch.

PR #1076, opened 2024-06-09
Support Early Exit Loss and/or Layer Dropout

mostafaelhoushi:
#### Context
- [X] add a new feature

Implements early exit loss and/pr layer dropout to reproduce experiments in various papers like [LayerSkip](https://arxiv.org/abs/2404.16710), [LITE](https://arxiv.org/abs/2310.18581), [LayerDrop](https://arxiv.org/abs/1909.11556), and [Progressive Layer Dropping](https://arxiv.org/abs/2010.13369).

#### Usage
Download Llama2 7B model:
```
tune download meta-llama/Llama-2-7b-hf --output-dir /tmp/Llama-2-7b-hf --hf-token <HF_TOKEN>
```

To reproduce experiments of various papers that use early exit loss and/or layer dropout:
- [LayerSkip](https://arxiv.org/abs/2404.16710) on TOPv2:
```
tune run --nnodes 1 --nproc_per_node 4 dev/early_exit_finetune_distributed --config recipes/dev/7B_full_early_exit.yaml early_exit_loss.scale=1.0 early_exit_loss.curriculum=torchtune.modules.early_exit_loss.GradualEarlyExitCurriculum early_exit_loss.scale_fn=torchtune.modules.early_exit_loss.linear_l_loss_scale layer_dropout.prob=0.2 layer_dropout.scale=exp
```
- [LITE](https://arxiv.org/abs/2310.18581):
```
tune run --nnodes 1 --nproc_per_node 4 dev/early_exit_finetune_distributed --config recipes/dev/7B_full_early_exit.yaml layer_dropout=null early_exit_loss.layers=8,12,16,20,24,28 early_exit_loss.scale_fn=torchtune.modules.early_exit_loss.uniform_loss_scale early_exit_loss.curriculum=null epochs=5
```

- [LayerDrop](https://arxiv.org/abs/1909.11556):
```
tune run --nnodes 1 --nproc_per_node 4 dev/early_exit_finetune_distributed --config recipes/dev/7B_full_early_exit.yaml early_exit_loss=null layer_dropout.prob=0.2 layer_dropout.layers=1::2
```

- [Progressive Layer Dropping](https://arxiv.org/abs/2010.13369) (The paper also implements a curriculum for layer drop probability which is not yet implemented.):
```
tune run --nnodes 1 --nproc_per_node 4 dev/early_exit_finetune_distributed --config recipes/dev/7B_full_early_exit.yaml early_exit_loss=null layer_dropout.prob=0.5 layer_dropout.scale=exp
```

#### Changelog
- Modified `TransformerDecoder` by refactoring the operations that happen after running all transformer layers that transform last hidden state to a probability vector across all tokens into a separate `unembed()` function. This was needed to re-use `unembed()` to obtain predictions at intermediate layers, and hence calculate losses at intermediate layers.
- Added to `torchtune/modules/common_utils.py` a `slice_str_to_array()` function that can convert numpy-like indexing string into a boolean array.
- Added `torchtune/modules/early_exit_loss.py` to implement early exit loss
- Added `torchtune/modules/layer_dropiut.py` to implement layer dropout
- Added separate recipe script and configuration file
- Added unit test cases

#### Test plan
1. Download Llama 2 7B:
```
tune download meta-llama/Llama-2-7b-hf --output-dir /tmp/Llama-2-7b-hf 
```

2. Finetune on TOPv2 instruction set (a small dataset that we can quickly finetune with early exit loss). W
```
tune run --nnodes 1 --nproc_per_node 8 dev/early_exit_finetune_distributed --config recipes/dev/7B_full_early_exit.yaml 
```
We will run different variants of this command in the Table below to show different speed ups.

3. Convert trained checkpoint to HF naming:
```
mv /tmp/Llama-2-7b-hf/hf_model_0001_0.pt /tmp/Llama-2-7b-hf/pytorch_model-00001-of-00002.bin
mv /tmp/Llama-2-7b-hf/hf_model_0002_0.pt /tmp/Llama-2-7b-hf/pytorch_model-00002-of-00002.bin
```

4. Install HF `transformers` nightly (to enable early exit speculative decoding as described [here](https://huggingface.co/blog/layerskip)):
```
pip install git+https://github.com/huggingface/transformers.git
pip install accelerate
```

5. Save this [code snippet](https://gist.github.com/mostafaelhoushi/bda38661da1738e3a01a07ccf69bde7d) to a file named `eval_hf.py`

6. Run `eval_hf.py` that benchmarks autoregressive decoding and early exit self-speculative decoding:
```
python eval_hf.py
```

You should get an output that looks like this:
```
[INST]Set alarm for 6am every day[\INST]Set alarm for 6 am every day [IN:CREATE_ALARMSet alarm for 6 am every day ]
Orig Time: 1.159
[INST]Set alarm for 6am every day[\INST]Set alarm for 6 am every day [IN:CREATE_ALARMSet alarm for 6 am every day ]
Layerskip Time: 0.500
```
Hence, LayerSkip's training recipe has enabled LayerSkip's (early exit self-speculative decoding) inference that leads to a speedup of 1.159/0.500 =  2.318x.

We ran different variants of training and obtained the results below. Please note that differences in generation time for different training configurations may be due to different number of tokens generated. Hence, to measure speedup, we need to look at the different generation times for the same training configuration:
[Link to Results Spreadsheets](https://docs.google.com/spreadsheets/d/1RrJFKGEeBgndoDKhH6DBl5SouFT-vIaMmvWaYbDc5yk/edit?usp=sharing)

#### Checklist

#### Future Work

In the future we can:
- **Implement early exit self-speculative decoding in `torchtune/generation/_generation.py`**: `_generation.py` uses similar code to `gpt-fast`. We have already implemented early-exit self-speculative decoding on gpt-fast [here](https://github.com/pytorch-labs/gpt-fast/compare/main...LayerSkip) so it should be straightforward to port to torchtune.
- **Enable users to add their own heads or adapter modules at early exits** rather than using a shared head for all early exits (this can enable implementing a paper like [Kangaroo](https://arxiv.org/abs/2404.18911)).
- **Make `early_exit_loss()` function support generic loss preprocessing**: different training scripts may perform different pre-processing of logits and labels. We need to refactor training scripts slightly to encapsulate such preprocessing into a function to invoke when calculating early exit loss.
- **Make `early_exit_loss()` function support generic loss functions**: support other types of losses like KL divergence loss, or losses represented as lists (as is the case with model's `num_output_chunks`>0).


torchtune/modules/transformer.py
```diff
@@ -240,9 +254,11 @@ def forward(
             # in most cases input_pos_len should be 1
             mask = self.causal_mask[None, input_pos]
 
-        for layer in self.layers:
+        for i, layer in enumerate(self.layers):
             # shape: [b, s, d]
-            h = layer(h, mask=mask, input_pos=input_pos)
+            h = self.layer_dropouts[i](layer, h, mask=mask, input_pos=input_pos)
```
ebsmothers, line 259: I'm wondering if there's a world where we can actually do all of this without even touching `TransformerDecoder`'s forward (except for the change to support `output_hidden_states`). Suppose we generalize the init arg `layer` in `TransformerDecoder` to take in a list of `nn.Module`s. Then we provide functions to build a list of layers and call those whenever we're initializing this class instead. So in the existing case it'd look like

```
def get_decoder_layers(
	attn: CausalSelfAttention,
	mlp: nn.Module,
	sa_norm: nn.Module,
	mlp_norm: nn.Module,
	n_layers: int
) -> List[TransformerDecoderLayer]:
	return [TransformerDecoderLayer(attn, mlp, sa_norm, mlp_norm) for _ in n_layers]
```

And for the `LayerDropout` version:

```
def get_layer_skip_decoder_layers(
	attn: CausalSelfAttention,
	mlp: nn.Module,
	sa_norm: nn.Module,
	mlp_norm: nn.Module,
	n_layers: int,
        layer_dropout_prob: float = 0.5,
        layer_dropout_prob_layer_scale: str = "exp",
        layer_dropout_str: str = ":",
) -> List[LayerDropout]:
	layers = [TransformerDecoderLayer(attn, mlp, sa_norm, mlp_norm) for _ in n_layers]
	return create_layer_dropout_modules(
		layers, 
		layer_dropout_prob, 
		layer_dropout_prob_layer_scale,
		layer_dropout_str
	)
```

I am making one big assumption here: that we can move the callable `function` from the `LayerDropout` `forward` into its `__init__`. Then a given `LayerDropout` module would correspond to a single `nn.Module` it is wrapping. I guess this is less flexible from the `LayerDropout` perspective, but I think it should help from a composability perspective. (I am also implicitly changing the signature of `create_layer_dropout_modules` to account for this change, replacing `n_layers` -> `layers`.) Let me know if this assumption makes sense (it's possible there are important details of the implementation I'm overlooking with this proposal).

If this does work, we would expose things in our model builders by replacing e.g. [this](https://github.com/pytorch/torchtune/blob/74fb5e4cceb6e64473608106b77f9f9308ec11c3/torchtune/models/llama3/_component_builders.py#L100-L105) with calls to the appropriate builder. This is similar to what we do for some of our LoRA components (see e.g. [here](https://github.com/pytorch/torchtune/blob/74fb5e4cceb6e64473608106b77f9f9308ec11c3/torchtune/models/llama3/_component_builders.py#L211-L221)).

Also cc @pbontrager since this is related to ideas we've been talking about for cross-attention support.

torchtune/modules/layer_dropout.py
```diff
@@ -0,0 +1,85 @@
...
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+from enum import Enum
+from typing import Callable, Optional
+import math
+import torch
+
+from .common_utils import slice_str_to_array
```
ebsmothers, line 12: nit: prefer absolute imports

torchtune/modules/transformer.py
```diff
@@ -150,6 +157,9 @@ def __init__(
         self.head_dim = head_dim
         self.causal_mask = None
 
+        self.layer_dropouts = create_layer_dropout_modules(num_layers, layer_dropout_prob, layer_dropout_prob_layer_scale, layer_dropout_str)
+        self.output_hidden_states = OrderedDict() # TODO: use tensordict?
```
ebsmothers, line 161: Just curious: why do we need a dict at all? And separately, why is this defined as an instance variable instead of something local to forward? Is it because you don't want to change the return signature? (If so I think we should just change it since returning hidden states is a fairly common case anyways)

recipes/full_finetune_distributed.py
```diff
@@ -516,14 +516,28 @@ def train(self) -> None:
...
+                    # logits_early = self._model.output(self._model.norm(torch.stack(tuple(self._model.output_hidden_states.values()))))
+                    for layer_id, hidden_state in self._model.output_hidden_states.items():
+                        h_early = self._model.norm(hidden_state)
+                        logits_early = self._model.output(h_early)
+                        # Shift so that tokens < n predict n
+                        logits_early = logits_early[..., :-1, :].contiguous()
+                        logits_early = logits_early.transpose(1, 2)
+                        # Compute early loss
+                        loss_early = self._loss_fn(logits_early, labels)
+                        loss += 0.1 / len(self._model.layers) * loss_early
```
ebsmothers, line 539: Is there a world where we can/should define this as a standalone loss function? I guess we would still need to apply norm and output to each hidden state outside the loss, but wonder if it'd help with the 0.1 magic number.

Also nit: I don't think layer_id is actually used here?

torchtune/modules/layer_dropout.py
```diff
@@ -0,0 +1,85 @@
...
+        ScaleType.EXP: math.exp(val * math.log(2) / scale_period) - 1,
+        ScaleType.LINEAR: val / scale_period,
+        ScaleType.LOG: math.log(val + 1) / math.log(scale_period + 1),
+        ScaleType.SIN: math.sin(0.5 * math.pi * val / scale_period),
+        ScaleType.SIGMOID: 1 / (1 + math.exp(-10 * (val / scale_period - 0.5))),
+    }[scale_type]
+
+def create_layer_dropout_modules(num_layers: int, prob_max: float= 0.0, prob_layer_scale: ScaleType = ScaleType.EXP, layers_str: Optional[str] = None, disable_on_eval: bool = True):
+    layer_dropouts = torch.nn.ModuleList()
+    has_dropout = slice_str_to_array(layers_str, num_layers) if layers_str else [True] * num_layers
```
ebsmothers, line 72: Can you share an example usage of `layers_str`? My assumption is that it's to filter which layers have dropout applied, but looking at `slice_str_to_array` I don't actually understand what format I should pass these in to accomplish that.

recipes/full_finetune_distributed.py
```diff
@@ -516,14 +516,28 @@ def train(self) -> None:
...
                 logits = logits[..., :-1, :].contiguous()
                 labels = labels[..., 1:].contiguous()
                 logits = logits.transpose(1, 2)
                 # Compute loss
                 loss = self._loss_fn(logits, labels)
 
+                # Compute early exit loss
+                if self._model.output_hidden_states:
+                    # TODO: calculate early_logits in one shot:
+                    # logits_early = self._model.output(self._model.norm(torch.stack(tuple(self._model.output_hidden_states.values()))))
```
ebsmothers, line 530: Is there something preventing this currently?

torchtune/modules/layer_dropout.py
```diff
@@ -0,0 +1,85 @@
...
+    def forward(self, function: Callable, input: torch.Tensor, *args, **kwargs):
+        n = input.shape[self.dim]
+
+        if self.prob == 0 or (self.disable_on_eval and self.training is False):
+            self.inferred = 1.0
+            return function(input, *args, **kwargs)
+
+        skip = torch.bernoulli(torch.Tensor((n) * [self.prob]), generator=self.generator).to(input.device).to(input.dtype)
+        self.inferred = 1 - torch.mean(skip)
+        ind_selected = (skip == 0).nonzero().squeeze().to(input.device)
```
ebsmothers, line 35: Won't this cause a sync? (I thought you already moved to input device in L33)

mostafaelhoushi:
> Also, is the idea to eventually support the self-speculative decoding technique as well?

I am not sure if you would like to support inference optimization techniques in torchtune, but we have implemented self-speculative decoding in gpt-fast here: https://github.com/pytorch-labs/gpt-fast/compare/main...LayerSkip.

torchtune/modules/transformer.py
mostafaelhoushi: I like this idea, and I will try to implement it.
One concern I have is if `LayerDropout` wraps `TransformerDecoderLayer`, then this means the weights of the model will look something like `layer_dropout0.layer.attn.wk`, right? Then this will cause problems if we want to load to or from a model that doesn't have dropout as its weights' names will be something like `layer0.atten.wk`.

recipes/full_finetune_distributed.py
mostafaelhoushi: Just implemented calculating early logits and losses in one shot without a for loop in this commit: 735d2a8023531a4b306dc59e150e3f20d56d0d9c

torchtune/modules/layer_dropout.py
mostafaelhoushi: Fixed here: be912a6f871af9efd2271b054a7ca43bde98efb9

torchtune/modules/layer_dropout.py
mostafaelhoushi: Thanks for noting this. I removed it in this commit: 0686dd28d602fb50bf57cf117f88314d49f58af2

mostafaelhoushi:
> Hi @mostafaelhoushi @ebsmothers @AkshatSh this feature is very interesting I see 2 branches with the name layer skip in the repository: SkipLayer ak/layer_skip Do they both aim the same goal of skipping layers or early-exit at some layer? Can you please elaborate more about the motivation behind this feature and the design of the solution
> 
> Thanks, Eshcar

I want to double check, I can't find those 2 branches in this torchtune repo. Are you referring to torchtune repo or another repo?

recipes/full_finetune_distributed.py
mostafaelhoushi: Moved to a standalone function here: 4e4783f12cb7a52fdca540097ed4138ca25ea6cd

torchtune/modules/transformer.py
mostafaelhoushi: I made hidden_states local to forward() and returned in this commit:
ccb4a5006ece5436505d8f7b1764734aa347d422

One disadvantage with returning hidden states is that it requires adding `if..else` conditions in the return statement and in the training script. If you prefer that, should I create a dedicated training script instead? I was hoping though that early exit and layer dropout become orthogonal to all of the existing training scripts: single node, multi node, lora, etc.

mostafaelhoushi: > why do we need a dict at all?

It is for a couple of reasons:
- we want different layers to have different loss scales (I implemented that in a separate function)
- I want to give the user flexibility which layers to return hidden state (in an upcoming commit or PR I might even have different iterations have different layers outputting hidden states)

Eshcar:
> > Hi @mostafaelhoushi @ebsmothers @AkshatSh this feature is very interesting I see 2 branches with the name layer skip in the repository: SkipLayer ak/layer_skip Do they both aim the same goal of skipping layers or early-exit at some layer? Can you please elaborate more about the motivation behind this feature and the design of the solution
> > Thanks, Eshcar
> 
> I want to double check, I can't find those 2 branches in this torchtune repo. Are you referring to torchtune repo or another repo?

apologies, my mistake it is from a different repository: pytorch-labs
https://github.com/pytorch-labs/gpt-fast/tree/LayerSkip by  @mostafaelhoushi
https://github.com/pytorch-labs/gpt-fast/tree/ak/layer_skip by @AkshatSh

is there a connection between the 2 branches and the current issue?
what is the motivation for skipping layers in the middle of a token generation?
If there is some paper about this I would be happy to read it

mostafaelhoushi:
> 
> apologies, my mistake it is from a different repository: pytorch-labs https://github.com/pytorch-labs/gpt-fast/tree/LayerSkip by @mostafaelhoushi https://github.com/pytorch-labs/gpt-fast/tree/ak/layer_skip by @AkshatSh
> 
> is there a connection between the 2 branches and the current issue? what is the motivation for skipping layers in the middle of a token generation? If there is some paper about this I would be happy to read it

https://github.com/pytorch-labs/gpt-fast/tree/LayerSkip is just a slightly cleaned up fork of https://github.com/pytorch-labs/gpt-fast/tree/ak/layer_skip
They implement the self-speculative decoding inference described in this paper: https://arxiv.org/abs/2404.16710

This PR we are on in torchtune is to implement the training recipe of the same paper.

torchtune/utils/early_exit.py
```diff
@@ -0,0 +1,141 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
```
ebsmothers, line 1: Let's move this to torchtune/modules/early_exit.py

recipes/full_finetune_distributed.py
```diff
@@ -30,6 +30,8 @@
 from torchtune.datasets import ConcatDataset
 from torchtune.recipe_interfaces import FTRecipeInterface
 from torchtune.utils.activations import apply_selective_activation_checkpointing
+from torchtune.utils.early_exit import early_exit_loss, build_early_exit_curriculum
```
ebsmothers, line 33: Let's create a separate recipe for this under recipes/dev

mostafaelhoushi:
Based on discussions with @ebsmothers and @RdoubleA :
- Rebase to get intermediate output functionality
- Move functionalities (e.g., early exit losses, curricula, etc.) to torchtune/modules/early_exit.py 
- Create a recipe under recipes/dev
- Layer Dropout: module replacement function. Could be in torchtune or torchao

torchtune/modules/layer_dropout.py
mostafaelhoushi: I have just updated the PR with docstring and unit test cases providing different example usages.

torchtune/modules/transformer.py
```diff
@@ -188,6 +198,7 @@ def forward(
         *,
         mask: Optional[Tensor] = None,
         input_pos: Optional[Tensor] = None,
+        output_hidden_states: Union[bool, List[bool]] = False,
```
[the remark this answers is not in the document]
mostafaelhoushi: Resolving this as `output_hidden_states` got supported in another PR

torchtune/modules/transformer.py
mostafaelhoushi: LayerDropout has been re-implemented in the PR by creating a wrapper module. Please take a look.

torchtune/modules/transformer.py
mostafaelhoushi: Resolving as `output_hidden_states` got supported in another PR.

recipes/full_finetune_distributed.py
mostafaelhoushi: Done

torchtune/modules/early_exit_loss.py
```diff
@@ -0,0 +1,293 @@
...
+    def get(self) -> np.ndarray:
+        """
+        Get the current output hidden states.
+        Returns:
+            np.ndarray: A list indicating whether we should calculate loss for each layer.
+        """
+        do_output_hidden_states = np.copy(self._do_output_hidden_states)
+        # Ensure last layer is trained
+        if self.train_last_layer:
+            do_output_hidden_states[-1] = True
```
ebsmothers, line 188: Out of curiosity, why do we provide special handling for this case? I also wonder whether we could just boil this class down to a protocol similar to [this](https://github.com/pytorch/torchtune/blob/b1aecb19ef5eee949c822e33ef4263807fb6f63e/torchtune/modules/peft/_utils.py#L17), but will defer to you on what you feel is the right way to expose things here.

torchtune/modules/early_exit_loss.py
```diff
@@ -0,0 +1,293 @@
...
+    Args:
+        *args: Positional arguments passed to the parent EarlyExitCurriculum class.
+        percent_scale (float, optional): A scaling factor to determine at which percentage
+            of steps, all the layers will be enabled. At `steps = max_steps / percent_scale`, all the layers will be enabled.
+        **kwargs: Keyword arguments passed to the parent EarlyExitCurriculum class.
+    """
+
+    def __init__(
+        self,
+        *args,
```
ebsmothers, line 228: Is the usage of args just for saving a few lines and not repeating the same args from the parent class? In this case I think mixing child args in the model makes it more confusing and not necessarily worth the lines saved

torchtune/modules/early_exit_loss.py
```diff
@@ -0,0 +1,293 @@
...
+        percent_scale: float = 2,
```
ebsmothers, line 229: nit but if we're calling it percent_scale I'd expect this to be 50.0 based on the docstring

ebsmothers (commented):
Just a partial review for now, will come back a bit later for the rest

recipes/dev/7B_full_early_exit.yaml
```diff
@@ -0,0 +1,140 @@
...
+# Tokenizer
+tokenizer:
+  _component_: torchtune.models.llama2.llama2_tokenizer
+  path: /tmp/Llama-2-7b-hf/tokenizer.model
+  max_seq_len: null
+
+# Dataset
+# dataset:
+#   _component_: torchtune.datasets.alpaca_dataset
+#   packed: False  # True increases speed
```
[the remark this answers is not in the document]
mostafaelhoushi: Done in b03cb574168534f631f1ec57655252496f5a3ae5

recipes/dev/early_exit_finetune_distributed.py
```diff
@@ -0,0 +1,1077 @@
...
+from torchtune.training import DummyProfiler, PROFILER_KEY
+from torchtune.training.activations import apply_selective_activation_checkpointing
+from torchtune.training.lr_schedulers import get_lr
+
+from tqdm import tqdm
+
+log = utils.get_logger("DEBUG")
+
+# TODO: add explanation of EE and LD and cite papers
+# TODO: add to .yaml file full test commands and different examples, citing commands to implement different papers.
```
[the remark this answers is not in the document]
mostafaelhoushi: Done in b03cb574168534f631f1ec57655252496f5a3ae5

recipes/dev/early_exit_finetune_distributed.py
```diff
@@ -0,0 +1,1077 @@
...
+            )
+
+        if (
+            cfg.get("fsdp_cpu_offload", False)
+            and cfg.optimizer.get("fused", False)
+            and not utils.torch_version_ge("2.4.0")
+        ):
+            raise RuntimeError(
+                "Using fused optimizer on CPU is only supported in PyTorch nightly."
+            )
```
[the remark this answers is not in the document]
mostafaelhoushi: Done in 199b8dd584e6b96a330b0fcb236d84bbfb5075b1

recipes/dev/7B_full_early_exit.yaml
```diff
@@ -0,0 +1,140 @@
...
+
+# Fine-tuning arguments
+batch_size: 2
+epochs: 1
+optimizer:
+  _component_: torch.optim.AdamW
+  fused: True
+  lr: 2e-5
+loss:
+  _component_: torch.nn.CrossEntropyLoss
```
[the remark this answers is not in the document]
mostafaelhoushi: In principal, early exit loss should work with any loss function including chunked cross entropy. However, the current implementation doesn't support it.
The current implementation is parallelized: it stacks the tensor outputs of various layers into one big tensor, applies the language model head on that tensor, and then applies the loss function once. 
For chunked cross entropy, applying the language model head creates a list rather than a tensor and I found that difficult to tensorize. So in this PR I didn't implement it (I added an assertion statement to throw an error if that's the case).

So we have 2 options:
Option A We can have a simple albeit less efficient implementation that doesn't parallelize: just loop over the outputs of different layers and apply the language model head. This way, we can guarantee supporting any type of loss. We can have a condition or heurestic to guide us whether we can apply early exit losses across layers in parallel or in sequence.
Option B: not sure if there is a way to parallelize some sort of lambda function across various layers. Does functorch help with that
Any other options?

Not sure if you would like to implement any of the options above in this PR or in a future PR. For this PR, I will need to re-run the training experiments to verify we get speedups during inference.

recipes/dev/early_exit_finetune_distributed.py
```diff
@@ -0,0 +1,1077 @@
...
+        All early exit loss related setup happens here.
+        """
+        do_output_hidden_states = None
+        early_exit_loss_curriculum = None
+
+        if cfg_early_exit_loss:
+            assert (
+                hasattr(self._loss_fn, "reduction")
+                and self._loss_fn.reduction == "mean"
+            ), "Currently early exit loss is only implemented for loss functions that apply a mean reduction."
```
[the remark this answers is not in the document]
mostafaelhoushi: This is related to the comment above: the parallelized implementation assumes these conditions.
We can though revisit the implementation to ensure it is more generic.

torchtune/modules/early_exit_loss.py
mostafaelhoushi: The idea here is that I am trying to ensure that we use the outputs of the last layer of the model. If we don't use the last layer's output, we are essentially pruning layers from the model, and the model's accuracy will drop a lot. So I created this parameter `train_last_layer` and set it to True by default.

There could be one scenario though where we want `train_last_layer` to be False: if we freeze the model's weights, and add early exit heads at different layers. (By head we mean the model's last layer norm and linear layer). However, currently this PR supports shared head only (all early exits use the same head).

I took a look at the protocol in the link provided. What I understood from it is that it is to support adding adapter modules to a model. I think we can use that in the future if we want to support each layer having it's own head, but now we are not adding any heads or layers or adapters to the model. What do you think?

torchtune/modules/early_exit_loss.py
mostafaelhoushi: Done repeating args in e5534eaf48ac8b4f42c7629ed76727d2d95e7b86

recipes/dev/early_exit_finetune_distributed.py
```diff
@@ -0,0 +1,1077 @@
...
+                cfg_early_exit_loss.get("layers", ":"), len(self._model.layers)
+            )
+            train_last_layer = cfg_early_exit_loss.get("include_last_layer", True)
+            verbose = cfg_early_exit_loss.get("verbose", False)
+
+            if cfg_early_exit_loss.curriculum:
+                early_exit_loss_curriculum = setup_early_exit_loss_curriculum(
+                    early_exit_curriculum=cfg_early_exit_loss.curriculum,
+                    do_output_hidden_states=do_output_hidden_states,
+                    max_steps=self.total_epochs * self._steps_per_epoch,
```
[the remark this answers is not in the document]
mostafaelhoushi: I followed `lr_scheduler` constructor that has `last_epoch` argument [here](https://github.com/pytorch/torchtune/blob/b5d2e6372017c163914b13b2514f29914e5dbb84/recipes/knowledge_distillation_distributed.py#L320) so I added a `last_step` argument to `EarlyExitCurriculum` in  6a2d79ba71b8813bcb4f0cdcefae12a0636c6ec7

torchtune/modules/early_exit_loss.py
mostafaelhoushi: I agree. In ac5616d0000ac575899ed85a90d4f52413b95e25 I renamed `percent_scale` to `fraction_scale` and then changed the condition and docstring to make it clearer.

torchtune/modules/early_exit_loss.py
ebsmothers: Oh yeah the adapter code pointer was more in terms of providing a looser protocol-style structure to inherit from instead of having e.g. the get method and various instance variables defined. It's probably more relevant if we don't need the `train_last_layer` variable since then `get` basically becomes just a getter for `self._do_output_hidden_states`. But if we do need some custom handling that is common to all curricula I think what you're doing here makes sense.

recipes/dev/7B_full_early_exit.yaml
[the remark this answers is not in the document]
ebsmothers: Thanks for the additional information. It sounds to me like either of these two options would involve some kind of extension of the early_exit_loss as you've defined it, so I think leaving it for a follow-up makes sense (especially given that you already assert to prevent accidentally using with our `CEWithChunkedOutputLoss`).

torchtune/modules/early_exit_loss.py
```diff
@@ -0,0 +1,334 @@
...
+        loss_scales = 1.0 / (layer_ids + 1)
+    elif loss_scale_type == LossScaleType.INV_SQRT_L:
+        loss_scales = torch.reciprocal(torch.sqrt(layer_ids + 1))
+    else:
+        raise ValueError(f"Unsupported loss_scale type {loss_scale_type}")
+
+    loss_scales = loss_scales * torch.where(layer_ids < n_layers - 1, e_scale, 1.0)
+    # normalize loss scales to ensure that their sum is 1.0
+    loss_scales = loss_scales / torch.sum(loss_scales)
+    assert torch.isclose(torch.sum(loss_scales), torch.Tensor([1.0]).to(loss_scales))
```
ebsmothers, line 136: nit but I don't think this should be needed given the preceding line (are you worried about some sort of numerical issue?)

tests/torchtune/modules/test_early_exit_loss.py
```diff
@@ -0,0 +1,180 @@
...
+    curriculum = RotationalEarlyExitCurriculum(
+        [True, False, False], max_steps=100, train_last_layer=train_last_layer
+    )
+    expected = np.array([True, False, train_last_layer])
+    assert np.array_equal(curriculum.get(), expected)
+    curriculum.step()
+    expected = np.array([False, True, train_last_layer])
+    assert np.array_equal(curriculum.get(), expected)
+    curriculum.step()
+    expected = np.array([False, False, True])
```
ebsmothers, line 112: Took me a minute to understand this line.. maybe add a comment explaining that since the last element is already true on this rotation the value of `train_last_layer` has no effect

torchtune/modules/early_exit_loss.py
```diff
@@ -0,0 +1,334 @@
...
+    return loss_scales
+
+
+class EarlyExitCurriculumType(str, Enum):
+    NONE = "none"
+    ROTATIONAL = "rot"
+    GRADUAL = "gradual"
+
+
+# TODO: create a base curriculum class that can be used for other aspects, e.g., dropout, datasets, etc.
```
ebsmothers, line 147: I don't think I fully understand this TODO. Isn't this the base class?

torchtune/modules/layer_dropout.py
```diff
@@ -0,0 +1,302 @@
...
+        out = input.clone()
+        assert (
+            self.dim == 0
+        ), "Currently only supporting dropping elements along the 0th dimension"
+        if ind_selected.numel() > 0:
+            out[ind_selected] = out_selected
+        return out
+
+
+class ModuleLayerDropoutWrapper(torch.nn.Module):
```
ebsmothers, line 99: I really like the solution you came up with here

torchtune/modules/layer_dropout.py
```diff
@@ -0,0 +1,302 @@
...
+            (Fan et al., https://arxiv.org/abs/1909.11556v1)
+        >>> prepare_layer_dropout(model, prob_max=0.2, prob_layer_scale=ScaleType.UNIFORM, layers_str="::2")
+        >>> # Apply layer dropout that increases linearly across layers, as described in Progressive Layer
+            Dropout paper (Zhang et al., https://arxiv.org/abs/2010.13369)
+        >>> prepare_layer_dropout(model, prob_max=0.2, prob_layer_scale=ScaleType.LINEAR)
+        >>> # Apply layer dropout that increases exponentially across layers, as described in
+            LayerSkip paper (Elhoushi et al., https://arxiv.org/abs/2404.16710)
+        >>> prepare_layer_dropout(model, prob_max=0.2, prob_layer_scale=ScaleType.EXP)
+    """
+    num_layers = len(model.layers)
```
ebsmothers, line 275: nit but you're assuming that model has `layers` attributed defined. I don't see an obvious way around it though, maybe just add an explicit assert to verify this?

torchtune/modules/layer_dropout.py
```diff
@@ -0,0 +1,302 @@
...
+        ScaleType.UNIFORM: 1.0,
+        ScaleType.EXP: math.pow(2, val / scale_period) - 1,
+        ScaleType.LINEAR: val / scale_period,
+        ScaleType.LOG: math.log(val + 1, scale_period + 1),
+        ScaleType.SIN: math.sin(0.5 * math.pi * val / scale_period),
+        ScaleType.SIGMOID: 1 / (1 + math.exp(-10 * (val / scale_period - 0.5))),
+    }[scale_type]
+
+    # after scale_period, scale should be 1
+    return min(scale, 1.0)
```
ebsmothers, line 216: At least for sin you may have to also include a `max(*, 0.0)` (though in that case for val > scale_period I guess you will start getting weird behavior anyways)

recipes/dev/early_exit_finetune_distributed.py
```diff
@@ -0,0 +1,1068 @@
...
+                and self._loss_fn.reduction == "mean"
+            ), "Currently early exit loss is only implemented for loss functions that apply a mean reduction."
+
+            do_output_hidden_states = slice_str_to_array(
+                cfg_early_exit_loss.get("layers", ":"), len(self._model.layers)
+            )
+            train_last_layer = cfg_early_exit_loss.get("include_last_layer", True)
+            verbose = cfg_early_exit_loss.get("verbose", False)
+
+            if cfg_early_exit_loss.curriculum:
```
ebsmothers, line 713: I think it's possible that `cfg_early_exit_loss` may not have the `curriculum` attribute

torchtune/modules/early_exit_loss.py
```diff
@@ -0,0 +1,334 @@
...
+import numpy as np
+import torch
+
+from torchtune import utils
+from torchtune.modules.transformer import TransformerDecoder
+
+log = utils.get_logger("DEBUG")
+
+
+class LossScaleType(str, Enum):
```
ebsmothers, line 20: General question about some of these enums (e.g. this one, `EarlyExitCurriculumType`, `ScaleType`). Do you find that it is easier to use these rather than just have e.g. `layer_ids_to_loss_scales` be an arbitrary `Callable[[torch.Tensor, int], torch.Tensor]` (i.e. that is taking in `layer_ids` and `n_layers` and returning the scales)? If so, why? Personally I feel it's a bit more flexible to do it the latter way but would be interested to know if you find the Enum approach to be better for experimentation (or for other reasons I'm missing). Anyways not a request for a change, just wanna hear your thoughts.

ebsmothers (commented):
OK made it through the rest of the code. This is looking good! I left a handful more comments but don't have any huge concerns. Btw thank you for adding the extensive unit tests, it's really helpful.

torchtune/modules/early_exit_loss.py
mostafaelhoushi: I wasn't aware of `Protocol` in Python and I just read some tutorial about it to understand it.
Yes, I see that the curricula classes are in the borderline between classes inheriting from a parent class, and classes implementing a Protocol.

But because this `get()` function does some handling (rather than just returning `self._do_output_hidden_states` it makes a copy of it and sets the last element to `True` if `self.train_last_layer` is `True`) then maybe for now we can keep it as a parent class rather than a protocol?

torchtune/modules/layer_dropout.py
mostafaelhoushi: Good point. In 78aff5af2cc79f74d977510dbfa703af97aab033 , I changed the function to take `layers` as input rather than `model`.
This way the function makes no assumptions, and we hand over both the flexibility and  responsibility to the user to provide a list of layers to the function.

mostafaelhoushi (commented):
Thanks @ebsmothers for your detailed review. Just made a batch of commits and comments to address the feedback.

torchtune/modules/__init__.py
```diff
@@ -51,4 +53,6 @@
     "local_kv_cache",
     "delete_kv_caches",
     "disable_kv_cache",
+    "LayerDropout",
+    "prepare_layer_dropout",
```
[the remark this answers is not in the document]
ebsmothers: Oh interesting. Have you chatted with torchao folks about this already? Personally I would still document it in the interim then just remove it later (unless for instance the migration is imminent), we try to keep doc coverage of any public APIs as close to 100% as possible.

recipes/dev/early_exit_finetune_distributed.py
ebsmothers: Oh the usage should be with the function/class you want to instantiate as part of the config (passed via the `_component_` field). So not too dissimilar to Hydra instantiate but `_target_` -> `_component_` (and we don't support nested instantiation or other advanced Hydra features).

So you would need to add the following to your config:

```
early_exit_loss:
  _component_: torchtune.modules.early_exit_loss.setup_early_exit_loss_curriculum
  ... # other existing fields
```

Then call as 

```
early_exit_loss_curriculum = config.instantiate(cfg_early_exit_loss, ...)
```

But this works best if the fields in your config are simply a passthrough to the function you're instantiating. Since you're modifying things in the recipe (e.g. renaming `curriculum` -> `early_exit_curriculum`, using a wrapper to cast the `layers` field into a list) you would need to change these behaviors. (Note that passing additional derived args like `max_steps` is fine as `instantiate` supports that. The bigger concern is that you do not have erroneous config keys floating around)

torchtune/modules/early_exit_loss.py
ebsmothers: Seems like the assert is still here?

torchtune/modules/early_exit_loss.py
ebsmothers: > I am trying to look for an example to follow but can't find. For example, how can layer_ids_to_loss_scales determine the scale type?

Basically I am thinking of something like this.

In the `early_exit_loss` function:

```
def early_exit_loss(
	# keep all the existing args except loss_scale_type, replace it with..
	loss_scale_fn: Callable[[Tensor, int, float], Tensor]
):
	# Everything in here is the same except the call to layer_ids_to_loss_scales becomes..
	losses_scales = loss_scale_fn(
	        torch.Tensor(hidden_layer_ids).to(losses_early),
	        len(model.layers),
		e_scale,
	)
```

Then we define all the different loss scale types like this:

```
def inv_sqrt_l_loss_scale(layer_ids: Tensor, n_layers: int, e_scale: float = 1.0) -> Tensor:
	loss_scales = torch.reciprocal(torch.sqrt(layer_ids + 1))
	loss_scales = loss_scales * torch.where(layer_ids < n_layers - 1, e_scale, 1.0)
	return loss_scales / torch.sum(loss_scales)
```

And finally the config would look like:

```
early_exit_loss:
  layers: ":"
  curriculum: "gradual"
  scale_type: torchtune.modules.early_exit_loss.inv_sqrt_l_loss_scale
  scale: 1.0
```

Note that the one final thing needed would be to convert the `scale_type` import path to the actual function it represents (since OmegaConf will just see this as a string). However, we have [_get_component_from_path](https://github.com/pytorch/torchtune/blob/e9b9ea568e169aea13ae2d03c0eb92ee241f47a1/torchtune/config/_utils.py#L37) for exactly this case.

The reason I like this approach is that then if I am a user who wants to experiment with my own scaling method, I just need to write the Python function for it and point the config to its path on my local machine. No need to go in and modify the enum.

> In test_layer_ids_to_loss_scales, I just made the PyTest fixture test all the possible values in the enum:

Oh that is quite nice. I guess we could define a global variable containing all the different supported functions, but I guess that's pretty much just a worse version of an enum 😅

recipes/dev/early_exit_finetune_distributed.py
```diff
@@ -0,0 +1,1068 @@
...
+    if not training.is_distributed():
+        raise RuntimeError(
+            "Distributed finetune recipe should be run via a distributed launcher."
+            "If using tune CLI, please specify --nnodes 1 and --nproc_per_node [num_gpus]"
+        )
+    init_process_group(backend="gloo" if cfg.device == "cpu" else "nccl")
+    if cfg.get("fsdp_cpu_offload", False):
+        # Utilize all available CPU cores for intra-op parallelism. This provides ~2x
+        # speed up when benchmarking fused AdamW on CPU
+        training.set_torch_num_threads()
```
ebsmothers, line 1057: (This is more as a note to myself but) we will need to incorporate the same set of changes from #2108 to get CPU offload + gradient clipping to work together here

torchtune/modules/early_exit_loss.py
ebsmothers: > But because this get() function does some handling (rather than just returning self._do_output_hidden_states it makes a copy of it and sets the last element to True if self.train_last_layer is True) then maybe for now we can keep it as a parent class rather than a protocol?

Yeah agreed, this makes sense to me

torchtune/modules/early_exit_loss.py
```diff
@@ -0,0 +1,334 @@
...
+    Args:
+        do_output_hidden_states (List[bool]): A list indicating whether each layer's hidden state
+            should be output to calculate their losses.
+        max_steps (int): The maximum number of steps in the curriculum.
+        train_last_layer (bool, optional): Whether to always calculate loss for the last layer. Defaults to True.
+        last_step (Optional[int]): The last step the curriculum stopped at in a previous run.
+            This is used when resuming training.
+        fraction_scale (float, optional): A scaling factor to determine at which fraction
+            of steps, all the layers will be enabled. At `steps = max_steps * fraction_scale`, all the layers will be enabled.
+        verbose (bool, optional): Whether to print verbose logs. Defaults to False.
```
ebsmothers, line 254: nit but in our docstrings I don't think we typically put optional unless we actually mean `Optional` type (i.e. `Union[T, None]`)

```suggestion
        train_last_layer (bool): Whether to always calculate loss for the last layer. Defaults to True.
        last_step (Optional[int]): The last step the curriculum stopped at in a previous run.
            This is used when resuming training.
        fraction_scale (float): A scaling factor to determine at which fraction
            of steps, all the layers will be enabled. At `steps = max_steps * fraction_scale`, all the layers will be enabled. Defaults to 0.5.
        verbose (bool): Whether to print verbose logs. Defaults to False.
```

ebsmothers (commented):
Left a few more comments (mostly just responses to open threads). But I think this is just about there, really excited to see it land!

torchtune/modules/__init__.py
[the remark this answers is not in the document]
mostafaelhoushi: OK. Added to API ref here b66e23b9e56dae6e10acf260fe6c22eec52fed0d
Yes, I spoke to Jesse Cai and there has been a PR but I need to follow up with them: https://github.com/pytorch/ao/issues/633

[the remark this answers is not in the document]
mostafaelhoushi: I also built docs and ensured it passes.

torchtune/modules/early_exit_loss.py
mostafaelhoushi: Oh Sorry. I made sure I removed the assertion in 2675b4c3fbe5ca4a9f0ea0ace4e9a3f485b87cd4

recipes/dev/early_exit_finetune_distributed.py
mostafaelhoushi: OK. I see, since we do some processing/casting rather than a simple passthrough then this might require some refactoring of the existing implementation. So you think we can have it in a future PR?

torchtune/modules/early_exit_loss.py
mostafaelhoushi: Changed scale_tyoe from enum to function in 00d8efa7ca63f6c10db0fc7ace1143f8f6ca71c1 and changed curriculum from enum to class in 78b8996545149433a69c1ba5a991c9a596c60708.

ebsmothers (approved):
Thanks @mostafaelhoushi for enabling early exit loss and layer dropout in torchtune! I'm really happy with how it turned out, this is one of my favorite features we've landed (especially given some of the more challenging design aspects you had to work through). Thanks for sticking through it, looking forward to seeing its applications in LayerSkip and elsewhere!

Issue #1072, opened 2024-06-07
FSDP2 + QLoRA: `NF4` dispatch error


jeromeku:
## FSDP2 + QLoRA: `NF4` Dispatch Error

Getting the following error when trying to run FSDP2 + QLoRA distributed fine-tuning:
```
[rank0]: NotImplementedError: NF4Tensor dispatch: attempting to run _c10d_functional.all_gather_into_tensor.default, this is not supported
```

Ran with the following command:
```
 tune run --nnodes=1 --nproc_per_node=2 lora_finetune_fsdp2 \
--config 7B_qlora_fsdp2.yaml
```
where `7B_qlora_fsdp2.yaml` is a local copy of `recipes/dev/7B_qlora_fsdp2.yaml`

`torch` and `torchao` are both nightly installs:
- torch: 2.4.0.dev20240603+cu121
- torchao: 0.2.0

## `fully_shard` wrapped units

When the model is wrapped with `fully_shard` [here](https://github.com/pytorch/torchtune/blob/1fa1f04baf124c074dcd93831fa38c8b657af1e9/recipes/dev/lora_finetune_fsdp2.py#L310-L321), the `lora` layers and `TransformerDecoderLayer` are individually wrapped, then the entire model (`TransformerDecoder`) is finally wrapped.  

Stepping through with a debugger, we can see that the top-level `FSDP` unit manages whatever has not been wrapped: the `tok_embeddings`, `RMSNorm` and `output projection` (same shape as the `embeddings` but not tied).  

Does this mean that these parameters will be `all-gathered` together?  That is, when the token embeddings are needed, the `RMSNorm` and `output projection` params will be gathered as well, even though they are at the start and end of the forward pass respectively and hence not needed at that time?

kartikayk:
@weifengpy or @awgu should have the most intelligent response to the the question. My understanding is that all modules in the root should be gathered together.

@weifengpy might also be able to comment on the NF4 issue

awgu:
@jeromeku regarding wrapped units, everything you said is correct! When you call `fully_shard(module)`, all parameters in `module.parameters()` becomes one group, except those already assigned to nested `fully_shard` call. (Here is the [code pointer](https://github.com/pytorch/pytorch/blob/5b3624117ac85a8dbd5486db27d23f2f5652289b/torch/distributed/_composable/fsdp/_fsdp_init.py#L73-L115) if you are interested.)

[...]

PR #1077, opened 2024-06-10
[QLoRA][FSDP2] saving state dict by upcasting NF4 to high precision

weifengpy:
saving state dict is the last piece of a functional distributed QLoRA. User need this to save model for inference or resume training
* unit test: `pytest -s tests/torchtune/utils/test_distributed.py -k state_dict`
* run recipe: `tune run --nnodes 1 --nproc_per_node 8 lora_finetune_fsdp2 --config llama2/7B_qlora epochs=1 max_steps_per_epoch=1 gradient_accumulation_steps=1`

This fixes NF4 dispatch error https://github.com/pytorch/torchtune/issues/1072
```[rank0]: NotImplementedError: NF4Tensor dispatch: attempting to run _c10d_functional.all_gather_into_tensor.default, this is not supported```

There are 2 options with pros and cons. This PR implements option 1

**option 1**: `get_full_model_state_dict` by calling `module.unshard()` -> `module.named_parameters()` -> `module.reshard()`
* Pros: un-blackbox upcasting so it's easy for user to customize. Avoid repeating FSDP2 all-gather logic by dispatching `all_gather_into` in `NF4Tensor` subclass
* Cons: we see `if NF4 ... else ...` in `get_full_model_state_dict`. We need to reserve memory for 1 TransformerBlock (it won't increase peak memory usage since fwd/bwd, see memory traces below)

**option 2**: `get_full_model_state_dict` by calling `DTensor(NF4).full_tensor()`. This requires dispatching `all_gather_into` in `NF4Tensor` subclass
* Pros: `get_full_model_state_dict` in TorchTune is the same across LoRA and QLoRA. The logic of  `DTensor(NF4).full_tensor()` will be hidden in TorchAO
* Cons: user need to change TorchAO in case upcasting NF4 to fp32 or turning off upcasting and save NF4 directly. dispatching `all_gather_into` repeats all-gather logic in FSDP2

**Memory traces for Option 1**: `module.unshard()` results in memory cost on 1 x TransformerBlock but it will not increase peak memory in fwd/bwd
Screenshot 2024-06-10 at 4 16 16 PM

### CODE REVIEW

File: torchtune/utils/_distributed.py [hunk 1]
```diff
@@ -345,12 +348,45 @@ def get_full_model_state_dict(
...
+        # Iterating from lowerer modules to higher
+        # Unsharding lora adapters before unsharding transformer block
+        for module_name, module in reversed(list(model.named_modules())):
+            if not isinstance(module, FSDPModule):
+                continue
+            module.unshard(async_op=False)
+            if is_rank_zero:
+                module_name = module_name.replace(f".{_CHECKPOINT_WRAPPED_MODULE}", "")
+                for local_fqn, param in module.named_parameters():
+                    local_fqn = local_fqn.replace(f".{_CHECKPOINT_WRAPPED_MODULE}", "")
```
ebsmothers, line 366: Is the need to replace `_CHECKPOINT_WRAPPED_MODULE` a consequence of our usage of the `apply_activation_checkpointing` API? And if so, can we remove this by switching to the torch.utils.checkpoint API? (Obviously not a concern for this PR, I am more just curious)

tests/torchtune/utils/test_distributed.py
```diff
@@ -397,6 +399,14 @@ def test_lora_state_dict(self):
 
     @gpu_test(gpu_count=2)
     def test_qlora_state_dict(self):
+        self.run_subtests(
+            {
+                "enable_activation_checkpointing": [False, True],
+            },
+            self._test_qlora_state_dict,
+        )
```
ebsmothers, line 407: nit: can just use `@pytest.mark.parametrize` for this? (Unless there's something about `FSDPTest` that precludes that, if so feel free to ignore this)

tests/torchtune/utils/test_distributed.py
```diff
@@ -411,28 +421,86 @@ def test_qlora_state_dict(self):
...
+        # fsdp model for saving state dict
+        fsdp_model_to_save = copy.deepcopy(base_model)
+        for m in fsdp_model_to_save.modules():
+            if enable_activation_checkpointing:
+                if isinstance(m, CheckpointWrapper):
+                    fully_shard(m)
+            else:
+                if isinstance(m, modules.TransformerDecoderLayer):
+                    fully_shard(m)
+        fully_shard(fsdp_model_to_save)
```
ebsmothers, line 444: The purpose here is to recreate the sharding of the QLoRA model, right? If so, why not shard LoRA layers first as is done elsewhere?

torchtune/utils/_distributed.py
```diff
@@ -345,12 +348,45 @@ def get_full_model_state_dict(
...
+    )
+    if has_nf4:
+        from torch.distributed._composable.fsdp.fully_shard import FSDPModule
+
+        # Iterating from lowerer modules to higher
+        # Unsharding lora adapters before unsharding transformer block
+        for module_name, module in reversed(list(model.named_modules())):
+            if not isinstance(module, FSDPModule):
+                continue
+            module.unshard(async_op=False)
```
ebsmothers, line 362: Is this line where your comment 

> We need to reserve memory for 1 TransformerBlock

is relevant? Also will we ever run into the case where we accidentally unshard the full model?

torchtune/utils/_distributed.py
```diff
@@ -345,12 +348,45 @@ def get_full_model_state_dict(
...
+            if is_rank_zero:
+                module_name = module_name.replace(f".{_CHECKPOINT_WRAPPED_MODULE}", "")
+                for local_fqn, param in module.named_parameters():
+                    local_fqn = local_fqn.replace(f".{_CHECKPOINT_WRAPPED_MODULE}", "")
+                    if len(module_name) > 0:
+                        full_fqn = module_name + "." + local_fqn
+                    else:
+                        full_fqn = local_fqn
+                    if full_fqn in cpu_state_dict:
+                        continue
```
ebsmothers, line 372: nit: I assume this is because we iterate over every param in every module bottoms-up? If so, can you just add a quick comment explaining this?

ebsmothers (approved):
Overall looks good! A few small questions/comments, but no major concerns.

tests/torchtune/utils/test_distributed.py
weifengpy: for exact QLoRA setup, I was actually thinking of using recipe test as the best place

for unit test and for best QPS, wrapping `TransformerDecoderLayer` is actually more optimal. I was wrapping LoRA to be fair with FSDP1 recipe. FSDP1 is base on flat-parameter and it has to wrap LoRA otherwise it cannot handle a mixing freezed and trainable parameters in one flat-parameter

torchtune/utils/_distributed.py
weifengpy: exactly! `module.unshard` is the reason of 1 x TransformerBlock

> accidentally unshard the full model

the unspoken knowledge is `module.unshard` is not recursive. if we call `.unshard` for root model, it will unshard tok_embeddings but NOT TransformerBlock

I also ensure we call unshard from lower module to higher module with `reversed(list(model.named_modules()))`. after `module.unshard(async_op=False)`, we always call `module.reshard()` to ensure freeing memory immediately

torchtune/utils/_distributed.py
weifengpy: you are right. `apply_activation_checkpointing` is polluting fqns and I have to get clean fqn by replacing str `_CHECKPOINT_WRAPPED_MODULE` 

If we use `torch.utils.checkpoint`, we would not need `_CHECKPOINT_WRAPPED_MODULE`

torchtune/utils/_distributed.py
weifengpy: you are right. I added comments to explain

PR #1082, opened 2024-06-12
Tokenizer redesign for better model-specific feature support

RdoubleA:
## Motivation

- Our tokenizers are not model specific and are rather tied to the base tokenizer, e.g., SentencePiece (SP) or TikToken (TT). These both implement `tokenize_messages` but this is not configurable for model-specific special tokens
- Both of these are highly coupled with llama2 and llama3. For example, all the special tokens for [TikToken are llama3 specific](https://github.com/pytorch/torchtune/blob/71741dfa34c9fcd6a3cae8deb56dc260a8be4ef0/torchtune/modules/tokenizers/_tiktoken.py#L83) and it is difficult to customize
- This has led to [Phi3 having its own tokenizer file](https://github.com/pytorch/torchtune/blob/71741dfa34c9fcd6a3cae8deb56dc260a8be4ef0/torchtune/models/phi3/_sentencepiece.py#L35) that reimplements SentencePieceTokenizer, but with custom special tokens. Ideally, this should be easily composable by configuring the phi3 specific tokens with SentencePiece as the base
- This problem will only get worse as foundation models add more capabilities such as tool calling to their respective tokenizers. For example, [Mistral v3 now uses their own special tokens on top of SentencePiece](https://docs.google.com/document/d/1zAa956bEz-CR6ca2BnDntwA1iB3qltU4SqH9GyoToaM/edit?usp=sharing), and we have no way to onboard this besides copying and reimplementing SentencePieceTokenizer. Redundancy is normally ok but not if it can be easily addressed with composability.
- If a new base tokenizer is used besides SentencePiece or TikToken, we should be flexible enough to quickly implement the underlying new encode and decode while still maintaining our higher level APIs (tokenize_messages, etc)

## Design proposal
Intuition: separate the two core APIs (encode/decode and tokenize_messages) that operate at different levels of abstraction:
- Encode/decode live in the base token encoding layer (SP or TT) and does not need model specific logic. This should be pulled out so any model can quickly use SP, TT, or any other base token encoding for encode/decode.
- tokenize_messages requires model specific special tokens placed in particular locations. Thus, there should be a model specific class that implements tokenize_messages

We can achieve the above with `BaseTokenizer` and `ModelTokenizer`.
- `BaseTokenizer` is the base abstract interface for any base tokenization model (SP or TT) that implements encode and decode
- `ModelTokenizer` is the base abstract interface for any model-specific tokenizer that implements tokenize_messages. All models will implement their own Tokenizer class based on this interface so they can control `tokenize_messages` logic

```

class BaseTokenizer(Protocol):
    """Abstract token encoding model"""

    def encode(self, text: str, **kwargs) -> List[int]:
        """
        Given a string, return the a list of token ids.
        """

    def decode(
        self, token_ids: List[int], include_special: bool = False, **kwargs
    ) -> str:
        """
        Given a list of token ids, return the decoded text.
        """
```
```
class ModelTokenizer(Protocol):
    """
    Abstract tokenizer that implements model specific special token logic in
    ``tokenize_message`` and ``tokenize_messages`` methods.
    """

    def tokenize_messages(
        self, messages: List[Message], **kwargs
        """
        Given a list of messages, return a list of tokens and list of masks for
        the concatenated and formatted messages.
        """
        pass
```

This means the SentencePieceTokenizer and TikTokenTokenizer will be refactored to separate out encode/decode and tokenize_messages logic.

```
class TikTokenBaseTokenizer(BaseTokenizer):
    def __init__(
        self,
        path: str,
        pattern: str,
        name: str,
        special_tokens: Dict[str, int],
    ):
        mergeable_ranks = load_tiktoken_bpe(path)
        self.tt_model = Encoding(
            name=name,
            pat_str=pattern,
            mergeable_ranks=mergeable_ranks,
            special_tokens=special_tokens,
        )
        ...

    def encode(
        self,
        text: str,
        add_bos: bool,
        add_eos: bool,
        allowed_special: Set[str] = set(),
    ) -> List[int]:
        ...

    def decode(
        self,
        token_ids: List[int],
        truncate_at_eos: bool = True,
    ) -> str:
        ...
```
```
class SentencePieceBaseTokenizer(BaseTokenizer):
    def __init__(
        self,
        path: str,
    ):
        spm_model = SentencePieceProcessor()
        spm_model.load(path)
        self.spm_model = spm_model
        # Special tokens are defined in the protobuf itself

    def encode(
        self,
        text: str,
        add_bos: bool,
        add_eos: bool,
        allowed_special: Set[str] = set(),
        trim_leading_whitespace: bool = False,
        prefix: Optional[str] = None,
    ) -> List[int]:
        ...

    def decode(
        self,
        token_ids: List[int],
    ) -> str:
        ...
```
And any model tokenizers would compose with the above classes:
```
class Llama3Tokenizer(ModelTokenizer):
    """
    tiktoken tokenizer configured with Llama3 Instruct's special tokens, as described in
    https://llama.meta.com/docs/model-cards-and-prompt-formats/meta-llama-3

    Args:
        path (str): Path to pretrained tiktoken tokenizer file.
    """

    def __init__(
        self,
        path: str,
    ):
        all_special_tokens_with_ids = self._get_all_special_tokens_with_ids()
        self.tt_model = TikTokenBaseTokenizer(
            path=path,
            name="llama3_tiktoken",
            pattern=CL100K_PATTERN,
            special_tokens=all_special_tokens_with_ids,
        )

    def tokenize_messages(
        self,
        messages: List[Message],
        tokenize_header: bool = True,
        add_eos: bool = True,
        ...
```
## Changelog
- Refactor SentencePieceTokenizer -> SentencePieceBaseTokenizer, TikTokenTokenizer -> TikTokenBaseTokenizer and only retain encode/decode logic
- Move tokenize_messages logic to `Llama2Tokenizer` and `Llama3Tokenizer` for SP and TT, respectively
- Create `GemmaTokenizer` and `MistralTokenizer` which leverage `SentencePieceBaseTokenizer`. The `tokenize_messages` logic is identical to `Llama2Tokenizer` (for now, Mistral needs to be updated with v3)
- Refactor Phi3's tokenizer to use `SentencePieceBaseTokenizer` and retain special token logic
- Revamp `DummyTokenizer` since it inherited previously from `SentencePieceTokenizer`
- Update all tests and docstrings

## Test plan
- Added unit tests for all tokenizers, `pytest tests` runs successfully
- Manual comparison of every updated tokenizer with the old version: https://gist.github.com/RdoubleA/a62855ae5440063d120de3c908c903e0

## Planned follow-ups
- Tokenizer live docs / API ref updates


torchtune/models/llama3/_special_tokens.json
```diff
@@ -0,0 +1,259 @@
+
```
joecummings, line 1: I like this format a lot b/c it mirrors how tokenizers are packaged with HF - can you explicitly call this a `_special_tokens_map.json` though?

torchtune/models/llama3/_tokenizer.py
```diff
@@ -0,0 +1,180 @@
...
+        self.start_header_id = all_special_tokens_with_ids["<|start_header_id|>"]
+        self.end_header_id = all_special_tokens_with_ids["<|end_header_id|>"]
+        self.eom_id = all_special_tokens_with_ids["<|eom_id|>"]
+        self.eot_id = all_special_tokens_with_ids["<|eot_id|>"]
+        self.python_tag = all_special_tokens_with_ids["<|python_tag|>"]
+
+        # During generation, stop when either eos_id or eot_id is encountered
+        self.stop_tokens = [self.eos_id, self.eot_id]
+
+    def _get_all_special_tokens_with_ids(self) -> Dict[str, int]:
```
joecummings, line 57: I would prefer this to be configurable with a `file_path` param rather than hardcoded in.

torchtune/models/llama3/_special_tokens.json
joecummings: Hmm, but should this technically be included in our repo? It gets packaged with the tokenizer we download so maybe we should just point to that one instead??

joecummings: Thus, this file would be deleted.

RdoubleA: That's an interesting idea, I'm just worried that gives us less control over when certain special tokens are officially supported in our library. For example, Mistral v3 has an updated special_tokens_map but we currently only have one MistralTokenizer class that doesn't yet use the new special tokens. If we loaded in the special_tokens_map.json from HF for Mistral v3 the tokenizer will not work as expected. Then we would basically need a separate tokenzier class for each iteration of the model. We could compare it against the HF special_tokens_map.json as part of our correctness check.

joecummings: I think this is the same as our guarantees around supported models in torchtune. We make no claims to support Mistral V3 ATM so users should not expect that it works OOTB, despite likely being able to load it with a little messing around. The same principle applies for tokenizers.

joecummings: And when we make claims that we fully support a model, then that means we should fully support all the special tokens needed to run that model. Hence, the special_tokens_map file that comes with the model should be guaranteed to work with torchtune.

joecummings: In addition, let's say someone wants to add a special token for their finetuned model. If we allow people to point to a file with the special tokens map supplied, it will be trivial to do so.

RdoubleA: OK, I am mostly convinced. I looked at the jsons that are typically on the model pages on HF and `tokenizer.json` is the one we want. The others don't seem to have the complete list of special tokens. The only other thing is we'd have to add an additional field for the path to the special tokens json for all tokenizers. I do like the ease of seeing all the special tokens just listed there on our repo though, but maybe we can just document it somewhere for each model...

kartikayk (commented):
Want to express my uncertainty on the ```TokenEncoding``` naming. This seems a bit unintuitive since I'd imagine the encoding to refer to something like UTF8.

Not the best, but should we consider Tokenizer (and so SentencePieceTokenizer derives from Tokenizer) and ModelTokenizer (and so Llama3Tokenizer derives from ModelTokenizer). I don't love ModelTokenizer, but with the right doc strings I think it's passable.

torchtune/data/tokenizers/_tiktoken.py
```diff
@@ -0,0 +1,132 @@
...
+
+
+class TikTokenEncoding(TokenEncoding):
+    """
+    A lightweight wrapper around tiktoken Encoding. This class additionally handles
+    breaking up the input text into substrings of a max length and splitting up long
+    repetitions to improve encode speed.
+
+    The core tiktoken.Encoding does NOT natively handle BOS and EOS ids, so this class
+    does not either. They must be handled by the caller of ``encode``.
```
ebsmothers, line 25: This is a bit confusing to me (I understand the motivation though). In practice I would strive for more consistency in the `encode` APIs -- obviously there are some pretty custom things in sentencepiece that we don't wanna recreate  across the board, so not saying the signatures should match exactly or anything. But BOS and EOS handling should imo be exposed in the same place

torchtune/data/tokenizers/_tiktoken.py
```diff
@@ -0,0 +1,132 @@
...
+            pat_str=pattern,
+            mergeable_ranks=mergeable_ranks,
+            special_tokens=special_tokens,
+        )
+        # Vocab size without special tokens
+        self.base_vocab_size = len(mergeable_ranks)
+        # Vocab size with special tokens
+        self.vocab_size = self.tt_model.n_vocab
+
+    def _split_long_repetitions(
```
ebsmothers, line 54: kind of a nit but strictly speaking this does not need to be a method on the class, right?

torchtune/models/llama3/_special_tokens.json
ebsmothers: A question from elsewhere on this PR that may be relevant to this discussion.. can we just define `special_tokens` as a property of the `Tokenizer` protocol? Then we have some formal contract about where it's exposed. (I agree that we should probably not store any actual JSON files like this in the repo)

ebsmothers:
Overall I like the proposal. Still going through all the code, but the division of `encode` and `tokenize_messages` APIs you've provided here definitely makes sense to me. One big question I have is whether we can do this in a way where we don't have to write an entirely new tokenizer class every time we onboard a new model. I get that if there's custom logic around special tokens we can't really avoid it, but for vanilla sentencepiece without much special formatting it'd be nice to make this process a bit more lightweight (e.g. Gemma and Llama2 tokenizers have the same `tokenize_messages` API, and imo copy-pasting this code is not as nice as than copy-pasting model builders, which are generally pretty straightforward to understand).


Also: 

> Move torchtune.modules.tokenizers to torchtune.data.tokenizers which is a more relevant location

I understand that this is a useful thing to do, but it makes reviewing this code harder and is kinda logically distinct from the important stuff in this PR (imo). With this move, I have to look at tokenizers in their entirety; without it I can actually see the diff.

RdoubleA:
> Want to express my uncertainty on the TokenEncoding naming. This seems a bit unintuitive since I'd imagine the encoding to refer to something like UTF8.

Yeah I agree I don't like the TokenEncoding naming either, but naming SP and TT as Tokenizer and everything else as ModelTokenizer is more confusing imo... there should be a clear distinction between a model tokenizer and the base tokenizer. Maybe we could call it `SentencePieceBaseTokenizer` and `TikTokenBaseTokenizer`

torchtune/data/tokenizers/_tiktoken.py
RdoubleA: this is only specific to tiktoken and will not be used elsewhere, correct? but I don't feel strongly either way

torchtune/data/tokenizers/_tiktoken.py
RdoubleA: I'm actually not sure why I didn't just move EOS and BOS handling to the TikToken class..

torchtune/models/llama2/_tokenizer.py
```diff
@@ -0,0 +1,184 @@
...
+            add_eos=add_eos,
+            trim_leading_whitespace=trim_leading_whitespace,
+        )
+
+    def decode(
+        self,
+        token_ids: List[int],
+        include_special: bool = False,
+    ) -> str:
+        return self._spm_model.decode(token_ids, include_special=include_special)
```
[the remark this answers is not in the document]
RdoubleA: Yeah I mainly surfaced the encode/decode in the model tokenizers for generation utilities. Otherwise we'd have to access the underlying model for encoding/decoding... which maybe is ok if we enforce a common class attribute name in the model tokenizer protocol.

torchtune/models/mistral/_tokenizer.py
```diff
@@ -0,0 +1,156 @@
...
+        Args:
+            messages (List[Message]): A list of messages, each containing role, content,
+                and masked attributes.
+            max_seq_len (Optional[int]): A max sequence length to truncate tokens to.
+                Default: None
+
+        Returns:
+            Tuple[List[int], List[bool]]: The tokenized messages
+        """
+        return tokenize_messages_no_special_tokens(
```
joecummings, line 149: Mistral has special tokens, no?

torchtune/models/gemma/_tokenizer.py
```diff
@@ -0,0 +1,129 @@
...
+        Args:
+            messages (List[Message]): A list of messages, each containing role, content,
+                and masked attributes.
+            max_seq_len (Optional[int]): A max sequence length to truncate tokens to.
+                Default: None
+
+        Returns:
+            Tuple[List[int], List[bool]]: The tokenized messages
+        """
+        return tokenize_messages_no_special_tokens(
```
joecummings, line 122: Gemma has special tokens, no?

torchtune/modules/tokenizers/_utils.py
```diff
@@ -4,58 +4,182 @@
...
+
+    # Finally, truncate if necessary
+    if max_seq_len:
+        tokenized_messages = truncate(tokenized_messages, max_seq_len, eos_id)
+        mask = truncate(mask, max_seq_len, message.masked)
+
+    return tokenized_messages, mask
+
+
+def parse_hf_tokenizer_json(tokenizer_json_path: str) -> Dict[str, int]:
```
joecummings, line 171: Is this used anywhere right now? Or just including it for future use? If the latter - while I do like it - maybe leave for when we need it.

RdoubleA: This is currently used in llama3 and phi3

RdoubleA:
@joecummings huh I guess Gemma does have special tokens, but our current tokenizer does not use them. what do you think about either punting the upgrade to later or parsing the special tokens from the HF json for now and then adding them in tokenize_messages as appropriate later? Same question for Mistral

torchtune/models/llama2/_tokenizer.py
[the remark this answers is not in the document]
joecummings: Could we call it "decode_messages" on the `ModelTokenizer` and have it handle filtering out the special tokens?

torchtune/models/llama2/_tokenizer.py
```diff
@@ -0,0 +1,141 @@
...
+
+        # Original tokenizer has no pad_id, which causes indexing errors when batch training
+        self._spm_model.pad_id = 0
+
+        # This is used in tokenize_messages: if the tokenizer does not
+        # encode whitespace, then we can more easily split strings
+        # on whitespace characters and encode them separately.
+        self.encodes_whitespace = any(
+            [self._spm_model.encode(c) for c in WHITESPACE_CHARS]
+        )
```
ebsmothers, line 56: I don't like this being in the model-specific tokenizers. It's very much a hack to get `encode(a) + encode(b) = encode(a+b)` to hold for sentencepiece. But if I am a user copy-pasting tokenizers I am gonna have no idea what the hell is going on here

torchtune/models/gemma/_tokenizer.py
```diff
@@ -0,0 +1,129 @@
...
+
+    @property
+    def pad_id(self):
+        return self._spm_model.pad_id
+
+    @property
+    def vocab_size(self):
+        return self._spm_model.vocab_size
+
+    def encode(
```
ebsmothers, line 69: Discussed a bit offline, but currently the encode/decode methods being defined in both Tokenizer and ModelTokenizer is confusing. Right now it's needed because we do call encode directly on ModelTokenizer in our generate recipe, but as @RdoubleA pointed out we should see if we can move everything onto tokenize_messages API (and deprecate encode/decode methods for ModelTokenizers) after #1106

torchtune/models/llama3/_tokenizer.py
```diff
@@ -0,0 +1,181 @@
...
+    Args:
+        path (str): Path to pretrained tiktoken tokenizer file.
+        special_tokens_path (str): Path to ``tokenizer.json`` from Hugging Face
+            model files that contains all registered special tokens.
+    """
+
+    def __init__(
+        self,
+        path: str,
+        special_tokens_path: str,
```
ebsmothers, line 36: So I understand the purpose of this and it is nice that we can just directly load from HF. But I also don't love that (a) we are very tied to their specific tokenizer representation (and not just generic json) and (b) it kinda abstracts away the actual special tokens a bit (though I guess you do define them below). Another approach here would be to pass `special_tokens: Dict[str, int]` and leave it to the builder to take the tokenizer json path and unpack it. Tbh I don't have a strong opinion here though

joecummings: Ohh the second option sounds good. We will provide the function to load and build the special tokens via the HF file but allow people to pass in a dict if they want. Smart cookie.

torchtune/models/llama2/_tokenizer.py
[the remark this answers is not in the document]
joecummings: BUmp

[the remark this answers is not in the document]
RdoubleA: sorry I missed this. Yes I agree, will do that

recipes/eleuther_eval.py
```diff
@@ -42,7 +42,7 @@ class _EvalWrapper(HFLM):
 
     Args:
         model (TransformerDecoder): The model to evaluate.
-        tokenizer (Tokenizer): The tokenizer to use.
+        tokenizer (ModelTokenizer): The tokenizer to use.
```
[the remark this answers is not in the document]
RdoubleA: oh no

torchtune/datasets/_chat.py
```diff
@@ -48,7 +48,7 @@ class ChatDataset(Dataset):
     turns does not fit within ``max_seq_len`` then it is truncated.
 
     Args:
-        tokenizer (Tokenizer): Tokenizer used to encode data. Tokenize must implement an ``encode`` and ``decode`` method.
+        tokenizer (ModelTokenizer): Tokenizer used to encode data. Tokenize must implement an ``encode`` and ``decode`` method.
```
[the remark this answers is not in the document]
RdoubleA: yes, will revise

torchtune/models/llama2/_tokenizer.py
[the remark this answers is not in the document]
RdoubleA: actually, I removed `include_special` from decode here, so this no longer applies right? and the generate recipe still requires using `decode` from the `ModelTokenizer`

tests/torchtune/datasets/test_alpaca_dataset.py
```diff
@@ -9,18 +9,15 @@
...
 from torchtune.datasets import alpaca_cleaned_dataset, alpaca_dataset
-from torchtune.modules.tokenizers import SentencePieceTokenizer
 
 
 class TestAlpacaDataset:
     @pytest.fixture
     def tokenizer(self):
-        # m.model is a pretrained Sentencepiece model using the following command:
-        # spm.SentencePieceTrainer.train('--input=<TRAIN_FILE> --model_prefix=m --vocab_size=2000')
-        return SentencePieceTokenizer(str(get_assets_path() / "m.model"))
```
ebsmothers, line 23: Any concerns that by removing these tokenizers we are now not actually testing any tokenizers that have their open proper encode method?

tests/torchtune/modules/tokenizers/test_sentencepiece.py
```diff
@@ -7,18 +7,19 @@
...
 
-class TestSentencePieceTokenizer:
+class TestSentencePieceBaseTokenizer:
     @pytest.fixture
     def tokenizer(self):
         # m.model is a pretrained Sentencepiece model using the following command:
         # spm.SentencePieceTrainer.train('--input=<TRAIN_FILE> --model_prefix=m --vocab_size=2000')
-        return SentencePieceTokenizer(str(ASSETS / "m.model"))
+        sp_tokenizer = SentencePieceBaseTokenizer(str(ASSETS / "m.model"))
+        sp_tokenizer.encodes_whitespace = True
```
ebsmothers, line 21: Why do we need to set this explicitly?

torchtune/models/mistral/_tokenizer.py
ebsmothers: I don't think so? https://huggingface.co/mistralai/Mistral-7B-v0.1/blob/main/tokenizer_config.json

torchtune/models/phi3/_tokenizer.py
```diff
@@ -4,101 +4,81 @@
...
-            "<|placeholder5|>": 32008,
-            "<|placeholder6|>": 32009,
-            "<|user|>": 32010,
-        }
-
-        self.vocab_size = spm_model.vocab_size()
-        self.bos_id = spm_model.bos_id()
+        self._spm_model = SentencePieceBaseTokenizer(path)
+
+        self.special_tokens = special_tokens or PHI3_SPECIAL_TOKENS
```
ebsmothers, line 52: nit: what if I want to explicitly pass an empty dict (i.e. no special tokens)? I think this will override that, right? (Maybe this isn't a realistic case, I'm not super familiar with phi3 tokenizer usage)

torchtune/modules/tokenizers/_utils.py
```diff
@@ -4,58 +4,178 @@
...
+            text (str): The text to encode.
+
+        Returns:
+            List[int]: The encoded list of token ids.
         """
+        pass
 
     def decode(
-        self, token_ids: List[int], add_bos: bool, add_eos: bool, **kwargs
+        self, token_ids: List[int], include_special: bool = False, **kwargs
```
ebsmothers, line 32: Maybe I'm being dense, I don't see `include_special` being used anywhere?

torchtune/modules/tokenizers/_utils.py
```diff
@@ -4,58 +4,178 @@
...
+        if start_of_turn:
+            tokenized_messages.append(bos_id)
+            mask.append(message.masked)
 
-        if current_slice_is_space ^ is_now_space:
-            current_slice_len = 1
-            current_slice_is_space = is_now_space
+        # We want to trim leading whitespace on the next message when
+        # (a) it is a continuation of the turn (i.e. not the first message)
+        # (b) the vocabulary explicitly encodes whitespace characters, and
```
ebsmothers, line 131: (b) is no longer checked here. Personally I think it's OK to leave this in as we want to be explicit about what we're doing here, but would at least update this comment to be a bit more clear about the fact that this should be checked inside the base tokenizer's encode method

torchtune/models/gemma/_tokenizer.py
ebsmothers: Yeah what's the latest on this? It seems to me that it does have special tokens: https://huggingface.co/google/gemma-2b/blob/main/tokenizer_config.json

torchtune/models/llama3/_tokenizer.py
```diff
@@ -0,0 +1,175 @@
...
+
+        # Encode BOS and EOS, define pad ID
+        self.bos_id = self.special_tokens["<|begin_of_text|>"]
+        self.eos_id = self.special_tokens["<|end_of_text|>"]
+        self.pad_id = PAD_ID
+
+        # Encode extra special tokens
+        self.start_header_id = self.special_tokens["<|start_header_id|>"]
+        self.end_header_id = self.special_tokens["<|end_header_id|>"]
+        self.eot_id = self.special_tokens["<|eot_id|>"]
```
ebsmothers, line 55: Not a huge deal, but I feel like this isn't the most configurable way to do this. Like we claim to support custom special tokens, but they also need to be a superset of LLAMA3_SPECIAL_TOKENS, which is not immediately obvious and requires some digging around to figure out. I guess the obvious alternative here is something like `extra_special_tokens` which tbh I don't love either. Maybe just add an assert or something with an explicit error message about which keys must be present when using this tokenizer.

torchtune/modules/tokenizers/_sentencepiece.py
```diff
@@ -60,19 +59,21 @@ def encode(
...
+                added to s2. This will only trim leading whitespace if the underlying
+                ``SentencePieceProcessor`` encodes whitespace. Default: False
             prefix (Optional[str]): Optional string to encode for trimming leading
                 whitespaces. Used only if trim_leading_whitespace=True. Default: None
+
         Returns:
             List[int]: The encoded token IDs.
         """
-        if trim_leading_whitespace:
+        if trim_leading_whitespace and self.encodes_whitespace:
```
ebsmothers, line 76: The rationale here is also less clear now that we've split encode and tokenize_messages. Please make sure to document why we're doing this

ebsmothers (approved):
A handful more comments but no huge concerns from my side. Accepting to unblock

tests/torchtune/modules/tokenizers/test_sentencepiece.py
RdoubleA: because the test model actually does not encode whitespaces and breaks the test. Alternatively, I can update the test values

torchtune/models/phi3/_tokenizer.py
RdoubleA: It will override it, but I don't think this is a case we need to worry about. If there's no special tokens, then why use the Phi3Tokenizer? But, to be extra cautious, it's easy to just make this check for None explicitly.

torchtune/modules/tokenizers/_utils.py
RdoubleA: It's not. I may just remove it. It's not straightforward to enable including special tokens for SentencePiece, at least.

tests/torchtune/modules/tokenizers/test_sentencepiece.py
ebsmothers: Sorry, what I mean is: why is this not already figured out by `SentencePieceBaseTokenizer`? Linking to where this should happen

torchtune/modules/tokenizers/_sentencepiece.py
```diff
@@ -38,16 +40,13 @@ def __init__(
...
         self.pad_id = spm_model.pad_id()
 
-        # This is used in tokenize_messages: if the tokenizer does not
-        # encode whitespace, then we can more easily split strings
+        # If the tokenizer does not encode whitespace,
+        # then we can more easily split strings
         # on whitespace characters and encode them separately.
         self.encodes_whitespace = any(
             [self.spm_model.encode(c) for c in WHITESPACE_CHARS]
         )
```
ebsmothers, line 48: Shouldn't the test tokenizer already have `encodes_whitespace` properly defined here?

tests/torchtune/models/phi3/test_phi3_tokenizer.py
```diff
@@ -0,0 +1,252 @@
...
+from pathlib import Path
+
+import pytest
+from torchtune.data import Message
+from torchtune.models.phi3 import phi3_mini_tokenizer
+
+ASSETS = Path(__file__).parent.parent.parent.parent / "assets"
+
+
+class TestPhi3MiniTokenizer:
```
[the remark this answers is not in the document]
RdoubleA: at least phi3 should be tested separately since the tokenize_messages is customized, but the others that simply use `tokenize_messages_no_special_tokens` there's not much difference. but it will be updated in the near future once they have their own special tokens (I need to look into Gemma more at least)

torchtune/models/gemma/_tokenizer.py
RdoubleA: these aren't used in any capacity in our current gemma implementation. Also I think we discussed elsewhere that the gemma and mistral special token updates will be a follow up: #1118

torchtune/models/mistral/_tokenizer.py
RdoubleA: not for our current implementation. Mistral v0.3 does.

docs/source/tutorials/chat.rst
```diff
@@ -114,14 +114,14 @@ If you look at our :class:`~torchtune.data.Llama2ChatFormat` class, you'll notic
 we don't include the :code:`<s>` and :code:`</s>` tokens. These are the beginning-of-sequence
 (BOS) and end-of-sequence (EOS) tokens that are represented differently in the tokenizer
 than the rest of the prompt template. Let's tokenize this example with the
-:class:`~torchtune.modules.tokenizers.SentencePieceTokenizer` used by Llama2 to see
+:class:`~torchtune.models.llama2.llama2_tokenizer` used by Llama2 to see
```
joecummings, line 117: nit: this is a function ref now

torchtune/models/llama3/_tokenizer.py
```diff
@@ -0,0 +1,175 @@
...
+        path (str): Path to pretrained tiktoken tokenizer file.
+        special_tokens (Optional[Dict[str, int]]): mapping containing special text tokens and
+            their registered token IDs. If left as None, this will be set to the canonical
+            Llama3 special tokens.
+    """
+
+    def __init__(
+        self,
+        path: str,
+        special_tokens: Optional[Dict[str, int]] = None,
```
joecummings, line 43: Why wouldn't we just default this to the LLAMA3_SPECIAL_TOKENS?

torchtune/models/mistral/__init__.py
```diff
@@ -23,8 +23,10 @@
     qlora_mistral_7b,
     qlora_mistral_classifier_7b,
 )
+from ._tokenizer import MistralTokenizer
 
 __all__ = [
+    "MistralTokenizer",
```
joecummings, line 29: This is in the `__all__`, but I don't see it updated in the model documentation. Therefore any links to this will not render and users won't be able to know what is going on with these tokenizers. Same applies to all model tokenizers.

torchtune/modules/tokenizers/__init__.py
```diff
@@ -4,8 +4,20 @@
...
+)
 
-__all__ = ["SentencePieceTokenizer", "TikTokenTokenizer", "Tokenizer"]
+__all__ = [
+    "SentencePieceBaseTokenizer",
+    "TikTokenBaseTokenizer",
+    "ModelTokenizer",
+    "BaseTokenizer",
+    "tokenize_messages_no_special_tokens",
+    "parse_hf_tokenizer_json",
```
joecummings, line 22: This needs to be documented.

joecummings (commented):
This is amazing work, but I see a few things that definitely need to be documented before landing.

torchtune/models/llama3/_tokenizer.py
RdoubleA: mutable types - if we put dict/list as default here it will be the same dict for all instances of Llama3Tokenizer. If it's ever updated it will updated for all of them. Very edge case, but better to just be safe.

docs/source/tutorials/chat.rst
RdoubleA: ah shoot

torchtune/modules/tokenizers/_sentencepiece.py
RdoubleA: Yes it is defined, it just turned out to be False

torchtune/models/llama3/_tokenizer.py
RdoubleA: Yeah, I think in the ruckus your original assert checks got thrown out. let me add that back in

PR #1083, opened 2024-06-13
[WIP] Sample packing deep dive

RdoubleA:
#### Context
Deep dive in live docs on sample packing with in-sample masking and position encoding and how it impacts training throughput and model quality

- [ ] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

#### Test plan

### DIFF
--- a/docs/source/deep_dives/sample_packing.rst
+++ b/docs/source/deep_dives/sample_packing.rst
@@ -1,0 +1,247 @@
+.. _sample_packing_deepdive:
+
+==========================================================================================
+Fine-tuning in torchtune with sample packing using in-sample position encoding and masking
+==========================================================================================
+
+torchtune now supports sample packing for significant boosts in training throughput. This
+can be enabled on CLI, in the yaml config, or in any dataset builder by setting ``packed=True``.
+
+.. code-block:: python
+
+    from torchtune.datasets import alpaca_dataset, PackedDataset
+
+    # Load in tokenizer
+    tokenizer = ...
+    dataset = alpaca_dataset(
+        tokenizer=tokenizer,
+        packed=True,
+    )
+    print(isinstance(dataset, PackedDataset))  # True
+
+.. code-block:: yaml
+
+    # YAML config
+    dataset:
+      _component_: torchtune.datasets.alpaca_dataset
+      packed: True
+
+.. code-block:: bash
+
+    # Command line
+    tune run full_finetune_single_device --config llama3/8B_full_single_device \
+    dataset.packed=True
+
+This will automatically adjust attention masks to avoid cross-sample attending and
+update position IDs to be relative, within-sample. In this write-up, we explore what
+the expected performance gains you should see are and how in-sample masking and position
+encoding can affect fine-tuned model quality.
+
+.. grid:: 1
+
+    .. grid-item-card:: :octicon:`mortar-board;1em;` Highlights
+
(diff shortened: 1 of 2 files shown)


ebsmothers:
doc build failed

docs/source/deep_dives/sample_packing.rst
```diff
@@ -0,0 +1,247 @@
...
+Context windows for modern LLMs are ever-increasing, from Llama3 at 8k to
+Gemini 1.5 Pro which is pushing 2 million. In practice, many OSS text datasets
+(such as Stanford Alpaca) consist of mostly short sequence length samples,
+averaging at ~100-1000, which is a small fraction of the full context. This can
+lead to significant padding compute and memory overhead, even if ignored or masked
+out in the loss function. Sample packing minimizes this overhead by jamming as
+many samples as possible in the extra space in the context window. This can massively
+increase throughput for model training, since the model will see more data per step.
+However, there are several caveats to consider when enabling sample packing for a
+transformer decoder model.
```
SalmanMohammadi, line 65: hope you don't mind me commenting here : )

small reformat:

```
However, there are several caveats to consider when enabling sample packing for a
transformer decoder model:

1. When concatenating multiple samples in a single context window, we need a way
to tell the model to not attend to all other previous samples and only attend to
the sample the current token belongs to. This can be achieved with a block causal
mask to prevent cross-contamination.

2. Similarly, for position encodings we need to encode relative position of a
token within its sample instead of absolute position across all samples in a
single context window.

3. Ideally, the block causal mask and relative position encoding is compatible
with flash attention to maximize speed-ups.

```

docs/source/deep_dives/sample_packing.rst
```diff
@@ -0,0 +1,247 @@
...
+tokens. This is sound for a single sample, where each token requires the context
+provided by all the previous. However, in a packed sample which may contain numerous
+different samples only separated by EOS tokens, all consecutive samples should not
+attend to previous samples. Instead, we can create a block causal mask where each
+sample within a pack has its own causal mask. This way the packs are less noisy
+to the model as it only has to focus on attending each sample to itself.
+
+[ADD MASK PICTURE]
+
+Block causal masks have been implemented in several publications (ref and ref) as
```
SalmanMohammadi, line 89: perhaps https://arxiv.org/pdf/2107.02027?

docs/source/deep_dives/sample_packing.rst
```diff
@@ -0,0 +1,247 @@
...
+a way to directly address issues with cross-contamination. Some users have observed
```
SalmanMohammadi, line 90: from section C: Related work in the above paper, they mention

> For instance, GPT-3  is trained with a maximum sequence length of 2048 where a single sequence may contain multiple segments of sentences separated by a special end of segment token ... GPT-3 does
not mask the attention between different segments in a sequence

and

>  RoBERTa and many other models in production like T5 [24 ] use a similar packing approach as GPT-3

docs/source/deep_dives/sample_packing.rst
```diff
@@ -0,0 +1,247 @@
...
+reasonable performance even in the absence of proper masking and position encoding,
+but this lacks systematic investigations. torchtune’s abstraction-free design makes
```
SalmanMohammadi, line 92: Torchtune's

PR #1084, opened 2024-06-13
[CLIP][IMAGE TRANSFORMS] Image transforms for clip encoder

felipemello1:
#### Context
- [x] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

Added the image transforms for clip.

Builder with the model specific values in:
`torchtune/models/clip/_model_builders.py
`

Clip specific use of transforms in:
`torchtune/models/clip/_transforms.py
`

Vision transforms in:
`torchtune/modules/transforms/vision
`

Every transform is a function.

Algorithm TLDR:
1. Gets list of possible resolutions based on tile_size and max_num_tiles;
2. Finds resolutions that best fits the image;
3. Resizes with distortion
4. Pads
5. Normalizes
6. tile_crop -> output is of shape [num_tiles, 3, tile_size, tile_size]

#### Changelog

Added torchvision to the requirements

#### Test plan

Every function is covered, some indirectly. All tests pass.

docs
image
image
image
image
image


felipemello1:
TODO: update examples with ">>>>"

torchtune/modules/transforms/pipelines.py
```diff
@@ -0,0 +1,154 @@
...
+
+ImageInput = Union["PIL.Image.Image", np.ndarray, "torch.Tensor"]
+
+
+class VariableImageSizeTransforms:
+    """
+    This class accepts images of any size and dynamically resize, pads and chunks it
+    based on the image aspect ratio and the number of image chunks we allow.
+
+    The algorithm will NOT distort the image fit a certain aspect ratio, because
```
kartikayk, line 34: nit: this sentence reads weirdly - do you mean "The algorithm will NOT distort the image to fit a certain aspect ratio"?

torchtune/modules/transforms/pipelines.py
```diff
@@ -0,0 +1,154 @@
...
+        best_resolution = self.get_best_resolution(image_size)
+
+        # resize to best_resolution while preserving aspect ratio
+        image = self.resize_without_distortion(image=image, target_size=best_resolution)
+
+        # pad to fit the best resolution
+        image = pad_image_top_left(image=image, target_size=best_resolution)
+
+        # Normalize and rescale
+        image = image.float()
```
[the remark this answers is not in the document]
felipemello1: What worries me are images from other domains like x-rays, for example, that are not bounded by 0:255. If we remove do_rescale, would we still be able to support those?

torchtune/modules/transforms/vision/get_canvas_best_fit.py
```diff
@@ -0,0 +1,178 @@
...
+
+    Args:
+        image (torch.Tensor): The image we want to fit into a canvas.
+        possible_resolutions (torch.Tensor): A tensor of shape (N, 2) where each
+            row represents a possible canvas.
+        resize_to_max_canvas (bool): If True, pick the canvas that allows maximum scaling.
+            If False, pick the canvas that minimizes downscaling, including no downscaling at all.
+    Returns:
+        Tuple[int, int]: The best resolution to fit the image into.
+    Examples:
```
joecummings, line 40: Nit: These examples are not rendering correctly in docs.

torchtune/modules/transforms/vision/resize_with_pad.py
```diff
@@ -0,0 +1,171 @@
...
+        image (torch.Tensor): The input image tensor in the format [..., H, W].
+        target_size (Tuple[int, int]): The desired resolution to fit the image into in the format [height, width].
+        resample (torchvision.transforms.InterpolationMode): Resampling method used when resizing images.
+            Supports ant torchvision.transforms.InterpolationMode, such as InterpolationMode.NEAREST,
+            InterpolationMode.NEAREST_EXACT, InterpolationMode.BILINEAR, InterpolationMode.BICUBIC.
+        max_upscaling_size (int): The maximum size to upscale the image to.
+            If None, will upscale up to target_size.
+    Returns:
+        torch.Tensor: The resized and padded image tensor in the format [..., H, W].
+    Examples:
```
joecummings, line 40: nit: These examples are not rendering correctly in docs.

torchtune/models/clip/_transforms.py
```diff
@@ -0,0 +1,176 @@
...
+            Should be the same used for the pre-trained model. If None, no normalization is performed.
+        possible_resolutions (Optional[List[Tuple[int, int]]]): List of possible resolutions as tuples (height, width).
+            If None, this will be calculated using max_num_tiles and tile_size.
+        tile_size (int): Size of the tiles to divide the image into
+        max_num_tiles (Optional[int]): Only used if possible_resolutions is NOT given.
+            Maximum number of tiles to break an image into.
+            This will be used to generate possible_resolutions,
+            e.g. [(224, 224), (224, 448), (448, 224)] if max_num_tiles = 2 and tile_size = 224.
+        resample (str): Resampling method used when resizing images. Supports any enum of
+            torchvision.transforms.InterpolationMode, e.g. "nearest", "nearest_exact", "bilinear", "bicubic".
```
RdoubleA, line 66: do you want to add a link to their docs?
```suggestion
            ``torchvision.transforms.InterpolationMode``, e.g. "nearest", "nearest_exact", "bilinear", "bicubic".
```

torchtune/models/clip/_transforms.py
```diff
@@ -0,0 +1,176 @@
...
+    image is resized without distortion (300,800) -> (448, 1194) #448 is the limiting side for the resize
+    image is padded (448, 1194) -> (448, 1344)
+    Image is tiled 2x5, for a final output shape of (10, 3, 224, 224)
+
+    Args:
+        image_mean (Optional[List[float]]): Mean values of each channel, used for normalization.
+            Should be the same used for the pre-trained model. If None, no normalization is performed.
+        image_std Union[float, List[float]]]): Standard deviation values of each channel, used for normalization.
+            Should be the same used for the pre-trained model. If None, no normalization is performed.
+        possible_resolutions (Optional[List[Tuple[int, int]]]): List of possible resolutions as tuples (height, width).
```
RdoubleA, line 58: maybe I'm still a bit noob but it's not clear to me what possible_resolutions mean, could you add another sentence describing what possible resolutions means in this case? and why multiple?

torchtune/models/clip/_transforms.py
```diff
@@ -0,0 +1,176 @@
...
+        assert (
+            possible_resolutions is not None or max_num_tiles is not None
+        ), f"Either possible_resolutions or max_num_tiles must be given. Got {possible_resolutions=} and {max_num_tiles=}"
+
+        # If possible_resolutions are not given, then calculate possible ones based on max_num_tiles
+        if not possible_resolutions and max_num_tiles:
+            possible_resolutions = find_supported_resolutions(
+                max_num_tiles=max_num_tiles, tile_size=tile_size
+            )
+        else:
```
RdoubleA, line 114: you don't need this else if possible_resolutions is not None, right?

torchtune/models/clip/_transforms.py
```diff
@@ -0,0 +1,176 @@
...
+            possible_resolutions = possible_resolutions
+
+        self.possible_resolutions = torch.tensor(possible_resolutions).reshape(-1, 2)
+        logger.info(f"possible_resolutions: {self.possible_resolutions}")
```
RdoubleA, line 118: nit: is this logging line helpful? seems a bit random, would say to make this more descriptive or remove

torchtune/models/clip/_transforms.py
```diff
@@ -0,0 +1,176 @@
...
+        self.max_upscaling_size = None if resize_to_max_canvas else tile_size
+        self.resample = torchvision.transforms.InterpolationMode[resample.upper()]
+
+        # tile_crop
+        self.tile_size = tile_size
+
+    def __call__(
+        self,
+        image: Image.Image,
+    ) -> Dict[str, torch.Tensor]:
```
RdoubleA, line 139: @pbontrager I think we want to standardize transforms around this call signature, right?
```
def __call__(self, *, image, **kwargs) -> Mapping[str, Any]
```
not blocking though since this is the first major transform

torchtune/models/clip/_transforms.py
```diff
@@ -0,0 +1,176 @@
...
+                image_tensor, mean=self.image_mean, std=self.image_std
+            )
+
+        # Divide the image into equally sized tiles
+        image_tensor = tile_crop(image=image_tensor, tile_size=self.tile_size)
+
+        aspect_ratio = torch.tensor(best_resolution).reshape(-1) // self.tile_size
+        return {
+            "image": image_tensor,
+            "ar": aspect_ratio,
```
RdoubleA, line 175: what will aspect ratio be used for down the pipeline?

torchtune/modules/transforms/vision/get_canvas_best_fit.py
RdoubleA: I think you need to add newlines between Args, Returns, Examples
Or you need to add a colon
```suggestion
    Examples::
```

torchtune/modules/transforms/vision/get_canvas_best_fit.py
```diff
@@ -0,0 +1,178 @@
...
+import torch
+
+logger = logging.getLogger(__name__)
+
+
+def get_canvas_best_fit(
+    image: torch.Tensor, possible_resolutions: torch.Tensor, resize_to_max_canvas: bool
+) -> Tuple[int, int]:
+    """
+    Determines the best canvas possible from a list of possible resolutions to
```
RdoubleA, line 21: the description for possible resolutions here is more informative, would be helpful to put it in the main class

torchtune/modules/transforms/vision/get_canvas_best_fit.py
```diff
@@ -0,0 +1,178 @@
...
+        ...     [224, 672],
+        ...     [672, 224],
+        ...     [224, 448],
+        ...     [448, 224],
+        ...     [224, 224]
+        ... ])
+        >>> get_canvas_best_fit(image, possible_resolutions, resize_to_max_canvas=False)
+        (224, 448)
+
+        In the example above, we calculate the scaling factors for each possible resolution
```
RdoubleA, line 53: or maybe these lines in between the code need to be tabbed back

torchtune/modules/transforms/vision/get_canvas_best_fit.py
```diff
@@ -0,0 +1,178 @@
...
+    for ar, resolution in asp_dict.items():
+        for height, width in resolution:
+            possible_resolutions.append((height * tile_size, width * tile_size))
+
+    return possible_resolutions
+
+
+def _get_factors(n: int) -> Set[int]:
+    """
+    Calculate all factors of a given number, i.e. a dividor that leaves no remainder.
```
RdoubleA, line 160:
```suggestion
    Calculate all factors of a given number, i.e. a divisor that leaves no remainder.
```

torchtune/modules/transforms/vision/get_canvas_best_fit.py
```diff
@@ -0,0 +1,178 @@
...
+        >>> _get_factors(n=12)
+        {1, 2, 3, 4, 6, 12}
+    """
+    factors_set = set()
+
+    for i in range(1, int(n**0.5) + 1):
+        if n % i == 0:
+            factors_set.add(i)
+            factors_set.add(n // i)
+    return factors_set
```
RdoubleA, line 178: maths

RdoubleA (commented):
Looks awesome overall, thanks for all the work you put into this. One overall question I had is the distinction between the CLIP transform which lives in torchtune/models and the other transforms in modules/transforms/vision. Will only the model transforms be classes with `__call__` and all the transforms that live in `modules/transforms/` be normal functions?

Also, what is the expected approach for a user to make a new model transform? I had originally thought there would be common image transforms that they can just `Compose` together in the model builder, but I see that CLIPImageTransform is a bit more involved than just piping transforms. So is this the overall user flow we want to enforce:
- all general transforms are standalone functions in modules
- to make a new model transform, user needs to define a class that uses the general transform functions in any way they want
- then they create a builder function that parametrizes the class and can be instantiated in the config

torchtune/models/clip/_transforms.py
felipemello1: This is used for positional embeddings in the vision encoder: https://github.com/felipemello1/torchtune/blob/9070944e1199f047bac86e3325695bd564d35000/torchtune/models/clip/_position_embeddings.py#L153

torchtune/models/clip/_transforms.py
felipemello1: correct! added it for the linter

torchtune/models/clip/_transforms.py
felipemello1: since this is calculated on the fly, it seemed reasonable to let the user know what are the options being considered, as a sanity check

RdoubleA (approved):
No major concerns on my end - thanks for pushing this through. You might have to wait until recipe tests is fixed on main before landing this. 

Also curious if we plan to add model transforms to the api_ref_models.rst, cc @pbontrager . But this can be a follow-up

torchtune/models/clip/_transforms.py
```diff
@@ -0,0 +1,174 @@
...
+        resize_to_max_canvas: bool = False,
+    ) -> None:
+
+        # get_canvas_best_fit
+        assert (
+            possible_resolutions is not None or max_num_tiles is not None
+        ), f"Either possible_resolutions or max_num_tiles must be given. Got {possible_resolutions=} and {max_num_tiles=}"
+
+        # If possible_resolutions are not given, then calculate possible ones based on max_num_tiles
+        if not possible_resolutions and max_num_tiles:
```
pbontrager, line 109: You have to explicitly check for None, otherwise 0 or empty tuples resolve as false

torchtune/models/clip/_transforms.py
pbontrager: This else doesn't make sense to me

pbontrager (approved):
Everything looks good here. I'll approve this, but I think you should update the check for possible_resolutions to explicitly check for None.

torchtune/models/clip/_transforms.py
felipemello1: If possible_resolutions is None or is empty or anything that is not a list with items, we must activate this condition to find possible_resolutions

felipemello1: image

PR #1085, opened 2024-06-13
[doc] update to the models documentation

calvinpelletier:
#### Context
- [ ] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

#### Changelog

- add missing models to the docs
- various other minor changes

#### Test plan
`make html` and examined the results

### DIFF
--- a/docs/source/api_ref_models.rst
+++ b/docs/source/api_ref_models.rst
@@ -26,6 +26,7 @@ All models from the `Llama3 family <https://llama.meta.com/llama3/>`_.
     llama3.llama3_70b
     llama3.lora_llama3_70b
     llama3.qlora_llama3_70b
+    llama3.llama3_tokenizer
 
 
 llama2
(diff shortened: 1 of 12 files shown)


docs/source/api_ref_models.rst
```diff
@@ -21,10 +21,11 @@ All models from the `Llama3 family <https://llama.meta.com/llama3/>`_.
     :nosignatures:
```
joecummings, line 21: We need the tokenizers in here, too.

pbontrager:
Thanks for adding all this in. One concern I have with the "models" docs page currently is that I believe the provided download commands provided are incomplete. I think if you use the command, the default download directory will not match the directory used in our configs. We could update the commands, but I think it'd be better to actually update the default folder to use the path name from huggingface hub which is the pattern we've been following.

calvinpelletier:
@pbontrager updated the `--output_dir` default of `tune download` to be `/tmp/<model_name>`

pbontrager (approved):

PR #1087, opened 2024-06-13
Fix precision + QLoRA state dict tests, DTensor init

ebsmothers:
#### Context
- [ ] add a new feature
- [X] fix a bug
- [ ] other (please add here)

#### Changelog
After #855 added usage of FSDPTest we cannot set `torch.backends.cudnn` in the same test env as an `FSDPTest`. The fix it to set `torch.backends.__allow_nonbracketed_mutation_flag` to True inside the test, similar to what was done in #855. Unfortunately we missed this test case because our unit tests that require GPUs do not currently run in CI. 

Also `test_qlora_state_dict` is failing as it requires a newer version of torchao. Current `torchao` does not have `__version__` defined, so we add the wonderful check `"__version__" not in dir(torchao)`. Separately, there is an issue where the `DTensor` API changed in a BC-breaking way. I bisected to the appropriate nightly and for now I gate based on that torch version. Once 2.4 releases we should be able to remove this check.

Oh also the way we were checking `version("torchao")` in `_register_nf4_dispatch_ops.py` doesn't really account for the possibility of having `torchao-nightly` installed (increasingly prevalent lately..). So I updated that bit of code to account for possible nightly torchao install. Once torchao has a version and we are on 0.3 all of this can go.

Follow-ups:
- Test GPU unit tests in our CI (there are quite a few of them and rn we have no signal on these, many of which are for more bleeding-edge features)
- Define proper version-based gating across our various dependencies and apply it consistently across the repo

#### Test plan

Prior to these changes:

```
pytest tests/
...
FAILED tests/torchtune/utils/test_distributed.py::TestFullyShardState::test_qlora_state_dict - RuntimeError: Process 0 exited with error code 10 and exception:
FAILED tests/torchtune/utils/test_precision.py::TestPrecisionUtils::test_set_float32_precision - RuntimeError: not allowed to set torch.backends.cudnn flags after disable_global_flags; please use flags() context manager instead
============== 2 failed, 242 passed, 34 skipped, 2 warnings in 55.96s ======================
```

After these changes (and on a nightly version of torchao)

```
pytest tests/
...
===== 244 passed, 34 skipped, 2 warnings in 63.99s (0:01:03) ========
```

### DIFF
--- a/tests/torchtune/utils/test_precision.py
+++ b/tests/torchtune/utils/test_precision.py
@@ -68,7 +68,7 @@ class TestPrecisionUtils:
 
         _set_float32_precision("high")
         setattr(  # noqa: B010
-            torch.backends, "__allow_nonbracketed_mutation_flag", True
+            torch.backends, "__allow_nonbracketed_mutation_flag", False
         )
         assert torch.get_float32_matmul_precision() == "high"
         assert torch.backends.cudnn.allow_tf32
--- a/tests/torchtune/utils/test_precision.py
+++ b/tests/torchtune/utils/test_precision.py
@@ -58,12 +58,18 @@ class TestPrecisionUtils:
 
     @pytest.mark.skipif(not cuda_available, reason="The test requires GPUs to run.")
     def test_set_float32_precision(self) -> None:
+        setattr(  # noqa: B010
+            torch.backends, "__allow_nonbracketed_mutation_flag", True
+        )
         _set_float32_precision("highest")
         assert torch.get_float32_matmul_precision() == "highest"
         assert not torch.backends.cudnn.allow_tf32
         assert not torch.backends.cuda.matmul.allow_tf32
 
         _set_float32_precision("high")
+        setattr(  # noqa: B010
+            torch.backends, "__allow_nonbracketed_mutation_flag", True
+        )
         assert torch.get_float32_matmul_precision() == "high"
         assert torch.backends.cudnn.allow_tf32
         assert torch.backends.cuda.matmul.allow_tf32
--- a/torchtune/utils/_distributed.py
+++ b/torchtune/utils/_distributed.py
@@ -321,6 +321,7 @@ def load_from_full_model_state_dict(
             sharded_param = full_tensor.new_zeros(chunk.size())
             sharded_param[: chunk.size(0)].copy_(chunk)
             # BC-breaking change to DTensor API in https://github.com/pytorch/pytorch/pull/128112
+            # TODO: change to from_local API (need to add view support for NF4)
             if version.parse(torch.__version__) >= version.parse("2.4.0.dev20240606"):
                 sharded_tensor = DTensor(
                     local_tensor=sharded_param,
--- a/torchtune/utils/_distributed.py
+++ b/torchtune/utils/_distributed.py
@@ -15,6 +15,7 @@ import torch.distributed as dist
 from torch import nn
 
 from torch.distributed._tensor import distribute_tensor, DTensor
+from torch.distributed._tensor.placement_types import DTensorSpec, TensorMeta
 from torch.distributed.algorithms._checkpoint.checkpoint_wrapper import (
     _CHECKPOINT_WRAPPED_MODULE,
 )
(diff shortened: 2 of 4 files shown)


tests/torchtune/utils/test_precision.py
```diff
@@ -58,12 +58,18 @@ def test_error_bf16_unsupported(self, mock_verify):
 
     @pytest.mark.skipif(not cuda_available, reason="The test requires GPUs to run.")
     def test_set_float32_precision(self) -> None:
+        setattr(  # noqa: B010
+            torch.backends, "__allow_nonbracketed_mutation_flag", True
```
pbontrager, line 62: Is this setting change contained within the test? If not, I we should grab the original value and set it back to that at the end of the test.

pbontrager (commented):
Good catch, just one comment and then good to go.

tests/torchtune/utils/test_precision.py
```diff
@@ -58,12 +58,18 @@ def test_error_bf16_unsupported(self, mock_verify):
...
+        )
         _set_float32_precision("highest")
         assert torch.get_float32_matmul_precision() == "highest"
         assert not torch.backends.cudnn.allow_tf32
         assert not torch.backends.cuda.matmul.allow_tf32
 
         _set_float32_precision("high")
+        setattr(  # noqa: B010
+            torch.backends, "__allow_nonbracketed_mutation_flag", True
```
felipemello1, line 71: is there a reason for adding "setattr" a second time and its placement here? Or doing it once at the start is enough?

torchtune/utils/_distributed.py
```diff
@@ -318,15 +320,32 @@ def load_from_full_model_state_dict(
...
-                sharded_param,
-                sharded_meta_param.device_mesh,
-                sharded_meta_param.placements,
-                shape=sharded_meta_param.size(),
-                dtype=sharded_meta_param.dtype,
-                requires_grad=sharded_meta_param.requires_grad,
-                stride=sharded_meta_param.stride(),
-            )
+            # BC-breaking change to DTensor API in https://github.com/pytorch/pytorch/pull/128112
+            if version.parse(torch.__version__) >= version.parse("2.4.0.dev20240606"):
```
ebsmothers, line 324: @wanchaol @weifengpy please let me know if this is the best way to handle this, also open to any other ideas either of you have here.

ebsmothers: Also regarding a longer-term plan here, are there any plans to make `DTensor` a public API? Or should we consider whether there's an alternative way to do this without directly calling into the `DTensor` API?

weifengpy: chatted about this. the long-term solution is to use public DTensor api

```
return DTensor.from_local(
    local_tensor,
    sharding_spec.mesh,
    sharding_spec.placements,
    shape=sharding_spec.shape,
    stride=sharding_spec.stride,
)
```

wanchaol: since this is state_dict logic, I think we should just directly use `DTensor.from_local`? `DTensor.__new__` is kinda private API of DTensor

ebsmothers: One more question here: if using with NF4, does it mean we will have to also add logic for aten.view in ao (based on [this line](https://github.com/pytorch/pytorch/blob/529bedad3e6cbd98387b9e46ce0c0e290ed12ab5/torch/distributed/_tensor/api.py#L151))?

weifengpy: good catch. i think so. do you know the args, kwargs after dispatching? if view_as(shape) matches NF4.shape, that’s trivial. otherwise we need a deeper look because the semantics of NF4.view_as(different shape)

weifengpy: btw, it does not have to be addressed in this PR, if it needs torchao change

ebsmothers: Yeah personally I don't know the args or kwargs so need to do a bit more investigation. Will leave this as a fast follow for now

joecummings (approved):
🫡🫡🫡

joecummings:
You can rebase on main now that #1095 has been merged and the workflows should pass.

torchtune/utils/_distributed.py
weifengpy: I just found pytorch from source does not work.  `torch.__version__` is `2.4.0a0+git....`. My commit hash is before DTensor BC breaking, but It goes into `DTensor(spec)` and complained cc @ebsmothers

PR #1088, opened 2024-06-13
fix torchtune import from torchao after refactor

jerryzh168:
Summary: att

Reviewed By: SLR722

Differential Revision: D58506712


pbontrager:
Thanks for this fix. Could you leave a comment specifying what that import is doing? Also it appears we're importing a lot of unused imports.

jerryzh168:
> Thanks for this fix. Could you leave a comment specifying what that import is doing? Also it appears we're importing a lot of unused imports.

sure, this is for deciding what quantizer to include, it's used in L48 in the file

pbontrager (approved):

PR #1089, opened 2024-06-13
Profiler v2

jeromeku:
@ebsmothers @RdoubleA 

Opening a new PR in place of #1070 as commit history from prior PR was corrupted after re-basing on main.

## Updates
- [x] Refactored `setup_torch_profiler`
  - this function now takes args that have been flattened, parsed from the top-level config
  - parsing of the top-level config is now handled by a `_setup_profiler` function defined in the `recipe`; see [here](https://github.com/jeromeku/torchtune/blob/c7b57ee94ea06f3e3788cf8905e5f177d9bd5290/recipes/lora_finetune_distributed.py#L283-L385) for such an example
- [x] Updated tests to reflect the modified API
- [x] Updated the following configs / recipes to demonstrate use of the updated the profiler
  - config / recipe: [llama2/7B_lora.yaml](https://github.com/jeromeku/torchtune/blob/profiler_v2/recipes/configs/llama2/7B_lora.yaml) and [lora_finetune_distributed](https://github.com/pytorch/torchtune/pull/1089/files#diff-9022ab99a2fca3e9bc0a2c07bd4262bd6c00ed9b198befc111f9565b21bb7c72)
  - config / recipe: [llama2/7B_lora_single_device](https://github.com/jeromeku/torchtune/blob/profiler_v2/recipes/configs/llama2/7B_lora_single_device.yaml) and [lora_finetune_single_device](https://github.com/jeromeku/torchtune/blob/profiler_v2/recipes/lora_finetune_single_device.py)

### Run single node `LoRA` finetune on Llama2
```
  tune run lora_finetune_single_device \
  --config llama2/7B_lora_single_device \
  max_steps_per_epoch=10 \
  profiler.enabled=True \
  profiler.schedule.wait=2 \
  profiler.schedule.warmup=2 \
  profiler.schedule.active=1 \
  profiler.schedule.repeat=0 \
  profiler.output_dir=./profile-test
```
This will profile in a 5-step cycle for 10 total steps and export the trace results to `profile-test`.  

Since there are 2 `active` steps, there will be two corresponding sub-directories within the export directory, labeled `iteration-{5,10}` respectively, each with: 1) gzipped tensorboard / chrome trace, 2) stack trace, 3) an html memory timeline.

### Run distributed `LoRA` fine-tune on on Llama2
```
tune run --nnodes=1 --nproc_per_node=2 lora_finetune_distributed \
--config llama2/7B_lora \
max_steps_per_epoch=3 \
gradient_accumulation_steps=2 \
profiler.enabled=True \
profiler.schedule.wait=1 \
profiler.schedule.warmup=1 \
profiler.schedule.active=1 \
profiler.output_dir=./profile-test
```
Should result in a single sub-directory within `profile-test` at `iteration-3` containing the exported traces labeled by rank along with the memory timeline for rank-0 (exporting multiple memory timelines results in export errors on multiple devices). 

Note that the chrome traces for the two examples will be different due to where the `profiler` is stepped within the `recipe` `train` loop.  In the former, the `profiler` is stepped [*per batch*](https://github.com/pytorch/torchtune/blob/c7b57ee94ea06f3e3788cf8905e5f177d9bd5290/recipes/lora_finetune_single_device.py#L670-L673) whereas in the latter, the `profiler` is stepped [*per optimizer step*](https://github.com/pytorch/torchtune/blob/c7b57ee94ea06f3e3788cf8905e5f177d9bd5290/recipes/lora_finetune_distributed.py#L764-L767), which will contain multiple batches when `gradient accumulation > 1`.  This nuance is included in the docstring for `setup_torch_profiler`.


recipes/lora_finetune_distributed.py
```diff
@@ -270,6 +276,114 @@ def setup(self, cfg: DictConfig) -> None:
             last_epoch=self.global_step - 1,
         )
 
+        # Set up profiler, returns FakeProfiler (nullcontext object with no-op `step` method)
+        # if cfg is missing profiler key or if `cfg.profiler.enabled = False`
+        self._profiler = self._setup_profiler(cfg.get(PROFILER_KEY, None), log_cfg=True)
+
+    def _setup_profiler(
+        self, cfg: DictConfig, log_cfg: bool = False
```
RdoubleA, line 284: nit: call this cfg_profiler, similar to the other setup methods, then it's very clear that it's the profiler field and not the full config

recipes/configs/llama2/7B_lora_single_device.yaml
```diff
@@ -80,9 +80,29 @@ device: cuda
...
+# Set enabled to False as it's only needed for debugging training
 # Show case the usage of pytorch profiler
 # Set enabled to False as it's only needed for debugging training
 profiler:
-  _component_: torchtune.utils.profiler
   enabled: False
-  output_dir: ${output_dir}/torchtune_perf_tracing.json
+
+  #Output directory of trace artifacts
+  output_dir: /tmp/profiling_outputs
```
RdoubleA, line 91: Would be great to reuse the output_dir we are already creating above so we don't make too many stray folders
```suggestion
  output_dir: ${output_dir}/profiling_outputs
```

recipes/lora_finetune_single_device.py
```diff
@@ -472,93 +581,99 @@ def train(self) -> None:
...
-                        self._optimizer.zero_grad(set_to_none=True)
-                        self._lr_scheduler.step()
-                        # Update the number of steps when the weights are updated
-                        self.global_step += 1
-
-                        loss_to_log = running_loss.item()
-                        pbar.update(1)
-                        pbar.set_description(
-                            f"{curr_epoch+1}|{self.global_step}|Loss: {loss_to_log}"
+        with self._profiler as prof:
```
RdoubleA, line 584: Why is this used as a context manager and in the other recipe we use start and stop, is that purposeful to show different ways to use it?

recipes/lora_finetune_single_device.py
```diff
@@ -472,93 +581,99 @@ def train(self) -> None:
...
+            # self.epochs_run should be non-zero when we're resuming from a checkpoint
+            for curr_epoch in range(self.epochs_run, self.total_epochs):
+                # Update the sampler to ensure data is correctly shuffled across epochs
+                # in case shuffle is True
+                self._sampler.set_epoch(curr_epoch)
+
+                # Optionally profile the training loop
+                with self._profiler:
```
RdoubleA, line 592: We don't need to define profiler context twice, right?

torchtune/utils/_profiler.py
```diff
@@ -48,3 +61,405 @@ def trace_handler(prof) -> None:
...
+        pass
+
+    def stop(self):
+        pass
+
+    def step(self):
+        pass
+
+
+def should_profile(cfg: DictConfig) -> bool:
```
RdoubleA, line 189: Why not just check cfg.profiler.enabled directly and remove this util? This would reduce one level of indirection

torchtune/utils/_profiler.py
```diff
@@ -48,3 +61,405 @@ def trace_handler(prof) -> None:
...
+    return (cfg is not None and len(cfg) > 0) and cfg.get("enabled", True)
+
+
+# NOTE: This is a reference implementation of a profiler setup method to be defined within a `recipe`.
```
RdoubleA, line 193: I understand the need of keeping a reference implementation for _setup_profiler, but imo the ones in the recipes should serve as the references, including this here and saying it should not be used can be confusing for users.

torchtune/utils/_profiler.py
```diff
@@ -6,28 +6,41 @@
...
+from torch._C._profiler import _ExperimentalConfig
+from torch.profiler import profile, tensorboard_trace_handler
+from torchtune.utils import get_world_size_and_rank
+
+from torchtune.utils.logging import get_logger
+
+log = get_logger("INFO")
 
 
 def profiler(
```
RdoubleA, line 28: It is not clear to me the relationship between this method and setup_torch_profiler, is this now deprecated? If so, let's just remove it entirely

torchtune/utils/_profiler.py
```diff
@@ -48,3 +61,405 @@ def trace_handler(prof) -> None:
...
+    if rank == 0:
+        log.warn(msg)
+
+
+def trace_handler(
+    prof: torch.profiler.profiler.profile,
+    output_dir,
+    metric="self_cuda_time_total",
+    row_limit=25,
+):
```
RdoubleA, line 100: Do you mind adding a docstring for this? :)

torchtune/utils/_profiler.py
```diff
@@ -48,3 +61,405 @@ def trace_handler(prof) -> None:
...
+
+    Additional notes:
+        - the profiler schedule updates with respect to an optimizer step:
+            - e.g., if `gradient_accumulation = 2`, then the profiler will step every 2 batches.
+        - sensible defaults will be chosen if the config is missing options
+            - if no activities are specified, profiler will default to CPU + CUDA
+            - if no schedule is specified, profiler will default to wait 10, warmup 5, active 3, repeat 1
+            - if a schedule is specified, profiler will validate that the schedule is valid and can be passed to `instantiate`
+            - certain options will be overridden (`with_stack` and `record_shapes`) depending on requirements of other options
+                - e.g., `profile_memory` requires `with_stack` and `record_shapes`
```
RdoubleA, line 371: Thank you for listing out potential overrides, this is awesome 👏

torchtune/utils/_profiler.py
```diff
@@ -48,3 +61,405 @@ def trace_handler(prof) -> None:
...
+    if output_dir is None:
+        _warn(
+            f" No output directory found in profiler config, defaulting to {_DEFAULT_PROFILE_DIR}"
+        )
+        output_dir = _DEFAULT_PROFILE_DIR
+
+    output_dir = Path(output_dir)
+    output_dir.mkdir(parents=True, exist_ok=True)
+    output_dir = str(output_dir)
+    callback = partial(trace_handler, output_dir=output_dir)
```
RdoubleA, line 433: A comment here for what the trace_handler is doing would be helpful

torchtune/utils/_profiler.py
```diff
@@ -48,3 +61,405 @@ def trace_handler(prof) -> None:
...
+        activities=activities,
+        profile_memory=profile_memory,
+        with_stack=with_stack,
+        record_shapes=record_shapes,
+        with_flops=with_flops,
+        schedule=schedule,
+        experimental_config=experimental_config,
+        on_trace_ready=callback,
+    )
+    if return_cfg:
```
RdoubleA, line 445: How about just return config by default and let recipe decide if it wants to log it?

recipes/lora_finetune_distributed.py
jeromeku: Fixed

recipes/configs/llama2/7B_lora_single_device.yaml
jeromeku: Done

recipes/lora_finetune_single_device.py
jeromeku: Yes -- idea is to showcase use as object and as a context manager.

recipes/lora_finetune_single_device.py
jeromeku: yeah - this was a mistake.  Removed second context manager.

torchtune/utils/_profiler.py
jeromeku: Done

torchtune/utils/_profiler.py
jeromeku: Removed

torchtune/utils/_profiler.py
jeromeku: Removed

torchtune/utils/_profiler.py
jeromeku: Done

torchtune/utils/_profiler.py
jeromeku: Done

torchtune/utils/_profiler.py
jeromeku: Done

jeromeku:
@RdoubleA 

Fixed all requests.

recipes/lora_finetune_distributed.py
```diff
@@ -270,6 +275,118 @@ def setup(self, cfg: DictConfig) -> None:
...
+            schedule:
+                wait: int
+                warmup: int
+                active: int
+                repeat: int
+        ```
+        """
+        # Check whether `profiler` key is present in the config and that it is not empty;
+        # if it is present check that `enabled = True`
+        if (cfg_profiler is not None and len(cfg_profiler) > 0) and cfg_profiler.get(
```
RdoubleA, line 330: This should be able to be simplified to simply `enabled = cfg_profiler.get("enabled", True), no? Since cfg_profiler is a required argument and will never be None
```suggestion
        if not cfg_profiler.get("enabled", True):
            if self._is_rank_zero:
                log.info(" Profiling disabled.")
            return FakeProfiler()
```

RdoubleA (commented):
I think overall this looks really clean to me, I'll go ahead and run the CI. Tagging @ebsmothers and @rohan-varma for any additional thoughts here

recipes/lora_finetune_distributed.py
jeromeku: I was accounting for the case when `cfg` is missing the `profiler` section altogether.

RdoubleA: ooh ya I see, maybe the type hint for the function needs to be updated instead?

jeromeku: fixed

torchtune/utils/_profiler.py
```diff
@@ -4,47 +4,350 @@
...
+
+    # Set up profiler schedule
+    use_default_schedule = not any([wait, warmup, active, repeat])
+
+    # Use default schedule if None, else validate that schedule is valid and can be passed to `instantiate`
+    if use_default_schedule:
+        wait = _DEFAULT_SCHEDULE["wait"]
+        warmup = _DEFAULT_SCHEDULE["warmup"]
+        active = _DEFAULT_SCHEDULE["active"]
+        repeat = _DEFAULT_SCHEDULE["repeat"]
```
ebsmothers, line 277: Any reason we handle the defaults for schedule differently than for other configurable fields? We could just fold the values from `_DEFAULT_SCHEDULE` into the corresponding default values in `setup_torch_profiler`. That way it'd be consistent with how we handle the other fields

torchtune/utils/_profiler.py
```diff
@@ -4,47 +4,350 @@
...
+        profile_memory=profile_memory,
+        with_stack=with_stack,
+        record_shapes=record_shapes,
+        with_flops=with_flops,
+        schedule=schedule,
+        experimental_config=experimental_config,
+        on_trace_ready=callback,
+    )
+
+    profiler_cfg = DictConfig(
```
ebsmothers, line 334: Sorry I actually disagree with @RdoubleA's [previous comment](https://github.com/pytorch/torchtune/pull/1089#discussion_r1639934356) here. I think it's better to keep the return type of `setup_torch_profiler` simple, I don't expect to get back a Tuple when calling `setup_torch_profiler`, I expect to just get back a profiler. Anyways, don't want to cause too much thrash on this so don't worry about changing it back for now; I think that's something we can sort out separately. But do lmk if you have any thoughts on this

recipes/lora_finetune_distributed.py
```diff
@@ -270,6 +275,118 @@ def setup(self, cfg: DictConfig) -> None:
...
+            #`torch.profiler.schedule` args
+            schedule:
+                wait: int
+                warmup: int
+                active: int
+                repeat: int
+        ```
+        """
+        # Check whether `profiler` key is present in the config and that it is not empty;
+        # if it is present check that `enabled = True`
```
ebsmothers, line 329: super nit: should it be this? Since if `profiler` key is present and `enabled` is missing we infer True
```suggestion
        # if it is present check that `enabled != False`
```

torchtune/utils/_profiler.py
```diff
@@ -4,47 +4,350 @@
...
+    "active": 2,
+    "repeat": 1,
+}
+
+DEFAULT_TRACE_OPTS: dict = {
+    "profile_memory": False,
+    "with_stack": False,
+    "record_shapes": True,
+    "with_flops": False,
+}
```
ebsmothers, line 44: I notice you use `DEFAULT_TRACE_OPTS` in the recipe to set defaults but not `_DEFAULT_SCHEDULE`. Is this just because it's more natural to use bools for default values than ints? My question is kinda similar to my other comment below.. ideally we can handle this consistently across the default values (I don't want to have half defined in the recipe and half defined in here).

recipes/lora_finetune_distributed.py
```diff
@@ -270,6 +275,118 @@ def setup(self, cfg: DictConfig) -> None:
...
+        if schedule_cfg is None:
+            wait = None
+            warmup = None
+            active = None
+            repeat = None
+        else:
+            wait = schedule_cfg.get("wait", None)
+            warmup = schedule_cfg.get("warmup", None)
+            active = schedule_cfg.get("active", None)
+            repeat = schedule_cfg.get("repeat", None)
```
ebsmothers, line 364: Do these fields have to be nested under schedule_cfg? (I get that it's natural for a Hydra-like instantiate API, but if we are just passing as flat args anyways idk how much it buys us)

recipes/lora_finetune_distributed.py
```diff
@@ -563,6 +680,7 @@ def train(self) -> None:
         running_loss = 0
         num_tokens = 0
 
+        self._profiler.start()
```
ebsmothers, line 683: It's really nice how cleanly this fits into our training loop. This is exactly what I was hoping for

ebsmothers (commented):
This is really shaping up nicely! A few more comments on how we expose defaults and consistency, but really love how non-invasive this is in the training loop itself.

jeromeku:
@ebsmothers 

Surfaced all default `profiler` settings to `recipe` per your suggestion.

torchtune/utils/_profiler.py
RdoubleA: How about handle logging the config here instead of in the recipe? In fact you can just pretty print a regular duct instead of a DictConfig object. Then you don't need to return a Tuple.

torchtune/utils/_profiler.py
```diff
@@ -4,47 +4,352 @@
...
+    )
+
+    # Use default schedule if None, else validate that schedule is valid and can be passed to `instantiate`
+    if use_default_schedule:
+        wait = DEFAULT_SCHEDULE["wait"]
+        warmup = DEFAULT_SCHEDULE["warmup"]
+        active = DEFAULT_SCHEDULE["active"]
+        repeat = DEFAULT_SCHEDULE["repeat"]
+        _warn(
+            " No schedule found in config, defaulting to wait {wait}, warmup {warmup}, active {active}, repeat {repeat}"
```
RdoubleA, line 281: Need an f-string here?

recipes/lora_finetune_distributed.py
```diff
@@ -270,6 +278,119 @@ def setup(self, cfg: DictConfig) -> None:
...
+
+        # Parse schedule specific args
+        schedule_cfg = cfg_profiler.get("schedule", None)
+
+        if schedule_cfg is None:
+            wait = DEFAULT_SCHEDULE["wait"]
+            warmup = DEFAULT_SCHEDULE["warmup"]
+            active = DEFAULT_SCHEDULE["active"]
+            repeat = DEFAULT_SCHEDULE["repeat"]
+        else:
```
RdoubleA, line 364: I'm a bit confused here, if the user does not specify the schedule the default will get used, but if they have a schedule field but the parameters are empty then wait, etc will be set to None. If the profiler is provided in the config and is enabled, don't we want to default schedule to always be the default unless overridden? Or am I misunderstanding

jeromeku: fixed

jeromeku:
@RdoubleA @ebsmothers 

Simplified profiler setup and schedule defaults per your comments.

torchtune/utils/_profiler.py
jeromeku: fixed

recipes/lora_finetune_distributed.py
jeromeku: fixed

torchtune/utils/_profiler.py
jeromeku: fixed

recipes/lora_finetune_distributed.py
```diff
@@ -270,6 +278,122 @@ def setup(self, cfg: DictConfig) -> None:
...
+                repeat: int
+        ```
+        """
+
+        enabled = cfg_profiler is not None and cfg_profiler.get("enabled", True)
+
+        if not enabled:
+            if self._is_rank_zero:
+                log.info(" Profiling disabled.")
+            return FakeProfiler()
```
RdoubleA, line 337: very clean 👌

recipes/lora_finetune_distributed.py
```diff
@@ -270,6 +278,122 @@ def setup(self, cfg: DictConfig) -> None:
...
+        # Parse schedule specific args
+        schedule_cfg = cfg_profiler.get("schedule", None)
+
+        # Check for schedule
+        # 1) If no schedule is provided, set to DEFAULT_SCHEDULE
+        # 2) else check for missing keys and warn if any are missing, setting these to defaults
+        if schedule_cfg is None:
+            schedule_cfg = DEFAULT_SCHEDULE
+        else:
+            schedule_cfg = {
```
RdoubleA, line 362: nit: I think this can be simplified further:
```
for k in DEFAULT_SCHEDULE.keys():
    if k not in schedule_cfg:
        schedule_cfg[k] = DEFAULT_SCHEDULE[k]
        log.warning(f"Missing key {k} in schedule config, defaulting to {k} = {schedule_cfg[k]}")
```

recipes/lora_finetune_distributed.py
```diff
@@ -270,6 +278,122 @@ def setup(self, cfg: DictConfig) -> None:
...
+        with_flops = cfg_profiler.get("with_flops", DEFAULT_TRACE_OPTS["with_flops"])
+        output_dir = cfg_profiler.get("output_dir", DEFAULT_PROFILE_DIR)
+
+        # Parse schedule specific args
+        schedule_cfg = cfg_profiler.get("schedule", None)
+
+        # Check for schedule
+        # 1) If no schedule is provided, set to DEFAULT_SCHEDULE
+        # 2) else check for missing keys and warn if any are missing, setting these to defaults
+        if schedule_cfg is None:
```
RdoubleA, line 359: nit: could just change the above line and remove this if-else logic
```
schedule_cfg = cfg_profiler.get("schedule", DEFAULT_SCHEDULE)
```
if you take my suggestion below as well, if schedule_cfg == DEFAULT_SCHEDULE then nothing will be replaced and it will be a no-op

RdoubleA (approved):
Thanks for pushing this through, couple of nit comments but nothing blocking and can be done as follow-ups. There's a few checks that are failing on CI, lmk if you have issues fixing those. You can fix the lint by running `pre-commit run --all-files` locally

RdoubleA:
@jeromeku went ahead and fixed the doc build since sphinx is a pain to deal with, though some bullets are not rendering correctly. mind taking a look? @joecummings

recipes/lora_finetune_distributed.py
```diff
@@ -270,6 +278,124 @@ def setup(self, cfg: DictConfig) -> None:
...
+        """
+
+        enabled = cfg_profiler is not None and cfg_profiler.get("enabled", True)
+
+        if not enabled:
+            if self._is_rank_zero:
+                log.info(" Profiling disabled.")
+            return FakeProfiler()
+
+        # Set up profiler activities
```
pbontrager, line 341: I believe all of these configs could be moved into the `setup_torch_profiler` util as optional args and then setup_torch_profiler could be called directly with config.instantiate

recipes/configs/llama2/7B_lora_single_device.yaml
```diff
@@ -83,6 +83,24 @@ enable_activation_checkpointing: True
...
+  CPU: True
+  CUDA: True
+
+  #trace options passed to `torch.profiler.profile`
+  profile_memory: False
+  with_stack: False
+  record_shapes: True
+  with_flops: False
+
+  `torch.profiler.schedule` options
```
pbontrager, line 101: Missing a #

torchtune/utils/_profiler.py
```diff
@@ -4,47 +4,358 @@
...
+
+def setup_torch_profiler(
+    enabled: bool = False,
+    cpu: bool = True,
+    cuda: bool = True,
+    profile_memory: bool = False,
+    with_stack: bool = False,
+    record_shapes: bool = False,
+    with_flops: bool = False,
+    wait: Optional[int] = None,
```
pbontrager, line 186: If we altered this to support `config.initialize(setup_torch_profiler)` from the recipe, then all of the schedule args could be renamed to "schedule_warmup", "schedule_active", etc allowing for a flat list of profiler configs. All of these config options that are None would be optional by default in the config and if they get a None value, they can be set to the default value in the init.

torchtune/utils/_profiler.py
```diff
@@ -4,47 +4,358 @@
...
+    if rank == 0:
+        log.info(f"Saving profiling results to {curr_trace_dir}")
+
+    # TODO: Is this necessary?
+    # see https://github.com/pytorch/torchtitan/blob/3050098dcee4901d88c712f9e8e9703d1735a29b/torchtitan/profiling.py#L48
+    if world_size > 1:
+        torch.distributed.barrier()
+
+
+class FakeProfiler:
```
pbontrager, line 138: nit: I think we use Dummy throughout the library instead of Fake

recipes/lora_finetune_distributed.py
```diff
@@ -270,6 +278,124 @@ def setup(self, cfg: DictConfig) -> None:
...
+                    wait: int
+                    warmup: int
+                    active: int
+                    repeat: int
+
+        """
+
+        enabled = cfg_profiler is not None and cfg_profiler.get("enabled", True)
+
+        if not enabled:
```
pbontrager, line 336: The logs can be moved into the util as well. I think we'd want the util to take "enabled" as a kwarg as well so we can  have it output the FakeProfiler if it's disabled or on the wrong rank.

pbontrager (changes requested):
Sorry for the late comments. This is a great addition and there has already been a lot of discussion on this PR, but I think it's important that we limit the amount of code this adds to the recipes as it adds a lot of boilerplate for the user. If we make our profiler util handle all of it's optional kwargs on init, then we should be able to directly utilize config.initialize. @RdoubleA does this match with your understanding too?

torchtune/utils/_profiler.py
jeromeku: What you're describing is more in-line with my original PR, where most of the code in `_setup_profiler` was abstracted away from the user in `setup_torch_profiler`.  

Per the comments from @RdoubleA and @ebsmothers, I refactored the original `setup_torch_profiler` such that most of the setup was extracted from `setup_torch_profiler` to the `recipe`, which admittedly results in more code duplication and config parsing on the part of the recipe writer / user. 

Not sure what direction to take at this point.

pbontrager: I am sorry for the confusion. In the original version, there was a lot of config parsing happening inside of the util, which we don't want for a utility since our principle is we want each function to work as code and to not assume that a config will be involved. So I agree with all the changes to to remove the config parsing from the util. But I believe that we can still get that same simplicity that you originally proposed by defining all the arguments inside of the __init__ instead of taking in a Dict. Something like this

```python
def setup_torch_profiler(
        self,
        *,
        enabled: bool = True
        output_dir: str = "profiler_output",
        #`torch.profiler.ProfilerActivity` types to trace
        activity_types: List[str] = ACTIVITY_TYPES.keys(),
        #trace options passed to `torch.profiler.profile`
        profile_memory: bool =  False
  	with_stack: bool =  False
  	record_shapes: bool = True
  	with_flops: bool =  True
  	# `torch.profiler.schedule` options
	schedule_wait: int = 5
	schedule_warmup: int = 5
        schedule_active: int = 2
	schedule_repeat: int = 1
        log_cfg: bool = False
) -> ContextManager:
```

Then `config.initialize(setup_torch_profiler)` could initialize it from the config automatically and setup_torch_profiler works as a stand alone function. Does this make sense? I'll tag @RdoubleA and @ebsmothers here to make sure I'm not contradicting them, because I don't want you to go back and forth on work.

ebsmothers: Yeah sorry for the thrash here @jeromeku. I think @pbontrager's last comment here generally makes sense to me. Basically we can define reasonable defaults for all params inside of `setup_torch_profiler`. That way we don't have to go through the work of parsing everything and setting defaults inside the recipe. Then the config would look like

```
profiler:
  _component_: torchtune.utils._profiler.setup_torch_profiler
  enabled: True
  cpu: True
  cuda: True
  ...
```

and in the recipe you can just call `config.instantiate(cfg.profiler)` in the recipe's `_setup_profiler` method. You then only have to check that the profiler field is defined, but by putting defaults in the `setup_torch_profiler` utility you don't really have to do a bunch of sanity checks on missing fields in the config.

jeromeku:
@RdoubleA @pbontrager @ebsmothers 
Refactored `_setup_profiler` in `recipes` to directly instantiate `torchtune.utils.setup_torch_profiler`.

torchtune/utils/_profiler.py
pbontrager: Yes to everything Evan said with the small change, that we wouldn't need a _setup_profiler method then and could just put it directly in the setup method.

pbontrager (approved):
Thank you for the quick turn around! I think this is looking good now, thank you for this great addition! (It should be good to go after the tests to pass)

jeromeku:
@RdoubleA @ebsmothers @pbontrager 

Any tips on how to fix the `docs` and `recipe` check failures?

ebsmothers:
@jeromeku for the recipe failure I think it's because `|` is only available for typing from Python 3.10 onwards, and our CI jobs also run on 3.8 and 3.9. So for that maybe just switch to `Union`. Btw if you do have access to one or more GPUs and wanna run locally to get quicker signal you can run `pytest tests -m integration_test` to run the recipe tests (maybe you already did this and are just on python >= 3.10).

For the doc signal I am not sure but the triple backtick on L220 looks suspicious to me. Similarly there you can do a local build of the docs by following [these instructions](https://github.com/pytorch/torchtune/blob/ef6e196d8e47e9bc584bc9f7ce836f646443381f/CONTRIBUTING.md#building-docs). (Admittedly some of the sphinx build errors can be quite annoying, if you are really banging your head against the wall here let me know and one of us can take a look at it.)

Also I guess linter signal is failing, just make sure to add newlines at the end of the config files for that one.

jeromeku:
@ebsmothers 

Thanks.  

Linting is easy enough to fix (mistake on my part for not running pre-commit hooks on config files).

Regarding recipe tests, I'm getting failures on tests on recipes / configs that were not affected by changes in this PR:
```
FAILED tests/recipes/test_eleuther_eval.py::TestEleutherEval::test_torchtune_checkpoint_eval_results
FAILED tests/recipes/test_lora_finetune_single_device.py::TestLoRAFinetuneSingleDeviceRecipe::test_loss[llama3/8B_lora_single_device-llama3-tune-True]
FAILED tests/recipes/test_lora_finetune_single_device.py::TestLoRAFinetuneSingleDeviceRecipe::test_loss[llama3/8B_lora_single_device-llama3-tune-False]
FAILED tests/recipes/test_lora_finetune_single_device.py::TestLoRAFinetuneSingleDeviceRecipe::test_loss_qlora[True-fp32]
FAILED tests/recipes/test_lora_finetune_single_device.py::TestLoRAFinetuneSingleDeviceRecipe::test_loss_qlora[True-bf16]
FAILED tests/recipes/test_lora_finetune_single_device.py::TestLoRAFinetuneSingleDeviceRecipe::test_loss_qlora[False-fp32]
FAILED tests/recipes/test_lora_finetune_single_device.py::TestLoRAFinetuneSingleDeviceRecipe::test_loss_qlora[False-bf16]
```

Also, any config that uses the previous `profiler` config will no longer be valid, since the `_component_` should now be `torchtune.utils.setup_torch_profiler` instead of `torchtune.utils.profiler` (shouldn't affect the tests, I don't think).  

Should I go ahead and change these in all config files with a `profiler` section (I only changed the ones for `llama2/7B_lora` and `llama2/7B_lora_single_device` as a demo).  I suppose any of the `recipes` which used the previous API will need to be changed as well.

ebsmothers:
> Also, any config that uses the previous profiler config will no longer be valid, since the _component_ should now be torchtune.utils.setup_torch_profiler instead of torchtune.utils.profiler (shouldn't affect the tests, I don't think).
>Should I go ahead and change these in all config files with a profiler section (I only changed the ones for llama2/7B_lora and llama2/7B_lora_single_device as a demo). I suppose any of the recipes which used the previous API will need to be changed as well.

@jeromeku good point, this will be BC-breaking for our LoRA recipes. Not the end of the world as we are actually planning to do a release soon (so now is actually the perfect time to get this in). I think updating any other configs containing profiler usage to point to the new API would be a good idea. I don't think any other recipes in the core repo use the profiler API though (but user recipes will need to change when they pull latest changes).

Re the failing tests, do you have a specific error message for any of them? I agree it's not caused by your changes so not a huge concern for you, but we do wanna get CI green to land. I can also take a pass on my end to repro (fwiw some tests run only locally and not in CI depending on which versions you have available, so you may see some test failures on our more bleeding-edge features locally)

jeromeku:
@ebsmothers 

Updated all configs with a `profiler` entry with the new profiler API.  Also, all integrated tests are now passing (except for `lm_eval` since I don't have that installed).

ebsmothers:
@jeromeku thanks, this is looking good! I am gonna run a couple sanity checks on my end and while waiting for GPU tests to run. After that I think this should be good to merge

ebsmothers (approved):

PR #1094, opened 2024-06-16
Fix the Gemma generation

solitude-alive:
Fix the problem in generation.
Aligned with https://github.com/pytorch/torchtune/blob/main/torchtune/modules/transformer.py .

#### Context
- [ ] add a new feature
- [x] fix a bug
- [ ] other (please add here)

#### Changelog

- Change `mask = self.causal_mask[None, None, input_pos]` to `mask = self.causal_mask[None, input_pos]`.
Make `GemmaTransformerDecoder` aligned with `TransformerDecoder` in [TransformerDecoder.py](https://github.com/pytorch/torchtune/blob/main/torchtune/modules/transformer.py) .

#### Test plan

### DIFF
--- a/torchtune/models/gemma/transformer.py
+++ b/torchtune/models/gemma/transformer.py
@@ -25,7 +25,7 @@ class GemmaTransformerDecoder(nn.Module):
 
     Args:
         tok_embeddings (nn.Embedding): PyTorch embedding layer, to be used to move
-            tokens to an embedding space and as the output projectionnv.
+            tokens to an embedding space and as the output projection.
         layer (TransformerDecoderLayer): Transformer Decoder layer.
         num_layers (int): Number of Transformer Decoder layers.
         max_seq_len (int): maximum sequence length the model will be run with, as used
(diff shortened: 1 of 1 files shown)


solitude-alive:
Using `tune generate` does not seem to cause any problems, but if I reproduce and rewrite a generation method, the `mask` will have an extra dimension at that point. After the modification, the problem will not occur.
Actually I'm not sure if this repository needs this modification, if you have any questions, please let me know.

RdoubleA (approved):
You're right that this should match the core TransformerDecoder module, since Gemma's is using the shared TransformerDecoderLayer and CausalSelfAttention. We enforced that the mask be a certain shape after the sample packing updates.

Actually while you're here, do you mind updating the docstrings as well to match TransformerDecoder? I'll approve and run CI in the meantime.

solitude-alive:
Yeah, I Update the docstrings.

solitude-alive:
> Yeah, I Update the docstrings.

Sorry, there are some problems with docstring. I will modify it soon.

solitude-alive (commented):
Fix the issue.

joecummings:
Failure is related to https://github.com/tensorflow/tensorboard/issues/6869, not anything wrong with this PR. Investigating now

joecummings:
@solitude-alive Can you rebase on main? I just merged #1095, which should allow the workflows here to pass.

solitude-alive:
Yeah, I merged main manually, is this work?

PR #1105, opened 2024-06-20
[doc] Add QAT tutorial

andrewor14:
Screenshot 2024-06-20 at 6 30 42 PM


andrewor14:
Note: This should be landed after https://github.com/pytorch/torchtune/pull/980, which is currently blocked on the torchao 0.3 release.

docs/source/tutorials/qat_finetune.rst
```diff
@@ -0,0 +1,371 @@
...
+to actual quantize operations after training, thereby producing a quantized model.
+Between these two steps, training can proceed exactly as before.
+
+.. image:: /_static/img/qat_diagram.png
+
+
+Applying QAT to Llama3 models
+-----------------------------
+
+We can easily apply the above QAT transformations to Llama3 in torchtune for tine-tuning:
```
ebsmothers, line 63: nit
```suggestion
We can easily apply the above QAT transformations to Llama3 in torchtune for fine-tuning:
```

docs/source/tutorials/qat_finetune.rst
```diff
@@ -0,0 +1,371 @@
...
+  x_fq = (x_fq - zp) * scale
+
+QAT typically involves applying a transformation to your model before and after training.
+For example, in the `torchao QAT implementation <https://github.com/pytorch/ao/blob/v0.2.0/torchao/quantization/prototype/qat.py>`_,
+these are represented as the prepare and convert steps: (1) prepare inserts fake quantize
+operations into linear layers, and (2) convert transforms the fake quantize operations
+to actual quantize operations after training, thereby producing a quantized model.
+Between these two steps, training can proceed exactly as before.
+
+.. image:: /_static/img/qat_diagram.png
```
ebsmothers, line 57: Nice diagram! One comment -- as a noob reader it's not immediately clear from the preceding text why the converted version does dequant on Wq (maybe this is obvious for someone familiar with quantization algos but idk how many readers of this tutorial that'll apply to). Might be good to add a sentence explaining forward logic in the quantized model.

docs/source/tutorials/qat_finetune.rst
```diff
@@ -0,0 +1,371 @@
...
+    name: en
+    split: train
+  seed: null
+  shuffle: True
+
+  ...
+
+  epochs: 1
+  max_steps_per_epoch: 2000
+  fake_quant_after_n_steps: 1000
```
ebsmothers, line 164: Might be good to briefly explain the idea behind this config

docs/source/tutorials/qat_finetune.rst
```diff
@@ -0,0 +1,371 @@
...
+  memory_efficient_fsdp_wrap: False
+
+.. note::
+
+  QAT in torchtune is currently not compatible with :code:`memory_efficient_fsdp_wrap`. This is a known issue and will be fixed in a future torchtune version.
+
+You can then use the following command to run fine-tuning with QAT using the above
+config. This workload requires at least 6 GPUs, each with VRAM of at least 80GB.
```
ebsmothers, line 172: Are there other techniques we can use to get below this? I know you mention below that `memory_efficient_fsdp_wrap` isn't supported, but what about low-precision optimizers or fusing optimizer and backward? If we can use these (or other memory-saving techniques) it might be a nice way to make the recipe accessible to a broader set of users. 

I think you mentioned previously that QAT typically uses about 30% more memory than a regular full-finetune, right? Might be worth at least briefly mentioning that the params need to get copied and so the expected memory usage for QAT will be higher than a regular full fine-tune (just to preempt any possible questions on that front).

docs/source/tutorials/qat_finetune.rst
```diff
@@ -0,0 +1,371 @@
...
+  ...
+
+  epochs: 1
+  max_steps_per_epoch: 2000
+  fake_quant_after_n_steps: 1000
+  memory_efficient_fsdp_wrap: False
+
+.. note::
+
+  QAT in torchtune is currently not compatible with :code:`memory_efficient_fsdp_wrap`. This is a known issue and will be fixed in a future torchtune version.
```
ebsmothers, line 169: This was the hang issue, right? Would be interested to see if we still see it running on FSDP2

docs/source/tutorials/qat_finetune.rst
```diff
@@ -0,0 +1,371 @@
...
+  quantizer:
+    _component_: torchtune.utils.quantization.Int8DynActInt4WeightQuantizer
+    groupsize: 256
+
+.. note::
+
+  Since we are passing in a quantized model, be sure to use the corresponding
+  post-training quantizer instead of the QAT quantizer. For example, if you
+  used the :code:`Int8DynActInt4WeightQATQuantizer` during fine-tuning, you
+  should specify :code:`Int8DynActInt4WeightQuantizer` in this step.
```
ebsmothers, line 274: Do we want to provide a reference somewhere to the currently-supported quantization methods?

docs/source/tutorials/qat_finetune.rst
andrewor14: Yeah this diagram contains some implementation details. I simplified it a bit and added a sentence to explain

docs/source/tutorials/qat_finetune.rst
andrewor14: Thanks, added a note on QAT overheads

docs/source/tutorials/qat_finetune.rst
andrewor14: added a link

ebsmothers (approved):

Issue #1111, opened 2024-06-23
ImportError: cannot import name 'packaging' from 'pkg_resources'


JadarTheObscurity:
### Environment:
- Ubuntu 24.04
- Python 3.12.3
- setuptools 70.1.0

### Problem:
Running `tune` throws the following error
```
Traceback (most recent call last):                                                                                                                                                                                  File "/home/ubuntu/llama3/venv/bin/tune", line 5, in <module>                                                                                                                                                       from torchtune._cli.tune import main                                                                                                                                                                            File "/home/ubuntu/llama3/venv/lib/python3.12/site-packages/torchtune/__init__.py", line 9, in <module>                                                                                                             from torchtune import datasets, models, modules, utils                                                                                                                                                          File "/home/ubuntu/llama3/venv/lib/python3.12/site-packages/torchtune/datasets/__init__.py", line 7, in <module>
    from torchtune.datasets._alpaca import alpaca_cleaned_dataset, alpaca_dataset
  File "/home/ubuntu/llama3/venv/lib/python3.12/site-packages/torchtune/datasets/_alpaca.py", line 10, in <module>
    from torchtune.datasets._instruct import InstructDataset
  File "/home/ubuntu/llama3/venv/lib/python3.12/site-packages/torchtune/datasets/_instruct.py", line 12, in <module>
    from torchtune.config._utils import _get_instruct_template
  File "/home/ubuntu/llama3/venv/lib/python3.12/site-packages/torchtune/config/__init__.py", line 7, in <module>
    from ._instantiate import instantiate
  File "/home/ubuntu/llama3/venv/lib/python3.12/site-packages/torchtune/config/_instantiate.py", line 12, in <module>
    from torchtune.config._utils import _get_component_from_path, _has_component
  File "/home/ubuntu/llama3/venv/lib/python3.12/site-packages/torchtune/config/_utils.py", line 16, in <module>
    from torchtune.utils import get_logger, get_world_size_and_rank
  File "/home/ubuntu/llama3/venv/lib/python3.12/site-packages/torchtune/utils/__init__.py", line 49, in <module>
    from .precision import (
  File "/home/ubuntu/llama3/venv/lib/python3.12/site-packages/torchtune/utils/precision.py", line 21, in <module>
    from pkg_resources import packaging
ImportError: cannot import name 'packaging' from 'pkg_resources' (/home/ubuntu/llama3/venv/lib/python3.12/site-packages/pkg_resources/__init__.py)
```

### Solution
[...]

PR #1114, opened 2024-06-24
Pin setuptools version

ebsmothers:
As pointed out in #1111, there is an import error when importing the setuptools `packaging` API in versions >= 70.0. This is discussed on the setuptools repo [here](https://github.com/pypa/setuptools/issues/4376) and on other downstream repos as well (see e.g. [here](https://github.com/pytorch/serve/issues/3176)). Going with the recommended solution from the second link and just pinning to 69.5.1 to resolve the issue.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -57,7 +57,7 @@ version = {attr = "torchtune.__version__"}
 
 # ---- Explicit project build information ---- #
 [build-system]
-requires = ["setuptools", "wheel"]
+requires = ["setuptools==69.5.1", "wheel"]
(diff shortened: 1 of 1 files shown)


RdoubleA (approved):

joecummings:
How did you test this? @ebsmothers

ebsmothers:
@joecummings I didn't yet. Will quickly create a new env to show this fixes the import error

ebsmothers:
Actually the `pkg_resources` API is no longer used in our repo so we don't need this pin. Closing this PR

Issue #1100, opened 2024-06-20
want dora and nef-tune supports!


jeffchy:
as the title

ebsmothers:
Hi @jeffchy thanks for the issue! We are currently working on adding DoRA to the library and hope to have it available very soon. For NEFTune I am a bit less familiar; if I understand correctly it is just adding noise to the embeddings? If that's the case is it possible to define an embeddings class like

```
class NEFTuneEmbeddings(nn.Module):
	def __init__(self, num_embeddings: int, embedding_dim: int, alpha: float):
		embedding = nn.Embedding(num_embeddings, embedding_dim)
        self.register_parameter("weight", nn.Parameter(embedding.weight))
		self.alpha = alpha
	
	def forward(self, input: Tensor) -> Tensor:
		raw_embeds = torch.nn.functional(input, self.weight)
		bsz, seq_len, embed_dim = raw_embeds.shape
		scaling_factor = self.alpha / math.sqrt(seq_len * embed_dim)
		noise = scaling_factor * torch.rand_like(bsz, seq_len, embed_dim)
		return raw_embeds + noise
```

Then if you want to build e.g. llama3 with this component, you can change [this line](https://github.com/pytorch/torchtune/blob/ef6e196d8e47e9bc584bc9f7ce836f646443381f/torchtune/models/llama3/_component_builders.py#L106) to point to your `NEFTuneEmbeddings` class (optionally adding `alpha` as a new parameter to the `llama3` function in case you want to experiment with it)

winglian:
@ebsmothers It may be worth considering adding [SymNoise](https://arxiv.org/abs/2312.01523v2) alongside NEFTune as components.

jeffchy:
Thanks for your response! These simple noise-based methods are useful in our experiments, it would be nice if they got officially supported.

ebsmothers:
Closing this old issue since we have DoRA support as of #1115

Issue #893, opened 2024-04-28
[FR] (Q)DoRA


DreamGenX:
(Q)DoRA, an alternative to (Q)LoRA is quickly proving to be a superior technique in terms of closing the gap between FFT and PEFT.

Known existing implementations:
- https://github.com/huggingface/peft -- enable qlora + set use_qdora
- https://github.com/AnswerDotAI/fsdp_qlora?tab=readme-ov-file#training-options

Field reports / comparisons to LoRA:
- https://arxiv.org/abs/2402.09353
- https://www.answer.ai/posts/2024-04-26-fsdp-qdora-llama3.html

rohan-varma:
Thanks for filing this issue! The torchtune team has definitely been following some of the progress around (Q)DoRA and are super interested in adding it. Stay tuned for more thoughts and discussion soon!

RdoubleA:
@DreamGenX If this is something you're interested in contributing to the library, we'd be happy to work with you on it :)

Prakyathkantharaju:
@RdoubleA I added a Dora-based update to my fork. If you approve, I can submit a pull request. 
 Link to the Dora update line here: https://github.com/Prakyathkantharaju/torchtune/blob/aefb8cbb02712177d690ca65cbac480fcb8ac429/torchtune/modules/peft/lora.py#L137

Prakyathkantharaju:
Hello @ebsmothers ,

Yes, that is correct. The method presented in the paper and that I wrote is slightly different, I based my code on this repo: https://github.com/rasbt/dora-from-scratch/blob/main/Using-LinearDoRA.ipynb 

I updated the code based on the author's recommendation [here](https://github.com/Prakyathkantharaju/torchtune/blob/cbafe85b56925fc51f75e338810e060af5c83270/torchtune/modules/peft/lora.py#L138), which apparently make this inference faster on GPU. 

Here is the reference for the author's comment: https://github.com/huggingface/peft/pull/1474#issuecomment-1963402710 

I have added references in the code as well. 

I also added a pull request https://github.com/pytorch/torchtune/pull/936 So that I can track any changes much clearer.

ebsmothers:
Thanks for the clarification @Prakyathkantharaju, and thanks for opening the PR. I agree, let's consolidate the discussion on the PR. I left a few initial comments, let's continue the conversation there.

RdoubleA:
Currently being worked on in #1115

ebsmothers:
[...]

PR #1115, opened 2024-06-24
DoRA

calvinpelletier:
#### Context

Adding support for DoRA: https://arxiv.org/abs/2402.09353

Also refactoring `LoRALinear`. The adapter logic is now encapsulated in a submodule `LowRankAdapter` while the base layer logic remains in `LoRALinear`. This lets us just wrap the `LowRankAdapter` module for FSDP, instead of separately wrapping the `lora_a`/`lora_b` projections and the new `magnitude` parameter.

Addresses #1100, #893, #936 

#### Changelog

- [x] Refactor `LoRALinear`/`LowRankAdapter`
- [x] Support DoRA in `LoRALinear` (enabled via `use_dora` in constructor)
- [x] Support DoRA in `get_merged_lora_ckpt`
- [x] Add `use_dora` to all model/component builders
- [x] Example configs for DoRA and QDoRA
- [x] Unit test for DoRA
- [x] Update LoRA recipes so that they notify the model when the base parameters have been loaded (for DoRA initialization)
- [ ] Fix FSDP unit tests
- [ ] Update `recipes/dev/lora_finetune_fsdp2.py`
- [ ] Support DoRA models in state dict conversion PEFT <-> torchtune (`*.lora_magnitude_vector.weight` <-> `*.lora.magnitude`)

#### Test plan

`pytest tests/torchtune/modules/peft/test_lora.py::TestLoRALinear::test_dora`
- checks parity against a reference implementation based on huggingface's
- checks that DoRA initializes correctly
- checks that the merged linear module is approximately the same as the unmerged DoRA linear module 

---

Compared finetunes for LoRA variants (config based on `llama3/8B_lora(_single_device)` for 2 epochs):

LoRA 1xH100
- [final loss: ~1.1](https://api.wandb.ai/links/torchtune/6h3ri4z0)
- time: 325min (~1400 tok/sec)
- GPU mem: 26.8gb

DoRA 1xH100
- [final loss: ~1.1](https://api.wandb.ai/links/torchtune/xp7fpfdu)
- time: 367min (~1300 tok/sec)
- GPU mem: 26.8gb

QLoRA 1xH100
- [final loss: todo](https://api.wandb.ai/links/torchtune/cg856nz6)
- time: todo
- GPU mem: 21.8gb

QDoRA 1xH100
- [final loss: ~1.1](https://api.wandb.ai/links/torchtune/00ggtewj)
- time: 425min (~1100 tok/sec)
- GPU mem: 21.8gb

DoRA 2xH100
- [final loss: ~1.1](https://api.wandb.ai/links/torchtune/gp5t7lyq)
- time: 1656min (~130 tok/sec/gpu)
- GPU mem: 27.1gb + 21.0gb

---

Attempted to replicate section 3.2 ("Weight Decomposition Analysis") of the DoRA paper.

[The code is here.](https://github.com/calvinpelletier/torchtune/blob/aa99fe6ee73468fcecfeea45b6eef2d729b6d0b5/tmp.py)

Results (for comparison with figure 2 of the paper):
Each point is a single query projection (avg direction delta on the X axis, avg magnitude delta on the Y axis).
Full finetune
full
LoRA finetune
lora
DoRA fintune
dora

The difference with the paper is probably due to a different experimental setup (Llama3 finetuned on alpaca instead of VL-BART finetuned on four image-text tasks).


SLR722:
Thanks for introducing DoRA to torchtune!
- could you add an example distributed yaml config to showcase DoRA? 
- Could you add more details on how you 'checks parity against a reference implementation based on huggingface's' and 'checks that DoRA initializes correctly' in your test plan? 
- In DoRA paper, DoRA outperformances LoRA on several tasks while they didn't mention the loss different. I saw in your comparison, the loss is pretty similar between LoRA and DoRA, could you also eval DoRA and LoRA on some eval benchmarks to see the difference?

torchtune/models/convert_weights.py
```diff
@@ -206,8 +206,8 @@ def _permute(t, n_heads):
 
 # Mapping from torchtune LoRA module names to PEFT LoRA module names
 _TO_PEFT_KEYS = {
-    "lora_a": "lora_A",
-    "lora_b": "lora_B",
+    "lora.a": "lora_A",
```
SLR722, line 209: why do we rename the LoRA keys name?

torchtune/modules/peft/lora.py
```diff
@@ -101,45 +92,112 @@ def _create_weight_and_bias(self):
...
+    def forward(self, x: Tensor, base_out: Tensor, base_weight: Tensor) -> Tensor:
+        if self.disabled:
+            return base_out
+        x = self.dropout(x)
+        lora_out = self.scaling * self.b(self.a(x))
+        if self.use_dora:
+            lora_out = self._dora_forward(x, base_weight, lora_out)
+        return base_out + lora_out
+
+    def _dora_forward(self, x: Tensor, base_weight: Tensor, lora_out: Tensor) -> Tensor:
```
SLR722, line 195: The DoRA logic looks good to me based on HF's implementation

torchtune/modules/peft/lora.py
```diff
@@ -15,7 +16,7 @@
 from torchtune.modules.peft.peft_utils import AdapterModule
 
 
-class LoRALinear(nn.Module, AdapterModule):
+class LoRALinear(nn.Module):
```
SLR722, line 19: For this LoRA refactor part, defer to @ebsmothers to check

ebsmothers:
**Update**: I commandeered this PR and pushed some changes with help from @weifengpy. Instead of the previous version where we split out `LoRALinear` into two classes to work with FSDP, I've kept our existing LoRA design and simply added a new `DoRALinear` class that's pretty similar to `LoRALinear` but with the addition of the `magnitude` parameter. 

A lot of the reason for the design of the original version was the usage of FSDP1 APIs, which were a bit too restrictive in terms of sharding params with different values of `requires_grad` in the same `nn.Module`. But with FSDP2 APIs, this is not a problem. However, loading the magnitude param on meta device was a bit tricky.. I wound up using the workaround suggested by Wei in https://github.com/pytorch/pytorch/issues/132721, calling `load_state_dict()` on the `magnitude` params to force the call to `reset_sharded_param` after we initialize them.

There is still some more work to do here though:

(1) Checkpoint save does not work (I think this should be relatively straightforward though)
(2) A lot of cleanup, I'm sure I broke a bunch of tests with my refactor

I think QDoRA will also need more extensive testing too. cc @SLR722

torchtune/modules/peft/peft_utils.py
```diff
@@ -340,3 +365,14 @@ def validate_missing_and_unexpected_for_lora(
                 raise AssertionError(f"Missing LoRA key {k} from adapter state dict")
     if lora_unexpected:
         raise AssertionError("Unexpected key loading adapter")
+
+
+def load_dora_magnitudes(model: nn.Module) -> None:
```
ebsmothers, line 370: Fyi after https://github.com/pytorch/pytorch/pull/132954/files landed we can now remove this hack by following the example there

recipes/dev/lora_finetune_fsdp2.py
```diff
@@ -314,6 +316,8 @@ def _setup_model(
         for m in reversed(list(model.modules())):
             if isinstance(m, nn.Linear) and m.weight.requires_grad:
                 fully_shard(m, **fsdp_kwargs)
+            if isinstance(m, DoRALinear):
```
SLR722, line 319: why do we treat DoRALinear separately here? why don't we treat it same as LoRALinear?

tests/torchtune/utils/test_distributed.py
```diff
@@ -194,7 +196,7 @@ def test_lora_fsdp_wrap(self):
 
         adapter_params = get_adapter_params(model)
         set_trainable_params(model, adapter_params)
-        num_lora_ab, num_transformer_layers = _get_n_lora_and_tformer_layers(model)
+        num_lora, num_transformer_layers = _get_n_lora_and_tformer_layers(model)
```
SLR722, line 199: I saw multiple changes to test_lora_fsdp_wrap test but the unit test doesn't pass on those changes.

ebsmothers: Sorry can you elaborate? Do you mean changes when you merge with latest main? Or changes from Calvin's initial version?

recipes/dev/lora_finetune_fsdp2.py
ebsmothers: I think it's necessary for proper initialization of the magnitude from the sharded LoRA A and B weights. Since otherwise they will be in different FSDP param groups so we would probably need to manually gather and then reshard the weights. You can doublecheck by commenting out L319-320 though to do the usual LoRA sharding, my guess is you will see some kind of DTensor error though.

tests/torchtune/utils/test_distributed.py
SLR722: There are several changes on test_distributed.py compared with master and I think it's either done by Calvin or your FSDP refactor work. Those tests are broken and I don't have context on why we need these changes

SLR722:
I made several updates to this PR to push it closer to the finish line. 
- fix and clean up several unit tests to let majority of the unit tests pass (there are some remaining issues will list below)
- split test_dora.py and test_lora and improve DoRA logic to make it match with reference implementation 
- fix the checkpointer logic so that dora_distributed, dora_single_device and qdora_single_device can run successfully E2E 

There are some remaining issues may need help from @ebsmothers or other people
- test_dora.py numerical doesn't match ref implementation when dropout p = 0.1 (maybe it's expected because of the non-deterministic of dropout?)
- some tests in test_distributed.py doesn't pass (see the discussions here https://github.com/pytorch/torchtune/pull/1115#discussion_r1725895692)
- 8B_dora_fsdp2.yaml takes long time in get_full_model_state_dict() with >2 ranks which causes NCCL timeout

cc: @ebsmothers

tests/torchtune/utils/test_distributed.py
ebsmothers: OK yeah I can take a look. The only changes I made were [these ones](https://github.com/pytorch/torchtune/pull/1115/commits/dd694cef19a340a0b181778ae296ee2da5336cc4#diff-a4e7291c0c5acab6a14537dd1c67bf0406396115dafb7459149fe08fd6502758)

SLR722: I fixed the test_distributed.py issue by reverting some of Calvin's change on _distributed.py, which are unnecessary after the DoRA refactor

SLR722:
To further verification the correctness of this DoRA work
- compare the loss between LoRA and DoRA 
  - LoRA: [0.8608](https://wandb.ai/torchtune_dev/torchtune/reports/loss-24-08-26-21-08-15---Vmlldzo5MTU4NTY3)
  - DoRA: [0.8631](https://wandb.ai/torchtune_dev/torchtune/reports/loss-24-08-26-21-09-51---Vmlldzo5MTU4NTg0)

- compare the correlation between changes in direction and magnitude
  - LoRA 
lora
  - DoRA 
dora
   - replicated the statement in Figure 2 in https://arxiv.org/pdf/2402.09353

- run eval on mmlu 
  - LoRA (0.6187)
Screenshot 2024-08-26 at 8 57 48 PM
  - DoRA (0.6214)
Screenshot 2024-08-26 at 9 03 30 PM
  - DoRA has slightly better eval results

recipes/configs/dev/llama3/8B_dora_fsdp2.yaml
```diff
@@ -0,0 +1,78 @@
...
+lr_scheduler:
+  _component_: torchtune.modules.get_cosine_schedule_with_warmup
+  num_warmup_steps: 100
+
+loss:
+  _component_: torch.nn.CrossEntropyLoss
+
+# Training
+epochs: 1
+max_steps_per_epoch: 50
```
ebsmothers, line 64: debug code sneaking in 😃

torchtune/modules/peft/dora.py
```diff
@@ -0,0 +1,167 @@
...
+        return weight_norm
+
+    def adapter_params(self) -> List[str]:
+        """
+        Return lora_a.weight and lora_b.weight as adapter params.
+        If bias is enabled, also return lora_a.bias and lora_b.bias.
+        """
+        # NOTE: this function has to be updated if the names of "lora_a" and "lora_b"
+        # in this module change.
+        # TODO: need to add back magnitude, but causing initial check to error out cause it's not defined yet
```
ebsmothers, line 115: Seems like this can be removed now

torchtune/modules/peft/__init__.py
```diff
@@ -10,21 +10,28 @@
...
     "get_adapter_params",
     "set_trainable_params",
     "validate_missing_and_unexpected_for_lora",
     "validate_state_dict_for_lora",
+    "load_dora_magnitudes",
     "disable_adapter",
     "get_merged_lora_ckpt",
     "get_lora_module_names",
+    "_lora_a_init_params",
+    "_lora_b_init_params",
```
ebsmothers, line 36: I don't think we want these to be public

tests/torchtune/modules/peft/test_dora.py
```diff
@@ -0,0 +1,270 @@
...
+        mag_norm_scale = (magnitude / weight_norm).view(1, -1)
+        result_dora = (mag_norm_scale - 1) * (
+            F.linear(x, weight)
+        ) + mag_norm_scale * lora_result * self.scaling
+        return result_dora
+
+    def _get_weight_norm(self, weight, lora_weight):
+        weight = weight + self.scaling * lora_weight
+        weight_norm = torch.linalg.norm(weight, dim=1).to(weight.dtype)
+        return weight_norm
```
ebsmothers, line 270: We should refactor this test. In general we try not to put the reference implementation directly in the unit test, but instead use it to determine the expected values, hardcode those in the test, and point to where we got them

tests/torchtune/modules/peft/test_dora.py
```diff
@@ -0,0 +1,195 @@
...
+        # Ensure forward passes are the same. This is because DoRALinear should use a special
+        # quantized linear operator that runs compute in higher prec (but only saves the 4 bit quantized tensor)
+        # for autograd.
+        inputs = torch.randn(BSZ, SEQ_LEN, 512, dtype=dtype)
+        torch.manual_seed(0)
+        dora_linear_out = dora_linear(inputs)
+        torch.manual_seed(0)
+        qdora_linear_out = qdora_linear(inputs)
+        torch.testing.assert_close(
+            dora_linear_out, qdora_linear_out, rtol=1e-01, atol=1e-01
```
SLR722, line 139: I need to make this assert very loose since I found to_nf4 will change the original value significantly

tests/torchtune/modules/peft/test_dora.py
```diff
@@ -0,0 +1,195 @@
...
+
+from functools import partial
+
+import pytest
+
+import torch
+from tests.test_utils import fixed_init_model
+from torch import nn
+from torchao.dtypes.nf4tensor import NF4Tensor, to_nf4
+from torchtune import utils
```
SLR722, line 15: Note: Need to change here to from torchtune import training after rebasing

tests/torchtune/modules/peft/test_dora.py
SLR722: Address this comment by only compare the numerical in unit test and move the reference implementation to tests/torchtune/models/llama2/scripts/compare_dora.py

tests/torchtune/modules/peft/test_dora.py
```diff
@@ -0,0 +1,195 @@
...
+            DoRALinear(
+                in_dim=512,
+                out_dim=512,
+                rank=RANK,
+                alpha=ALPHA,
+                use_bias=True,
+                quantize_base=False,
+            )
+
+    # @pytest.mark.parametrize("dtype", [torch.bfloat16, torch.float32])
```
ebsmothers, line 100: How come this is commented out?

tests/torchtune/modules/peft/test_dora.py
ebsmothers: Hmm this is a bit suspicious to me. For comparison in the [corresponding LoRA test](https://github.com/pytorch/torchtune/blob/a83eeff0079a73ee04a11e8fc2573ed8f671b231/tests/torchtune/modules/peft/test_lora.py#L144) we do not have to do this. Can we make sure nothing unexpected is happening with the magnitude vector here?

ebsmothers (commented):
A few small things, otherwise want to make sure there are no correctness around quantization with the magnitude. After that this looks good to go!

ebsmothers (approved):
This one is finally ready to go! Huge thanks to both @calvinpelletier for all the initial work and to @SLR722 for taking this one the last mile (really quite a few miles). Really excited to be able to support DoRA!

PR #1117, opened 2024-06-24
[WIP] Multimodal dataset + template transforms

RdoubleA:


torchtune/datasets/_finetune.py
```diff
@@ -0,0 +1,116 @@
...
+import numpy as np
+
+from datasets import load_dataset
+from torch.utils.data import Dataset
+from torchtune.config._utils import _get_component_from_path
+from torchtune.data import CROSS_ENTROPY_IGNORE_IDX, Message
+from torchtune.modules.transforms import Transform
+
+
+class FinetuneDataset(Dataset):
```
joecummings, line 18: Not a huge fan of this name - why not something like `CustomMultiturnDataset`?

torchtune/datasets/_the_cauldron.py
```diff
@@ -0,0 +1,57 @@
...
+
+from typing import Any, Dict, List
+
+from torchtune.data import Message
+
+from torchtune.datasets._finetune import FinetuneDataset
+from torchtune.modules.transforms import Transform
+
+
+class CauldronMessages:
```
joecummings, line 15: Was this asked for? Or just an example of using the `FinetuneDataset`?

torchtune/datasets/_finetune.py
joecummings: Ah I see this is used for `Cauldron`, which I don't think I would count as a "custom" multiturn dataset so maybe the name for this should be `MultiturnChatDataset`??

Issue #1093, opened 2024-06-14
Support NF4 quantization of linear layers without LoRA applied


ebsmothers:
As pointed out by @janeyx99, our `quantize_base` argument will only quantize the base model weights of linear layers with LoRA applied to them (see e.g. [here](https://github.com/pytorch/torchtune/blob/59d36e633ed9f1055f08498805df2ea0486657c2/torchtune/models/llama3/_component_builders.py#L313-L324) in our Llama3 self-attention builder). But this is kind of an artificial constraint.. there's no reason we can't just use the same `to_nf4` API we use in LoRA [here](https://github.com/pytorch/torchtune/blob/59d36e633ed9f1055f08498805df2ea0486657c2/torchtune/modules/peft/lora.py#L94) to quantize other `nn.Linear`s to save memory. We can also define e.g. an `NF4Linear` class if we want (in fact we previously had such a class but ultimately didn't use it, see #465). We just need to figure out the right way to expose this. We could either

- modify the default behavior with `quantize_base` to quantize all linear layers, or
- add additional configurability, either e.g. a `quantize_lora_layers_only` bool or further per-linear configurability similar to what we currently have for our LoRA configs today ([cf.](https://github.com/pytorch/torchtune/blob/59d36e633ed9f1055f08498805df2ea0486657c2/torchtune/models/llama3/_component_builders.py#L133-L136)). Personally I think the latter is overgeneralizing but open to discussion here

cc @joecummings, @winglian

rohan-varma:
Thought it might be useful to share a bit of context here on the original basis for this - we initially enabled `quantize_base` applying only for LoRA modules as we wanted to focus on replicating the original QLoRA implementation in the paper (https://arxiv.org/abs/2305.14314) and planned to follow up generalizing this if it was a popular user request. 

Agreed there are no technical blockers in generalizing this, though pointing out that users probably can't directly use the `to_nf4` API OOTB to achieve this currently as there's probably some things we do around the state_dict that would also need to be changed.

What about configuring it as we do for lora modules? So we specify `lora_modules = [k_proj, v_proj, output_proj]` and can also specify `quantized_modules = [...]` in a similar way.

janeyx99:
By the way, there is also this config `apply_lora_to_output` that might be relevant--I'm not sure how it plays with `lora_attn_modules` if `output_proj` is specified. In the same way, what's the current mechanism to specify qlora over lora? Will that conflict semantically with the new proposed `quantized_modules`?

[...]

PR #1119, opened 2024-06-25
NF4 quantization of linear layers without LoRA applied

winglian:
#### Context
- [X ] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

Please link to any issues this PR addresses. #1093 

#### Changelog
Reverts #658 to bring back FrozenNF4Linear. When `quantize_base` is set to true, all base weights for linear layers are quantized, even if they do not have LoRA applied to them.

#### Test plan


joecummings:
cc @msaroufim

janeyx99 (commented):
Clarification question: does thid mean that `quantize_base` is the boolean with which one specifies the q in qLoRA, which is globally applied to all projs? (versus how LoRA could be applied to a subset of the parameters of q, k, v and not o).

winglian:
running `tune run lora_finetune_single_device --config llama3/8B_qlora_single_device`
but only with `lora_attn_modules: ['q_proj']` instead of all the attention linear layers, uses `15810MiB` on main and `12688MiB` on this branch.

torchtune/modules/low_precision/nf4_linear.py
```diff
@@ -0,0 +1,59 @@
...
+    ):
+        if "bias" in kwargs and kwargs.pop("bias"):
+            raise RuntimeError("FrozenNF4Linear does not currently support biases!")
+
+        super().__init__(in_dim, out_dim, device=device, bias=False, **kwargs)
+        self.weight.requires_grad_(False)
+        self.nf4_weight = to_nf4(self.weight.data)
+        # re-register self.weight as the nf4 weight, so that the nf4 weight
+        # shows up as expected in .parameters, state_dict, etc.
+        self.weight = torch.nn.Parameter(self.nf4_weight, requires_grad=False)
```
janeyx99, line 46: Could we avoid accessing `.data` of `self.weight`?

I'm curious if there's a way to use swap_tensors here to accomplish this more cleanly. https://pytorch.org/docs/stable/generated/torch.utils.swap_tensors.html  cc @mikaylagawarecki  any ideas?

janeyx99 (commented):
Rest of PR looks fine; we should implement FrozenNF4Linear as cleanly as possible though

janeyx99 (commented):
Also, what test cases can we write to verify functionality?

torchtune/modules/low_precision/nf4_linear.py
mikaylagawarecki: @winglian Could you explain the rationale for the `.data` access?

ebsmothers: Yeah can we do something similar to what's currently done in [LoRALinear](https://github.com/pytorch/torchtune/blob/58d58bce65836c23865e6a9aa683475631527d82/torchtune/modules/peft/lora.py#L94) for this?

ebsmothers (commented):
Overall the changes look reasonable to me. Re @janeyx99's testing comment, a couple things I think we should add here:

(1) a unit test for nf4_linear
(2) some kind of e2e test. I am thinking: run one of our QLoRA recipes before and after the change, confirm that (a) the peak memory reduces as expected, and (b) we see no regression in eval metrics on the resulting fine-tuned checkpoint.

It's also likely that this will break some of our existing QLoRA tests. So e.g. the values [here](https://github.com/pytorch/torchtune/blob/58d58bce65836c23865e6a9aa683475631527d82/tests/recipes/test_lora_finetune_single_device.py#L55-L58) would need to be updated

torchtune/modules/low_precision/nf4_linear.py
winglian: This is simply reverting back the code that was deleted in https://github.com/pytorch/torchtune/pull/658/files#diff-74b0d911936ebd1d0e216004577afed27b84b6bfdff9c6a9a1a28f6fac054850L45

Removing the `.data` does seem to work as well, so I've checked in that change.

torchtune/modules/low_precision/nf4_linear.py
```diff
@@ -0,0 +1,59 @@
...
+
+    def __init__(
+        self, in_dim: int, out_dim: int, device: Optional[torch.device] = None, **kwargs
+    ):
+        if "bias" in kwargs and kwargs.pop("bias"):
+            raise RuntimeError("FrozenNF4Linear does not currently support biases!")
+
+        super().__init__(in_dim, out_dim, device=device, bias=False, **kwargs)
+        self.weight.requires_grad_(False)
+        self.nf4_weight = to_nf4(self.weight)
```
janeyx99, line 43: nice

tests/torchtune/modules/low_precision/test_nf4_linear.py
```diff
@@ -0,0 +1,126 @@
...
+        nf4_linear = FrozenNF4Linear(dim, dim, device="cuda", dtype=dtype)
+        orig_weight = nf4_linear.weight.get_original_weight().clone().detach()
+        bnb_nf4_linear = _build_bnb_linear(input_weight=orig_weight)
+
+        # From https://github.com/drisspg/transformer_nuggets/blob/f05afad68ad9086d342268f46a7f344617a02314/test/test_qlora.py#L65
+        bnb_reconstruction = bnb_nf4_linear(
+            torch.eye(dim, dim, dtype=dtype, device="cuda")
+        )
+        # Ensure nf4_linear and bnb reconstructions are close to each other.
+        diff = (
```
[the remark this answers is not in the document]
janeyx99: and agreed with @ebsmothers  that there should be a test to ensure memory is as expected, either in the code using torch.cuda.max_memory_allocated or with a local e2e test summarized in the PR body.

torchtune/modules/low_precision/nf4_linear.py
```diff
@@ -0,0 +1,59 @@
...
+        # re-register self.weight as the nf4 weight, so that the nf4 weight
+        # shows up as expected in .parameters, state_dict, etc.
+        self.weight = torch.nn.Parameter(self.nf4_weight, requires_grad=False)
```
janeyx99, line 46:
```suggestion
        torch.utils.swap_tensors(self.weight, torch.nn.Parameter(self.nf4_weight, requires_grad=False))
```
Not a big expert of swap_tensor, but something like this should ensure that the non-data portions of self.weight are preserved + all other references to it are preserved.

tests/torchtune/modules/low_precision/test_nf4_linear.py
```diff
@@ -0,0 +1,126 @@
...
+from torchtune.modules.low_precision import FrozenNF4Linear
+from torchtune.utils.seed import set_seed
+
+try:
+    from torchao.dtypes.nf4tensor import NF4Tensor
+except ImportError as e:
+    raise RuntimeError(
+        "Please install torchao to run this test."
+        "Example: pip install git+https://github.com/pytorch-labs/ao.git"
+    ) from e
```
ebsmothers, line 19: Do we still need this?

tests/torchtune/modules/low_precision/test_nf4_linear.py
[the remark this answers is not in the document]
ebsmothers: Yeah I also realized @winglian had already done this but only after my previous comment. I think in that case it should be fine to just document it in the PR summary with the command used to get those numbers.

ebsmothers (commented):
One other thing -- this will change the default behavior of recipes using `quantize_base`. Can you run some QLoRA fine-tunes and report eval numbers before and after the change to confirm it doesn't cause any model quality regressions? (E.g. if I only apply LoRA to `q_proj` and `v_proj` and set `quantize_base=True` I will now be converting a lot more linears into lower precision than I did before. I don't think this should ultimately matter much but would be good to document that in the PR summary)

tests/torchtune/modules/low_precision/test_nf4_linear.py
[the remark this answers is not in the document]
winglian: yeah, that was using `tune run lora_finetune_single_device --config llama3/8B_qlora_single_device`, with the change of `lora_attn_modules: ['q_proj']` so that we have some linear layers left over to quantize that aren't LoRALinear.

winglian:
here's the evals. slight drop from quantization `0.5682` -> `0.5546`

```
main

|    Tasks     |Version|Filter|n-shot|Metric|   |Value |   |Stderr|
|--------------|------:|------|-----:|------|---|-----:|---|-----:|
|truthfulqa_mc2|      2|none  |     0|acc   |↑  |0.5682|±  |0.0149|

nf4 - quantized linear
|    Tasks     |Version|Filter|n-shot|Metric|   |Value |   |Stderr|
|--------------|------:|------|-----:|------|---|-----:|---|-----:|
|truthfulqa_mc2|      2|none  |     0|acc   |↑  |0.5546|±  |0.0149|
```

ebsmothers:
@winglian thanks for adding the eval results, I think it looks reasonable. One remaining thing before we merge: seems like one of the bnb comparison unit test cases is failing. See [here](https://github.com/pytorch/torchtune/actions/runs/9878727250/job/27283298692?pr=1119)

joecummings (approved):
WOoooooooooooooooooooooooooo

PR #1125, opened 2024-06-26
Pin ao==0.3

ebsmothers:
PR to upgrade ao to 0.3. Clean up a couple ao version checks in the code that are no longer needed. We also need to provide more explicit guidance to users on how to run our FSDP2 recipes using stable ao + pytorch nightlies

Tested 

```
tune run --nnodes 1 --nproc_per_node 2 lora_finetune_fsdp2 --config llama2/7B_qlora
...
1|14|Loss: 1.9024205207824707:   0%|▎                                                                                                                                                | 14/6470 [01:02<7:35:25,  4.23s/it]
```

### DIFF
--- a/recipes/dev/fsdp2_recipes.md
+++ b/recipes/dev/fsdp2_recipes.md
@@ -1,6 +1,6 @@
 ## FSDP2 Recipes
 
-This directory contains training recipes for LoRA and QLoRA using FSDP2.
+This directory contains distributed training recipes for LoRA and QLoRA using [FSDP2](https://github.com/pytorch/pytorch/issues/114299).
 Currently FSDP2 is only available in PyTorch nightly releases.
 
 To set up your environment to run these recipes, you should first install torchtune dependencies,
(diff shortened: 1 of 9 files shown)


recipes/dev/fsdp2_recipes.md
```diff
@@ -0,0 +1,12 @@
+## FSDP2 Recipes
+
+This directory contains training recipes for LoRA and QLoRA using FSDP2.
```
RdoubleA, line 3: Want to add brief details here or a link for what FSDP2 is?

RdoubleA (approved):

recipes/dev/fsdp2_recipes.md
msaroufim: There's a good doc on torchtitan https://github.com/pytorch/torchtitan/blob/main/docs/fsdp.md

weifengpy:
nice! congrats for moving the pin. I was worried about everything moving fast. all good then :D

PR #1129, opened 2024-06-27
Updated CausalAttention doc string

pbontrager:
#### Context
- [ ] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

Incorrect docstring for num_kv_heads is updated.

### DIFF
--- a/torchtune/modules/attention.py
+++ b/torchtune/modules/attention.py
@@ -47,9 +47,9 @@ class CausalSelfAttention(nn.Module):
         embed_dim (int): embedding dimension for the model
         num_heads (int): number of query heads. For MHA this is also the
             number of heads for key and value
-        num_kv_heads (int): number of key and value heads. If specified,
-            user should ensure `num_heads` % `num_kv_heads` == 0. Default value is
-            `None`, in which case this is the same as MHA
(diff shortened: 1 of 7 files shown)


torchtune/modules/attention.py
```diff
@@ -47,9 +47,9 @@ class CausalSelfAttention(nn.Module):
         embed_dim (int): embedding dimension for the model
         num_heads (int): number of query heads. For MHA this is also the
             number of heads for key and value
-        num_kv_heads (int): number of key and value heads. If specified,
-            user should ensure `num_heads` % `num_kv_heads` == 0. Default value is
-            `None`, in which case this is the same as MHA
+        num_kv_heads (int): number of key and value heads. User should ensure
+            `num_heads` % `num_kv_heads` == 0. For standard MHA set `num_kv_heads` == `num_heads`,
+            for GQA `num_kv_heads` < `num_heads`, and for MQA set `num_kv_heads` == 1.
```
kartikayk, line 52: Probably should call out that num_heads % num_kv_heads should be 0 (thats the first check in code too)

kartikayk (approved):
left one comment, but thanks for fixing this!

PR #1138, opened 2024-07-02
Interleaved image support in tokenizers

RdoubleA:
Adding support for image special tokens in tokenizers and Message dataclass across the library. Starting with Llama3Tokenizer as an example, as multimodal Flamingo will be using this.

This PR will enabled support for **multiple images per sample, interleaved throughout the text**. This is done by overhauling how content is represented in the `Message` class:
```
[
    Message(role="user", content=[{"type": "image"}, {"type": "text", "content":"hello"}, ...]),
    ...
]
```
For backwards compatibility, passing a string into the content field will automatically get casted into the list of dictionaries. Using dictionaries is visually clunky, but it enables the most flexibility for any non-text content that needs to be tokenized differently and future metadata fields that may affect tokenization.

We must also use a different paradigm for chat formats. Now, treat it as extra tags that are prepended/appended to the Message content list based on the role:
```
for message in sample:
      content = (
          [{"type": "text", "content": cls.template[message.role][0]}]
          + message.content
          + [{"type": "text", "content": cls.template[message.role][1]}]
      )
      formatted_dialogue.append(
          Message(
              role=message.role,
              content=content,
              masked=message.masked,
              ipython=message.ipython,
              eot=message.eot,
          ),
      )
```

## Changelog
- Overhaul to Message class. Update `Message` content field to be multimodal - can have text or image content in a list of dictionaries. Added properties for easily checking if there's image content or getting the text only content
- Update all chat format tests
- Update all chat formats to use the new Message content structure. Now we simply prepend and append tags as needed
- If `Message` content contains type==image, then add image special token in Llama3Tokenizer

## Test Plan
Compared to a reference implementation token by token. Made this into new unit tests for text only messages, text + image, text + interleaved image, tool calls

## Other considered approaches
There are other approaches I considered for enabling interleaved images in a message. The important workflow to consider is that users will have to use our utilities or create their own transforms to get their datasets into our Message format with interleaved image content. We should make this workflow as easy as possible.

#### 1
Use an enum, `Media.IMAGE`, and make `Message.content` a `List[Union[str, Media]]` that combines texts and places images in between strings
```
[
    Message(role="user", content=[Media.IMAGE, "hello", Media.IMAGE, "world"]),
    ...
]
```
**Pros**: slightly less tedious to convert user datasets into this structure and also process this (versus dictionaries)
**Cons**: not easy to support image metadata or other fields that may affect image tokenization in the future

#### 2
Keep everything as a string and allow special tokens in the text itself. Tokenizer splits on the indicated image token and adds the correct special token id and encodes the strings normally
```
[
    Message(role="user", content="<img> hello <img> world", media={"image": "<img>"})
    ...
]
```
**Pros**: More prompt template / chat format friendly, as we can still format the string as is. In other approaches, since the text is broken up, prompt templating is considerably more difficult. Easier to process string only content, no impact on text-only workflows
**Cons**: we are breaking our rule of no special tokens in the text


recipes/generate.py
```diff
@@ -14,7 +14,7 @@
 
 from torchtune import config, utils
 from torchtune.config._utils import _get_component_from_path
-from torchtune.data import ChatFormat, InstructTemplate, Message
+from torchtune.data import apply_chat_format, ChatFormat, InstructTemplate, Message
```
joecummings, line 17: copycat copycat copycat

tests/torchtune/datasets/test_chat_dataset.py
```diff
@@ -59,30 +23,44 @@ def dialogue(self):
...
+                        {"role": "assistant", "content": "I agree.", "masked": False}
+                    ),
                 ],
             },
         ]
 
     @mock.patch("torchtune.datasets._chat.load_dataset")
     def test_get_item(self, mock_load_dataset, chat_format, dialogue):
-        mock_load_dataset.return_value = Dataset.from_list(dialogue)
+        mock_load_dataset.return_value = dialogue
```
ebsmothers, line 63: Why this change? Shouldn't load_dataset be returning a `Dataset` object?

tests/torchtune/models/llama3/test_llama3_tokenizer.py
```diff
@@ -23,22 +23,157 @@ def tokenizer(self):
...
+                        "content": "I can see the sun. But even if I cannot see the sun, I know that it exists.",
+                    },
+                ],
+                masked=True,
+                eot=True,
+            ),
+            Message(
+                role="assistant",
+                content=[
+                    {"type": "text", "content": "locate_sun(radius=100_000_000)"},
```
ebsmothers, line 124: lol

tests/torchtune/models/llama3/test_llama3_tokenizer.py
```diff
@@ -64,119 +199,361 @@ def token_ids(self):
...
-        assert tokenizer.start_header_id == 128006
-        assert tokenizer.end_header_id == 128007
-        assert tokenizer.eom_id == 128008
-        assert tokenizer.eot_id == 128009
-        assert tokenizer.python_tag == 128255
-
-    def test_tokenizer_vocab_size(self, tokenizer):
-        assert tokenizer.base_vocab_size == 2000
-        assert tokenizer.vocab_size == 128256
+    def test_tokenize_image_and_text_messages(self, tokenizer, image_and_text_messages):
```
ebsmothers, line 290: nit: these are all pretty much just combos of like 3-5 different token sequences, right? Might be clearest to just define them once (like `text_1_tokens`, `header_tokens`, `tool_call_tokens`, etc) then explicitly piece them together in each test to get the expected value (and can just use e.g. `len(tokens)` to build expected masks similarly). Currently reading this unit test is needlessly painful

torchtune/models/llama3/_tokenizer.py
```diff
@@ -181,17 +234,18 @@ def tokenize_messages(
         # bos and eos are always masked
         mask = [True]
         for message in messages:
-            tokenized_message = self.tokenize_message(
-                message, tokenize_header=tokenize_header
-            )
+            tokenized_message = self.tokenize_message(message)
```
ebsmothers, line 237: So we don't actually pass tokenize_header through anymore? Is this just because it's always true for llama3 anyways?

torchtune/models/llama3/_tokenizer.py
```diff
@@ -125,45 +149,75 @@ def decode(
...
-        else:
-            tokenized_message = tokenized_message + [self.eom_id]
+
+        tokenized_header = self._tokenize_header(message) if tokenize_header else []
+
+        tokenized_body = self._tokenize_body(message)
+
+        tokenized_end = self._tokenize_end(message) if tokenize_end else []
+
+        tokenized_message = tokenized_header + tokenized_body + tokenized_end
```
ebsmothers, line 213: very nice

torchtune/models/llama3/_tokenizer.py
```diff
@@ -83,11 +100,18 @@ def _validate_special_tokens(
...
+        for token in [
+            "<|begin_of_text|>",
+            "<|end_of_text|>",
+            "<|start_header_id|>",
+            "<|end_header_id|>",
+            "<|eom_id|>",
+            "<|eot_id|>",
+            "<|python_tag|>",
+            "<|image|>",
+        ]:
```
ebsmothers, line 114: Isn't the whole point of this tokenizer design that we can use different builders to support different sets of special tokens? It feels weird that we would then assert on tokens that may not be present in all llama3 tokenizers

torchtune/data/_chat_formats.py
```diff
@@ -5,21 +5,21 @@
...
+from torchtune.data._types import Message, Role
 
 
 class ChatFormat(ABC):
     """
-    Interface for chat formats. Each chat format should include template
-    prompts with placeholders for the data inputs. There should be a template
-    for each role: system, user, assistant.
+    Interface for chat formats. Each chat format should include tags for system,
+    user, and assistant messages. that are prepended or appended to the message
```
ebsmothers, line 16: nit: not a complete sentence

torchtune/data/_chat_formats.py
```diff
@@ -79,27 +79,38 @@ def format(
...
+        system_message = []
         formatted_dialogue = []
         for message in sample:
-            content = ""
             if message.role == "system":
-                content = cls.system.format(content=message.content)
-                system_message = content
+                system_message = (
+                    [{"type": "text", "content": cls.template["system"][0]}]
+                    + message.content
```
ebsmothers, line 88: Sorry maybe I'm being dense here. But isn't `message.content` now a list of dicts? How are we able to do this?

ebsmothers (commented):
Thanks for updating the design based on our discussions. Have a handful of comments but no huge concerns. Can you also fix the doc build job failure and update the PR summary to reflect the design we settled on here?

tests/torchtune/datasets/test_chat_dataset.py
RdoubleA: I made the sample use Message objects, which Dataset.from_list doesn't like. Otherwise I need to write a quick `convert_to_messages`. I'm ok either way, this just seemed easier

torchtune/models/llama3/_tokenizer.py
RdoubleA: yeah, kind of an artifact from previous approaches I've tried. I left it in there in case it may be useful in the future. but I'm ok to remove

torchtune/models/llama3/_tokenizer.py
RdoubleA: as discussed offline, all but the image token is required. so will remove that one

torchtune/data/_chat_formats.py
RdoubleA: yep, now we are concatenating lists directly

ebsmothers: OK yes I was in fact being dense then

ebsmothers (approved):

PR #1141, opened 2024-07-03
Vision cross attention mask transform

RdoubleA:
## Context
Multimodal models that use cross-attention layers in the text LLM backbone to attend to outputs of a vision encoder on the images require a cross-attention mask to ensure the correct text tokens attend to the right images. For multiple, interleaved images in a single sample, we follow the approach in Flamingo (Fig. 7 of https://arxiv.org/pdf/2204.14198) where text tokens after an image attend to that particular image (more details in docstrings). To create this mask, this PR adds a transform class that takes a sample dictionary (i.e., during data preprocessing in one of torchtune's dataset classes), computes this cross-attention mask and adds it to the dictionary.

Handling variable number of tiles and number of images across samples will be done in the batch collator in a follow-up PR. For now, we return the masks as a list of tensors, one mask per image, which can have varied n_tiles.

## Changelog
- Add base Transform interface
- Add `VisionCrossAttentionMask`
- Add unit tests for `VisionCrossAttentionMask`

## Test plan
`pytest tests/torchtune/modules/transforms/test_transforms.py`

## Docs
image

### DIFF
--- a/tests/torchtune/modules/transforms/test_transforms.py
+++ b/tests/torchtune/modules/transforms/test_transforms.py
@@ -38,11 +38,16 @@ class TestVisionCrossAttentionMask:
 
     @pytest.fixture
     def images(self, num_tiles, tokens):
-        n_img = len([i for i in tokens if i == IMAGE_TOKEN_ID])
-        return [torch.ones(num_tiles, 3, 2, 2) for _ in range(n_img)]
+        n_img = len([token_id for token_id in tokens if token_id == IMAGE_TOKEN_ID])
+        n_channels = 3
+        tile_size = 2
+        return [
+            torch.ones(num_tiles, n_channels, tile_size, tile_size)
+            for _ in range(n_img)
+        ]
 
     @pytest.fixture
-    def transform(self, tile_size, patch_size):
+    def cross_attn_mask_transform(self, tile_size, patch_size):
         # patches per tile = 4
         return VisionCrossAttentionMask(
             tile_size=tile_size,
@@ -50,14 +55,16 @@ class TestVisionCrossAttentionMask:
             image_token_id=IMAGE_TOKEN_ID,
         )
 
-    def test_get_image_attention_intervals(self, transform, tokens):
-        actual = transform._get_image_attention_intervals(tokens)
+    def test_get_image_attention_intervals(self, cross_attn_mask_transform, tokens):
+        actual = cross_attn_mask_transform._get_image_attention_intervals(tokens)
         expected = [[2, 6], [3, 6], [6, 9]]
         assert actual == expected
 
-    def test_call(self, transform, tokens, images, image_num_tokens):
+    def test_call(self, cross_attn_mask_transform, tokens, images, image_num_tokens):
+        sample = {"tokens": tokens, "images": images}
         dummy_kwargs = {"hello": 8}
-        actual = transform(tokens=tokens, images=images, **dummy_kwargs)
+        sample.update(dummy_kwargs)
+        actual = cross_attn_mask_transform(sample)
         expected = [
             torch.zeros(len(tokens), image_num_tokens, dtype=torch.bool)
             for _ in range(len(images))
--- a/tests/torchtune/modules/transforms/test_transforms.py
+++ b/tests/torchtune/modules/transforms/test_transforms.py
@@ -1,0 +1,11 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+from typing import Any, List, Mapping, Protocol
+
+from torchtune.data import Message
+from torchtune.modules.tokenizers import ModelTokenizer
+from torchtune.modules.transforms import Pipeline, TokenizeMessages, CrossAttentionMask
--- a/tests/torchtune/modules/transforms/test_transforms.py
+++ b/tests/torchtune/modules/transforms/test_transforms.py
@@ -4,8 +4,70 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
-from typing import Any, List, Mapping, Protocol
+import pytest
+import torch
+from torchtune.modules.transforms import VisionCrossAttentionMask
 
-from torchtune.data import Message
-from torchtune.modules.tokenizers import ModelTokenizer
-from torchtune.modules.transforms import CrossAttentionMask, Pipeline, TokenizeMessages
+
+IMAGE_TOKEN_ID = 1
+
+
+class TestVisionCrossAttentionMask:
+    @pytest.fixture
+    def num_tiles(self):
+        return 2
+
+    @pytest.fixture
+    def tile_size(self):
+        return 4
+
+    @pytest.fixture
+    def patch_size(self):
+        return 2
+
+    @pytest.fixture
+    def image_num_tokens(self, num_tiles, tile_size, patch_size):
+        return ((tile_size // patch_size) ** 2 + 1) * num_tiles
+
+    @pytest.fixture
+    def tokens(self):
+        # This tests image tokens not at start, consecutive images, and image
+        # with text until end.
+        # text = 2, image = 1
+        return [2, 2, IMAGE_TOKEN_ID, IMAGE_TOKEN_ID, 2, 2, IMAGE_TOKEN_ID, 2, 2]
+
+    @pytest.fixture
+    def images(self, num_tiles, tokens):
+        n_img = len([i for i in tokens if i == IMAGE_TOKEN_ID])
+        return [torch.ones(num_tiles, 3, 2, 2) for _ in range(n_img)]
+
+    @pytest.fixture
+    def transform(self, tile_size, patch_size):
+        # patches per tile = 4
+        return VisionCrossAttentionMask(
+            tile_size=tile_size,
+            patch_size=patch_size,
+            image_token_id=IMAGE_TOKEN_ID,
+        )
+
+    def test_get_image_attention_intervals(self, transform, tokens):
+        actual = transform._get_image_attention_intervals(tokens)
+        expected = [[2, 6], [3, 6], [6, 9]]
+        assert actual == expected
+
+    def test_call(self, transform, tokens, images, image_num_tokens):
+        dummy_kwargs = {"hello": 8}
+        actual = transform(tokens=tokens, images=images, **dummy_kwargs)
+        expected = [
+            torch.zeros(len(tokens), image_num_tokens, dtype=torch.bool)
+            for _ in range(len(images))
+        ]
+        expected[0][2:6, :] = True
+        expected[1][3:6, :] = True
+        expected[2][6:9, :] = True
+        for i in range(len(images)):
+            torch.testing.assert_close(actual["encoder_mask"][i], expected[i])
+            torch.testing.assert_close(actual["images"][i], images[i])
+
+        assert actual["tokens"] == tokens
+        assert actual["hello"] == dummy_kwargs["hello"]
--- a/torchtune/modules/transforms/_transforms.py
+++ b/torchtune/modules/transforms/_transforms.py
@@ -63,7 +63,7 @@ class VisionCrossAttentionMask(Transform):
         self.patches_per_tile = patch_grid_size**2
         self.image_token_id = image_token_id
 
-    def _get_image_attention_intervals(self, tokens: List[int]) -> List[List[int, int]]:
+    def _get_image_attention_intervals(self, tokens: List[int]) -> List[List[int]]:
         """
         Returns a list of lists of the form [start, end) where start is the index
         of the current image token and end is the index of the next image token, exclusive.
@@ -72,7 +72,7 @@ class VisionCrossAttentionMask(Transform):
             tokens (List[int]): List of token IDs in the text sequence
 
         Returns:
-            List[List[int, int]]: List of lists of the form [start, end) indicating
+            List[List[int]]: List of lists of the form [start, end) indicating
                 range of positions in text sequence that should attend to the image
 
         Example:
--- a/torchtune/modules/transforms/_transforms.py
+++ b/torchtune/modules/transforms/_transforms.py
@@ -4,7 +4,7 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
-from typing import Any, Dict, List, Mapping, Protocol
+from typing import Any, List, Mapping, Protocol
 
 import torch
 
@@ -12,12 +12,10 @@ import torch
 class Transform(Protocol):
     """
     Loose interface for all data and model transforms. Transforms operate at the
-    sample level and perform operations on a sample dict which is contained in
-    kwargs. Any fields that will be processed are unfolded with explicit keyword-arguments,
-    then the updated dict is returned.
+    sample level and perform operations on a sample dict, returning the updated dict.
     """
 
-    def __call__(self, **kwargs) -> Mapping[str, Any]:
+    def __call__(self, sample: Mapping[str, Any]) -> Mapping[str, Any]:
         pass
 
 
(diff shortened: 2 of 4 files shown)


torchtune/modules/transforms/_transforms.py
```diff
@@ -0,0 +1,142 @@
...
+    and follow these rules:
+    1) Text tokens immediately following the image token up until the next image token
+    2) Consecutive image tokens attend to all subsequent text tokens
+
+    Resultant mask is of shape (text_seq_len, image_seq_len), where True indicates
+    that the token outputted from the image encoder attends to the token in the
+    text sequence.
+
+    Args:
+        num_patches (int): Number of patches per image, excluding class token.
```
felipemello1, line 77: I wonder if we should add a link to modules/VisionTransformer for a in-depth explanation of what num_patches mean. For better clarity, would it make sense to rename it num_patches_per_tile, since later we multiply it by n_tiles?

If we say "number of patches per image", it may be confusing, because an image can have a variable number of patches.

later on you say:
```
single_image_seq_len = n_tiles * self.num_patches + 1
image_seq_len = single_image_seq_len * n_img
```

So tile != image. Image is a set of tiles.

torchtune/modules/transforms/_transforms.py
```diff
@@ -0,0 +1,142 @@
...
+    sample level and perform operations on a sample dict which is contained in
+    kwargs. Any fields that will be processed are unfolded with explicit keyword-arguments,
+    then the updated dict is returned.
+    """
+
+    def __call__(self, **kwargs) -> Mapping[str, Any]:
+        pass
+
+
+class Compose(Transform):
```
felipemello1, line 25: Since the torchvision compose has a different behavior, I wonder if it makes sense to change Compose to something else, so users dont get confused with tv.Compose. Maybe "ComposeTransforms"?

torchtune/modules/transforms/_transforms.py
```diff
@@ -0,0 +1,142 @@
...
+        self.num_patches = num_patches
+        self.image_token_id = image_token_id
+
+    def _get_image_attention_intervals(
+        self, tokens: List[int]
+    ) -> List[Tuple[int, int]]:
+        """
+        Returns a list of tuples of the form (start, end) where start is the index
+        of the current image token and end is the index of the next image token, exclusive.
+        If the image token attends until the end of the sequence, end will be -1.
```
felipemello1, line 91: nit: should we add Args:,Returns, Examples?

torchtune/modules/transforms/_transforms.py
```diff
@@ -0,0 +1,142 @@
...
+        if len(vision_token_locations) == 0:
+            return []
+        # If there is only one image, it will attend to subsequent text until end
+        if len(vision_token_locations) == 1:
+            return [[vision_token_locations[0], -1]]
+
+        vision_masks = [
+            [tok1, tok2]
+            for tok1, tok2 in zip(
+                vision_token_locations[:-1], vision_token_locations[1:]
```
felipemello1, line 106: nit: 

Maybe add a comment:
"offset by one and get consecutive indices"

I guess renaming tok1, tok2 to something like tok_idx_prev, tok_idx_next may be more intuitive

torchtune/modules/transforms/_transforms.py
```diff
@@ -0,0 +1,142 @@
...
+        last_mask_end = vision_masks[-1][1]
+        for vision_mask in vision_masks[::-1]:
+            if vision_mask[0] == vision_mask[1] - 1:
+                vision_mask[1] = last_mask_end
+            last_mask_end = vision_mask[1]
+        return vision_masks
+
+    def __call__(self, *, tokens, images, **kwargs):
+        # We are still at sample level pre-collating
+        n_img, n_tiles, _, _, _ = images.shape
```
felipemello1, line 123: nit: Maybe add a comment with the other dimensions, so we know what they are, but keep the "_", so we know they are not used?

You said "# We are still at sample level pre-collating"
So is n_img == bsz? If so, for consistency with VisionTransformer, should we rename it?

torchtune/modules/transforms/_transforms.py
```diff
@@ -0,0 +1,142 @@
...
+        text_seq_len = len(tokens)
+        single_image_seq_len = n_tiles * self.num_patches + 1
```
felipemello1, line 125: nit: maybe add comment explaining that +1 is for CLS, if thats the case

felipemello1 (commented):
it looks great! I left a few comments more around naming, since tiles/patches/image is so confusing. I think that this class would benefit from a visual small example too.

regarding the interval parts, IMO we should try to make the operations as explicit as possible. Instead of tok1, tok2; i; vision_mask[0], we should try to give them names, like idx_start, idx_end, etc.

torchtune/modules/transforms/_transforms.py
```diff
@@ -0,0 +1,142 @@
...
+        tokens, mask = self.tokenizer.tokenize_messages(
+            messages, max_seq_len=self.max_seq_len
+        )
+        kwargs.update({"tokens": tokens, "mask": mask})
+        return kwargs
+
+
+class CrossAttentionMask(Transform):
+    """
+    Computes the cross-attention mask for text + image inputs. Text tokens that
```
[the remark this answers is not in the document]
RdoubleA: Yeah that's a good point. Will rename

torchtune/modules/transforms/_transforms.py
RdoubleA: ah sorry, probably my shallow understanding of patches/images/tiles. What I intended was num_patches per tile. If it makes sense I'd like to keep the name consistent with your vision transformer (either patch_grid_size or patch_size maybe?), whichever parameter you use to compute num patches.

I also assumed at this point tiles is padded to the max in all the images. Is this incorrect? Where does the padding happen?

felipemello1: Agreed, patch_size is good! I would just add to the documentation that its the number of patches **per tile**, instead of per image. In other words, whenever we use the word "image", i think it should mean the whole thing.

felipemello1: patches_per_tile is a fixed size, and its calculated as (tile_size // patch_size)**2.

What I did in VisionTransformer was to ask the user to pass tile_size and patch_size, and I calculated it for them. The VisionTransformer has a helper function that saves this value: https://github.com/felipemello1/torchtune/blob/f683812626ad4559464840112ddce516487bea5c/torchtune/modules/vision_transformer.py#L249

Maybe get it form the model, or ask for tile_size and patch_size, to avoid user confusion?

torchtune/modules/transforms/_transforms.py
RdoubleA: how about Pipeline?

ebsmothers: Just for my own understanding, the main difference with torchvision compose is that we support multiple inputs and multiple outputs here? Can we not just use torchvision compose with a single dict?

torchtune/modules/transforms/_transforms.py
ebsmothers: yeah please add type and shape info for arguments to `__call__`

torchtune/modules/transforms/_transforms.py
```diff
@@ -0,0 +1,142 @@
...
+
+        mask = torch.zeros(text_seq_len, image_seq_len, dtype=torch.bool)
+        for image_num, interval in enumerate(intervals):
+            start, end = interval
+            end = text_seq_len if end == -1 else end
+            mask[
+                start:end,
+                image_num
+                * single_image_seq_len : (image_num + 1)
+                * single_image_seq_len,
```
ebsmothers, line 138: nit: split line differently if linter allows it, as written this is confusing

torchtune/modules/transforms/_transforms.py
```diff
@@ -0,0 +1,142 @@
...
+            ] = True
+
+        kwargs.update({"encoder_mask": mask, "tokens": tokens, "images": images})
```
ebsmothers, line 141: Why do we need to also update with tokens and images? Isn't this a no-op for those args?

torchtune/modules/transforms/_transforms.py
felipemello1: I tried naming something as Pipeline, as Kartikay said it would confuse people, because it is also used by other libraries :P. I guess sklearn?

torchtune/modules/transforms/_transforms.py
RdoubleA: Since they are explicit keyword args, they get unfolded from kwargs and you have to add them back in

torchtune/modules/transforms/_transforms.py
RdoubleA: @ebsmothers our Compose needs to have a slightly different forward signature to unfold dictionary inputs. From [torchvision](https://github.com/pytorch/vision/blob/main/torchvision/transforms/transforms.py#L93C5-L96C19):
```
def __call__(self, img):
        for t in self.transforms:
            img = t(img)
        return img
```
but to avoid confusion, I agree should name it something else. Just haven't figured out what yet

torchtune/modules/transforms/_transforms.py
```diff
@@ -0,0 +1,164 @@
...
+        """
+
+        # One sample can have multiple images - verify the number of image tokens
+        # is the same
+        n_img = len(images)
+        intervals = self._get_image_attention_intervals(tokens)
+        assert len(intervals) == n_img
+
+        # Create mask for each individual image based on its number of tokens,
+        # which can vary based on number of tiles. The masks are padded and concatenated
```
kartikayk, line 146: noob question: what does "and concatenated" mean here?

torchtune/modules/transforms/_transforms.py
```diff
@@ -0,0 +1,164 @@
...
+        # together in the batch collator
+        text_seq_len = len(tokens)
+        masks = []
+        for image_num, interval in enumerate(intervals):
+            # Identify what part of text sequence should be attended
+            start, end = interval
+            # Compute this image's number of tokens based on num tiles, patches per tile
+            n_tiles = images[image_num].shape[0]
+            image_seq_len = n_tiles * (self.patches_per_tile + 1)  # +1 for CLS token
```
kartikayk, line 155: So we have a CLS token per tile?

torchtune/modules/transforms/_transforms.py
```diff
@@ -0,0 +1,164 @@
...
+        img2 │   │ │ ■ │ │ ■ │ │ ■ │ │ ■ │ │ ■ │ │   │ │   │ │   │ │   │ │   │
+             └───┘ └───┘ └───┘ └───┘ └───┘ └───┘ └───┘ └───┘ └───┘ └───┘ └───┘
+             ┌───┐ ┌───┐ ┌───┐ ┌───┐ ┌───┐ ┌───┐ ┌───┐ ┌───┐ ┌───┐ ┌───┐ ┌───┐
+        img3 │   │ │   │ │   │ │   │ │   │ │   │ │ ■ │ │ ■ │ │ ■ │ │ ■ │ │ ■ │
+             └───┘ └───┘ └───┘ └───┘ └───┘ └───┘ └───┘ └───┘ └───┘ └───┘ └───┘
+            <img1> <img2>These  are   two  dogs. <img3> This   is    a    cat.
+
+
+
+    Resultant mask is of shape (text_seq_len, image_seq_len), where True indicates
```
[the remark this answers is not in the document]
RdoubleA: this mask is per image, so what is returned is a list of these tensors. I can make this more clear

torchtune/modules/transforms/_transforms.py
```diff
@@ -0,0 +1,164 @@
...
+        end = len(tokens)
+        vision_token_locations = [
+            i for i, token in enumerate(tokens) if token == self.image_token_id
+        ]
+        # Return empty list if there are no images
+        if len(vision_token_locations) == 0:
+            return []
+        # If there is only one image, it will attend to subsequent text until end
+        if len(vision_token_locations) == 1:
+            return [[vision_token_locations[0], end]]
```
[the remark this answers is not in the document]
RdoubleA: ooh yeah, good catch

torchtune/modules/transforms/_transforms.py
RdoubleA: if we have multiple images in a sample, their masks need to be concatenated after padding upto max tiles per image

torchtune/modules/transforms/_transforms.py
```diff
@@ -0,0 +1,164 @@
...
+            n_tiles = images[image_num].shape[0]
+            image_seq_len = n_tiles * (self.patches_per_tile + 1)  # +1 for CLS token
+            # Mask will be block of 1s at the corresponding interval in the text.
+            # It is not a causal block because all the image tokens correspond
+            # to a single image, so text tokens attend to all the image's tokens
+            mask = torch.zeros(text_seq_len, image_seq_len, dtype=torch.bool)
+            mask[start:end, :] = True
+            masks.append(mask)
+
+        kwargs.update({"encoder_mask": masks, "tokens": tokens, "images": images})
```
ebsmothers, line 163: Maybe it's just me but I find this whole kwargs pattern kinda unintuitive. Like really we are covertly using this transform to also add its inputs into the dictionary (does that mean that it doesn't already contain them?). Maybe I need to see the callsite, but I don't understand why we can't just have e.g. 

```
sample: Dict[str, Any]
mask_transform = VisionCrossAttentionMask(...)
sample["encoder_mask"] = mask_transform(sample["tokens"], sample["images"])
``` 

Alternatively if you want to go more of the compose route:

```
class VisionCrossAttentionMask:
	def __call__(sample: Dict[str, Any]) -> Dict[str, Any]:
		# just key into sample directly in the transform
		sample["encoder_mask"] = construct_mask(sample["tokens"], sample["images"])
		return sample

mask_transform = VisionCrossAttentionMask(...)
sample = mask_transform(sample)
```

But rn we are kinda halfway in between the two and it feels clunky to me.

RdoubleA: cc @pbontrager  who'sthought about this a lot more

We were between two approaches, the kwargs approach here and the sample approach you described where you key into the fields you need for the transform. One issue with that was this implicit assumption that token ids are always under "tokens", image tensors are always under "images", etc., so we thought we may have to provide these keys as attributes to the transform class. Then the key management gets a bit messy to handle and error prone.

Although, thinking about this more, the collator will also implicitly assume the keys "tokens", "labels", "images" etc will be present. If we're consistent with these across the library, and the fact that the number of these keys in the sample dict will remain quite small ~5-10 max, it might be fine to keep these assumptions, or do something similar to checkpointing with model key, optim key, etc as constants.

I don't have a strong opinion either way, although it would be nice to keep a forward signature of just sample.

torchtune/modules/transforms/_transforms.py
felipemello1: yes

felipemello1 (commented):
mostly nits about naming/docs. Feel free to ignore.

felipemello1 (commented):
mostly nits about naming/docs. Feel free to ignore.

tests/torchtune/modules/transforms/test_transforms.py
```diff
@@ -0,0 +1,73 @@
...
+    @pytest.fixture
+    def tokens(self):
+        # This tests image tokens not at start, consecutive images, and image
+        # with text until end.
+        # text = 2, image = 1
+        return [2, 2, IMAGE_TOKEN_ID, IMAGE_TOKEN_ID, 2, 2, IMAGE_TOKEN_ID, 2, 2]
+
+    @pytest.fixture
+    def images(self, num_tiles, tokens):
+        n_img = len([i for i in tokens if i == IMAGE_TOKEN_ID])
```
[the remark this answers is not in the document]
kartikayk: +1 I think this is a great point

torchtune/modules/transforms/_transforms.py
ebsmothers: Yeah imo you're gonna need the keys somewhere anyways, better to just be explicit about it than do this weird back and forth between dict values and standalone arguments.

tests/torchtune/modules/transforms/test_transforms.py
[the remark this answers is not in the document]
RdoubleA: ah, bad habits die hard

tests/torchtune/modules/transforms/test_transforms.py
```diff
@@ -0,0 +1,73 @@
...
+        return [torch.ones(num_tiles, 3, 2, 2) for _ in range(n_img)]
```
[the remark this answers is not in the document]
RdoubleA: I had that before, didn't realize I changed it... will update

tests/torchtune/modules/transforms/test_transforms.py
```diff
@@ -0,0 +1,73 @@
...
+
+    @pytest.fixture
+    def transform(self, tile_size, patch_size):
```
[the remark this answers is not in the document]
RdoubleA: sure, just for you Felipe

torchtune/modules/transforms/_transforms.py
```diff
@@ -118,25 +116,49 @@ def _get_image_attention_intervals(
...
+            tokens (List[int]): List of token IDs in the text sequence. Number of
+                image token IDs in the sequence must match the number of images.
+            images (List[torch.Tensor]): List of image Tensors post-tiling of shape
+                (n_tiles, c, h, w) each.
+            **kwargs (Dict[str, Any]): all other keys within the sample that will
+                not be altered by this transform.
+
+        Returns:
+            Dict[str, Any]: updated sample with the following keys:
+                - encoder_mask (List[torch.Tensor]): masks of shape (text_seq_len, image_seq_len)
```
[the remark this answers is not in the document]
RdoubleA: yeah agreed

torchtune/modules/transforms/_transforms.py
RdoubleA: made the comment more clear

ebsmothers (approved):

PR #876, opened 2024-04-26
Add Phi3 Mini 4K Instruct Model to torchtune

kartikayk:
Imp Note: The tokenizer still needs some work, this will be a follow up PR.

#### Context
What is the purpose of this PR? Is it to
- [x] add a new feature
- [ ] fix a bug
- [ ] update tests and/or documentation
- [ ] other (please add here)

Please link to any issues this PR addresses.

#### Changelog

This PR adds support for the Phi3 Mini 4K Instruct model to torchtune. Specifically we add the following:
- Phi3's RoPE module. This is not numerically equivalent to the Llama2 or Llama3 models and care needs to be taken to have correct behavior for bf16 training
- State dict conversion logic accounting for the fused qkv and gate_up projection matrices
- Config for multi-gpu full finetuning

#### Test plan
Please make sure to do each of the following if applicable to your PR. (If you're not sure about any one of these just ask and we will happily help.)

- [x] run pre-commit hooks and linters (make sure you've first installed via `pre-commit install`)
- [x] add unit tests for any new functionality
- [x] update docstrings for any new or updated methods or classes
- [x] run unit tests via `pytest tests`
- [ ] run recipe tests via `pytest tests -m integration_test`
- [x] manually run any new or modified recipes with sufficient proof of correctness
	- [ ] include relevant commands and any other artifacts in this summary (pastes of loss curves, eval results, etc.)

Unit Tests

Detailed comparisons with [reference implementation](https://gist.github.com/kartikayk/87db294c9c9791622140cf4c9dc5a628)

```
pytest tests/torchtune
```

Full-finetune Recipe

Screenshot 2024-04-25 at 7 57 45 PM

PR #1143, opened 2024-07-05
Add Qwen2-7B model to torchtune.

fyabc:
#### Context

This PR add support to Qwen2-7B, including model components and generation / full-parameter finetune / LoRA finetune recipes.

- [x] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

#### Changelog

- Add support to Qwen2
- Fix code format

#### Test plan

#### Implementation Details & Notes

- **transformer decoder**: In Qwen2-0.5B and Qwen2-1.5B, the output projection is replaced with token embedding weights ([`tie_word_embeddings = True`](https://huggingface.co/Qwen/Qwen2-0.5B-Instruct/blob/main/config.json#L21)), so we add a new `torchtune.models.qwen2.transformer.Qwen2TransformerDecoder` module to support this feature (set `output=None`).
- **tokenizer**:  `torchtune.models.qwen2._tokenizer.Qwen2Tokenizer` construct a "fast" tokenizer backed by HuggingFace's *tokenizers* library (follow [official transformers implementation](https://github.com/huggingface/transformers/blob/v4.40.1/src/transformers/models/qwen2/tokenization_qwen2_fast.py)).
  - The official implemtation `torchtune.data._chat_formats.ChatMLFormat` will add an extra `<|IM_END|>` tag at the end of the last assistant message in generation (see [this discussion](https://github.com/pytorch/torchtune/pull/1019/files#r1621540801)), so we override the `tokenize_message()` method to fix this problem (will not add `<|IM_END|>` and EOS token if `messages[-1].role == 'assistant' and not messages[-1].content`)
- **verify the correctness**: We compare the output logits and generated responses between torchtune and official transformer implementation under the same prompts and infer hyperparameters.
- **QLoRA**: Since quantized `torchtune.modules.peft.lora.LoRALinear` does not support bias yet, QLoRA finetune recipes a not added into this PR (will support in future).
- More Qwen2 models (0.5B, 1.5B, etc.) will be added in future.


pbontrager:
Thank you for this great contribution. Could you provide a bit more information in the PR Description around some of the decisions you made when implementing Qwen2? I'm interested in particular why the TransformerDecoder wasn't sufficient for implementing Qwen. I am also interesting in understanding better how you tested and verified that this implementation is correct. Looking forward to getting this into our library and thank you again.

fyabc:
> Thank you for this great contribution. Could you provide a bit more information in the PR Description around some of the decisions you made when implementing Qwen2? I'm interested in particular why the TransformerDecoder wasn't sufficient for implementing Qwen. I am also interesting in understanding better how you tested and verified that this implementation is correct. Looking forward to getting this into our library and thank you again.

Hi, I have updated some implementation details in the description, please check it again.

fyabc:
> Hi @fyabc, thanks so much for this PR. A couple initial questions:
> 
> 1. The transformer decoder class looks pretty similar to the one we have for [Gemma](https://github.com/pytorch/torchtune/blob/main/torchtune/models/gemma/transformer.py#L18), which also ties the embedding weights. I see there is also an optional post-embedding norm in your implementation, but other than that are these two the same? Basically I am wondering if we should provide a single general `TiedEmbeddingTransformerDecoder` (or something like that) that can be used by both Gemma and Qwen2 models.
> 2. What are the specific differences between Qwen2 RoPE and Llama2/Llama3 RoPE?
> 3. What's the underlying tokenizer used by Qwen2? Is it just GPT-2 BPE tokenization? We currently do not have a dependency in torchtune on the tokenizers library, so adding this may require a bit more discussion.

Thank you for your suggestions!

1. Yes, Qwen2 decoder is the same as Gemma decoder, I think a general `TiedEmbeddingTransformerDecoder` class is a good solution. (or change `TransformerDecoder` directly?)
2. Like [Phi3](https://github.com/pytorch/torchtune/pull/876), Qwen2 RoPE is not numerically equivalent to the Llama2/Llama3 RoPE in bf16 training.
3. Yes, Qwen2 tokenizer is based on GPT-2 BPE tokenization, and I am implementing another version to remove the dependency on tokenizers (derived from [Qwen2Tokenizer](https://github.com/huggingface/transformers/blob/main/src/transformers/models/qwen2/tokenization_qwen2.py#L83)).

ebsmothers:
> 1. Yes, Qwen2 decoder is the same as Gemma decoder, I think a general `TiedEmbeddingTransformerDecoder` class is a good solution. (or change `TransformerDecoder` directly?)
> 2. Like [Phi3](https://github.com/pytorch/torchtune/pull/876), Qwen2 RoPE is not numerically equivalent to the Llama2/Llama3 RoPE in bf16 training (Qwen2 compute rotary embedding in input dtype, Llama2/3 will convert to fp32).
> 3. Yes, Qwen2 tokenizer is based on GPT-2 BPE tokenization, and I am implementing another version to remove the dependency on tokenizers (derived from [Qwen2Tokenizer](https://github.com/huggingface/transformers/blob/main/src/transformers/models/qwen2/tokenization_qwen2.py?rgh-link-date=2024-07-09T10%3A01%3A49Z#L83)).

Thanks @fyabc for your responses! For (1) let's go with `TiedEmbeddingTransformerDecoder`, no need to change `TransformerDecoder` (since that change would impact a lot more models). If you want to also point Gemma to this new class that'd be great, but if not we can also do it as a follow-up. 

Thanks for clarifying on (2). @joecummings has also been looking at our RoPE implementation so can weigh in with any thoughts he has.

(3) sounds great! Please let me know if you need any pointers from us on adding a new tokenizer; otherwise just ping on here when you are ready for a more detailed review of everything.

fyabc:
@ebsmothers Hi, I have updated this PR based on your suggestions:

1. Add a new `torchtune.modules.transformer.TiedEmbeddingTransformerDecoder` class to handle models with tied embedding weight. (I have not refactored Gemma implementation)
2. Rewrite `Qwen2Tokenizer`, remove the dependency on `tokenizers` library.

torchtune/modules/transformer.py
```diff
@@ -245,3 +246,121 @@ def forward(
         # shape: [b, s, out_dim] - out_dim is usually the vocab size
         output = self.output(h).float()
         return output
+
+
+class TiedEmbeddingTransformerDecoder(TransformerDecoder):
```
SalmanMohammadi, line 251: This PR is awesome!

Small comment here: similar to how `GemmaTransformerDecoder` is a separate component to `TransformerDecoder`, is it worth removing the inheritance here too and just inheriting from `nn.Module`? It should be fine to replicate the code in the constructor and `setup_caches`.

torchtune/models/qwen2/_model_builders.py
```diff
@@ -0,0 +1,124 @@
...
+from torchtune.modules.tokenizers import parse_hf_tokenizer_json
+
+"""
+Model builders build specific instantiations using component builders. For example
+the qwen2_7b model builder uses the qwen2 component builder to create the
+qwen2 7B model.
+"""
+
+
+def qwen2_7b() -> TransformerDecoder:
```
SalmanMohammadi, line 23: similar typing comment here : )

torchtune/models/qwen2/_model_builders.py
```diff
@@ -0,0 +1,124 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
```
SalmanMohammadi, line 1: Is there generally interest in the 1.5B model too? If so, is it worth adding?

Perhaps as a more general point, I noticed in the Qwen2 technical report that weight tying is only used for the 0.5B and 1.5B models, right? Would users want to configure weight tying in the larger models where weight tying is enabled by default?

torchtune/models/qwen2/_tokenizer.py
```diff
@@ -0,0 +1,353 @@
...
+        >>> print(tokenized_text)
+        []
+    """
+
+    system = f"{IM_START}system\n{{content}}{IM_END}\n"
+    user = f"{IM_START}user\n{{content}}{IM_END}\n"
+    assistant = f"{IM_START}assistant\n{{content}}{IM_END}\n"
+    assistant_for_generation = f"{IM_START}assistant\n"
+
+    def __init__(
```
SalmanMohammadi, line 95: Is it worth thinking about trying to unify the constructor args here to help all our tokenizers look and feel the same?

tests/torchtune/models/qwen2/test_qwen2.py
```diff
@@ -0,0 +1,48 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
```
SalmanMohammadi, line 1: I'm sure @ebsmothers will have more thoughts on whether this is necessary here - but I'll share an excellent comment from him about the testing process in torchtune https://github.com/pytorch/torchtune/pull/840#discussion_r1579801172.

tests/assets/tiny_bpe_vocab.json
```diff
@@ -0,0 +1,2002 @@
+{
```
SalmanMohammadi, line 1: I've not looked too closely at where this is used, but could we minify this? : )

tests/assets/tiny_bpe_tokenizer.json
```diff
@@ -0,0 +1,3986 @@
+{
```
SalmanMohammadi, line 1: Similar to my other comment, is it also worth minifying this?

torchtune/modules/transformer.py
joecummings: + 1 to this!

torchtune/models/qwen2/_tokenizer.py
```diff
@@ -0,0 +1,353 @@
...
+    pairs = set()
+    prev_char = word[0]
+    for char in word[1:]:
+        pairs.add((prev_char, char))
+        prev_char = char
+    return pairs
+
+
+class Qwen2Tokenizer(ModelTokenizer):
+    """This class construct a "fast" Qwen2 tokenizer (backed by HuggingFace's *tokenizers* library).
```
joecummings, line 75: nit: no longer backed by HuggingFace's tokenizers library

torchtune/models/qwen2/_tokenizer.py
```diff
@@ -0,0 +1,353 @@
...
+
+    See <https://github.com/huggingface/transformers/blob/v4.40.1/src/transformers/models/qwen2/tokenization_qwen2_fast.py>.
+
+    Args:
+        vocab_file (str): Path to vocab.json file.
+        merges_file (str): Path to merges.txt file.
```
joecummings, line 81: Can you add a quick comment here on what "merges.txt" is?

Also, missing a few args in the docstrings.

joecummings (commented):
Overall, this is coming together very nicely and I appreciate your hard work to add a great new model to torchtune! Most of the comments I had are nits, but I did have one last thing to say on testing.

The individual unit tests are great for proving correctness of individual models, but a E2E run of this recipe on alpaca and comparing the loss curves between this and HuggingFace is a fantastic sanity check to make sure the torchtune users are getting the expected experience. It's also useful to output both loss curves to a Weights & Biases run so we can visually see that they're similar.

torchtune/modules/transformer.py
fyabc: > This PR is awesome!
> 
> Small comment here: similar to how `GemmaTransformerDecoder` is a separate component to `TransformerDecoder`, is it worth removing the inheritance here too and just inheriting from `nn.Module`? It should be fine to replicate the code in the constructor and `setup_caches`.

OK, I will remove the inheritance.

tests/assets/tiny_bpe_vocab.json
fyabc: I follow the setting in `tiktoken_small.model` in assets (vocab_size == 2000).
I will reduce it to 500.

torchtune/models/qwen2/_model_builders.py
fyabc: Yes, I will add other size of models (0.5B / 1.5B / 72B).
In addition, whether weight tying is enabled must be consistent between model initialization (in `_model_builders.py` here) and weight loading (in `_convert_weights.py`, reading configuration from `config.json` file), so I think we should fix the weight tying configuration in default model providers to avoid users from bugs.
Based on the experience of the open source community, most users have no need to modify weight tying when tuning Qwen models. For expert users, they can directly use `torchtune.models.qwen2.qwen2()` to provide correct hyperparameters.

torchtune/models/qwen2/_tokenizer.py
fyabc: merges.txt contains all BPE merge operations, and this file is required to split a single word into byte-level BPE tokens.
You can find a description at <https://github.com/huggingface/transformers/issues/4777>.

torchtune/models/qwen2/_tokenizer.py
fyabc: Can I set default value of `merges_file` to `None`, and infer it from `vocab_file` (`merges_file = merges_file or os.path.join(os.path.dirname(vocab_file), 'merges.txt')`)?

Some users may change `eos_token` and other special tokens when tuning Qwen2, so I keep these keyword arguments for flexibility.

fyabc:
> Overall, this is coming together very nicely and I appreciate your hard work to add a great new model to torchtune! Most of the comments I had are nits, but I did have one last thing to say on testing.
> 
> The individual unit tests are great for proving correctness of individual models, but a E2E run of this recipe on alpaca and comparing the loss curves between this and HuggingFace is a fantastic sanity check to make sure the torchtune users are getting the expected experience. It's also useful to output both loss curves to a Weights & Biases run so we can visually see that they're similar.

Thank you for you suggestions! I will run all recipes and report the results and comparisons between HuggingFace.

tests/assets/tiny_bpe_vocab.json
SalmanMohammadi: Sorry @fyabc, my meaning was to use something like https://codebeautify.org/jsonminifier which just removes any unnecessary characters and formatting. The vocab size itself was probably okay!

torchtune/models/qwen2/_tokenizer.py
SalmanMohammadi: HF will always pull the `merges_file` into the same directory as the vocabulary right? That seems like a sensible suggestion, particularly if you think there are cases when a tokenizer comes with a `merges_file`, but a user wouldn't want to use it. If that's never going to be the case we could maybe just remove it from the `__init__` and hardcode loading it if it exists.

Could we also change `vocab_file` to just `path`, which means we can specify the config for this tokenizer the same as we would any other?

This is a slight tangent, and correct me if I'm wrong here, but the `tokenizer.json` format is for the fast fast BPE format, and from a quick look around, we don't support this yet for Llama models, right? Is it worth opening a broader discussion about generalising this tokenizer?

I think we also have some precedent for generally specifying custom tokens (like `eos_id`) in the codebase. cc @joecummings for thoughts here (and to generally make sure I'm not tripping).

torchtune/models/qwen2/_model_builders.py
SalmanMohammadi: Thanks for the insight!

This brings me to a broader point and feel free to ignore me if this sounds crazy. If we don't think there would be interest from the community in the 0.5B/1.5B models, which do use weight tying, would we be able to simplify this PR a bunch by just using the original `TransformerDecoder` for the 7B and 72B models, which don't need weight tying? 

If you think it's worth offering this flexibility I think the added decoder is fine, and it'll be nice to share it with Gemma. (also cc @joecummings @ebsmothers here).

SalmanMohammadi:
I've also added a few comments (but feel free to take these with a pinch of salt since I don't bring the experience that the maintainers do!). 

This PR is great and you've generally fit everything into the codebase well. I've raised some general points about whether it's worth generalising the fast BPE tokenizer that you've implemented, and whether there's interest in the smaller weight-tied models, or a need for flexibility around weight tying in the larger models; I'd love to get some input on these points.

RdoubleA (commented):
Sorry I'm late to the party - thank you for all the excellent work here! This was a huge effort to add an important model to the library, and you've integrated it into the codebase well.

One general comment I had on tied embeddings - do you know if users will ever want to use Qwen2 without tied embeddings? I noticed on HF this defaults to False. Asking because assuming it's always tied will simplify a lot of the builder logic and typing. For Gemma for example, we took a stance that it will always be tied, but I don't know enough about Qwen to make that call.

On generalizing the fast BPE tokenizer here - I would vote to keep it in Qwen until another model needs to use the same GPT-2 BPE tokenizer (cc @SalmanMohammadi)

torchtune/models/qwen2/_model_builders.py
joecummings: Yeah, we've made it pretty far in this PR so I think best course of action is to land this with the weight tying; however, for future reference @SalmanMohammadi is absolutely correct that we prefer to start by landing *only* what is needed, which - in this case - would just be regular Qwen2 without weight tying.

This code will make it easy to add the small models, which is a huge benefit for our memory constrained users, so I'm excited for the quick follow up with those configs :)

torchtune/models/qwen2/_tokenizer.py
```diff
@@ -0,0 +1,357 @@
...
+    return pairs
+
+
+class Qwen2Tokenizer(ModelTokenizer):
+    """This class construct a Qwen2 tokenizer, based on GPT-2 byte-level BPE tokenization.
+
+    See <https://github.com/huggingface/transformers/blob/v4.40.1/src/transformers/models/qwen2/tokenization_qwen2.py>.
+
+    Args:
+        vocab_file (str): Path to vocab.json file.
```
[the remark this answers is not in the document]
fyabc: sorry, forget to update the docstring... I will fix it later

torchtune/models/qwen2/_tokenizer.py
```diff
@@ -0,0 +1,357 @@
...
+        merges_file (str): Path to merges.txt file.
+
+    Example:
+        >>> tokenizer = Qwen2Tokenizer(vocab_file="/path/to/vocab.json", merges_file="/path/to/merges.txt")
+        >>> tokenized_text = tokenizer.encode("Hello world!")
+        >>> print(tokenized_text)
+        []
```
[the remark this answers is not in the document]
fyabc: sorry, the doctest is directly copied from other class. I will replace it with correct output.

torchtune/models/qwen2/_tokenizer.py
```diff
@@ -0,0 +1,357 @@
...
+    """
+
+    system = f"{IM_START}system\n{{content}}{IM_END}\n"
+    user = f"{IM_START}user\n{{content}}{IM_END}\n"
+    assistant = f"{IM_START}assistant\n{{content}}{IM_END}\n"
+    assistant_for_generation = f"{IM_START}assistant\n"
```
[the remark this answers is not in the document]
fyabc: As mentioned in this PR, Qwen2 uses ChatML template, but the exist `torchtune.data._chat_formats.ChatMLFormat` class will add an extra eos token (see notes [here](https://github.com/pytorch/torchtune/pull/1143#issue-2392523051)), so I reimplement the template in Qwen2Tokenizer.
(If `ChatMLFormat` can fix the bug of eos token, I think it is also a good choice to use `ChatMLFormat` directly.)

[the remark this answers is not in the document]
RdoubleA: Ah ok, thanks for explaining. Sorry I missed that. I created an issue to track this: #1194. Let's leave the template you have here as is.

recipes/configs/qwen2/7B_full.yaml
```diff
@@ -0,0 +1,78 @@
+# Config for multi-device full finetuning in full_finetune_distributed.py
```
SalmanMohammadi, line 1: Could you add these to the recipe registry please?

https://github.com/pytorch/torchtune/blob/main/torchtune/_recipe_registry.py

fyabc:
> Hi @fyabc! Thanks for your patience as we reviewed this PR - we're still really excited to get the Qwen2 model into torchtune.
> 
> Would you mind taking another quick look at the comments by @SalmanMohammadi and @RdoubleA ? Let me know how else I can help.

Sorry, I am busy on other work last week and forgot to reply. I will update my code on these comments in few days.

joecummings:
> Sorry, I am busy on other work last week and forgot to reply. I will update my code on these comments in few days.

No worries at all! Please do let us know if there's anyway we can help support you further - we appreciate all the work you've already done here

torchtune/models/qwen2/_tokenizer.py
```diff
@@ -0,0 +1,357 @@
...
+                line = line.strip()
+                if (i == 0 and line.startswith("#version:")) or not line:
+                    continue
+                bpe_merges.append(tuple(line.split()))
+        self.bpe_ranks = dict(zip(bpe_merges, range(len(bpe_merges))))
+        # NOTE: the cache can grow without bound and will get really large for long running processes
+        # (esp. for texts of language that do not use space between word, e.g. Chinese); technically
+        # not a memory leak but appears as one.
+        # GPT2Tokenizer has the same problem, so let's be consistent.
+        self.cache = {}
```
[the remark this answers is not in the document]
fyabc: Already changed to `lru_cache()` and add a configurable cache size argument.

fyabc:
> Thanks so much for the updates @fyabc. This is looking great. As per @joecummings's comment above, would it be possible to verify training works as expected?
> 
> > The individual unit tests are great for proving correctness of individual models, but a E2E run of this recipe on alpaca and comparing the loss curves between this and HuggingFace is a fantastic sanity check to make sure the torchtune users are getting the expected experience. It's also useful to output both loss curves to a Weights & Biases run so we can visually see that they're similar.
> 
> I think once you've had a chance to compare a training run, we should be good to land this (though Joe and Rafi may have some more input). Please let us know if there's more we can do to help - we really appreciate all the work you've put in here.
> 

Thank you for your suggestion! I am running finetune recipes and compare them with the official huggingface finetune script. I will report experiment results to wandb and add the share link to PR description.

torchtune/modules/transformer.py
```diff
@@ -246,3 +247,152 @@ def forward(
         # shape: [b, s, out_dim] - out_dim is usually the vocab size
         output = self.output(h).float()
         return output
+
+
+class TiedEmbeddingTransformerDecoder(nn.Module):
```
ebsmothers, line 252: Now that this is in transformer.py as a core module we need to add a unit test for it. I don't think we need to block this PR on it, so filed #1241 as a follow-up.

recipes/configs/qwen2/7B_full.yaml
```diff
@@ -0,0 +1,78 @@
...
+# using a Qwen2 7B model
+#
+# This config assumes that you've run the following command before launching
+# this run:
+#   tune download Qwen/Qwen2-7B-Instruct --output-dir /tmp/Qwen2-7B-Instruct
```
ebsmothers, line 6: I believe the download commands also need to add the `--ignore-patterns` flag to properly download safetensors files (this is slightly annoying but we need to do things this way for the time being)
```suggestion
#   tune download Qwen/Qwen2-7B-Instruct --output-dir /tmp/Qwen2-7B-Instruct --ignore-patterns ""
```

torchtune/models/qwen2/_component_builders.py
```diff
@@ -0,0 +1,450 @@
...
+    num_layers: int,
+    num_heads: int,
+    num_kv_heads: int,
+    embed_dim: int,
+    intermediate_dim: int,
+    max_seq_len: int,
+    attn_dropout: float = 0.0,
+    norm_eps: float = 1e-5,
+    rope_base: float = 1_000_000.0,
+    tie_word_embeddings: bool = False,
```
ebsmothers, line 50: Out of curiosity, is this something we anticipate people to experiment with? I think it's fine to support but (a) it can have implications on checkpointing and/or FSDP wrapping logic, and (b) we are relaxing the contract of the builder function a bit to do it. Not a huge concern, but wanna make sure that it's valuable to add this kind of configurability.

ebsmothers (approved):
This looks great! I left a few small comments but none of them are blocking. Would like clarification on the point about making `tie_word_embeddings` configurable, but regardless I think we can address anything else in a follow-up. Thank you so much for contributing Qwen2 to our library!

torchtune/models/qwen2/_component_builders.py
ebsmothers: Edit: nevermind on this point. I see that you mention it being relevant for 0.5B and 1.5B size models. So I think the approach makes sense

PR #1147, opened 2024-07-08
Making `add_eos` optional in text completion dataset

SalmanMohammadi:
#### Context
- [x] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

#### Changelog

Adding optional `add_eos` to text completion dataset with default set to `True`. Adding this as a seperate PR to support my PPO work https://github.com/pytorch/torchtune/pull/1005. CC @RdoubleA 

#### Test plan


tests/torchtune/datasets/test_text_completion_dataset.py
```diff
@@ -45,3 +45,23 @@ def test_get_item(self, mock_load_dataset):
             prompt, label = dataset[i]["tokens"], dataset[i]["labels"]
             assert prompt == self.expected_tokenized_prompts[i]
             assert label == expected_labels[i]
+
+    @mock.patch("torchtune.datasets._text_completion.load_dataset")
+    def test_get_item_no_eos(self, mock_load_dataset):
+        mock_load_dataset.return_value = self.get_samples()
+        expected_labels = self.expected_tokenized_prompts
+
+        dataset = TextCompletionDataset(
```
joecummings, line 54: IIUC, this test is just ensuring that a user *can* call the `TextCompletionDataset` with the param `add_eos=False`. However, we would actually want to be able to test that the EOS token is not added in this case.

joecummings (commented):
Overall this looks fantastic and I think is a great value add to the API! Just one comment on how we test this :)

tests/torchtune/datasets/test_text_completion_dataset.py
SalmanMohammadi: apologies if I'm missing something - does`assert prompt == self.expected_tokenized_prompts[i][:-1]` not test this? I'm trimming the expected output to omit the EOS token

joecummings: Ahhhh, I missed that you were trimming it to omit the EOS token - smart! Can you leave a comment in the test calling that out?

SalmanMohammadi: for sure! all good now

joecummings (approved):
lgtm - will merge once CI is green.

PR #1146, opened 2024-07-08
Fused SelfAttention and Cross Attention Decoder

pbontrager:
#### Context
What is the purpose of this PR? Is it to
- [x] add a new feature
- [ ] fix a bug
- [ ] update tests and/or documentation
- [ ] other (please add here)

This adds a prototype implementation of a flamingo model. This introduces fused layers and fused embeddings for combining cross attention layers with a decoder language model backbone. In the included component builders, this PR uses Llama3 as the reference.

#### Changelog
- Flamingo component builders - llama3 based decoder component
- modules/model_fusion - modules for fusing layers between pre-trained model, loosely follows peft design

Out of scope: flamingo encoder, utilities for manipulating fused model parameters

#### Test plan
This is a new experimental component builder, it will be tested once a model builder is put together.

- [x] run pre-commit hooks and linters (make sure you've first installed via `pre-commit install`)
- [ ] add unit tests for any new functionality
- [x] update docstrings for any new or updated methods or classes
- [x] run unit tests via `pytest tests`
- [x] run recipe tests via `pytest tests -m integration_test`
- [ ] manually run any new or modified recipes with sufficient proof of correctness
	- [ ] include relevant commands and any other artifacts in this summary (pastes of loss curves, eval results, etc.)

PR #1150, opened 2024-07-08
[Flamingo][multimodal] Vision encoder + text decoder

felipemello1:
#### Context
- [x] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

- Added flamingo vision encoder
- Ported text decoder from: https://github.com/pytorch/torchtune/pull/1146
- Updated CLIP attention module

TODO: 
- [x] clip transformer has to be replaced with TransformerSelfAttentionLayer
- [ ] Confirm location and naming of all new transformer related modules
- [ ] regression tests
- [ ] parity check
- [ ] docstrings
- [ ] builder with text decoder + vision encoder (prob in another PR)

#### Changelog

torchtune/models/clip/_component_builders.py
- Replaced pytorch native transformer with torchtunes version;

torchtune/models/flamingo/_encoders.py
- FlamingoVisionAdapter: takes in clip embedding and outputs final vision embedding
- FlamingoVisionEncoder: wrapper to call Clip + adapter

torchtune/models/flamingo/_component_builders.py
- flamingo_vision_encoder: instantiates clip_vision_encoder + FlamingoVisionAdapter and returns FlamingoVisionEncoder
- flamingo_text_decoder: instantiates llama3 + TransformerCrossAttentionLayer using FusionLayer and returns MMTransformerDecoder

torchtune/modules/feedforward.py
- MLP: simple feed forward layer for the transformer MLP. Used by CLIP and Flamingo.

torchtune/modules/multimodal_transformer.py
- TransformerSelfAttentionLayer: used in flamingo with gates. Used in CLIP as vanillas transformer.
- TransformerCrossAttentionLayer: used in flamingo_text_decoder
- MMTransformerDecoder: used in flamingo_text_decoder

torchtune/modules/attention.py
- GroupedQueryAttention: used in flamingo_text_decoder, clip, flamingo_vision_encoder. IMPORTANT: llama3 in flamingo will use this, while llama3-text-only will use CausalSelfAttention

torchtune/modules/model_fusion.py
- fusion_embed.py: used in flamingo_text_encoder
- fusion_layer.py: used in flamingo_text_encoder

torchtune/modules/tanh_gate.py
- TanhGate: used in flamingo vision adapter and cross encoder.

#### Test plan
- Clip tests pass, except one regression.
- Flamingo vision encoder shapes pass, need regression.
- Flamingo text decoder instantiates, but no shape or regression tests yet.
- Need tests for the individual new modules. But testing the models as whole should work for now.


torchtune/models/flamingo/_encoders.py
```diff
@@ -0,0 +1,121 @@
...
+        self,
+        in_dim: int,
+        hidden_dim: int,
+        out_dim: int,
+        act_layer: nn.Module,
+        dropout=0.0,
+    ):
+        super().__init__()
+        self.layer1 = nn.Linear(in_dim, hidden_dim)
+        self.layer2 = nn.Linear(hidden_dim, out_dim)
```
ebsmothers, line 113: Any reason we can't pass in nn.Modules to the class as in the [existing FeedForward class](https://github.com/pytorch/torchtune/blob/069b12bef0b9cf735d5fb7cdc4192bfbf9abd764/torchtune/modules/feed_forward.py#L25-L27)? That'll make it easier for us to provide builders for LoRA-fied versions of these

torchtune/models/flamingo/_encoders.py
```diff
@@ -0,0 +1,121 @@
...
+                mlp_scale=TanhGate(),
+            )
+
+            transformer_layers.append(layer)
+
+        self.projection = nn.Linear(proj_in, proj_out)
+
+    def forward(self, x: torch.Tensor, hidden_states: torch.Tensor) -> torch.Tensor:
+
+        bsz, n_ims, n_tiles, n_tokens, embed_dim = x
```
ebsmothers, line 73: Should be this?
```suggestion
        bsz, n_ims, n_tiles, n_tokens, embed_dim = x.shape
```

torchtune/models/flamingo/_encoders.py
```diff
@@ -0,0 +1,121 @@
...
+
+        # transformer
+        x = x.view(bsz * n_ims, n_tiles * n_tokens, embed_dim)
+        for layer_idx, transformer_layer in enumerate(self.transformer_layers):
```
ebsmothers, line 77: nit: don't enumerate if you're not actually using layer_idx

torchtune/models/flamingo/_encoders.py
```diff
@@ -0,0 +1,121 @@
...
+            x = transformer_layer(x)
+
+        # projection
+        x = x.view(bsz, n_ims, n_tiles, n_tokens, embed_dim)
+        x = torch.cat([x, hidden_states], dim=-1)
```
ebsmothers, line 82: Add shape comment here (and more generally need docstrings for these classes)

torchtune/models/flamingo/_encoders.py
```diff
@@ -0,0 +1,121 @@
...
+
+from torchtune.modules import (
+    Fp32LayerNorm,
+    GroupedQueryAttention,
+    TanhGate,
+    TransformerSelfAttentionLayer,
+)
+
+
+class FlamingoVisionAdapter(nn.Module):
```
ebsmothers, line 19: I don't think this follows the design pattern of our other components. Why can't we just pass layer and projection as nn.Modules to the init so that we keep things generic? Then use builders to actually construct the lower-level components and piece them all together

torchtune/modules/attention.py
```diff
@@ -227,3 +227,233 @@ def forward(
         # reshape the output to be the same shape as the input
         output = output.transpose(1, 2).contiguous().view(bsz, seq_len, -1)
         return self.output_proj(output)
+
+
+class GroupedQueryAttention(nn.Module):
```
ebsmothers, line 232: It's not clear to me why this is in modules/attention.py but other stuff goes in models/flamingo. I would suggest moving this under models/flamingo too to be consistent unless there's a clear reason not to.

torchtune/modules/multimodal_transformer.py
```diff
@@ -0,0 +1,206 @@
...
+
+    def __init__(
+        self,
+        attn: GroupedQueryAttention,
+        mlp: nn.Module,
+        *,
+        attn_norm: Optional[nn.Module] = None,
+        mlp_norm: Optional[nn.Module] = None,
+        attn_scale: Optional[nn.Module] = None,
+        mlp_scale: Optional[nn.Module] = None,
```
ebsmothers, line 35: Sorry to keep harping on this, but I still don't understand why we don't just provide separate versions of self-attention and MLP modules with scaling (see e.g. [here](https://gist.github.com/ebsmothers/7ed8a319b7566b76868299e245f5995b) for what the MLP would look like). Then we can provide different builders for `TransformerSelfAttentionLayer` with and without scaling and users don't have to try and figure out `attn_scale` and `mlp_scale` mean

torchtune/models/flamingo/_encoders.py
kartikayk: Yeh it would be good to add some quick tests to make sure these are working as expected

torchtune/modules/multimodal_transformer.py
```diff
@@ -0,0 +1,206 @@
...
+
+from typing import Optional
+
+import torch
+from torch import nn, Tensor
+
+from torchtune.modules import GroupedQueryAttention
+
+
+class TransformerSelfAttentionLayer(nn.Module):
```
kartikayk, line 15: Would be good to just rename this tp have "Multimodal" in the prefix

torchtune/modules/attention.py
kartikayk: So our "primary" attention implementation already supports GQA and MQA [[ref](https://github.com/pytorch/torchtune/blob/main/torchtune/modules/attention.py#L189)]. I guess the question here is if this entire module is specific to flamingo, should this just reside within that folder till we figure out how to merge back? That said, I'll raise the same concern as I did in Philip's PR that GQA isn't a great name for this module. I'd prefer something like MMMultiHeadAttention or similar

ebsmothers: Yeah our naming here is kinda unclear. Like we already support GQA but via our `CausalSelfAttention` class. So any new functionality here is mainly to support cross-attention. So as is the naming of these two classes is not actually conveying how they differ. My two cents is that we should just call them `MultiHeadAttention` and `MultiHeadCrossAttention` to get this point across, but I think the renaming of `CausalSelfAttention` can be saved for another day.

torchtune/modules/multimodal_transformer.py
felipemello1: This is also used in clip. I think that we should only use MM for modules that touch at the same time text and image. This is not the case for this module. What do you think?

torchtune/modules/multimodal_transformer.py
felipemello1: I think that both ways work. I like your idea, but i also think its convenient if the module already provides it. I guess that one benefits of this logic being in the transformer module is that we dont have to have 2x the implementation of every scale and MLP module, one gated and one that isnt. The con is that this module gets a bit bloated.

torchtune/modules/attention.py
felipemello1: will rename it after we are 100% sure :P

PR #1151, opened 2024-07-09
Miscellaneous CI, dependency, and version fixes

ebsmothers:
Consolidating a bunch of small changes into this PR:

- run unit tests on GPUs in our CI
- define a util for ao version checks (this will come in handy once https://github.com/pytorch/ao/pull/485 lands)
- remove torchvision from pyproject.toml and add separate install instructions to fix [this issue](https://github.com/pytorch/torchtune/commit/06a125efb7d1c65bbfe59763d57761bf46966eb7#r143974775) with forced downgrade of PyTorch nightlies
- run CI on the combo of ao + PyTorch nightlies
  - Another thing we could consider is to remove ao from our pyproject.toml, similar to what I've added here for torchvision, then give the separate install instructions depending on whether users want nightly or stable versions. Otherwise it's a bit hacky because we technically install both stable and nightly versions of ao in our install flow (but nightly will supersede stable). Punting this for now since I don't think it'll break anything

CI should be green


torchtune/utils/quantization.py
```diff
@@ -10,10 +10,18 @@
...
     Int4WeightOnlyGPTQQuantizer,
     Int4WeightOnlyQuantizer,
-    quantize,
     Quantizer,
 )
 
+from torchtune.modules.low_precision._utils import _get_torchao_version
+
+ao_version, is_nightly = _get_torchao_version()
+print(ao_version, is_nightly)
```
rohan-varma, line 19: nit: remove this before landing?

tests/torchtune/utils/test_distributed.py
```diff
@@ -262,6 +262,10 @@ def world_size(self) -> int:
         return 2
 
     @gpu_test(gpu_count=2)
+    @pytest.mark.skipif(
+        version.parse(torch.__version__).base_version < "2.4.0",
+        reason="torch >= 2.4 required",
```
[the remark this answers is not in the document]
ebsmothers: Some of the DTensor APIs we use in [load_from_full_model_state_dict](https://github.com/pytorch/torchtune/blob/069b12bef0b9cf735d5fb7cdc4192bfbf9abd764/torchtune/utils/_distributed.py#L287) were not stable prior to 2.4. We already addressed this for the QLoRA state dict test in #1087. In this case it's OK because we are testing FSDP2 functionality which is not available until 2.4 anyways. cc @weifengpy in case I'm missing any important points here.

pyproject.toml
```diff
@@ -10,8 +10,6 @@ authors = [
 ]
 keywords = ["pytorch", "finetuning", "llm"]
 dependencies = [
-    # multimodality
-    "torchvision",
```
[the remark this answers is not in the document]
ebsmothers: Just chatted with @NicolasHug on this and he confirmed it's not possible since there's no way to point pyproject.toml to a specific conda channel or PyPI repo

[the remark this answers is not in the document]
felipemello1: got it! i am just afraid that that in the long run it may cause issues, since we may want to pin torchvision version. For example, in ClipTransforms, older versions will break. 

brainstorming:
Maybe we can also force it to also install torchvision nightlies, like you force pytorch nightlies?

For example, in the [link you shared](https://gist.github.com/ebsmothers/dc78853f03d89c156673a9e84a321238), it installed version 0.18. [But nightly is 0.20](https://pytorch.org/vision/versions.html).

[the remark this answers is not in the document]
ebsmothers: It should be installed in a single command, this is why I updated the readme to clarify this. So if the user runs `pip install torch torchvision` they will get stable versions of both torch and torchvision; if they run `pip3 install --pre torch torchvision --index-url https://download.pytorch.org/whl/nightly/cu121` they will get torch nightly and torchvision 0.20. 

I can't find the link to the original comment but I believe that was demonstrating what happens when we install using the existing pyproject.toml. For reference, [here](https://gist.github.com/ebsmothers/5f377798d834fc76b4823c5557584c08) is my pip list after running the first command; [here](https://gist.github.com/ebsmothers/9d4ed23ab72955be639fdc20d03a5809) is my pip list after running the second command. You can see that the versions are as expected.

Btw regarding pinning versions -- we do not test on anything older than the latest stable version of PyTorch and I don't think we want to worry about breaking folks on older versions than that. By the same logic, I don't think we should be pinning to older versions of torchvision. Then the best way to keep things in sync is just install the two together using these commands.

.github/workflows/gpu_test.yaml
```diff
@@ -1,4 +1,4 @@
-name: Multi-GPU Recipe Tests
```
joecummings, line 1: Why the rename here?

.github/workflows/recipe_test_nightly.yaml
```diff
@@ -4,6 +4,7 @@ on:
   schedule:
     # Runs at midnight every day
     - cron:  '0 0 * * *'
+  workflow_dispatch:
```
joecummings, line 7: yay

README.md
```diff
@@ -156,7 +156,16 @@ You can find a full list of all our Llama3 configs [here.](recipes/configs/llama
 
 ## Installation
 
-**Step 1:** [Install PyTorch](https://pytorch.org/get-started/locally/). torchtune is tested with the latest stable PyTorch release as well as the preview nightly version.
+**Step 1:** [Install PyTorch](https://pytorch.org/get-started/locally/). torchtune is tested with the latest stable PyTorch release as well as the preview nightly version. For multimodality
+be sure to also install torchvision.
+
+```
+# Install stable version of PyTorch using pip
+pip3 install torch torchvision
```
joecummings, line 164: I don't think we need to explicitly say `pip3` anymore

joecummings (approved):

pyproject.toml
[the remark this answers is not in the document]
felipemello1: thanks for the links and explanation!

README.md
```diff
@@ -156,7 +156,16 @@ You can find a full list of all our Llama3 configs [here.](recipes/configs/llama
 
 ## Installation
 
-**Step 1:** [Install PyTorch](https://pytorch.org/get-started/locally/). torchtune is tested with the latest stable PyTorch release as well as the preview nightly version.
+**Step 1:** [Install PyTorch](https://pytorch.org/get-started/locally/). torchtune is tested with the latest stable PyTorch release as well as the preview nightly version. For multimodality
```
kartikayk, line 159:
```suggestion
**Step 1:** [Install PyTorch](https://pytorch.org/get-started/locally/). torchtune is tested with the latest stable PyTorch release as well as the preview nightly version. For fine-tuning the multimodal LLMs available in the repo, you'll need to install torchvision as well
```

kartikayk (commented):
Did we align on AO being an optional dependency? If so, why not do what we do with BnB example and ask users to manually install?

ebsmothers:
> Did we align on AO being an optional dependency? If so, why not do what we do with BnB example and ask users to manually install?

@kartikayk I think we aligned on actually testing ao nightlies but not on having it as an optional dependency. So I'm doing the former here and not the latter. But cc @msaroufim @joecummings if either of you have thoughts on this

.github/workflows/gpu_test.yaml
ebsmothers: It's no longer just a recipe test, right? Now it's recipe + unit test

PR #1156, opened 2024-07-09
Multimodal collater with interleaved image, cross-attention mask padding

RdoubleA:
## Context
Add the batch collater for multimodal image + text datasets. The collater supports the following:
- Tiled images
- Multiple images per sample
- Cross-attention masks

Inputs must be samples from the multimodal dataset post tiling, post transform:
- "tokens": List[int] of length text_seq_len, varies across samples
- "labels": List[int] of length text_seq_len, varies across samples
- "images": List[Tensor], each with shape (n_tiles, c, h, w)
- "encoder_mask": List[Tensor], each with shape (text_seq_len, image_seq_len)
- "aspect_ratio": List[Tensor], each with shape (h_ratio, w_ratio)

It performs the following actions:
(1) Pad text sequence and encoder mask to the longest sequence length in the batch
(2) Pad image tensors in the tile dimension with zeros to the largest number of tiles in the batch
(3) Add empty images of zeros to samples up to max number of images in the batch
(4) Pad aspect ratios with (1,1) for all added padding images

**Feedback requested**: 
- The name is a bit verbose, but padded_collate_multimodal is too vague as well
- We are padding numerous dimensions here, requiring nested for loops with runtime O(total num of images in batch) and running it twice. Would be great to see if we can simplify/optimize/vectorize it further

## Changelog
- Added the collate function + unit test

## Test plan
`pytest tests/torchtune/utils/test_collate.py`

## Docs
image
image


torchtune/data/_collate.py
```diff
@@ -214,6 +214,142 @@ def padded_collate_sft(
...
+) -> Dict[str, torch.Tensor]:
+    """Pad a batch of text sequences, tiled image tensors, aspect ratios,
+    and cross attention masks.
+
+    ``batch`` is expected to be a list of sample dicts containing the following:
+        - "tokens": List[int] of length text_seq_len, varies across samples
+        - "labels": List[int] of length text_seq_len, varies across samples
+        - "images": List[Tensor], each with shape (n_tiles, c, h, w)
+        - "encoder_mask": List[Tensor], each with shape (text_seq_len, image_seq_len)
+        - "aspect_ratio": List[Tensor], each with shape (h_ratio, w_ratio)
```
ebsmothers, line 230: The shape is just (2,) right? Also maybe worth saying that len(images) == len(encoder_mask) == len(aspect_ratio) for each element in the batch

torchtune/data/_collate.py
```diff
@@ -214,6 +214,142 @@ def padded_collate_sft(
...
+    Args:
+        batch (List[Dict[str, Any]]): A list of sample dicts containing tokens,
+            labels, images, encoder_mask, and aspect_ratio.
+        padding_idx (int): Padding index for input token ids. Defaults to 0.
+        ignore_idx (int): Padding index for labels. Defaults to -100.
+
+    Returns:
+        Dict[str, Tensor]: Collated tokens, labels, images, encoder_mask, aspect_ratio tensors.
+
+    Example:
```
ebsmothers, line 248: 10/10 example

torchtune/data/_collate.py
```diff
@@ -214,6 +214,142 @@ def padded_collate_sft(
...
+        (4) Pad aspect ratios with (1,1) for all added padding images
+
+    Args:
+        batch (List[Dict[str, Any]]): A list of sample dicts containing tokens,
+            labels, images, encoder_mask, and aspect_ratio.
+        padding_idx (int): Padding index for input token ids. Defaults to 0.
+        ignore_idx (int): Padding index for labels. Defaults to -100.
+
+    Returns:
+        Dict[str, Tensor]: Collated tokens, labels, images, encoder_mask, aspect_ratio tensors.
```
ebsmothers, line 246: Can you add shapes here? Esp for the image fields

torchtune/data/_collate.py
```diff
@@ -214,6 +214,162 @@ def padded_collate_sft(
...
+        >>> image_id = 1
+        >>> tokens_per_tile = 5
+        >>> c, h, w = 1, 1, 1
+        >>> batch = [
+        ...     {
+        ...         "tokens": [1, 2, 1, 3], "labels": [4, 5, 6, 7],
+        ...         "encoder_input": {
+        ...             # One image with two tiles, one image with three tiles
+        ...             "images": [torch.ones(2, c, h, w), torch.ones(3, c, h, w)],
+        ...             "aspect_ratio": [torch.tensor([1, 2]), torch.tensor([1, 2])],
```
pbontrager, line 268: If there are three tiles, the aspect ratio would multiply to 3. So [3, 1] or [1, 3] are the two options for the second image.

torchtune/data/_collate.py
```diff
@@ -214,6 +214,162 @@ def padded_collate_sft(
...
+        ...         },
+        ...         "encoder_mask": [torch.ones(4, 5 * 2), torch.ones(4, 5 * 3)],
+        ...     },
+        ...     {
+        ...         "tokens": [1, 4], "labels": [8, 9],
+        ...         "encoder_input": {
+        ...             # One image with four tiles
+        ...             "images": [torch.ones(4, c, h, w)],
+        ...             "aspect_ratio": [torch.tensor([1, 2])],
```
pbontrager, line 277: [2, 2]

torchtune/data/_collate.py
```diff
@@ -214,6 +214,162 @@ def padded_collate_sft(
...
+        ...         },
+        ...         "encoder_mask": [torch.ones(2, 5 * 4)],
+        ...     },
+        ... ]
+        >>> model_inputs = padded_collate_vision_text(batch=batch)
```
pbontrager, line 282: This name doesn't match the current name. I actually prefer padded_collate_vision_text as it's more straight forward and we can either generalize this function or split and rename as we get more vision_text models in the future.

torchtune/data/_collate.py
```diff
@@ -214,6 +214,162 @@ def padded_collate_sft(
...
+        >>> print(model_inputs["tokens"])
+        tensor([[1, 2, 1, 3],
+                [1, 4, 0, 0]])
+        >>> print(model_inputs["labels"])
+        tensor([[4, 5, 6, 7],
+                [8, 9, -100, -100]])
+        >>> print(model_inputs["encoder_input"]["images"].shape)  # (bsz, max_num_images, max_num_tiles, c, h, w)
+        torch.Size([2, 2, 4, 1, 1, 1])
+        >>> print(model_inputs["encoder_mask"].shape)  # (bsz, max_num_images, max_num_tiles, tokens_per_tile * max_num_tiles)
```
pbontrager, line 291: This should actually be [2, 4, 40] since cross attention is text vs image sequence and the image sequence is num_images*num_tiles*tokens_per_tile.

torchtune/data/_collate.py
```diff
@@ -214,6 +214,162 @@ def padded_collate_sft(
...
+        torch.Size([2, 2, 4, 20])
+        >>> print(model_inputs["encoder_input"]["aspect_ratio"].shape)  # (bsz, max_num_images, 2)
```
pbontrager, line 293: I'm not sure if this should be [2, 2, 2] or [2, 4]. @felipemello1 ?

pbontrager (commented):
Really good way to handle a complicated topic. I think we should name this padded_collate_vision_text for now and pretend that it's general until it's not. My suspicion is that we'll have to move this into the model folder in the future as it's so model specific. After you address the comments I left, I'll test this for you on my script.

torchtune/data/_collate.py
felipemello1: aspect_ratio should be (bsz, max_num_images, 2), and then in the clip we reshape:

aspect_ratio = aspect_ratio.reshape(bsz_and_n_imgs, 2)

https://github.com/pytorch/torchtune/blob/82c232d0679ddef3fc419cdc18af758b98b4da05/torchtune/modules/vision_transformer.py#L354C9-L354C63

torchtune/data/_collate.py
```diff
@@ -214,6 +214,162 @@ def padded_collate_sft(
...
+        tensor([[[[0.]]], [[[0.]]], [[[0.]]], [[[0.]]]])
+    """
+    # Text tokens can be handled independently by existing collater
+    text_only = [
+        {"tokens": sample["tokens"], "labels": sample["labels"]} for sample in batch
+    ]
+    collated_text = padded_collate_sft(text_only, padding_idx, ignore_idx)
+    max_seq_len = collated_text["tokens"].shape[-1]
+
+    # TODO: Figure out how to make this more efficient or vectorized. Setting
```
felipemello1, line 311: didnt think too much about it, but maybe:
1. do a first pass to check the max of each dimension.
2. create a tensor with all zeros
3. Add the input to the tensor correct line: eg. tensor[0] += sample

torchtune/data/_collate.py
RdoubleA: I still don't fully understand aspect ratio so I will take your word

RdoubleA:
> I think we should name this padded_collate_vision_text for now and pretend that it's general until it's not.

I actually think we should go the other way and keep this overspecified until we have a concrete use case to generalize it, I'd rather be overspecific than mislead users into thinking this can be used for any multimodal model

torchtune/data/_collate.py
RdoubleA: but what about batch size?

torchtune/data/_collate.py
RdoubleA: Pre-allocating would definitely simplify the code. I would still need to loop through each individual image though

RdoubleA: I will leave this as a follow-up though in the interest of time

torchtune/data/_collate.py
```diff
@@ -214,6 +214,162 @@ def padded_collate_sft(
...
+    ignore_idx: int = CROSS_ENTROPY_IGNORE_IDX,
+) -> Dict[str, torch.Tensor]:
+    """Pad a batch of text sequences, tiled image tensors, aspect ratios,
+    and cross attention masks.
+
+    ``batch`` is expected to be a list of sample dicts containing the following:
+        - "tokens": List[int] of length text_seq_len, varies across samples
+        - "labels": List[int] of length text_seq_len, varies across samples
+        - "encoder_input": Dict[str, List[torch.Tensor]]
+            - "images": List[torch.Tensor], each with shape (n_tiles, c, h, w)
```
SalmanMohammadi, line 231: Can you say somewhere `c, h, w` = `channel, height, width`?

torchtune/data/_collate.py
SalmanMohammadi: `+ 1`

pbontrager (commented):
I've tested this and everything seems to be working accurately. The only issue right now is that this can't be used for inference. We need to expose padded_direction for text and not expect "labels" when padded_direction=="left". Also, I'd like to propose "padded_collate_tiled_images_and_mask".

pbontrager (approved):
Thank you!

PR #1158, opened 2024-07-10
[7/7] Multimodal datasets (The Cauldron, LLaVA-Instruct-150K)

RdoubleA:
## Changelog
- Add two example multimodal dataset builders, each with their own dataset transform: [The Cauldron](https://huggingface.co/datasets/HuggingFaceM4/the_cauldron) and [LLaVA-Instruct-150K](https://huggingface.co/datasets/liuhaotian/LLaVA-Instruct-150K). Both require slightly different preprocessing and serve as good examples
- Add a utility to quickly map raw text with image tags into what's expected for Message content field
- Upgrade `ShareGPTToMessages` with support for an image column

## Test plan
- Unit test for The Cauldron, LLaVA Instruct datasets
- Unit test for split_text_by_image_tag


torchtune/data/_utils.py
```diff
@@ -33,43 +31,30 @@ def truncate(
...
-
-    - There is a system message that's not the first message
-    - There are two consecutive user messages
-    - An assistant message comes before the first user message
-    - The message is empty
-    - Messages are shorter than length of 2 (min. one user-assistant turn)
-
+    Given a raw text string, split by the specified ``image_tag``
+    and form into list of dictionaries to be used in the ``Message`` content
+    field.
```
ebsmothers, line 38: Can you be a bit more explicit about the format? E.g. at least "image tags will be represented as a dictionary {"type": "image"}"

torchtune/data/_utils.py
```diff
@@ -33,43 +31,30 @@ def truncate(
...
+        >>> print(content)
+        [{"type": "image"}, {"type": "text", "content": "hello "}, {"type": "image"}, {"type": "text", "content": "world"}]
+    """
+    split_content = content.split(image_tag)
+    final_content_list = []
+    for i, substr in enumerate(split_content):
+        if len(substr) > 0:
+            final_content_list.append({"type": "text", "content": substr})
+        if i < len(split_content) - 1:
+            final_content_list.append({"type": "image"})
```
ebsmothers, line 58: So this will not support multiple consecutive image tags, right? If so that should be explicitly stated somewhere

torchtune/data/_messages.py
```diff
@@ -361,11 +380,31 @@ def __call__(self, sample: Mapping[str, Any]) -> Mapping[str, Any]:
...
+            image = Image.open(image_path)
+        elif "image" in sample and isinstance(sample["image"], Image.Image):
+            image = sample["image"]
+        else:
+            image = None
+
+        processed_sample = {"messages": messages}
+        if image is not None:
+            processed_sample["images"] = [image]
+        return processed_sample
```
ebsmothers, line 407: Maybe controversial but why do we need to put all this in the existing `ShareGPTToMessages` class? Then as a text-only user I have extra stuff to sort through, plus when we expose at the dataset level it feels kinda overloaded. Idk maybe there are obvious pros to doing it this way that I'm missing here (apart from the obvious one of less code duplication, which tbh I don't care too much about in this case)

torchtune/datasets/multimodal/_llava_instruct.py
```diff
@@ -0,0 +1,113 @@
...
+    Then, you must pass in the directory containing all the images to the ``image_dir`` argument
+    so each file can be loaded as PIL images in the transform. In the example above, you would
+    set ``image_dir="coco/train2017"``.
+
+    Args:
+        model_transform (Transform): model-specific transform that takes in a sample dict and applies custom
+            transforms on the keys. The tokenizer used by the model should be encapsulated in the model transform
+            and should operate on the "messages" field. Any model-specific image transforms should operate on
+            the "images" field. The keys returned by the model should be aligned with the
+            expected inputs into the model.
```
ebsmothers, line 51: This is very general and probably the most important part of setting this up correctly. I would try to give an explicit example here or something

torchtune/datasets/multimodal/_llava_instruct.py
```diff
@@ -0,0 +1,113 @@
...
+        column_map (Optional[Dict[str, str]]): a mapping from the expected columns ("conversations")
+            to the new column names in the dataset. If None, assume these are identical.
+            Default is None.
+        new_system_prompt (Optional[str]): if specified, prepend a system message. This can
+            serve as instructions to guide the model response. Setting this will OVERRIDE any system
+            messages already present in the dataset. Default is None.
+        image_tag (Optional[str]): if specified, split the raw text content by the specified ``image_tag``
+            and use placeholders for where the images are present in the text for proper tokenization. Set
+            this if your dataset contains images and uses a specific string (ex: "<image>") to indicate the
+            presence of an image. Leave this as None if your dataset does not contain images. Default is "<image>".
```
ebsmothers, line 66: When would one of these not contain images though?

torchtune/datasets/multimodal/_the_cauldron.py
```diff
@@ -0,0 +1,165 @@
...
+            define source as the data file type (e.g. "json", "csv", "text") and pass
+            in the filepath in ``data_files``. See `Hugging Face's
+            <https://huggingface.co/docs/datasets/en/package_reference/loading_methods#datasets.load_dataset.path>`_
+            ``load_dataset`` for more details. Default is ``HuggingFaceM4/the_cauldron``.
+        column_map (Optional[Dict[str, str]]): a mapping to change the expected "images"
+            and "texts" column names to the actual column names in the dataset. Default is None,
+            keeping the default column names.
+        new_system_prompt (Optional[str]): if specified, prepend a system message. This can
+            serve as instructions to guide the model response. Setting this will OVERRIDE any system
+            messages already present in the dataset. Default is None.
```
ebsmothers, line 124: I think I missed the original PR here, but this naming feels a bit unclear to me (though I guess the purpose of "new" is to emphasize that we're overriding any other system prompts). Also again would give an example of this

torchtune/datasets/multimodal/_the_cauldron.py
```diff
@@ -0,0 +1,165 @@
...
+        >>> cauldron_ds = the_cauldron_dataset(model_transform=model_transform, subset="ai2d")
+        >>> for batch in Dataloader(cauldron_ds, batch_size=8):
+        >>>     print(f"Batch size: {len(batch)}")
+        >>> Batch size: 8
+    """
+
+    message_transform = TheCauldronToMessages(
+        train_on_input=train_on_input,
+        column_map=column_map,
+    )
```
ebsmothers, line 149: Should be passing `new_system_prompt`? Also nit but would call it `to_messages` or something, since `model_transform` also involves `Message` but as an input

torchtune/datasets/multimodal/_the_cauldron.py
```diff
@@ -0,0 +1,165 @@
...
+                    content=[{"type": "text", "content": message["assistant"]}],
+                )
+            )
+
+        if self.new_system_prompt is not None:
+            messages = [
+                Message(
+                    role="system", content=self.new_system_prompt, masked=True, eot=True
+                )
+            ] + messages
```
ebsmothers, line 84: Based on this I take it Cauldron never has its own system prompt?

torchtune/datasets/multimodal/_the_cauldron.py
```diff
@@ -0,0 +1,165 @@
...
+
+
+class TheCauldronToMessages(Transform):
+    """
+    Construct messages from a sample formatted similarly to
+    `The Cauldron dataset<https://huggingface.co/datasets/HuggingFaceM4/the_cauldron>_`.
+
+    Image placeholders are prepended to the text in the ``Message`` content. Images in the
+    dataset are expected to be a list of a single PIL image, so they are simply passed through
+    with an optional column remapping if ``column_map`` is specified.
```
ebsmothers, line 22: Give examples please

torchtune/datasets/multimodal/_the_cauldron.py
```diff
@@ -0,0 +1,165 @@
...
+        train_on_input: bool = True,
+        column_map: Optional[Dict[str, str]] = None,
+        new_system_prompt: Optional[str] = None,
+    ):
+        self.train_on_input = train_on_input
+        self.new_system_prompt = new_system_prompt
+        if column_map is not None:
+            if "images" not in column_map:
+                raise ValueError(
+                    "column_map must contain 'images' as a key if specified"
```
ebsmothers, line 49: nit: this wording is unclear

torchtune/datasets/multimodal/_the_cauldron.py
```diff
@@ -0,0 +1,165 @@
...
+
+from typing import Any, Dict, Mapping, Optional, Union
+
+from torchtune.data import Message
+from torchtune.datasets._packed import PackedDataset
+from torchtune.datasets._sft import SFTDataset
+from torchtune.modules.transforms import Transform
+
+
+class TheCauldronToMessages(Transform):
```
ebsmothers, line 15: I do worry a bit that we're overfitting to these two datasets in our choice of naming. Like really iiuc llava_instruct means you have a local directory of images and the sample points to filenames and the cauldron means you have actual PILs in each sample. That seems more general than what we're indicating

torchtune/datasets/multimodal/_the_cauldron.py
```diff
@@ -0,0 +1,165 @@
...
+                Message(
+                    role="system", content=self.new_system_prompt, masked=True, eot=True
+                )
+            ] + messages
+
+        return {"messages": messages, "images": sample[self._column_map["images"]]}
+
+
+def the_cauldron_dataset(
+    model_transform: Transform,
```
ebsmothers, line 90: Maybe I'm being dense but looking at this I have no idea how I should define the model_transform to construct one of these

ebsmothers (commented):
Need to fix CI

torchtune/data/_messages.py
RdoubleA: It seemed general enough to include in a general transform, but yeah I can see this being overloaded. Some options are: 1) make a separate ShareGPT transform with images, 2) keep the llava instruct transform its own class until we find more examples that fit this format (don't generalize until we need to ;) )

torchtune/data/_utils.py
RdoubleA: hmmm.... it should. let me revisit

torchtune/datasets/multimodal/_llava_instruct.py
RdoubleA: once the flamingo transform lands, it would be easy to point to this

torchtune/datasets/multimodal/_llava_instruct.py
RdoubleA: mmm... I guess you're right, I suppose it doesn't make sense to make this configurable

torchtune/datasets/multimodal/_the_cauldron.py
RdoubleA: well, we only have one dataset example that fits this mold. so does it make sense to make it general at this point?

torchtune/datasets/multimodal/_the_cauldron.py
RdoubleA: the cauldron doesn't by default but we'd want it to be configurable if the user wants to add it

torchtune/datasets/multimodal/_the_cauldron.py
RdoubleA: Yep that's exactly it

torchtune/datasets/multimodal/_llava_instruct.py
ebsmothers: Can we add a TODO or a follow-up then? I wanna make sure it's very clear how to use these builders, rn I feel like it's not

torchtune/data/_messages.py
ebsmothers: OK yeah let's do (2). Plus we already have a Cauldron transform so I feel like it's most consistent to do it this way anyways

torchtune/datasets/multimodal/_the_cauldron.py
ebsmothers: Yeah I guess this is kinda inconsistent with my other comment anyways. Let's go with what you proposed there, keep the naming as is and have a separate transform for each of the cauldron and llava-instruct datasets

torchtune/datasets/multimodal/_llava_instruct.py
```diff
@@ -0,0 +1,191 @@
...
+        message_transform=message_transform,
+        split=split,
+        data_files=data_files,
+        **load_dataset_kwargs,
+    )
+    if packed:
+        if model_transform.max_seq_len is None:
+            raise ValueError(
+                "PackedDataset requires a max_seq_len to be set on the tokenizer."
+            )
```
ebsmothers, line 189: This part is a bit confusing to me. So model transform has a max_seq_len, but that's really just associated with the tokenizer, right? So do we pass that up from the tokenizer to the model transform (since it's composed of both tokenizer and image transform)?

torchtune/datasets/multimodal/_the_cauldron.py
```diff
@@ -0,0 +1,197 @@
...
+    from Hugging Face Datasets.
+
+    The Cauldron consists of numerous datasets. You must specify one of the datasets
+    using the ``subset`` argument.
+
+    Args:
+        model_transform (Transform): model-specific transform class that takes in a sample dict and applies custom
+            transforms on the keys. It should consist of at minimum two components: text tokenization (called
+            on the "messages" field) and image transform (called on the "images" field). The keys returned by
+            the model transform should be aligned with the expected inputs into the model.
```
ebsmothers, line 142: Sorry to harp on this, but can we have a more explicit example until the Flamingo transform is available? Even if it's just a mock tokenizer and image transform with clearly-defined API contracts in a code snippet, I think that will help a lot

torchtune/data/_messages.py
```diff
@@ -463,3 +463,45 @@ def __call__(self, sample: Mapping[str, Any]) -> Mapping[str, Any]:
             updated_messages.append(Message.from_dict(message))
 
         return {"messages": updated_messages}
+
+
+def validate_messages(
```
ebsmothers, line 468: If we're moving this here, should also delete from data/_utils.py?

ebsmothers: Also for multimodal, do we need to do any extra validation? E.g. number of images == number of messages with type == 'image'? (Not saying we should do it here btw)

ebsmothers (approved):
A few more comments but no major concerns from me. Stamping to unblock

tests/torchtune/datasets/multimodal/test_llava_instruct_dataset.py
```diff
@@ -0,0 +1,136 @@
...
+            5: 8,
+            7: 8,
+            6: 5,
+            1: 5,
+            9: 2,
+            0: 1,
+            -2: 1,
+            12: 1,
+            10: 1,
+            -1: 1,
```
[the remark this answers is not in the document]
RdoubleA: image token, eos id

torchtune/datasets/multimodal/_llava_instruct.py
```diff
@@ -0,0 +1,191 @@
...
+    new_system_prompt: Optional[str] = None,
+    train_on_input: bool = True,
+    packed: bool = False,
+    split: str = "train",
+    data_files: str = "llava_instruct_150k.json",
+    **load_dataset_kwargs: Dict[str, Any],
+) -> Union[SFTDataset, PackedDataset]:
+    """
+    Support for family of image + text datasets similar to
+    `LLaVA-Instruct-150K <https://huggingface.co/datasets/liuhaotian/LLaVA-Instruct-150K>_`
```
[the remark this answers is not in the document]
RdoubleA: oop

torchtune/data/_messages.py
RdoubleA: good point... I am not sure yet where this should happen. maybe note this as a followup once the whole end-to-end is established

torchtune/datasets/multimodal/_llava_instruct.py
RdoubleA: excellent question, @pbontrager and I were considering enforcing some contract at the model transform level, but maybe for now we just pass that up from the tokenizer

PR #1162, opened 2024-07-11
Add version check for FSDP2 recipe

ebsmothers:
Add a check that users have at least PyTorch 2.4 installed to run FSDP2 recipe


joecummings (commented):
Can you add a comment at the top of the config files too?

ebsmothers:
> Can you add a comment at the top of the config files too?

I think they're already in there, just needed to pull from latest main

PR #1163, opened 2024-07-11
Experimental Flag for tune ls

pbontrager:
#### Context
- [ ] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

Currently we have a number of recipes under dev that we would like to make accessible to users but still clearly signal to users that these are not as well tested and may have special dependencies. The current behavior is to list these all in "tune ls" which isn't ideal as it doesn't disclose the experimental nature of these recipes. One option would be to exclude them entirely from "tune ls" but then it makes it very difficult to use them outside of github.

The proposal here is to add an "experimental" flag to tune ls that displays the extra recipes. All of the recipes are always available in tune run, but if you want to view them in ls you must pass the extra flag. This is accomplished by adding an extra parameter to the recipe dataclass in _recipe_registry that allows some recipes to be marked as experimental. Then get_all_recipes can accept an argument "include_experimental". This is True for "tune run" and False by default for "tune ls".

#### Changelog
- include_experimental added to get_all_recipes
- experimental added to recipe dataclass
- experimental flag added to tune ls
- Updated unit test and cli rst
- Fixed bug where dev recipes weren't included in project package

#### Test plan
Ran tests and test tune ls in terminal for expected outputs

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -66,7 +66,7 @@ where = [""]
 include = ["torchtune*", "recipes*"]
 
 [tool.setuptools.package-data]
-recipes = ["configs/*.yaml", "configs/*/*.yaml"]
+recipes = ["configs/*.yaml", "configs/*/*.yaml", "configs/*/*/*.yaml"]
 
 
 # ---- Tooling specifications ---- #
--- a/docs/source/tune_cli.rst
+++ b/docs/source/tune_cli.rst
@@ -105,7 +105,7 @@ with matching names. By default we ignore safetensor files, but if you want to i
 List built-in recipes and configs
 ---------------------------------
 
-The ``tune ls`` command lists out all the built-in recipes and configs within torchtune.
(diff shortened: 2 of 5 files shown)


pyproject.toml
```diff
@@ -66,7 +66,7 @@ where = [""]
 include = ["torchtune*", "recipes*"]
 
 [tool.setuptools.package-data]
-recipes = ["configs/*.yaml", "configs/*/*.yaml"]
+recipes = ["configs/*.yaml", "configs/*/*.yaml", "configs/*/*/*.yaml"]
```
ebsmothers, line 69: Can we just use something like `configs/**.yaml` here?

RdoubleA: it should be `configs/**/*.yaml` to match any number of subdirectories

ebsmothers: I think we already landed the existing version in another PR. Also after thinking about it more maybe it's better to be explicit here anyways

PR #1167, opened 2024-07-12
update models docs

ebsmothers:
A bunch of miscellaneous documentation changes to torchtune/models. Main changes:

- Provide explicit download instructions for all model variants (at least all the ones that have training configs. We can consider adding the others too, but left it out since there's not an easy one-liner to use them anyways)
- Actually document lora_dropout in builders where it's used (we should add it to the other builders where it's missing, can do that later though)
- Minor clarifications and formatting in clip_vision_encoder and TiledTokenPositionalEmbedding
- Making clip_vit_224_transform private since I don't think it's worth documenting rn (and it's not used anywhere atm)
- Improve rendering of our model-specific SentencePiece tokenizers' `tokenize_messages` docstrings
- Remove `scale_hidden_dim_for_mlp` from public APIs (similarly, I don't think it's worth documenting/exposing rn)
- Fix return type for Mistral 7B classifier

### DIFF
--- a/docs/source/api_ref_models.rst
+++ b/docs/source/api_ref_models.rst
@@ -11,9 +11,18 @@ llama3
 
 All models from the `Llama3 family <https://llama.meta.com/llama3/>`_.
 
+To download the Llama3-8B-Instruct model:
+
 .. code-block:: bash
 
-    tune download meta-llama/Meta-Llama-3-8B-Instruct --hf-token <ACCESS_TOKEN>
+    tune download meta-llama/Meta-Llama-3-8B-Instruct --hf-token <HF_TOKEN>
+
+To download the Llama3-70B-Instruct model:
+
+.. code-block:: bash
+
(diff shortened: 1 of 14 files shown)


pbontrager:
Is `-hf-token <HF_TOKEN>` required for phi3, gemma, and mistral? I thought it was only required for models where you had to get special approval for the weights.

ebsmothers:
> Is `-hf-token <HF_TOKEN>` required for phi3, gemma, and mistral? I thought it was only required for models where you had to get special approval for the weights.

Oh interesting.. I just copied from the example commands in our configs, all of which have it included. Just tested these 3 -- seems like Phi3 does not require it but the other two do. If it's all the same to you, maybe I'll leave it in for now and we can separately update the Phi3 commands across the board later on; what do you think?

joecummings (approved):
We need to do an overhaul of our entire models page to help people find what models we support, but this is a good improvement for our current release.

PR #1174, opened 2024-07-12
Fix llama3 chat tutorial

RdoubleA:
The recent tokenizer refactors have not been reflected in the chat tutorial. This PR updates all the code paths accordingly.

Importantly, this PR also clarifies Llama3 Instruct vs Base.

### DIFF
--- a/docs/source/tutorials/chat.rst
+++ b/docs/source/tutorials/chat.rst
@@ -2,24 +2,25 @@
 Fine-tuning Llama3 with Chat Data
 =================================
 
-Llama3 introduced a new prompt template for fine-tuning with chat data. In this tutorial,
+Llama3 Instruct introduced a new prompt template for fine-tuning with chat data. In this tutorial,
 we'll cover what you need to know to get you quickly started on preparing your own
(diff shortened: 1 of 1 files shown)


docs/source/tutorials/chat.rst
```diff
@@ -132,9 +142,9 @@ as IDs 1 and 2. We can verify that these are our BOS and EOS tokens.
 
 .. code-block:: python
 
-    print(tokenizer.spm_model.piece_to_id("<s>"))
+    print(tokenizer._spm_model.spm_model.piece_to_id("<s>"))
```
ebsmothers, line 145: Should we add a quick comment to explain the nested spm_model attributes? As is it looks a little weird, but also don't wanna get into a whole digression about our tokenizer abstractions either

ebsmothers (approved):

docs/source/tutorials/chat.rst
RdoubleA: not sure how to do this without requiring a long digression. maybe there's a better way to organize this in the tokenizer class

PR #1176, opened 2024-07-13
Fix incorrect date check for ao nightlies

ebsmothers:
Our nightly recipe test CI job is failing because we aren't parsing the dates in ao nightlies correctly. We gate a feature on the 7/3 nightly, which is represented as "2024.7.3+cu121" (e.g.). But "2024.7.3+cu121" >= e.g. "2024.7.13+cu121", so we can't compare the raw strings.

This PR adds a util to convert everything to datetime objects so that we're comparing across a standardized format.

Test plan: added PR trigger to nightly recipe test, confirmed CI was green: 

Screenshot 2024-07-12 at 7 43 00 PM

### DIFF
--- a/.github/workflows/recipe_test_nightly.yaml
+++ b/.github/workflows/recipe_test_nightly.yaml
@@ -4,6 +4,7 @@ on:
   schedule:
     # Runs at midnight every day
     - cron:  '0 0 * * *'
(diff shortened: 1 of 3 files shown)


RdoubleA (approved):
This is indeed one of the PRs

PR #1184, opened 2024-07-16
fix torchtune internal CI after the `quantize_` API refactor

jerryzh168:
Summary: att

Differential Revision: D59822075


RdoubleA (approved):
Thanks for the patch. Looks like just a lint error

Issue #1044, opened 2024-06-03
Refactor InstructTemplate and ChatFormat


RdoubleA:
Prompt templates have been trending towards being handled as special tokens in the tokenizer (llama2 -> llama3, mistral v2 -> mistral v3), making the classes `InstructTemplate` and `ChatFormat` less practical. Instruct data can also be considered a form of single-turn chat data, and more models are leaning into the chat / conversation format especially with adding tool-calling capabilities.

- Investigate and determine if `InstructTemplate` is still useful for quick mapping of data table columns to a prompt and should remain a standalone class
- Consider merging `InstructTemplate` and `ChatFormat` into one `PromptTemplate` class to support legacy templating behavior
- Figure out plan for customizing special tokens in a tokenizer for any arbitrary template a user may want to fine-tune with

See discussion in #1019 for more context. cc @ebsmothers

joecummings:
@RdoubleA  Can this be closed?

Issue #1002, opened 2024-05-18
Chat tutorial doesn't work as is


christobill:
Hello 👋 
I have an error trying to follow instruction on this page: https://pytorch.org/torchtune/main/tutorials/chat.html 

when I run:
`tune run --nproc_per_node 2 lora_finetune_distributed --config custom_config_lora.yaml`

I get: 
```
INFO:torchtune.utils.logging:Dataset and Sampler are initialized.
INFO:torchtune.utils.logging:Learning rate scheduler is initialized.
Traceback (most recent call last):
  File "/opt/conda/lib/python3.10/site-packages/recipes/lora_finetune_distributed.py", line 615, in <module>
    sys.exit(recipe_main())
  File "/opt/conda/lib/python3.10/site-packages/torchtune/config/_parse.py", line 50, in wrapper
    sys.exit(recipe_main(conf))
  File "/opt/conda/lib/python3.10/site-packages/recipes/lora_finetune_distributed.py", line 610, in recipe_main
    recipe.train()
  File "/opt/conda/lib/python3.10/site-packages/recipes/lora_finetune_distributed.py", line 523, in train
    for idx, batch in enumerate(
  File "/opt/conda/lib/python3.10/site-packages/tqdm/std.py", line 1166, in __iter__
    for obj in iterable:
  File "/opt/conda/lib/python3.10/site-packages/torch/utils/data/dataloader.py", line 631, in __next__
    data = self._next_data()
  File "/opt/conda/lib/python3.10/site-packages/torch/utils/data/dataloader.py", line 675, in _next_data
    data = self._dataset_fetcher.fetch(index)  # may raise StopIteration
  File "/opt/conda/lib/python3.10/site-packages/torch/utils/data/_utils/fetch.py", line 51, in fetch
    data = [self.dataset[idx] for idx in possibly_batched_index]
  File "/opt/conda/lib/python3.10/site-packages/torch/utils/data/_utils/fetch.py", line 51, in <listcomp>
    data = [self.dataset[idx] for idx in possibly_batched_index]
  File "/opt/conda/lib/python3.10/site-packages/torchtune/datasets/_chat.py", line 81, in __getitem__
    sample = self._data[index]
  File "/opt/conda/lib/python3.10/site-packages/datasets/dataset_dict.py", line 81, in __getitem__
    raise KeyError(
KeyError: "Invalid key: 0. Please first select a split. For example: `my_dataset_dictionary['train'][0]`. Available splits: ['train']"
```

[...]

PR #1186, opened 2024-07-16
[RFC] Unified dataset with data and model transforms

RdoubleA:
Thanks to @pbontrager for all the discussions to help converge to this design.

**TLDR:** 
- Let’s make a general fine-tuning dataset class that takes in a data transform class and a model transform class and supports **both text-only and multimodal**
- Instruct, chat, multimodal neatly fall into this structure - refactored grammar dataset as an example. Preference and TextCompletion we will not touch.
- All multimodal transforms (e.g. image transforms) can be captured by the model_transform parameter
- Tokenizer can double as a model transform by inheriting the Transform protocol - meaning text datasets can continue to just use tokenizers
- Instruct templates and chat formats can be merged into a singular Template class that can be used when building data transform classes
- Config and recipe changes are very minimal

## Context
The challenge with designing a dataset class for fine-tuning is the sheer breadth of all the different data preprocessing, prompt templates, dataset structure, and custom preprocessing methods that could be supported. For a library solution that users can import and configure for their own data, we need to impose a loose structure to provide some consistency across the codebase. We’ve currently landed halfway to a general dataset class in torchtune by having task-specific dataset classes:
- InstructDataset uses InstructTemplates and converts columns in a dataset directly to user and assistant messages for tokenization
- ChatDataset uses ChatFormats and restructures a column with a “conversation” into the torchtune Message structure, which then gets tokenized

This is confusing and inconsistent because the overall pipeline is similar (some preprocessing, create Messages, tokenize), yet these are arbitrarily separated with different abstractions in both - InstructTemplates and ChatFormats - and it’s hard for users to reason about when or how to use each when customizing their dataset. And this does not easily extend to multimodal training which involves model specific image/video/audio transforms. We’ve somewhat constricted users to follow a rigid data pipeline structure that is simply not extensible enough.

**Core intuition: the data pipeline can be split into two phases: data transform and model transform.**

Instead of trying to build a class/abstractions that fit every single possible dataset, data preprocessing, and model preprocessing method, we can partition the class into two steps that the users themselves define and provide: a data-specific transform and a model specific transform. All transforms can be categorized into either a transform that is specific to the dataset (getting the right columns, applying prompt templates, restructuring the list of messages) or specific to the model (tokenization, image transforms). Both the data transform and the model transform are relegated to classes that user’s can construct and then pass into the dataset class, constructing the pipeline themselves. This provides the loose structure with full flexibility for any customization/edge cases. The other added benefit is that the model transform can be easily ported into a config for inference or eval on a different dataset, since it’s agnostic to the input data.

Screenshot 2024-07-16 at 2 17 19 PM

The anchor point between data and model transform (and the reason why it’s not just one amorphous transform class) is the Message class. This provides a unified contract for users to design their data transforms with:
- Data transforms should take in a sample dict with any columns and do whatever processing needed to output a list of Messages which all the model tokenizers understand. 
- Model transforms should take in the Messages (and images/video/audio) and apply the correct tokenizer or image transform based on the model and output the tokens/labels used by the model for training. 

We can keep the dictionary as a unified representation of a single sample that all parts of the pipeline operate on.

This structure lets each model and dataset have fully custom processing instead of creating a middle abstraction to categorize a subset of model or datasets (Instruct and Chat, for example), covering the entire breadth of possible data transforms.

## Proposal
With this structure we can define a general FinetuneDataset for all of fine-tuning. The dataset class itself simply calls the components in the right order; the magic happens in the custom data and model transforms. Now this class can focus on general dataset-specific tasks, such as iterable, distributed, integration with HF datasets, etc, and let the transform classes handle the diversity of options. See FinetuneDataset in code changes for details.

### Instruct and chat datasets

The original instruct and chat dataset distinction is no longer necessary since all data will be transformed into Messages in order to be tokenized. But conceptually, here’s how they would be reworked in the new design:
```python
# Old designs
class InstructDataset(Dataset):
    def __init__(
        self,
        tokenizer: ModelTokenizer,
        source: str,
        template: InstructTemplate,
        transform: Optional[Callable] = None,
        train_on_input: bool = False,
    ) -> None:
        ...

    def _prepare_sample(self, sample: Mapping[str, Any]) -> Dict[str, List[int]]:
        # This falls under message transform
        transformed_sample = self._transform(sample) if self._transform else sample

        prompt = self.template.format(transformed_sample, self._column_map)
        key_output = (
            self._column_map["output"]
            if self._column_map and "output" in self._column_map
            else "output"
        )
        messages = [
            Message(role="user", content=prompt, masked=(not self.train_on_input)),
            Message(role="assistant", content=transformed_sample[key_output]),
        ]

        ...
        # This falls under model transform
        tokens, mask = self._tokenizer.tokenize_messages(
            messages, max_seq_len=self.max_seq_len
        )

        ...

class ChatDataset(Dataset):
    def __init__(
        self,
        *,
        tokenizer: ModelTokenizer,
        source: str,
        convert_to_messages: Callable[[Mapping[str, Any]], List[Message]],
        chat_format: Optional[ChatFormat] = None,
        max_seq_len: int,
        train_on_input: bool = False,
    ) -> None:
        ...

    def _prepare_sample(self, sample: Mapping[str, Any]) -> Dict[str, List[int]]:
        # This falls under data transform
        messages = self._convert_to_messages(sample, self.train_on_input)
        if self.chat_format is not None:
            messages = self.chat_format.format(messages)
        ...
        # This falls under model transform
        tokens, mask = self._tokenizer.tokenize_messages(
            messages, max_seq_len=self.max_seq_len
        )
        ...


# New designs
instruct_dataset = FinetuneDataset(
    source=source,
    # Getting right columns, applying prompt template, creating messages falls here
    message_transform=AlpacaMessages(),  # or something similar
    model_transform=llama3_tokenizer(),
)

chat_dataset = FinetuneDataset(
    source=source,
    # Converting to the correct message structure and applying chat templates
    message_transform=SlimOrcaMessages(),  # or something similar
    model_transform=llama3_tokenizer(),
)
```
### What happened to the tokenizer argument?
Tokenization can be generalized as a model transform. In text-only datasets, the model transform is only tokenization. In multimodal datasets, the model transform includes tokenization for text data + image/video/audio transforms for other modalities. To support the original behavior of passing the tokenizer into a text-only dataset builder, we can update all tokenizer classes to additionally inherit from the Transform protocol (multiple inheritance?? how spicy). The tokenizer can now be passed into the model_transform argument. See Llama3Tokenizer in files changed for details.

**Alternative:** Have a TokenizerTransform class that takes in a tokenizer and applies its tokenize_messages. This would reduce redundancy (instead of updating all model tokenizers with a call method) but may make defining dataset.model_transform in the config a bit more complicated by requiring nested components

**Alternative:** Keep tokenizer as a standalone required argument and make model transform optional. Tokenization will ALWAYS be a required step in the model transform, so it may make sense to make that explicit. However, this is less flexible for multimodal models that require post-tokenization transforms (ex: generating the cross-attention mask requires the text tokens and images post-transform). Although, we could sequence tokenization to always come first and then apply the model transform.

### What happened to instruct templates and chat formats?
These abstractions were originally intended to support both quickly mapping columns to a prompt string or structuring system/user/assistant messages with specific tags. But if both instruct and chat data can be generalized into creating messages from data columns, there’s no need to keep these separate with different signatures. It can cause complications for inference especially (see https://github.com/pytorch/torchtune/issues/1044) and just adds to user burden of things they need to learn. Instead, let’s merge them and show examples of how they can be used as plug-in components in the data transforms. See _templates.py in files changed.

All prompt templates can be generalized as prepending and appending flavor text to each of system, user, and assistant messages depending on the role. This is the purpose of the QuickTemplate class. For more complex behavior (like llama2 and mistral with weird behavior with system prompts), a template class can inherit from the protocol directly.

### Preference and text completion datasets
Both of these do not neatly fall into the proposed design. Ideally this can be the go to dataset solution for all types of datasets, but an argument can be made that different training paradigms (SFT [instruct/chat] vs RLHF [preference] vs CPT [text completion]) should have separately evolving dataset abstractions
- Preference has separate chosen user+assistant messages and rejected user+assistant messages, although this could be handled in a specific message transform class for preference datasets
- Preference calls tokenize messages on chosen and rejected conversations separately and returns them in different fields. This is different from the default behavior of when model tokenizers are called as a transform which returns just tokens, mask and have no notion of chosen/rejected. This would require separate model transforms for preference (Llama3PreferenceTransform, etc).
- Text completion does not use tokenize messages and encodes the raw text with eos or bos. We currently do not support any templating or data transforms in text completion, although it wouldn’t be difficult to add this. Extra would would be required to make a tokenizer model transform call encode instead of tokenize messages

The current proposal will only refactor instruct/chat because of the additional complexities for Preference, TextCompletion, but this might be worth revisiting in the future.

### Example dataset
We've revisited the Grammar dataset (liweili/c4_200m) with the new design, please see the code changes.

We’ve essentially upgraded the original GrammarErrorCorrectionTemplate which simply takes the input and output columns and constructs a prompt with returning the full list of Messages. Templates can be plugged into the data transform classes and used as needed. And each dataset will have its own specific transform class that structures the Messages from the right columns with the right template, instead of baking that into the FinetuneDataset and having it generalize to all cases.

### Where do max_seq_len, train_on_input go?
These have always been floating arguments without a clear home. In fact, train_on_input has been a source of confusion since all the message converters require it as an argument (https://github.com/pytorch/torchtune/issues/1002) but it is defined in ChatDataset. With the two transforms approach, max_seq_len (which is generally tied to the model’s context window) stays within the tokenizer / model transform and train_on_input (somewhat dataset specific, e.g., alpaca requires this to be true) stays with the data transform and is exposed in the builder.

### Config updates
Simply replace tokenizer with model_transform.
```yaml
model_transform:
  max_seq_len: 8192

dataset:
  _component_: torchtune.datasets.grammar_dataset
  train_on_input: True
```
### Recipe updates
One line change.
```yaml
# old
ds = config.instantiate(cfg_dataset, tokenizer=self._tokenizer)

# new
ds = config.instantiate(cfg_dataset, model_transform=self._tokenizer)
```

## Caveats
The design relies on the assumption of a few keys that are used across the data transform, model transform, dataset, and recipe. Data transform is expected to return the same sample dict with the updated field “messages” that contains the list of messages representing that row. Model transform is expecting this “messages” field to be present, and outputs fields “tokens” and “mask” which the dataset needs to generate “labels”. For image transforms, we assume “images” is the field in the sample dict to operate on.

There are a few ways we can address this:
- Leave it as is and just be consistent throughout the codebase to make it a convention
- Codify these expected keys into global constants, similar to utils.MODEL_KEY, utils.OPTIM_KEY for checkpointing
- Use these as defaults but let them be configurable in each of the transform classes: __init__(self, …, key_messages: str = “messages”)
- Do the kwargs approach instead (see discussion [here](https://github.com/pytorch/torchtune/pull/1141#discussion_r1669709525))


felipemello1 (commented):
First impressions: I like the concept of DataTransforms and ModelTransforms! We probably need to stress-test it a bit. I dont think that, by just looking at these names, it is clear to me what should belong in DataTransforms and what should belong in ModelTransforms, and that tokenizer is a type of ModelTransform. Maybe thats ok, but if we can lower the cognitive barrier, that would be nice.

1) I felt that it was a bit hard to understand all the pieces i have to define/use to create a new dataset. Eg:

```
from torch.utils.data import Dataset
from torchtune.data import Message
from torchtune.data._templates import GrammarErrorCorrectionTemplate
from torchtune.datasets._finetune import FinetuneDataset
from torchtune.modules.transforms import Transform
from torchtune.datasets._grammar import GrammarMessages, grammar_dataset
```

I wonder if we have a chance to aggregate some of those, for example, a Transform and a Template, if they will always and only be used together.

Maybe having a DAG with all the ingredients necessary would make it easy to visualize.

2) We probably need explicit naming conventions. e.g. GrammarMessages is a transform, but its hard to know that by the naming alone.

3) I appreciate the detailed discussion. I feel that there are important questions there, and it would be nice to have a TLDR: Q1, Q2, ..., Qn, so we can make sure we are not forgetting any.

torchtune/models/llama3/_tokenizer.py
```diff
@@ -237,14 +239,32 @@ def tokenize_messages(
...
+
+        Args:
+            sample (Mapping[str, Any]): A sample with a "messages" field containing
+                a List[Message] to tokenize
+
+        Returns:
+            Mapping[str, Any]: The sample with added "tokens" and "mask" fields
+                and the "messages" field removed.
+        """
+        messages = sample.pop("messages")
```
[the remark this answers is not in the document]
RdoubleA: we won't need messages once they are tokenized, so I was popping to save some memory

torchtune/datasets/_grammar.py
```diff
@@ -47,12 +84,11 @@ def grammar_dataset(
         >>> Batch size: 8
     """
 
-    return instruct_dataset(
-        tokenizer=tokenizer,
+    message_transform = GrammarMessages(train_on_input=train_on_input)
```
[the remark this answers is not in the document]
RdoubleA: hm, I see what you mean but if this transform is changed then there's no point of using this builder function is there?

torchtune/datasets/_grammar.py
```diff
@@ -4,17 +4,54 @@
...
+from torchtune.modules.transforms import Transform
+
+
+class GrammarMessages(Transform):
+    def __init__(
+        self, train_on_input: bool = False, column_map: Optional[Dict[str, str]] = None
+    ):
+        self.train_on_input = train_on_input
+        self.column_map = column_map
+        self.template = GrammarErrorCorrectionTemplate()
```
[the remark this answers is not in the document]
RdoubleA: I was initially leaning towards that, but in some cases the prompt templates may be model specific (e.g. MistralChatFormat, Llama2ChatFormat, ChatMLFormat) so I decided to make them standalone components. Also, prompt templates are more task specific rather than dataset specific. Actually, I think making this the default template but also changeable is preferrable.

RdoubleA:
@felipemello1 Thanks for the detailed review, I agree with all your points:

>  I dont think that, by just looking at these names, it is clear to me what should belong in DataTransforms and what should belong in ModelTransforms, and that tokenizer is a type of ModelTransform.

Do you think this can be addressed by thorough documentation? Or would having explicit abstractions for DataTransforms and ModelTransforms be more clear?

> We probably need explicit naming conventions. e.g. GrammarMessages is a transform, but its hard to know that by the naming alone

Agreed that this will probably help distinguish them a bit more, although I did want to be clear that we are returning a list of messages from the data transform.

> Maybe having a DAG with all the ingredients necessary would make it easy to visualize.

Yep I can add a quick version of this to the summary, and a nicer version of this to the eventual doc updates

torchtune/data/_templates.py
```diff
@@ -0,0 +1,80 @@
...
+            messages (List[Message]): a single conversation, structured as a list
+                of :class:`~torchtune.data.Message` objects
+
+        Returns:
+            The formatted list of messages
+        """
+        pass
+
+
+class QuickTemplate(Template):
```
joecummings, line 40: What is `QuickTemplate`? I don't understand

felipemello1:
> Do you think this can be addressed by thorough documentation? Or would having explicit abstractions for DataTransforms and ModelTransforms be more clear?

I am not sure. For example, when you use instagram, you dont have to read documentation to understand the main functionality. Code is different, but if we can use conventions that the users are already familiar with, then this cognitive barrier will be lower. 

Can you remind me if ModelTransforms will always be a type of tokenization? If that is true, maybe we can just name it SomethingTokenizerStrategy?

kartikayk (commented):
First off, I love this RFC. As these go, it's definitely up there in terms of quality and readability - thank you!

A few thoughts overall:

> the data pipeline can be split into two phases: data transform and model transform

I think this makes sense. But we'll need to be very careful with naming and with how we compose different transforms together. I think some of the naming as I skimmed through the code below was not very intuitive (I'd stress check this with other folks though).

One noob question. I'm trying to understand the ```message_transform```. IIUC, this does the following:
- Convert the raw data into ```List[Message]```
- For this list, apply the appropriate template which maybe model specific and/or task specific

Is this correct? If so, why is the second step i.e. applying the template combined with this transform? At first, I assumed this is dataset specific which is counter-intuitive for the model specific template?

> What happened to the tokenizer argument?

TBH, I think making the tokenizer as a required argument makes sense? Do we expect there to be a use case where we don't have textual inputs? Or let me know if I misunderstand?

> Preference and text completion datasets

Keeping these separate makes a ton of sense

felipemello1:
> @felipemello1 It will always include tokenization, but for multimodal it will additionally include the image transform, cross attn mask transform, etc. So calling it `TokenizerStrategy` wouldn't make sense

I was thinking that ImageTransform was part of data transform, since we are literally transforming the data. Had to read your definition again.

Regarding "cross attn mask transform", i was considering any sort of masking as part of the tokenization process, since we always mask tokens. Is that a fair assumption?

RdoubleA:
> I was thinking that ImageTransform was part of data transform, since we are literally transforming the data. Had to read your definition again.

Sorry I guess I jumped a logical step there, I was thinking image transforms would always be model specific, like the CLIPImageTransform, Flamingo, etc, every model will do it slightly differently. But let me know if this is not the case.

>  i was considering any sort of masking as part of the tokenization process

Not necessarily, we are adding a mask field to the sample dict that is computed based on the tokens and transformed images, and this is used in SDPA. I considered that a "transform", but I guess that's just semantics. My main point was that folding image transform + tokenizer + mask all into tokenizer strategy feels too narrow.

torchtune/datasets/_finetune.py
```diff
@@ -0,0 +1,51 @@
...
+
+import numpy as np
+
+from datasets import load_dataset
+from torch.utils.data import Dataset
+from torchtune.data import CROSS_ENTROPY_IGNORE_IDX
+from torchtune.modules.transforms import Transform
+
+
+class FinetuneDataset(Dataset):
```
joecummings, line 17: I don't like the idea of differentiating between a "finetune" dataset and any other kind of dataset - doesn't feel like the right name. 

Still working out what would be the best convention here.

torchtune/datasets/_finetune.py
```diff
@@ -0,0 +1,51 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
```
joecummings, line 1: Also since this is BC breaking, what is our plan to deprecate the old APIs? 2 release cycles like Core? Can this RFC include a warning in all the old datasets?

torchtune/datasets/_finetune.py
RdoubleA: Yeah I couldn't come to a good name. I did "finetune" because preference = RLHF, text completion = CPT, and everything else falls under finetune. Maybe could keep it as ChatDataset? MessageDataset?

torchtune/datasets/_grammar.py
```diff
@@ -4,17 +4,54 @@
 # This source code is licensed under the BSD-style license found in the
```
joecummings, line 4: Would this kind of file belong in datasets anymore??

torchtune/datasets/_finetune.py
RdoubleA: Need to think about this more... do you have any suggestions for the rollout? The dataset builders will remain the same so users shouldn't need to change their configs. Especially if we keep tokenizer as a required argument and make model transform optional, it's only the underlying abstractions changing.

torchtune/datasets/_finetune.py
joecummings: Related to the above point, I do find that instruct/chat/text-completion dataset notion is easy to understand. Would it make sense to still provide those as higher level APIs that build above this new implementation? That way we kind of meet users wherever they are. 1) with specific implemnentaitno like alpaca. 2) with instruct/chat/text-completion and 3) for advanced users, just the base notion of a TorchTuneDataset

torchtune/datasets/_grammar.py
```diff
@@ -47,12 +84,11 @@ def grammar_dataset(
         >>> Batch size: 8
     """
 
-    return instruct_dataset(
-        tokenizer=tokenizer,
+    message_transform = GrammarMessages(train_on_input=train_on_input)
+    ds = FinetuneDataset(
```
SalmanMohammadi, line 88: If a user would like access to the underlying dataset prior to any transformers, e.g. to filter the dataset before sampling, they would have to just define their own dataset? 
This is probably not super relevant here, but just thinking ahead if we want to eventually provide similar kind of abstractions for other kinds of datasets/use-cases, that don't fall neatly into per-sample custom processing.

RdoubleA: The raw data should still be accessible in the class attribute self.data. Filtering can also be folded into the data transform. Although I may need to add a bit of logic to support ignoring certain samples.

RdoubleA:
Getting merged in starting with #1234. Discussion is resolved.

PR #1193, opened 2024-07-17
Integrate flex attention

RdoubleA:
**TLDR: We can nearly 7x (+560%) our throughput by turning compile + sample packing + flex attention on (based on Alpaca). Compared to packing with compiled SDPA, compiled flex attention boosts throughput by over 2x (140%) for Llama3 8B at max sequence length of 8192. This effect is more pronounced as sequence length increases as it scales more efficiently.**

## Context
This is a proposal of what the integration with the prototype API `flex_attention` available in PyTorch Core's nightlies would look like in torchtune. It requires replacing the `nn.scaled_dot_product_attention` call in `CausalSelfAttention` with `flex_attention` and the usage of `BlockMask` instead of a standard tensor mask. The primary challenges are handling versioning since this is not available in the latest stable release of torch. I've tried to minimize changes to the attention/transformer modules and the recipes as much as possible besides version gating logic. Most of the core updates occur in `PackedDataset` and a new collate function for sample packing to construct the `BlockMask`.

## What is FlexAttention?
FlexAttention is an alternative to SDPA that can enable better performance for users that are utilizing different attention variants and are comfortable with using torch.compile. It allows users to specify custom modifications to attention scores within the Fused Scaled Dot Product Attention Kernel. This enables various attention patterns and biases to be implemented efficiently, with potential runtime and memory savings. This includes support for arbitrary masks with flash attention, which will enable performant attention kernels for sample packing (which requires a block causal mask) and vision cross attention masks.
```
torch.nn.attention.flex_attention.flex_attention(
    q,
    k,
    v,
    block_mask=mask,
)
```
The signature is largely similar to SDPA, except the expected mask is now a `BlockMask` (see [here](https://github.com/pytorch/pytorch/blob/main/torch/nn/attention/flex_attention.py#L168)). This takes in a `mask_mod` function that will materialize a kernel-level mask ad-hoc, preventing us from holding large 2D attention masks in memory that scale quadratically with sequence length.

Thus, we expect FlexAttention with sample packing to be 1) more memory-efficient and 2) increase throughput due to faster mask construction and ability to use flash attention.

## How does sample packing with SDPA compare to Flex?
This gist covers the details of this experiment: https://gist.github.com/RdoubleA/012409f7919973d6ba7e9ca3efd5c237. Losses were observed to be equivalent before and after this change.

TLDR is sample packing with FlexAttention scales significantly better with sequence length:
- Throughput actually decreases as we increased max sequence length from 2048 -> 8192 for SDPA, likely due to quadratic increase in compute for assembling a [b x 8192 x 8192] block causal mask
- At a sequence length of 2048, we observed +50% increase in WPS when switching to FlexAttention.
- At a sequence length of 8192, WPS dropped by 23% for SDPA compared to 2048
- At a sequence length of 8192, FlexAttention was a +140% increase in WPS compared to SDPA
- Without packing, we achieve a baseline of ~500 WPS per gpu, meaning we can 7x our throughput by turning packing + compile + flex attention on
- We also reduce peak memory reserved by 1 GB by using flex attention

From these observations, we can extrapolate that flex attention will be a dramatic improvement over SDPA with sample packing for context lengths 100k - 1M. For these reasons, it is worthwhile integrating this in our core modules even if it may add some slightly complexity.

image

## Changelog
- Instead of creating a 2D mask on the fly in `PackedDataset`'s `__getitem__`, we create a list of `document_ids` that indicate which tokens belong to which samples in a packed sequence
- We use `document_ids` to create a `BlockMask` in the batch collater. A new collate function for packing is added, and mask utilities were added to `utils/attention_bias.py`
- `CausalSelfAttention` will swap which attention to use depending on if a normla tensor mask is passed or a `BlockMask`
- All flex attention logic is gated by version. I was on `torch==2.5.0.dev20240717+cu124` when testing.

## Test plan
- [x] Ensure loss curves are identical
- [x] Add test for packed collate
- [x] Update attention tests for flex attention
- [x] Update `PackedDataset` tests
- [x] Test with compile
- [x] Test affected recipes

cc @drisspg @Chillee @kartikayk


drisspg:
One nit: Flex is not intended to be a replacement for sdpa rather it is designed to enable better performance for users that are utilizing different attention variants and are comfortable with using torch.compile

torchtune/modules/attention.py
```diff
@@ -10,6 +10,85 @@
...
+from torchtune.utils.attention_bias import _MaskType
+from torchtune.utils.logging import get_logger, log_once
+
+_log: logging.Logger = get_logger()
+
+
+if torch_version_ge("2.5.0"):
+    from torch.nn.attention.flex_attention import BlockMask, flex_attention
+
+    flex_attention_compiled = torch.compile(flex_attention, dynamic=False)
```
felipemello1, line 23: Not a review yet, just a reminder that this should probably be controlled by cfg.compile. If compile=False, and we use flex_attention, maybe it  should raise an error, or default back to efficient attention

felipemello1: @kartikayk mentioned that there are machines that are not compatible with compile. 

option 1: raise and error

options 2: let users control compile_model, compile_loss, compile_attention (i dont like it too much)

option 3:This may be too much, but its probably the optimal way. Today we work with SDPA, and if the mask is not causal, we default back to efficient attention, right?

We could add flexattention to this logic.

if mask == is_causal:
    att = SPDA
elif compile==True:
     att = flexattention
else:
     att = efficient_attention

kartikayk: I don't think any of our components should have a hard dependency on compile since this limits usability to backends which don't support compile. I think any design should account for this and provide escape hatches. I prefer the flash-attention approach of falling back to the less performant option rather than erroring out. In this case the "less performant" way would actually be to do what we do today i.e. no FlexAttention instead of flex attention without compile (unless you have evidence that flex without compile is ok). Is there a way we can support multiple attention APIs - kinda like option 3 from @felipemello1

RdoubleA: I see - we can do a variant of option 3 that also checks for pytorch version. it will be messy, but I don't see any other way

drisspg: I do not reccomend using FlexAttention without compile, the EfficientAttention backend and soon CuDNN backend with full bias support will have the same memory usage and be much more performant as using flex in eager

torchtune/utils/collate.py
```diff
@@ -69,3 +71,73 @@ def padded_collate(
...
+    block_mask = packed_block_causal_mask(
+        seq_lens=seq_lens,
+        device=device,
+    )
+
+    return {
+        "tokens": tokens,
+        "labels": labels,
+        "input_pos": input_pos,
+        "mask": block_mask,
```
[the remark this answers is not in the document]
felipemello1: I see that you added it to _convert_to_tensors, so probably its not necessary to have it here?

torchtune/modules/transformer.py
```diff
@@ -96,7 +99,11 @@ def forward(
         # Input tensor and attention output have the same shape
         # [b, s, d]
         # Norm applied before self-attention
-        attn_out = self.attn(self.sa_norm(x), mask=mask, input_pos=input_pos)
+        attn_out = self.attn(
+            self.sa_norm(x),
```
felipemello1, line 103: This is not using the new attention, right? I guess we would have to merge/solve conflicts

torchtune/utils/attention_bias.py
```diff
@@ -0,0 +1,126 @@
...
+            shape (batch_size, n), where n is the max number of sequences across packs.
+        device (torch.device): Device to create the mask on.
+
+    Returns:
+        _MaskType: BlockMask or Tensor if torch version < 2.5.0.
+    """
+    if torch_version_ge("2.5.0"):
+        document_ids = _get_document_ids_from_seq_lens(seq_lens)
+        batch_size, max_seq_len = document_ids.shape
+        document_ids = document_ids.to(device)
```
felipemello1, line 110: i dont think we need ".to(device)", if seq_lens is already in the correct device. But reiterating my other comment, i dont fully understand why have device here if we do it in the recipe

torchtune/modules/attention.py
```diff
@@ -10,6 +10,85 @@
...
 from torchtune.modules.kv_cache import KVCache
+from torchtune.utils._version import torch_version_ge
+from torchtune.utils.attention_bias import _MaskType
+from torchtune.utils.logging import get_logger, log_once
+
+_log: logging.Logger = get_logger()
+
+
+if torch_version_ge("2.5.0"):
+    from torch.nn.attention.flex_attention import BlockMask, flex_attention
```
felipemello1, line 21: Can you clarify: If the user doesnt have torch>=2.5, they cannot use block causal. Is that right?

If the answer is yes, then i think we need to express it a bit better in this if/else block

tests/torchtune/utils/test_collate.py
```diff
@@ -6,9 +6,13 @@
...
+
+import pytest
 import torch
+from torchtune.utils import torch_version_ge
 
-from torchtune.utils.collate import padded_collate
+from torchtune.utils.collate import padded_collate, padded_collate_packed
 
 
 class TestBatchPadSequence:
```
felipemello1, line 18: given how these tests were modified, i am understanding that we you were planning to support efficient attention anymore. Maybe thats ok, but it implies that user must have:

torch>=2.5
compile will always be on

In principle, that works for me, but maybe its worth some discussion? Sorry if it was already

felipemello1 (commented):
hey, thanks for PR! I am excited to see it land!

The examples in the docstring are very nice. I left some comments, some are nit, but others are more structural, like:
- Where should padding be done
- When to send to device
- Will we ever fallback to efficient attention?

torchtune/utils/attention_bias.py
```diff
@@ -0,0 +1,126 @@
...
+            document_mask = document_ids[b, q_idx] == document_ids[b, kv_idx]
+            return causal_mask & document_mask
+
+        return create_block_mask(
+            mask_mod,
+            batch_size,
+            1,
+            max_seq_len,
+            max_seq_len,
+            device=device,
```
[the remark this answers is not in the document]
drisspg: You can't wrap currently wrap create_block_mask in compile for a few reasons, I would recommend setting that to true by default, it makes a very large difference in performance and is required in order not to OOM on larger sequences

torchtune/modules/attention.py
drisspg: So the scenarios where compilation will fail are pretty much the same for triton - older gpus. Efficient attention has slightly better older gpus suppport. I would say if you are newer than v100 flex should work and v100 or older (in terms of sm version) efficient would work

torchtune/utils/attention_bias.py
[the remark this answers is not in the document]
drisspg: On changing block_size, if you have verrry sparse blocks it may be worth to bump on very large sequences; I think sweeping in real world use cases is the best bet for figuring out if there is more optimal choice

torchtune/modules/attention.py
felipemello1: ok, update from Driss:

> [https://github.com/pytorch/pytorch/blob/758d78790164bfb041555daed380de96e06f78a3/torch/testing/_internal/common_cuda.py#L29-L32](https://l.workplace.com/l.php?u=https%3A%2F%2Fgithub.com%2Fpytorch%2Fpytorch%2Fblob%2F758d78790164bfb041555daed380de96e06f78a3%2Ftorch%2Ftesting%2F_internal%2Fcommon_cuda.py%23L29-L32&h=AT2SiPER_AEu74JpqaJrnqKSKKj3VJRl_FKR0wiKuL_fYyOrMlaDwNocwYxfQ7Rh_4YlepFVd4JKK5WfOVrlUf-kpJv1eSFDrasGq7E-LG5nu7DvsAkLQntV2xFrhsEsKsRyYZqyOddN0pWqirHHEz0uFOtIuh3sIq3mog)
> 
> ignore the lazyval stuff and just do:
> 
> should_use =  torch.cuda.is_available() and torch.cuda.get_device_capability() >= (7, 5)
> + [https://arnon.dk/matching-sm-architectures-arch-and-gencode-for-various-nvidia-cards/](https://l.workplace.com/l.php?u=https%3A%2F%2Farnon.dk%2Fmatching-sm-architectures-arch-and-gencode-for-various-nvidia-cards%2F&h=AT2SiPER_AEu74JpqaJrnqKSKKj3VJRl_FKR0wiKuL_fYyOrMlaDwNocwYxfQ7Rh_4YlepFVd4JKK5WfOVrlUf-kpJv1eSFDrasGq7E-LG5nu7DvsAkLQntV2xFrhsEsKsRyYZqyOddN0pWqirHHEz0uFOtIuh3sIq3mog)
> 
> Honestly might be a question for: https://fb.workplace.com/groups/385893200869952
> 
> As to what they recommend the min version that should be used

RdoubleA: ok, to summarize the discussion here, only run flex attention if the following are true:
- We are sample packing
- User has torch >= 2.5
- torch.cuda.get_device_capability() >= (7, 5)

and since it doesn't make sense to run flex without compile, we will just compile by default since we are already checking for a certain GPU capability

torchtune/utils/collate.py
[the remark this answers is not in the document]
RdoubleA: probably best to be consistent and do it in the collater

torchtune/utils/attention_bias.py
```diff
@@ -0,0 +1,126 @@
...
+        )
+    else:
+        return create_block_causal_mask(seq_lens=seq_lens).to(device)
```
ebsmothers, line 126: A small thing, but can we make the naming clearer between these two cases (maybe alias the flex_attention API if we need to)? Otherwise it's not super clear that the diff between `create_block_mask` and `create_block_causal_mask` is around flex attention and not causal vs non-causal

torchtune/modules/transformer.py
```diff
@@ -65,20 +67,22 @@ def forward(
...
                 [batch_size x seq_length x embed_dim]
-            mask (Optional[torch.Tensor]): Optional boolean tensor which contains the attention mask
+            mask (Optional[_MaskType]): Optional boolean tensor which contains the attention mask
                 with shape [batch_size x seq_length x seq_length]. This is applied after
                 the query-key multiplication and before the softmax. A value of True in row i
                 and column j means token i attends to token j. A value of False means token i
                 does not attend to token j. If no mask is specified, a causal mask
-                is used by default. Default is None.
+                is used by default. If a BlockMask is passed for document masking in a packed
+                sequence, we use :func:`~torch.nn.attention.flex_attention.flex_attention` when
```
ebsmothers, line 84: The link to flex_attention doesn't seem to render in the live docs, fyi

torchtune/modules/transformer.py
```diff
@@ -373,19 +380,22 @@ def forward(
...
-                with shape [b x s x s]. This is applied after the query-key multiplication and
-                before the softmax. A value of True in row i and column j means token i attends
-                to token j. A value of False means token i does not attend to token j. If no
-                mask is specified, a causal mask is used by default. Default is None.
+            mask (Optional[_MaskType]): Optional boolean tensor which contains the attention mask
+                with shape [b x s x s]. This is applied after
+                the query-key multiplication and before the softmax. A value of True in row i
+                and column j means token i attends to token j. A value of False means token i
+                does not attend to token j. If no mask is specified, a causal mask
+                is used by default. If a BlockMask is passed for document masking in a packed
```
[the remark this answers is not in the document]
ebsmothers: Personally I would split this up more clearly. Like say up front "either a boolean tensor or a `BlockMask`, which means flex attention is being used". Then describe each case in more detail

torchtune/utils/attention_bias.py
```diff
@@ -0,0 +1,126 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
```
ebsmothers, line 1: Couple comments on locations of things: (1) why is this stuff in `attention_bias.py`? This is just mask construction, right? We are not actually adding biases to the scores? And while I'm at it, (2) I wonder about the definition of `_attention_call` inside modules/attention.py. Call me crazy, but why not put all this in e.g. `torchtune/modules/attention_utils.py` or something

torchtune/utils/attention_bias.py
```diff
@@ -0,0 +1,126 @@
...
+
+        def mask_mod(b, h, q_idx, kv_idx):
+            causal_mask = q_idx >= kv_idx
+            document_mask = document_ids[b, q_idx] == document_ids[b, kv_idx]
+            return causal_mask & document_mask
+
+        return create_block_mask(
+            mask_mod,
+            batch_size,
+            1,
```
ebsmothers, line 120: Noob question: what's the difference between setting one of the dimensions to 1 and setting it to None? Are they both just broadcasting?

tests/torchtune/utils/test_attention_bias.py
```diff
@@ -0,0 +1,81 @@
...
+                    [0, 0, 0, 0, 1, 0],
+                    [0, 0, 0, 0, 1, 1],
+                ],
+            ],
+            dtype=torch.bool,
+        )
+        torch.testing.assert_close(actual, expected)
+
+    @mock.patch("torchtune.utils.attention_bias.torch_version_ge")
+    def test_packed_block_causal_mask_sdpa(self, mock_version, seq_lens):
```
ebsmothers, line 57: Do we want to do a comparison against `create_block_mask` anywhere in here?

torchtune/modules/attention.py
ebsmothers: Right and if we are sample packing and either of (2) or (3) don't hold then we will just fall back to SDPA with memory-efficient attention

torchtune/utils/collate.py
```diff
@@ -69,3 +71,73 @@ def padded_collate(
...
     return {"tokens": input_ids.long(), "labels": labels.long()}
+
+
+def padded_collate_packed(
+    batch: List[PACK_TYPE],
+    device: torch.device = torch.device("cuda"),
+) -> Dict[str, torch.Tensor]:
+    """Collate packed sequences into a batch. Only convert the seq lens into
+    a block mask for use with attention. Tokens, labels, and input_pos are
+    already padded to the same length within :class:`~torchtune.datasets.PackedDataset`.
```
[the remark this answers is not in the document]
ebsmothers: Yeah I am also confused by the padding logic, but have a slightly different question. Iiuc:

(1) we pad each pack to `max_seq_len` in the packed dataset. We also append an element to the pack's seq_len field telling us how many padding tokens there are
(2) in the collator we then pad the seq_len field for each pack in the batch to be the same length
(3) we construct the mask using these padded seq_len fieldsing 

Is there something in (3) that necessitates the padding done in (2)? Maybe I'm being dense but I don't see it.. we either use flex attention, in which case we call `_get_document_ids_from_seq_lens`, or we don't, in which case we call `create_block_causal_mask`. In both cases we are just iterating over `seq_lens[batch_idx]`, right? So if I've already padded everything to a fixed length inside of the packed dataset, what is the value of also padding the seq lengths? 

Concretely, if I have pack 0 with (unpadded) sequence lengths [1, 3, 5] and pack 1 with (unpadded) sequence lengths [1, 2, 3, 4, 5] and my max_seq_len is 20, then in the packed dataset I will pad each pack and have new seq lengths [1, 3, 5, 11] and [1, 2, 3, 4, 5, 5], respectively. But then iiuc the collator will also pad [1, 3, 5, 11] -> [1, 3, 5, 11, 0, 0]. Why?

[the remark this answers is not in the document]
ebsmothers: Also @felipemello1's comment on moving all the padding into collate is interesting. I'm kinda inclined to agree; the one drawback could be that we lose that nice fixed-sequence-length property of our packed datasets that makes compile's life easier (though it's done fine with our unpacked variable-length sequences so maybe nbd). I do wonder if we should decouple those changes from this PR since it involves more nontrivial changes to the packed dataset (unless ofc it winds up being straightforward to do on top of the other stuff here)

torchtune/utils/attention_bias.py
drisspg: Early on we only supported the broadcasting case and not None, I think that its better to actually use None now, this has been updated since this PR was opened

torchtune/utils/collate.py
```diff
@@ -69,3 +71,73 @@ def padded_collate(
...
+
+    Args:
+        batch (List[PACK_TYPE]): A list of pack dictionaries containing the following keys:
+            - tokens: input token ids
+            - labels: label token ids
+            - input_pos: relative position ids for each sequence in pack
+            - seq_lens: lengths of each sample within the pack
+        device (torch.device): device to create block mask on. Default is "cuda".
```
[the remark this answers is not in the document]
RdoubleA: this was done before `.to(device)` support was added for block mask. so it should be doable to just keep it in the recipe only.

torchtune/utils/collate.py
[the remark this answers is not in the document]
felipemello1: If we want to keep this property, we can still have it in the collator

torchtune/modules/attention.py
```diff
@@ -10,6 +10,85 @@
...
 from torch import nn
 from torchtune.modules.kv_cache import KVCache
+from torchtune.utils._version import torch_version_ge
+from torchtune.utils.attention_bias import _MaskType
+from torchtune.utils.logging import get_logger, log_once
+
+_log: logging.Logger = get_logger()
+
+
+if torch_version_ge("2.5.0"):
```
SalmanMohammadi, line 20: can we remove the next 70 or so lines after 2.5 is out?

torchtune/utils/attention_bias.py
RdoubleA: document_ids needs to be on the same device as `create_block_mask`. by default it creates on cuda. we could just hardcode this and move document_ids to cuda, or we can pass down the device all the way from the recipe down here. I guess we're not planning to support flex on non-cuda devices, so it should be ok to just assume cuda here.

torchtune/modules/transformer.py
```diff
@@ -568,19 +578,22 @@ def forward(
...
```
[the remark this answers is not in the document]
RdoubleA: this is already stated in the beginning of the sentence?

torchtune/modules/attention.py
RdoubleA: once stable version is 2.5, yes. until then we need to keep this check unfortunately. will move it to a separate util

torchtune/utils/collate.py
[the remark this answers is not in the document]
RdoubleA: cc @gau-nernst mentioned that keep seq len fixed in PackedDataset might still be more efficient for compile even if it does compile successfully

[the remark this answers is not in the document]
RdoubleA: > Concretely, if I have pack 0 with (unpadded) sequence lengths [1, 3, 5] and pack 1 with (unpadded) sequence lengths [1, 2, 3, 4, 5] and my max_seq_len is 20, then in the packed dataset I will pad each pack and have new seq lengths [1, 3, 5, 11] and [1, 2, 3, 4, 5, 5], respectively. But then iiuc the collator will also pad [1, 3, 5, 11] -> [1, 3, 5, 11, 0, 0]. Why?

@ebsmothers I think your assessment is correct. We shouldn't need to pad seq lens as long as they each sum to max seq len

tests/torchtune/utils/test_attention_bias.py
RdoubleA: yep, adding that test

recipes/full_finetune_distributed.py
```diff
@@ -500,14 +500,15 @@ def _setup_data(
...
-                else None
+            collate_fn=partial(
+                padded_collate_sft,
+                padding_idx=self._tokenizer.pad_id,
+                ignore_idx=self._loss_fn.ignore_index,
+            )
+            if not packed
+            else partial(
+                padded_collate_packed,
+                device=self._device,
```
ebsmothers, line 511: no longer need device here?

tests/torchtune/datasets/test_packed_dataset.py
```diff
@@ -145,10 +137,15 @@ def test_packed_dataset(
...
-        expected_mask, expected_input_pos = self._get_expected_mask_and_input_pos(
+        (
+            expected_seq_lens,
+            expected_input_pos,
+        ) = self._get_expected_seq_lens_and_input_pos(
             max_seq_len, sample_size, split_across_pack
         )
-        torch.testing.assert_close(packed[0]["mask"], expected_mask)
+        print(max_seq_len, sample_size, max_packs, split_across_pack)
+        print(packed[0]["seq_lens"])
```
ebsmothers, line 147: remove

torchtune/modules/transformer.py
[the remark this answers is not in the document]
ebsmothers: Bumping this comment, I still think this explanation should be clearer. Something like

```
mask (Optional[_MaskType]): Used to mask the scores in QK^T. Either a boolean tensor or a `BlockMask`.
    If a boolean tensor: (basically just use what you already wrote)

    If a block mask: ... (as is it's not clear to me how I would even construct a block mask, maybe point to the FlexAttention blog for examples here? 
```

My main gripe is that most of what is written here does not actually apply to the case of passing a `BlockMask` iiuc, but that's not clear from the way that it's structured.

tests/torchtune/data/test_collate.py
```diff
@@ -47,6 +52,110 @@ def test_batch_pad_sequence(self):
...
+        not _SUPPORTS_FLEX_ATTENTION,
+        reason="Please install a nightly build of torch to run this test.",
+    )
+    @gpu_test(gpu_count=1)
+    def test_padded_collate_packed_flex(self):
+        # create_block_mask requires that seq_len be divisible by 128, the default block size.
+        # see https://github.com/pytorch/pytorch/blob/main/torch/nn/attention/flex_attention.py#L636
+        batch = [
+            {
+                "tokens": torch.ones(128, dtype=torch.long),
```
ebsmothers, line 121: nit: can we use torch.arange or something slightly more nontrivial?

tests/torchtune/data/test_collate.py
```diff
@@ -47,6 +52,110 @@ def test_batch_pad_sequence(self):
...
+        )
+        torch.testing.assert_close(
+            collated["input_pos"],
+            torch.stack(
+                [torch.zeros(128, dtype=torch.long), torch.zeros(128, dtype=torch.long)]
+            ),
+        )
+        torch.testing.assert_close(
+            collated["mask"].to_dense(),
+            torch.tensor([[[[1]]], [[[1]]]], dtype=torch.int32, device="cuda"),
```
ebsmothers, line 156: Sorry maybe I'm being dense, how do we get from [2] * 64 + [4] * 32 to this?

RdoubleA: I honestly don't know. something to do with how the blocks are represented: https://github.com/pytorch/pytorch/blob/e49c0acc396e89baf8c6450e1fa0571d4ce2d4ed/torch/nn/attention/flex_attention.py#L360

torchtune/datasets/_packed.py
```diff
@@ -209,14 +208,12 @@ def _add_pack(self, pack: PACK_TYPE) -> None:
...
+        """Converts a pack into tensors. Pack comes in as a dict of lists and is converted to tensors."""
         return {
-            "tokens": torch.tensor(pack["tokens"]),
-            "labels": torch.tensor(pack["labels"]),
-            "input_pos": torch.tensor(pack["input_pos"]),
-            "seq_lens": pack["seq_lens"],
+            "tokens": torch.tensor(pack["tokens"], dtype=torch.long),
+            "labels": torch.tensor(pack["labels"], dtype=torch.long),
+            "input_pos": torch.tensor(pack["input_pos"], dtype=torch.long),
+            "seq_lens": torch.tensor(pack["seq_lens"], dtype=torch.long),
```
ebsmothers, line 216: I thought the point of `PACK_TYPE` was to let this be a List. Isn't it just `Dict[str, Tensor]` now?

torchtune/datasets/_packed.py
```diff
@@ -235,6 +232,14 @@ def _pad_pack(self, pack: PACK_TYPE, padding_idx: int) -> PACK_TYPE:
             value=CROSS_ENTROPY_IGNORE_IDX,
         )
 
+        # Add padding tokens as a last seq len to ensure sum is max_seq_len
+        num_padding_tokens = self.max_seq_len - len(pack["tokens"])
```
ebsmothers, line 236: nit: just calculate this once (you also use it in L224)

torchtune/datasets/_packed.py
RdoubleA: the input is still a list. we need to convert to tensors so we can pad the individual packs independent of collater.

torchtune/modules/attention_utils.py
```diff
@@ -0,0 +1,243 @@
...
+        flex_attention,
+    )
+
+    flex_attention_compiled = torch.compile(flex_attention, dynamic=False)
+
+    # We cannot do nested compile, but flex attention only has perf benefits
+    # when compiled. To insulate it from the compiler, we wrap it with
+    # compiler.disable so that it can be used regardless of whether the model
+    # is compiled or not, and flex attention always remains compiled.
+    @torch.compiler.disable(recursive=False)
```
ebsmothers, line 34: So where did we land on this? Recursive is not necessary, right?

torchtune/modules/attention_utils.py
```diff
@@ -0,0 +1,243 @@
...
+import torch
+
+from torch import nn
+from torchtune.utils._version import torch_version_ge
+from torchtune.utils.logging import get_logger, log_once
+
+# We can only use flex attention / BlockMask if torch version >= 2.5.0 and GPU is Turing / SM75 and above
+_SUPPORTS_FLEX_ATTENTION = torch_version_ge(
+    "2.5.0"
+) and torch.cuda.get_device_capability() >= (7, 5)
```
ebsmothers, line 19: This will error if we're running purely on CPU I think

torchtune/modules/attention_utils.py
RdoubleA: it works either way. I think it makes sense to keep recursive just to be overly cautious

tests/torchtune/modules/test_attention_utils.py
```diff
@@ -0,0 +1,150 @@
...
+        torch.testing.assert_close(actual, expected)
+
+    @pytest.mark.skipif(
+        not _SUPPORTS_FLEX_ATTENTION,
+        reason="Please install a nightly build of torch (>=2.5.0) to run this test.",
+    )
+    @gpu_test(gpu_count=1)
+    def test_packed_block_causal_mask_flex(self):
+        # create_block_mask requires that seq_len be divisible by 128, the default block size.
+        # see https://github.com/pytorch/pytorch/blob/main/torch/nn/attention/flex_attention.py#L636
```
ebsmothers, line 92: I don't think this pointer is right. Should probably include the commit hash in the URL

torchtune/modules/attention_utils.py
RdoubleA: can add `torch.cuda.is_available()`

torchtune/modules/attention_utils.py
```diff
@@ -0,0 +1,243 @@
...
+        v: torch.Tensor,
+        block_mask: BlockMask,
+    ) -> torch.Tensor:
+        return flex_attention_compiled(q, k, v, block_mask=block_mask)
+
+    _MaskType = Union[torch.Tensor, BlockMask]
+else:
+    _MaskType = torch.Tensor
+
+_log: logging.Logger = get_logger()
```
ebsmothers, line 47: Any reason not to move this up to the top?

tests/torchtune/modules/test_attention_utils.py
RdoubleA: don't even remember what this is supposed to point to

torchtune/modules/attention_utils.py
```diff
@@ -0,0 +1,243 @@
...
+
+        # Instead of passing a tensor mask, flex attention requires a mask_mod function
+        # that determines which elements of QK^T should be included in the attention
+        # computation prior to the softmax. For sample packing, we need both the
+        # logic for both causal mask and document mask. See PyTorch's official
+        # blog post for more details: https://pytorch.org/blog/flexattention/#mask-mods
+        def mask_mod(b, h, q_idx, kv_idx):
+            causal_mask = q_idx >= kv_idx
+            document_mask = document_ids[b, q_idx] == document_ids[b, kv_idx]
+            return causal_mask & document_mask
```
ebsmothers, line 150: Nit: this part is pretty important, maybe add a short docstring for the function itself explaining the actual mask this is meant to construct

tests/torchtune/data/test_collate.py
ebsmothers: Yeah I saw that, still don't fully understand it. But may just need to look at it a bit more closely

ebsmothers (approved):
You really had to jump through a lot of hoops on this one.. one of these days I swear you will have a PR that is nice and easy where everything goes smoothly. This was not that PR. Honestly great work though, the perf wins on this are gonna be incredible

PR #1196, opened 2024-07-18
[RFC][Docs][WIP] Custom Recipe Tutorial

SalmanMohammadi:
#### Context
- [ ] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

This tutorial will follow along as I (henceforth referred to as "The Author"), The Author, implements a simple recipe (henceforth referred to as "The Tutorial") for the express purpose of educating and enlightening the user (henceforth known as "The Reader"), in the implementation and optimization of language model training code. The Reader, upon accessing and engaging with The Tutorial contained therein, shall be privy to the methodologies, best practices, and advanced optimization techniques as demonstrated by The Author.

Okay, enough nonesense. I briefly mentioned this offline (online?) to @kartikayk and @joecummings. Torchtune's design philosophy makes things hackable, easy-to-understand, and extend. I'd like to surface how straightforward this can make adding custom fine-tuning recipes for our more intermediate users. Readers will get to see the process for how to get started with a new recipe, and generally lots of tips and tricks for writing training code for LLMs. I'd also like to showcase the awesome optimisation features we expose, which I'll demonstrate in the recipe by providing readers some insight into e.g. what `torch.compile` can practically do for them here, without necessarily needing to understand the internals.

Initially I'm thinking of killing two birds with one stone here and documenting my process for adding a (very simple) reward model finetuning recipe which I can later extend and contribute.

As you can see, this is a very rough overview of my vision here, just opening this up early to make sure we're aligned and to field any ideas or input.

EDIT: interest from #1249

#### Test plan
 👀  👀 
 👁️ 👁️

Perhaps someone could guinea pig and try follow along the final tutorial?


pbontrager:
Thank you for going in this direction, this is awesome and something I've been wanting better exposed. Looking at the template you laid out, do you plan to focus on building the recipe from scratch or starting from an existing recipe? From scratch might be cleaner but less realistic in most real scenarios. I like the idea of extending a supervised fine-tuning to reward training because training with multiple models is usually where it's nicest to be able to just script things yourself. Some other potential customizations people would be interested in would be integrating their own logging/custom metrics, adding eval, or any kind of data manipulation during training. Not advocating for making this complicated, just some additional ideas if one direction ends up being too complicated. I look forward to watching the progress.

SalmanMohammadi:
Thanks for your thoughts  @pbontrager : )

> Looking at the template you laid out, do you plan to focus on building the recipe from scratch or starting from an existing recipe? From scratch might be cleaner but less realistic in most real scenarios. 

I doubt I'll need to change too much of the setup code - the recipe setups as they are now are really neat. Adding support for for reward fine-tuning (which is just training a base model with a classifier head on a preference dataset) shouldn't need much too much logic outside of the core train loop. I'm thinking that even if I'm copy-pasting a bunch of code, I'll give the user a rough idea of what it does incase it's relevant for them to modify it (or even point out why you might need to modify it). 

> Some other potential customizations people would be interested in would be integrating their own logging/custom metrics, adding eval, or any kind of data manipulation during training. 

Yeah I dig this, big time. I think the eval pipeline is something I'd like to see fleshed out a bit more in torchtune - we've had some community interest in this. If I'm going down the reward modelling route; there's reward-modelling benchmarks like https://huggingface.co/spaces/allenai/reward-bench. 

Data manipulation is also something I've seen come up quite a bit; even things like specifying configs correctly. I think our docs do explain this well. Imo it'd help to have concrete end-to-end examples with enough context to understand why we're doing things the way we are.

RdoubleA:
@SalmanMohammadi Do you think it's worth separating optimizations / memory saving tricks in a separate tutorial? I feel like that individually is a very useful resource. But overall, I think this outline makes a lot of sense.

SalmanMohammadi:
> @SalmanMohammadi Do you think it's worth separating optimizations / memory saving tricks in a separate tutorial? I feel like that individually is a very useful resource. But overall, I think this outline makes a lot of sense.

I definitely think it's worth going into optimizations / memory saving tricks in a separate tutorial. As I've been writing this, a lot of this tutorial is supported by the awesome tutorials and deep dives which are already in the repo. It'll likely be the same for optimizations / memory stuff, where this is more a high-level look at which components you'd need to grab to enable some of the most generally useful ones, without going into detail about how they work.

PR #1197, opened 2024-07-18
DPO Loss refactor

SalmanMohammadi:
#### Context
- [ ] add a new feature
- [ ] fix a bug
- [x] refactor

I was initially looking at this refactor to add SimPO into the DPO recipe, but got slightly carried away. I think we might need a separate SimPO recipe down the line anyway, since DPO-style losses use a reference model, whereas SimPO doesn't.

This PR refactors the DPO loss module into separate classes for each of the loss types it supported. Each separate loss is now documented with a reference to its corresponding paper, a little intuition about how it works, and comes with a corresponding unit test.

#### Test plan

### DIFF
--- a/docs/source/api_ref_modules.rst
+++ b/docs/source/api_ref_modules.rst
@@ -81,6 +81,8 @@ Loss
    :nosignatures:
 
    loss.DPOLoss
+   loss.RSOLoss
+   loss.IPOLoss
 
 
 Vision Transforms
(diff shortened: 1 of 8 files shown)


RdoubleA (approved):
Excellent refactor. I honestly don't have any concerns here.

As a DPO noob, do you mind adding some details in the docstrings about the training setup for each of these losses? Like what is the policy/reference model, and the core value prop of DPO compared to PPO is that it does not require a reward model. 

Also, are we completely removing the kto pair loss?

SalmanMohammadi:
> Also, are we completely removing the kto pair loss?

Yeah, sorry, I forgot to point this out. The reference implementation in TRL seems to differ enough (it uses a separate trainer with a different dataset format), and I couldn't see tests for the KTO loss in the original DPO PR. I personally would re-consider this when we have tests against a reference implementation.

PR #1204, opened 2024-07-22
Fix check for valid config for 70B training with fused optimizer

joecummings:
#### Context
- [ ] add a new feature
- [x] fix a bug
- [ ] other (please add here)

#### Changelog
* Check for "fused" on the optimizer in the config itself

#### Test plan

**Check that it doesn't hit the check when PyTorch is high enough and options are selected**
1. Install pytorch nightlies
2. `tune run --nproc_per_node 8 full_finetune_distributed --config llama3/70B_full`
3. Confirm it runs ✅

**Check that it doesn't hit the check when options are not selected**
1. Set fused to False
2. `tune run --nproc_per_node 8 full_finetune_distributed --config llama3/70B_full`
3. Confirm it runs  ✅

**Check that it does hit the check when PyTorch is not high enough and options are selected**
1. Install pytorch 2.3 (latest stable)
3. `tune run --nproc_per_node 8 full_finetune_distributed --config llama3/70B_full`
4. Confirm it hits the error ✅

### DIFF
--- a/recipes/full_finetune_distributed.py
+++ b/recipes/full_finetune_distributed.py
@@ -105,8 +105,8 @@ class FullFinetuneRecipeDistributed(FTRecipeInterface):
             )
 
         if (
-            cfg.get("fsdp_cpu_offload", False)
-            and cfg.get("fused", False)
+            cfg.get("fsdp_cpu_offload", True)
+            and cfg.get("optimizer.fused", True)
             and not utils.torch_version_ge("2.4.0")
         ):
             raise RuntimeError(
--- a/recipes/full_finetune_distributed.py
+++ b/recipes/full_finetune_distributed.py
@@ -105,8 +105,8 @@ class FullFinetuneRecipeDistributed(FTRecipeInterface):
             )
 
         if (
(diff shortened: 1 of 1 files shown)


recipes/full_finetune_distributed.py
```diff
@@ -106,7 +106,7 @@ def __init__(self, cfg: DictConfig) -> None:
 
         if (
             cfg.get("fsdp_cpu_offload", False)
-            and cfg.get("fused", False)
+            and cfg.optimizer.get("fused", False)
```
joecummings, line 109: It's fine to directly check the optimizer here, b/c we require it for any configs trying to use this recipe.

RdoubleA (approved):

recipes/full_finetune_distributed.py
pbontrager: Technically shouldn't we check that device='cuda'? Though it's probably unlikely that you'd train with anything else.

pbontrager (approved):
Thanks!

PR #1210, opened 2024-07-23
Checkpointer Save "adapter_only" Option

pbontrager:
#### Context
- [x] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

#### Changelog
This adds a new option to save_checkpoint to only save the adapter if it's available. This adds the option to all three checkpointers. This is needed to better support large checkpoint models like 405B where it doesn't make sense to save the entire checkpoint after each epoch for a lora recipe.

This PR was intended as a minimal change on the checkpoint PR without breaking recipes, hence adding this option. An alternative approach would be to have each checkpointer implement save_checkpoint, save_adapter, and save_recipe. This would simplify conditional logic for saving but I leave this to the checkpoint refactor work.

This implementation raises an error if adapter_only=True and there is no adapter. adapter_only has no affect on intermediate_checkpoint.

#### Test plan

### DIFF
--- a/tests/torchtune/utils/test_checkpointer.py
+++ b/tests/torchtune/utils/test_checkpointer.py
@@ -306,7 +306,7 @@ class TestHFLlama2FullModelCheckpointer:
     ):
         """ """
         state_dict = single_file_checkpointer.load_checkpoint()
-        # state_dict = {MODEL_KEY: orig_state_dict}
+
         with pytest.raises(
             ValueError, match="Adapter checkpoint not found in state_dict"
         ):
@@ -324,6 +324,7 @@ class TestHFLlama2FullModelCheckpointer:
             output_state_dict_1 = safe_torch_load(output_file_1)
 
         output_state_dict_2 = safe_torch_load(output_file_2)
+        # Check that the empty adapter we saved is the one loaded succesfully
         assert len(output_state_dict_2.keys()) == 0
 
     def test_save_checkpoint_in_peft_format(
--- a/tests/torchtune/utils/test_checkpointer.py
+++ b/tests/torchtune/utils/test_checkpointer.py
@@ -301,6 +301,31 @@ class TestHFLlama2FullModelCheckpointer:
         assert len(output_state_dict_1.keys()) + 1 == len(orig_state_dict_1.keys())
         assert len(output_state_dict_2.keys()) + 1 == len(orig_state_dict_2.keys())
 
+    def test_load_save_adapter_only(
+        self, tmp_path, single_file_checkpointer, llama2_hf_checkpoints
+    ):
+        """ """
+        state_dict = single_file_checkpointer.load_checkpoint()
+        # state_dict = {MODEL_KEY: orig_state_dict}
+        with pytest.raises(
+            ValueError, match="Adapter checkpoint not found in state_dict"
+        ):
+            single_file_checkpointer.save_checkpoint(
+                state_dict, epoch=2, adapter_only=True
+            )
+
+        state_dict[ADAPTER_KEY] = {}
+        single_file_checkpointer.save_checkpoint(state_dict, epoch=2, adapter_only=True)
+
+        output_file_1 = Path.joinpath(tmp_path, "hf_model_0001_2.pt")
+        output_file_2 = Path.joinpath(tmp_path, "adapter_2.pt")
+
+        with pytest.raises(ValueError, match="Unable to load checkpoint from"):
+            output_state_dict_1 = safe_torch_load(output_file_1)
+
+        output_state_dict_2 = safe_torch_load(output_file_2)
+        assert len(output_state_dict_2.keys()) == 0
+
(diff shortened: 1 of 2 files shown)


tests/torchtune/utils/test_checkpointer.py
```diff
@@ -301,6 +301,31 @@ def test_save_load_checkpoint_multiple_file(
         assert len(output_state_dict_1.keys()) + 1 == len(orig_state_dict_1.keys())
         assert len(output_state_dict_2.keys()) + 1 == len(orig_state_dict_2.keys())
 
+    def test_load_save_adapter_only(
+        self, tmp_path, single_file_checkpointer, llama2_hf_checkpoints
+    ):
+        """ """
+        state_dict = single_file_checkpointer.load_checkpoint()
+        # state_dict = {MODEL_KEY: orig_state_dict}
```
joecummings, line 309: What's this comment?

tests/torchtune/utils/test_checkpointer.py
```diff
@@ -301,6 +301,31 @@ def test_save_load_checkpoint_multiple_file(
...
+        single_file_checkpointer.save_checkpoint(state_dict, epoch=2, adapter_only=True)
+
+        output_file_1 = Path.joinpath(tmp_path, "hf_model_0001_2.pt")
+        output_file_2 = Path.joinpath(tmp_path, "adapter_2.pt")
+
+        with pytest.raises(ValueError, match="Unable to load checkpoint from"):
+            output_state_dict_1 = safe_torch_load(output_file_1)
+
+        output_state_dict_2 = safe_torch_load(output_file_2)
+        assert len(output_state_dict_2.keys()) == 0
```
joecummings, line 327: What is this part testing? That there's no keys found? 

Leave a comment in the code.

torchtune/utils/_checkpointing/_checkpointer.py
```diff
@@ -264,6 +267,10 @@ def save_checkpoint(
                 f"{os.path.getsize(output_path) / 1000**3:.2f} GB "
                 f"saved to {output_path}"
             )
+        elif adapter_only:
```
joecummings, line 270: This logic doesn't make sense to me. Why wouldn't this go above?

torchtune/utils/_checkpointing/_checkpointer.py
```diff
@@ -463,54 +471,58 @@ def save_checkpoint(
...
-            split_state_dicts[cpt_idx].update({key: weight})
-
-        # write the partitioned state dicts to the right checkpoint file
-        for cpt_idx, model_state_dict in split_state_dicts.items():
-            if not self._safe_serialization:
-                output_path = Path.joinpath(
-                    self._output_dir, f"hf_model_{cpt_idx}_{epoch}"
-                ).with_suffix(".pt")
-                torch.save(model_state_dict, output_path)
+        if not adapter_only:
```
joecummings, line 479: super nit: check for positive affirmation, rather than negative.

joecummings (commented):
Can you also run a llama3.1 recipe and check that it only saves the adapter weights? 10 steps should be fine.

Bonus points for loading up the adapter weights into PEFT, too.

torchtune/utils/_checkpointing/_checkpointer.py
pbontrager: This branch gets called if there wasn't an adapter key in the config (the first if is false) but adapter_only is true

pbontrager:
> Can you also run a llama3.1 recipe and check that it only saves the adapter weights? 10 steps should be fine.

Confirmed by hardcoding adapter_only=True in lora_finetune_singledevice.

joecummings (approved):
lgtm

cc @ebsmothers  tho

ebsmothers (commented):
This looks reasonable to me. Will we expose `adapter_only` in the recipe in a separate PR?

pbontrager:
> This looks reasonable to me. Will we expose `adapter_only` in the recipe in a separate PR?

Yes, I plan to add it to the dev fsdp2 recipe to enable 405B, and that can get rolled into the other LoRA recipes when that recipe is moved out of dev. The alternative is to update all Lora recipes, in this PR, to save with adapter only until the last epoch.

ebsmothers (approved):

brunodoamaral:
I just trained a model with `save_adapter_weights_only: True` and it worked fine.

My question is if there is a way to **generate** from a model that was trained with this option.

I tried to specify the `adapter_checkpint` as shown bellow, but have no success. I got a _missing key exception_ for all lora weights.

```yaml
checkpointer:
  _component_: torchtune.utils.FullModelHFCheckpointer
  checkpoint_dir: ./models/Meta-Llama-3.1-8b
  adapter_checkpoint: adapter_0.pt
```

Is there any example of generation for models trained with  `save_adapter_weights_only: True`?

Thanks for your help!

PR #1211, opened 2024-07-23
Prevent pad ids, special tokens displaying in generate

RdoubleA:
#### Context
- [ ] add a new feature
- [x] fix a bug
- [ ] other (please add here)

Pad ID is implicitly assumed to be 0 in `utils.generate`, and any instances of "0" in the generated tokens is replaced by the tokenizer's pad id. This is very apparent for Llama3 which has a reserved special token for pad id:
```
Tell me a joke, please. I could use a laugh.
Here's one: Why couldn't the bicycle stand up by itself?

Because it was two-tired!

Hope that brought a smile to your face<|finetune_right_pad_id|> Do you have a favorite joke you'd like to share? I'd love to hear it<|finetune_right_pad_id|> 

(And if you're feeling extra playful, I can try to come up with a joke on the spot. Just let me know!)<|eot_id|>
```

Here, I simply remove the logic of replacing 0 with pad id. I additionally filter out any special tokens when decoding with the TikToken base tokenizer, with an option to let them be displayed in the decoded string. With these changes, the same generated output for Llama3 is now cleaned up:
```
Tell me a joke, please. I could use a laugh.
Here's one: Why couldn't the bicycle stand up by itself?

Because it was two-tired!

Hope that brought a smile to your face! Do you have a favorite joke you'd like to share? I'd love to hear it! 

(And if you're feeling extra playful, I can try to come up with a joke on the spot. Just let me know!)
```
Notably, the 0 ID for Llama3 is the token for "!"

#### Test plan
`tune run generate --config generation` for Llama3, which uses tiktoken. Also tried other models that use SentencePiece like Phi3 and confirmed it does not impact any behavior.


torchtune/modules/tokenizers/_tiktoken.py
```diff
@@ -153,5 +156,11 @@ def decode(
                 k = None
             if k:
                 token_ids = token_ids[:k]
-        token_ids = [token_id for token_id in token_ids if token_id != self.bos_id]
+        if not show_special:
```
SalmanMohammadi, line 159: This is slightly far out, but could we move the logic from here to the generation recipe and move the flag for `show_special` there, too? Unless we anticipate there's other places we could use this logic, but if so, could we throw it in a utility function instead?

This way the tokenizers remain agnostic to how they are being used, and all calls to `tokenizer.decode` will behave identically.

RdoubleA: I see what you mean, but I do like keeping this with the tokenizer so I can quickly debug even in the middle of the recipe by calling `tokenizer.decode(sample, show_special=True)`. I find myself doing this a lot when debugging in general, and it's especially useful to see the exact text that's going into the model . If it was a separate utility it wouldn't be straightforward to do this in a pdb debugger

torchtune/models/llama3/_tokenizer.py
```diff
@@ -134,6 +134,7 @@ def decode(
         self,
         token_ids: List[int],
         truncate_at_eos: bool = True,
+        show_special: bool = False,
```
SalmanMohammadi, line 137: nit: `show_special_tokens` pls

SalmanMohammadi: Or `skip_special_tokens` like HF? https://github.com/huggingface/transformers/blob/47c29ccfaf56947d845971a439cbe75a764b63d7/src/transformers/tokenization_utils_base.py#L4007

torchtune/modules/tokenizers/_tiktoken.py
SalmanMohammadi: Okay, yeah, IGYWM.

SalmanMohammadi:
One nit; lgtm

SalmanMohammadi (approved):

joecummings (approved):

PR #1216, opened 2024-07-24
Add test for Llama3.1 ScaledRoPE

joecummings:
#### Context
- [ ] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

#### Changelog
* Rename `Llama3_1RotaryPositionalEmbeddings` --> `Llama3ScaledRoPE`
* Add test for `Llama3ScaledRoPE`

#### Test plan

### DIFF
--- a/tests/torchtune/models/llama3_1/test_position_embeddings.py
+++ b/tests/torchtune/models/llama3_1/test_position_embeddings.py
@@ -99,7 +99,7 @@ class TestLlama3ScaledRoPE:
         # check shapes
         assert_expected(x_out.shape, input.shape)
 
-    def test_forward_with_packed_pos(self, input, rope) -> None:
+    def test_forward_with_2d_pos_ids(self, input, rope) -> None:
         """
         Use input_pos to indicate positions of each token relative to its sequence
         when sample is packed.
@@ -135,3 +135,6 @@ class TestLlama3ScaledRoPE:
         meta_rope._rope_init()
         for p1, p2 in zip(rope_on_device.buffers(), meta_rope.buffers()):
             torch.testing.assert_close(p1, p2)
+
+        # Assert meta_rope cache is no longer on meta device
+        assert meta_rope.cache.device != torch.device("meta")
--- a/tests/torchtune/models/llama3_1/test_position_embeddings.py
+++ b/tests/torchtune/models/llama3_1/test_position_embeddings.py
@@ -4,8 +4,6 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
-from typing import Tuple
-
 import pytest
 import torch
 
@@ -35,7 +33,7 @@ class TestLlama3ScaledRoPE:
 
     EXPECTED_FREQS_CIS_MEAN = tensor(0.1738)
     EXPECTED_FREQS_CIS_SUM = tensor(91141.7656)
-    EXPECTED_FREQS_CIS_MAX = tensor(1.)
+    EXPECTED_FREQS_CIS_MAX = tensor(1.0)
 
     EXPECTED_X_OUT_MEAN = tensor(-2.4781e-06)
     EXPECTED_X_OUT_SUM = tensor(-83.1523)
@@ -57,12 +55,10 @@ class TestLlama3ScaledRoPE:
         return torch.randn(bsz, seq_len, num_heads, head_dim)
 
     @pytest.fixture
-    def rope(
-        self, input_params
-    ) -> Llama3ScaledRoPE:
+    def rope(self, input_params) -> Llama3ScaledRoPE:
         _, _, head_dim, _, max_seq_len = input_params
         return Llama3ScaledRoPE(dim=head_dim, max_seq_len=max_seq_len)
-    
+
     def test_cache_equality(self, input, rope) -> None:
         # Have to explicitly call _rope_init() to initialize theta matrix
         rope._rope_init()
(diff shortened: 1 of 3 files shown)


tests/torchtune/models/llama3_1/test_position_embeddings.py
```diff
@@ -0,0 +1,137 @@
...
+        when sample is packed.
+        """
+        (
+            bsz,
+            seq_len,
+            _,
+            _,
+        ) = input.shape
+        x_out = rope(
+            input, input_pos=torch.arange(seq_len).unsqueeze(0).expand(bsz, seq_len)
```
felipemello1, line 114: n00b question: The input here is different than in the previous test, but we still expect the same sum() self.EXPECTED_X_OUT_SUM? Should they be different?

joecummings: L117 I believe answers your question

felipemello1: I see, you do broadcasting inside of rope implementation

```
# tensor has shape [b, s, 1, h_d // 2, 2] if packed samples,
# otherwise has shape [1, s, 1, h_d // 2, 2]
rope_cache = rope_cache.view(-1, xshaped.size(1), 1, xshaped.size(3), 2)
```

So the sum should be the same, because the input pos is the same for both:
```
>>> seq_len=3
>>> bsz=2
>>> print(torch.arange(seq_len).unsqueeze(0).expand(bsz, seq_len))
tensor([[0, 1, 2],
        [0, 1, 2]])

>>> print(torch.arange(seq_len))       
tensor([0, 1, 2])
```

Maybe I am missing the point, but I am not sure how this is testing packed samples. Should we try to have a more complex input_pos for this case?, e.g.:
tensor([[0, 1, 2, 0, 1, 2, 3, 4],
        [0, 1, 2, 0, 0, 0 ,0]])

Other than that, LGTM!

joecummings: I definitely agree with your second point and would be in favor of filing an issue for a more complex test! For this, I just wanted to copy-paste from the original test_positional_embeddings and make sure the scaling worked as expected.

felipemello1: Maybe add a TODO before merging, if you think it is worth it

tests/torchtune/models/llama3_1/test_position_embeddings.py
```diff
@@ -0,0 +1,137 @@
...
+    https://gist.github.com/joecummings/4f1331a9c1e5aa15bad1641acb74fe0e
+    """
+
+    EXPECTED_FREQS_CIS_MEAN = tensor(0.1738)
+    EXPECTED_FREQS_CIS_SUM = tensor(91141.7656)
+    EXPECTED_FREQS_CIS_MAX = tensor(1.0)
+
+    EXPECTED_X_OUT_MEAN = tensor(-2.4781e-06)
+    EXPECTED_X_OUT_SUM = tensor(-83.1523)
+    EXPECTED_X_OUT_MAX = tensor(5.4625)
```
ebsmothers, line 40: nit: I feel like it's overkill to assert on all of these. Nbd for this PR though

tests/torchtune/models/llama3_1/test_position_embeddings.py
```diff
@@ -0,0 +1,137 @@
...
+        return torch.randn(bsz, seq_len, num_heads, head_dim)
+
+    @pytest.fixture
+    def rope(self, input_params) -> Llama3ScaledRoPE:
+        _, _, head_dim, _, max_seq_len = input_params
+        return Llama3ScaledRoPE(dim=head_dim, max_seq_len=max_seq_len)
+
+    def test_cache_equality(self, input, rope) -> None:
+        # Have to explicitly call _rope_init() to initialize theta matrix
+        rope._rope_init()
```
ebsmothers, line 64: Maybe I'm being dense but why don't we have to do this on every test case?

tests/torchtune/models/llama3_1/test_position_embeddings.py
ebsmothers: Based on this discussion would also consider renaming from `test_forward_with_packed_pos` as I think it's a bit misleading (maybe `test_forward_with_2d_pos_ids` or something)

tests/torchtune/models/llama3_1/test_position_embeddings.py
```diff
@@ -0,0 +1,137 @@
...
+
+    def test_rope_init_meta_device(self, input_params):
+        _, _, head_dim, _, max_seq_len = input_params
+        rope_on_device = Llama3ScaledRoPE(dim=head_dim, max_seq_len=max_seq_len)
+        with torch.device("meta"):
+            meta_rope = Llama3ScaledRoPE(dim=head_dim, max_seq_len=max_seq_len)
+
+        meta_rope._rope_init()
+        for p1, p2 in zip(rope_on_device.buffers(), meta_rope.buffers()):
+            torch.testing.assert_close(p1, p2)
```
ebsmothers, line 137: Can also explicitly assert that nothing is on meta device

ebsmothers (approved):
Couple small comments, stamping to unblock

tests/torchtune/models/llama3_1/test_position_embeddings.py
joecummings: BUT HOW WILL WE KNOW IT'S THE SAME

tests/torchtune/models/llama3_1/test_position_embeddings.py
joecummings: This is the only test that doesn't call the rope forward method, which is where the cache is actually set. So for this one and only this one, we need to manually initialize the cache.

PR #1218, opened 2024-07-24
Allow BF16 dtype support on CPU

sanchitintel:
#### Description
PyTorch supports BF16 dtype for CPUs. If CPUs don't support some BF16-related ISAs such as AVX512_BF16 & AMX_BF16, BF16 <-> FP32 conversions are done (compute happens in FP32, in these cases). 

#### Changelog
BF16 dtype support check would now return True for CPUs. 
It was returning `False` earlier in the `quantize.py` recipe if PyTorch was installed without CUDA support and device type was set to `cpu` in a config yaml file because no value for the `device` argument was being provided in the invocation to `utils.get_dtype()`, so it was defaulting to `None`, and `utils.get_dtype()` was throwing an error about the device not supporting BFloat16 .
Added `device` argument in calls to `utils.get_dtype()`.

#### Test plan

Verified manually that the quantization recipe on CPU is not failing due to `utils.get_dtype()` returning False.
I could add a UT for CPU device, if required. Thanks!

### DIFF
--- a/docs/source/deep_dives/recipe_deepdive.rst
+++ b/docs/source/deep_dives/recipe_deepdive.rst
@@ -137,7 +137,7 @@ Initialize recipe state including seed, device, dtype, metric loggers, relevant 
     def __init__(...):
 
         self._device = utils.get_device(device=params.device)
-        self._dtype = utils.get_dtype(dtype=params.dtype)
+        self._dtype = utils.get_dtype(dtype=params.dtype, device=self._device)
         ...
 
 Load checkpoint, update recipe state from checkpoint, initialize components and load state dicts from checkpoint
--- a/recipes/eleuther_eval.py
+++ b/recipes/eleuther_eval.py
@@ -182,7 +182,7 @@ class EleutherEvalRecipe(EvalRecipeInterface):
 
     def setup(self) -> None:
         self._device = utils.get_device(device=self._cfg.device)
-        self._dtype = utils.get_dtype(dtype=self._cfg.dtype)
+        self._dtype = utils.get_dtype(dtype=self._cfg.dtype, device=self._device)
         self._limit = self._cfg.limit
         self._tasks = list(self._cfg.tasks)
         self._quantizer = config.instantiate(self._cfg.quantizer)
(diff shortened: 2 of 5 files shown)


joecummings:
@ebsmothers Any reason not to do this?

joecummings (commented):
@sanchitintel Great PR! Could you just run lint for these files?

sanchitintel:
Hi @joecummings, sorry, I haven't verified this change. I'll request review after verifying it, and will remove Draft mode for this PR. Thanks!

sanchitintel:
Hi @joecummings, the PR is now ready for review. Thanks!

At my end, though,[ running `pre-commit install` to set up the linter](image) results in a run-time error `importlib.metadata.PackageNotFoundError: No package metadata was found for pre-commit`, although I installed `pre-commit`:

image

Please advise if you've encountered such an issue before.

SalmanMohammadi:
Thanks for adding this @sanchitintel : ) pretty neat change

ebsmothers:
Thanks for the PR! I partially agree with @SalmanMohammadi -- `get_default_device` is probably the cleanest way to handle this. But I also think we should still pass the device explicitly once we infer it in the recipe (otherwise we have to infer it in the dtype utilities which creates needless separation on where defaults are defined).

So I think we can delete [_get_device_type_from_env](https://github.com/pytorch/torchtune/blob/0057fe7cf83e14f0b62538a8d4d20719e0a88639/torchtune/utils/_device.py#L52) and just replace its usage [here](https://github.com/pytorch/torchtune/blob/0057fe7cf83e14f0b62538a8d4d20719e0a88639/torchtune/utils/_device.py#L115) with a call to `get_default_device`. Then when we call `get_device` in the recipe (e.g. [here](https://github.com/pytorch/torchtune/blob/0057fe7cf83e14f0b62538a8d4d20719e0a88639/recipes/generate.py#L38) for the generate recipe) we can pass that explicitly to `get_dtype` as you've done here.

sanchitintel:
Thanks a lot for your feedback, @SalmanMohammadi & @ebsmothers!

I reverted the change pertaining to `torch.get_default_device`, and also removed the related comment.

ebsmothers (approved):
Looks good, thanks for fixing this!

PR #1220, opened 2024-07-25
Added adapter_only option to LoRA

### LINKED ISSUE #1212: [Feature] Enable Adapter Only Checkpointing for LoRA
LoRA recipes don’t need to save the merged checkpoint at each epoch as the much smaller adapter weights plus the original checkpoint is all you need. Checkpointer has an option on save_checkpoint called adapter_only that only saves the adapter weights. This PR would be to update all Lora based recipes to pass adapter_only=true for each epoch except the final one to get a final saved model.

spider-man-tm:
#### Context
- [x] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

close #1212

#### Changelog
In #1210, the "adapter_only" (boolean) option was added to the save_checkpoint method of each Checkpointer class. When this option is set to True, only the adapter weights are saved instead of the entire model weights.
This PR applies that change to LoRA fine-tuning.

#### Test plan

### DIFF
--- a/docs/source/deep_dives/checkpointer.rst
+++ b/docs/source/deep_dives/checkpointer.rst
@@ -404,6 +404,11 @@ looks something like this:
     # set to True if restarting training
     resume_from_checkpoint: True
 
+    # adapter_only option
+    # Set to True to save only the adapter weights for intermediate epochs.
+    # For the final epoch, the entire model weights will be saved regardless of this option.
+    adapter_only: False
+
 |
 
 Putting this all together
--- a/docs/source/deep_dives/checkpointer.rst
+++ b/docs/source/deep_dives/checkpointer.rst
@@ -404,10 +404,8 @@ looks something like this:
     # set to True if restarting training
     resume_from_checkpoint: True
 
-    # adapter_only option
-    # Set to True to save only the adapter weights for intermediate epochs.
-    # For the final epoch, the entire model weights will be saved regardless of this option.
-    adapter_only: False
+    # Set to True to save only the adapter weights
+    save_adapter_weights_only: False
 
 |
 
--- a/docs/source/deep_dives/checkpointer.rst
+++ b/docs/source/deep_dives/checkpointer.rst
@@ -361,7 +361,7 @@ Checkpointing for LoRA
 In torchtune, we output both the adapter weights and the full model "merged" weights
 for LoRA. The "merged" checkpoint can be used just like you would use the source
 checkpoint with any post-training tools. For more details, take a look at our
-:ref:`LoRA Finetuning Tutorial <lora_finetune_label>`.
+:ref:`LoRA Finetuning Tutorial <lora_finetune_label>`.Additionally, by setting the option "save_adapter_weights_only" to True when saving a checkpoint, you can choose to save only the adapter weights.
 
 The primary difference between the two use cases is when you want to resume training
 from a checkpoint. In this case, the checkpointer needs access to both the initial frozen
--- a/recipes/lora_dpo_distributed.py
+++ b/recipes/lora_dpo_distributed.py
@@ -539,7 +539,7 @@ class LoRADPORecipeDistributed(FTRecipeInterface):
                     "Saving final model checkpoint."
                     "Please note that you have set save_adapter_weights_only=True, so only adapter weights will be saved."
                     "You need to merge the adapter weights into your base model for further use. "
-                    "See {where do we point them for now??}"
+                    "See 'https://github.com/pytorch/torchtune/blob/main/docs/source/deep_dives/checkpointer.rst'"
                 )
 
     def concatenated_forward(
--- a/recipes/lora_dpo_distributed.py
+++ b/recipes/lora_dpo_distributed.py
@@ -143,6 +143,7 @@ class LoRADPORecipeDistributed(FTRecipeInterface):
         self.global_step = 0
 
         self._resume_from_checkpoint = cfg.resume_from_checkpoint
+        self._adapter_only = cfg.get("adapter_only", False)
         self._gradient_accumulation_steps = cfg.gradient_accumulation_steps
 
     def load_checkpoint(self, cfg_checkpointer: DictConfig) -> Dict[str, Any]:
@@ -526,11 +527,20 @@ class LoRADPORecipeDistributed(FTRecipeInterface):
                     }
                 )
 
-            self._checkpointer.save_checkpoint(
-                checkpoint_dict,
-                epoch=epoch,
-                intermediate_checkpoint=intermediate_checkpoint,
-            )
+            # If the option was True, save only the adapter except for the last epoch
+            if self._adapter_only and epoch + 1 < self.total_epochs:
+                self._checkpointer.save_checkpoint(
+                    checkpoint_dict,
+                    epoch=epoch,
+                    intermediate_checkpoint=(epoch + 1 < self.total_epochs),
+                    adapter_only=True,
+                )
+            else:
+                self._checkpointer.save_checkpoint(
+                    checkpoint_dict,
+                    epoch=epoch,
+                    intermediate_checkpoint=(epoch + 1 < self.total_epochs),
+                )
 
     def concatenated_forward(
         self, model: nn.Module, batch: Tuple[torch.Tensor, torch.Tensor]
--- a/recipes/lora_dpo_distributed.py
+++ b/recipes/lora_dpo_distributed.py
@@ -539,7 +539,7 @@ class LoRADPORecipeDistributed(FTRecipeInterface):
                     "Saving final model checkpoint."
                     "Please note that you have set save_adapter_weights_only=True, so only adapter weights will be saved."
                     "You need to merge the adapter weights into your base model for further use. "
-                    "See 'https://github.com/pytorch/torchtune/blob/main/docs/source/deep_dives/checkpointer.rst'"
+                    f"See {type(self._checkpointer).__name__}"
                 )
 
     def concatenated_forward(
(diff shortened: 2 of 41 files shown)


spider-man-tm:
To the reviewer,

Please let me know if there are any missing parts in my implementation or if there are any parts that should be removed. It's perfectly fine if the maintainer edits it directly.

recipes/lora_dpo_distributed.py
```diff
@@ -526,11 +531,21 @@ def save_checkpoint(
                     }
                 )
 
-            self._checkpointer.save_checkpoint(
-                checkpoint_dict,
-                epoch=epoch,
-                intermediate_checkpoint=intermediate_checkpoint,
-            )
+            # If the option was True, save only the adapter except for the last epoch
+            is_intermediate_epoch = epoch + 1 < self.total_epochs
```
SalmanMohammadi, line 535: Thanks so much for your contribution : )
Small q: do we want to support the case where merged weights are never saved, even in the final epoch, and let the user merge them? Might be helpful when working with particularly large models. @joecummings

docs/source/deep_dives/checkpointer.rst
```diff
@@ -404,6 +404,11 @@ looks something like this:
     # set to True if restarting training
     resume_from_checkpoint: True
 
+    # adapter_only option
+    # Set to True to save only the adapter weights for intermediate epochs.
+    # For the final epoch, the entire model weights will be saved regardless of this option.
+    adapter_only: False
```
joecummings, line 410: nit: Considering `save_adapter_weights_only` instead of `adapter_only` for clarity. 

I know it's different from the param name on the checkpointer, but the extra context isn't needed there b/c you already know you're dealing with saving a checkpointing.

cc @SalmanMohammadi  and @pbontrager  for thoughts.

recipes/lora_dpo_distributed.py
joecummings: Yep, I think @SalmanMohammadi's point is inline with what I was thinking here. This does mean that we need to do a better job explaining how to merge weights or what exactly to do with them afterwards.

recipes/lora_dpo_distributed.py
```diff
@@ -143,6 +143,7 @@ def __init__(self, cfg: DictConfig) -> None:
         self.global_step = 0
 
         self._resume_from_checkpoint = cfg.resume_from_checkpoint
+        self._adapter_only = cfg.get("adapter_only", False)
```
joecummings, line 146: Can you add a logging INFO statement saying that only the adapter weights will be saved so the user knows right away?

joecummings (commented):
This is looking great - thanks so much :)

Just a couple comments.

recipes/lora_dpo_distributed.py
joecummings: Hmmmm, this does make sense to me. I'd much rather have the error propagated from the checkpointer, rather than the recipe file.

I think the complexity arises in how we actually instantiate the checkpointer. I don't think we want `save_adapter_weights_only` as a param on the initialization of the checkpointer b/c the user should be able to take the component and save adapters or the whole file if they want without re-initializing it. Therefore, it has to be a method on `save`.

So if we don't want it being passed to the constructor, we would have to parse out the `save_adapter_weights_only` from the config file before creating the checkpointer, which feels a little messy. Maybe we push the warning to the `save` method of the checkpointer? The only issue with this is that then the user could accidentally have the `save_adapter_weights_only=True`, train their whole model, and then save the checkpoints without actually wanting that feature. But that might be too hand-holdy to worry about?

Thanks for coming on this journey of vomiting all my thoughts on this PR. I think my TL;DR is that this config variable should be separate from the instantiation of the checkpointer, but maybe we push the logging info to the `save` method instead of in the main recipe. 

Thoughts?

SalmanMohammadi: >  maybe we push the logging info to the save method instead of in the main recipe.

The `save_checkpoint_method` of the checkpointer right? I agree here - we also validate `adapter_only` there.

recipes/lora_dpo_distributed.py
SalmanMohammadi: Concretely, should we fix `adapter_only=self._save_adapter_only` here, and add something to the effect of:
```
if not is_intermediate_epoch:
    log.info("Saving final model checkpoint. Please note that you have set save_adapter_only=True, so only adapter weights will be saved. You will need to merge the adapter weights into your base model for further use. See {where do we point them for now??}")
```

recipes/lora_dpo_distributed.py
ebsmothers: Generally agree with @joecummings here.. the checkpointer class itself should mostly care about stuff that's global across a given fine-tuning run (input and output checkpoint formats). The flag to save only adapter weights is a more local thing since it can in theory vary depending on where we are in training. So imo it makes sense to expose only in `save_checkpoint` and not in init.

A separate (but still relevant) point: our checkpointer naming still seems to imply it's only used for full fine-tuning. This is confusing and another reason why it'd look weird to put `save_adapter_weights_only` in its init. (I don't think that's a good reason to keep `save_adapter_weights_only` outside the checkpointer init, I actually think we should just rename the checkpointer instead)

spider-man-tm:
Thank you to all reviewers. I have made the following revisions:

- The name of the option `adapter_only` was changed to `save_adapter_weights_only`. While `adapter_only` is understandable within the `save_checkpoint` method, it becomes ambiguous when used independently.
- To avoid complications, the option is retained as a method option for `save_checkpoint` rather than being included in the Checkpointer constructor.
- Adjusted the code to maintain consistency by removing the special case where all weights were saved only for the last epoch.
- Added log output to inform the user that only adapter weights are being saved.
  - Due to a too long error occurring when running `pre-commit run --all-files`, the logs are spread over 4 lines.

Feel free to review the changes and provide further feedback!

recipes/lora_dpo_distributed.py
```diff
@@ -143,6 +143,9 @@ def __init__(self, cfg: DictConfig) -> None:
         self.global_step = 0
 
         self._resume_from_checkpoint = cfg.resume_from_checkpoint
+        self._save_adapter_weights_only = cfg.get("save_adapter_weights_only", False)
+        log.info(f"save_adapter_weights_only: {self._save_adapter_weights_only}")
```
ebsmothers, line 147: This is more of a nit since I know @joecummings gave contrary advice here already, but I don't like logging config fields in the recipe like this. We already log the full config, no need to just directly re-log an individual config field unless there is something more non-trivial happening (e.g. we use some feature that depends on a particular version of PyTorch or something).

ebsmothers (commented):
Really appreciate you adding this! Just two minor things then I think this is good to go

spider-man-tm:
@ebsmothers 

Thank you for your review! I have made the corrections based on your feedback.

> Maybe I'm being dense but is this log correct? I don't see where we actually check that save_adapter_weights_only=True prior to logging this

You’re absolutely right, this could indeed confuse the user. I’ve now modified the code to log the message based on the value of the option.

```python
            if not is_intermediate_epoch:
                log.ingo("Saving final epoch checkpoint.")
                if self._save_adapter_weights_only:
                    log.info(
                        "Please note that you have set save_adapter_weights_only=True, so only adapter weights will be saved."
                        "You need to merge the adapter weights into your base model for further use. "
                        f"See {type(self._checkpointer).__name__}"
                    )
                else:
                    log.info(
                        "The full model checkpoint, including all weights and configurations, has been saved successfully."
                        "You can now use this checkpoint for further training or inference."
                    )
```

> We already log the full config, no need to just directly re-log an individual config field unless there is something more non-trivial happening

I’ve removed the log output for that section. I agree that it’s best not to deviate too much from the overall logging pattern for the other options.

recipes/lora_finetune_single_device.py
```diff
@@ -514,11 +518,20 @@ def save_checkpoint(self, epoch: int) -> None:
...
+            intermediate_checkpoint=is_intermediate_epoch,
+            adapter_only=self._save_adapter_weights_only,
         )
+        if not is_intermediate_epoch:
+            log.info(
+                "Saving final model checkpoint."
+                "Please note that you have set save_adapter_weights_only=True, so only adapter weights will be saved."
+                "You need to merge the adapter weights into your base model for further use. "
+                f"See {type(self._checkpointer).__name__}"
+            )
```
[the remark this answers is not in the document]
SalmanMohammadi: Good catch. Sorry, my original suggestion missed this.

Do we still think it's a good idea to move this into the checkpointer, rather than recipes?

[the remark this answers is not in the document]
spider-man-tm: @SalmanMohammadi 

Indeed, rather than writing similar code each time we create a new recipe, it might be better to create it with a checkpointer. What do you think?

spider-man-tm:
I moved the log output to checkpointer.
https://github.com/pytorch/torchtune/pull/1220/commits/e8e8757783ce2a784680734306c9c4676a2f47fc

recipes/lora_finetune_single_device.py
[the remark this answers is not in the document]
ebsmothers: > Indeed, rather than writing similar code each time we create a new recipe, it might be better to create it with a checkpointer. What do you think?

Good point, I'm inclined to agree with this (kinda similar to my other comment, it's nice to keep the recipe files themselves as clean as possible)

SalmanMohammadi:
LGTM. Thanks so much for your contribution, and your patience in addressing our comments : )

SalmanMohammadi (approved):

spider-man-tm:
@SalmanMohammadi 
PyTorch is my favorite ML package, and I'm happy to have contributed to it. Thanks so much!

joecummings (approved):
Awesome work!

Issue #1017, opened 2024-05-24
[RFC] TransformerDecoder refactor


SalmanMohammadi:
# `TransformerDecoder` Refactor

**Authors:**
* @SalmanMohammadi
with input from:
* @kartikayk 
* @ebsmothers 
* @pbontrager 
## **Summary**

Refactoring `TransformerDecoder` to offer additional flexibility for new use-cases.

## **Motivation/Prior art**

- https://github.com/pytorch/torchtune/issues/968 - not sure if this is in scope.
- https://github.com/pytorch/torchtune/pull/840

Currently, `TransformerDecoder` can only be used for language-modelling tasks. There is interest in additional use-cases, such as:
- Classifier models akin to HF's [`AutoModelForSequenceClassification`](https://huggingface.co/transformers/v3.0.2/model_doc/auto.html#automodelforsequenceclassification)
- Value head models akin to TRL's [`AutoModelForCausalLMWithValueHead`](https://github.com/huggingface/trl/blob/d1aa0b6b2c8dfd78c0f771759d1ff2469c0e5ed2/trl/models/modeling_value_head.py#L62) 
- Reward models with custom initialisation as in [`lm_human_preference_details`](AutoModelForCausalLMWithRewardHead)
- Using arbitrary hidden states from the transformer backbone
- ...

Such a refactor could allow users to easily adapt a transformer backbone for a variety of down-stream tasks; `lm_human_preference_details` demonstrates how HF's transformer backbone can be extended [in just 8 lines](https://github.com/vwxyzjn/lm-human-preference-details/blob/ccc19538e817e98a60d3253242ac15e2a562cb49/lm_human_preference_details/train_policy_accelerate.py#L314). While this refactor initially targets recipes which will be provided within Torchtune, such as [PPO](https://github.com/pytorch/torchtune/pull/1005), or sequence-classification training recipes (e.g. for reward models), it would allow users to write custom recipes for many fine-tuning tasks whilst utilising underlying Torchtune features.

## **Proposed Implementation**

A small-scale implementation for Mistral models exists in [this draft PR](https://github.com/pytorch/torchtune/pull/1005). In summary:

[...]

PR #1224, opened 2024-07-25
[RFC] TransformerDecoderLayer Refactor

pbontrager:
# [RFC] TransformerDecoderLayer Refactor

Refactor TransformerDecoder so it can be used for multimodal architectures.

**TLDR**
- Replace TransformerDecoderLayer with TransformerSelfAttention and TransformerCrossAttention
- Replace CausalSelfAttention with GroupedQueryAttention
- Support legacy Checkpoints

**WHY**: because we need to support a mix of cross attention and self attention layers to support encoders and deep fusion multimodal models in the library. Allowing different transformer layer types allows us to support most variants of transformers while keeping the complexity of the attention module with a reusable GroupedQueryAttention.# [RFC] TransformerDecoderLayer Refactor

**NOTE** The RFC is also includes in an md file so that everyone can comment on specific lines more easily.

## Context

Currently `TransformerDecoder` with `TransformerDecoderLayer` is a decoder only style transformer. As opposed to the original transformer (image below), decoder only style models don't take inputs from an encoder and remove the cross attention block (green) from the decoder block (`TransformerDecoderLayer`). Since most SOTA LLMs have been GPT style decoder only models for the last few years, only supporting Decoder layers have worked well until now.

<p align="center">
  
</p>

But looking forward at the advances in multimodal LLMs, it's clear that encoder-decoder model architectures will become important again [ref](https://scontent-iad3-1.xx.fbcdn.net/v/t39.2365-6/452387774_1036916434819166_4173978747091533306_n.pdf?_nc_cat=104&ccb=1-7&_nc_sid=3c67a6&_nc_ohc=t6egZJ8QdI4Q7kNvgGwBp2W&_nc_ht=scontent-iad3-1.xx&oh=00_AYC2OwSKC1f5iJMNAiAj48_5_u1cmYjC1nPcIizQmBu7HQ&oe=66A6EB8D). This brings with it several new requirements for TransformerDecoderLayers:
- The need to support self attention (decoder x decoder) and cross attention (decoder x encoder). Another way to think of this is that decoder blocks need to be able to support their primary input (tokens) and a conditional input (multimodal embeddings).
- The need to support both self attention masks and cross attention (aka conditional) masks. To work with encoders, layers also need to relax the assumption that self attention is always causal.
- The need to support mixtures of layer types where some might require conditional inputs, some might not.

**TLDR** Layers need to support a primary input and mask (current inputs) and an optional conditional input and mask. This would support any standard variation of TransformerBlock.

**Note on Attention** In the diagram above, it should be noted that all three attention implementations (red, green, and blue) use the same attention module, Multi Headed Attention. Most of the implementation complexity lie inside this module, and it's only getting more complex [#1193](https://github.com/pytorch/torchtune/pull/1193). The only difference between them is what inputs are provided as k, q, v, and what mask is provided. All of that to say, to support cross attention and self attention, the attention module doesn't change but the transformer block itself. See Figure 2, Deep Fusion, for an overview of some popular layer variations for multimodal models [ref](https://arxiv.org/pdf/2405.17927).
## Proposal
**High Level List:**
- Extend the TransformerDecoder forward signature to include `encoder_input` and `encoder_mask`. This sets a standard signature for all layers to follow so they can all be mixed and matched and stacked together.
- Modify TransformerDecoder to accept a list of layers instead of a layer and and a count. This allows us to support models where only some layers are conditional and other are not.
- Implement two initial layers `TransformerSelfAttentionLayer` and `TransformerCrossAttentionLayer`. `TransformerSelfAttentionLayer` would replace `TransformerDecoderLayer`. These would just be two possible layers, but the design would support more variations.
- Generalize `CausalSelfAttention` work in all three scenarios above (red, green, blue) and rename to `GroupedQueryAttention`. This includes changing the default mask behavior to not always be causal.
- **Extra**:
	- Add TanhGate module to support gated attention [ref](https://paperswithcode.com/method/attention-gate)
	- Modify `TransformerDecoder` to output it's hidden layers and make `model.output` optional to support the changes proposed [#1017](https://github.com/pytorch/torchtune/issues/1017) and make only one BC breaking change.
	- Implement legacy checkpoint conversion to support old checkpoints

### TransformerDecoder
TransformerDecoder only highlighting the changes, current module [here](https://github.com/pytorch/torchtune/blob/0057fe7cf83e14f0b62538a8d4d20719e0a88639/torchtune/modules/transformer.py#L99). Discussion in the comments.

```python
class TransformerDecoder(nn.Module):
    def __init__(
        self,
        *,
        ...
        # as opposed to layer and num_layers
        layers: List[nn.Module],
        ...
        # output is generalized and hidden_idx is added to address #1017
        output: Callable
        hidden_idx: Optional[List[int]] = None
    ) -> None:
        super().__init__()
        # We could keep an optional layer + num_layers args for BC
        # Otherwise this change requires builders to go from
        # TransformerDecoder(layer, 32)
        # To
        # for _ in range(32):
	    #    layers.append(layer)
	    # TransformerDecoder(layers)
        ...

    def forward(
        self,
        tokens: Tensor,
        *,
        encoder_input: Optional[Tensor] = None,
        encoder_mask: Optional[Tensor] = None,
    ) -> [Tensor, List[Tensor]]:
        ...

		# Move max seq length check from attention to here
        if seq_len > self.max_seq_len:
            raise ValueError("Too long")

        ...

		hidden = []
        for i, layer in enumerate(self.layers):
	        if i in self.hidden_idx:
		        hidden.append(h)
            h = layer(
                h,
                mask=mask,
                encoder_input=encoder_input,
                encoder_mask=encoder_mask,
                input_pos=input_pos,
            )

        # shape: [b, s, d]
        h = self.norm(h)

        # this unifies the output with ViT but breaks recipes
        # Alternative:
        # output = output if not hidden else hidden + [output]
        return output, hidden
```
@SalmanMohammadi does this design address all your needs from #1017
### TransformerSelfAttentionLayer
TransformerDecoderLayer only highlighting the changes, current module [here](https://github.com/pytorch/torchtune/blob/0057fe7cf83e14f0b62538a8d4d20719e0a88639/torchtune/modules/transformer.py#L15). Discussion in the comments.

```python
class TransformerSelfAttentionLayer(nn.Module):
    def __init__(
        self,
        attn: GroupedQueryAttention, # Updated Attn module
        mlp: nn.Module,
        *,
        # sa_norm renamed since it's no longer only self attention
        # this requires repo wide changes for sa_norm -> attn_norm
        attn_norm: Optional[nn.Module] = None,
        mlp_norm: Optional[nn.Module] = None,
        # New scale modules to allow scaling/gating attn output
        attn_scale: Optional[nn.Module] = None,
        mlp_scale: Optional[nn.Module] = None,
    ) -> None:
        super().__init__()
        self.attn = attn
        self.mlp = mlp
        self.attn_norm = attn_norm or nn.Identity()
        self.mlp_norm = mlp_norm or nn.Identity()
        self.attn_scale = attn_scale or nn.Identity()
        self.mlp_scale = mlp_scale or nn.Identity()

    def forward(
        self,
        x: Tensor,
        *,
        **kwargs, # encoder_input and encoder_mask is not used for this module
    ) -> Tensor:
	    # Identical to TransformerDecoderLayer with addition of *_scale
        attn_out = self.attn(self.attn_norm(x), mask=mask, input_pos=input_pos)

        h = self.attn_scale(attn_out) + x

        mlp_out = self.mlp(self.mlp_norm(h))

        out = h + self.mlp_scale(mlp_out)
        return out
```

### TransformerCrossAttentionLayer
New layer. Discussion in the comments.
```python
class TransformerCrossAttentionLayer(nn.Module):
    def __init__(
		...
    ) -> None:
        super().__init__()
        assert (
            attn.pos_embeddings is None
        ), "Positions are not computed for encoder inputs"
        ...

    def _skip_mask(self, mask: Optional[Tensor]) -> Optional[Tensor]:
	    """Some tokens in x may not attend to any encoder inputs
        due to the cross attention mask (encoder_mask). This results in
        a full row of the attention matrix being masked out.

        In the example below, the word "the" is masked from every embedding.

        .. code-block:: text

            |emb||emb||emb|
        |The| x    x    x
        |red|      x
        |car| x

        This results in no inputs into the softmax layer which causes a NaN.
        The skip mask removes the output of the attention module and
        mlp resulting in the token being skipped.

        """
		# There is more discussion around this here
		# https://gist.github.com/drisspg/547648ded500d078961b7a3b3f11c310
		# We might be able to optimize this approach more
        if mask is None:
            return None
        if mask.dtype == torch.bool:
            mask = ~mask
        else:
            mask = torch.isneginf(mask)
        mask = torch.all(mask, dim=-1, keepdim=True)
        return mask

    def forward(
        self,
        x: Tensor,
        *,
        encoder_input: Optional[Tensor] = None,
        encoder_mask: Optional[Tensor] = None,
        **kwargs,
    ) -> Tensor:
        # Cross attention only layer (see red only) has no
        # purpose if no conditional input is provided
        if encoder_input is None:
            return x

        # A mask of tokens (x) with no encoder_input
        skip_mask = self._skip_mask(encoder_mask)

        # Input tensor and attention output have the same shape
        # [b, s, d]
        # Norm applied before self-attention
        attn_out = self.attn(
            self.attn_norm(x),
            encoder_input,
            mask=encoder_mask,
            input_pos=input_pos
        )
        if skip_mask is not None:
            attn_out.masked_fill_(skip_mask, 0)

        # Residual connection; shape: [batch_size, seq_length, embed_dim]
        h = self.attn_scale(attn_out) + x

        # Norm applied before the feedforward layer
        mlp_out = self.mlp(self.mlp_norm(h))
        if skip_mask is not None:
            mlp_out.masked_fill_(skip_mask, 0)

        # Residual connection; shape: [batch_size, seq_length, embed_dim]
        out = h + self.mlp_scale(mlp_out)
        return out
```
@Chillee the `_skip_mask` is an example of a fully masked out row. Is applying the mask this way the most efficient way to handle this?
### GroupedQueryAttention
CausalSelfAttention only highlighting the changes, current module [here](https://github.com/pytorch/torchtune/blob/main/torchtune/modules/attention.py). Discussion in the comments.
```python
class GroupedQueryAttention(nn.Module):
    def __init__(
        self,
        *,
        ...
        # made this optional to support encoders
        pos_embeddings: Optional[nn.Module] = None,
        # support qk normalization
        # https://arxiv.org/abs/2010.04245
        q_norm: Optional[nn.Module] = None,
		k_norm: Optional[nn.Module] = None,
		# sets the default mask wehn none is provided
		# to be causal or None
        default_causal_mask: bool = True,
    ) -> None:
        super().__init__()
        ...
        if bool(q_norm) ^ bool(k_norm):
            raise ValueError("q and k norm must be set together")
		...

    def forward(
        self,
        x: Tensor,
        # Optinal for cross attention but we could
        # make it required instead and self attention
        # would be GQA(x, x)
        y: Optional[Tensor] = None,
        *,
    ) -> Tensor:
        ...
        bsz, seq_len_x, _ = x.shape
        y = y if y is not None else x
        _, seq_len_y, _ = y.shape

        # q has shape [b, s, num_heads * head_dim]
        # k has shape [b, s, num_kv_heads * head_dim]
        # v has shape [b, s, num_kv_heads * head_dim]
        q = self.q_proj(x)
        k = self.k_proj(y)
        v = self.v_proj(y)

        ...

        # Apply positional embeddings
        if self.pos_embeddings is not None:
            q = self.pos_embeddings(q, input_pos=input_pos)
            k = self.pos_embeddings(k, input_pos=input_pos)

        ...

        # Normalize k and q
        if self.q_norm is not None:
            q = self.q_norm(q)
            k = self.k_norm(k)

        ...

        output = nn.functional.scaled_dot_product_attention(
            q,
            k,
            v,
            attn_mask=mask,
            dropout_p=self.attn_dropout,
            # if you're using a kv_cache or a mask is provided
            # then don't apply a default causal mask
            # if there is no mask,
            # then only apply a causal mask if GQA has default_causal_mask
            is_causal=self.kv_cache is None and mask is None and self.is_causal,
        )

        ...

```

### Legacy Checkpointing

Since GQA changes sa_norm to attn_norm, this will break existing checkpoints that are in the tune format. It could be argued that since we don't really let users save checkpoints in the tune format by default, that this would not affect anyone. But to ensure that no checkpoint are broken, we can add conversion mappings to [convert_weights.py](https://github.com/pytorch/torchtune/blob/main/torchtune/models/convert_weights.py). These would be exactly the same as `_FROM_HF` or `_FROM_META` but instead map from older version to new.

```python
_FROM_TUNE_0_2_1 = {
    "layers.{}.sa_norm.scale": "layers.{}.attn_norm.scale",
}
```
The above code block is assuming that the mapping function is updated to treat missing keys as an identity

The state dict can be converted based on the provided version string `#.#.#`
```python
def _legacy_to_tune(
    state_dict: Dict[str, torch.Tensor], version: str
) -> Dict[str, torch.Tensor]:
    converted_state_dict = {}
    v1, v2, v3 = version.split(".")
    for key, value in state_dict.items():
        new_key = get_mapped_key(key, eval(f"_FROM_TUNE_{v1}_{v2}_{v3}"))
        converted_state_dict[new_key] = value

    return converted_state_dict

```

Finally, FullModelTorchTuneCheckpointer would be updated to check for signatures of older checkpoints, and then call `_legacy_to_tune`. This change is only applied to 'load_checkpoint' so that a users checkpoint is converted to the newer format but never back to the older format.
```python

class FullModelTorchTuneCheckpointer(_CheckpointerInterface):
	...

    def load_checkpoint(self, weights_only: bool = True) -> Dict[str, Any]:
        ...

		if "layers.0.sa_norm.scale" in model_state_dict:
			logger.info(
				"This is an older checkpoint format. Converting \
				to the current version."
			)
			model_state_dict = convert_weights._legacy_to_tune(
				model_state_dict, "0.2.1"
			)
			state_dict[utils.MODEL_KEY] = model_state_dict
		...

```
### Alternatives
@kartikayk expressed concerns over exposing SelfAttention and CrossAttention at the build level which might add complexity for users to parse. While we still requires users to pass in the attention module anyway, another way we could approach this would be to parametrize TransformerDecoderLayer to handle both cross attention and self attention.

```python
class TransformerDecoderLayer(nn.Module):
    def forward(
        self,
        x: Tensor,
        *,
        encoder_input: Optional[Tensor] = None,
        encoder_mask: Optional[Tensor] = None,
    ) -> Tensor:

        if encoder_input is None and self.cross_attn_layer:
            return x

		# return None if not self.cross_attn_layer
        skip_mask = self._skip_mask(encoder_mask)

		y = encoder_input if self.cross_attn_layer else x
		m = encoder_mask if self.cross_attn_layer else mask
        attn_out = self.attn(
            self.attn_norm(x),
            y,
            mask=m,
            input_pos=input_pos
        )
        if skip_mask is not None:
            attn_out.masked_fill_(skip_mask, 0)

        # Residual connection; shape: [batch_size, seq_length, embed_dim]
        h = self.attn_scale(attn_out) + x

        # Norm applied before the feedforward layer
        mlp_out = self.mlp(self.mlp_norm(h))
        if skip_mask is not None:
            mlp_out.masked_fill_(skip_mask, 0)

        out = h + self.mlp_scale(mlp_out)
        return out
```

This is a reasonable alternative but I do fear it'll get more and more complex to extend as we go forward and want to support more attention types.

### Risks/Concerns
- Attention module and model builders both get a bit more complex with this. The more variation we support the more complexity we introduce. I believe that supporting multiple transformer layers at least allows us to keep individual layers simpler.
- This change breaks BC both in the checkpoint itself but also in model builders. We can update all of our model builders, but if users have their own custom ones they will have to update them too. Do we have a mechanism for clearly communicating this in the release notes along with how to update a model builder?
- The cross attention layer here does not work with sample packing. Should we make layers that support sample packing through an error if they're used with packed data? @RdoubleA
- This proposal continues the use of the TransformerDecoder though technically it has been generalized that it could also be used for an encoder (it would need to ignore the kv cache stuff.). I think we should stick to the current name though as that is its primary purpose.


RFC.md
```diff
@@ -0,0 +1,413 @@
...
+```python
+class TransformerSelfAttentionLayer(nn.Module):
+    def __init__(
+        self,
+        attn: GroupedQueryAttention, # Updated Attn module
+        mlp: nn.Module,
+        *,
+        # sa_norm renamed since it's no longer only self attention
+        # this requires repo wide changes for sa_norm -> attn_norm
+        attn_norm: Optional[nn.Module] = None,
```
RdoubleA, line 114: can this still remain `sa_norm` since we're still in TransformerSelfAttentionLayer? I get that you'd want to keep the inits uniform across SA and XA layers, but if we are designing to accept any type of attention layer, would it be reasonable to relax this constraint? And then you won't have any checkpoint / BC breaking changes.

RdoubleA (commented):
Excellent breakdown. At a high level this approach makes sense to me. I do prefer the split `TransformerSelfAttentionLayer` and `TransformerCrossAttentionLayer` because it's very explicit when building a model and would make parsing model architectures more clear. Combining them into a single decoder layer is a bit opaque imo.

> The cross attention layer here does not work with sample packing.

Could you elaborate here? I had thought that it would still work as long as the images / encoder_masks are also packed

RFC.md
pbontrager: I guess it could remain as sa_norm. I prefer having these be the same for both layer types but there's nothing enforcing that.

pbontrager:
> Could you elaborate here? I had thought that it would still work as long as the images / encoder_masks are also packed

What I meant was it wouldn't work with packed "out of the box". We could extend dataset packing to support this, but in its current form it wouldn't. And there could likely be future layers that don't work with an existing implementation of packed.

RFC.md
```diff
@@ -0,0 +1,413 @@
...
+    def __init__(
+        self,
+        *,
+        ...
+        # as opposed to layer and num_layers
+        layers: List[nn.Module],
+        ...
+        # output is generalized and hidden_idx is added to address #1017
+        output: Callable
+        hidden_idx: Optional[List[int]] = None
```
joecummings, line 51: wth is this

pbontrager: List of hidden layers to return. It could also be called `hidden_layers`, `output_layers`, `output_indices`, etc.

kartikayk (commented):
Thanks for putting together this detailed RFC! Two high level comments:

- BC. I don't think we should prioritize preserving builder function BC over simplifying the design. We have not yet guaranteed BC as an alpha library and dont have very high library-style usage (which is going to change as torchchat and ET take dependency on the builders soon). I would like to see us drop the builder BC consideration, provide ways for folks to be able to reuse old checkpoints and simplify the APIs where we can

- GQA. I don't like this name. The module still supports all of MHA, MQA and GQA and calling this GQA seems very confusing to me (came up in some other discussions as well). I'd still like to see this be something more general such as MultiHeadAttention and clarify in the doc string what we support

pbontrager:
Thanks for the feedback. To ask two follow questions:
- Could you clarify this a bit more? Are there ways you think this could be simplified more? Most of the bc stuff is just reusing our existing checkpoint mapping to load in older checkpoints, but I wasn’t trying to limit the design too much. If there’s ways you think it’s too complex, or could be simplified please let me know.
- I’m okay with MHA, really what I wanted to communicate is that it’s a generalized MHA.

RFC.md
ebsmothers: Personally I would be inclined to break BC here for the sake of consistency across self- and cross-attention, otherwise it will just introduce extra remapping across the repo moving forward that will be more of a headache (especially since we are already breaking BC with these changes, we might as well do this while we can get away with it).

RFC.md
ebsmothers: Ya could be named better. Personally I just like `return_hidden_states` as a `Union[bool, List[int]]` that we convert to a `List[bool]` (or `List[int]`) on the backend; that way we can still easily support the common case of returning all hidden layers pretty easily.

RFC.md
```diff
@@ -0,0 +1,413 @@
...
+### Legacy Checkpointing
+
+Since GQA changes sa_norm to attn_norm, this will break existing checkpoints that are in the tune format. It could be argued that since we don't really let users save checkpoints in the tune format by default, that this would not affect anyone. But to ensure that no checkpoint are broken, we can add conversion mappings to [convert_weights.py](https://github.com/pytorch/torchtune/blob/main/torchtune/models/convert_weights.py). These would be exactly the same as `_FROM_HF` or `_FROM_META` but instead map from older version to new.
+
+```python
+_FROM_TUNE_0_2_1 = {
+    "layers.{}.sa_norm.scale": "layers.{}.attn_norm.scale",
+}
+```
+The above code block is assuming that the mapping function is updated to treat missing keys as an identity
```
ebsmothers, line 323: This is a secondary point to some of the other things in this RFC but I might reconsider this. It is probably best to be explicit about our conversions, especially with respect to external checkpoint formats (e.g. if some new key is added and we assume the identity mapping it can get tricky to debug). We can also consider our internal torchtune checkpoint versioning as a separate API, in which case we can assume the identity mapping for our own internal versioning but not for any other checkpoint remappings.

RFC.md
```diff
@@ -0,0 +1,413 @@
...
+        # Norm applied before the feedforward layer
+        mlp_out = self.mlp(self.mlp_norm(h))
+        if skip_mask is not None:
+            mlp_out.masked_fill_(skip_mask, 0)
+
+        # Residual connection; shape: [batch_size, seq_length, embed_dim]
+        out = h + self.mlp_scale(mlp_out)
+        return out
+```
+@Chillee the `_skip_mask` is an example of a fully masked out row. Is applying the mask this way the most efficient way to handle this?
```
ebsmothers, line 232: To confirm, can we fold `_skip_mask` into flex attention after https://github.com/pytorch/pytorch/pull/131012?

RFC.md
```diff
@@ -0,0 +1,413 @@
...
+New layer. Discussion in the comments.
+```python
+class TransformerCrossAttentionLayer(nn.Module):
+    def __init__(
+		...
+    ) -> None:
+        super().__init__()
+        assert (
+            attn.pos_embeddings is None
+        ), "Positions are not computed for encoder inputs"
```
ebsmothers, line 157: Is there a nicer way to do this? This will limit the flexibility of this module a bit, right?

RFC.md
```diff
@@ -0,0 +1,413 @@
...
+        *,
+        encoder_input: Optional[Tensor] = None,
+        encoder_mask: Optional[Tensor] = None,
+        input_pos: Optional[Tensor] = None,
+        **kwargs,
+    ) -> Tensor:
+        # Cross attention only layer (see red only) has no
+        # purpose if no conditional input is provided
+        if encoder_input is None:
+            return x
```
ebsmothers, line 203: I know why you decided to do this, just wanna point out that it's a bit unintuitive for generic cross-attention layer imo. We should make it abundantly clear that this is the behavior, otherwise people will get confused

RFC.md
```diff
@@ -0,0 +1,413 @@
...
+        skip_mask = self._skip_mask(encoder_mask)
+
+        # Input tensor and attention output have the same shape
+        # [b, s, d]
+        # Norm applied before self-attention
+        attn_out = self.attn(
+            self.attn_norm(x),
+            encoder_input,
+            mask=encoder_mask,
+            input_pos=input_pos
```
ebsmothers, line 215: Also unintuitive that we're passing this if position embeddings are forced to be None

RFC.md
SalmanMohammadi: 75% noob question: should `hidden_idx` be a property of `TransformerDecoder`? Or are users expected to have to re-initialise if they want different states?

ebsmothers (commented):
Overall no major concerns from my side. My main question is around the intersection of these changes with #1193, do you and @RdoubleA have a clear idea of how to integrate these?

RFC.md
```diff
@@ -0,0 +1,413 @@
...
+        # shape: [b, s, d]
+        h = self.norm(h)
+
+        output = self.output(h).float()
+        # this unifies the output with ViT but breaks recipes
+        # Alternative:
+        # output = output if not hidden else hidden + [output]
+        return output, hidden
+```
+@SalmanMohammadi does this design address all your needs from #1017
```
SalmanMohammadi, line 101: So the only thing we're changing here is the addition of returning hidden states, right?

I think the other half of #1017 was flexibility around how the output from the last hidden layer is used for some downstream task(s), particularly cases which could modify the return signature of `TransformerDecoder.forward`.

I admit since we're omitting joint policy-value head networks in PPO we currently don't have a concrete use case for this, so I'm not really concerned about this atm. If this isn't something that's on your horizon for supporting down the line, then I don't think we actually need #1017.

pbontrager: For joint policy heads, couldn't you just define your head as

```python
PolicyValueHead(nn.Module):
	...
def forward(self, h):
	p = self.policy(h)
	v = self.value(h)
	return p, v
```
and then set `output=PolicyValueHead()`? But also if we don't have a concrete use case for it, I'm fine with not worrying about it until we do.

SalmanMohammadi: Agreed, I don't the added complexity is worth it. Overall this is really great. Really clearly written and presented - thank you : )

RFC.md
ebsmothers: @SalmanMohammadi I was wondering about this too. I anticipate it may help with compile to have a fixed set of layers that are deterministically returned on each forward pass (though haven't actually tested this so please take with a huge grain of salt). On the other hand, `output_hidden_states` is usually defined in HF as part of the forward signature so it may be what users are accustomed to. 

But there are also certain things that may necessitate this being a static property of the class. E.g. if we are using hidden states as part of a fusion down the line and we concatenate them along the sequence dimension before passing to an MLP or something, we would actually need to know the number of hidden states to infer the MLP dim. I am not sure if this is something @pbontrager is trying to solve for, but if so we would probably need to do it this way.

SalmanMohammadi: ¿por qué no los dos¿

Couldn't we override in the forwad pass like we do `self.causal mask` and `mask`? Preferring `output_hidden_states: Optional[...] = None` over `self._output_hidden_states`?

joecummings: + 1 to the name being `output_hidden_states` and consideration for having this be defined on the main module, but possible to override in the forward pass if needed. 

The latter could be overkill for this refactor and I would settle for a code comment mentioning that this is possible to enable if necessary.

RFC.md
joecummings: Should we save a torchtune version or hash in each model config?

RFC.md
```diff
@@ -0,0 +1,413 @@
...
+        out = h + self.mlp_scale(mlp_out)
+        return out
+```
+
+This is a reasonable alternative but I do fear it'll get more and more complex to extend as we go forward and want to support more attention types.
+
+### Risks/Concerns
+- Attention module and model builders both get a bit more complex with this. The more variation we support the more complexity we introduce. I believe that supporting multiple transformer layers at least allows us to keep individual layers simpler.
+- This change breaks BC both in the checkpoint itself but also in model builders. We can update all of our model builders, but if users have their own custom ones they will have to update them too. Do we have a mechanism for clearly communicating this in the release notes along with how to update a model builder?
+- The cross attention layer here does not work with sample packing. Should we make layers that support sample packing through an error if they're used with packed data? @RdoubleA
```
joecummings, line 412: So to be clear, no sample packing will be enable for models that utilize cross attention?

pbontrager: Not until sample packing is updated to update cross attention masks.

RFC.md
pbontrager: I can move that into **kwargs

RFC.md
pbontrager: Could you clarify more here? The flexibility of the attention blocks it can be used with? We could remove the check all together but thought it was better to have it until we didn't need it anymore.

RFC.md
pbontrager: We can't get rid of it. flex attention can remove the need for applying it to the attention output but we still need it then for the feedforward.

RFC.md
pbontrager: I don't know why this would ever be dynamically changing in the forward pass. I'd rather just expose it in the builder function if we needed too.  I'll use `output_hidden_states`.

Do people prefer the output `return output, hidden` or 
```
output = output if not hidden else hidden + [output]
return output
```
?

RdoubleA:
done by #1312 and #1139

PR #1227, opened 2024-07-25
Unify configs for Phi3

joecummings:
#### Context

I couldn't take it anymore. Every one of our configs looks different, has different names for things, has different params, and use different datasets as examples. Some of this makes sense and is related to the specific use-case. The majority though is just our fast moving codebase and increases the mental strain on the user to find something in a config. 

At last I can rest.

#### Changelog
* Impose an order to the elements in the config: model, tokenizer, checkpointer, dataset, fine-tuning arguments, training env, memory management, logging, profiler (optional)
* Move epoch related things and sim batch size related things close to each other
* Add `log_peak_memory_stats` to every config
* Remove max_seq_len from the configs that had it - no need in this example
* Set output-dir to the common tmp name we've been using.

#### Test plan

1. EYES and that this makes sense
2. I will be running each of these configs for 10 steps and confirming everything looks right 

#### Next steps

Once we agree on everything here, I will either do the rest of the configs myself or ask the community for help.

### DIFF
--- a/recipes/configs/phi3/mini_full.yaml
+++ b/recipes/configs/phi3/mini_full.yaml
@@ -17,22 +17,16 @@
 # Single device full finetuning requires more memory optimizations. It's
 # best to use mini_low_memory.yaml for those cases
 
+# Model arguments
+model:
+  _component_: torchtune.models.phi3.phi3_mini
+
 # Tokenizer
 tokenizer:
   _component_: torchtune.models.phi3.phi3_mini_tokenizer
   path: /tmp/Phi-3-mini-4k-instruct/tokenizer.model
 
-# Dataset
-dataset:
-  _component_: torchtune.datasets.alpaca_cleaned_dataset
-  max_seq_len: 4096
-seed: null
-shuffle: True
-
-# Model Arguments
-model:
-  _component_: torchtune.models.phi3.phi3_mini
-
+# Checkpointer
 checkpointer:
   _component_: torchtune.utils.FullModelHFCheckpointer
   checkpoint_dir: /tmp/Phi-3-mini-4k-instruct
@@ -45,16 +39,22 @@ checkpointer:
   model_type: PHI3_MINI
 resume_from_checkpoint: False
 
+# Dataset
+dataset:
+  _component_: torchtune.datasets.alpaca_cleaned_dataset
+seed: null
+shuffle: True
+
 # Fine-tuning arguments
-batch_size: 2
 epochs: 1
+max_steps_per_epoch: null
+batch_size: 2
+gradient_accumulation_steps: 16
 optimizer:
   _component_: torch.optim.AdamW
   lr: 5e-6
 loss:
   _component_: torch.nn.CrossEntropyLoss
-max_steps_per_epoch: null
-gradient_accumulation_steps: 16
 
 # Training env
 device: cuda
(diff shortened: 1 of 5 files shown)


recipes/configs/phi3/mini_qlora_single_device.yaml
```diff
@@ -40,51 +42,49 @@ checkpointer:
...
-
-# Show case the usage of pytorch profiler
+# Showcase the usage of PyTorch profiler
 # Set enabled to False as it's only needed for debugging training
 profiler:
   _component_: torchtune.utils.setup_torch_profiler
   enabled: False
 
-  #Output directory of trace artifacts
+  # Output directory of trace artifacts
```
RdoubleA, line 87: you removed output_dir from this config, so need to write out the full directory path

recipes/configs/phi3/mini_qlora_single_device.yaml
```diff
@@ -40,51 +42,49 @@ checkpointer:
...
+# Memory management
+enable_activation_checkpointing: True
+dtype: bf16
+
 # Logging
-output_dir: /tmp/qlora_finetune_output/
 metric_logger:
   _component_: torchtune.utils.metric_logging.DiskLogger
-  log_dir: ${output_dir}
+  log_dir: /tmp/Phi-3-mini-4k-instruct
```
RdoubleA:
```suggestion
  log_dir: /tmp/Phi-3-mini-4k-instruct/logs
```

RdoubleA (approved):
This appeases my inner OCD, can you verify these configs can still run? (maybe you can use `tune validate` ;))

joecummings:
> This appeases my inner OCD, can you verify these configs can still run? (maybe you can use `tune validate` ;))

Yep! I'll run all of these before merging.

recipes/configs/phi3/mini_full.yaml
```diff
@@ -45,30 +39,34 @@ checkpointer:
   model_type: PHI3_MINI
 resume_from_checkpoint: False
 
+# Dataset
+dataset:
+  _component_: torchtune.datasets.alpaca_cleaned_dataset
```
joecummings, line 44: Should we default to a chat dataset since our base models are already instruct tuned?

recipes/configs/phi3/mini_qlora_single_device.yaml
```diff
@@ -40,52 +42,50 @@ checkpointer:
...
 
-# Environment
-device: cuda
-dtype: bf16
-enable_activation_checkpointing: True
-
-# Show case the usage of pytorch profiler
+# Showcase the usage of PyTorch profiler
 # Set enabled to False as it's only needed for debugging training
 profiler:
```
joecummings, line 83: I would love an explicit rule-of-thumb for when we include profiler by default. 

Anyone? @kartikayk @ebsmothers

recipes/configs/phi3/mini_qlora_single_device.yaml
```diff
@@ -40,52 +42,50 @@ checkpointer:
...
   _component_: torchtune.datasets.alpaca_cleaned_dataset
 seed: null
 shuffle: True
-batch_size: 2
 
-# Optimizer and Scheduler
+# Fine-tuning arguments
+epochs: 1
+max_steps_per_epoch: null
+batch_size: 2
```
joecummings, line 54: bsz + grad acc + default gpus called out in example command above should be 128 if possible. 

source: me, anecdotally. Fight me

SalmanMohammadi:
We love to see this. Could we add a blank copy of this format to the repo as a source-of-truth template? In `contributing/`?

joecummings:
> We love to see this. Could we add a blank copy of this format to the repo somewhere as a source-of-truth template?

I think we do need source-of-truth somewhere, but each recipe actually potentially has different possible config defaults therefore I would want those to be documented in your recipe docs refactor actually. 

So each recipe page would be 1) a description of what the recipe does and how it works best  2) all the knobs/keys you can mess with OOTB and 3) a list of all the default configs that pair with it.

PR #1235, opened 2024-07-27
[memory Improvement] delete logits before bwd

felipemello1:
#### Context
- [ ] add a new feature
- [ ] fix a bug
- [x] other (please add here)

Inspired by https://github.com/pytorch/torchtune/pull/1046

Releasing logits before backward releases memory and reduces peak allocated memory

image

#### Changelog
delete the logits before backward for lora/FFT/QAT recipes. I didn't do it for RL, since there is a bit more complexity there. Lora distributed already had it.

#### Test plan
ran it for 5 epochs with/without the change. Same loss and tok/s, but lower memory

### DIFF
--- a/recipes/full_finetune_distributed.py
+++ b/recipes/full_finetune_distributed.py
@@ -523,6 +523,8 @@ class FullFinetuneRecipeDistributed(FTRecipeInterface):
                 logits = logits.transpose(1, 2)
                 # Compute loss
                 loss = self._loss_fn(logits, labels)
+                # free logits otherwise it peaks backward memory
+                del logits
 
                 loss = loss / self._gradient_accumulation_steps
                 running_loss += loss
--- a/recipes/full_finetune_single_device.py
+++ b/recipes/full_finetune_single_device.py
@@ -452,6 +452,8 @@ class FullFinetuneRecipeSingleDevice(FTRecipeInterface):
                 logits = logits.transpose(1, 2)
                 # Compute loss
                 loss = self._loss_fn(logits, labels)
+                # free logits otherwise it peaks backward memory
+                del logits
 
                 loss = loss / self._gradient_accumulation_steps
                 running_loss += loss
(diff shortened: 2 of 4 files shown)


ebsmothers (commented):
Nice find! I think this should be a free win. Looks like some CI jobs are failing but otherwise no concerns from me

joecummings (approved):

musabgultekin:
Hi @felipemello1 I saw that you added "nproc8" config to the table. Could you elaborate what is that. Im asking as I see it has a lower memory footprint on FFT

felipemello1:
> Hi @felipemello1 I saw that you added "nproc8" config to the table. Could you elaborate what is that. Is that FSDP with 8x GPUs?. Im asking as I see it has a lower memory footprint on FFT

@musabgultekin you are correct, 8xA100 with FSDP

felipemello1:
@SalmanMohammadi since you are touching some RL recipes, if you have a chance to test it there too and you dont mind, please let us know.

SalmanMohammadi:
Hey @felipemello1. Thanks for your work here - a very neat change.
The main blocker in RL recipes is that the logits are also used for metric logging as a way to measure the extent of divergence for the model being trained, or to measure how well differentiated logits/logprobs for "preferred" vs "non-preferred" are. 

I can take a look to see if there's a neat way around this. As an aside, how did the improvements you saw in memory usage scale with batch size/max seq len?

felipemello1:
> The main blocker in RL recipes is that the logits are also used for metric logging

I think that doing .detach().cpu() should do the trick? But then it may make the code a bit ugly/weird. Before going down the rabit hole of how to organize the code, I guess we could just test it, ignoring the metrics logging, and see if it impacts memory?

> how did the improvements you saw in memory usage scale with batch size/max seq len?

13GB less for 24k sequence len with LoRA, bsz=1, but i dont have a graph showing the % diff for multiple seq_len and bsz :/

image

SalmanMohammadi:
Thanks so much for pointing me towards this. I made a hopefully minimal change to maintain the logging behaviour whilst deleting the unnecessary logits.

This only affects DPO, since RLHF is still in the works, but I've done my best to be ruthless with freeing memory wherever I can when implementing RLHF. If you're interested I wouldn't say no to a review in the relevant functions : )

PR #1236, opened 2024-07-27
[memory improvement] pre-allocate memory before training starts

felipemello1:
#### Context
- [ ] add a new feature
- [ ] fix a bug
- [x] other (please add here)

As suggested in [pytorch tuning guide](https://pytorch.org/tutorials/recipes/recipes/tuning_guide.html#preallocate-memory-in-case-of-variable-input-length): 

> "reallocate memory in case of variable input length
> Models for speech recognition or for NLP are often trained on input tensors with variable sequence length. Variable length can be problematic for PyTorch caching allocator and can lead to reduced performance or to unexpected out-of-memory errors. If a batch with a short sequence length is followed by an another batch with longer sequence length, then PyTorch is forced to release intermediate buffers from previous iteration and to re-allocate new buffers. This process is time consuming and causes fragmentation in the caching allocator which may result in out-of-memory errors.
> 
> A typical solution is to implement preallocation. It consists of the following steps:
> 
> - generate a (usually random) batch of inputs with maximum sequence length (either corresponding to max length in the training dataset or to some predefined threshold)
> - execute a forward and a backward pass with the generated batch, do not execute an optimizer or a learning rate scheduler, this step preallocates buffers of maximum size, which can be reused in subsequent training iterations
> - zero out gradients
> - proceed to regular training"

It greatly decreases max reserved memory, and if the batch or max_seq is too large, it should raise OOM before training starts, instead of during training.

To reproduce: https://gist.github.com/felipemello1/17871d69435a2f6f22fa333f31dfe020

#### Changelog
- Added memory pre-allocation to Full FT, LoRA and QAT scripts. It is the same function for all scripts. I wonder if this should be some utils function.
- Changed `self._optimizer.zero_grad()` to `self._optimizer.zero_grad(set_to_none=True)`

notes:
- IMPORTANT: I noticed that in the distributed frameworks, memory consumption still increases after batch one (images below). I think its worth investigating if some memory is not being released, or if thats just a property of distributed training. @ebsmothers 
- It requires for max_seq_len to be defined in the config. One additional thing we could do is check if dataset == packed, and if so, use the model hardcoded max_seq_len. But I am not sure if its worth the trouble.
- RL scripts are a bit more complex and would need to be addressed in another PR

#### Test plan
Ran it before and after the change
image
image
image
image
image
image

### DIFF
--- a/recipes/full_finetune_distributed.py
+++ b/recipes/full_finetune_distributed.py
@@ -135,6 +135,10 @@ class FullFinetuneRecipeDistributed(FTRecipeInterface):
         self.max_steps_per_epoch = cfg.max_steps_per_epoch
         self.global_step = 0
 
+        # Arguments to pre allocte memory
+        self._batch_size = cfg.batch_size
+        self._max_seq_len = cfg.dataset.max_seq_len
+
     def load_checkpoint(self, cfg_checkpointer: DictConfig) -> Dict[str, Any]:
         """
         Extract the checkpoint state from file and validate. If resume_from_checkpoint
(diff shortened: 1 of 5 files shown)


felipemello1:
IMPORTANT: I noticed that in the distributed frameworks, memory consumption still increases after batch one (images below). I think its worth investigating if some memory is not being released, or if thats just a property of distributed training. @ebsmothers

felipemello1:
Do you guys know why adding "utils.cleanup_before_training()" to full_finetune_single_device raises this error: https://github.com/pytorch/torchtune/actions/runs/10126372565/job/28002955028?pr=1236

ebsmothers (commented):
This is an interesting technique. Tbh I'm not sure about enabling this for a couple of reasons:

(1) I don't understand the value in this greatly decreasing maximum reserved memory. Iiuc, PyTorch will greedily reserve extra memory sometimes, but if there is actual memory pressure approaching an OOM then unused cache memory will be freed. In that case it should be allocated memory that we care about for OOMs, not reserved memory. The bit about preventing "unexpected OOMs" is compelling, but I thought that PyTorch can request CUDA to do some reorganizing via `cudaFree` for some perf hit. I do wonder if there are still cases where pre-allocation actually prevents an OOM though. On this note, would also recommend giving [this blog](https://zdevito.github.io/2022/08/04/cuda-caching-allocator.html) a read if you haven't already.
 
(2) There are cases where max seq len is quite high but the actual sequence length of samples in the dataset is much smaller. In that case applying this could cause an OOM up front that we otherwise might not see (I know that we can simply disable this feature, but I think it can cause unnecessary confusion).

Anyways this is not to sound too pessimistic cause I can imagine this helping performance if we don't have to reallocate a bunch of blocks due to a small sequence length in the first batch. But I wanna make sure we think carefully before enabling such a feature.

felipemello1:
Closed it by mistake when cleaning up some branches, but I agree with what you said. Knowing that is OOMs at the start is interesting, but it adds more code to the recipe, requires max_seq_len to be defined and doesnt seem to add benefits. I tried running some code that OOMs unless i do `os.environ["PYTORCH_CUDA_ALLOC_CONF"] = "expandable_segments:True"`, and I was expecting that preallocation would solve it, but it didnt.

I am fine with just keeping it closed. Thanks for the comment! @ebsmothers

PR #1238, opened 2024-07-28
Remove unused variables

stsouko:
Unused shape unpacking blocks customization of input and embedding layer.

### DIFF
--- a/torchtune/modules/transformer.py
+++ b/torchtune/modules/transformer.py
@@ -220,9 +220,6 @@ class TransformerDecoder(nn.Module):
             - d: embed dim
             - m_s: max seq len
         """
-        # input tensor of shape [b, s]
-        bsz, seq_len = tokens.shape
-
         # shape: [b, s, d]
         h = self.tok_embeddings(tokens)
 


ebsmothers:
Hi @stsouko thanks for creating the PR. The change looks fine to me since, as you point out, neither `bsz` nor `seq_len` are actually used later in the class. Out of curiosity, what sort of customization are you looking to do here? I'd be interested to know what shape your tokens are and what you're using as an embedding layer (assuming you wouldn't mind sharing this). I assume the output of your embeddings will still be shape `[b, s, d]`, but it'd be helpful to ensure we're supporting these more custom cases properly.

Also please make sure to sign the CLA if you haven't already.

stsouko:
Hi @ebsmothers,
I would like to modify embedding to implement Llava like multimodal adaptation, thus, input can be a custom data structure which contains minibatch of text token ids and vectors of different modality.

ebsmothers:
Thanks for the info! Kicked off a CI run, once everything is green I think this is good to merge. Also cc @pbontrager who has been looking at our multimodality support

ebsmothers (approved):

PR #1242, opened 2024-07-29
Enable export for CLIPImageTransform

lucylq:
#### Context
- [x] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

#### Changelog

**Refactor CLIPImageTransform so that it can be exported, and shared with other repos (eg. ExecuTorch). This includes**:

`torchtune/models/clip/_transforms.py`
- Create CLIPImageTransformCore, an nn.Module containing resize, pad, reshape logic. ExecuTorch will export this nn.Module.
- Replace F.pad with Pad, which is an nn.Module. This allows ExecuTorch to swap it for a custom op.
- Move logic in resize_with_pad.py to the top level in CLIPImageTransform, pass the information to the nn.Module.
- Add arg for antialias to make it configurable on ET side.

`torchtune/modules/transforms/vision_utils/tile_crop.py`
- Move this into an nn.Module. This allows ExecuTorch to swap it for a custom op.

Update tests
- tests/torchtune/models/clip/test_clip_image_transform.py
- tests/torchtune/modules/transforms/test_tile_crop.py

Update code references
- torchtune/modules/transforms/__init__.py
- torchtune/modules/vision_transformer.py
- docs/source/api_ref_modules.rst

- I think we can remove resize_with_pad.py with associated tests after this change.

#### Test plan
Run pytest in directories:
- torchtune/tests/torchtune/models/clip/
- torchtune/tests/torchtune/modules/transforms/

- [x] add unit tests for any new functionality (used existing tests)
- Error: seems unrelated to this PR (also see this error on main, without my changes)
	- ``` E   AttributeError: `np.string_` was removed in the NumPy 2.0 release. Use `np.bytes_` instead.```
	- Full error log: P1503938971
- [na] manually run any new or modified recipes with sufficient proof of correctness


torchtune/models/clip/_transforms.py
```diff
@@ -5,24 +5,131 @@
...
+        torch._check(target_h > 1)
+        torch._check(target_h <= self.config.max_image_size)
+        torch._check(target_w > 1)
+        torch._check(target_w <= self.config.max_image_size)
+
+        image = resize(
+            image,
+            size=[target_h, target_w],
+            interpolation=self.config.resample,
+            antialias=True,
```
lucylq, line 104: Note, to lower to ExecuTorch we have to set this to False.

torchtune/models/clip/_transforms.py
```diff
@@ -5,24 +5,111 @@
...
+
+        # Checks to allow the model to export via torch.export.
+        self.check_variable_bounds_for_export(target_h, 2, self.max_image_size)
+        self.check_variable_bounds_for_export(target_w, 2, self.max_image_size)
+
+        image = resize(
+            image,
+            size=[target_h, target_w],
+            interpolation=self.resample,
+            antialias=self.antialias,
```
lucylq, line 87: Add as an argument, so we can configure it on ExecuTorch side as well.

torchtune/models/clip/_transforms.py
```diff
@@ -5,24 +5,112 @@
...
 )
+from torchvision.transforms import v2
+from torchvision.transforms._functional_tensor import resize
 
 from torchvision.transforms.v2 import functional as F
 
 logger = logging.getLogger(__name__)
 
 
+class CLIPImageTransformCore(torch.nn.Module):
```
ebsmothers, line 27: Just to confirm after our discussion today, the idea is to fully make `CLIPImageTransform` exportable, but we are not there yet, right? If so, do we intend to land this as an intermediate version first? The reason I ask is because with this PR some of our utilities like `resize_with_pad` will no longer have any callsites in the repo, and I want to know whether this is expected to be the final state or just a temporary one.

torchtune/models/clip/_transforms.py
```diff
@@ -5,24 +5,112 @@
...
+            size=[target_h, target_w],
+            interpolation=self.resample,
+            antialias=self.antialias,
+        )
+
+        # Pad.
+        canvas_h, canvas_w = canvas_size.tolist()
+        # Checks to allow the model to export via torch.export.
+        self.check_variable_bounds_for_export(canvas_h, 2, self.max_image_size)
+        self.check_variable_bounds_for_export(canvas_w, 2, self.max_image_size)
```
ebsmothers, line 95: Can we just consolidate these checks together with target_h and target_w up front? Something like

```
def check_variable_bounds_for_export(self, vars: List[int]):
	for var in vars:
	        torch._check(var >= 2)
	        torch._check(var <= self.max_image_size)
```

Then we just call this as a one-liner at the beginning of the function. (Can even consider

ebsmothers (commented):
Mostly small comments, will leave it for @felipemello1 to confirm details of the actual transforms

torchtune/models/clip/_transforms.py
felipemello1: I sent a message to someone in GenAI to see what they think

torchtune/models/clip/_transforms.py
lucylq: Yeah - ideally we can fully export `CLIPImageTransform` but right now that would require more support on export side. So I think we can remove `resize_with_pad` and its tests.

torchtune/models/clip/_transforms.py
```diff
@@ -5,24 +5,112 @@
...
+        self.max_image_size = max_image_size
+        self.antialias = antialias
+
+    """
+    Performs torch._checks to confirm a value is within the specified lower and upper bounds.
+    Note: this is used to export the model. For eager mode usage, please disregard.
+
+    The check mitigates data dependent errors that may occur during torch.export. It installs
+    a deferred runtime assert, instead of a compile-time guard. For more context:
+    https://docs.google.com/document/d/1HSuTTVvYH1pTew89Rtpeu84Ht3nQEFTYhAX3Ypa_xJs/edit
```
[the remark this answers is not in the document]
lucylq: Yes, the doc is publicly available, though we can also give a tl;dr if that's preferred? Something like:

Data dependent errors usually occur in models with data-dependent control flow, eg. via .item(), tolist(), nonzero().

torchtune/models/clip/_transforms.py
```diff
@@ -5,24 +5,112 @@
...
+
+        padding = [0, canvas_h - target_h, 0, canvas_w - target_w]
+        output = v2.Pad(padding=padding)(image)
+
+        # Normalize.
+        if self.image_mean is not None and self.image_std is not None:
+            output = v2.functional.normalize(output, self.image_mean, self.image_std)
+
+        # Reshape.
+        tiles = TileCrop()(output, self.tile_size)
```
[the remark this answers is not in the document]
lucylq: updated, thank you

torchtune/models/clip/_transforms.py
```diff
@@ -151,28 +244,46 @@ def __call__(self, *, image: Image.Image, **kwargs) -> Mapping[str, Any]:
...
+        # Calculate the largest aspect ratio preserving size that fits best_resolution.
+        scale_h = target_size[0] / image_height
+        scale_w = target_size[1] / image_width
 
-        aspect_ratio = torch.tensor(best_resolution).reshape(-1) // self.tile_size
+        new_target_height = min(math.floor(image_height * scale_w), target_size[0])
+        new_target_width = min(math.floor(image_width * scale_h), target_size[1])
+
+        # Call ClipImageTransformCore to perform resize, pad, normalize and reshape transforms.
+        core_transform = CLIPImageTransformCore(
```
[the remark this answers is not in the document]
lucylq: updated, thank you

torchtune/modules/transforms/vision_utils/tile_crop.py
```diff
@@ -24,36 +27,37 @@ def tile_crop(image: torch.Tensor, tile_size: int) -> torch.Tensor:
 
     Examples:
         >>> image = torch.rand(3, 200, 300)
-        >>> tiles = tile_crop(image, tile_size=50)
+        >>> tiles = TileCrop(image, tile_size=50)
```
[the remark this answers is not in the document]
lucylq: updated, thank you

torchtune/models/clip/_transforms.py
lucylq: updated to consolidate all the checks together

torchtune/models/clip/_transforms.py
```diff
@@ -69,6 +158,7 @@ class CLIPImageTransform:
             resolution from possible_resolutions.
             If False, it will pick the resolution that minimizes downscaling, including no downscaling at all.
             In this case, the image will only be upscaled if it's size < tile_size. Default False.
+        max_image_size (int): Used to upper-bound the input image size for export.
```
felipemello1, line 161: For a regular user, i am not sure if they understand what "for export" means. Will users change this? If not, should it be a global constant in the script? Maybe if we could add a bit more context, it would solve it

torchtune/models/clip/_transforms.py
```diff
@@ -5,24 +5,113 @@
...
+        """
+        Performs the core transformations involved in CLIPImageTransform;
+        1. Resize the image to target_size.
+        2. Pad the image to canvas_size.
+        3. Normalize the image using image_mean and image_std.
+        4. Reshape the image tensor into [n, channels, tile_size, tile_size].
+        Args:
+            image (torch.Tensor): image as a 3D tensor in form [C, H, W].
+            target_size (torch.Tensor): tensor of shape [1, 2] containing the target_height and target_width for resize.
+            canvas_size (torch.Tensor): tensor of shape [1, 2] containing the canvas_height and canvas_width for padding.
```
felipemello1, line 74: jsut wanted to confirm, is the shape here [1,2] or just (2,)? because we do target_size[0] and target_size[1] in another part of the code

torchtune/models/clip/_transforms.py
```diff
@@ -5,24 +5,113 @@
...
+
+        # Resize.
+        image = resize(
+            image,
+            size=[target_h, target_w],
+            interpolation=self.resample,
+            antialias=self.antialias,
+        )
+
+        # Pad.
```
felipemello1, line 97: nit: can you add a small comment here clarifying that image will be on top-left and is being padded on the right/bottom? (feel free to word it as you wish). It may sound simple, but when comparing implementations, i was confused about why some implementations padded with image on the center, and some on top-left.

torchtune/modules/transforms/vision_utils/tile_crop.py
```diff
@@ -11,49 +11,53 @@
 logger = logging.getLogger(__name__)
```
felipemello1, line 11: not part of your refactor, but i dont think we are using it. Do you mind deleting?

torchtune/models/clip/_transforms.py
```diff
@@ -130,11 +221,23 @@ def __init__(
 
         # resize_with_pad
```
felipemello1, line 222: nit: maybe change it to just "resize"?

torchtune/models/clip/_transforms.py
```diff
@@ -130,11 +221,23 @@ def __init__(
...
         self.max_upscaling_size = None if resize_to_max_canvas else tile_size
-        self.resample = torchvision.transforms.InterpolationMode[resample.upper()]
+        self.resample = resample
 
         # tile_crop
         self.tile_size = tile_size
 
+        # image size upper bound for export
```
felipemello1, line 229: same comment here about export. Maybe just doing torch.export makes it a bit more intuitive for the user, who has no clue that exporting is a thing

felipemello1 (commented):
LGTM, the refactor simplified some parts by removing resize_with_pad. Do you mind updating the unit tests (tilecrop, if there is any) and deleting the unused code (e.g. resize_with_pad)?

You would also need to update the function here: https://github.com/pytorch/torchtune/blob/main/docs/source/api_ref_modules.rst#vision-transforms

torchtune/models/clip/_transforms.py
```diff
@@ -5,24 +5,112 @@
...
+        canvas_size (torch.Tensor): tensor of shape [1, 2] containing the canvas_height and canvas_width for padding.
+
+    Returns:
+        Tuple[torch.Tensor, torch.Tensor]: Image tensor of shape [n, channels, tile_size, tile_size]
+            and aspect ratio tensor of shape [1, 2].
+    """
+
+    def forward(
+        self, image: torch.Tensor, target_size: torch.Tensor, canvas_size: torch.Tensor
+    ):
```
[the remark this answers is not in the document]
lucylq: thank you!

torchtune/models/clip/_transforms.py
```diff
@@ -5,24 +5,116 @@
...
+            image,
+            size=[target_h, target_w],
+            interpolation=self.resample,
+            antialias=self.antialias,
+        )
+
+        # Pad, such that the image is on the top-left and padded on the right-bottom.
+        sizes = [3, canvas_h, canvas_w]
+        padding = [0, canvas_h - target_h, 0, canvas_w - target_w]
+        output = v2.Pad(padding=padding)(image)
```
felipemello1, line 103: should we use a functional here, or add pad to __init__, so it follows the same pattern as the other transforms?

lucylq: Oh hmn, I'm not sure we can place it into __init__, because we initialize the nn.Module with potentially different padding for each call

The reason we're using an nn.Module instead of the functional version is so we can do a module swap with our custom op (have some issues exporting pad).

torchtune/models/clip/_transforms.py
```diff
@@ -5,24 +5,116 @@
...
+        # Resize.
+        image = resize(
+            image,
+            size=[target_h, target_w],
+            interpolation=self.resample,
+            antialias=self.antialias,
+        )
+
+        # Pad, such that the image is on the top-left and padded on the right-bottom.
+        sizes = [3, canvas_h, canvas_w]
```
felipemello1, line 101: is "sizes" used somewhere? I think we can delete it

lucylq: thanks - updated

felipemello1:
hey @lucylq , out of curiosity, can you remind me why we had to refactor resize_with_pad?

lucylq:
> hey @lucylq , out of curiosity, can you remind me why we had to refactor resize_with_pad? In other words, could it have been just a nn.module?

The main export issues are coming from the calculations we do to resize, pad and tile_crop the image tensor. This is because we're performing a bunch of calculations on values (rather than tensors), and using those values to transform the image tensor. I think export currently can't handle cases when we extract values from tensors / use those values to create/transform other tensors. cc @angelayi 

So I tried to place all the calculations outside of the nn.Module, and for cases where we can't (eg. tile_crop, and the impl of pad), made them into a custom op to hide the data dependent places / prevent export from tracing in.

We can make resize_with_pad into an nn.Module, but prefer not to because that means we have to implement resize and pad in C++ on our end. With the refactor, we implement pad, which is a lot simpler. Though if antialias=True then we may need resize anyway.

felipemello1:
> Though if antialias=True then we may need resize anyway.

I asked a researcher, but no answers yet. @pbontrager think it probably is. To unblock testing, maybe its ok to go like this for now, and come back to "antialias=True" when we have an answer? I just dont know if this is productive.

felipemello1:
btw, we would have to delete this unit test too: https://github.com/pytorch/torchtune/blob/main/tests/torchtune/modules/transforms/test_resize_with_pad.py

lucylq:
> btw, we would have to delete this unit test too: https://github.com/pytorch/torchtune/blob/main/tests/torchtune/modules/transforms/test_resize_with_pad.py

oh - I think I removed it, though the docs build is still failing. It might be due to TileCrop? Repro'd the error locally though still can't pinpoint where the error is coming from.

felipemello1:
FYI: we realized that the unit tests are missing regression tests. I will add them tomorrow so we are 100% show about parity.

lucylq:
> 1. erase anything related to vision from the .rst and confirm that this is the problem.
> 2. re-add one by one to see which is the problematic

Thanks, that helped - TileCrop was missing a docstring

torchtune/modules/transforms/vision_utils/tile_crop.py
```diff
@@ -4,56 +4,58 @@
...
-    # Flatten the tiles
-    tiles = transposed.contiguous().view(
-        tiles_height * tiles_width, channel_size, tile_size, tile_size
-    )
-
-    return tiles
+    def __init__(self):
+        super().__init__()
+
+    def forward(self, image: torch.Tensor, tile_size: int) -> torch.Tensor:
```
pbontrager, line 18: If this is a module now instead of function. I think we can follow torchvision style Module transforms and move tile_size into __init__

lucylq:
Rebased and ran tests from https://github.com/pytorch/torchtune/pull/1253

lucylq:
working on https://github.com/pytorch/torchtune/pull/1269 which refactors exported version into a separate file, after discussion with torchtune team.

PR #1255, opened 2024-08-02
Move setup_cache and reset_cache to the layer and attention module

pbontrager:
#### Context
- [x] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

In order to support kv_cache for heterogenous layer types [see here](https://github.com/pytorch/torchtune/pull/1224), we can't naively add kv_cache the same way inside of TransformerDecoder. This is currently done with

```python
for layer in self.layers:
            layer.attn.kv_cache = KVCache(
                batch_size=batch_size,
                max_seq_len=self.max_seq_len,
                num_heads=self.num_heads,
                head_dim=self.head_dim,
                dtype=dtype,
            )
```

which both assumes that the kv_cache is the same for every layer and that every layer has the format `layer.attn.kv_cache`. Neither of these assumptions hold true for MM models. For CrossAttentionLayers, the `max_seq_len=encoder_max_seq_len`. Also, more generally, having kv_cache defined in the attention module and layer gives much more flexibility for future layers. This is a problem when fusing two layers together (regular and adapter layer) and you need custom logic on how kv_cache is applied to the fused layer.

Related to this, input_pos, is not the same for every layer type during inference as you're dealing with multiple sequences; decoder sequence and encoder sequence. Since the input_pos information is already contained within kv_cache, this PR also automatically computes input_pos if not provided.

#### Changelog
Add create_cache, cache_enabled, and reset_cache to TransformerDecoderLayer. Also add create_cache and enable_cache to attention module. The kv_cache creation logic is within the attention module and TransformerDecoderLayer simply passes it through. TransformerDecoder then calls each layers create_cache/reset_cache.

In TransformerDecoderLayer, if kv_cache is set and input_pos is not provided, input_pos is created given the cache size.

#### Test plan


torchtune/modules/transformer.py
```diff
@@ -36,6 +36,23 @@ def __init__(
...
+    def setup_cache(self, batch_size: int, dtype: torch.dtype) -> None:
+        """Setup key value caches for attention calculation.
+
+        Args:
+            batch_size (int): batch size for the caches.
+            dtype (torch.dtype): dtype for the caches.
+        """
+        self.attn.setup_cache(batch_size, dtype)
+
+    def cache_enabled(self) -> bool:
```
RdoubleA, line 48: nit: make this a property?

torchtune/modules/attention.py
```diff
@@ -117,6 +117,37 @@ def __init__(
...
+        Args:
+            batch_size (int): batch size for the caches.
+            dtype (torch.dtype): dtype for the caches.
+
+        Raises:
+            RuntimeError: if called when kv_cache is already setup.
+        """
+        # Don't overwrite user defined kv_cache from init
+        if self.kv_cache is not None:
+            raise RuntimeError(
```
RdoubleA, line 132: should this throw an error or just raise a warning and skip setup if called twice?

RdoubleA (commented):
Keeping KVCache with the attention module makes a lot more sense to me. My main issue is that the setup_cache is being propagated up from attention to decoder layer and to decoder and there's nothing enforcing that each of these have the methods. Before we assumed attn had a kv_cache from TransformerDecoder and set it up directly, which was also odd, but if a user wants to swap out the TransformerDecoderLayer or the CausalSelfAttention, how do we make sure they know to add these methods?

torchtune/modules/transformer.py
```diff
@@ -36,6 +36,23 @@ def __init__(
         self.mlp_norm = mlp_norm
         self.mlp = mlp
 
+    def setup_cache(self, batch_size: int, dtype: torch.dtype) -> None:
```
ebsmothers, line 39: Maybe a controversial q but if we are gonna have multiple different types of layers that can be passed to `TransformerDecoder` should we define some kind of Protocol to ensure any individual layer has this method defined? (I don't have a strong opinion on this tbh, just putting it out there)

torchtune/modules/attention.py
```diff
@@ -161,6 +192,10 @@ def forward(
         # input has shape [b, s, d]
         bsz, seq_len, _ = x.shape
 
+        if self.kv_cache and input_pos is None:
+            cache_size = self.kv_cache.size
+            input_pos = torch.arange(cache_size, cache_size + seq_len, device=x.device)
```
ebsmothers, line 197: This part feels a bit weird to me. So we are relying on KV cache to infer the number of tokens seen by the positional embeddings thus far during inference? Why not just define a variable in positional embeddings for this? And just as a sanity check: we never have to worry about sample packing because this is an inference-only path, right?

ebsmothers (commented):
> This is a problem when fusing two layers together (regular and adapter layer) and you need custom logic on how kv_cache is applied to the fused layer.

I don't fully understand this point. Fusing here == stacking?

But overall this looks good to me

joecummings:
@pbontrager Does this need to land before #1283

torchtune/modules/transformer.py
pbontrager: In a early version of the TransformerDecoderLayer RFC I had a Protocol and generally got poor reception. But I'm open to adding it back in.

torchtune/modules/attention.py
pbontrager: We're relaying on kv_cache to track the input_pos during coding because, the amount cached and amount decoded are the same thing. For me the positional embeddings make less sense for this because they have nothing to do with input_pos, they are just converting positional information into embedding form.

torchtune/modules/transformer.py
ebsmothers: Lol oh yeah I remember that now. Let's leave it like this

joecummings (approved):
i WaNt to ApPRoVe ThiS pR tOO

Issue #1257, opened 2024-08-03
Pre training


etemiz:
Is there a way to do pre training using txt files?  I am seeing mostly fine tuning recipes. Thanks

SalmanMohammadi:
Hey @etemiz! Thanks for checking out torchtune : ))

We support continued pre-training through our [`text_completion_dataset`](https://pytorch.org/torchtune/main/tutorials/datasets.html#custom-unstructured-text-corpus).

The example I linked actually shows you how to specify this dataset for your own txt file. You can then use this dataset configuration for any of our SFT recipes: LoRA/full-finetune and single-device/distributed.

Let me know if this makes sense.

etemiz:
That worked well. But I am stuck again.

 File "/home/dead/ml/torchtune/torchtune/v/lib/python3.10/site-packages/torchtune/datasets/_text_completion.py", line 55, in _prepare_sample
    prompt = sample[self._column]
KeyError: None


Do I have to convert txt file to something of a json nature ?

{ "text": "unstructured text goes here" }

RdoubleA:
Hi @etemiz, assuming you just have a raw text file with no column names, can you try using the default value for column? From the docstring for `text_completion_dataset`:

> For local datasets with a single column, use the default "text", which is what is assigned by Hugging Face datasets when loaded into memory. Default is "text".

etemiz:
I didn't touch the default. My yaml:

```
dataset:
  _component_: torchtune.datasets.text_completion_dataset
  source: text
  data_files: /tmp/all.txt
  split: train
```

all.txt is a normal text file with long lines. Each line is a topic.

RdoubleA:
Hmm I see. Do you mind sharing the torchtune version you are on as well?

etemiz:
torchtune              0.2.1

SalmanMohammadi:
Could you try change your yaml to:

etemiz:
That worked. One warning and another error appeared.
Warning
```
torchao/dtypes/nf4tensor.py:870: UserWarning: Attempting to use hipBLASLt on an unsupported architecture! Overriding blas backend to hipblas (Triggered internally at ../aten/src/ATen/Context.cpp:288.)
```

Error
```
  File "/home/dead/ml/torchtune/torchtune/v/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1736, in _wrapped_call_impl
    return self._call_impl(*args, **kwargs)
[...]

Issue #1191, opened 2024-07-17
`wikitext_dataset` fails in model forward with "Expected Long, Int; but got torch.cuda.FloatTensor"


RdoubleA:
Thanks to @felipemello1 for pointing this out.

Repro: `tune run lora_finetune_single_device --config llama3/8B_lora_single_device max_seq_len=4096 dataset=torchtune.datasets.wikitext_dataset gradient_accumulation_steps=2 log_every_n_steps=1 log_peak_memory_stats=True max_steps_per_epoch=10 epochs=1`

Because the wikitext dataset (https://huggingface.co/datasets/Salesforce/wikitext) splits samples by line in an article, some of the rows will be empty, leading to entire samples with pad_id. Ideally, these rows should just be dropped.

But this leads to our bug - if the first sample is empty, `pad_sequence` in `padded_collate` will actually return a FloatTensor instead of keeping the ints (https://github.com/pytorch/torchtune/blob/main/torchtune/utils/collate.py#L46). You can repro this adding `pdb.set_trace()` above the linked code and continuing until you see a batch where the first sample is empty.

The fix is twofold:
- Enforce int tensors with `long()` in collate
- Use the document level version of wikitext: https://huggingface.co/datasets/EleutherAI/wikitext_document_level, which is a cleaner dataset

pbontrager:
Can you just modify wikitext_dataset itself to skip empty sequences? My concern with `long` is that it won't respect the int type used by the other tokens.

RdoubleA:
addressed in #1258

PR #1258, opened 2024-08-04
Better wikitext

RdoubleA:
#### Context
- [ ] add a new feature
- [x] fix a bug
- [ ] other (please add here)

Addresses #1257 and #1191.

#### Changelog
- Use wikitext document level dataset from Hugging Face, which is cleaner and uses whole articles as samples instead of lines - more sensible for text completion datasets
- Expose `split_across_pack` for text completion datasets
- Update text completion dataset builder to use `column="text"` as default
- Enforce long int tokens in batch collater

#### Test plan
`tune run --nproc_per_node 8 full_finetune_distributed --config llama3/8B_full dataset.max_seq_len=4096 dataset=torchtune.datasets.wikitext_dataset max_steps_per_epoch=200 epochs=1` runs successfully without previously seen errors


torchtune/datasets/_text_completion.py
```diff
@@ -89,15 +90,24 @@ def text_completion_dataset(
...
+        source (str): path to dataset repository on Hugging Face. For local datasets,
+            define source as the data file type (e.g. "json", "csv", "text") and pass
+            in the filepath in ``data_files``. See Hugging Face's ``load_dataset``
             (https://huggingface.co/docs/datasets/en/package_reference/loading_methods#datasets.load_dataset.path)
-        column (Optional[str]): name of column in the sample that contains the text data. This is typically required
-            for Hugging Face datasets or tabular data, but can be omitted for local datasets. Default is None.
+            for more details.
+        column (str): name of column in the sample that contains the text data. This is typically required
+            for Hugging Face datasets or tabular data. For local datasets with a single column, use the default "text",
+            which is what is assigned by Hugging Face datasets when loaded into memory. Default is "text".
```
SalmanMohammadi, line 100:
```suggestion
        column (str): name of column in the sample that contains the text data. This is typically required
            for Hugging Face datasets or tabular data. For local datasets with a single column (e.g. unstructured txt files), use the default "text" which is used by Hugging Face datasets when loaded into memory. Default is "text".
```

torchtune/datasets/_text_completion.py
```diff
@@ -75,10 +75,11 @@ def _prepare_sample(self, sample: Mapping[str, Any]) -> Dict[str, List[int]]:
...
     tokenizer: ModelTokenizer,
     source: str,
-    column: Optional[str] = None,
+    column: str = "text",
     max_seq_len: Optional[int] = None,
     add_eos: bool = True,
     packed: bool = False,
+    split_across_pack: bool = True,
     **load_dataset_kwargs: Dict[str, Any],
 ) -> TextCompletionDataset:
```
SalmanMohammadi, line 84:
```suggestion
) -> Union[TextCompletionDataset, PackedDataset]:
```

SalmanMohammadi: nit: to also update the docstring, please

torchtune/datasets/_wikitext.py
```diff
@@ -40,10 +52,10 @@ def wikitext_dataset(
         TextCompletionDataset: the configured TextCompletionDataset
```
SalmanMohammadi, line 52: nit: 
```suggestion
        Union[TextCompletionDataset, PackedDataset]: the configured  :class:`~torchtune.datasets.TextCompletionDataset` or :class:`~torchtune.datasets.PackedDataset` if ``packed=True``.
```

SalmanMohammadi (approved):
minor nits; thanks for this : )

torchtune/utils/collate.py
```diff
@@ -68,7 +68,7 @@ def padded_collate(
             (0, labels_seq_len - input_ids_seq_len),
             value=padding_idx,
         )
-    return {"tokens": input_ids, "labels": labels}
+    return {"tokens": input_ids.long(), "labels": labels.long()}
```
SalmanMohammadi, line 71: and this deals with the empty line issue?

felipemello1 (commented):
LGTM! I would just address Salman comments first. 

Just one question: I believe that exposing "split_across_pack" is only relevant to text_completion. But are there any others that would be worth updating too, for consistency?

torchtune/utils/collate.py
RdoubleA: yes, when the line is empty these tensors become float tensors. so even if the line is empty this will ensure that the model still trains.

I was thinking about whether we should add a warning if a sample is completely empty (== all padding). It will get entirely ignored in loss anyway but it may be useful for debugging.

RdoubleA:
> Just one question: I believe that exposing "split_across_pack" is only relevant to text_completion. But are there any others that would be worth updating too, for consistency?

let me take a look, I think the other datasets for SFT instead of CPT should keep split across pack as False, but I don't have any hard evidence suggesting that SFT should never split across pack

torchtune/datasets/_wikitext.py
```diff
@@ -4,46 +4,59 @@
...
-        subset (str): name of subset of data to use, see the `wikitext page <https://huggingface.co/datasets/wikitext#data-fields>`_
-            for available subsets.
+            for more details.
+        subset (str): name of subset of data to use, see the `wikitext page
+            <https://huggingface.co/datasets/EleutherAI/wikitext_document_level#data-instances>`_
+            for available subsets. Default is ``"wikitext-103-v1"``.
         max_seq_len (Optional[int]): Maximum number of tokens in the returned input and label token id lists.
             Default is None, disabling truncation. We recommend setting this to the highest you can fit in memory
             and is supported by the model. For example, llama2-7B supports up to 4096 for sequence length.
+        packed (bool): Whether or not to pack the dataset to ``max_seq_len`` prior to training. Default is False.
```
SalmanMohammadi, line 46: @RdoubleA spotted that this isn't used in wikitext, is it just to match the signature of text completion builder?

RdoubleA: oh this should be passed into `text_completion_dataset`. will update.

PR #1263, opened 2024-08-05
Add mora layer

leighgable:
#### Context
- [x ] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

Here's a link to the relevant [discussion](https://github.com/pytorch/torchtune/discussions/1043#discussioncomment-10116971)
#### Changelog

#### Test plan

### DIFF
--- a/torchtune/modules/peft/mora.py
+++ b/torchtune/modules/peft/mora.py
@@ -140,7 +140,7 @@ class MoRALinear(nn.Module, AdapterModule):
             self.cos, self.sin = self.precompute_freqs(r)
         rh_in_x = torch.cat((-in_x[..., r // 2:], in_x[..., :r // 2]), dim=-1)
         in_x = in_x * self.cos + rh_in_x * self.sin
-        out_x = self.lora_a(in_x)
+        out_x = self.lora_a(in_x.to(self.lora_a.weight.dtype))
         out_x = out_x.view(*x.shape[:-1], -1)[..., :out_f]
         if out_x.shape[-1] < out_f:
             repeat_time = out_f // out_x.shape[-1]
--- a/torchtune/modules/peft/mora.py
+++ b/torchtune/modules/peft/mora.py
@@ -97,7 +97,7 @@ class MoRALinear(nn.Module, AdapterModule):
         if self.use_bias:
             if self._quantize_base:
                 raise NotImplementedError(
-                    "Quantized LoRALinear does not support bias at the moment."
+                    "Quantized MoRALinear does not support bias at the moment."
                 )
             bias = linear.bias
         return weight, bias
@@ -158,12 +158,12 @@ class MoRALinear(nn.Module, AdapterModule):
             Tensor: output tensor with shape ``(..., out_dim)``
 
         """
+        if self.disabled:
+            return x
         if self._quantize_base:
             out = linear_nf4(input=x, weight=self.weight)
         else:
             out = F.linear(x, self.weight, self.bias)
-        if self.disabled:
-            return out
 
         # out = self.dropout(out)
 
--- a/torchtune/modules/peft/mora.py
+++ b/torchtune/modules/peft/mora.py
@@ -7,6 +7,7 @@ import math
 from typing import List
 
 import torch.nn.functional as F
+import torch
 
 from torch import nn, Tensor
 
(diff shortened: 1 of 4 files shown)


SalmanMohammadi:
Thanks so much for adding this @leighgable. This is really cool and we'd love to help you get this landed in torchtune.

At a high level, we like to ensure features which we support are thoroughly tested to meet our bar for correctness. You've made an excellent start here with some unit tests. At a high level though, we like to make sure features which we add are benchmarked against some reference implementation. For example, by running this technique with as close-to-identical parameters/configuration as possible against a reference implementation and ensuring that the loss curves look similar. 

To start off, could you please attach some evidence of correctness to this PR? For example, comparing against the repo in the original discussion (https://github.com/kongds/MoRA), or through HF. 

Secondly, memory saving techniques are a core value proposition of torchtune - we want to ensure hardware isn't a barrier for people who wish to finetune LLMs. Usually, with our memory saving techniques we perform some degree of benchmarking to provide users with an idea of which techniques they can use to optimise their fine-tuning for their hardware. This also serves as ensuring we can provide configurations out-of-the-box which are tested against common hardware setups (e.g. 2x4090 or 4x4090).

Of course, this kind of benchmarking can be prohibitive if you don't easily have access to this kind of resource, so cc @ebsmothers for thoughts on this when we get to that point. I also don't want to block you here for things we can do on our side as a follow-up, so we can agree on a reasonable level of testing here.

Please let me know if I can help further - we can take a closer look at your code once we've come to some consensus about the overall end-to-end corectness of the implementation.

SalmanMohammadi:
To add to my above comment, it might be worth first doing a much more granular test by running forward on your implementation and a reference implementation with the same input tensor and ensuring the outputs are the same. This might help keep you sane if any issues come up during end-to-end testing.

I think, similar to the implementation of DoRA in #1115, we need to think about how we expose this in our model builders and configs in a sane way, particularly as we scale to more PEFT approaches. Currently, it doesn't look like you've integrated this into one of our models @leighgable - this will be a necessary step, both for verifying correctness, and to ensure we provide similar support for this feature as we do our other PEFT offerings.

cc @ebsmothers for more thoughts here since you've been working on DoRA.

torchtune/modules/peft/mora.py
```diff
@@ -0,0 +1,194 @@
...
+            self.cos, self.sin = self.precompute_freqs(r)
+        rh_in_x = torch.cat((-in_x[..., r // 2 :], in_x[..., : r // 2]), dim=-1)
+        in_x = in_x * self.cos + rh_in_x * self.sin
+        out_x = self.lora_a(in_x.to(self.lora_a.weight.dtype))
+        out_x = out_x.view(*x.shape[:-1], -1)[..., :out_f]
+        if out_x.shape[-1] < out_f:
+            repeat_time = out_f // out_x.shape[-1]
+            if out_f % out_x.shape[-1] != 0:
+                repeat_time += 1
+            out_x = torch.cat([out_x] * repeat_time, dim=-1)[..., :out_f]
```
ebsmothers, line 155: Can we clean this up a bit? Personally I find it a bit hard to parse what's happening here. Imo it's best to just write from scratch with shape comments and explicit variable names along the way

torchtune/modules/peft/mora.py
```diff
@@ -0,0 +1,194 @@
...
+        self.register_parameter("weight", nn.Parameter(weight))
+        self.register_parameter(
+            "bias", nn.Parameter(bias) if bias is not None else None
+        )
+        self.dropout = nn.Dropout(p=dropout)
+        self.lora_a = nn.Linear(
+            in_features=self.rank, out_features=self.rank, bias=False
+        )
+        # self.lora_b = nn.Linear(in_features=self.rank, out_features=self.rank, bias=False)
+        self.cos, self.sin = self.precompute_freqs(self.rank)
```
ebsmothers, line 75: Do these need to be stored as buffers or anything?

torchtune/modules/peft/mora.py
```diff
@@ -0,0 +1,194 @@
...
+from torch import nn, Tensor
+
+from torchao.dtypes.nf4tensor import linear_nf4, to_nf4
+from torchtune.modules.low_precision import _register_nf4_dispatch_ops  # noqa: F401
+from torchtune.modules.peft.peft_utils import AdapterModule
+
+
+class MoRALinear(nn.Module, AdapterModule):
+    """MoRA linear layer as introduced in `MoRA: High-Rank Updating for
+    Parameter-Efficient Fine-Tuning <https://arxiv.org/abs/2405.12130>`_.
```
ebsmothers, line 22: I remember we had talked about the 6 or so different implementations in kongds. Looks like here you're using the RoPE-inspired version. I thought a lot of the results in the paper use the group-based decomposition from equation 6, do you know if this version performs better? Mainly I want to make sure we are balancing providing the canonical version from the paper and (as much as possible) giving an implementation that's extensible to other cases.

ebsmothers (commented):
Thanks for the PR @leighgable! I agree with @SalmanMohammadi's point around testing -- did you set `EXPECTED_VAL` based on the kongds (or some other existing) implementation? Basically we should confirm that when we load identical weights into our MoRA layer and some reference implementation and pass some dummy tensor we get the same output. (Maybe you've done that already, if so great! If not I can help provide pointers here as needed.) 

Re integrating into an end-to-end workflow, I think both the `__init__` and `forward` signatures of this are the same as LoRA (though correct me if I'm mistaken here)? If so that would make the actual integration easy, you can follow an approach similar to what I did for DoRA [here](https://github.com/calvinpelletier/torchtune/blob/dora2/torchtune/models/llama3/_component_builders.py#L322). We should think about what this will look like and have a plan for some of the things that will inevitably come up (e.g. the state dict will now be different without the `lora_b` parameter, we need to make sure we can load and save weights appropriately). Again, happy to provide pointers on how to hook everything up so we can start training some models when the time comes.

I left some other inline comments as well, please let me know if anything's unclear.

PR #1240, opened 2024-07-29
Fixing pydoclint 

SalmanMohammadi:
#### Context
What is the purpose of this PR? Is it to
- [ ] add a new feature
- [x] fix a bug
- [x] update tests and/or documentation
- [ ] other (please add here)

Have you noticed that our docstring linter has been significantly less annoying lately? Perhaps you've observed this phenomenon for a while. Infact, [the keen eyed amongst you](https://github.com/pytorch/torchtune/pull/840#discussion_r1585341219) may even associate it with April 30th, 2024. The day I merged #840.

Indeed, my dear friends, on that fateful day, [I changed this single line](https://github.com/pytorch/torchtune/blame/2dc11d933387f04870586ed6da30413eef2617bf/pyproject.toml#L82). This effectively disabled our docstring linter hook.

I come to you to atone for my sins. I've updated the docstring linter config, which now correctly uses regex for `exclude`. I've also fixed docstrings this has now caught. 

#### Test plan

- [x] run pre-commit hooks and linters (make sure you've first installed via `pre-commit install`)
- [ ] add unit tests for any new functionality
- [x] update docstrings for any new or updated methods or classes
- [x] run unit tests via `pytest tests`
- [ ] run recipe tests via `pytest tests -m integration_test`
- [ ] manually run any new or modified recipes with sufficient proof of correctness
- [ ] include relevant commands and any other artifacts in this summary (pastes of loss curves, eval results, etc.)

PR #1168, opened 2024-07-12
Reverting Gemma checkpointing logic

SalmanMohammadi:
#### Context
What is the purpose of this PR? Is it to
- [ ] add a new feature
- [x] fix a bug
- [x] update tests and/or documentation
- [ ] other (please add here)

#1062 #1122  

#### Changelog

Reverting #1064, leaving Gemma checkpointing test in.

#### Test plan
Please make sure to do each of the following if applicable to your PR. (If you're not sure about any one of these just ask and we will happily help.)

- [x] run pre-commit hooks and linters (make sure you've first installed via `pre-commit install`)
- [x] add unit tests for any new functionality
- [ ] update docstrings for any new or updated methods or classes
- [x] run unit tests via `pytest tests`
- [ ] run recipe tests via `pytest tests -m integration_test`
- [ ] manually run any new or modified recipes with sufficient proof of correctness
- [ ] include relevant commands and any other artifacts in this summary (pastes of loss curves, eval results, etc.)

PR #1267, opened 2024-08-05
readme updates for PPO and Qwen2

ebsmothers:
Thanks to @SalmanMohammadi we now have a PPO recipe! We should show it off.

Also we should have added Qwen2 to our README already so doing that now too.

Also add acknowledgment on the readme to both @SalmanMohammadi for the PPO recipe and @fyabc for Qwen2 models.
(And fix a broken DPO hyperlink while I'm in here)


README.md
```diff
@@ -306,6 +306,8 @@ torchtune provides well-tested components with a high-bar on correctness. The li
 
 We really value our community and the contributions made by our wonderful users. We'll use this section to call out some of these contributions! If you'd like to help out as well, please see the [CONTRIBUTING](CONTRIBUTING.md) guide.
 
+- [@SalmanMohammadi](https://github.com/salmanmohammadi) for adding a comprehensive end-to-end recipe for [PPO](recipes/lora_dpo_single_device.py) training to torchtune
```
joecummings, line 309: We need a better way to handle this. These are definitely some of the best contributions we have, but we should highlight everyone who contributes to the repo. What about axolotl's acknowledgements: https://github.com/axolotl-ai-cloud/axolotl#contributing-?

ebsmothers: I don't intend for this to be an exhaustive list, but rather a more high-signal place to list the most valuable contributions we've got. The axolotl approach works well for them, but in our case it'll be 50% various PyTorch devs which imo defeats the purpose. I agree it's not an approach we should continue indefinitely, but this gives direct access to a GitHub user profile which I think is more valuable (on the Axolotl page it's basically just a JPEG with a hyperlink to the contributors page). We can consider keeping only the ~5 latest community contributions here so it doesn't become overwhelming though

felipemello1 (approved):

README.md
joecummings: Awesome, thanks for volunteering to keep this list updated at all times!

joecummings: That came off snarkier than I intended

joecummings: Or did it...

felipemello1: It may also introduces some weirdness/politics in who decides what is important enough.

joecummings: All contributors with the name Joe automatically go to the top of this list

joecummings: Anyone named Evan is permabanned

SalmanMohammadi: > It may also introduces some weirdness/politics in who decides what is important enough.

Yeah totally agreed here. IMO PPO is some of my worst work on the repo. Why can't we call out when I [broke the checkpointer for gemma](https://github.com/pytorch/torchtune/pull/1168) and then [fixed it for sweet sweet github stats](https://github.com/pytorch/torchtune/pull/1064), or [pYdOcLiNt](https://github.com/pytorch/torchtune/pull/1240)??

Maybe we can have a monthly vote on Discord for PRs to be added. It's the community contributions, why not put it in the community's hands?

joecummings: Another idea would be to feature all contributors in the CONTRIBUTING.md file and just do a better job highlighting our added features on a monthly basis, with an emphasis on community contributions.

This would be like an "Updates" section that holds a memory of 1-3 months or so.

SalmanMohammadi: > This would be like an "Updates" section on the README that holds a memory of 1-3 months or so.

https://github.com/pytorch/torchtune/pulls?q=merged%3A%3E2024-05-01+is%3Apr+sort%3Areactions+

PR #1271, opened 2024-08-06
[3/n] SFTDataset: refactor chat formats as prompt templates

RdoubleA:
#### Context
- [x] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

As discussed in the RFC in #1186, we will merged instruct and chat datasets to the following unified pipeline that can better support multimodal:
1. `message_transform` to create `List[Message]` from dataset with full flexibility on columns, ad-hoc modifications, etc. For multimodal, additionally images are loaded from the path
2. `prompt_template` as a optional way to add structured text around specific roles in the list of messages
3. `model_transform` that takes the list of messages and tokenizes it. For multimodal, it will additionally apply model-specific image transforms to the images associated with the sample

**For ease of review, we will stage this as multiple moderate-sized PRs**. This PR refactors the current `ChatFormat`s to the merged `PromptTemplate` classes. We keep the old ChatFormats to preserve BC for now

#### Test plan
- [ ] update live docs and check rendering
- [x] unit tests for new chat templates


felipemello1 (commented):
The changes make sense to me. I just worry about outdated docs. I searched for ChatFormat, and found some tutorials and mentions. Is there anything that we should delete/edit? E.g. should we keep a tutorial based on ChatFormat if we are deprecating it?

https://github.com/pytorch/torchtune/blob/5c7246e05b56729324625dee8a886727cfefeac0/docs/source/tutorials/datasets.rst#L318

RdoubleA:
> E.g. should we keep a tutorial based on ChatFormat if we are deprecating it?

Yeah great question. I was planning to keep all ChatFormat and Instruct/ChatDataset related tutorials until the refactor is entirely done, then I can update all the tutorials at the end. Let me know if there's any concerns with this

torchtune/data/_prompt_templates.py
```diff
@@ -120,6 +120,207 @@ def __call__(self, messages: List[Message]) -> List[Message]:
...
+                    content=content,
+                    masked=message.masked,
+                    ipython=message.ipython,
+                    eot=message.eot,
+                ),
+            )
+        return formatted_dialogue
+
+
+class MistralChatTemplate(PromptTemplateInterface):
```
joecummings, line 197: Crazy idea: What if we add these to the mistral/ dir and llama2/ dir? After all, these are model specific chat templates.

RdoubleA: That.... sounds pretty good actually. Let's do that

joecummings (approved):
LGTM, keep em coming

Issue #1185, opened 2024-07-16
expandable_segments with PYTORCH_CUDA_ALLOC_CONF reduces VRAM


winglian:
In [Unsloth](https://github.com/unslothai/unsloth/blob/main/unsloth/__init__.py#L53-L54):
```python
# Reduce VRAM usage by reducing fragmentation
os.environ["PYTORCH_CUDA_ALLOC_CONF"] = "expandable_segments:True"
```

using the defaults in torchtune:

`tune run lora_finetune_single_device --config llama3/8B_qlora_single_device`

- A100 80GB
- Measure @ 20steps 
- PYTORCH_CUDA_ALLOC_CONF="expandable_segments:True"

We should dig into why this provides such a large memory improvement and consider making it the default setting, or documenting it somewhere if not.

baseline: 16.39GiB @ 21.6sec/it
w/ compile: 16.63GiB @ 8.5sec/it
w/ expandable_segments: 10.831GiB @ 20.56sec/it
w/ compile + expandable_segments:  10.919GiB @ 8.5sec/it

felipemello1:
I wonder if we can squeeze more from it, and make it model + machine specific. E.g. in the first pass of the training, run a scrip that optimizes for it, since there are a few parameters associated with this flag

```
    os.environ["PYTORCH_CUDA_ALLOC_CONF"] = "backend:native,\
                                            max_split_size_mb:1024,\
                                            roundup_power2_divisions:8,\
                                            roundup_bypass_threshold_mb:256,\
                                            garbage_collection_threshold:0.8"
```

https://www.educative.io/answers/memory-management-using-pytorchcudaallocconf
https://dev.to/shittu_olumide_/how-can-i-set-maxsplitsizemb-to-avoid-fragmentation-in-pytorch-37h9

pbontrager:
Adding a comment to track this discussion in Pytorch core https://github.com/pytorch/pytorch/issues/130330

felipemello1:
I did some tests. It doesnt impact allocated memory, only reserved memory (cache). Pre-allocating memory before training starts has the same result. To pre-allocate, we create a batch with max_seq_len, fun a forward/backward, but no step. 

If we dont do it, as batch_sizes change, the memory gets fragmented, resulting in higher reserved memory. But it didnt seem harmful. I will put a PR suggesting we do it in our recipes. Worst case scenario, the user will get OOM in the very first batch, instead in the middle of training.

[...]

PR #1273, opened 2024-08-06
Updating Qwen2 small model configs and component/model builders

SalmanMohammadi:
#### Context
- [ ] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

Removing low-memory config params from 0.5B and 1.5B Qwen2 models, since they probably don't need them and they will slow training down.
Also ensure validity between `apply_lora_to_output` and `tie_word_embeddings`

Tests:

```
                         __..--.._
  .....              .--~  .....  `.
.":    "`-..  .    .' ..-'"    :". `
` `._ ` _.'`"(     `-"'`._ ' _.' '
     ~~~      `.          ~~~
              .'
             /
            (
             ^---'

```


joecummings (commented):
Definitely agree with the approach, but eyes aren't gunna get this one by me, no matter how cool the ASCII art is. 
* Can you run 10 steps for each of these so we can get a sense of the max memory allocated and reserved?
* Can you test the batch sizes with convergence? I think as long as memory is below 24 GB, you can increase the batch size. Otherwise might be worth keeping in a smaller grad acc value.

recipes/configs/qwen2/0.5B_lora.yaml
```diff
@@ -23,8 +23,7 @@ model:
   _component_: torchtune.models.qwen2.lora_qwen2_0_5b
   lora_attn_modules: ['q_proj', 'v_proj']
   apply_lora_to_mlp: False
-  apply_lora_to_output: False
-  lora_rank: 8
+  lora_rank: 64
```
joecummings, line 26: Any reason we don't just want to allow them to LoRA-ify the output, with default to False?

SalmanMohammadi: 0.5B and 1.5B have tied embeddings

SalmanMohammadi:
> Definitely agree with the approach, but eyes aren't gunna get this one by me, no matter how cool the ASCII art is.
> 
>     * Can you run 10 steps for each of these so we can get a sense of the max memory allocated and reserved?
> 
>     * Can you test the batch sizes with convergence? I think as long as memory is below 24 GB, you can increase the batch size. Otherwise might be worth keeping in a smaller grad acc value.

```
⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⢀⣴⠆⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀
⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⣰⠟⠁⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⣀⣠⡄
⣄⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⡀⠀⢰⠃⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠐⢀⡄⠚⢁⣠⠴⠚⠋⠉⠉⠁
⠈⠑⢦⡀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⡇⠀⡎⠀⠀⠀⣀⡀⡀⠀⠀⠀⠠⢀⠔⠁⣠⡶⠋⠁⠀⠀⠀⠀⠀⠀
⠀⠀⠀⠈⠻⢤⣀⠀⠈⠓⢤⣀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⡇⠀⢇⠀⠈⠁⠀⠀⠀⠀⢸⠟⡔⢁⣤⣾⣿⣿⡶⢶⣤⣤⡀⠀⠀⠀
⠀⠀⠀⠀⠀⠀⠉⠳⢄⡀⠀⠈⠓⢤⡀⠀⠀⠀⠀⠀⠀⣸⡆⢳⡀⢸⡄⠀⠀⠀⠀⠀⠀⡇⢀⣶⣿⡿⠛⣻⣿⣯⢾⣷⡉⣷⡄⠀⠀
⠀⠀⠀⠀⠀⠀⠀⠀⠀⠉⠢⣄⠀⠀⠙⢦⡀⠀⢀⡤⠾⠎⢱⡄⢣⠀⢳⡀⠀⠀⠀⠀⠀⢠⠏⣼⠉⠛⠶⢧⣤⠼⠏⠛⠛⠉⢻⠀⠀
⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠈⠙⠦⣀⠀⠙⠰⠋⠀⠀⠀⠀⠙⠂⢣⡈⢷⡀⠀⠀⠀⢀⡟⢰⠇⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠘⡇⠀
⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠈⠓⢦⣀⠀⠀⠀⠀⠀⠀⠀⠀⠱⣄⠻⣄⠀⠀⢸⡇⣼⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⢠⡇⠀
⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠈⠓⢤⡀⠀⠀⠀⠀⠀⠀⠈⠀⠈⢷⣄⠀⢿⣿⠆⠀⠀⠀⠀⠀⠀⠀⠀⠀⢀⡾⠁⠀
⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠙⣦⣄⠀⠀⠀⠀⠀⠀⠀⠀⠈⢹⡿⠿⢦⣀⡀⠀⠀⠀⠀⠀⣀⡴⠟⠁⠀⠀
⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⢠⠞⣴⣟⢧⣄⠀⠀⠀⠀⠀⠀⢠⠏⠁⠀⣠⣄⣉⡉⠓⠒⠋⠉⠉⠀⠀⠀⠀⠀
⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⡴⢋⡞⠁⠈⠙⢿⠀⠀⠀⠀⠀⢀⡟⠀⠀⠀⡿⠀⠈⠉⠉⠉⢻⠆⠀⠀⠀⠀⠀⠀
⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⢀⠞⣱⢋⡀⢀⠀⠀⢸⡀⠀⠀⠀⠀⣾⠁⠀⠀⢠⠃⠀⠀⠀⠀⠀⡟⠀⠀⠀⠀⠀⠀⠀
⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⣠⢏⡞⢁⠞⠀⡈⠳⡀⢰⠃⠀⠀⠀⠀⢸⡀⠀⠀⢸⠀⠀⠀⠀⠀⣼⠃⠀⠀⠀⠀⠀⠀⠀
⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⢀⡼⣱⠋⢠⠊⠀⡼⠁⠀⡇⠸⠀⠀⠀⠀⠀⠈⣧⠀⠀⣼⠀⠀⠀⠀⢠⡏⠀⠀⠀⠀⠀⠀⠀⠀
⠓⢄⡀⠀⠀⠀⠀⠀⠀⠀⠀⠀⣠⢏⡼⠁⣠⠋⠢⣼⠁⠀⢀⣷⣀⣀⣀⣰⣟⡳⣄⡘⣦⢀⡟⠀⠀⠀⠀⣼⠀⠀⠀⠀⠀⠀⠀⠀⠀
⠀⠀⠙⢦⣀⠀⠀⠀⠀⠀⠀⣴⣣⠏⠀⢠⠃⠀⠀⠉⢧⣰⠟⢳⣤⠾⠛⣿⠛⠛⡛⡛⠛⠋⠉⠉⠉⠉⣷⣇⠀⠀⠀⠀⠀⠀⠀⠀⠀
⠀⠀⠀⠀⠈⠳⣄⠀⠀⢠⣾⡿⠁⠀⠀⠛⢦⡀⠀⠀⠈⡟⠀⠀⠙⢶⢶⢿⠒⠚⣿⣿⠛⠛⠋⠉⠉⡟⣻⠁⠀⠀⠀⠀⠀⠀⠀⠀⠀
⠀⠀⠀⠀⠀⠀⠀⠙⢶⣿⣿⡄⠀⠀⠀⠀⠀⠹⡄⠀⢸⡇⠀⠀⠀⢸⢸⣼⠀⠀⣿⣿⠀⠀⠀⠀⠀⡇⡿⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀
```

you got it. fwiw I grabbed these values from the gemma 2b configs

SalmanMohammadi:
Repporting: maximum peak memory (GB), ~median toks/sec, over 250 steps. 
Tested on a 2080 8GB.
[WandB runs here](https://wandb.ai/salman-mohammadi/torchtune/)

| **Config**              | Active | Allocated | Reserved | toks/sec |
| ----------------------- | ------ | --------- | -------- | -------- |
| 0.5B full single device | 6.7    | 6.7       | 7.8      | 800      |
| 0.5B LoRA single device | 6      | 6         | 8        | 1200     |
| 1.5B full single device | 8      | 8         | 8        | 400      |
| 1.5B LoRA single device | 6.5    | 6.5       | 7        | 450      |

If you'd prefer to see higher throughput we can bump bsz and announce with support for 12GB.

recipes/configs/qwen2/0.5B_lora_single_device.yaml
```diff
@@ -19,11 +19,10 @@
 # Model Arguments
 model:
   _component_: torchtune.models.qwen2.lora_qwen2_0_5b
-  lora_attn_modules: ['q_proj', 'v_proj']
+  lora_attn_modules: ['q_proj', 'k_proj', 'v_proj']
```
joecummings, line 22: why did you add k_proj here?

recipes/configs/qwen2/0.5B_lora.yaml
```diff
@@ -23,8 +23,7 @@ model:
...
   lora_alpha: 16
```
joecummings, line 27: I think rule of thumb to start is that alpha should be half of rank?

recipes/configs/qwen2/1.5B_full_single_device.yaml
```diff
@@ -51,12 +52,15 @@ batch_size: 2
...
+  lr: 2e-5
+
 optimizer_in_bwd: True
+
 loss:
   _component_: torch.nn.CrossEntropyLoss
+
 max_steps_per_epoch: null
-gradient_accumulation_steps: 16
+gradient_accumulation_steps: 1
```
joecummings, line 63: What's the reasoning for bumping this down?

recipes/configs/qwen2/1.5B_lora_single_device.yaml
```diff
@@ -19,10 +19,9 @@
 # Model Arguments
 model:
   _component_: torchtune.models.qwen2.lora_qwen2_1_5b
-  lora_attn_modules: ['q_proj', 'v_proj']
+  lora_attn_modules: ['q_proj', 'k_proj']
```
joecummings, line 22: Why did you change this to 'k_proj'?

recipes/configs/qwen2/1.5B_lora_single_device.yaml
```diff
@@ -19,10 +19,9 @@
...
   apply_lora_to_mlp: False
-  apply_lora_to_output: False
-  lora_rank: 8
+  lora_rank: 64
   lora_alpha: 16
```
joecummings, line 25: Sam comment as before with alpha being half of rank

recipes/configs/qwen2/0.5B_lora_single_device.yaml
SalmanMohammadi: I had memory spare since increasing batch size OOM-ed, thought it might help performance

recipes/configs/qwen2/1.5B_full_single_device.yaml
SalmanMohammadi: opt in bwd

joecummings (commented):
Thanks for doing the tests, the WandB runs are extremely helpful. I think the 8GB looks reasonable, but I have a couple questions as to the minute changes you made to the configs. 

Not a blocker for merging this PR, but did you check how much compile helped for the 1.5B full single device and the 0.5 full single device?

recipes/configs/qwen2/1.5B_lora_single_device.yaml
SalmanMohammadi: mistake!

SalmanMohammadi:
> Not a blocker for merging this PR, but did you check how much compile helped for the 1.5B full single device and the 0.5 full single device?

Just gave it a quick try. No difference. Might make a difference on more modern hardware?

recipes/configs/qwen2/0.5B_lora_single_device.yaml
```diff
@@ -19,11 +19,10 @@
...
   apply_lora_to_mlp: False
-  apply_lora_to_output: False
-  lora_rank: 8
-  lora_alpha: 16
+  lora_rank: 64
+  lora_alpha: 32
```
felipemello1, line 25: it should be the opposite :P 

alpha is double of rank. Check llama configs for example 
image

btw, I think that having LoRA on the MLP is better than adding 'k_proj'.

recipes/configs/qwen2/0.5B_lora_single_device.yaml
```diff
@@ -39,20 +38,22 @@ checkpointer:
...
 shuffle: True
-batch_size: 2
+batch_size: 4
 
 # Optimizer and Scheduler
 optimizer:
   _component_: torch.optim.AdamW
   weight_decay: 0.01
-  lr: 3e-4
+  lr: 2e-3
```
felipemello1, line 55: is the change in LR is because we tested it / saw it somewhere, or is it just a guess?

recipes/configs/qwen2/1.5B_lora.yaml
```diff
@@ -23,9 +23,8 @@ model:
   _component_: torchtune.models.qwen2.lora_qwen2_0_5b
```
felipemello1, line 23: i am misreading it? The config name says 1.5, but the component and instructions are 0.5

I personally dont like having "." in the middle of a file name. I wonder if it should be 1_5B

recipes/configs/qwen2/0.5B_lora_single_device.yaml
SalmanMohammadi: I tested when benchmarking

recipes/configs/qwen2/1.5B_lora.yaml
SalmanMohammadi: good catch! I hadn't tested distributed. I'm not married to the config name but I agree with you

recipes/configs/qwen2/0.5B_lora_single_device.yaml
SalmanMohammadi: I don't think this is consistent across our configs. For example, Gemma2B has:

```
model:
  _component_: torchtune.models.gemma.lora_gemma_2b
  lora_attn_modules: ['q_proj', 'k_proj', 'v_proj']
  apply_lora_to_mlp: True
  lora_rank: 64
  lora_alpha: 16
```

I think you're right though. Happy to change this

SalmanMohammadi: note: `apply_lora_to_mlp` did impact training speed/mem usage for me without significantly improving loss

SalmanMohammadi:
small note: I needed https://github.com/pytorch/torchtune/issues/1185 to make these experiments fit on my GPU.

SalmanMohammadi:
> small note: I needed #1185 to make these experiments fit on my GPU.

not sure if this pushes the needle on that issue? it's listed on the docs in stable (https://pytorch.org/docs/stable/notes/cuda.html#environment-variables) but denoted "experimental"

joecummings:
> small note: I needed #1185 to make these experiments fit on my GPU.

All experiments? Or just the 1.5 full finetuning?

recipes/configs/qwen2/0.5B_lora.yaml
```diff
@@ -65,20 +66,21 @@ loss:
...
   log_dir: ${output_dir}
+
 log_every_n_steps: 1
 log_peak_memory_stats: False
 
 # Environment
 device: cuda
 dtype: bf16
-enable_activation_checkpointing: False
+enable_activation_checkpointing: True
```
joecummings, line 83: We need activation checkpointing now?

recipes/configs/qwen2/0.5B_lora_single_device.yaml
joecummings: Ah misspoke - definitely go with Felipe's suggestion. Alpha being double rank.

recipes/configs/qwen2/0.5B_lora.yaml
SalmanMohammadi: I think the tradeoff for double batch size was net positive, but I can double check

SalmanMohammadi:
> All experiments? Or just the 1.5 full finetuning?

I'm going to say all to be safe.

joecummings (approved):
This looks good to me - thanks for doing those tests. 

A merge on main should fix the failing CI.

PR #1214, opened 2024-07-24
[RFC] Chat preference datasets

SalmanMohammadi:
#### Context
What is the purpose of this PR? Is it to
- [x] add a new feature
- [ ] fix a bug
- [ ] update tests and/or documentation
- [ ] other (please add here)

#### Not all preference datasets were created equal

Some of the most popular preference datasets on HF [are chat datasets](https://huggingface.co/datasets?other=preference&sort=likes). Let's show them some love. This PR continues to flesh out our support for alignment fine-tuning by adding support for chat-style preference datasets. This also provides a concrete example for when we consider integrating preference datasets with #1186.

```python
from torchtune.datasets import chat_preference_dataset
from torchtune.data import ChatMLFormat, get_openai_messages
from torchtune.models import llama2

tokenizer = llama2.llama2_tokenizer("./target/dummy/tokenizer.model")

ds = chat_preference_dataset(
    tokenizer=tokenizer,
    source = "mlabonne/orpo-dpo-mix-40k",
    conversation_style="openai",
    chat_format="torchtune.data.ChatMLFormat",
    max_seq_len=512,
    train_on_input=False,
    split="train[:10%]"
)
print(ds[0].keys())

"""
dict_keys(['chosen_input_ids', 'chosen_labels', 'rejected_input_ids', 'rejected_labels'])
"""

print(f"chosen_input_ids:\n \t {tokenizer.decode(ds[0]["chosen_input_ids"])}")
print("-----------")
print(f"rejected_input_ids:\n \t {tokenizer.decode(ds[0]["rejected_input_ids"])}")

"""
chosen_input_ids:
 	 <|im_start|>user
 The setting is an otherworldly, yet eerily familiar, metropolis known as "Zephyria." It's a city suspended in the ether, floating amidst nebulous clouds of cosmic dust. The architecture here is surreal and alien, with buildings that twist and spiral like strands of DNA, reaching towards the cosmos. Streets are paved with luminescent cobblestones, casting soft hues of blues and purples, reflecting off iridescent structures. Strange vegetation, vibrant and bioluminescent, thrives everywhere, creating a stark contrast against the deep indigo sky.

Now, immerse me in this ethereal journey through Zephyria. <|im_end|>
<|im_start|>assistant
[...]

PR #1276, opened 2024-08-06
[4/7] Refactor preference dataset with transforms design

RdoubleA:
#### Context
Following the RFC in https://github.com/pytorch/torchtune/pull/1186, we will use the unified message_transform -> template -> tokenization data pipeline in all our datasets. This PR updates `PreferenceDataset` to follow this pipeline. We can't use `SFTDataset` here because chosen and rejected messages are treated separately.

Covers #1214 

#### Changelog
- Refactor `PreferenceDataset` and `stack_exchange_paired_dataset`
- Add `ChosenRejectedToMessages` message transform
- `StackExchangedPairedTemplate` -> `QuestionAnswerTemplate`
- Update tests

#### Test plan
- Added unit tests
- DPO runs with stack exchange paired against main, DPO run with hh_rlhf_helpful_dataset

image


torchtune/datasets/_stack_exchange_paired.py
```diff
@@ -0,0 +1,76 @@
...
+    Family of preference datasets similar to `StackExchangePaired data
+    <https://huggingface.co/datasets/lvwerra/stack-exchange-paired>`_.
+
+    Args:
+        tokenizer (ModelTokenizer): Tokenizer used by the model that implements the ``tokenize_messages`` method.
+        source (str): path to dataset repository on Hugging Face. For local datasets,
+            define source as the data file type (e.g. "json", "csv", "text") and pass
+            in the filepath in ``data_files``. See Hugging Face's ``load_dataset``
+            (https://huggingface.co/docs/datasets/en/package_reference/loading_methods#datasets.load_dataset.path)
+            for more details. Default is ``liweili/c4_200m``.
```
SalmanMohammadi, line 37: should be  lvwerra/stack-exchange-paired?

SalmanMohammadi:
I've included a couple examples of preference datasets below below, [see here](https://huggingface.co/collections/RLHFlow/standard-format-preference-dataset-662eec0252e194d5d40c252a) for a good list, too.
 
### Chat Format

[Anthropic HH-RLHF (processed)](https://huggingface.co/datasets/RLHFlow/HH-RLHF-Helpful-standard?row=1*)
image

### [Instruct-format](https://huggingface.co/datasets/weqweasdas/preference_dataset_mixture2_and_safe_pku/viewer/default/train?row=0)
image

These datasets are nice examples of processed datasets which are ready for consumption in a standard format. This is one of the things I was unclear about - it seems like we currently have a prompt template for supporting each specific dataset type OOTB? Do you think this scales in a sensible way, or are we just being deliberate with the datasets we provide OOTB to ensure there's sufficient examples for users to specify their own? For example, with preference datasets with standard chat/instruct formats as above, would it be possible for the user specify a standard template we provide (such as ChatML?) rather than having to write one for their specific usecase?

Continuing on from above RE the default datasets we provide, I would also prefer the default exemplar preference dataset to be chat/instruct style. Chat/instruct style preference datasets are the standard for open source preference/DPO-trained models; I think it's a far more common use case (the stack exchange paired dataset is pretty specific), and one of the motivators for this refactor is to unify chat+instruct preference datasets, right? I'd like to make sure there's at least one example of a chat/instruct dataset which allows users to replicate e.g. instruct models tuned with DPO in torchtune.

Sorry to make you explain probably obvious things here - I'm trying to get the full picture of how an end-user would configure one of these datasets with the current example. 

Thank you so much for helping out with this, I'm really enjoying seeing your datasets refactor masterplan fall into place so neatly.

SalmanMohammadi:
To further motivate my points above, my grand plan for alignment in torchtune is for a user to roughly follow the fine-tuning steps used by e.g. Llama2 or InstructGPT. The dataset in the the PPO config I provide is the chat-style HHRLHF dataset above - I want to demonstrate an end-to-end journey for what I anticipate will be the most common use-case for someone using torchtune for alignment: from SFT, to reward modelling, and finally RLHF using PPO. If I were to write a tutorial for this, I would love to show users how to follow this process with minimal changes to our configs (in an ideal world, without modifying any code at all!!).

RdoubleA:
@SalmanMohammadi Thanks for sharing these examples. I suppose I generalized a bit based off of stack exchange paired but as you mention it is rather niche, although already present in our library. I'm happy to add one or both of the datasets you recommended as builders, both of those seem to share the same `message_transform` and that might be more generalizable than the one I put for stack exchange.

RdoubleA:
> it seems like we currently have a prompt template for supporting each specific dataset type OOTB? 

Not exactly, the prompt template is not required. A lot of the exemplar datasets have it just to show plenty of different examples of how you would configure a custom dataset. Users can use any or no prompt template when making their own builders.

torchtune/datasets/_preference.py
```diff
@@ -10,64 +10,98 @@
...
     ) -> None:
         self._tokenizer = tokenizer
+        self._prompt_template = prompt_template
+        self._message_transform = message_transform
         self._data = load_dataset(source, **load_dataset_kwargs)
-        self.template = template
-        self._transform = transform
-        self._column_map = column_map
-        self.max_seq_len = max_seq_len
-        self._data = self._data.filter(
```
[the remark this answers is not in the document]
felipemello1: never mind, i see that stack exchange has it. But if we expect that this logic will have to be in most datasets by default, i think it may be better to keep it here, no?

torchtune/datasets/_preference.py
```diff
@@ -10,64 +10,98 @@
...
+        prompt_template (Optional[PromptTemplate]): template used to format the messages based on their role. This is used
+            to add structured text around the actual messages and is called on both chosen and rejected messages.
+            The structured text is used in three scenarios:
+
+            - Task-specific templates to gear models for a particular task that it will expect after training
+            - Model-specific templates that are required whenever the model is prompted, such as the [INST]
+              tags in Llama2 and in Mistral
+            - Community standardized templates, such as :class:`~torchtune.data.ChatMLFormat`
+
+            The extra text will still get tokenized as normal text, not as special tokens.
```
felipemello1, line 80: i think i understand what "Extra text" means, but maybe we can make it super clear? in my mind, at first glance, if there is a template, there shouldnt be extra text.

felipemello1 (dismissed):
some minor comments. Changes make sense to me. I did not build the docs. Have you had a chance to make sure it renders well? There is also a build doc error

torchtune/data/_prompt_templates.py
```diff
@@ -152,3 +152,18 @@ def __call__(self, messages: List[Message]) -> List[Message]:
 
 Please see :class:`~torchtune.data.PromptTemplate` for full API arguments.
 """
+QuestionAnswerTemplate = partial(
```
ebsmothers, line 155: These do render awkwardly in the live docs, just fyi: 

Screenshot 2024-08-08 at 7 24 58 AM

Not sure if there's a way around this but just something to be aware of

torchtune/datasets/_preference.py
```diff
@@ -10,64 +10,98 @@
...
 
 from torchtune.modules.tokenizers import ModelTokenizer
+from torchtune.modules.transforms import Transform
 
 
 class PreferenceDataset(Dataset):
     """
-    Class that supports any custom dataset with instruction-based prompts and a
-    configurable template.
+    Primary class for fine-tuning via RLHF/PPO or DPO methods on a preference dataset
```
ebsmothers, line 21: Is this true for PPO? I thought there we use text completion dataset

torchtune/datasets/_preference.py
```diff
@@ -10,64 +10,98 @@
...
+    sourced from Hugging Face Hub, local files, or remote files. This class requires
+    the dataset to have "chosen" and "rejected" model responses. At a high level, this
```
ebsmothers, line 23: Can consider giving an example raw data format here

tests/torchtune/datasets/test_preference_dataset.py
```diff
@@ -0,0 +1,117 @@
...
+
+        rejected_messages = [
+            Message.from_dict(sample["prompt"][0]),
+            Message.from_dict(sample["rejected"][0]),
+        ]
+
+        return {"chosen": chosen_messages, "rejected": rejected_messages}
+
+
+class TestSFTDataset:
```
ebsmothers, line 33:
```suggestion
class TestPreferenceDataset:
```

torchtune/datasets/_preference.py
```diff
@@ -77,34 +111,26 @@ def __getitem__(self, index: int) -> Dict[str, List[int]]:
...
+        transformed_sample = self._message_transform(sample)
+        if self._prompt_template is not None:
+            transformed_sample["chosen"] = self._prompt_template(
+                transformed_sample["chosen"]
+            )
+            transformed_sample["rejected"] = self._prompt_template(
+                transformed_sample["rejected"]
+            )
+
+        # TODO: Truncation differs from original DPO repo
```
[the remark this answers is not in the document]
ebsmothers: Let's file a follow-up task for it

torchtune/datasets/_stack_exchange_paired.py
```diff
@@ -0,0 +1,76 @@
...
+        "rejected": "response_k",
+    }
+
+    def _filter(x):
+        return (
+            len(x[column_map["prompt"]]) + len(x[column_map["chosen"]])
+            <= tokenizer.max_seq_len
+            and len(x[column_map["prompt"]]) + len(x[column_map["rejected"]])
+            <= tokenizer.max_seq_len
+        )
```
ebsmothers, line 62: I know this was in the original implementation, but is this really the correct usage? Like we are truncating a string to max seq len here, not the tokens? (Unless I'm missing something obvious)

ebsmothers (approved):
A few comments but no major concerns, please make sure other folks' comments are addressed as well

torchtune/datasets/_preference.py
SalmanMohammadi: Should be:

fine-tuning for preference modelling techniques (e.g. training a preference model for RLHF, or directly optimizing a model through DPO).

SalmanMohammadi:
> I'm happy to add one or both of the datasets you recommended as builders

Could we please add a builder for https://huggingface.co/datasets/RLHFlow/HH-RLHF-Helpful-standard?row=1*, which is a chat-style preference dataset?

Let me know if you'd rather see this as a followup.

torchtune/datasets/_stack_exchange_paired.py
SalmanMohammadi: yeah, see here for the original useage

https://github.com/eric-mitchell/direct-preference-optimization/blob/f8b8c0f49dc92a430bae41585f9d467d3618fe2f/preference_datasets.py#L241

SalmanMohammadi: This is something that's been on the back of my mind that has bothered me a little - this is specific to:

- the dataset format used in the original DPO experiments
- the use-case for the original DPO PR

There are many preference datasets which won't have the "prompt" column. I'd personally rather just see this removed as the default.

tests/torchtune/datasets/test_preference_dataset.py
RdoubleA: oop

torchtune/datasets/_stack_exchange_paired.py
ebsmothers: Wait but this pointer is happening after tokenization, right? Aren't we doing this filtering prior to tokenization? In that case we're talking about len(str) here not len(List[int])

RdoubleA: ooh good point. I vote to just remove.

SalmanMohammadi: Sorry, yeah, I'm getting muddled up here. I'd PREFER to remove.

What's the workflow for a user wanting to filter on the tokenized chosen/rejected? Define their own class?

torchtune/datasets/_preference.py
RdoubleA: sorry, didn't realize PPO is text completion

torchtune/datasets/_stack_exchange_paired.py
RdoubleA: > to filter on the tokenized chosen/rejected

dont think this is a real use case, because what can you filter if everything is in integer ids? truncating past max seq len is already handled by the tokenizer

SalmanMohammadi: Yeah okay, I guess with our default preference dataset we already have the chosen/rejected and they're just being tokenized - you don't to construct them and truncate correctly.

torchtune/data/_prompt_templates.py
RdoubleA: so it renders correctly if I did a partial of a function, but not a partial of a class :(

torchtune/datasets/_stack_exchange_paired.py
```diff
@@ -0,0 +1,76 @@
...
+    ChosenRejectedToMessages,
+    PromptTemplate,
+    QuestionAnswerTemplate,
+)
+from torchtune.datasets._preference import PreferenceDataset
+from torchtune.modules.tokenizers import ModelTokenizer
+
+
+def stack_exchange_paired_dataset(
+    tokenizer: ModelTokenizer,
```
SalmanMohammadi, line 19: Why is this `tokenizer: ModelTokenizer`, but all other builders use `model_transform: Transform`?

torchtune/datasets/_preference.py
```diff
@@ -10,64 +10,98 @@
...
         self,
-        tokenizer: ModelTokenizer,
+        *,
         source: str,
-        template: InstructTemplate,
-        transform: Optional[Callable] = None,
-        column_map: Optional[Dict[str, str]] = None,
-        max_seq_len: Optional[int] = None,
+        message_transform: Transform,
+        tokenizer: ModelTokenizer,
```
SalmanMohammadi, line 94: similarly to my comment in the stack exchange builder: why aren't we also using `model_transform (Transform)` here?

torchtune/datasets/_stack_exchanged_paired.py
```diff
@@ -1,46 +0,0 @@
-# Copyright (c) Meta Platforms, Inc. and affiliates.
```
SalmanMohammadi, line 1: this whole time we've used "stack exchange**d**"? lol

SalmanMohammadi: I wouldn't have noticed this if not for the filename diff, no shade meant

SalmanMohammadi: case in point: I have to look up how to `git reset --hard` at least three times a week on stack overflowed

torchtune/datasets/_preference.py
```diff
@@ -10,64 +10,98 @@
...
     Args:
-        tokenizer (ModelTokenizer): Tokenizer used by the model that implements the ``tokenize_messages`` method.
         source (str): path to dataset repository on Hugging Face. For local datasets,
             define source as the data file type (e.g. "json", "csv", "text") and pass
-            in the filepath in ``data_files``. See Hugging Face's ``load_dataset``
-            (https://huggingface.co/docs/datasets/en/package_reference/loading_methods#datasets.load_dataset.path)
+            in the filepath in ``data_files``. See `Hugging Face's
+            <https://huggingface.co/docs/datasets/en/package_reference/loading_methods#datasets.load_dataset.path>`_
+            ``load_dataset`` for more details.
+        message_transform (Transform): callable that keys into the desired fields in the sample
```
SalmanMohammadi, line 67: pretty tangential point, but the behaviour in this docstring isn't enforced by the ``Transform`` protocol itself right? In that the transform should always have `def __call__(...) -> List[Message]:`

torchtune/datasets/_preference.py
RdoubleA: we could make it `model_transform`, but preference will always use pure text tokenization. unless you know of multimodal examples of preference datasets?

torchtune/datasets/_stack_exchanged_paired.py
RdoubleA: the name was bothering me for a long time 😅

felipemello1: "overflowed" lol

torchtune/datasets/_preference.py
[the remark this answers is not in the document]
RdoubleA: this should be removed because it is actually incorrect, max seq length is based on the tokens not on the string length, and so we shouldn't filter on string length here

PR #1277, opened 2024-08-06
Expose max_seq_len in tokenizer builders

RdoubleA:
#### Context
We moved the `max_seq_len` parameter from dataset builders to tokenizer classes, but we never exposed it in the tokenizer builders. This PR updates all tokenizers with max_seq_len so you can modify max seq len from the config right under the tokenizer. Also fixed a bug in phi3 tokenizer, and removed redundant llama3_tokenizer for Llama3.1

#### Test plan
EYES

### DIFF
--- a/torchtune/models/gemma/_model_builders.py
+++ b/torchtune/models/gemma/_model_builders.py
@@ -3,7 +3,7 @@
 #
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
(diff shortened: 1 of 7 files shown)


joecummings (approved):
What's the bug in Phi3?

RdoubleA:
`parse_hf_tokenizer_json` was never imported despite being used. we never hit this bug because we never had custom special tokens :)

PR #1280, opened 2024-08-07
Adding Intel XPU backend support using device-agnostic way

shiyan1121:
#### Context
This PR enhances the Torchtune framework by introducing support for the Intel XPU backend.

#### Changelog
To avoid the redundant hard-code for XPU logic and simplify the device abstraction, we follows the implementation in PyTorch FSDP:
[Making fsdp device-agnostic for custom-backend which implement cuda-semantics #99024](https://github.com/pytorch/pytorch/pull/99024)

In the `torchtune/utils/_device.py` file, we introduce the `_DeviceHandle` class. Users can initialize this class with their device by using `device_handle = _DeviceHandle.from_device(my_device)`.  Since XPU also follows the CUDA semantics and provides CUDA-like APIs, e.g. `torch.xpu.empty_cache()`, most `torch.cuda.` semantics could be replaced by `device_handle.`, which are available both for CUDA and XPU, and even for other CUDA-like devices.

For some logics that the `device_handle` strategy cannot be implemented, we resort to using XPU related judgement to handle different scenarios.


ebsmothers:
Thanks @HJG971121 for the PR. Overall I think the `_DeviceHandle` abstraction is pretty nice. 

One very basic question is around the usage of `intel_extension_for_pytorch` and `oneccl_bindings_for_pytorch`, are these just defining all the necessary bindings to the XPU backend? And if so, do they need to be imported globally, or can we put them in a utility somewhere? 

And a second question: does torchtune run on XPU with the changes in this PR? It'd be nice to add some sample runs confirming things run and have reasonable performance.

HJG971121:
Here is our example of command line for llama2-7b full finetuning on single device, where the device is specified as XPU:

`tune run full_finetune_single_device --config llama2/7B_full_low_memory device=xpu`

recipes/full_finetune_distributed.py
```diff
@@ -332,7 +331,7 @@ def _setup_model(
             # Initialize empty modules on all non-zero ranks
             param_init_fn=(
                 lambda module: module.to_empty(
-                    device=torch.device("cuda"), recurse=False
+                    device=torch.device("xpu") if torch.xpu.is_available() else torch.device("cuda"), recurse=False
```
RdoubleA, line 334: If device_handle is serving as an intermediary wrapper to handle all device related APIs agnostic of whether that is CPU, CUDA, or XPU, shouldn't we work with device_handle directly here instead of torch.device? My understanding is that we only need to find the appropriate device one time in the recipe (via `get_device`), then call `get_device_handle`, and use the `device_handle` for all these operations. Or am I misunderstanding?

recipes/full_finetune_distributed.py
```diff
@@ -588,8 +587,10 @@ def recipe_main(cfg: DictConfig) -> None:
             "Distributed finetune recipe should be run via a distributed launcher."
             "If using tune CLI, please specify --nnodes 1 and --nproc_per_node [num_gpus]"
         )
-
-    init_process_group(backend="gloo" if cfg.device == "cpu" else "nccl")
+    if torch.xpu.is_available():
```
RdoubleA, line 590: same comment here

recipes/lora_finetune_single_device.py
```diff
@@ -540,10 +541,7 @@ def train(self) -> None:
                                 "lr": self._optimizer.param_groups[0]["lr"],
                                 "tokens_per_second_per_gpu": num_tokens / time_per_step,
                             }
-                            if (
-                                self._device.type == "cuda"
-                                and self._log_peak_memory_stats
-                            ):
+                            if self._log_peak_memory_stats:
```
RdoubleA, line 544: do we want to guard against CPU here? `get_memory_stats` will throw an error but maybe we shouldn't log at all if on CPU. (actually, none of our recipes support CPU training, so maybe we should just let it error out, @ebsmothers)

torchtune/utils/_device.py
```diff
@@ -5,9 +5,50 @@
...
 
 import torch
+if torch.xpu.is_available():
+    import intel_extension_for_pytorch
+
+class _DeviceHandle:
+    """
+    This is a simple abstraction for computing devices,
+    which enables custom backends that implement CUDA-like
+    semantics.
```
RdoubleA, line 18: Could you provide a few more details here on how this should be used? MY understanding is that this is used in place of device. A few examples would be helpful.

torchtune/utils/_device.py
```diff
@@ -5,9 +5,50 @@
...
+        Return an device handle corresponding to the device, and through this handle,
+        operations with the same semantics as CUDA can be performed on the device.
+        Just return torch.cuda if the device is cuda to make attribute-access faster.
+        Custom backend must first register a module with the same name with {device.type} on torch.
+        """
+        if device.type == "cuda":
+            return cast(_DeviceHandle, torch.cuda)
+        return cls(device)
+
+    def __getattr__(self, __name: str) -> Any:
```
RdoubleA, line 45: a quick docstring here would be helpful, something along the lines of "retrieve attribute from correct backend based on registered device"

torchtune/utils/_device.py
```diff
@@ -37,28 +78,35 @@ def _setup_cuda_device(device: torch.device) -> torch.device:
...
     # Ensure index is available before setting device
-    if device.index >= torch.cuda.device_count():
-        raise RuntimeError(
-            "The local rank is larger than the number of available GPUs."
-        )
-
-    torch.cuda.set_device(device)
+    device_handle = get_device_handle(device)
+    if device.index >= device_handle.device_count():
+            raise RuntimeError(
```
RdoubleA, line 90: nit: extra tab here

RdoubleA (commented):
Like the idea of the DeviceHandle a lot. Had a few questions around its usage.

recipes/full_finetune_distributed.py
HJG971121: Thank you for your comment! 
Your understanding is correct，and it's my mistake that I did't use unified device handle in this instance. 

Additionally, I suggest that code here could be rewritten as 
`module.to_empty(device = self._device.type, recurse=False)`
to avoid redundant device verification, given that the device attribute is already specified.

torchtune/utils/_device.py
HJG971121: This `device_handle`is designed to eliminate redundant code when implementing CUDA-like APIs for your customized device. For instance, consider the XPU. To empty the cache, a simplistic approach involves using conditional statements:

`torch.xpu.empty_cache() if device == 'xpu' else torch.cuda.empty_cache()`

This method, however, lacks efficiency and flexibility, particularly as the number of GPU devices increases in the future. Therefore, we adopt the `device_handle` strategy to integrate both CUDA APIs and the CUDA-like APIs of the customized device.

To utilize this strategy, you should first define your `device_handle` based on your device:

`device_handle = get_device_handle (device) # device could be ‘cuda’, ‘xpu’ or your customed device`

Then, the code for emptying the cache can be rewritten as:

`device_handle.empty_cache()`

In this setup, `device_handle` replaces direct calls to `torch.cuda` or `torch.xpu`, streamlining the process across different device types. Commonly used CUDA APIs, such as `torch.cuda.memory_allocated()` and `torch.cuda.synchronize()`, can be utilized in the same manner.

recipes/full_finetune_distributed.py
HJG971121: Thank you for pointing that out! 

Indeed, there are scenarios where the `device_handle` might not be applicable, for example `backend` needs to be determined based on the device type. 

And I think using `cfg.device` as the condition variable may be a simpler way. I will rewrite the code, thank you.

torchtune/utils/_device.py
HJG971121: Thank you for your suggestion! I will add docstring to it.

zhuhong61:
Convert to Draft. We are working on preparing the RFC for the discussion about the Intel PyTorch distributed design. We will push forward the process after reaching an alignment :)

Ankur-singh:
> Convert to Draft. We are working on preparing the RFC for the discussion about the Intel PyTorch distributed design. We will push forward the process after reaching an alignment :)

Hi @zhuhong61 any updates on this?

PR #1284, opened 2024-08-07
[5/7] SFTDataset: Refactor Alpaca

RdoubleA:
#### Context
- [x] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

As discussed in the RFC in #1186, we will merged instruct and chat datasets to the following unified pipeline that can better support multimodal:
1. `message_transform` to create `List[Message]` from dataset with full flexibility on columns, ad-hoc modifications, etc. For multimodal, additionally images are loaded from the path
2. `prompt_template` as a optional way to add structured text around specific roles in the list of messages
3. `model_transform` that takes the list of messages and tokenizes it. For multimodal, it will additionally apply model-specific image transforms to the images associated with the sample

**For ease of review, we will stage this as multiple moderate-sized PRs**. This PR refactors our darling `alpaca_dataset` to use the new `SFTDataset`. Tangibly:
- Update alpaca_dataset builder
- Create alpaca specific message transform (because of the weird template switch when you don't have an input)
- Update some stale docstrings
- Update the `dummy_alpaca_dataset` used in integration tests
- Fix some bugs in message transforms

#### Test plan
- [x] update live docs and check rendering
- [x] unit tests for new `AlpacaToMessages`
- [x] update alpaca tests
- [x] loss curves before and after
image


torchtune/datasets/_alpaca.py
```diff
@@ -6,20 +6,90 @@
...
+from typing import Any, Dict, Mapping, Optional, Union
 
-from torchtune.modules.tokenizers import ModelTokenizer
+from torchtune.data import Message
+from torchtune.datasets._packed import PackedDataset
+from torchtune.datasets._sft import SFTDataset
+from torchtune.modules.transforms import Transform
+
+
+class AlpacaToMessages(Transform):
```
ebsmothers, line 17: This I find a bit confusing. Why are most message transforms in `_messages.py` but then this one is here? (I know there is some logic custom to Alpaca, but still)

torchtune/datasets/_alpaca.py
```diff
@@ -6,20 +6,90 @@
...
+        key_output = column_map.get("output", "output")
+
+        if key_input in sample and sample[key_input]:
+            prompt = self.template["prompt_input"].format(
+                instruction=sample[key_instruction], input=sample[key_input]
+            )
+        else:
+            prompt = self.template["prompt_no_input"].format(
+                instruction=sample[key_instruction]
+            )
```
ebsmothers, line 65: Yeah this is a bit awkward.. so what's the core of the issue here? Is it that for Alpaca there is not a guaranteed 1:1 mapping of raw strings to messages, and so the usual templating on messages does not apply?

torchtune/datasets/_alpaca.py
RdoubleA: the others are generic that can apply to any dataset of similar structure. Alpaca's is specific to alpaca and should remain here. kind of like how model specific components are in model folders but all generic components go on modules

torchtune/datasets/_alpaca.py
RdoubleA: Yeah exactly... open to alternative suggestions. It doesn't work as a standard prompt template that takes `List[Message]` and returns `List[Message]` because it depends on the columns

joecummings:
Are these test failures relevant?

RdoubleA:
@joecummings seems to be NCCL errors :( but I don't think it's related

joecummings (approved):
This LGTM as long as CI is green

PR #1286, opened 2024-08-07
[6/7] SFTDataset: deprecate instruct/chat

RdoubleA:
#### Context
Add a new decorator function `deprecated`, the grim reaper coming to announce the expiration of your favorite classes. The following are currently on the chopping block:
- `ChatFormat` (replaced by `PromptTemplateInterface`)
- `InstructTemplate` (replaced by `PromptTemplateInterface`)
- `ChatDataset` (replaced by `SFTDataset`)
- `InstructDataset` (replaced by `SFTDataset`)
- `get_openai_messages` (replaced by `JSONToMessages`)
- `get_sharegpt_messages` (replaced by `ShareGPTToMessages`)

Rest In Power to the following:
- `Llama2ChatFormat` (replaced by `Llama2ChatTemplate`)
- `MistralChatFormat` (replaced by `MistralChatTemplate`)
- `ChatMLFormat` (replaced by `ChatMLTemplate`)

`AlpacaInstructTemplate` will be removed by #1284 and `StackExchangedPairedTemplate` removed by #1276.

#### Test plan
unit test for `deprecated`. you can also see how the logs look when you run the tests locally:
```
tests/torchtune/data/test_instruct_templates.py:43
  /data/users/rafiayub/torchtune-rafiayub/tests/torchtune/data/test_instruct_templates.py:43: FutureWarning: AlpacaInstructTemplate is deprecated and will be removed in future versions. 
    template = AlpacaInstructTemplate()

tests/torchtune/data/test_converters.py::TestShareGPTToLlama2Messages::test_conversion
  /data/users/rafiayub/torchtune-rafiayub/tests/torchtune/data/test_converters.py:35: FutureWarning: get_sharegpt_messages is deprecated and will be removed in future versions. Please use `torchtune.data.ShareGPTToMessages` with `torchtune.datasets.SFTDataset` instead.
    converted_messages = get_sharegpt_messages(self.samples)

tests/torchtune/data/test_converters.py::TestOpenAIToLlama2Messages::test_conversion_conversations_key
  /data/users/rafiayub/torchtune-rafiayub/tests/torchtune/data/test_converters.py:81: FutureWarning: get_openai_messages is deprecated and will be removed in future versions. Please use `torchtune.data.JSONToMessages` with `torchtune.datasets.SFTDataset` instead.
    converted_messages_1 = get_openai_messages(self.samples_1)
```


torchtune/utils/logging.py
```diff
@@ -25,3 +29,34 @@ def get_logger(level: Optional[str] = None) -> logging.Logger:
         level = getattr(logging, level.upper())
         logger.setLevel(level)
     return logger
+
+
+def deprecated(msg: str = "") -> Callable[[T], T]:
```
joecummings, line 34: To avoid blowing up people's logs, is it possible to make sure that:
1. This only logs once (not every time it is hit)
2. This only logs on rank0

ebsmothers: Nice-to-have: include an explicit version number when we will remove support for the API

ebsmothers: Also this is kinda orthogonal, but I don't like that our utils directory == trainer utils. Because it takes deps on pretty much every other directory, which means we may run into circular import issues if we try to use this in our lower-level components. I would think about whether there's a place further upstream in our dependency graph we can put this (if only we didn't already commandeer the utils folder name with our furthest downstream components..)

torchtune/data/_instruct_templates.py
```diff
@@ -37,6 +39,7 @@ def format(
         pass
 
 
+@deprecated()
```
ebsmothers, line 42: Why no message for this one?

torchtune/data/_converters.py
```diff
@@ -62,6 +66,9 @@ def get_sharegpt_messages(
     return messages
 
 
+@deprecated(
+    msg="Please use `torchtune.data.JSONToMessages` with `torchtune.datasets.SFTDataset` instead."
+)
```
ebsmothers, line 71: Maybe slight overkill (and also the usefulness of this depends on how much people actually use a given API, which we don't have clear insights into), but for any migrations that are nontrivial we could consider updating the docstring with a quick example of how to make the change in user code.

torchtune/data/_instruct_templates.py
RdoubleA: re our discussion on the alpaca PR, the template will just be absorbed into the message transform. this class will be removed anyway instead of deprecated

torchtune/utils/logging.py
RdoubleA: 1. the lru cache ensures it's logged only once per class, also verified it in the unit test
2. good point, will add this

RdoubleA: > there's a place further upstream in our dependency graph we can put this

not entirely sure what this would practically mean, do you mean a different file?

torchtune/data/_converters.py
RdoubleA: yeah let's do that. can add a small note to all the docstrings of deprecated classes

torchtune/datasets/_instruct.py
```diff
@@ -18,8 +18,10 @@
 )
 from torchtune.datasets._packed import PackedDataset
 from torchtune.modules.tokenizers import ModelTokenizer
+from torchtune.utils.logging import deprecated
 
 
+@deprecated(msg="Please use `torchtune.datasets.SFTDataset` for custom chat data.")
```
SalmanMohammadi, line 24: I think @ebsmothers already raised this, but worth pointing the user to a concrete example of a replacement?
If not, this says "chat" instead of instruct btw

torchtune/data/__init__.py
```diff
@@ -4,12 +4,7 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
-from torchtune.data._chat_formats import (
-    ChatFormat,
-    ChatMLFormat,
-    Llama2ChatFormat,
-    MistralChatFormat,
-)
```
SalmanMohammadi, line 12: think you gotta remove from api ref

torchtune/utils/logging.py
ebsmothers: Probably a separate directory. Because otherwise anywhere we add `from utils.file_name import SomeClass` will go through `utils/__init__.py`, right? And that will import all our APIs with all their upstream dependencies in data, datasets, modules, models. So we'll get stuck with a circular dependency unless we move this out of utils entirely

torchtune/data/_converters.py
RdoubleA: I'm going to leave this out actually, a lot of concepts (like prompt template) have moved out of the dataset class completely, and it might be more confusing to try to show hard equivalents

torchtune/datasets/_grammar.py
```diff
@@ -74,7 +76,13 @@ def grammar_dataset(
     ds = SFTDataset(
         source=source,
         message_transform=message_transform,
-        model_transform=model_transform,
+        model_transform=tokenizer,
```
ebsmothers, line 79: So we use model_transform in some places and tokenizer in others?

tests/torchtune/utils/test_logging.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+def test_deprecated():
+    @deprecated(msg="Please use `TotallyAwesomeClass` instead.")
+    class DummyClass:
+        pass
+
+    with pytest.warns(
+        FutureWarning,
+        match="DummyClass is deprecated and will be removed in future versions. Please use `TotallyAwesomeClass` instead.",
```
ebsmothers, line 18: this is a nit but it's a bit awkward to me that we split the log message across the utility and the arg we pass. Like your passing msg="Please use `TotallyAwesomeClass` instead" assumes that you know exactly what the first half of the message is, which is somewhat annoying to go check every time

torchtune/datasets/_instruct.py
```diff
@@ -124,73 +134,120 @@ def instruct_dataset(
...
+
+    Raises:
+        ValueError: If ``packed=True`` and ``tokenizer.max_seq_len`` is not set.
     """
-    ds = InstructDataset(
-        tokenizer=tokenizer,
+
+    message_transform = InputOutputToMessages(
+        train_on_input=train_on_input, column_map=column_map
+    )
```
ebsmothers, line 239: Did we discuss renaming this or did I imagine that? Also I'm a bit out of the loop on #1366 but why do we use a single static system prompt in `InputOutputToMessages` now? Is that sufficient? (Sorry if you guys already hashed this all out)

torchtune/data/_prompt_templates.py
```diff
@@ -6,7 +6,7 @@
 from functools import partial
 from typing import Dict, List, Protocol, Tuple, Union
 
-from torchtune.data import Message, Role
+from torchtune.data._messages import Message, Role
```
ebsmothers, line 9: Are these hacks for circular import issues?

tests/torchtune/datasets/test_instruct_dataset.py
```diff
@@ -4,146 +4,91 @@
...
 
-from datasets import Dataset
+import pytest
 from tests.test_utils import DummyTokenizer
 from torchtune.data import InstructTemplate
 from torchtune.data._common import CROSS_ENTROPY_IGNORE_IDX
-from torchtune.datasets import InstructDataset
+from torchtune.datasets import instruct_dataset, InstructDataset
+
+ASSETS = Path(__file__).parent.parent.parent / "assets"
```
ebsmothers, line 15: Not a new problem to this PR but why do we define this in every test file instead of just importing from test_utils or something?

ebsmothers (commented):
Can you update the PR summary? I think it's not actually clearly emphasizing the latest set of changes you've made

tests/torchtune/datasets/test_grammar_dataset.py
```diff
@@ -36,7 +36,7 @@ def test_label_no_masking(self, load_dataset, tokenizer):
             ]
         )
 
-        grammar_ds = grammar_dataset(model_transform=tokenizer, train_on_input=True)
+        grammar_ds = grammar_dataset(tokenizer=tokenizer, train_on_input=True)
```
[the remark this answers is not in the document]
RdoubleA: I'm honestly on the fence here, but discussed with @pbontrager and agreed that we'll take a look at these holistically at the end and make a call once we work on multimodal recipes. For now I've been following text datasets = tokenizer, SFTDataset and multimodal = model transform

torchtune/data/_utils.py
```diff
@@ -73,3 +77,34 @@ def validate_messages(
                 f"System message at index {i} in messages, but system messages must come first"
             )
         last_turn = message.role
+
+
+def deprecated(msg: str = "") -> Callable[[T], T]:
```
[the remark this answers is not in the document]
RdoubleA: I can try this

torchtune/data/_instruct_templates.py
```diff
@@ -10,6 +10,10 @@
 
 class InstructTemplate(ABC):
     """
+    Note:
```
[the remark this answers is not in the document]
RdoubleA: This interface is meant to be inherited and the deprecated decorator doesn't like when it's triggered on an inherited class, wasn't able to drill down on why

torchtune/datasets/_grammar.py
RdoubleA: See other comment

torchtune/datasets/_instruct.py
RdoubleA: Don't recall that discussion but @joecummings was interested in potentially renaming this.

As for the system prompt, yes a single prompt is sufficient to add a system message for every sample conversation.

torchtune/data/_prompt_templates.py
RdoubleA: Indeed

torchtune/data/_converters.py
```diff
@@ -7,12 +7,28 @@
 from typing import Any, List, Mapping
 
 from torchtune.data._messages import Message
+from torchtune.data._utils import deprecated
 
 
+@deprecated(
+    msg="Please use `torchtune.data.ShareGPTToMessages` with `torchtune.datasets.SFTDataset` instead."
```
SalmanMohammadi, line 14:
```suggestion
    msg="Please use `torchtune.data.ShareGPTToMessages` with the `message_transform` argument with `torchtune.datasets.SFTDataset` instead."
```

torchtune/data/_converters.py
```diff
@@ -62,11 +78,26 @@ def get_sharegpt_messages(
     return messages
 
 
+@deprecated(
+    msg="Please use `torchtune.data.JSONToMessages` with `torchtune.datasets.SFTDataset` instead."
```
SalmanMohammadi, line 82:
```suggestion
    msg="Please use `torchtune.data.JSONToMessages` with the `message_transform` argument with `torchtune.datasets.SFTDataset` instead."
```

torchtune/datasets/_alpaca.py
```diff
@@ -114,7 +114,8 @@ def alpaca_dataset(
             <https://huggingface.co/docs/datasets/en/package_reference/loading_methods#datasets.load_dataset.path>`_
             ``load_dataset`` for more details. Default is ``tatsu-lab/alpaca``.
         column_map (Optional[Dict[str, str]]): a mapping from the expected columns in the message transform
-            :class:`~torchtune.data.AlpacaToMessages` to the new column names in the dataset. If None, use
+            :class:`~torchtune.data.AlpacaToMessages` to the new column names in the dataset. Keys should be
+            "instruction", "input", and "output" and values should be the new column names. If None, use
```
SalmanMohammadi, line 118:
```suggestion
            "instruction", "input", and "output" and values should be the new column names. If None, uses
```

torchtune/datasets/_chat.py
```diff
@@ -115,84 +115,150 @@ def chat_dataset(
...
     train_on_input: bool = False,
     packed: bool = False,
     **load_dataset_kwargs: Dict[str, Any],
-) -> ChatDataset:
+) -> Union[SFTDataset, PackedDataset]:
     """
-    Build a configurable dataset with conversations. This method should be
-    used to configure a custom chat dataset from the yaml config instead of
-    using :class:`~torchtune.datasets.ChatDataset` directly, as it is made to be config friendly.
+    Quickly configure a custom dataset with conversations between user and model assistant.
```
SalmanMohammadi, line 125: EPIC nit: remove the qualifier

tests/torchtune/utils/test_logging.py
RdoubleA: Yeah, I could just pass in the full message everytime but that gets tedious. I supposed I could make it a required argument and do that for simplicity.

torchtune/data/_utils.py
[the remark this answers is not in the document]
RdoubleA: unfortunately, any import to `torchtune.utils` requires hitting the `__init__.py` file, so this does not work

SalmanMohammadi (approved):
Looks great @RdoubleA. Thanks for this. Only 14.285714285% of the work left!

torchtune/data/_utils.py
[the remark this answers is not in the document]
ebsmothers: ugh should've known it was too good to be true..

PR #1287, opened 2024-08-08
Full finetune FSDP2 recipe

ebsmothers:
This PR migrates our full finetune recipe onto FSDP2. 

## Main changes
- Define a utility `shard_model` for sharding our models with FSDP2. The recommended way to shard FSDP2 models is by iterating over modules in a bottom-up fashion. As a result, this utility takes in a list of `shard_conditions`. Each shard condition is a callable taking a module name and the module itself (i.e. a pair from `model.named_modules()`) and returning True if the module should be sharded (and False otherwise). We use a list of shard conditions to more easily decouple different things we might want to check (e.g. whether a module is using activation checkpointing, whether we are doing any custom sharding, or whether LoRA is applied (coming in a subsequent PR)). If any of `shard_conditions` return True, the module will be sharded. `shard_model` will also take any FSDP-related fields like `cpu_offload` and `reshard_after_forward` (~fka `sharding_strategy`)
- Deprecate `memory_efficient_fsdp_wrap` flag. Instead, we can provide an optional config `custom_sharded_layers` which is more flexible. For our existing Llama3 8B sharding, we can pass `['tok_embeddings', 'output']` to tell FSDP to shard these modules separately

### Other smaller changes:
- Start the process of moving off of `reset_parameters()` and onto `rope_init()`
- Document a couple places we need to change after all our recipes are on FSDP2
- Remove a version check in `load_model_from_full_state_dict` that's no longer needed

## Test plan

Apart from green CI, 

### Test Llama3 8B perf and memory

On both this PR and main, run 

```
tune run --nproc_per_node 2 full_finetune_distributed --config llama3/8B_full \
max_steps_per_epoch=500 epochs=1 log_peak_memory_stats=True
```

Results are as follows (note that tok/sec plots are smoothed in all figures for clarity)

Screenshot 2024-08-10 at 11 57 12 AM

### Other model tests

**Note:** the losses curves for Gemma 7B differ from those with FSDP1. However, Gemma 7B loss curves appear to be non-reproducible with FSDP1, see #1303.

#### Phi3

Command:

```
tune run --nproc_per_node 2 full_finetune_distributed --config phi3/mini_full \
log_peak_memory_stats=True max_steps_per_epoch=500 epochs=1
```

Screenshot 2024-08-10 at 12 15 23 PM

#### Gemma 2B

```
tune run --nnodes 1 --nproc_per_node 2 full_finetune_distributed --config gemma/2B_full  \
max_steps_per_epoch=500 epochs=1 log_peak_memory_stats=True
```

Screenshot 2024-08-10 at 1 56 49 PM

#### Qwen2 0.5B

Screenshot 2024-08-10 at 1 46 37 PM

### Test checkpoint resume

First run

```
tune run --nproc_per_node 2 full_finetune_distributed --config llama3/8B_full max_steps_per_epoch=10 epochs=2
```

Then run

```
tune run --nproc_per_node 2 full_finetune_distributed --config llama3/8B_full max_steps_per_epoch=10 epochs=2 \
resume_from_checkpoint=True checkpointer.checkpoint_dir=/tmp/Meta-Llama-3-8B-Instruct \
checkpointer.recipe_checkpoint=recipe_state.pt checkpointer.checkpoint_files=[meta_model_0.pt] \
...
Model checkpoint of size 16.06 GB saved to /tmp/Meta-Llama-3-8B-Instruct/meta_model_1.pt
2|20|Loss: 1.0202215909957886: 100%|██████████████████=██| 10/10 
```

#### Custom wrapping

Tried out with Llama2 7B, can see some reduction in reserved memory with no perf impact

```
tune run --nproc_per_node 2 full_finetune_distributed --config llama2/7B_full \
max_steps_per_epoch=500 epochs=1 log_peak_memory_stats=True \
custom_sharded_layers=['tok_embeddings','output']
```

Screenshot 2024-08-10 at 2 23 19 PM

#### Selective AC

```
tune run --nproc_per_node 2 full_finetune_distributed --config llama3/8B_full \
max_steps_per_epoch=500 epochs=1 log_peak_memory_stats=True ac_mode=selective ac_option=3
...
1|45|Loss: 0.9101344347000122:   9%|████████████████▋                                                                                                                                                                         
```

(Not testing memory or perf here, just making sure the run succeeds, since this is still an experimental feature)

### DIFF
--- a/recipes/full_finetune_distributed.py
+++ b/recipes/full_finetune_distributed.py
@@ -4,12 +4,11 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
-import os
 import sys
 import time
 
 from functools import partial
-from typing import Any, Dict, Optional, Tuple
+from typing import Any, Dict, List, Optional, Tuple
 from warnings import warn
 
 import torch
@@ -17,10 +16,6 @@ from omegaconf import DictConfig, ListConfig
 
 from torch import nn
 from torch.distributed import destroy_process_group, init_process_group
-from torch.distributed._composable.fsdp import CPUOffloadPolicy, fully_shard
-from torch.distributed.algorithms._checkpoint.checkpoint_wrapper import (
-    CheckpointWrapper,
-)
 
 from torch.optim import Optimizer
 from torch.utils.data import DataLoader, DistributedSampler
@@ -189,7 +184,7 @@ class FullFinetuneRecipeDistributed(FTRecipeInterface):
 
     def setup(self, cfg: DictConfig) -> None:
         """
-        Setup the recipe state. This includes recipe state (if resume_from_checkpoint is True),
+        Setup the recipe. This includes training state (if resume_from_checkpoint is True),
         model, tokenizer, loss, optimizer, sampler, and dataloader.
         """
         if self._is_rank_zero:
@@ -203,8 +198,9 @@ class FullFinetuneRecipeDistributed(FTRecipeInterface):
         self._model = self._setup_model(
             cfg_model=cfg.model,
             enable_activation_checkpointing=cfg.enable_activation_checkpointing,
-            memory_efficient_fsdp_wrap=cfg.get("memory_efficient_fsdp_wrap", False),
+            custom_sharded_layers=cfg.get("custom_sharded_layers", None),
             fsdp_cpu_offload=cfg.get("fsdp_cpu_offload", False),
+            reshard_after_forward=cfg.get("reshard_after_forward", True),
             model_state_dict=checkpoint_dict[utils.MODEL_KEY],
             ac_mode=cfg.get("ac_mode", None),
             ac_option=cfg.get("ac_option", None),
@@ -249,8 +245,9 @@ class FullFinetuneRecipeDistributed(FTRecipeInterface):
         self,
         cfg_model: DictConfig,
         enable_activation_checkpointing: bool,
-        memory_efficient_fsdp_wrap: bool,
+        custom_sharded_layers: Optional[List[str]],
         fsdp_cpu_offload: bool,
+        reshard_after_forward: bool,
         model_state_dict: Dict[str, Any],
         ac_mode: Optional[str] = None,
         ac_option: Optional[int] = None,
@@ -261,7 +258,6 @@ class FullFinetuneRecipeDistributed(FTRecipeInterface):
               the right dtype
            b. All ranks calls ``load_state_dict`` without peaking CPU RAMs since
               full state dicts are loaded with ``torch.load(mmap=True)``
-           c. We register (pre-)forward hooks with ``fully_shard`` instead of wrapping `nn.Module`
         """
 
         if self._is_rank_zero:
@@ -279,9 +275,6 @@ class FullFinetuneRecipeDistributed(FTRecipeInterface):
         # ac_mode and ac_option together control selective AC. This is only enabled
         # when these are set AND ``enable_activation_checkpointing`` is set to False
         # We'll clean this up as soon as testing of AC is complete
-        ac_mode = ac_mode
-        ac_option = ac_option
-
         if (not enable_activation_checkpointing) and (ac_mode is not None):
             apply_selective_activation_checkpointing(
                 model,
@@ -294,35 +287,44 @@ class FullFinetuneRecipeDistributed(FTRecipeInterface):
                 model, auto_wrap_policy={modules.TransformerDecoderLayer}
             )
 
-        fsdp_kwargs = {}
-        if fsdp_cpu_offload:
-            fsdp_kwargs["offload_policy"] = CPUOffloadPolicy()
+        # For FSDP sharding, we can condition on either the module or its name
+        # Shard conditions should be callables taking name (relative to model root)
+        # and the module itself and returning a bool on whether to shard the given module
+        fsdp_shard_conditions = []
 
-        # Shard the model with FSDP
-        for m in reversed(list(model.modules())):
-            # For large vocab size, we can additionally shard
-            # the token embeddings and output projections individually
-            if memory_efficient_fsdp_wrap:
-                if isinstance(m, nn.Embedding):
-                    fully_shard(m, **fsdp_kwargs)
-                if isinstance(m, modules.TransformerDecoder):
-                    fully_shard(m.output, **fsdp_kwargs)
-            # TransformerDecoderLayer is wrapped by CheckpointWrapper
-            # when enable_activation_checkpointing
-            if enable_activation_checkpointing:
-                if isinstance(m, CheckpointWrapper):
-                    fully_shard(m, **fsdp_kwargs)
-            else:
-                if isinstance(m, modules.TransformerDecoderLayer):
-                    fully_shard(m, **fsdp_kwargs)
-        fully_shard(model, **fsdp_kwargs)
+        # Shard transformer decoder layers (or AC-wrapped versions)
+        # Alternatively we could condition on the module type (TransformerDecoder or CheckpointWrapper)
+        # But directly using the name is more concise
+        def _is_layer_fqn(s: str) -> bool:
+            """
+            Return True for layers.i and False for all other module names
+            Covers sharding for both AC-wrapped and non-AC-wrapped modules in one shot
+            """
+            s_list = s.split(".")
+            return len(s_list) == 2 and s_list[0] == "layers" and str.isdigit(s_list[1])
+
+        fsdp_shard_conditions = [lambda n, m: _is_layer_fqn(n)]
+
+        # If wrapping any layers separately, we can add another shard condition
+        # A layer will be sharded if any of the fsdp_shard_conditions are met
+        if custom_sharded_layers:
+            fsdp_shard_conditions += [lambda n, m: n in custom_sharded_layers]
+
+        utils.shard_model(
+            model=model,
+            shard_conditions=fsdp_shard_conditions,
+            cpu_offload=fsdp_cpu_offload,
+            reshard_after_forward=reshard_after_forward,
+        )
 
         with utils.set_default_dtype(self._dtype), self._device:
             for m in model.modules():
                 # RoPE is not covered in state dict
-                if isinstance(m, modules.RotaryPositionalEmbeddings):
-                    m.reset_parameters()
+                if hasattr(m, "rope_init"):
+                    m.rope_init()
 
+        # This method will convert the full model state dict into a sharded state
+        # dict and load into the model
         utils.load_from_full_model_state_dict(
             model, model_state_dict, self._device, self._is_rank_zero, strict=True
         )
@@ -582,7 +584,6 @@ def recipe_main(cfg: DictConfig) -> None:
             "Distributed finetune recipe should be run via a distributed launcher."
             "If using tune CLI, please specify --nnodes 1 and --nproc_per_node [num_gpus]"
         )
-    os.environ["TORCH_NCCL_AVOID_RECORD_STREAMS"] = "1"
     init_process_group(backend="gloo" if cfg.device == "cpu" else "nccl")
     if cfg.get("fsdp_cpu_offload", False):
         # Utilize all available CPU cores for intra-op parallelism. This provides ~2x
--- a/recipes/full_finetune_distributed.py
+++ b/recipes/full_finetune_distributed.py
@@ -584,6 +584,10 @@ def recipe_main(cfg: DictConfig) -> None:
         )
     os.environ["TORCH_NCCL_AVOID_RECORD_STREAMS"] = "1"
     init_process_group(backend="gloo" if cfg.device == "cpu" else "nccl")
+    if cfg.get("fsdp_cpu_offload", False):
+        # Utilize all available CPU cores for intra-op parallelism. This provides ~2x
+        # speed up when benchmarking fused AdamW on CPU
+        utils.set_torch_num_threads()
 
     config.log_config(recipe_name="FullFinetuneRecipeDistributed", cfg=cfg)
 
--- a/recipes/full_finetune_distributed.py
+++ b/recipes/full_finetune_distributed.py
@@ -4,6 +4,7 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
+import os
 import sys
 import time
 
@@ -15,24 +16,19 @@ import torch
 from omegaconf import DictConfig, ListConfig
 
 from torch import nn
-from torch.distributed import init_process_group
-from torch.distributed.fsdp import (
-    CPUOffload,
-    FullOptimStateDictConfig,
-    FullStateDictConfig,
-    FullyShardedDataParallel as FSDP,
-    StateDictType,
+from torch.distributed import destroy_process_group, init_process_group
+from torch.distributed._composable.fsdp import fully_shard
+from torch.distributed.algorithms._checkpoint.checkpoint_wrapper import (
+    CheckpointWrapper,
 )
+
 from torch.optim import Optimizer
 from torch.utils.data import DataLoader, DistributedSampler
-
 from torchtune import config, modules, utils
 from torchtune.datasets import ConcatDataset
 from torchtune.recipe_interfaces import FTRecipeInterface
-from torchtune.utils.activations import apply_selective_activation_checkpointing
 
 from tqdm import tqdm
-
 
 log = utils.get_logger("DEBUG")
 
@@ -43,8 +39,8 @@ class FullFinetuneRecipeDistributed(FTRecipeInterface):
     distributed training and can be run on a single node (1 to 8 GPUs).
 
     Features:
-        - FSDP. Supported using PyTorch's FSDP APIs. DDP is currently not supported. Training on CPU
-            is not supported.
+        - FSDP. Supported using PyTorch's FSDP APIs. DDP is currently not supported. Traning on CPU is not
+            supported.
 
         - Activation Checkpointing. This can be controlled using the ``activation_checkpointing``
             flag. Activation checkpointing helps reduce the memory footprint since we no longer keep
@@ -73,14 +69,18 @@ class FullFinetuneRecipeDistributed(FTRecipeInterface):
             checkpointing.
 
         - Checkpointing. Model weights are checkpointed both at the end of each epoch and at the end of
-            training. Optimizer state and recipe state (seed, total_epochs, number of epochs run etc) are
-            only saved at the end of a given epoch and used in case of resuming training.
+            training. Currently we checkpoint both the adapter weights (trainable params only) and the
+            complete merged weights (adapter weights added back to the base model). For more details
+            please take a look at our LoRA tutorial
+            (https://pytorch.org/torchtune/main/tutorials/lora_finetune.html).
 
-            Resuming training is controlled by the ``resume_from_checkpoint`` flag. Mid-epoch checkpointing is
+            Optimizer State and recipe state (seed, total_epochs, number of epochs run etc) are
+            only saved at the end of a given epoch and used in case of resuming training. Resuming
+            training is controlled by the ``resume_from_checkpoint`` flag. Mid-epoch checkpointing is
             currently not supported.
 
             For more details on the checkpointer, please take a look at
-            our checkpointer deepdive (https://pytorch.org/torchtune/main/deep_dives/checkpointer.html).
+            our checkpointer deepdive (https://pytorch.org/torchtune/main/tutorials/checkpointer.html).
 
         - Logging. Terminal, Disk, WandB and TensorBoard are all supported.
 
@@ -92,10 +92,11 @@ class FullFinetuneRecipeDistributed(FTRecipeInterface):
 
     Raises:
         ValueError: If ``dtype`` is set to fp16.
+        ValueError: If world_size is 1
+        RuntimeError: If ``dtype`` is set to bf16 and the hardware does not support bf16.
     """
 
     def __init__(self, cfg: DictConfig) -> None:
-
         self._device = utils.get_device(device=cfg.device)
         self._dtype = utils.get_dtype(cfg.dtype, device=self._device)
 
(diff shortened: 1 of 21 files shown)


recipes/full_finetune_distributed.py
```diff
@@ -590,12 +582,8 @@ def recipe_main(cfg: DictConfig) -> None:
             "Distributed finetune recipe should be run via a distributed launcher."
             "If using tune CLI, please specify --nnodes 1 and --nproc_per_node [num_gpus]"
         )
-
+    os.environ["TORCH_NCCL_AVOID_RECORD_STREAMS"] = "1"
     init_process_group(backend="gloo" if cfg.device == "cpu" else "nccl")
-    if cfg.get("fsdp_cpu_offload", False):
-        # Utilize all available CPU cores for intra-op parallelism. This provides ~2x
-        # speed up when benchmarking fused AdamW on CPU
-        utils.set_torch_num_threads()
```
weifengpy, line 598: setting num of threads is a critical lesson we learned in cpu offloading. the perf indication is huge. curious why removing it ? do not need cpu offloading anymore?

weifengpy (commented):
thanks for driving this! The code looks good. Happy to see the similarity between lora and full finetuning and many of the code can be reused

I did notice the perf difference in snapshot (wandb). would love to learn more from profilers traces and compare FSDP1 and FSDP2 side by side

I did not stamp only because this PR seems to for discussion purpose

recipes/full_finetune_distributed.py
ebsmothers: Ah thanks, yes this should stay in. Will update

ebsmothers:
Thanks @weifengpy for the review! We intend to land this (hopefully soon), mainly I left the PR in draft status because there is some additional testing I'd like to do. Re the perf difference, this run is for Llama3 which does some slightly custom sharding. Let me re-run on Llama2 and see if the results persist. The suggestion to profile to understand the difference makes sense to me -- while I'm at it I may even integrate the profiler from our LoRA recipes into full finetune (since this is something folks have been asking about anyways)

recipes/full_finetune_distributed.py
```diff
@@ -314,41 +289,51 @@ def _setup_model(
...
+                    fully_shard(m, **fsdp_kwargs)
+            # For large vocab size, we can additionally shard
+            # the token embeddings and output projections individually
+            if memory_efficient_fsdp_wrap:
+                if isinstance(m, nn.Embedding):
+                    fully_shard(m, **fsdp_kwargs)
+                if isinstance(m, modules.TransformerDecoder):
+                    fully_shard(m.output, **fsdp_kwargs)
+            else:
+                if isinstance(m, modules.TransformerDecoderLayer):
```
felipemello1, line 316: n00b question: why else?

Should it be:
```
if isinstance(m, modules.TransformerDecoderLayer):
    fully_shard(m, **fsdp_kwargs)
if memory_efficient_fsdp_wrap:
    if isinstance(m, nn.Embedding):
        fully_shard(m, **fsdp_kwargs)
    if isinstance(m, modules.TransformerDecoder):
        fully_shard(m.output, **fsdp_kwargs)

File: recipes/full_finetune_distributed.py [hunk 3]
```diff
@@ -425,32 +402,48 @@ def _setup_data(
...
+    def save_checkpoint(
+        self,
+        epoch: int,
+    ) -> None:
         """
-        Save state dict to file. The recipe save_checkpoint method is responsible for
-        correctly creating the checkpoint dict and passing to the checkpointer.
+        Checkpoint the state of the recipe. The constructed checkpoint state dict
+        contains the following information:
+        - Model weights with key MODEL_KEY
```
felipemello1, line 412:
```suggestion
        - Model weights with key utils.MODEL_KEY
```

recipes/full_finetune_distributed.py
ebsmothers: Ah I need to update this, the memory efficient wrapping logic should be disentangled from the AC logic. We want to wrap the transformer decoder layer. If activation checkpointing is enabled this is already wrapped in `CheckpointWrapper` so we check that condition. Otherwise we wrap `TransformerDecoderLayer` directly

recipes/full_finetune_distributed.py
```diff
@@ -314,41 +289,51 @@ def _setup_model(
...
-
-        # original activation checkpointing (full) - flip the condition above
         if enable_activation_checkpointing and ac_mode is None:
             utils.set_activation_checkpointing(
                 model, auto_wrap_policy={modules.TransformerDecoderLayer}
             )
 
+        fsdp_kwargs = {}
+        if fsdp_cpu_offload:
+            fsdp_kwargs["offload_policy"] = CPUOffloadPolicy()
```
felipemello1, line 299: should we expose the shard strategy like in https://github.com/pytorch/torchtune/pull/1024

ebsmothers: We can, but imo at this stage it's a nice-to-have rather than a must-have

weifengpy:
> We intend to land this (hopefully soon), mainly I left the PR in draft status because there is some additional testing I'd like to do. 

sounds good. feel free to request another review when it's ready

> while I'm at it I may even integrate the profiler from our LoRA recipes into full finetune (since this is something folks have been asking about anyways)

great! I am happy to take a deeper look at the trace once we have it. hopefully to gain new knowledge to make FSDP2 better

recipes/full_finetune_distributed.py
pbontrager: It does feel wrong to remove a feature that was just recently contributed though. It feels the goal of moving this out of dev is to integrate it with everything we already have.

recipes/full_finetune_distributed.py
```diff
@@ -590,7 +582,7 @@ def recipe_main(cfg: DictConfig) -> None:
             "Distributed finetune recipe should be run via a distributed launcher."
             "If using tune CLI, please specify --nnodes 1 and --nproc_per_node [num_gpus]"
         )
-
+    os.environ["TORCH_NCCL_AVOID_RECORD_STREAMS"] = "1"
     init_process_group(backend="gloo" if cfg.device == "cpu" else "nccl")
```
pbontrager, line 586: I know this is old, but technically I think we only want nccl if device is cuda. The default for init_process_group just does both which works really well.

recipes/full_finetune_distributed.py
```diff
@@ -590,7 +582,7 @@ def recipe_main(cfg: DictConfig) -> None:
             "Distributed finetune recipe should be run via a distributed launcher."
             "If using tune CLI, please specify --nnodes 1 and --nproc_per_node [num_gpus]"
         )
-
+    os.environ["TORCH_NCCL_AVOID_RECORD_STREAMS"] = "1"
```
pbontrager, line 585: Can you explain what this does?

pbontrager (commented):
Thanks for this update! I just added some comments and general questions as this is my first time really looking at FSDP2.

recipes/full_finetune_distributed.py
```diff
@@ -192,37 +189,31 @@ def _update_recipe_state(self, ckpt_dict: Dict[str, Any]) -> None:
...
+        model, tokenizer, loss, optimizer, sampler, and dataloader.
         """
         if self._is_rank_zero:
             self._metric_logger = config.instantiate(cfg.metric_logger)
 
             # log config with parameter override
             self._metric_logger.log_config(cfg)
 
-        ckpt_dict = self.load_checkpoint(cfg.checkpointer)
+        checkpoint_dict = self.load_checkpoint(cfg_checkpointer=cfg.checkpointer)
```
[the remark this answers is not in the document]
ebsmothers: > Why aren't we calling this state_dict?

To me state_dict == model weights only, but this can also include recipe state. But no strong preference here tbh

recipes/full_finetune_distributed.py
ebsmothers: It hasn't landed yet though

recipes/full_finetune_distributed.py
```diff
@@ -314,41 +289,51 @@ def _setup_model(
...
             utils.set_activation_checkpointing(
                 model, auto_wrap_policy={modules.TransformerDecoderLayer}
             )
 
+        fsdp_kwargs = {}
+        if fsdp_cpu_offload:
+            fsdp_kwargs["offload_policy"] = CPUOffloadPolicy()
+
+        # Shard the model with FSDP
+        for m in reversed(list(model.modules())):
```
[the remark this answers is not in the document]
ebsmothers: Yeah I think we should have utils. And to be clear, we are not actually doing any AC wrapping here, that happens above. Here we are just checking which modules have been wrapped by AC already.

recipes/full_finetune_distributed.py
```diff
@@ -314,41 +289,51 @@ def _setup_model(
...
+            # For large vocab size, we can additionally shard
+            # the token embeddings and output projections individually
+            if memory_efficient_fsdp_wrap:
```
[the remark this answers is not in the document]
ebsmothers: That's not what this is though. This is just for the case where we want to shard token embeddings and output projections in their own unit (instead of just in the top-level model sharding). We found this saves a fair bit of memory for Llama3. I do agree the naming is a bit awkward, definitely open to other suggestions here

recipes/full_finetune_distributed.py
ebsmothers: Oh yeah this is a good question. This was originally used to save memory when we were using FSDP with `sync_module_states=True` (see #843). I suspect it shouldn't be needed with our current FSDP2 usage but can also run a quick sanity check here. @weifengpy can you confirm whether we hit `dist._broadcast_coalesced` in any part of `fully_shard`?

recipes/full_finetune_distributed.py
ebsmothers: What other devices do we support right now though?

> The default for init_process_group just does both which works really well

Not sure I understand this.. do you mean we should just be calling `init_process_group()`?

recipes/full_finetune_distributed.py
```diff
@@ -314,41 +289,51 @@ def _setup_model(
...
+                if isinstance(m, nn.Embedding):
+                    fully_shard(m, **fsdp_kwargs)
+                if isinstance(m, modules.TransformerDecoder):
+                    fully_shard(m.output, **fsdp_kwargs)
+            # TransformerDecoderLayer is wrapped by CheckpointWrapper
+            # when enable_activation_checkpointing
```
kartikayk, line 311:
```suggestion
            # when enable_activation_checkpointing is set to True
```

recipes/full_finetune_distributed.py
```diff
@@ -314,41 +289,51 @@ def _setup_model(
...
+            else:
+                if isinstance(m, modules.TransformerDecoderLayer):
+                    fully_shard(m, **fsdp_kwargs)
+        fully_shard(model, **fsdp_kwargs)
+
+        with utils.set_default_dtype(self._dtype), self._device:
+            for m in model.modules():
+                # RoPE is not covered in state dict
+                if isinstance(m, modules.RotaryPositionalEmbeddings):
+                    m.reset_parameters()
```
kartikayk, line 324: How should I conceptually parse this block? IIUC, at this stage all of the model params are init on meta device (```load_from_full_model_state_dict``` is where we bring them over to GPU unless I'm misremembering). So at this point we're constructing the RoPE buffers on CUDA? Is there a reason this is happening before we load the state dict? Or doesn't matter? 

Also, we called this ```reset_parameters``` because of an idiosyncrasy with FSDP1. Do we need to carry that over? Can we just make ```rope_init``` a public API? It's really hard to reason about ```reset_parameters``` if you don't know the details.

recipes/full_finetune_distributed.py
```diff
@@ -314,41 +289,51 @@ def _setup_model(
...
+
+        utils.load_from_full_model_state_dict(
```
kartikayk, line 326: Add a brief comment here saying what this is actually doing - helps make sure this function is self explanatory which is how we intended it to be

recipes/full_finetune_distributed.py
```diff
@@ -360,17 +345,13 @@ def _setup_model(
...
-        """
-        Set up the optimizer. This method also handles transforing the state dict
-        for FSDP.
-        """
         optimizer = config.instantiate(cfg_optimizer, self._model.parameters())
-
         if opt_state_dict:
-            opt_state_dict = FSDP.optim_state_dict_to_load(
-                self._model, optimizer, opt_state_dict
+            utils.load_from_full_optimizer_state_dict(
```
kartikayk, line 350: @weifengpy IIRC there were some APIs related to optimizer and state_dict that were in the process of being moved over to distributed? Is that still the plan? If so, we can just import these from distributed instead of from utils?

kartikayk (commented):
Do the recipe tests need any updates? I'd guess not? Can you also test checkpoint reads and writes and make sure memory looks fine?

recipes/full_finetune_distributed.py
```diff
@@ -266,37 +257,21 @@ def _setup_model(
...
-               model weights from checkpoint.
-            c. While wrapping the model with FSDP, we set ``sync_module_states``
-               to TRUE and broadcast module params and buffers from rank 0.
-            d. The ``device_id`` param ensures that the FSDP initialization happens on
-               the correct device.
+           a. To minimize GPU peak memory, we initialize the model on meta device with
+              the right dtype
+           b. All ranks calls ``load_state_dict`` without peaking CPU RAMs since
+              full state dicts are loaded with ``torch.load(mmap=True)``
+           c. We register (pre-)forward hooks with ``fully_shard`` instead of wrapping `nn.Module`
```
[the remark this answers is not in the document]
ebsmothers: Yeah these are both fair points, lemme update

recipes/full_finetune_distributed.py
weifengpy: We can remove setting TORCH_NCCL_AVOID_RECORD_STREAMS in FSDP2 because we get rid of recordstream. `dist._broadcast_coalesced ` was from state dict loading for FSDP1 and it's gone as well

recipes/full_finetune_distributed.py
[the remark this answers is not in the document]
ebsmothers: So a couple thoughts here:

(1) This is an optimization that we only apply to Llama 3, but it should be generally useful for any model with a very large vocab size. Ofc I agree that as we've named it none of these points are obvious.. my initial intention here was to be as non-BC-breaking as possible, but if you guys want a rename of this I am happy to do it.

(2) On passing modules we want to wrap separately as a string: I think this is an interesting idea -- if we iterate over named modules we could probably just do a check on the FQN? The only concern I have is that it could be a bit easy for someone to shoot themselves in the foot here.. we are giving a lot more freedom over configurations that we have not really tried out ourselves.

Either way I do plan to move this into a separate utility. I'm still on the fence about how exactly to expose the individual wrapping of these layers, will try a couple approaches and see what makes sense here.

recipes/full_finetune_distributed.py
weifengpy: I worked with @mori360 last month to upstream the implementation into PTD API. Numerically it's on-par now, but we need to verify memory behavior before formal migration https://github.com/pytorch/pytorch/issues/12874

recipes/full_finetune_distributed.py
ebsmothers: > Is there a reason this is happening before we load the state dict? Or doesn't matter?

It shouldn't matter because the RoPE buffers are not in the state dict anyways. This becomes more relevant for LoRA though, where we maybe have existing adapter weights to load. In that case, we either move from meta device on state dict load or on the usual param init, but not both. But even in this case since we only do one or the other I think the order shouldn't matter (lmk if this makes sense or not though)

> Can we just make rope_init a public API?

Yeah I think we can do this

recipes/full_finetune_distributed.py
```diff
@@ -314,41 +289,51 @@ def _setup_model(
...
         if enable_activation_checkpointing and ac_mode is None:
             utils.set_activation_checkpointing(
                 model, auto_wrap_policy={modules.TransformerDecoderLayer}
             )
 
+        fsdp_kwargs = {}
+        if fsdp_cpu_offload:
+            fsdp_kwargs["offload_policy"] = CPUOffloadPolicy()
+
+        # Shard the model with FSDP
```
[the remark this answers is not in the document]
ebsmothers: Yeah this is the top thing on my plate to refactor here

recipes/full_finetune_distributed.py
```diff
@@ -15,25 +16,21 @@
...
-from torch.distributed import init_process_group
-from torch.distributed.fsdp import (
-    CPUOffload,
-    FullOptimStateDictConfig,
-    FullStateDictConfig,
-    FullyShardedDataParallel as FSDP,
-    StateDictType,
+from torch.distributed import destroy_process_group, init_process_group
+from torch.distributed._composable.fsdp import CPUOffloadPolicy, fully_shard
+from torch.distributed.algorithms._checkpoint.checkpoint_wrapper import (
```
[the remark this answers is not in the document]
ebsmothers: If we're thinking about the same thing, it may have been the suggestion to move to [torch.utils.checkpoint](https://pytorch.org/docs/stable/checkpoint.html) instead of relying on distributed APIs for AC. If it's all the same to you I would punt on that here

recipes/full_finetune_distributed.py
```diff
@@ -192,37 +189,31 @@ def _update_recipe_state(self, ckpt_dict: Dict[str, Any]) -> None:
...
-        # ``_setup_model`` handles initialization and loading the state dict. This method
-        # should be called before ``_setup_optimizer`` since transforming the optimizer
-        # state dict requires the model
         self._model = self._setup_model(
             cfg_model=cfg.model,
             enable_activation_checkpointing=cfg.enable_activation_checkpointing,
             memory_efficient_fsdp_wrap=cfg.get("memory_efficient_fsdp_wrap", False),
             fsdp_cpu_offload=cfg.get("fsdp_cpu_offload", False),
-            model_state_dict=ckpt_dict[utils.MODEL_KEY],
+            model_state_dict=checkpoint_dict[utils.MODEL_KEY],
```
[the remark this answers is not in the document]
ebsmothers: Yes I think we do still need some version of it, otherwise our peak memory will be higher on Llama3 models

recipes/full_finetune_distributed.py
```diff
@@ -266,37 +257,21 @@ def _setup_model(
...
-            log.info("FSDP is enabled. Instantiating Model on CPU for Rank 0 ...")
-            init_start = time.perf_counter()
-
-            with utils.set_default_dtype(self._dtype):
-                model = config.instantiate(cfg_model)
 
+        if self._is_rank_zero:
             log.info(
-                f"Model instantiation took {time.perf_counter() - init_start:.2f} secs"
+                "FSDP is enabled. Instantiating model and loading checkpoint on Rank 0 ..."
```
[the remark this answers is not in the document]
ebsmothers: Sorry not sure I fully understand this question. The `_is_rank_zero` here is just preventing log spew, and the `init_process_group` logic remains unchanged from the existing version of the recipe

recipes/full_finetune_distributed.py
```diff
@@ -314,41 +289,51 @@ def _setup_model(
...
-        # Ensure no params and buffers are on meta device
-        utils.validate_no_params_on_meta_device(model)
-
-        # original activation checkpointing (full) - flip the condition above
         if enable_activation_checkpointing and ac_mode is None:
             utils.set_activation_checkpointing(
                 model, auto_wrap_policy={modules.TransformerDecoderLayer}
             )
 
+        fsdp_kwargs = {}
```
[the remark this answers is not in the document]
ebsmothers: Oh yeah I can see how this is confusing. A bit awkward but auto_wrap_policy for AC != auto_wrap_policy from our dear old friend FSDP. You can see the usage [here](https://github.com/pytorch/torchtune/blob/0a407712eda252573326074d33af0a66c2d2990e/torchtune/utils/memory.py#L42-L43)

recipes/full_finetune_distributed.py
```diff
@@ -266,37 +257,21 @@ def _setup_model(
...
-               we initialize on meta_device for all ranks other than rank 0.
-            b. Rank 0 is also responsible for calling ``load_state_dict`` and loading the
-               model weights from checkpoint.
-            c. While wrapping the model with FSDP, we set ``sync_module_states``
-               to TRUE and broadcast module params and buffers from rank 0.
-            d. The ``device_id`` param ensures that the FSDP initialization happens on
-               the correct device.
+           a. To minimize GPU peak memory, we initialize the model on meta device with
+              the right dtype
+           b. All ranks calls ``load_state_dict`` without peaking CPU RAMs since
```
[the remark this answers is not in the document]
ebsmothers: Because we shard as we load. Previously we only loaded on rank 0 because we relied on sync_module_states to broadcast to all the other ranks. I think this way should be more memory-efficient because we never need a full copy of the state dict on rank 0

ebsmothers:
> Do the recipe tests need any updates? I'd guess not? Can you also test checkpoint reads and writes and make sure memory looks fine?

Yeah recipe tests are all passing. Will update with tests for checkpoints and memory results

recipes/full_finetune_distributed.py
ebsmothers: To follow up on this, there is not a 1:1 mapping between FSDP1 sharding strategy and FSDP2 APIs. But 

(1) for single-node training (all we currently support), the only relevant sharding strategies should be FULL_SHARD, SHARD_GRAD_OP, and NO_SHARD. HYBRID_SHARD and _HYBRID_SHARD_ZERO2 are only distinct from these in the case of multi-node training.
(2) FSDP2 does not support NO_SHARD, in that case they just redirect users to DDP

So then we are just down to FULL_SHARD and SHARD_GRAD_OP, which correspond to setting [reshard_after_forward](https://github.com/pytorch/pytorch/blob/345bea01dc69d9c31a4954c299767ccecd702d3d/torch/distributed/_composable/fsdp/fully_shard.py#L33) to True and False, respectively. So I can expose this flag in the recipe.

Thanks to @weifengpy for helping to clear this up!

kartikayk (approved):

Issue #1246, opened 2024-07-30
Llama3 qlora load_state_dict takes forever


l-berg:
When working on my customized LoRAFinetuneRecipeSingleDevice recipe and upgrading from `torchtune` version 0.1.1 to 0.2.1 and `torchao` 0.1 to 0.3.1, I noticed that model loading times went up dramatically when using QLoRA. Now, loading llama3-8b takes about 5 minutes, where it used to only be a few seconds in version 0.1.1. I was able to pinpoint it to the call `model.load_state_dict(base_model_state_dict, strict=False)`.

Here are the steps I took to reproduce this issue in a new conda environment (pytorch version 2.4).
```bash
conda create --name tune python=3.11
conda install pytorch torchvision torchaudio pytorch-cuda=12.4 -c pytorch -c nvidia
conda activate tune
pip install torchtune
tune download meta-llama/Meta-Llama-3-8B-Instruct --output-dir . --hf-token <token>
tune run lora_finetune_single_device --config llama3/8B_qlora_single_device checkpointer.checkpoint_dir=$CHECKPOINT_DIR tokenizer.path=$CHECKPOINT_DIR/tokenizer.model checkpointer.output_dir=$CHECKPOINT_DIR output_dir=$CHECKPOINT_DIR
```
outputs
```
...
DEBUG:torchtune.utils.logging:Setting manual seed to local seed 4061266822. Local seed is seed + rank = 4061266822 + 0
Writing logs to /.../testtune/instruct/original/log_1722353553.txt
<WAITING FOR 5 MINUTES, RAM usage slowly rising to 15GB>
INFO:torchtune.utils.logging:Model is initialized with precision torch.bfloat16.
INFO:torchtune.utils.logging:Memory stats after model init:
        GPU peak memory allocation: 6.79 GB
        GPU peak memory reserved: 6.97 GB
        GPU peak memory active: 6.79 GB
...
```

When I downgrade to version 0.1.1, it's fast again:
```bash
pip install torchtune==0.1.1
tune run lora_finetune_single_device --config llama3/8B_qlora_single_device checkpointer.checkpoint_dir=$CHECKPOINT_DIR tokenizer.path=$CHECKPOINT_DIR/tokenizer.model checkpointer.output_dir=$CHECKPOINT_DIR output_dir=$CHECKPOINT_DIR
```
outputs
```
[...]

PR #1294, opened 2024-08-09
Delete torchtune's inplace copy definition for NF4

ebsmothers:
Thanks to @msaroufim for root causing this tricky bug. This addresses https://github.com/pytorch/ao/issues/642 on our end as well as #1246.

**Update**: Instead of reenabling our inplace copy, @gau-nernst has found the upstream fix in torchao (described [here](https://github.com/pytorch/ao/issues/642#issuecomment-2306089086)). With [this fix](https://github.com/pytorch/ao/pull/737), we can fully delete the inplace copy from torchtune. Until we're on the ao version with the fix our QLoRA will continue to be slow, but after that things should be back to normal.

###Stuff below this line is outdated but kept as background###

We were defining our own override of aten copy op in torchtune prior to torchao version 0.2, when they added their own. After that we version-gated this override to ao < 0.2. But it turns out that the version we have is faster. Tbh idk why that is yet, but it should be safe to re-enable this to get us back to our previous perf on QLoRA state dict load. 

Tested by logging time from recipe startup to the beginning of the train loop.

On main: `init time: 101.48360734200105`
On this PR: `init time: 15.445241551031359`


torchtune/modules/low_precision/_register_nf4_dispatch_ops.py
```diff
@@ -6,7 +6,6 @@
 
 import torch
 from torchao.dtypes.nf4tensor import implements as nf4_tensor_impl, to_nf4
-from torchtune.modules.low_precision._utils import _get_torchao_version
```
ebsmothers, line 9: Strictly speaking this util is now no longer used anywhere in our library, but I am inclined to keep it in for possible (likely?) future usage

ebsmothers:
@weifengpy seems we may need to add an override for `aten.to.dtype_layout` to handle [this line](https://github.com/pytorch/pytorch/blob/da65cfbdea4f1f2176f6242004bda940a24f9ddb/torch/distributed/_composable/fsdp/_fsdp_param.py#L344) when creating FSDP params. Please lmk if I have the right understanding here

weifengpy:
> @weifengpy seems we may need to add an override for `aten.to.dtype_layout` to handle [this line](https://github.com/pytorch/pytorch/blob/da65cfbdea4f1f2176f6242004bda940a24f9ddb/torch/distributed/_composable/fsdp/_fsdp_param.py#L344) when creating FSDP params. Please lmk if I have the right understanding here

yes, the error msg looks relevant. but I do not fully understand why torchtune implementation dispatch aten.to.dtype_layout but torchao’s implementation does not need it

FSDP2 needs to support around 9 tensor ops but aten.to.dtype_layout is totally new

torchtune/modules/low_precision/_register_nf4_dispatch_ops.py
joecummings: Nope - get rid of it until we need it again. It's easy enough to find in git history.

joecummings (approved):

weifengpy:
I see the PR is accepted. curious how did we resolve the local_tensor=nan issue?

msaroufim (approved):
very nice! glad the version check code is also gone 💀

Issue #1228, opened 2024-07-26
[FEATURE REQUEST] Compile model+loss_fn together


gau-nernst:
Right now when `compile=True`, only the model is compiled

https://github.com/pytorch/torchtune/blob/e10142016798cf84f2e5c638a985014384f400a7/recipes/lora_finetune_single_device.py#L383-L386

We can further boost performance by including loss calculations in compile step.

From my benchmarks, the improvement is pretty significant. Running `tune run lora_finetune_single_device --config llama3/8B_qlora_single_device compile=True` on 4080, the improvements is ~1100 tok/s -> ~1450 tok/s (30% improvement)

Baseline - main e101420

```
Step 1 | loss:1.9794920682907104 lr:2.9999999999999997e-06 tokens_per_second_per_gpu:36.1405614689982 
Step 2 | loss:2.132385015487671 lr:5.999999999999999e-06 tokens_per_second_per_gpu:1057.5127828685158 
Step 3 | loss:2.0779435634613037 lr:8.999999999999999e-06 tokens_per_second_per_gpu:1107.9911968768445 
Step 4 | loss:2.107637643814087 lr:1.1999999999999999e-05 tokens_per_second_per_gpu:1050.867526697978 
Step 5 | loss:1.9378103017807007 lr:1.4999999999999999e-05 tokens_per_second_per_gpu:1122.8913807208219 
Step 6 | loss:2.0037074089050293 lr:1.7999999999999997e-05 tokens_per_second_per_gpu:1114.3397000842317 
Step 7 | loss:2.049872636795044 lr:2.1e-05 tokens_per_second_per_gpu:1108.8569776038305 
Step 8 | loss:2.1449005603790283 lr:2.3999999999999997e-05 tokens_per_second_per_gpu:1027.7933852631434 
Step 9 | loss:1.9285515546798706 lr:2.6999999999999996e-05 tokens_per_second_per_gpu:1113.8025743191083 
Step 10 | loss:2.0167834758758545 lr:2.9999999999999997e-05 tokens_per_second_per_gpu:1094.9996764298014 
Step 11 | loss:1.9725338220596313 lr:3.2999999999999996e-05 tokens_per_second_per_gpu:975.2314694129668 
Step 12 | loss:1.974663496017456 lr:3.5999999999999994e-05 tokens_per_second_per_gpu:1044.7823733103148 
[...]

PR #1296, opened 2024-08-09
Compile model+loss for LoRA single device recipe

gau-nernst:
#### Context
- [x] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

Address #1228

To limit the scope of this PR, I will only update the LoRA single device recipe. We can keep #1228 open and discuss how to do model+loss compile for other recipes.

#### Changelog

Compile model+loss together for LoRA single device recipe. This has shown substantial speed improvements.

#### Test plan

Run the following command on 4070Ti SUPER (16GB VRAM)

>`tune run lora_finetune_single_device --config llama3/8B_qlora_single_device compile=True metric_logger._component_=torchtune.utils.metric_logging.WandBLogger batch_size=4 gradient_accumulation_steps=1 log_peak_memory_stats=True`

image

15% improvement in tok/s on average, generally lower memory consumption.


recipes/lora_finetune_single_device.py
```diff
@@ -557,31 +576,10 @@ def train(self) -> None:
...
-                    logits = self._model(tokens, mask=mask, input_pos=input_pos)
-                    # Shift so that tokens < n predict n
-                    logits = logits[..., :-1, :].contiguous()
-                    labels = labels[..., 1:].contiguous()
-                    logits = logits.transpose(1, 2)
-                    # Compute loss
-                    loss = self._loss_fn(logits, labels)
-                    # free logits otherwise it peaks backward memory
-                    del logits
+                    batch = {k: v.to(self._device) for k, v in batch.items()}
```
felipemello1, line 579: My concern is that the batch may contain keys that may not be necessary in the gpu. Can we keep the .to(device) like it was before? Or did you have some other motivation for it?

gau-nernst: I tried putting the old `.to(device)` code within `self._loss_step()`, but `torch.compile()` shows some warnings. Should be harmless though. I will double-check it again.

Another option is to keep the old `.to(device)` outside `self._loss_step()`. Then, it is a bit more clunky to pass `tokens`, `labels`, `mask`, `input_pos` to `self._loss_step()` instead of just `batch` (or we can just modify the `batch` object in-place). Lmk which one you prefer, then I will just change to that.

felipemello1: i think it should be fine to keep as it is then. I dont think that our batches output extra terms, so that scenario I raised would be an exception.

felipemello1:
Before i approve it, would it make sense to add it to the other recipes, so they are all aligned? For example, should we test it in distributed? We probably want to avoid having different versions of loss_step for each recipe.

out of curiosity: Did you notice any major difference in compiling time?

gau-nernst:
> would it make sense to add it to the other recipes, so they are all aligned?

I think we can add this to other recipes in separate PR, since each requires good testing to confirm no regression and expected speed up. For example, even for this recipe (single device QLoRA), took me 10min+ for each mini run with Lllama3-8B (well most of the time is spent on quantization due to #1246 and compilation).

> should we test it in distributed

I don't have a multi-gpu setup, so I can't run and test it myself 😅.

> We probably want to avoid having different versions of loss_step for each recipe.

I think the point is that `self._loss_step()` will be different for each recipe? so that each recipe can have their custom logic. But yea at least among the ones that trains on autogressive modelling loss (not DPO/PPO), their `self._loss_step()` should be the same.

> Did you notice any major difference in compiling time?

Not on my machine. I also checked it myself. Didn't put a timer, but this wandb graph suggests they have the same compile time (power usage goes up at the same time -> training starts at the same time)

image

Another question. Realised that I deleted this line of yours

https://github.com/pytorch/torchtune/blob/c1fabb958873f5356133e897b66b889a6477de57/recipes/full_finetune_single_device.py#L523

Since now we compile model+loss together, I don't think this will make a difference for `compile=True` anymore. But probably still has an impact in eager mode (`compile=False`). I will add it back. Is it ok with you?

felipemello1:
sounds good! gonna let the tests run and approve the PR. Thanks for addressing my questions

felipemello1 (approved):

PR #1304, opened 2024-08-10
fix duplicate logging

### LINKED ISSUE #1302: Duplicate logging
Should be relatively quick to debug but I may not get to it today.. seems like our logging is happening multiple times.

For example if I run
```
```

the first part of my logs look like [this](https://gist.github.com/ebsmothers/b9d3a4b5254075e99540d71b845d3747).

My env is PyTorch + TorchVision stable install and dev install on torchtune main.

ebsmothers:
Fixes #1302 

Tbh I'm not that familiar with the logging module so not sure why this is happening. But our other logs inside modules/ use logging's getLogger. Also **using anything from utils/ inside modules/** creates a circular dependency. Though it seems to resolve somehow (again not familiar with how or why), it also seems to cause our logger to get created twice.

Not to get up on my soapbox here, but **utils/ is a terrible name for our training utilities directory**. As it stands today, our training utilities have dependencies on both our modules and data directories. We should really either

(1) remove these dependencies (the data one is easy, the modules one requires re-thinking both our distributed and generation utilities, which is probably nontrivial), or
(2) rename this directory to something that is clearly training-related, so that we we can have an actual utils directory taking no other dependencies from within our repo (so that they can then be used wherever we want)

Or we can just keep using two separate loggers 😃

### DIFF
--- a/torchtune/modules/attention.py
+++ b/torchtune/modules/attention.py
@@ -4,14 +4,14 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
+import logging
 from typing import Optional
 
 import torch
 from torch import nn, Tensor
 from torchtune.modules.kv_cache import KVCache
-from torchtune.utils.logging import get_logger
 
-logger = get_logger("DEBUG")
+logger = logging.getLogger(__name__)
 
 
 class CausalSelfAttention(nn.Module):


joecummings (approved):

joecummings:
I would honestly prefer 2 but with a mix of 1.

pbontrager:
I don’t see what you are saying is a circular dependency. The python logger module is meant to act as a singleton if you initialize it with __name__ which we do. The reason for our util is to have a single stream for all our logs, even in a distributed run. This is a feature not a bug, you aren’t losing logs from distributed processes. This PR is just hiding the logs from other ranks, the correct solution is to only log on rank 0 something that you only want logged on rank 0. This needs to be reverted. @ebsmothers

PR #1305, opened 2024-08-11
Fix eval in regression test

ebsmothers:
The config name and the results parsing in our regression test job are incorrect.

This is actually a bit awkward to test now that we (a) don't allow creating PRs from a fork, and (b) don't let forks access the S3 bucket containing regression test artifacts.

So for now I've tested it locally, which I guess is better than nothing?

```
pytest tests/regression_tests/test_llama2_7b.py -m slow_integration_test
...
====== 1 passed, 1 warning in 267.94s (0:04:27) ======
```

### DIFF
--- a/tests/regression_tests/test_llama2_7b.py
+++ b/tests/regression_tests/test_llama2_7b.py
@@ -27,7 +27,7 @@ EVAL_CONFIG_PATH = Path.joinpath(
 @gpu_test(gpu_count=2)
 class TestLoRA7BDistributedFinetuneEval:
     @pytest.mark.slow_integration_test
-    def test_finetune_and_eval(self, tmpdir, caplog, monkeypatch):
+    def test_finetune_and_eval(self, tmpdir, capsys, monkeypatch):
 
         ckpt_path = Path(CKPT_MODEL_PATHS[CKPT])
         ckpt_dir = ckpt_path.parent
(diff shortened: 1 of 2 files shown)


tests/regression_tests/test_llama2_7b.py
```diff
@@ -50,7 +50,7 @@ def test_finetune_and_eval(self, tmpdir, caplog, monkeypatch):
         runpy.run_path(TUNE_PATH, run_name="__main__")
         eval_cmd = f"""
         tune run eleuther_eval \
-            --config eleuther_eval \
+            --config eleuther_evaluation \
```
SalmanMohammadi, line 53: How was this test working before??

SalmanMohammadi:
> This is actually a bit awkward to test now that we (a) don't allow creating PRs from a fork, and (b) don't let forks access the S3 bucket containing regression test artifacts.

I missed this, but we're no longer running integration tests in our CI?

tests/regression_tests/test_llama2_7b.py
ebsmothers: It wasn’t, it’s been failing in CI for a while now. But it doesn’t run on PRs or anything, only nightly

ebsmothers:
> > This is actually a bit awkward to test now that we (a) don't allow creating PRs from a fork, and (b) don't let forks access the S3 bucket containing regression test artifacts.
> 
> I missed this, but we're no longer running integration tests in our CI?

So we actually have two types of integration tests: recipe tests and regression tests. Recipe tests always run on PRs but only use small checkpoints. Regression tests use the full-size model and run nightly. Currently this is the only regression test we have, but we’ve been wanting to add more and just haven’t had time (e.g. it’d be nice if we could test memory or perf of some of our models too).

SalmanMohammadi (approved):

SalmanMohammadi:
I'll raise an issue : )

PR #1309, opened 2024-08-12
adding expandable segments

SalmanMohammadi:
#### Context
- [x] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

#1185 

#### Changelog

Adding `os.environ["PYTORCH_CUDA_ALLOC_CONF"] = "expandable_segments:True"` to all our recipes.

#### Test plan
Run on 2080 Super 8GB.
```bash
(tune) salman@combuter:~/torchtune$ echo $PYTORCH_CUDA_ALLOC_CONF

(tune) salman@combuter:~/torchtune$ tune run full_finetune_single_device --config qwen2/0.5B_full_single_device log_peak_memory_stats=True metric_logger=torchtune.utils.metric_logging.WandBLogger metric_logging.project=torchtune_mem checkpointer.checkpoint_dir=/home/salman/models/Qwen2-0.5B-Instruct tokenizer.path=/home/salman/models/Qwen2-0.5B-Instruct/vocab.json tokenizer.merges_file=/home/salman/models/Qwen2-0.5B-Instruct/merges.txt max_steps_per_epoch=250
...
1|250|Loss: 1.0745091438293457: 100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 250/250 [22:37<00:00,  5.43s/it]
```
[WandB](https://wandb.ai/salman-mohammadi/torchtune/runs/urcs5jpl?nw=nwusersalmanmohammadi) of successful run with all peak memory stats <= 8GB.

### DIFF
--- a/recipes/full_finetune_distributed.py
+++ b/recipes/full_finetune_distributed.py
@@ -4,9 +4,9 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
+import os
 import sys
 import time
-
 from functools import partial
 from typing import Any, Dict, Optional, Tuple, Union
 from warnings import warn
@@ -36,6 +36,8 @@ from tqdm import tqdm
 
 
 log = utils.get_logger("DEBUG")
+
+os.environ["PYTORCH_CUDA_ALLOC_CONF"] = "expandable_segments:True"
 
 
 class FullFinetuneRecipeDistributed(FTRecipeInterface):
@@ -397,11 +399,11 @@ class FullFinetuneRecipeDistributed(FTRecipeInterface):
             sync_module_states=True,
             # Initialize empty modules on all non-zero ranks
             param_init_fn=(
-                lambda module: module.to_empty(
-                    device=torch.device("cuda"), recurse=False
+                lambda module: (
+                    module.to_empty(device=torch.device("cuda"), recurse=False)
+                    if not self._is_rank_zero
+                    else None
                 )
-                if not self._is_rank_zero
-                else None
             ),
         )
 
@@ -477,13 +479,15 @@ class FullFinetuneRecipeDistributed(FTRecipeInterface):
             dataset=ds,
             batch_size=batch_size,
             sampler=sampler,
-            collate_fn=partial(
-                utils.padded_collate,
-                padding_idx=self._tokenizer.pad_id,
-                ignore_idx=self._loss_fn.ignore_index,
-            )
-            if not packed
-            else None,
+            collate_fn=(
+                partial(
+                    utils.padded_collate,
+                    padding_idx=self._tokenizer.pad_id,
+                    ignore_idx=self._loss_fn.ignore_index,
+                )
+                if not packed
+                else None
+            ),
         )
 
         if self._is_rank_zero:
--- a/recipes/full_finetune_single_device.py
+++ b/recipes/full_finetune_single_device.py
@@ -28,6 +28,8 @@ from tqdm import tqdm
 
 log = utils.get_logger("DEBUG")
 
+os.environ["PYTORCH_CUDA_ALLOC_CONF"] = "expandable_segments:True"
+
 
 class FullFinetuneRecipeSingleDevice(FTRecipeInterface):
     """
@@ -424,13 +426,15 @@ class FullFinetuneRecipeSingleDevice(FTRecipeInterface):
             dataset=ds,
             batch_size=batch_size,
             sampler=sampler,
-            collate_fn=partial(
-                utils.padded_collate,
-                padding_idx=self._tokenizer.pad_id,
-                ignore_idx=self._loss_fn.ignore_index,
-            )
-            if not packed
-            else None,
+            collate_fn=(
+                partial(
+                    utils.padded_collate,
+                    padding_idx=self._tokenizer.pad_id,
+                    ignore_idx=self._loss_fn.ignore_index,
+                )
+                if not packed
+                else None
+            ),
         )
 
         log.info("Dataset and Sampler are initialized.")
--- a/recipes/lora_dpo_distributed.py
+++ b/recipes/lora_dpo_distributed.py
@@ -4,9 +4,9 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
+import os
 import sys
 import time
-
 from functools import partial
 from typing import Any, Dict, Optional, Tuple
 from warnings import warn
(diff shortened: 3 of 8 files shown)


felipemello1:
I think that if we were to add it, we can probably do it in torchtune/__init__.py, so it is in a single place. Wdyt?

SalmanMohammadi:
> I think that if we were to add it, we can probably do it in torchtune/**init**.py, so it is in a single place. Wdyt?

I was thinking similarly - Phil pointed out that having it at the top of the recipes makes it more visible and easier to disable since it's an experimental feature, which I agree with.

felipemello1:
@SalmanMohammadi , wanna make this change and add it to __init__.py? I would be ok with approving it, since so many people writing PRs are setting it to True to not OOM, and no one reported problems. 

@ebsmothers @pbontrager  , do you think we would be ok with approving it?

SalmanMohammadi:
> @SalmanMohammadi , wanna make this change and add it to **init**.py? I would be ok with approving it, since so many people writing PRs are setting it to True to not OOM, and no one reported problems.

Sorry, that was poorly worded on my part. 

> Phil pointed out that having it at the top of the recipes makes it more visible and easier to disable since it's an experimental feature, which I agree with.

I meant I agreed that having it in the top of recipes is more visible than having it set somewhere non-obvious/hidden like an init or util function : )

cc @joecummings too for landing this

felipemello1:
that makes sense! I edited my initial comment before I saw your reply. I do think that this should be controlled by a flag, and leave as False until pytorch merge their PR.

Apparently the only problem with leaving this a true by default is because they dont have it implemented for CUDA IPC, which I have no clue if is relevant for our users.

SalmanMohammadi:
> that makes sense! I edited my initial comment before I saw your reply. I do think that this should be controlled by a flag, and leave as False until pytorch merge their PR.

What about something like `if not "PYTORCH_CUDA_ALLOC_CONF" in os.environ:...` so we respect the user's choice for those who have deliberately set it, if it OOMs due to that issue they'll see hint from torch to set it anyway, and for 99% (pulling this out of the air) of users who don't care about this variable we set this sensible default?

Since this is hasn't landed in torch stable, this is still slightly experimental (I agree w you that it seems like a no-brainer and we've seen a lot less OOMs due to it), so we want it to be easily removable s.t. there are only shallow references to it in our codebase.

ebsmothers:
Personally I'm OK with these changes assuming we understand the implications across our recipes. If we have enough datapoints that this doesn't obviously regress any of our recipes I think we should just enable it. I would say we should run each model at least once and each recipe at least once then link the memory numbers in the test plan (you can take the testing of #1287 as an example here). @SalmanMohammadi I know you may not be able to run on distributed setups, but maybe someone can help you out here? 

Only other thing: maybe just move the definition of the environment variable into `recipe_main()` (unless there is some technical reason not to do this?)

SalmanMohammadi:
> I would say we should run each model at least once and each recipe at least once then link the memory numbers in the test plan (you can take the testing of #1287 as an example here)

As in, n_models * n_recipes? D :

If so @felipemello1 I might ask for your config profiling script if it can be adjusted for running multiple models.

felipemello1:
@SalmanMohammadi , i dont think that we need extensive testing. My understand is that we just want each recipe once, with one model. Can you confirm @ebsmothers ?

ebsmothers:
@SalmanMohammadi @felipemello1 sorry I could have stated this more clearly. Really I mean max(n_models, n_recipes), i.e. run every recipe at least once and every model at least once. Definitely do not need to do all possible combinations of the two

felipemello1:
I had issues with the command below when using expandable=True. It worked when using chunkedCE though.

```
tune run --nnodes 1 --nproc_per_node 8 lora_finetune_distributed --config gemma/2B_lora optimizer_in_bwd=False enable_activation_checkpointing=True loss=torch.nn.CrossEntropyLoss optimizer._component_=torch.optim.AdamW compile=True dataset.packed=False dataset.train_on_input=True tokenizer.max_seq_len=2048 metric_logger=torchtune.utils.metric_logging.WandBLogger metric_logger.project=chunked_CE log_every_n_steps=1 log_peak_memory_stats=True gradient_accumulation_steps=1 epochs=1 batch_size=32 max_steps_per_epoch=200 metric_logger.name=gemma2b__vanillaCE__nrpoc8_lora__AC checkpointer.output_dir=/tmp/Meta-Llama-3.1-8B-Instruct/gemma2b__vanillaCE__nrpoc8_lora__AC
```

recipes/full_finetune_distributed.py
```diff
@@ -29,6 +29,8 @@
 
 log = utils.get_logger("DEBUG")
 
+os.environ["PYTORCH_CUDA_ALLOC_CONF"] = "expandable_segments:True"
```
felipemello1, line 32: I think that it should be a parameter in the config, but we should not expose it. The default is true.

If some user faces an issue, they can just turn it off with a flag, instead of having to create a new recipe. Something like:

```
# Expandable segments save reserved memory and cant prevent OOM.
# If you have an issue because of this flag, you can set it to false in your config or add a flag to
# you command `tune run <your configurations> use_expandable_segments=False`
# For more info, check: <link to the PR or issue>
use_expandable_segments = cfg.get("use_expandable_segments", "True)
if use_expandable_segments:
	os.environ["PYTORCH_CUDA_ALLOC_CONF"] = "expandable_segments:True"
```

I have just tested qwen/gemma/llama on lora/full single/dist and it worked well for the most part. Just some warnings with qwen/gemma.

SalmanMohammadi:
Thank you so very much for helping out here @felipemello1.

> [W828 08:50:04.996578516 CUDAAllocatorConfig.h:28] Warning: expandable_segments not supported on this platform (function operator())
Do you remember what hardware you were running this on? This is a bit strange.

> That is a good point. Maybe we can add both?

Given that there's some still odd behaviour I agree it'd be good to highlight that this is slightly experimental and make it easy to disable. The error above looks to be related to this issue https://github.com/pytorch/pytorch/issues/122057. I guess my only reservation here is that if a user sees that error message, they wouldn't really know that it's something that we enabled and that you can configure through torchtune, unless we warn by default that we've enabled this experimental feature. 

The impact of this seems to be pretty minimal though - it doesn't look like it breaks things, right? It just means the user might not be able to reproduce some of our advertised hardware benchmarks.

Sorry, this is a bit of thought vomit. Probably overkill.

PR #1312, opened 2024-08-12
TransformerDecoderLayer Refactor

pbontrager:
#### Context
- [x] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

This is an implementation of the similarly named RFC #1224.

#### Changelog
This follows all of the proposed changes from the RFC with the following exceptions:
- sa_norm remains unchanged based on feedback to avoid breaking checkpoints. This pattern is used for Cross attention too with ca_norm
- hidden_idx is renamed output_hidden_state based on feedback
- TransformerDecoder returns either 'output' or the list [*hidden, output] to simplify to a single output
- TransformerDecoder num_layers was kept to keep single layer initialization convenience option
- CausalSelfAttention was renamed to MultiHeadedAttention
- Didn't implement a legacy checkpoint mapping, since it was no longer needed

#### Test plan

All unit tests pass. Since TransformerSelfAttentionLayer is a drop in replacement, with the same logic, for TransformerDecoderLayer, recipe functionality hasn't changed. Along with unit and integration_tests, I'll also test launching one config per model.

#### Test Runs
Reporting Loss, TPS, and Memory for the 10th step
Config Overrides: `gradient_accumulation_steps=1 max_steps_per_epoch=12 epochs=1 log_peak_memory_stats=True batch_size=4 seed=1`

Model: Command | Original Results | Refactor Results
- Llama3 8B:
	- `tune run full_finetune_single_device --config llama3/8B_full_single_device
	- loss:1.0483 tokens_per_second_per_gpu:680.4253 peak_memory_active:16.4158
	- loss:1.0483 tokens_per_second_per_gpu:655.1651 peak_memory_active:16.4158 
- Llama2 7B:
	- `tune run lora_finetune_single_device --config llama2/7B_lora_single_device`
	- loss:1.0986 tokens_per_second_per_gpu:2947.4527 peak_memory_active:13.7403
	- loss:1.0999 tokens_per_second_per_gpu:2946.9523 peak_memory_active:13.7403
- Code Llama 2:
	- `tune run lora_finetune_single_device --config code_llama2/7B_qlora_single_device`
	- loss:1.3836 tokens_per_second_per_gpu:1540.0405 peak_memory_active:5.2585
	- loss:1.3843 tokens_per_second_per_gpu:1540.2066 peak_memory_active:5.2585
- Phi 3 Mini:
	- `tune run --nproc_per_node 4 full_finetune_distributed --config phi3/mini_full`
	- loss:0.8311 tokens_per_second_per_gpu:718.9642 peak_memory_active:9.0107
	- loss:0.8328 tokens_per_second_per_gpu:750.9670 peak_memory_active:9.0107
- Mistral 7B:
	- `tune run full_finetune_single_device --config mistral/7B_full_low_memory`
	- loss:0.8734 tokens_per_second_per_gpu:989.0916 peak_memory_active:14.4627
	- loss:0.873 tokens_per_second_per_gpu:975.2271 peak_memory_active:14.4627
- Gemma 2B:
	- `tune run --nproc_per_node 4 lora_finetune_distributed --config gemma/2B_lora`
	- loss:2.3675 tokens_per_second_per_gpu:245.2158 peak_memory_active:4.3747
	- loss:2.3717 tokens_per_second_per_gpu:261.6916 peak_memory_active:4.5800


joecummings (commented):
I see that all the CI passes but I want two more things:
1) Output from the manual tests you ran with 7B checkpoints
2) Running before and after Llama3.1 8B looking at memory and toks/sec b/c that isn't covered by the other tests and you did change the wrapping of FSDP

torchtune/modules/transformer.py
```diff
@@ -145,20 +290,32 @@ class TransformerDecoder(nn.Module):
...
-        layer: TransformerDecoderLayer,
-        num_layers: int,
+        layers: Union[nn.Module, List[nn.Module]],
+        num_layers: Optional[int],
         max_seq_len: int,
         num_heads: int,
         head_dim: int,
         norm: nn.Module,
         output: nn.Linear,
+        output_hidden_states: Optional[List[int]] = None,
```
SalmanMohammadi, line 300: tiny nit, why not?
```suggestion
        output_hidden_states: Optional[List[int]] = [],
```

torchtune/modules/transformer.py
```diff
@@ -145,20 +290,32 @@ class TransformerDecoder(nn.Module):
...
+                layers, nn.Module
+            ), "num_layers is defined, layers must be a module"
+            layers = _get_clones(layers, num_layers)
 
         self.tok_embeddings = tok_embeddings
-        self.layers = _get_clones(layer, num_layers)
+        self.layers = layers
         self.norm = norm
         self.output = output
+        self.output_hidden_states = output_hidden_states or []
```
SalmanMohammadi, line 318: then you dont have to do this

torchtune/modules/transformer.py
```diff
@@ -218,13 +381,16 @@ def forward(
         Note: At the very first step of inference, when the model is provided with a prompt,
         ``input_pos`` would contain the positions of all of the tokens in the prompt
         (eg: ``torch.arange(prompt_length)``). This is because we will need to compute the
-        KV values for each position.
+        KV values for each position.M
 
         Returns:
-            Tensor: output tensor with shape [b x s x v]
+            Tensor: output tensor with shape [b x s x v] or a list of layer
```
SalmanMohammadi, line 387: typing should be Union[Tensor, List[Tensor]?

torchtune/modules/transformer.py
pbontrager: You don't want to use mutable types as a default value in Python. Otherwise, all instances of TransformerDecoder that use the default value will actually share the same list object. Since we don't modify the list here in the class, it's not actually a problem, but it's a risky pattern nonetheless that is best avoided.

torchtune/modules/transformer.py
```diff
@@ -209,6 +368,10 @@ def forward(
                 before the softmax. A value of True in row i and column j means token i attends
                 to token j. A value of False means token i does not attend to token j. If no
                 mask is specified, a causal mask is used by default. Default is None.
+            encoder_input (Optional[Tensor]): Optional second input to cross attend with x. Shape
+                [batch_size x seq_length x embed_dim] (seq_length and embed_dim may very from x)
```
SalmanMohammadi, line 372:
```suggestion
                [batch_size x seq_length x embed_dim] (seq_length and embed_dim may vary from x)
```

torchtune/modules/transformer.py
```diff
@@ -218,13 +381,16 @@ def forward(
         Note: At the very first step of inference, when the model is provided with a prompt,
         ``input_pos`` would contain the positions of all of the tokens in the prompt
         (eg: ``torch.arange(prompt_length)``). This is because we will need to compute the
-        KV values for each position.
+        KV values for each position.M
```
SalmanMohammadi, line 384: Mmmmmmmmm

torchtune/modules/attention.py
```diff
@@ -162,6 +176,8 @@ def forward(
         Args:
             x (Tensor): input tensor with shape
                 [batch_size x seq_length x embed_dim]
+            y (Optional[Tensor]): second input tensor for cross attentin with shape
```
SalmanMohammadi, line 179:
```suggestion
            y (Optional[Tensor]): second input tensor for cross attention with shape
```

torchtune/modules/transformer.py
```diff
@@ -90,13 +99,147 @@ def forward(
...
+            **kwargs (Dict): transformer layer inputs not relevant to self attention.
+
+        Returns:
+            Tensor: output tensor with same shape as input
+                [batch_size x seq_length x embed_dim]
+        """
+        # During decoding, it's possible encoder_input is None because the embeds
+        # are already stored in the kv cache.
+        empty_cache = not self.cache_enabled or self.attn.kv_cache.size == 0
+        # Skip cross attention when no secondary input as it's primary purpose
```
kartikayk, line 217: Sorry if you already explained this before but why does the cross-attention layer skip cross-attention?

torchtune/modules/transformer.py
```diff
@@ -90,13 +99,147 @@ def forward(
...
+                [batch_size x x_seq_len x encoder_seq_len]
+            **kwargs (Dict): transformer layer inputs not relevant to self attention.
+
+        Returns:
+            Tensor: output tensor with same shape as input
+                [batch_size x seq_length x embed_dim]
+        """
+        # During decoding, it's possible encoder_input is None because the embeds
+        # are already stored in the kv cache.
+        empty_cache = not self.cache_enabled or self.attn.kv_cache.size == 0
```
kartikayk, line 216: I don't think I follow?

torchtune/modules/transformer.py
```diff
@@ -90,13 +99,147 @@ def forward(
...
+        # A mask of tokens (x) with no encoder_input
+        skip_mask = self._skip_mask(encoder_mask)
+
+        # Input tensor and attention output have the same shape
+        # [b, s, d]
+        # Norm applied before self-attention
+        # TODO: Add support for sample packing and bring ack input_pos
+        attn_out = self.attn(self.ca_norm(x), encoder_input, mask=encoder_mask)
+        if skip_mask is not None:
+            attn_out.masked_fill_(skip_mask, 0)
```
kartikayk, line 231: I don't think I follow?

torchtune/modules/transformer.py
```diff
@@ -145,20 +290,32 @@ class TransformerDecoder(nn.Module):
...
         max_seq_len: int,
         num_heads: int,
         head_dim: int,
         norm: nn.Module,
         output: nn.Linear,
+        output_hidden_states: Optional[List[int]] = None,
     ) -> None:
         super().__init__()
+        if num_layers is None:
+            assert isinstance(
```
kartikayk, line 304: Same comment as above around the use of assert

torchtune/modules/transformer.py
kartikayk:
```suggestion
                [batch_size x seq_length x embed_dim] where seq_length and embed_dim may very from x
```

torchtune/modules/transformer.py
```diff
@@ -209,6 +368,10 @@ def forward(
                 before the softmax. A value of True in row i and column j means token i attends
                 to token j. A value of False means token i does not attend to token j. If no
                 mask is specified, a causal mask is used by default. Default is None.
+            encoder_input (Optional[Tensor]): Optional second input to cross attend with x. Shape
+                [batch_size x seq_length x embed_dim] (seq_length and embed_dim may very from x)
+            encoder_mask (Optional[Tensor]):  Cross attention boolean tensor with shape
```
kartikayk, line 373: Add some information around what the 1s and 0s in the mask mean. We had some users open up issues around this previously and we committed to describing this in more detail

torchtune/modules/transformer.py
```diff
@@ -249,15 +424,27 @@ def forward(
...
+            )
 
         # shape: [b, s, d]
         h = self.norm(h)
 
         # shape: [b, s, out_dim] - out_dim is usually the vocab size
         output = self.output(h).float()
+
+        # Output list if hidden states are requested, otherwise just the output
+        output = output if not hidden else [*hidden, output]
```
kartikayk, line 447: Generally, I don't think I like the output type being dynamically determined. I'd much rather prefer if we always returned a list and and had output[-1] refer to the output. Otherwise the downstream users will always need to add checks related to the type. Ideally this isn't a list but a dataclass

kartikayk (commented):
Some initial comments - overall looks much cleaner! Testing would need to be comprehensive for this to land

torchtune/modules/attention.py
```diff
@@ -14,9 +14,9 @@
 logger = logging.getLogger(__name__)
 
 
-class CausalSelfAttention(nn.Module):
-    """Multi-headed grouped query self-attention (GQA) layer introduced
-    in https://arxiv.org/abs/2305.13245v1.
+class MultiHeadedAttention(nn.Module):
```
RdoubleA, line 17: nit: isn't the canonical name `MultiHeadAttention`?

torchtune/modules/attention.py
```diff
@@ -105,20 +112,26 @@ def __init__(
...
+            raise ValueError("q and k norm must be set together")
+
         # Set attributes
         self.num_heads = num_heads
         self.num_kv_heads = num_kv_heads
         self.embed_dim = embed_dim
         self.attn_dropout = attn_dropout
         self.head_dim = head_dim
         self.max_seq_len = max_seq_len
+        self.is_causal = default_causal_mask
```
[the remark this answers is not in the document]
RdoubleA: would also vote for just keeping the name `is_causal` everywhere

torchtune/modules/transformer.py
SalmanMohammadi: + 1

torchtune/modules/transformer.py
```diff
@@ -233,6 +399,15 @@ def forward(
             - d: embed dim
             - m_s: max seq len
         """
+        # input tensor of shape [b, s]
+        bsz, seq_len = tokens.shape
+
+        if seq_len > self.max_seq_len:
```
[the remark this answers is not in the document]
RdoubleA: I've hit this many times, but it's a bit confusing because the current suggestion is to set max_seq_len in the tokenizer. In the llama models the max_seq_len in the TransformerDecoder is hardcoded to the max for the model. Maybe we should make it more clear that the max_seq_len here is the limit for the model (if we want to enforce a limit) rather than what is actually truncated - probably as a separate follow up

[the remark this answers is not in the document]
pbontrager: This becomes an issue when we look at kv_cache as the true max_seq_len might require much more memory allocated than we plan to use. On the other hand, this check should primarily check that the sequence isn't longer than what the model was trained for. We could expose a parameter in our builders for max_seq_len_cached or something, but I think that's out of scope of this work.

torchtune/modules/transformer.py
pbontrager: skip_mask represents rows in the original mask that are fully masked out, i.e a given token can't attend to any encoder embeds. This results in NaN values in that row from attention. This replaces them with 0s in place.

torchtune/modules/tanh_gate.py
```diff
@@ -0,0 +1,27 @@
...
+
+from torch import nn, Tensor
+
+
+class TanhGate(nn.Module):
+    """Implements a basic learnable gate to scale layer outputs"""
+
+    def __init__(self) -> None:
+        super().__init__()
+        self.scale = nn.Parameter(torch.zeros(1))
```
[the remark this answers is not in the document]
pbontrager: The idea is for the gate to start closed and default towards skipping each layer and then let the model learn to progressively "open up" the activation outputs as needed.

torchtune/modules/transformer.py
```diff
@@ -76,6 +84,7 @@ def forward(
                 of each token relative to its sample when packed, shape [b x s].
                 During inference, this indicates the position of the current token.
                 If none, assume the index of the token is its position id. Default is None.
+            **kwargs (Dict): transformer layer inputs not relevant to self attention.
```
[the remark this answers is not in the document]
pbontrager: encoder_input and encoder_mask

torchtune/modules/transformer.py
```diff
@@ -4,37 +4,44 @@
...
     ) -> None:
         super().__init__()
-        self.sa_norm = sa_norm
         self.attn = attn
-        self.mlp_norm = mlp_norm
         self.mlp = mlp
+        self.sa_norm = sa_norm or nn.Identity()
+        self.mlp_norm = mlp_norm or nn.Identity()
+        self.sa_scale = sa_scale or nn.Identity()
+        self.mlp_scale = mlp_scale or nn.Identity()
```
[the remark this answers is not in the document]
pbontrager: Yes

torchtune/modules/transformer.py
```diff
@@ -90,13 +99,147 @@ def forward(
...
+        attn: MultiHeadedAttention,
+        mlp: nn.Module,
+        *,
+        ca_norm: Optional[nn.Module] = None,
+        mlp_norm: Optional[nn.Module] = None,
+        ca_scale: Optional[nn.Module] = None,
+        mlp_scale: Optional[nn.Module] = None,
+    ) -> None:
+        super().__init__()
+        assert (
```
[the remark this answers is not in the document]
pbontrager: Updated this, but a quick search over the library shows that we're primarily using "assert" now and should probably update on this.

torchtune/modules/transformer.py
```diff
@@ -90,13 +99,147 @@ def forward(
...
+            attn.pos_embeddings is None
+        ), "Doesn't support positional embeddings for cross attention, because q and k are different sequences."
```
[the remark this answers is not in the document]
pbontrager: This would just be correct in the builder function as we'd be building a specific model. This is more a safety check for  if you're defining your own model and try to apply pos_embeddings naively to every layer. There may be a way to make it work with cross attention but I'd rather someone ran into this check and solved the problem instead of it silently happening incorrectly.

torchtune/modules/transformer.py
```diff
@@ -90,13 +99,147 @@ def forward(
...
+        mlp resulting in the token being skipped.
+
+        """
+        if mask is None:
+            return None
+        if mask.dtype == torch.bool:
+            mask = ~mask
+        else:
+            mask = torch.isneginf(mask)
+        mask = torch.all(mask, dim=-1, keepdim=True)
```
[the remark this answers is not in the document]
pbontrager: Is it better now?

torchtune/modules/transformer.py
pbontrager: You're supposed to skip the cross attention layer if there is no encoder input since there's nothing to cross attend to. But during decoding, there might be no encoder_input because it's already cached. So this checks both that the cache is empty and that there's no current input.

torchtune/modules/attention.py
pbontrager: You're right, I'll update this. 😢

torchtune/modules/transformer.py
pbontrager: After some external discussion, I'm going to leave this like this for now so that the default behavior of the model doesn't change through the library. And I'll revisit this in a separate PR.

pbontrager:
@joecummings I added all of the tests that I ran with their commands and their outputs. I ran most of our small models trying a variety of tests (full single device, full distributed, lora single device, lora distributed, qlora single device). I reported here the logged output for the 10th step which looks consistent between main and this branch (much smaller difference than between consecutive steps). I ran into zero issues running these (though our Gemma recipe doesn't seem to be learning).

Did I miss anything or would you be confident to land this now?

torchtune/modules/transformer.py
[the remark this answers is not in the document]
kartikayk: Thank you!

torchtune/modules/transformer.py
kartikayk: I'll trade you an "Approve" for a "TODO" in the docstring which leaves a reminder to clean this up :)

kartikayk (approved):

kartikayk:
Looks really clean and the testing is amazing - thank you for the attention to detail and for patiently answering my many noob questions!

PR #1313, opened 2024-08-12
Adding classifier checkpointing utils - renaming `peft/peft_utils.py`

### LINKED ISSUE #1300: Rename `peft/_peft_utils.py` to `peft/_utils.py`
We already know it's for PEFT, it's in the directory. Also, consistency with our other work.

SalmanMohammadi:
#### Context

- [x] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

closes #1300.

#### Changelog

I renamed peft_utils while I was here, since I was trying to unify the two.

This PR adds a small utility for loading checkpoints into classifier models. In all cases, we want to load all but the `output.weight` weights into the classifier. I've added a utility function in `torchtune.utils._checkpointing._utils` (what a wordy namespace huh?), `update_state_dict_for_classifier`, which will over-write `output.weight` in the state dict to be loaded, with the model's own `output.weight` if the shapes of the weights do not match. It also sprinkles a little validation on top.

To contextualize this, you'll be able to:
```python
from torchtune.models.llama2 import llama2_classifier
from torchtune import utils

model = llama2_classifier(
    num_classes=1,
    ...
)
# loading in a base llama2 checkpoint
checkpointer = utils.FullModelHFCheckpointer(
    ...
)

state_dict = checkpointer.load_checkpoint()["model"]
utils.update_state_dict_for_classifier(state_dict, model.named_parameters())
model.load_state_dict(state_dict=state_dict)
# UserWarning: Found output.weight with shape torch.Size([32000, 2048]) 
# in checkpoint. This will be overwritten with model's output.weight 
# with shape torch.Size([1, 2048])
```

in any recipe where you're fine-tuning a classifier. This also defers to our usual lora state dict validation:
```python
from torchtune.modules.peft._utils import (
    validate_missing_and_unexpected_for_lora,
)

from torchtune.models.llama2 import lora_llama2_classifier
from torchtune import utils

model = lora_llama2_classifier(
    ...
)
# load in base llama2 checkpoint
checkpointer = utils.FullModelHFCheckpointer(
    ...
)

state_dict = checkpointer.load_checkpoint()["model"]
utils.update_state_dict_for_classifier(state_dict, model.named_parameters())
base_missing, base_unexpected = model.load_state_dict(state_dict=state_dict, strict=False)
lora_missing, lora_unexpected = None, None
validate_missing_and_unexpected_for_lora(
    ...
)
# UserWarning: Found output.weight with shape torch.Size([32000, 2048]) 
# in checkpoint. This will be overwritten with model's output.weight 
# with shape torch.Size([1, 2048])

```

Is this overkill? I don't know. This will take out the validation from recipes involving classifiers (atm PPO, but soon classification and reward modelling), but also provide a relatively straightforward util for users.

#### Test plan


recipes/dev/lora_finetune_fsdp2.py
```diff
@@ -27,7 +27,7 @@
 from torchtune import config, modules, utils
 from torchtune.datasets import ConcatDataset
 from torchtune.modules.peft import LoRALinear
-from torchtune.modules.peft.peft_utils import (
+from torchtune.modules.peft._utils import (
```
joecummings, line 30: This is actually incorrect. Our recipes should really only be accessing public APIs (no underscore). Therefore, the proper way to do this would be to update the `__init__.py` file in `modules/peft` to import the APIs from `_utils.py`. Then, these import statements would look like:

```
from torchtune.modules.peft import LoRALinear, get_adapter_params, ...
```

SalmanMohammadi: whoops. thank you

recipes/ppo_full_finetune_single_device.py
```diff
@@ -426,24 +426,15 @@ def _setup_model(
         policy_model.load_state_dict(policy_state_dict)
         ref_policy_model.load_state_dict(ref_policy_state_dict)
 
-        reward_missing, reward_unexpected = reward_model.load_state_dict(
-            reward_model_state_dict, strict=False
-        )
-        value_missing, value_unexpected = value_model.load_state_dict(
-            value_model_state_dict, strict=False
+        utils.update_state_dict_for_classifier(
```
joecummings, line 429: Can you add an explicit comment here explaining what's happening?

tests/torchtune/utils/_checkpointing/test_utils.py
```diff
@@ -0,0 +1,69 @@
...
+            num_classes=6,
+            vocab_size=VOCAB_SIZE,
+            num_layers=N_LAYERS,
+            num_heads=NUM_KV_HEADS,
+            num_kv_heads=NUM_KV_HEADS,
+            embed_dim=EMBED_DIM,
+            max_seq_len=MAX_SEQ_LEN,
+        )
+
+    def test_loading_base_checkpoint_into_classifier(
```
joecummings, line 47: Might be my small brain but this test is confusing to me.

L50 - We update the llama2_state_dict with the classification params
L53 - We assign `expected_output_weight` from `llama2_classifier_model.state_dict` which should just be random weights??
L54 - We load the actual weights into the `llama2_classifier_model`
L55 - We check the output weight of the `llama2_classifier_model`
L56 - We assert they are equal

Okay after writing it out, I think I see how this is working. Maybe it would make more sense in my mind by moving L53 to the beginning of the test function. Then can you put some comments in there?

torchtune/utils/_checkpointing/_utils.py
```diff
@@ -0,0 +1,53 @@
...
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+from typing import Dict, Iterable, Tuple
+from warnings import warn
+
+import torch
+
+
+def update_state_dict_for_classifier(
```
joecummings, line 13: This should be a public API and also exposed in our documentation.

joecummings (commented):
Couple comments.

torchtune/utils/_checkpointing/_utils.py
```diff
@@ -0,0 +1,53 @@
...
+    state_dict: Dict[str, torch.Tensor],
+    model_named_parameters: Iterable[Tuple[str, torch.nn.Parameter]],
+):
+    """
+    Validates the state dict for checkpoint loading for a classifier model, to be
+    used prior to a call to ``model.load_state_dict(state_dict)``.
+    When fine-tuning a classifier model from the checkpoint of a base language model
+    which has ``output.weight`` of shape ``[vocab_dim, embed_dim]``, we overwrite
```
felipemello1, line 21: super nit: we dont check for [vocab_dim, embed_dim]`, so it could be of any shape (except the checkpoint's shape). Is that correct?

Just for inspiration, in case you decide to make changes, maybe we could do something like below (this is really just an idea. I am not asking you to replace the docstring :P )

"""
Used to replace the output.weight of a base model when the checkpoint output.weight has a different shape.

This is useful for situations like reward modeling, when the base model originally has shape ``[vocab_dim, embed_dim]``, but we want it to be something like ``[num_classes, embed_dim]``.

Notes:
- output.bias will be ignored
- If the output.shape is already of the same shape as the checkpoint, this function doesn't replace it.
"""

Btw, just confirming: The reason we only swap the weights if their shape is different is because the ``model.load_state_dict`` will have already loaded the weights correctly?

felipemello1 (commented):
just small nits.

torchtune/utils/_checkpointing/_utils.py
SalmanMohammadi: > The reason we only swap the weights if their shape is different is because the model.load_state_dict will have already loaded the weights correctly?

This is a good point, maybe we always want to swap the weights. If the user is starting from another classifier model with the same `num_classes` but is using it for a different task, does it make sense to start off with the pretrained weights?

> super nit: we dont check for [vocab_dim, embed_dim]`, so it could be of any shape (except the checkpoint's shape). Is that correct?

Sorry, what do you mean here? We're assuming the base model is identical aside from the output layer.

SalmanMohammadi: To answer your question:
>  The reason we only swap the weights if their shape is different is because the model.load_state_dict will have already loaded the weights correctly?

this is before load state dict

tests/torchtune/utils/_checkpointing/test_utils.py
SalmanMohammadi: Yeah maybe this isn't a well-designed test. It's a relatively trivial check. I'll update with your suggestions, ty

torchtune/utils/_checkpointing/utils.py
```diff
@@ -15,39 +15,46 @@ def update_state_dict_for_classifier(
...
 
     Raises:
         AssertionError: if ``state_dict`` does not contain ``output.weight``.
+        AssertionError: if ``model_named_parameters`` does not contain ``output.weight``.
+
     """
-    output_weight = [
-        (k, v) for (k, v) in model_named_parameters if k == "output.weight"
-    ]
+    output_weight = dict(model_named_parameters).get("output.weight", None)
```
felipemello1, line 42: just a thought: if we have to do dict, does it create a copy? If so, would it be super slow for larger models? Not sure whats happening behind the scenes

Maybe there is some trick like model_named_parameters.args.get("output.weight", None)?

tests/torchtune/utils/_checkpointing/test_utils.py
joecummings: The comments make this super clear - thank yoU!!

torchtune/utils/_checkpointing/utils.py
SalmanMohammadi: I don't think it does a deep copy of the underlying tensor, I had a quick test. model_named_parameters returns an iterable so needs to be consumed in some way
I agree I'd like for there to be a nice way. Open to suggestions.

joecummings (approved):
This looks good to me.

cc @felipemello1  if you have any issues merging this

felipemello1 (approved):

SalmanMohammadi:
Pls don't merge yet, I broke the PPO integration tests (already worth the pain of writing those)

PR #1315, opened 2024-08-12
Prevent OOM during checkpoint save on colab for llama3-8b qlora recipe

mikaylagawarecki:
#### Context
This PR prevents OOM during checkpoint save on colab for the following recipe, 

`tune run recipes/lora_finetune_single_device.py --config recipes/configs/llama3/8B_qlora_single_device.yaml`

I do believe it is possible to do better (perf improvements for this colab case) if we refactor the checkpointing logic, but that would be a more invasive refactor imo for now this is the most minimally invasive change that unblocks this use case.

#### Changelog
- Added a new hook `_low_ram_reparametrize_as_dtype_state_dict_post_hook` and `_register_reparametrize_state_dict_hooks` that toggles between the regular reparametrize state_dict hook and this one
- added a model config `low_cpu_ram` -- when `low_cpu_ram` is`True` `_register_reparametrize_state_dict_hooks` toggles to the `low_ram` version of the hook
- PyTorch side changes were landed in https://github.com/pytorch/pytorch/pull/134504 [was reverted, remerging right now as of 9/05] and https://github.com/pytorch/pytorch/pull/134371 **both should be available in the 20240830 nightly**

Old changelog (keeping around for posterity)
- ~Set a seed  in `8B_qlora_single_device.yaml` to make dataloader samples (and hence weights) deterministic~
- ~[for testing velocity purposes]Changed number of epochs to 8, reduced `max_steps_per_epoch` --> 20, `gradient_accumulation_steps` --> 2 to make `save_checkpoint` be called sooner~
- ~[for loss curves] some changes to lora_finetune_single_device.py to mimic a user doing `resume_from_checkpoint` after each epoch (while preserving logger)~
- ~[Not to be landed in torchtune, for PoC purpose] patched [`FakeTensor.__reduce_ex__ `](https://github.com/pytorch/torchtune/pull/1315/files#diff-734dbb110089223adfc465adb1594c2645bc025171e23c67f7b7bece857edbf0R42-R56) which is needed to ensure the `write_record_metadata` utility to create empty checkpoints is called (requires changes in https://github.com/pytorch/pytorch/pull/133272) I need to figure out how to land this piece :)~ 
- ~Skip registration of `reparametrize_as_dtype_state_dict_post_hook` for llama3~
~- Showed an example of how to do the corresponding (using mmap to prevent OOM) in `lora_finetune_single_device.py:save_checkpoint`~

##### Sanity check 
Ran `tune run` command on devgpu  (with only the changes in `8B_qlora_single_device.yaml` to set seed and decrease steps per epoch)  and verified that `meta_model_0.pt` generated is the same before and after the changes in this PR with small [snippet](https://gist.github.com/mikaylagawarecki/60cf8e77f6230fcd45c62a4a7f479370)

| | Before ckpt save time (s) | After ckpt save time (s) |
|-|-|-|
| devgpu | 178s | ~200s (ranged from 190 to 220ish) |
|colab | OOM (no baseline) | 510s (ranged from 421s to 610s) |

Verified that colab does not OOM
https://colab.research.google.com/drive/1y7Az78ATauK7gkewZkcMO3cNgVWm1233?usp=sharing

##### Loss Curves
The validation was run on commit `abdbd7` which has special logic to mimic `resume_from_checkpoint` for each epoch

Config is per the changes in `recipes/configs/llama3/8B_qlora_single_device.yaml` (8 epochs with 20 steps per epoch and gradient accumulation every 2 steps) on 6cf31b6. For the reloading checkpoint case [I modified `lora_fine_tune_single_device.py:recipe_main` to mimic `resume_from_checkpoint` after each epoch](https://github.com/pytorch/torchtune/pull/1315/files#diff-734dbb110089223adfc465adb1594c2645bc025171e23c67f7b7bece857edbf0R711-R727)

Devgpu:
Screenshot 2024-09-05 at 10 29 09 PM

Colab:
Screenshot 2024-09-05 at 10 26 19 PM

### DIFF
--- a/recipes/configs/llama3/8B_qlora_single_device.yaml
+++ b/recipes/configs/llama3/8B_qlora_single_device.yaml
@@ -45,7 +45,7 @@ save_adapter_weights_only: False
 # Dataset and Sampler
 dataset:
   _component_: torchtune.datasets.alpaca_cleaned_dataset
-seed: 1372749820
+seed: null
 shuffle: True
 batch_size: 2
 
@@ -62,17 +62,16 @@ loss:
   _component_: torchtune.modules.loss.CEWithChunkedOutputLoss
 
 # Training
-epochs: 10
-max_steps_per_epoch: 20
-gradient_accumulation_steps: 2
+epochs: 1
+max_steps_per_epoch: null
+gradient_accumulation_steps: 16
 compile: False
 
 # Logging
 output_dir: /tmp/qlora_finetune_output/
 metric_logger:
-  _component_: torchtune.utils.metric_logging.WandBLogger
-  # the W&B project to log to
-  project: torchtune
+  _component_: torchtune.utils.metric_logging.DiskLogger
+  log_dir: ${output_dir}
 log_every_n_steps: 1
 log_peak_memory_stats: False
 
--- a/recipes/configs/llama3/8B_qlora_single_device.yaml
+++ b/recipes/configs/llama3/8B_qlora_single_device.yaml
@@ -45,7 +45,7 @@ save_adapter_weights_only: False
 # Dataset and Sampler
 dataset:
   _component_: torchtune.datasets.alpaca_cleaned_dataset
-seed: null
+seed: 1372749820
 shuffle: True
 batch_size: 2
 
@@ -62,16 +62,18 @@ loss:
   _component_: torchtune.modules.loss.CEWithChunkedOutputLoss
 
 # Training
-epochs: 1
-max_steps_per_epoch: null
-gradient_accumulation_steps: 16
+epochs: 5
+max_steps_per_epoch: 10
+gradient_accumulation_steps: 2
 compile: False
 
 # Logging
 output_dir: /tmp/qlora_finetune_output/
+# enable logging to the built-in WandBLogger
 metric_logger:
-  _component_: torchtune.training.metric_logging.DiskLogger
-  log_dir: ${output_dir}
+  _component_: torchtune.training.metric_logging.WandBLogger
+  # the W&B project to log to
+  project: torchtune
 log_every_n_steps: 1
 log_peak_memory_stats: False
 
@@ -105,5 +107,4 @@ profiler:
   active_steps: 2
   num_cycles: 1
 
-# For colab use True
-low_cpu_ram: False
+low_cpu_ram: True
--- a/recipes/configs/llama3/8B_qlora_single_device.yaml
+++ b/recipes/configs/llama3/8B_qlora_single_device.yaml
@@ -105,4 +105,5 @@ profiler:
   active_steps: 2
   num_cycles: 1
 
-low_cpu_ram: True
+# For colab
+low_cpu_ram: False
--- a/recipes/configs/llama3/8B_qlora_single_device.yaml
+++ b/recipes/configs/llama3/8B_qlora_single_device.yaml
@@ -23,7 +23,7 @@ model:
   apply_lora_to_output: False
   lora_rank: 8
   lora_alpha: 16
-  low_cpu_ram: True
+  low_cpu_ram: False
 
 # Tokenizer
 tokenizer:
--- a/recipes/configs/llama3/8B_qlora_single_device.yaml
+++ b/recipes/configs/llama3/8B_qlora_single_device.yaml
@@ -45,7 +45,7 @@ save_adapter_weights_only: False
 # Dataset and Sampler
 dataset:
   _component_: torchtune.datasets.alpaca_cleaned_dataset
-seed: 1372749820
+seed: null
 shuffle: True
 batch_size: 2
 
@@ -59,20 +59,19 @@ lr_scheduler:
   num_warmup_steps: 100
 
 loss:
-  _component_: torch.nn.CrossEntropyLoss
+  _component_: torchtune.modules.loss.CEWithChunkedOutputLoss
 
 # Training
-epochs: 10
-max_steps_per_epoch: 20
-gradient_accumulation_steps: 2
+epochs: 1
+max_steps_per_epoch: null
+gradient_accumulation_steps: 16
 compile: False
 
 # Logging
 output_dir: /tmp/qlora_finetune_output/
 metric_logger:
-  _component_: torchtune.utils.metric_logging.WandBLogger
-  # the W&B project to log to
-  project: torchtune
+  _component_: torchtune.utils.metric_logging.DiskLogger
+  log_dir: ${output_dir}
 log_every_n_steps: 1
 log_peak_memory_stats: False
 
--- a/recipes/configs/llama3/8B_qlora_single_device.yaml
+++ b/recipes/configs/llama3/8B_qlora_single_device.yaml
@@ -70,7 +70,7 @@ compile: False
 # Logging
 output_dir: /tmp/qlora_finetune_output/
 metric_logger:
-  _component_: torchtune.utils.metric_logging.DiskLogger
+  _component_: torchtune.training.metric_logging.DiskLogger
   log_dir: ${output_dir}
 log_every_n_steps: 1
 log_peak_memory_stats: False
--- a/recipes/configs/llama3/8B_qlora_single_device.yaml
+++ b/recipes/configs/llama3/8B_qlora_single_device.yaml
@@ -46,7 +46,7 @@ save_adapter_weights_only: False
 # Dataset and Sampler
 dataset:
   _component_: torchtune.datasets.alpaca_cleaned_dataset
-seed: 1372749820
+seed: null
 shuffle: True
 batch_size: 2
 
@@ -63,18 +63,16 @@ loss:
   _component_: torchtune.modules.loss.CEWithChunkedOutputLoss
 
 # Training
-epochs: 8
-max_steps_per_epoch: 20
-gradient_accumulation_steps: 2
+epochs: 1
+max_steps_per_epoch: null
+gradient_accumulation_steps: 16
 compile: False
 
 # Logging
 output_dir: /tmp/qlora_finetune_output/
-# enable logging to the built-in WandBLogger
 metric_logger:
-  _component_: torchtune.utils.metric_logging.WandBLogger
-  # the W&B project to log to
-  project: torchtune
+  _component_: torchtune.utils.metric_logging.DiskLogger
+  log_dir: ${output_dir}
 log_every_n_steps: 1
 log_peak_memory_stats: False
 
--- a/recipes/configs/llama3/8B_qlora_single_device.yaml
+++ b/recipes/configs/llama3/8B_qlora_single_device.yaml
@@ -105,5 +105,5 @@ profiler:
   active_steps: 2
   num_cycles: 1
 
-# For colab
+# For colab use True
 low_cpu_ram: False
--- a/recipes/configs/llama3/8B_qlora_single_device.yaml
+++ b/recipes/configs/llama3/8B_qlora_single_device.yaml
@@ -23,7 +23,6 @@ model:
   apply_lora_to_output: False
   lora_rank: 8
   lora_alpha: 16
-  low_cpu_ram: False
 
 # Tokenizer
 tokenizer:
@@ -46,7 +45,7 @@ save_adapter_weights_only: False
 # Dataset and Sampler
 dataset:
   _component_: torchtune.datasets.alpaca_cleaned_dataset
-seed: null
+seed: 1372749820
 shuffle: True
 batch_size: 2
 
@@ -63,16 +62,17 @@ loss:
   _component_: torchtune.modules.loss.CEWithChunkedOutputLoss
 
 # Training
-epochs: 1
-max_steps_per_epoch: null
-gradient_accumulation_steps: 16
+epochs: 10
+max_steps_per_epoch: 20
+gradient_accumulation_steps: 2
 compile: False
 
 # Logging
 output_dir: /tmp/qlora_finetune_output/
 metric_logger:
-  _component_: torchtune.utils.metric_logging.DiskLogger
-  log_dir: ${output_dir}
+  _component_: torchtune.utils.metric_logging.WandBLogger
+  # the W&B project to log to
+  project: torchtune
 log_every_n_steps: 1
 log_peak_memory_stats: False
 
@@ -105,3 +105,5 @@ profiler:
   warmup_steps: 5
   active_steps: 2
   num_cycles: 1
+
+low_cpu_ram: True
--- a/recipes/configs/llama3/8B_qlora_single_device.yaml
+++ b/recipes/configs/llama3/8B_qlora_single_device.yaml
@@ -45,7 +45,7 @@ save_adapter_weights_only: False
 # Dataset and Sampler
 dataset:
   _component_: torchtune.datasets.alpaca_cleaned_dataset
-seed: null
+seed: 1372749820
 shuffle: True
 batch_size: 2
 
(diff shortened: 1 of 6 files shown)


recipes/lora_finetune_single_device.py
```diff
@@ -491,20 +515,54 @@ def save_checkpoint(self, epoch: int) -> None:
...
+        state_dict = self._model.state_dict()
+        mode = FakeTensorMode()
+        converter = FakeTensorConverter()
+        fake_state_dict = OrderedDict()
+        for k, v in state_dict.items():
+            # remove adapters (we don't want them in merged_state_dict)
+            if "lora_a" in k or "lora_b" in k:
+                continue
+            if isinstance(v, NF4Tensor):
+                fake_state_dict[k] = converter.from_real_tensor(mode, v).to(torch.bfloat16).cpu()
```
ebsmothers, line 530: Might want to parametrize dtype here, I don't think Colab will support bfloat16

recipes/lora_finetune_single_device.py
```diff
@@ -491,20 +515,54 @@ def save_checkpoint(self, epoch: int) -> None:
...
+            else:
+                fake_state_dict[k] = converter.from_real_tensor(mode, v).cpu()
+
+        # === Save and load fake state_dict ===
+        # dest_state_dict is now backed by an mmap-ed file that can be written to
+        dest_state_dict_path = "/tmp/llama_state_dict.pt"
+        torch.save(fake_state_dict, dest_state_dict_path)
+        torch.serialization.set_default_mmap_options(MAP_SHARED)
+        dest_state_dict = torch.load(dest_state_dict_path, mmap=True, weights_only=True)
+        torch.serialization.set_default_mmap_options(MAP_PRIVATE)
```
ebsmothers, line 540: Maybe worth doing this as a context manager or something. Also noob q: why are we able to put back into `MAP_PRIVATE` mode before we actually copy `dest_state_dict`?

recipes/lora_finetune_single_device.py
```diff
@@ -491,20 +515,54 @@ def save_checkpoint(self, epoch: int) -> None:
...
+
+        # This bit does D2H one by one and since dest_state_dict is backed by mmap --> won't OOM :)
+        for k in state_dict.keys():
+            if 'lora_a' in k or 'lora_b' in k:
+                # adapters are very small ~0.04GB so we don't need to use mmap when offloading to CPU
+                dest_state_dict[k] = state_dict[k].cpu()
+            if isinstance(state_dict[k], NF4Tensor):
+                dest_state_dict[k].copy_(state_dict[k].to(torch.bfloat16))
```
ebsmothers, line 548: similar comment here about bfloat16

recipes/lora_finetune_single_device.py
```diff
@@ -491,20 +515,54 @@ def save_checkpoint(self, epoch: int) -> None:
...
+            else:
+                dest_state_dict[k].copy_(state_dict[k])
 
         # Construct the full state dict with LoRA weights merged into base LLM weights
         merged_state_dict = get_merged_lora_ckpt(
             state_dict,
             rank=self._lora_rank,
             alpha=self._lora_alpha,
+            dest_state_dict=dest_state_dict,
```
ebsmothers, line 557: (another noob q:) so after this, we can save the checkpoint normally (e.g. in the call to `save_checkpoint` on L580), even though we are now saving something like `{"model": dest_state_dict, ... (other stuff)}` to a separate file from the mmapped one?

ebsmothers (commented):
This is awesome! Left a few basic questions, but overall really excited to see that we'll be able to provide proper end-to-end Colab support with these changes. One high-level comment is that we should think about writing a utility for the portion in ~L520-L550. Then we can gate behind a config like `low_memory_save` or something like that. But that's more of a UX thing, happy to help out there if needed.

recipes/configs/llama3/8B_qlora_single_device.yaml
```diff
@@ -44,7 +44,7 @@ save_adapter_weights_only: False
 # Dataset and Sampler
 dataset:
   _component_: torchtune.datasets.alpaca_cleaned_dataset
-seed: null
+seed: 1372749820
```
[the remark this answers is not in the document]
mikaylagawarecki: This might have been a wrong assumption on my part, I assumed that a seed was needed for testing purposes so the DataLoader would feed the same samples and we would see the same loss curve between runs, but perhaps this is not necessary

recipes/lora_finetune_single_device.py
mikaylagawarecki: Good point, I should add a context manager for this, this is a global config that sets `MAP_SHARED` for all torch.load calls (the implementation is like this https://github.com/pytorch/pytorch/blob/main/torch/serialization.py#L1245-L1251), so changing back to `MAP_PRIVATE`  would not affect the preceding torch.load call

recipes/lora_finetune_single_device.py
mikaylagawarecki: We should think of dest_state_dict as being "mmap-backed"

The final `save_checkpoint` on L580 is resaving a new checkpoint with `{"model": dest_state_dict, ... (other stuff)}` to a new file

We could potentially do something smarter than what we're doing right now but the code changes would be more invasive than they are now, given that the checkpointers seem to do some remapping though I refrained from doing this as a v1, wdyt?

recipes/lora_finetune_single_device.py
```diff
@@ -507,6 +507,14 @@ def save_checkpoint(self, epoch: int) -> None:
         # Move to CPU to avoid a copy on GPU
         state_dict = {k: v.cpu() for k, v in self._model.state_dict().items()}
 
+        # Construct the adapter weights
+        # Do this using the state_dict to avoid running upcast and H2D in state_dict post hook twice
+        # Must be before get_merged_lora_ckpt because get_merged_lora_ckpt will remove lora keys
+        adapter_key_filter = lambda x: x in self.adapter_params
+        adapter_state_dict = {
+            k: v for k, v in state_dict.items() if adapter_key_filter(k)
+        }
```
mikaylagawarecki, line 516: Ideally we want to only run state_dict post hooks once, so we should reuse the state_dict

this does change the semantic slightly though -- before `adapter_0.pt` contained weights tagged with CUDA, but now it contains weights tagged with CPU.

 Not sure whether the old behavior was intended/whether this change is ok (but no CI seems to fail :D) Also when loading, we [`map_location="cpu"` regardless](https://github.com/pytorch/torchtune/blob/81bc3daaf15e84c0a40d813f6eed996595eb422a/torchtune/utils/_checkpointing/_checkpointer_utils.py#L110)

recipes/configs/llama3/8B_qlora_single_device.yaml
```diff
@@ -23,6 +23,7 @@ model:
   apply_lora_to_output: False
   lora_rank: 8
   lora_alpha: 16
+  low_cpu_ram: False
```
ebsmothers, line 26: Given that we have to modify the state dict hook, I see why it makes sense to put this in the model config. But it does feel a bit weird to me since it's not really a property of the model (more just the model is a convenient place for us to know that we're gonna have to upcast NF4 tensors). 

I wonder if we can instead define a standalone config, parse it in the recipe with e.g. `low_cpu_ram = cfg.get("low_cpu_ram", False)`, then use that to overwrite the `reparametrize_as_dtype_state_dict_post_hook`. Maybe a bit hacky, but we can at least assert that the expected state dict hook is there before replacing it to ensure that we aren't adding this onto any old non-QLoRA model. Is that obviously worse? (Mainly I want to avoid our model classes having to know or care about low-level details of how they're gonna be checkpointed)

torchtune/models/llama3/_component_builders.py
```diff
@@ -250,8 +257,20 @@ def lora_llama3(
     if quantize_base:
         # For QLoRA, we reparametrize 4-bit tensors to bf16, and offload to CPU on the fly
         # so as to not increase peak memory
+        if low_cpu_ram and _low_ram_reparametrize_as_dtype_state_dict_post_hook is None:
+            if sys.platform == "win32":
+                raise RuntimeError(
+                    "low_cpu_ram=True not supported on Windows."
+                )
+            else:
+                raise RuntimeError("low_cpu_ram=True requires torch.__version__ >= 2.5.0.dev20240830.")
```
ebsmothers, line 266: Similar comment here: ideally we can do these checks in the recipe (or in a utility) rather than in the builder of the model

torchtune/modules/common_utils.py
```diff
@@ -48,3 +52,81 @@ def reparametrize_as_dtype_state_dict_post_hook(
             state_dict[k] = v.to(dtype)
             if offload_to_cpu:
                 state_dict[k] = state_dict[k].cpu()
+
+
+# mmap.MAP_SHARED is not supported on Windows but this change targets colab.
+if hasattr(torch.serialization, "skip_data") and not sys.platform == "win32":
```
ebsmothers, line 58: Is the first half of this check just a proxy for a particular torch version? If so maybe better to just directly gate on that (with `torch_version_ge` or something)

mikaylagawarecki: `torch_version_ge` seems to cause circular import when imported in this file :/

ebsmothers: Oh yeah we are working to fix that, `__torch_version__` is good too

recipes/lora_finetune_single_device.py
ebsmothers: Sorry I missed this comment before. But yeah I think this change makes sense, I don't think there's any reason we need to require CUDA weights. And as you point out since we load on CPU when resuming it was probably never really an issue. Plus not re-running the state dict post hooks is a nice bonus.

recipes/lora_finetune_single_device.py
```diff
@@ -710,6 +721,8 @@ def recipe_main(cfg: DictConfig) -> None:
         - Overwritten by arguments from the command-line
     """
     config.log_config(recipe_name="LoRAFinetuneRecipeSingleDevice", cfg=cfg)
+    if cfg.get("low_cpu_ram", False):
+        common_utils._use_low_cpu_ram = True
```
ebsmothers, line 725: Does this have to be in `recipe_main`? Can we instead do it somewhere inside the recipe class (before the model gets instantiated)? Also would add a one-line comment explaining this

torchtune/modules/common_utils.py
```diff
@@ -48,3 +54,118 @@ def reparametrize_as_dtype_state_dict_post_hook(
             state_dict[k] = v.to(dtype)
             if offload_to_cpu:
                 state_dict[k] = state_dict[k].cpu()
+
+
+# mmap.MAP_SHARED is not supported on Windows but this change targets colab.
+if torch.__version__ >= "2.5.0.dev20240906" and not sys.platform == "win32":
```
ebsmothers, line 60: Just a sanity check here: is this if/else just to ensure that no one tries to directly import the `_low_ram_reparametrize_as_dtype_state_dict_post_hook` API on an unsupported environment? Mainly asking because we have the equivalent checks in `_register_reparametrize_state_dict_hooks` now

torchtune/models/llama3/_component_builders.py
```diff
@@ -250,9 +250,7 @@ def lora_llama3(
     if quantize_base:
         # For QLoRA, we reparametrize 4-bit tensors to bf16, and offload to CPU on the fly
         # so as to not increase peak memory
-        model._register_state_dict_hook(
-            partial(reparametrize_as_dtype_state_dict_post_hook, offload_to_cpu=True)
-        )
+        _register_reparametrize_state_dict_hooks(model)
```
ebsmothers, line 253: Doesn't have to be done in this PR, but we can think about adding this for other models that would have a similar memory situation when running QLoRA (Llama 3.1 8B is an obvious choice, but there are a handful of other similarly-sized models supported in our repo that could benefit from this)

recipes/lora_finetune_single_device.py
```diff
@@ -694,7 +698,14 @@ def train(self) -> None:
                     prof.step()
 
                 self.epochs_run += 1
+                start_save_checkpoint = time.time()
```
ebsmothers, line 701: nit: would use `time.perf_counter()` here unless there's a strong reason not to. If I understand the two APIs correctly it shouldn't make a huge difference for this application, and we use `time.perf_counter()` elsewhere in the recipe, so nice to be consistent. But do lmk if there's a specific reason we should be using `time.time()`

torchtune/modules/common_utils.py
```diff
@@ -48,3 +54,118 @@ def reparametrize_as_dtype_state_dict_post_hook(
...
+            else:
+                fake_state_dict[k] = converter.from_real_tensor(mode, v)
+
+            if offload_to_cpu:
+                fake_state_dict[k] = fake_state_dict[k].cpu()
+
+        # Create a state_dict on disk with space reserved for storage bytes
+        # Then load with mmap and MAP_SHARED (can writeback to disk file)
+        dest_state_dict_path = "/tmp/fake_state_dict.pt"
+        with torch.serialization.skip_data(materialize_fake_tensors=True):
```
ebsmothers, line 112: noob q: what does `materialize_fake_tensors` mean in this context?

torchtune/modules/common_utils.py
```diff
@@ -48,3 +54,118 @@ def reparametrize_as_dtype_state_dict_post_hook(
...
+        with torch.serialization.set_default_mmap_options(mmap.MAP_SHARED):
+            dest_state_dict = torch.load(
+                dest_state_dict_path, mmap=True, weights_only=True
+            )
+
+        # Do D2H and upcast one by one and since dest_state_dict is backed by mmap --> won't OOM
+        # even when there is no swap space (e.g. colab)
+        for k in state_dict.keys():
+            if isinstance(state_dict[k], NF4Tensor):
+                dest_state_dict[k].copy_(state_dict[k].to(torch.bfloat16))
```
ebsmothers, line 123: should be this?
```suggestion
                dest_state_dict[k].copy_(state_dict[k].to(dtype))
```

torchtune/modules/common_utils.py
```diff
@@ -48,3 +54,118 @@ def reparametrize_as_dtype_state_dict_post_hook(
...
+            else:
+                dest_state_dict[k].copy_(state_dict[k])
+
+        # In place update original state_dict object. Although the private state dict
+        # post hook supports out of place behavior, the semantic actually buggy. We eventually want
+        # to use the public state_dict post hook which does not support out of place behavior.
+        for k in state_dict.keys():
+            state_dict[k] = dest_state_dict[k]
```
ebsmothers, line 131: I think I have some misunderstanding here. If we inplace update the state dict to the upcasted version of the weights, why won't it cause an OOM?

ebsmothers (commented):
A few more small comments and questions, but otherwise this looks good to go!

torchtune/modules/common_utils.py
mikaylagawarecki: When we do `sd = torch.load(..., mmap=True)`, the *storages* of the tensors in `sd` are mmap-backed

`state_dict[k] = dest_state_dict[k]` does not access any pages of the storage of the tensor given by `dest_state_dict[k]`, so the storage is not materialized, and no OOM will happen

torchtune/modules/common_utils.py
mikaylagawarecki: It means that FakeTensors in the object passed to `torch.save` will be treated as if they were real tensors. The implications are

- `torch.load` will load a tensor on the FakeTensor's device with storage allocated but uninitialized (0s)

torchtune/modules/common_utils.py
mikaylagawarecki: You're right, removing this if else

torchtune/models/llama3/_component_builders.py
mikaylagawarecki: will do in followup

torchtune/modules/common_utils.py
ebsmothers: Oh yeah nvm I get it now, I was not thinking it through carefully enough. Thanks for the explanation!

ebsmothers (approved):
Thank you for enabling this, can't wait to put together some torchtune Colab notebooks for our users!

mikaylagawarecki:
Added changes from https://github.com/pytorch/torchtune/pull/1535 + one more round of loss curves

Screenshot 2024-09-10 at 3 27 23 PM
Screenshot 2024-09-10 at 3 27 12 PM

PR #1317, opened 2024-08-13
Add llama3.1 405B model definition and builder

lessw2020:
This PR:
1 - adds the model definition and base model builder for llama3_1 405B.  
2 - Have verified it's ability to produce a 405B model via import and running within torchchat/distributed:

Screenshot 2024-08-12 at 7 23 40 PM

3 - This PR is effectively a subset (main model only, rather than QLora/Lora) of https://github.com/pytorch/torchtune/pull/1232

#### Context
- [X ] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

#### Changelog

#### Test plan

### DIFF
--- a/torchtune/models/llama3/_model_builders.py
+++ b/torchtune/models/llama3/_model_builders.py
@@ -62,25 +62,7 @@ def llama3_70b() -> TransformerDecoder:
         rope_base=500000.0,
     )
 
-def llama3_1_405b() -> TransformerDecoder:
-    """
-    Builder for creating a Llama3.1 model initialized w/ the default 405B parameter values.
-    Returns:
-        TransformerDecoder: Instantiation of Llama3.1 405B model
-    """
-    return llama3_1(
-        vocab_size=128_256,
-        num_layers=126,
-        num_heads=128,
-        num_kv_heads=8,
-        embed_dim=16384,
-        max_seq_len=8192,
-        intermediate_dim=53248,
-        attn_dropout=0.0,
-        norm_eps=1e-5,
-        rope_base=500000.0,
-    )
-    
+ 
 def llama3_tokenizer(path: str, special_tokens_path: Optional[str] = None, max_seq_len: Optional[int] = None) -> Llama3Tokenizer:
     """
     Tokenizer for Llama3.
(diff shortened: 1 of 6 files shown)


torchtune/models/llama3/_model_builders.py
```diff
@@ -62,7 +62,25 @@ def llama3_70b() -> TransformerDecoder:
         rope_base=500000.0,
     )
 
-
+def llama3_1_405b() -> TransformerDecoder:
```
joecummings, line 65: I think this should be removed - this is in the Llama3 folder, not 3.1

joecummings (commented):
Just one comment! Otherwise looks great

felipemello1 (commented):
Maybe this was intentionally left out, but shouldnt we also add a config in torchtune/configs/llama3_1/405B_full.yaml, 405B_lora.yaml?

If so, they would also need to be registered here https://github.com/pytorch/torchtune/blob/main/torchtune/_recipe_registry.py

joecummings:
> Maybe this was intentionally left out, but shouldnt we also add a config in torchtune/configs/llama3_1/405B_full.yaml, 405B_lora.yaml?
> 
> If so, they would also need to be registered here [`main`/torchtune/_recipe_registry.py](https://github.com/pytorch/torchtune/blob/main/torchtune/_recipe_registry.py?rgh-link-date=2024-08-13T13%3A31%3A22Z)

Nope - we're just getting the model builder in here for now so that torchchat can use it.

torchtune/models/llama3/_model_builders.py
lessw2020: great catch - this was inadvertent.  Removing!

torchtune/models/llama3/_model_builders.py
```diff
@@ -62,7 +62,7 @@ def llama3_70b() -> TransformerDecoder:
         rope_base=500000.0,
     )
 
-
```
joecummings, line 65: 🙃

joecummings (approved):
Looks great! Will merge when CI is green.

torchtune/models/llama3/_model_builders.py
lessw2020: lol - after removing the inadvertent 405B builder, I was trying to keep the same formatting of two blank lines between each model_builder.

PR #1323, opened 2024-08-13
[EZ] fix DPO distributed recipe

RdoubleA:
## Context
Some variables were left undefined, causing an error when trying to run the lora_dpo_distributed recipe on main when logging metrics. Copied the missing lines over from the single device recipe.

## Test plan
`tune run --nnodes 1 --nproc_per_node 8 lora_dpo_distributed --config llama2/7B_lora_dpo` works now

### DIFF
--- a/recipes/lora_dpo_distributed.py
+++ b/recipes/lora_dpo_distributed.py
@@ -620,6 +620,9 @@ class LoRADPORecipeDistributed(FTRecipeInterface):
                     policy_rejected_logits,
                 ) = self.concatenated_forward(self._model, batch)
 
+                policy_chosen_logits_mean = policy_chosen_logits.detach().mean()
+                policy_rejected_logits_mean = policy_rejected_logits.detach().mean()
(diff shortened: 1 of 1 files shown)


recipes/lora_dpo_distributed.py
```diff
@@ -620,6 +620,9 @@ def train(self) -> None:
                     policy_rejected_logits,
                 ) = self.concatenated_forward(self._model, batch)
 
+                policy_chosen_logits_mean = policy_chosen_logits.detach().mean()
+                policy_rejected_logits_mean = policy_rejected_logits.detach().mean()
+
```
felipemello1, line 625: I think that the idea to do detach.mean was so we could delete the original ones and release memory. If that is true, are we missing the deletion part?

RdoubleA: good catch

felipemello1 (approved):

PR #1338, opened 2024-08-14
Deep Fusion Modules

pbontrager:
#### Context
- [x] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

This implements modules from the Fusion Models RFC #1283 necessary for building Deep Fusion Models.

#### Changelog
This introduces the model_fusion folder in modules and the following moduels
- fusion_models.DeepFusionModel
- fusion_layer.FusionLayer
- fusion_embed.FusionEmbedding
- fusion_utils.regsiter_fusion_module

Along with these modules there is an __init__.py, a new section in the modules api docs, and a unit test for each module/function added. No existing code was edited.

#### Test plan

#### UX


torchtune/modules/model_fusion/fusion_embed.py
```diff
@@ -0,0 +1,74 @@
...
+# LICENSE file in the root directory of this source tree.
+
+from typing import List
+
+import torch
+from torch import nn, Tensor
+
+
+class FusionEmbedding(nn.Module):
+    """Fusion embedding supports training additional special tokens while keeping
```
RdoubleA, line 14: could you give an example of additional special tokens? and also an example of how a user might use this layer in their own model?

torchtune/modules/model_fusion/fusion_embed.py
```diff
@@ -0,0 +1,74 @@
...
+    the original embedding frozen. When fusing new models with a langauge model,
```
RdoubleA, line 15:
```suggestion
    the original embedding frozen. When fusing new models with a language model,
```

torchtune/modules/model_fusion/fusion_embed.py
```diff
@@ -0,0 +1,74 @@
...
+    there may be some additional tokens needed to support the fused langauge model.
```
RdoubleA, line 16:
```suggestion
    there may be some additional tokens needed to support the fused language model.
```

torchtune/modules/model_fusion/fusion_embed.py
```diff
@@ -0,0 +1,74 @@
...
+    The FusionEmbedding keeps the original embeddings frozen while learning a much smaller
+    second embedding for the additional tokens. During forward this module routes
+    the tokens to the appropriate embedding table.
+
+    Args:
+        vocab_size (int): language model vocab size
+        additional_tokens (int): additional tokens for the fused model
```
RdoubleA, line 23: nit: would prefer `num_fusion_tokens` or `additional_vocab_size` or `fusion_vocab_size`

torchtune/modules/model_fusion/fusion_embed.py
```diff
@@ -0,0 +1,74 @@
...
+            input (Tensor): input integer tensor with shape
+                [batch_size x seq_length]
+
+        Returns:
+            Tensor: output tensor embedding with shape
+                [batch_size x seq_length x embed_dim]`
+
+        """
+        bs, seq_len = input.size()
+        vocab_size = self.embedding.num_embeddings
```
RdoubleA, line 61: nit: could just save vocab_size as attribute directly in init?

torchtune/modules/model_fusion/fusion_embed.py
```diff
@@ -0,0 +1,74 @@
...
+
+        mask = input < vocab_size
+        tokens = torch.masked_select(input, mask)
+        additional_tokens = torch.masked_select(input, ~mask) - vocab_size
+
+        embeds = self.embedding(tokens)
```
RdoubleA, line 67: could you add comments indicating shapes here?

torchtune/modules/model_fusion/fusion_layer.py
```diff
@@ -0,0 +1,123 @@
...
+from torch import nn, Tensor
+
+
+class FusionLayer(nn.Module):
+    """Fusion layer as introduced in `Flamingo: a Visual Language Model for Few-Shot Learning <https://arxiv.org/abs/2204.14198>`_.
+
+    Deep Fusion model architectures combine pretrained encoder models with pretrained
+    language models by infusing the encoder outputs into the middle layers of the LLM.
+    This allows the language model to interpret the enocder outputs as text and
+    "understand" any modality where you can train an encoder. To enable the language model
```
RdoubleA, line 19: nit:
```suggestion
    "understand" any modality on which you can train an encoder. To enable the language model
```

torchtune/modules/model_fusion/fusion_layer.py
```diff
@@ -0,0 +1,123 @@
...
+    to adapt to the encoder outputs, the FusionLayer fuses a new learnable layer to an existing
+    decoder (language model) layer. This additional layer can take the encoder embeddings and
+    learn to combine them with the token embeddings from the decoder. The module supports fusing
+    the new layer before or after the original, in Flamingo the new layer is fused to the input.
```
RdoubleA, line 23: flamingo is fused before or after?
```suggestion
    the new layer before or after the original; in Flamingo the new layer is fused before the original.
```

torchtune/modules/model_fusion/fusion_layer.py
```diff
@@ -0,0 +1,123 @@
...
+
+    The original layer is wrapped in FusionLayer such that it maintains it's original state_dict
```
RdoubleA, line 25: tsk tsk
```suggestion
    The original layer is wrapped in FusionLayer such that it maintains its original state_dict
```

torchtune/modules/model_fusion/fusion_layer.py
```diff
@@ -0,0 +1,123 @@
...
+    key and the pre-trained checkpoint isn't broken. The new layer parameters are available
+    through fusion_params to separately control if they're trainable or not.
```
RdoubleA, line 27:
```suggestion
    through ``fusion_params`` to separately control if they're trainable or not.
```

torchtune/modules/model_fusion/fusion_layer.py
```diff
@@ -0,0 +1,123 @@
...
+                [batch_size x seq_length x embed_dim]
+            **kwargs (Dict): all additional layer args
+
+        Returns:
+            Tensor: output tensor with same shape as input
+                [batch_size x seq_length x embed_dim]`
+
+        """
+        if self.fusion_first:
+            x = self.fusion_layer(x, **kwargs)
```
RdoubleA, line 118: should the kwargs be shared across both the original layer and the fusion layer? can they be different types of nn.Modules?

torchtune/modules/model_fusion/fusion_models.py
```diff
@@ -0,0 +1,114 @@
...
+
+import torch
+from torch import nn, Tensor
+from torchtune.modules import TransformerDecoder
+
+
+class DeepFusionModel(nn.Module):
+    """DeepFusion is a type of fused model architecture where a pretrained encoder is combined
+    with a pretrained decoder (LLM). This is a popular architecture for multimodal models, with
+    a full overview availble in `The Evolution of Multimodal Model Architectures <https://arxiv.org/abs/2405.17927>`_.
```
RdoubleA, line 17:
```suggestion
    a full overview available in `The Evolution of Multimodal Model Architectures <https://arxiv.org/abs/2405.17927>`_.
```

torchtune/modules/model_fusion/fusion_models.py
```diff
@@ -0,0 +1,114 @@
...
+
+    This module has the same methods and forward signature as TransformerDecoder and can be used
```
RdoubleA, line 19:
```suggestion
    This module has the same methods and forward signature as :class:`~torchtune.modules.TransformerDecoder` and can be used
```

torchtune/modules/model_fusion/fusion_models.py
```diff
@@ -0,0 +1,114 @@
...
+    interchangeably where TransformerDecoder is. It combines the encoder with the decoder as a
```
RdoubleA, line 20: same comment

torchtune/modules/model_fusion/fusion_models.py
```diff
@@ -0,0 +1,114 @@
...
+        # in the decoder cache.
+        encoder_embed = None
+        if encoder_input is not None:
+            encoder_embed = self.encoder(**encoder_input)
+
+        # input_pos specifies the sequence position of the provided tokens
+        # we slice the encoder_mask to only include those same positions
+        if input_pos is not None and encoder_mask is not None:
+            encoder_mask = encoder_mask[:, input_pos]
+        output = self.decoder(tokens, mask, encoder_embed, encoder_mask, input_pos)
```
RdoubleA, line 113: nit: use keyword arguments explicitly here

torchtune/modules/model_fusion/fusion_utils.py
```diff
@@ -0,0 +1,26 @@
...
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+import functools
+from typing import List
+
+from torch import nn
+
+
+def register_fusion_module(module: nn.Module):
```
RdoubleA, line 13: an example here would be helpful, especially to understand how this is used in combination with the other fusion classes

RdoubleA (commented):
very clean overall. excited to see this come together with the model builders.

tests/torchtune/modules/model_fusion/test_fusion_embed.py
```diff
@@ -0,0 +1,75 @@
...
+    @torch.no_grad()
+    def test_forward(self, embed, vocab_size, additional_tokens, dim):
+        """
+        Test that the forward pass of the FusionEmbedding works as expected.
+        """
+        tokens = torch.randint(0, vocab_size + additional_tokens, (2, 10))
+        out = embed(tokens)
+
+        assert out.shape == (2, 10, dim)
+        assert_expected(out.mean(), torch.tensor(0.3409), atol=1e-3, rtol=1e-3)
```
kartikayk, line 55: noob question: where's this value coming from?

tests/torchtune/modules/model_fusion/test_fusion_layer.py
```diff
@@ -0,0 +1,121 @@
...
+        """
+        fused_layer = FusionLayer(layer, fusion_layer, fusion_first=False)
+
+        x = torch.rand((1, dim))
+        out = fused_layer(x)
+
+        assert out.shape == (1, dim)
+        assert_expected(out.mean(), torch.tensor(-0.0816), atol=1e-3, rtol=1e-3)
+
+    def test_get_and_load_state_dict(self, fused_layer):
```
kartikayk, line 86: nice!

torchtune/modules/model_fusion/fusion_embed.py
pbontrager: Each embedding keeps its own size already, it feels redundant for us to as well. Though I should probably add the property "num_embeddings" so this could be used like an embedding.

torchtune/modules/model_fusion/fusion_embed.py
```diff
@@ -0,0 +1,74 @@
...
+            Tensor: output tensor embedding with shape
+                [batch_size x seq_length x embed_dim]`
+
+        """
+        bs, seq_len = input.size()
+        vocab_size = self.embedding.num_embeddings
+
+        mask = input < vocab_size
+        tokens = torch.masked_select(input, mask)
+        additional_tokens = torch.masked_select(input, ~mask) - vocab_size
```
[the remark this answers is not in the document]
pbontrager: I didn't think of this situation. The main reason to use this module is because you need to add extra tokens that the decoder training didn't account for because it wasn't trained to be fused with another model. I think if you had extra "capacity" in your original embedding table you could use those but you'd have to unfreeze and update it. What would generalizing this look like, defining a range of integers where the fusion embedding is used?

I feel it'd be easier to explain this consideration in the docstring instead of trying to generalize for that case for now, and update this module if needed.

torchtune/modules/model_fusion/fusion_layer.py
pbontrager: This is a layer meant to be used in TransformerDecoder. So it would be expected to follow the same input/output. I'd be in favor of creating a TransformerLayer protocol though. @ebsmothers

torchtune/modules/model_fusion/__init__.py
```diff
@@ -0,0 +1,17 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+from .fusion_embed import FusionEmbedding
```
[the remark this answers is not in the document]
pbontrager: I was following the peft folder's lead here but I could update the files.

torchtune/modules/model_fusion/fusion_embed.py
```diff
@@ -0,0 +1,74 @@
...
+
+        embeds = self.embedding(tokens)
+        additional_embeds = self.fusion_embedding(additional_tokens)
+
+        out = self._fused_embed(bs, seq_len)
+        mask = mask.unsqueeze(-1).expand(bs, seq_len, self.dim)
+        out.masked_scatter_(mask, embeds)
```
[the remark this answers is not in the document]
pbontrager: I have "out" which shape [b x s x d]. I also have embeds and additional_embeds who's combined sequence length equals the full sequence length. Masked scatter updates "out" wherever mask is true.

torchtune/modules/model_fusion/fusion_models.py
```diff
@@ -0,0 +1,114 @@
...
+    a full overview availble in `The Evolution of Multimodal Model Architectures <https://arxiv.org/abs/2405.17927>`_.
+
+    This module has the same methods and forward signature as TransformerDecoder and can be used
+    interchangeably where TransformerDecoder is. It combines the encoder with the decoder as a
+    single module for checkpointing and finetuning. It is expected that the encoder and decoder
+    are already defined with any extra learnable "fusion_params"; learnable parameters to help
+    adapt the pre-trained encoder to the pre-trained decoder.
+
+    Example::
+        >>> model = DeepFusionModel(LLama3(), CLIP())
```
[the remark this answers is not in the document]
pbontrager: It's just meant to be a stand in for any decoder language model. I could use LLM() instead.

tests/torchtune/modules/model_fusion/test_fusion_embed.py
pbontrager: This is just grabbed from the first run. There's nothing specific that needs to be parity checked here, I just want to make sure that this value never changes in the future.

docs/source/api_ref_modules.rst
```diff
@@ -65,6 +65,20 @@ PEFT Components
...
+Components for building models that are a fusion of two+ pre-trained models
+
+.. autosummary::
+    :toctree: generated/
+    :nosignatures:
+
+    model_fusion.DeepFusionModel
+    model_fusion.FusionLayer
+    model_fusion.FusionEmbedding
+    model_fusion.regsiter_fusion_module
```
joecummings, line 79: typo police

joecummings (commented):
Why do all these need to be in separate files? We don't separate out these in the LLM world and I think Soumith said at one point (and I agree) that too many redirects can be fatiguing to understanding what the code is actually doing.

torchtune/modules/model_fusion/fusion_models.py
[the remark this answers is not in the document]
kartikayk: maybe just use one of our builder functios so this isn't confusing?

kartikayk (approved):
Stamping this since it overall looks good, but please address all of the outstanding comments!

kartikayk:
@joecummings my guess is that each of these files are going to include many more classes, especially as we add support for early fusion models (extrapolating from the RFC). If this is true, it might make sense to keep these lightweight

torchtune/modules/model_fusion/fusion_embed.py
```diff
@@ -0,0 +1,74 @@
...
+        Returns:
+            Tensor: output tensor embedding with shape
+                [batch_size x seq_length x embed_dim]`
+
+        """
+        bs, seq_len = input.size()
+        vocab_size = self.embedding.num_embeddings
+
+        mask = input < vocab_size
+        tokens = torch.masked_select(input, mask)
```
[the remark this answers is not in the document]
pbontrager: Add the shape comment. And yes it's [bs x num_token] where num_tokens = (input < vocab_size).sum()

Issue #1343, opened 2024-08-14
Decouple dataset from prompt template


joecummings:
Right now if I want to use a Llama2 model with a custom dataset I have to tie that dataset directly to the Llama2 prompt template in code. I can't change it from the config. This means my experiments are incredibly hard to maintain. In addition, the prompt template from training should be used in inference and therefore I have to implement it again? Huge potential for silent bugs here.

In code:

```python
def custom_dataset(source, tokenizer, **kwargs):
    # The 'user' column is the input, and the 'code' column is the output.
    message_transform = InputOutputToMessages(
        train_on_input = False,
        column_map = {
            "input": "statement",
            "output": "code",
        }
    )

    prompt_template = Llama2PromptTemplate()

    # Build our dataset!
    return SFTDataset(
        source=source,
        message_transform=message_transform,
        model_transform=tokenizer,
        prompt_template=prompt_template,
        split=split,
    )
        
```

Ideally, I'd like to be able to do something like this in the config:

```yaml
dataset:
  _component_: path.to.custom_dataset
  prompt_template: torchtune.models.llama2.Llama2ChatTemplate
  split: sample
```

cc @RdoubleA

pbontrager:
I agree it should be decoupled from the dataset so people can customize it. But I don't think it should be a dataset parameter at all, it should be part of the model_transform/tokenizer. So for Joe's example:

```yaml
model_transform:
  _component_: torchtune.models.llama3.llama3_tokenizer
  prompt_template: torchtune.models.llama2.Llama2ChatTemplate
  max_seq_len: 8192
  path: /tmp/llama3/tokenizer.model
```

This allows it to follow the model for eval and generate where you won't have the dataset.

[Off topic] I'm uncomfortable with the tokenizer truncating sequences. That feels like a dataset responsibility. I feel this could lead to unintuitive behavior during decoding.

RdoubleA:
Good flagging this. This was possible in the old instruct/chat builders. I am currently working on revamping those with the new SFTDataset. In my mind the ideal behavior would be I can specify a custom prompt template directly from config or specify a string to an existing prompt template.

```
dataset:
  _component_: path.to.custom_dataset
  prompt_template: torchtune.models.llama2.Llama2ChatTemplate
  split: train
```
or
```
dataset:
  _component_: path.to.custom_dataset
  prompt_template:
    user:
    - "Question: "
    - " Answer: "
  split: train
```
this requires a lot of awkward type checking logic though

[...]

PR #1347, opened 2024-08-14
Move prompt templating to tokenizer

RdoubleA:
#### Context
Prompt templating is often associated with the model or the finetuning task and needs to be carried over to inference/eval. To make this easier, we move prompt template as part of the tokenizer instead of the dataset.

Addresses #1343.

#### Changelog
- Update `ChatMLTemplate` for use in Qwen2Tokenizer. Simplify Qwen's tokenize_messages with `ChatMLTemplate`
- Add prompt template argument to all model tokenizers and builder functions
- Remove prompt template from all dataset classes and builders
- Add utility to define a prompt template from config via a string to the class name or a custom dict in the config
- Update all tests
- Add `__call__` to Qwen2Tokenizer as per #1353 

#### Test plan

#### UX


torchtune/data/_prompt_templates.py
```diff
@@ -226,3 +242,29 @@ def __call__(
 
 Please see :class:`~torchtune.data.PromptTemplate` for full API arguments.
 """
+
+
+def _get_prompt_template(
```
joecummings, line 247: This is super cool - we'll need some examples to make it clear to the user all the ways this can work.

torchtune/models/qwen2/_tokenizer.py
```diff
@@ -10,7 +10,7 @@
```
joecummings, line 10: Is there a prompt template file missing or is that coming in a separate PR?

torchtune/data/_prompt_templates.py
```diff
@@ -161,12 +163,24 @@ def __call__(
             The formatted list of messages
         """
         formatted_dialogue = []
-        for message in messages:
-            content = (
-                [{"type": "text", "content": self.template[message.role][0]}]
-                + message.content
-                + [{"type": "text", "content": self.template[message.role][1]}]
-            )
+        for index, message in enumerate(messages):
```
joecummings, line 166: Was this a bug from before?

joecummings (approved):
I see no big issues with this - just one question and waiting for green CI.

torchtune/data/_prompt_templates.py
RdoubleA: kind of, see this discussion: https://github.com/pytorch/torchtune/pull/1019/files#r1621540801. it was also called out in the Qwen2 PR (#1143): 
>The official implemtation torchtune.data._chat_formats.ChatMLFormat will add an extra <|IM_END|> tag at the end of the last assistant message in generation (see [this discussion](https://github.com/pytorch/torchtune/pull/1019/files#r1621540801)), so we override the tokenize_message() method to fix this problem (will not add <|IM_END|> and EOS token if messages[-1].role == 'assistant' and not messages[-1].content)

PR #1348, opened 2024-08-14
Update peak memory stats to be logged in GiB

joecummings:
#### Context

The difference between GB and GIB is usually pretty small and can be used interchangeably. However, this causes a strange issue when really pushing the boundaries of available memory. For example, we could report that a finetuning run has used up 83 GB of memory on an 80 GB machine. If I'm a user seeing that, I'm thinking that there's a glitch or a bug. 

If we report in GiB, we show that the user is only using ~ 77 GiB of an 80 GiB machine.

#### Changelog
* Log in GiB, not GB

#### Test plan

Check my math:
https://stackoverflow.com/questions/33837689/convert-bytes-into-gigabytes-with-javascript-math

Peak stats when starting for code-llama2 7B now:

```
GPU peak memory allocation: 13.19 GiB
GPU peak memory reserved: 13.27 GiB
GPU peak memory active: 13.19 GiB
```

### DIFF
--- a/torchtune/utils/memory.py
+++ b/torchtune/utils/memory.py
@@ -207,9 +207,9 @@ def get_memory_stats(device: torch.device, reset_stats: bool = True) -> dict:
             f"Logging memory stats is only supported on CUDA devices, got {device}"
         )
 
-    peak_memory_active = torch.cuda.memory_stats().get("active_bytes.all.peak", 0) / 1e9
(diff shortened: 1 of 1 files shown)


RdoubleA (approved):

felipemello1 (approved):

joecummings:
@pytorch-bot merge

joecummings:
> @pytorch-bot merge

Darn - was worth a try.

Issue #1308, opened 2024-08-12
Add LR Scheduler to full finetune recipes


joecummings:
Following [how we do it in LoRA recipes](https://github.com/pytorch/torchtune/blob/afd23fd0b2f9051958affae20890396e2594756f/recipes/lora_finetune_distributed.py#L475), we should add the ability to use a LR scheduler to the full finetune recipes

Here's everything that would need to be changed:
- [ ] Add appropriate functions to recipes/full_finetune_distributed.py (follow recipes/lora_finetune_distribute.py for example)
- [x] Add appropriate functions to recipes/full_finetune_single_device.py (follow recipes/lora_finetune_single_device.py for example)
- Run an example run with Llama3.1 8B and log to Weights & Biases. Post screenshot confirming this works in your PR

I don't think we need to update all our configs yet, just need to make sure the functionality exists.

felipemello1:
@joecummings , I was thinking that this could be a function in utils that takes an optimizer as input, and just does the step for, that way it can be shared for any recipe. Maybe it need a flag for "is_distributed".

This complicates the task, but maybe we could follow a similar pattern for model checkpointing and early stopping, which are also frequently asked features.

joecummings:
We CaN aLWayS dO iT IN a FollOw Up REfaCtOR

joecummings:
Kidding, I know that will never happen. I like the idea.

user074:
I guess it kinda work for the full_finetune_single_device.py. I will update after I finish the training to see how it goes
<img width="1245" alt="Screenshot 2024-08-13 at 12 39 08 AM" src="https://github.com/user-attachments/assets/59e8f7bf-3631-40b1-92da-bf9983c5bdd7">

SalmanMohammadi:
Can we close this?

ebsmothers:
@SalmanMohammadi technically I think we still need to add to the distributed recipe to fully close it out

user074:
I can have a check on the distributed case but probably after mid November

felipemello1:
fyi, reference implementation from a fork: https://github.com/homebrewltd/torchtune/blob/1ff4fc217e578a1656dc60c343dcee98e7ac013e/recipes/full_finetune_fsdp2.py#L285

PR #1350, opened 2024-08-15
Add LR Scheduler to single device full finetune

user074:
#### Context
- [x] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

Please link to any issues this PR addresses: #1308 

#### Changelog
I added LR scheduler to single device full fine tune (full_finetune_single_device.py). For example we can add following to the config of 8B_full_single_device.yaml:
```
optimizer:
  _component_: bitsandbytes.optim.PagedAdamW8bit
  weight_decay: 0.01
  lr: 1e-5

lr_scheduler:
  _component_: torchtune.modules.get_cosine_schedule_with_warmup
  num_warmup_steps: 500
  
```

If _optimizer_in_bwd, then I create a dummy optimizer in order to update
If not then I just follow lora_funetune_single_device.py

Be aware that 2e-5 of get_cosine_schedule_with_warmup with PagedAdamW8bit would cause instability, and I suspect it might due to momentum. 1e-5 would be stable in my case.

Screenshot 2024-08-14 at 8 31 14 PM

#### Test plan

#### UX


ebsmothers:
Hi @user074 thanks for this PR! Sorry, somehow the review here seems to have fallen through the cracks a bit. Overall the implementation looks reasonable though. One request: for the case where optimizer_in_bwd is True, can we offload the logic into a standalone utility (e.g. in `utils/memory.py` so that it's colocated with the `OptimizerInBackwardWrapper`? Maybe a signature like 

```
def get_lr_scheduler_for_optim_in_bwd(
	lr_scheduler: LRScheduler,
	optim_in_bwd_wrapper: OptimizerInBackwardWrapper
) -> LRScheduler:
```

(This is just one suggestion, open to any alternative ideas if you think this can be done more clearly another way)

Also I think some tests are failing, you may need to update the test configs [here](https://github.com/pytorch/torchtune/blob/05aeb71bec637f26a5b14e16638d445c1f7a8deb/tests/recipes/test_full_finetune_single_device.py#L47). Once you do that you can also run the tests locally via `pytest tests -m integration_test` (do let me know if you need any other help on this though)

user074:
Thanks @ebsmothers. I will update it by the weekend.
Additional note: I can try to do the same implementation for the distributed version but i do not have enough GPUs to test it out. Do you have any suggestions on how to test to ensure it works for the distributed version?

ebsmothers:
> Thanks @ebsmothers. I will update it by the weekend. Additional note: I can try to do the same implementation for the distributed version but i do not have enough GPUs to test it out. Do you have any suggestions on how to test to ensure it works for the distributed version?

@user074 good question. Actually the implementation of the distributed version may be more straightforward given that we don't have the option to fuse optimizer and backward there. We may be able to help out with testing, if you share a repro command for the loss curves in the PR summary we can use that for the distributed recipe to ensure we get similar behavior there.

recipes/configs/llama2/7B_full_low_memory.yaml
```diff
@@ -53,7 +53,10 @@ batch_size: 2
 epochs: 3
 optimizer:
   _component_: bitsandbytes.optim.PagedAdamW
-  lr: 2e-5
+  lr: 1e-5
+lr_scheduler:
+  _component_: torchtune.modules.get_cosine_schedule_with_warmup
+  num_warmup_steps: 1
```
ebsmothers, line 59: I saw your commit about needing to do this to pass the test. Which test is that referring to? I don't think we wanna make this change to our configs unless we know the results will be better, plus it's kinda weird to say we are using a scheduler "with warmup" when the warmup is just one step

user074: > I saw your commit about needing to do this to pass the test. Which test is that referring to? I don't think we wanna make this change to our configs unless we know the results will be better, plus it's kinda weird to say we are using a scheduler "with warmup" when the warmup is just one step

It should be `test_loss` of `tests/recipes/test_full_finetune_single_device.py`. So when I ran the test, the increase of the warm up steps would result higher loss tolerance of loss values than expected loss discrepancy. 
`torch.testing.assert_close(
            loss_values, expected_loss_values, rtol=1e-4, atol=1e-4
        )` 
I suspected that might due to warm up steps. However, if I change the tolerance the value might be too high. It might even reach 1e-1 level. However when warmup step is large we are effectively have a much smaller lr. Observed from my own training this might cause initial loss to have large differences. As we can see from my own training the initial loss have like 2e-1 difference but the difference would shrink with more training steps after warmup finished. I havent find a good way to resolve this though. Do you have any suggestions? I hope it isnt some bug from my code causing the tolerance difference thats' why I tested it with 1 warmup step...

Screenshot 2024-09-05 at 7 00 44 AM

ebsmothers: @user074 ah I see, I think I was a bit confused about the exact issue but it makes sense now. Note that our recipe tests actually use test-specific overrides; that way we can decouple the configs we use for real fine-tunes from what we use for testing. You can see the overrides [here](https://github.com/pytorch/torchtune/blob/d31649e75e51cf15dbdca004288c32550e8f7f8d/tests/recipes/test_full_finetune_single_device.py#L36-L50) -- I would suggest just adding an override like `lr_scheduler=torch.optim.ConstantLr` there. Hopefully that'll fix things!

user074: Do you know how to deal with 'num_warmup_steps' when I set to ConstantLr. It seems to give error due to this extra argument. But I couldnt find a way to remove this keyword argument from test script. model_config does not contain it explicitly so I cannot remove.

user074: Still have problem with the larger loss with the ConstantLR. Kinda odd since warmup steps shouldnt have any affect now. What is left is OptimizerInBackwardWrapper behavior. Maybe I implemented something wrong there?

Screenshot 2024-09-07 at 11 00 10 AM
Screenshot 2024-09-07 at 11 09 33 AM

SalmanMohammadi: I noticed you changed the learning rate in the config? Could that be it?

user074: > I noticed you changed the learning rate in the config? Could that be it?

Not really since the test recipe overrides the lr to 2e-5. I also just tested the 2e-5 as the original lr. It is the same problem with loss discrepancy.

SalmanMohammadi: Looks to be working on the build machines. Could you push your latest changes, even if they break the tests, and I can check it out?

ebsmothers: Agree it'd be helpful to just push the changes and see what's happening. But also if using ConstantLR make sure you're setting `factor=1`

torchtune/utils/memory.py
```diff
@@ -136,6 +137,62 @@ def get_optim_key(self, key: str) -> Any:
...
+            RuntimeError: If the LR scheduler has not been set.
+        """
+        if self.lr_scheduler:
+            self.lr_scheduler.step(epoch)
+        else:
+            raise RuntimeError(
+                "LR scheduler has not been set. Call set_lr_scheduler first."
+            )
+
+    def get_last_lr(self):
```
SalmanMohammadi, line 179: nit:
```suggestion
    def get_last_lr(self) -> float:
```

Have you correctly configured your pre-commit hooks? Our linters should pick this up

recipes/full_finetune_single_device.py
```diff
@@ -429,6 +436,45 @@ def _setup_optimizer(
             log.info("Optimizer is initialized.")
             return optimizer
 
+    def _setup_lr_scheduler(
+        self,
+        cfg_lr_scheduler: DictConfig,
+        num_training_steps: int,
+        last_epoch: int,
+    ) -> Optional[Optimizer]:
```
SalmanMohammadi, line 444: Could you add docstrings here?

torchtune/utils/memory.py
ebsmothers: > Have you correctly configured your pre-commit hooks? Our linters should pick this up

I think there are merge conflicts, so linters will not run in CI until you resolve this. Also for missing type hints idk if flake8 actually catches those anyways? (Not positive though)

ebsmothers:
Hi @user074 just wanted to check in if you're still working on this. If not let me know and one of us can help to take it over the finish line (I think from here there should be very little left)

user074:
> Hi @user074 just wanted to check in if you're still working on this. If not let me know and one of us can help to take it over the finish line (I think from here there should be very little left)



> Hi @user074 just wanted to check in if you're still working on this. If not let me know and one of us can help to take it over the finish line (I think from here there should be very little left)

Just updated suggested changes. Sorry for the delay

ebsmothers:
@user074 no problem. Can you also merge latest changes from main into your branch? There are merge conflicts and we can't run CI until those get resolved

recipes/full_finetune_single_device.py
```diff
@@ -418,6 +425,58 @@ def _setup_optimizer(
...
+                optimizer,
+                num_training_steps=num_training_steps,
+                last_epoch=last_epoch,
+            )
+        else:
+            lr_scheduler = config.instantiate(
+                cfg_lr_scheduler,
+                optimizer,
+                last_epoch=last_epoch,
+            )
```
ebsmothers, line 471: Ah is this a consequence of using `ConstantLR` in the test? I think it's a bit confusing to condition on whether the component is coming from torchtune.modules or not. In general there will be things from torch.optim that don't match the signature in the else statement, and there will probably also (in the future) be things from torchtune.modules that don't match the signature in the if statement.

In that case (and sorry for changing my mind on this), is there any reason we can't just make `lr_scheduler` optional? Then if it's not passed we just continue to use the usual constant learning rate defined in the optimizer and set `self._lr_scheduler=None` in the recipe (and in the training loop check that it's not None prior to stepping). This will make the recipe test a non-issue and keep the setup clean. 

Separately though it's clear that our LR scheduler instantiation is not flexible enough. We should figure out how to better support schedulers from torch.optim moving forward (though that's a problem for another PR).

user074: Yes. I was thinking to make it more flexible also and support other types of torch.optim.lr_scheduler. I find that torchtune.modules.lr_scheduler only has get_cosine_schedule_with_warmup and num_training_steps input would not be compatible with other torch.optim.lr_scheduler. 

I can check and work with lr_scheduleer as None by the weekend. It is mainly due to the testing config that it always would input num_training_steps even when there is no args for that.

user074: Updated `self._lr_scheduler=None`
Also I modified the testing. I still use the cosine instead of constant but effectively change it equivalent to constant by setting `num_cycles=0`. It should be a bit more elegant imo

recipes/full_finetune_single_device.py
```diff
@@ -586,6 +640,9 @@ def train(self) -> None:
                         self._optimizer.step()
                         self._optimizer.zero_grad(set_to_none=True)
 
+                    # Need to fix `lr_scheduler.step()` before `optimizer.step()` warning
```
SalmanMohammadi, line 643: What is this comment referring to? : )

recipes/full_finetune_single_device.py
```diff
@@ -267,6 +267,13 @@ def setup(self, cfg: DictConfig) -> None:
             self._steps_per_epoch = self.max_steps_per_epoch
         self.global_step = self.epochs_run * self._steps_per_epoch
 
+        # Setup lr scheduler
+        self._lr_scheduler = self._setup_lr_scheduler(
+            cfg_lr_scheduler=cfg.lr_scheduler,
```
ebsmothers, line 272: I might be wrong but I think this will throw an attribute error if LR scheduler is not passed, should probably do something like this instead

```suggestion
            cfg_lr_scheduler=cfg.get("lr_scheduler", None),
```

ebsmothers (commented):
Thanks, I think this is much cleaner! I left one more comment, can you run a test with no LR scheduler (running manually is fine) just to make sure nothing is broken after applying my suggestion? After that I think this should be good to go

user074:
> Thanks, I think this is much cleaner! I left one more comment, can you run a test with no LR scheduler (running manually is fine) just to make sure nothing is broken after applying my suggestion? After that I think this should be good to go

I just ran a llama 2 7B without lr_scheduler and it works.

ebsmothers (approved):
Looks good, thanks for adding this!

Issue #1278, opened 2024-08-07
[RFC] Optimizer CPU offload from torchao for single GPU low memory config


gau-nernst:
The recent addition of optimizer CPU offload in torchao can be useful for single GPU low memory config.

https://github.com/pytorch/ao/tree/main/torchao/prototype/low_bit_optim#optimizer-cpu-offload

In my brief testing https://github.com/pytorch/torchtune/compare/main...gau-nernst:torchtune:optim_offload, there is **~25% increase in tok/s**. Wandb project: https://wandb.ai/gau-nernst/torchtune. My system: 4070Ti SUPER (16GB VRAM), Ryzen 5600, DDR4.

<img width="1025" alt="image" src="https://github.com/user-attachments/assets/f114f20f-a600-4759-8ea3-8e18f368e1e8">

There is also a difference in handling gradients memory.
- For CPU offload, I use `offload_gradients=True` in `CPUOffloadOptimizer`, which free gradients once device-to-host transfer finishes.
- For paged Adam, it is done via `optimizer_in_bwd=True`.

Regarding memory usage, it's pretty strange since in nvidia-smi, paged Adam run also occupies a lot of memory (near 16GB). Perhaps because bnb manages their own unified memory so PyTorch doesn't report it? Also, for RAM usage, htop reports 55.5GB for paged Adam, and 64.1GB for offload Adam.

We probably need more testing. In particular:
- **Different system configurations**. CPU offload Adam can be dependent on RAM and CPU speed, since optim step is done on CPU. Paged Adam might be faster when there is more spare GPU memory, since paged Adam does optim step on GPU. The optimal batch size (to maximize tok/s) for each config might be different too.
- **Memory spike behavior**. For CPU offload Adam, I had to add `expandable_segments:True` to prevent OOM in the middle of training. Memory spike behavior might be unpredictable with CPU offload Adam, since it is not well tested. The spike might come from gradients offloading (ref: https://github.com/pytorch/ao/pull/584#discussion_r1704667190, not 100% sure). I haven't tested paged Adam without `expandable_segments:True` yet.

Regardless, I think adding an extra option for low memory single GPU training is beneficial, even if it is not well-tested yet.

cc @msaroufim

gau-nernst:
Profiling trace for CPU offload Adam, bs=4, Llama2-7B

![image](https://github.com/user-attachments/assets/a1945769-b9b2-4759-9acc-86c74abde779)

From the screenshot
[...]

PR #1230, opened 2024-07-26
[RFC][DOCS] Recipe [DOCS] ([DOC]umentation)

SalmanMohammadi:
What is the purpose of this PR? Is it to
- [x] add a new feature
- [ ] fix a bug
- [x] update tests and/or documentation
- [ ] other (please add here)

Let's map out a user journey here:

> I'm a l33t gamer with a [2xRTX 4090  battlestation](https://old.reddit.com/r/battlestations/comments/1e8m7g7/i_dont_do_much_battling_here_but_i_do_enjoy/). I've just beat Cyberpunk (on max settings, mind you), and now I'd like to get into fine-tuning LLMs. I'm excited about the latest release of my favourite animal-based LLM, Llama3.1. What do I want to do? Not 100% sure - fine-tuning an LLM for my Discord? The LLama 3.1 documentation has helped me find ✨torchtune✨! What can I do with it?? How can I quickly discover this? 
> Oh, cool, there's a quick tutorial for fine-tuning Llama3? And it uses LoRA? Great! And there's 🔥🔥🔥 documentation on customizing my own datasets? Amazing!!
> Now, what else can I do with this LoRA "recipe"? Isn't there a page where I can understand what all these parameters mean, at once?
> ...
> Well, what other recipes are there? What else can I do with ✨torchtune✨?
> ...

Right now, don't have a clear way to communicate to our users which recipes we support, and how to quickly configure them. Documentation for recipes is:
- Hidden inside recipe files. Recipe-specific features are listed amongst features common to all recipes. This makes the documentation a bit difficult to parse.
- Hidden inside our config files, which are also replicated across our configs. Our config files contain some of the more crucial details when using recipes; which commands to run, which levers to pull.

I want to maximise the surface area of the features we expose in torchtune. Our design philosophy keeps things flat and modular, users can swap out components, models, and datasets freely. I wish for the ML PhD and the l33t gamer to be able to discover what we offer, and how to use it, with equal ease.

My contribution addresses this in the following ways:
- I propose a [glossary of features](https://docs-preview.pytorch.org/pytorch/torchtune/1230/tutorials/memory_optimisations.html) which are common across all of our recipes, and also common amongst recipes with specialised fine-tuning features such as PEFT or FSDP(2).
[...]

PR #1351, opened 2024-08-15
Add CPU offload optimizer from torchao

gau-nernst:
#### Context
- [x] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

Please link to any issues this PR addresses. #1278 

#### Changelog

- Add CPU offload optimizer from torchao
- Add a sample config for Llama2-7B. Can be extend to other configs as well.

#### Test plan

#### UX

#### Results

**Llama2-7B**

```
tune run full_finetune_single_device --config llama2/7B_full_low_memory metric_logger._component_=torchtune.utils.metric_logging.WandBLogger batch_size=4 log_peak_memory_stats=True
tune run full_finetune_single_device --config llama2/7B_full_offload_optimizer metric_logger._component_=torchtune.utils.metric_logging.WandBLogger batch_size=4 log_peak_memory_stats=True
```

image

~30% faster than paged Adam from bnb.

Without `PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True`, at bs=4, paged AdamW does not OOM, but CPU offload AdamW OOM at step 62. So indeed the gradient offload mechanism is not perfect, as it does not guarantee gradient of the current parameter is cleared before the next backward.

**Phi3-mini-4B** (example of overriding config)

```
tune run full_finetune_single_device --config phi3/mini_full_low_memory metric_logger._component_=torchtune.utils.metric_logging.WandBLogger batch_size=8 log_peak_memory_stats=True
tune run full_finetune_single_device --config phi3/mini_full_low_memory metric_logger._component_=torchtune.utils.metric_logging.WandBLogger batch_size=8 log_peak_memory_stats=True optimizer._component_=torchtune.modules.optim.CPUOffloadOptimizer optimizer.optimizer_class=torch.optim.AdamW optimizer.fused=True optimizer.offload_gradients=True
```

image

~20% speedup.

(Speedup will be greater with bs=16, but bs=16 requires `PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True` to prevent OOM for both paged AdamW and CPU offload AdamW).

**Test resume**

Initial run

```
tune run full_finetune_single_device --config phi3/mini_full_low_memory metric_logger._component_=torchtune.utils.metric_logging.WandBLogger batch_size=8 log_peak_memory_stats=True optimizer._component_=torchtune.modules.optim.CPUOffloadOptimizer optimizer.optimizer_class=torch.optim.AdamW optimizer.fused=True optimizer.offload_gradients=True max_steps_per_epoch=100 epochs=2
```

Resume

```
tune run full_finetune_single_device --config phi3/mini_full_low_memory metric_logger._component_=torchtune.utils.metric_logging.WandBLogger batch_size=8 log_peak_memory_stats=True optimizer._component_=torchtune.modules.optim.CPUOffloadOptimizer optimizer.optimizer_class=torch.optim.AdamW optimizer.fused=True optimizer.offload_gradients=True max_steps_per_epoch=100 epochs=2 resume_from_checkpoint=True checkpointer.recipe_checkpoint=/tmp/Phi-3-mini-4k-instruct/recipe_state.pt "checkpointer.checkpoint_files=[hf_model_0001_0.pt,hf_model_0002_0.pt]"
```

image

Loss match exactly.

All benchmarks done with 4070Ti SUPER and Ryzen 5600. Caveats from https://github.com/pytorch/ao/tree/main/torchao/prototype/low_bit_optim#optimizer-cpu-offload apply.


felipemello1:
This is nice! Let me know when its ready for reviewing. n00b question: Is there any difference on peak allocated memory?

gau-nernst:
These are the graphs logged in WandB

image

Peak memory alloc match exactly. Since paged Adam allocates its own memory using unified memory, I'm not sure if other numbers reported here are correct. When I check in nvidia-smi, the numbers are roughly the same.

Also wanted to ask you. I want to add a test to make sure this optimizer can be saved and resumed correctly with torchtune. Where and how should I write such test? Should it be under integration test?

Second thing. If you have time, can you help me run this benchmark on a server also? From what I understand, CPU offload optimizer should be much faster with server CPUs thanks to faster RAM and CPU (AVX-512 and more cores). Want to check if it is true and if this recipe can be useful for server scenario also.

felipemello1:
>I want to add a test to make sure this optimizer can be saved and resumed correctly with torchtune. Where and how should I write such test? Should it be under integration test? I will mark this as ready for review once I add this test. 

I dont think that it should be under regression. It reminds me of what checkpointer does, i.e. load a model, save, load again: https://github.com/pytorch/torchtune/blob/main/tests/torchtune/utils/test_checkpointer.py

Can you create a lightweight test with a dummy model, do a couple of fwd/bwd pass, and test it in the same place you already have your other test? I just dont know if we have a flag for GPU_REQUIRED that we can use. Would that be necessary?

@joecummings @pbontrager any thoughts?

gau-nernst:
Discussed with some of the people in torchtune discord. I'm not adding such test anymore, but ran a sanity check with saving and resuming. It worked fine. I added the result in the PR description.

torchtune/modules/optim/_cpu_offload.py
```diff
@@ -0,0 +1,62 @@
...
+
+    _CPU_OFFLOAD_OPTIM_AVAILABLE = True
+
+    class CPUOffloadOptimizer(_CPUOffloadOptimizer):
+        """Offload optimizer to CPU for single-GPU training. This will reduce GPU memory by the size of optimizer
+        state. Optimizer step will be done on CPU.
+
+        Args
+            params: a list of parameters or parameter groups.
+            optimizer_class: constructor of the base optimizer.
```
felipemello1, line 22: Adding an example here would be great. E.g. 'torch.optim.AdamW', so its clear for the user how to use it. Maybe even the an instantiation example would be nice, like:

```
Example:
	from torchtune.modules.optim import CPUOffloadOptimizer
	optimizer=CPUOffloadOptimizer(
		params=model.params(), 
		optimizer_class='torch.optim.AdamW', 
		offload_gradients=False, 
		lr=2e-5
	)

File: torchtune/modules/optim/_cpu_offload.py [hunk 2]
```diff
@@ -0,0 +1,62 @@
...
+                accumulation.
+            kwargs: other keyword arguments to be passed to the base optimizer e.g. `lr`, `weight_decay`.
+
+        NOTE: This is a light wrapper around `torchao.prototype.low_bit_optim.CPUOffloadOptimizer` that parses
+        optimizer class string into a concrete optimizer class.
+        """
+
+        def __init__(
+            self,
+            params,
```
felipemello1, line 33: We should add notation, list List[torch.Tensor], or something like that. What do you think? Maybe just copy from pytorchs implementation

torchtune/modules/optim/_cpu_offload.py
```diff
@@ -0,0 +1,62 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+from torchtune.config._utils import _get_component_from_path
+
+try:
+    from torchao.prototype.low_bit_optim import (
```
felipemello1, line 10: Should we have the try/except only on the import, instead of the whole function? Then we can also simply this file

torchtune/modules/optim/_cpu_offload.py
```diff
@@ -0,0 +1,62 @@
...
+        CPUOffloadOptimizer as _CPUOffloadOptimizer,
+    )
+
+    _CPU_OFFLOAD_OPTIM_AVAILABLE = True
```
felipemello1, line 14: I am not a big fan of adding this logic here only for the sake of the test. Would it be possible to handle this check in the test file?

torchtune/modules/optim/_cpu_offload.py
```diff
@@ -0,0 +1,62 @@
...
+                optimizer_class=_get_component_from_path(optimizer_class),
+                offload_gradients=offload_gradients,
+                **kwargs,
+            )
+
+except ImportError:
+
+    _CPU_OFFLOAD_OPTIM_AVAILABLE = False
+
+    class CPUOffloadOptimizer:
```
felipemello1, line 50: i think it should just raise an error when trying to import, instead of loading a dummy Class that will raise the error later anyway. wdyt?

torchtune/modules/optim/_cpu_offload.py
felipemello1, line 1: I am not sure how I feel about this filename. If we need the "_" at the start, and what pattern we should follow for optimizers. Should it maybe be cpu_off_load_optimizer.py? (honest question)

tests/torchtune/modules/optim/test_cpu_offload.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+class TestCPUOffloadOptimizer:
+    @pytest.mark.skipif(
+        not torch.cuda.is_available(), reason="CPU offload optimizer requires CUDA"
+    )
+    @pytest.mark.parametrize("offload_gradients", [False, True])
+    def test_cpu_offload_optimizer(self, offload_gradients):
+        model_ref = nn.Sequential(
+            nn.Linear(128, 256), nn.GELU(), nn.Linear(256, 128)
```
felipemello1, line 26: for the sake of speed, would it make sense to make this smaller, e.g. 8, 16? I guess this shouldnt impact the test robustness. Is that right?

tests/torchtune/modules/optim/test_cpu_offload.py
```diff
@@ -0,0 +1,50 @@
...
+import torch
+from torch import nn
+from torchtune.modules.optim import CPUOffloadOptimizer
+from torchtune.modules.optim._cpu_offload import _CPU_OFFLOAD_OPTIM_AVAILABLE
+
+if not _CPU_OFFLOAD_OPTIM_AVAILABLE:
+    pytest.skip("CPU offload optimizer is not available", allow_module_level=True)
+
+
+class TestCPUOffloadOptimizer:
```
felipemello1, line 19: should we add a test for memory saving? Check: https://github.com/pytorch/pytorch/blob/52ad41a9c5c9cb44c54cc632433c6a5bbb23b37a/test/inductor/test_scatter_optimization.py#L184

recipes/configs/llama2/7B_full_offload_optimizer.yaml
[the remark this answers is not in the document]
felipemello1: For doc reference, take a look here: https://github.com/pytorch/torchtune/pull/1230/files#diff-b005b8f603902bc186738a4760d8575c31e9a18e997b06c4a9a17ff444c366c2

cc: @SalmanMohammadi

felipemello1 (commented):
first pass on the PR. Thanks for implementing it! I love the results. I left some comments. Beyond those, a few questions:

1. Would it be possible to support it for distributed too?
2. You said: "So indeed the gradient offload mechanism is not perfect, as it does not guarantee gradient of the current parameter is cleared before the next backward.". Is there a way to avoid random crashes in the run? For example, by preallocating memory or using GC?
3. Part of me wished that we could just use a flag "off_load_CPU_optimizer", instead of requiring the user to redefine the optimizer. But i tried to think a bit about it, and I think what you wrote makes sense. Just food for thought.

recipes/configs/llama2/7B_full_offload_optimizer.yaml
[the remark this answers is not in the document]
gau-nernst: This is simply because it is the largest model my machine can run. I can't run Llama3-8B/Llama3.1-8B. I can remove the config file.

Okay I will add instructions on how to use this optimizer in `docs/source/tutorials/memory_optimizations.rst` instead.

torchtune/modules/optim/_cpu_offload.py
```diff
@@ -0,0 +1,62 @@
...
+    _CPU_OFFLOAD_OPTIM_AVAILABLE = True
+
+    class CPUOffloadOptimizer(_CPUOffloadOptimizer):
+        """Offload optimizer to CPU for single-GPU training. This will reduce GPU memory by the size of optimizer
+        state. Optimizer step will be done on CPU.
+
+        Args
+            params: a list of parameters or parameter groups.
+            optimizer_class: constructor of the base optimizer.
+            offload_gradients: free GPU gradients once they are moved to CPU. Not compatible with gradient
```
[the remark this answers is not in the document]
gau-nernst: There are gradients memory and optimizer states memory.
- When `offload_gradients=False`, gradients memory will build up during backward pass as we do back-prop for the parameters. Optimizer states are still in CPU.
- When `offload_gradients=True`, gradients memory (on GPU) will be freed once they are copied to CPU. This mechanism is similar to `optimizer_in_bwd`.

Will add this explanation to `docs/source/tutorials/memory_optimizations.rst`.

recipes/configs/llama2/7B_full_offload_optimizer.yaml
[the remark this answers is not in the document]
felipemello1: we have some optimizations cooking, and you may be able to run llama3 8b by end of next week :)

torchtune/modules/optim/_cpu_offload.py
gau-nernst: I see some files have underscore in their names (e.g. model defs) so I follow it. I will rename it as you like.

torchtune/modules/optim/_cpu_offload.py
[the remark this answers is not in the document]
felipemello1: adding to rst is nice. But i think that also having it in the docstring is nice too.

torchtune/modules/optim/_cpu_offload.py
felipemello1: @joecummings , can you help me on this one?

torchtune/modules/optim/_cpu_offload.py
gau-nernst: How do you propose to handle it? The annoying thing is torchao nightly doesn't bump the minor version (if you install torchao nightly now, the version will be `0.4.0+gitxxxx`). And there might be patch fix also (i.e. bump to `0.4.1`, which will not include CPU offload optimizer). In other words, I have to check for `0.4` only (don't check for bug fix) and check for `+gitxxx`). Feel kinda clunky for such a small thing.

torchtune/modules/optim/_cpu_offload.py
gau-nernst: So are you proposing that if there is CPU offload optimizer from torchao, `CPUOffloadOptimizer` in torchtune shouldn't be defined at all? If it is the case, the problem is that I won't be able to import it in `__init__.py` (or I need to add extra version check/try-except logic again). Maybe I don't need to re-import it in `__init__.py`?

felipemello1: Good point! I didnt think about the init.

@joecummings, is there any case like this in torchtune?

We can do:
`from torchtune.modules.optim.cpu_offload_optimizer import CPUOffloadOptimizer`

or

`from torchtune.modules.optim import CPUOffloadOptimizer`

But in the second case, importing it in the __init__ requires the implementation of a DummyClass with NotImplementedError. Do you see anyway around it? Or should we just have the DummyClass?

gau-nernst:
> Would it be possible to support it for distributed too?

FSDP1/2 already has built-in logic for optimizer offload. There is no need for this. I originally wrote this in torchao specifically for single-GPU setup, since such functions are not available/not good for single-GPU.

> Is there a way to avoid random crashes in the run

I don't have a solution at the moment. One reason the CPU offload logic is fast because I overlap GPU->CPU transfer of gradients with backward pass. If I don't do overlapping, perhaps the memory behavior will be similar to `optimizer_in_bwd`, but it will be slower. I can try and profile it.

But anyway, I don't expect the memory usage is significantly more than paged Adam + `optimizer_in_bwd`. That's why when I set `PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True`, it didn't OOM anymore. I was testing at the edge of OOM, so it's still possible that later in the training, both optimizers will OOM (due to activations/gradients...). Generally as a rule of thumb when I train models, I leave some VRAM headroom in case there are abnormal memory spikes.

> Part of me wished that we could just use a flag "off_load_CPU_optimizer", instead of requiring the user to redefine the optimizer. But i tried to think a bit about it, and I think what you wrote makes sense. Just food for thought.

Initially I want to design it like this too. But that means adding extra logic to all current and future single-device recipes recipes. So I think just having a separate optimizer class is cleaner to maintain.

recipes/configs/llama2/7B_full_offload_optimizer.yaml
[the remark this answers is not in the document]
SalmanMohammadi: > Okay I will add instructions on how to use this optimizer in `docs/source/tutorials/memory_optimizations.rst` instead.

I'd be happy to see it in a follow up and sufficiently documented in the docstring for now to be added, until the memory optimization tutorial lands.

felipemello1:
will review tomorrow. Thanks for the changes!

SalmanMohammadi:
Thanks so much for adding this @gau-nernst. The speed improvements look great.

Regarding the importing, do we need inheritance here? Couldn't we simplify the whole file by doing something like:

```python

from functools import partial
from torchtune.config._utils import _get_component_from_path
from torch.optim.optimizer import ParamsT


def CPUOffloadOptimizer(params: ParamsT, optimizer_class: str, *, offload_gradients: bool = True, **kwargs):
        """Offload optimizer to CPU for single-GPU training. This will reduce GPU memory by the size of optimizer
        state. Optimizer step will be done on CPU.
        Args:
            params (ParamsT): a list of parameters or parameter groups.
            optimizer_class (str): name of the base optimizer.
            offload_gradients (bool): free GPU gradients once they are moved to CPU. Setting this to True will further
                reduce GPU memory by the size of gradients. This is not compatible with gradient accumulation.
            **kwargs: other keyword arguments to be passed to the base optimizer e.g. `lr`, `weight_decay`.
        Example:
            >>> from torchtune.modules.optim import CPUOffloadOptimizer
            >>> optimizer = CPUOffloadOptimizer(
            ...    params=model.parameters(),
            ...    optimizer_class='torch.optim.AdamW',
            ...    offload_gradients=False,
            ...    lr=2e-5
            ... )
        To use this optimizer from the command line, you can write
            optimizer=torchtune.modules.optim.CPUOffloadOptimizer
            optimizer.optimizer_class=torch.optim.AdamW
        NOTE: This is a light wrapper around `torchao.prototype.low_bit_optim.CPUOffloadOptimizer` that parses
        optimizer class string into a concrete optimizer class. See the documentation here for caveats
        https://github.com/pytorch/ao/tree/main/torchao/prototype/low_bit_optim#optimizer-cpu-offload
        """
    try:
        from torchao.prototype.low_bit_optim import CPUOffloadOptimizer as TorchAOCPUOffloadOptimizer

        return partial(
            TorchAOCPUOffloadOptimizer(
                params=params,
                optimizer_class=optimizer_class,
                offload_gradients=offload_gradients,
                **kwargs,
            )
        )

    except ImportError:
        raise NotImplementedError(
            "CPU offload optimizer requires torchao>=0.5 or torchao-nightly. Try update torchao or install "
            "torchao from source: pip install git+https://github.com/pytorch/ao.git"
        )

```

In your test, you can do something similar to [here](https://github.com/pytorch/torchtune/blob/2f8ed7a34457eaddce59a58e8c37dd51d6ba65f8/tests/torchtune/modules/low_precision/test_nf4_linear.py#L7) to mark skip without needing to expose a boolean from the module, because that's the only reason we needed it, right?

cc @joecummings @ebsmothers 

I also noticed the latest release of torchao is 0.4.0. Should we reflect this in the raise until the 0.5 release?

Could you also please add a sentence or two in the docstring to highlight when you might want to use this/the tradeoffs? Your insight will be invaluable if someone else wants to expose the docs for this in a tutorial.

torchtune/modules/optim/cpu_offload.py
```diff
@@ -0,0 +1,77 @@
...
+    )
+
+    _CPU_OFFLOAD_OPTIM_AVAILABLE = True
+
+    class CPUOffloadOptimizer(_CPUOffloadOptimizer):
+        """Offload optimizer to CPU for single-GPU training. This will reduce GPU memory by the size of optimizer
+        state. Optimizer step will be done on CPU.
+
+        Args:
+            params (ParamsT): a list of parameters or parameter groups.
```
SalmanMohammadi, line 22: nit: explicitly say this is usually `model.parameters()`

torchtune/modules/optim/cpu_offload.py
```diff
@@ -0,0 +1,77 @@
...
+            optimizer_class (str): name of the base optimizer.
```
SalmanMohammadi, line 23: nit: explicitly say e.g. `"torch.optim.AdamW"`.
Are there any optimizers this wouldn't work with?

gau-nernst:
I feel like we are having too much unnecessary discussion just to "parse string class name into actual class". Would you be open to modify the behavior of `_instantiate_node()` instead?

https://github.com/gau-nernst/torchtune/blob/51d5a9e3414f79ebfcace824656299c8e6e2fad4/torchtune/config/_instantiate.py#L23-L36

Perhaps we can add a new marker to tell `_instantiate_node()` to parse string into a class/object import. This would eliminate documentation and testing burden in torchtune (torchtune can still mention `torchao.prototype.low_bit_optim.CPUOffloadOptimizer` in a guide/tutorial).

Otherwise, I will apply what Salman suggested. I don't think overriding `__new__` is better in any way - you are returning an object that is not its own class! And the class itself only acts like a constructor function.

SalmanMohammadi:
> Otherwise, I will apply what Salman suggested. I don't think overriding `__new__` is better in any way - you are returning an object that is not its own class! And the class itself only acts like a constructor function.

Let's go with this for now to keep things simple. I'm happy whatever Felipe and yourself are here. Approving for now, if you could address the comments and get the OK from Felipe before merging too.

SalmanMohammadi (approved):

felipemello1:
>Perhaps we can add a new marker to tell _instantiate_node() to parse string into a class/object import. This would eliminate documentation and testing burden in torchtune (torchtune can still mention torchao.prototype.low_bit_optim.CPUOffloadOptimizer in a guide/tutorial).

Thats a great question: cant we just use it the same way we use torch.optim.Adamw, and delete all this code from torchtune? e.g.

```
optimizer:
		__component__: torchao.prototype.low_bit_optim.CPUOffloadOptimizer
```

Is the issue that it wouldnt raise a proper error?

felipemello1 (changes requested):
question about just solving it in the config

gau-nernst:
Yes, that's what I have in mind too. The problem is `'torch.optim.AdamW'` is interpreted as a string (from the yaml parser). We need a way to convert it to the actual class. Right now `_instantiate_node()` doesn't do it.
Consequently, we need a way to tell `_instantiate_node()` that we want to convert a specific string to an object (via `_get_component_from_path()`). I don't think it's a good idea for `_instantiate_node()` to try inspecting if a string can/should be converted into an object automatically.

felipemello1:
>The problem is 'torch.optim.AdamW' is interpreted as a string (from the yaml parser). We need a way to convert it to the actual class

Long shot: Would it make sense to do it in "CPUOffloadOptimizer", in torchao? Its ok if the answer is no. In that case, I will approve the PR. But I imagine that many libraries have a hydra-like config, so they will have similar problems.

torchtune/modules/optim/cpu_offload.py
```diff
@@ -0,0 +1,77 @@
...
+
+from torch.optim.optimizer import ParamsT
+from torchtune.config._utils import _get_component_from_path
+
+try:
+    from torchao.prototype.low_bit_optim import (
+        CPUOffloadOptimizer as _CPUOffloadOptimizer,
+    )
+
+    _CPU_OFFLOAD_OPTIM_AVAILABLE = True
```
ebsmothers, line 15: For what it's worth we do have [this utility](https://github.com/pytorch/torchtune/blob/9e65fa9ae67ce09e953cb68b5093d42cb6b31310/torchtune/modules/low_precision/_utils.py#L32) to help with gating on a specific ao version. Advantage being it's a bit more explicit and that way once we upgrade to the necessary ao version it'll be easy to see that we can remove the try/except

torchtune/modules/optim/cpu_offload.py
```diff
@@ -0,0 +1,77 @@
...
+            self,
+            params: ParamsT,
+            optimizer_class: str,
+            *,
+            offload_gradients: bool = False,
+            **kwargs,
+        ) -> None:
+            super().__init__(
+                params=params,
+                optimizer_class=_get_component_from_path(optimizer_class),
```
ebsmothers, line 56: As if there aren't enough cooks in the kitchen already on this discussion.. one option could be to just do this all in the recipe:

```
if hasattr(cfg.optimizer, "optimizer_class"):
	torchao_version, is_nightly = _get_torchao_version()
	if not (**some condition on ao version**):
		raise ValueError("CPU offload optimizer not supported")
		cfg.optimizer.optimizer_class = _get_component_from_path(optimizer_class)
		self._optimizer = instantiate(cfg.optimizer)
else:
	# normal optimizer instantiation
```

Not 100% sure this will work btw, please lmk if it's a dumb suggestion. But the idea is to hopefully avoid the need for a thin wrapper class just to do str -> type. 

To weigh in on the other parts of the conversation though, I do think your usage of `_get_component_from_path` is the way to go here. If we're gonna use it like this we should make it a public API though (but that doesn't have to be your problem)

torchtune/modules/optim/cpu_offload.py
gau-nernst: I'm aware of that function. But the required gating logic would be:
1. If version >= 0.5 (note that 0.4.1 will not include CPU offload optimizer, since it is a bugfix, though I don't think there will be a 0.4.1 release now)
2. OR, version == 0.4.x+gitxxxxxx i.e. install ao from source

As you can see, the check is quite burdensome, thus I want to avoid that.

`_get_torchao_version()` was mainly for handling is_fbcode and torchao-nightly, so it's not particularly helpful in this situation I think.

torchtune/modules/optim/cpu_offload.py
gau-nernst: Something like you suggested should work, but I want to avoid adding extra optimizer parsing logic to ALL recipes :sweat_smile:.

ebsmothers: Yeah that's understandable. Not that this is a sound argument, but there is precedent for doing something like this in torchtune already: see our [ConcatDataset](https://github.com/pytorch/torchtune/blob/e568b671bf4dd043649611c3d0905967257ccac2/recipes/full_finetune_distributed.py#L448-L457), which is pretty much the same in every recipe. I think this is kind of what the `_setup_*` methods are for anyways.. in an ideal world we get that nice one-line instantiation of every component, but in practice without recursive instantiation that's unlikely to happen (I am also happy to revisit that topic but iiuc it won't solve this particular case anyways)

torchtune/modules/optim/cpu_offload.py
ebsmothers: Ah good point, we need to update it to account for source installs. I've been meaning to do that but haven't gotten around to it yet. No need to worry about that here then, I think what you have is fine

gau-nernst:
Closing due to https://github.com/pytorch/ao/issues/725. To use AdamW CPU Offload from ao, simply set

```
optimizer=torchtune.modules.optim.CPUOffloadOptimizer
```

felipemello1:
Thanks for all the work on this @gau-nernst !!

SalmanMohammadi:
I wonder if you'd be interested in just contributing the tests for this @gau-nernst, since they've already been written here?

PR #1356, opened 2024-08-16
[regression test] make nightlies and stable independent from each other

felipemello1:
#### Context
- [ ] add a new feature
- [x] fix a bug
- [ ] other (please add here)

Currently when nightlies break, it also cancels stable regression tests (https://github.com/pytorch/torchtune/actions/runs/10412450496/job/28838240610)

#### Changelog
The GPT gods told me that if I add {{matrix}} to the group, then one flow failing wont affect the other.

#### Test plan
CI

### DIFF
--- a/.github/workflows/regression_test.yaml
+++ b/.github/workflows/regression_test.yaml
@@ -6,7 +6,7 @@ on:
     - cron:  '0 0 * * *'
 
 concurrency:
-  group: regression-test-${{ github.workflow }}-${{ github.ref == 'refs/heads/main' && github.run_number || github.ref }}
+  group: regression-test-${{ github.workflow }}-${{ github.ref == 'refs/heads/main' && github.run_number || github.ref }}-${{ matrix.torch-version }}
   cancel-in-progress: true
 
 permissions:


atalman:
hi @felipemello1 try adding ``fail-fast: false`` under the matrix.
Here is the example from builder repo: https://github.com/pytorch/builder/blob/main/.github/workflows/validate-linux-binaries.yml#L133

```
 linux:
    needs: generate-linux-matrix
    strategy:
      matrix: ${{ fromJson(needs.generate-linux-matrix.outputs.matrix) }}
      fail-fast: false
```

ebsmothers:
I think relying on CI for testing here won't be sufficient. You may need to hack in a trigger on pull request to get this to actually be tested in CI on your PR (then remove when ready to land). Note that you will also need to hack the test itself to not actually load the checkpoint from S3, since your fork will not have the requisite permissions (happy to provide a pointer here if necessary). Also did you try Andrey's suggestion?

felipemello1:
closed in favor of https://github.com/pytorch/torchtune/pull/1413#event-14007317860

PR #1357, opened 2024-08-16
Flamingo Model Components

pbontrager:
#### Context
- [x] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

Reimplementation of #1150 based on refactor

#### Changelog
Adds new flamingo folder

#### Test plan

#### UX


torchtune/models/flamingo/_component_builders.py
```diff
@@ -0,0 +1,227 @@
...
+the building blocks simple.
+"""
+
+def flamingo_vision_encoder(
+    # clip encoder parameters
+    patch_size: int,
+    num_heads: int,
+    clip_embed_dim: int,
+    clip_num_layers: int,
+    output_hidden_states: Optional[List[int]] = None,
```
tarun292, line 41: Should this come a little later as it's a non-default kw argument?

torchtune/models/flamingo/_encoder.py
```diff
@@ -0,0 +1,107 @@
...
+        Args:
+            x (Tensor): input tensor with shape
+                [batch_size x num_imgs x num_tiles x num_embeds x encoder_dim]
+            hidden_states (Optional[List[Tensor]]): list of hidden states
+                from the encoder. Each hidden state has the same shape as x.
+
+        Returns:
+            Tensor: output tensor of a sequence of embedings
+                [batch_size x encoder_seq_length x decoder_dim]
+                where sequence length is num_imgs*num_tiles+num_embeds
```
SalmanMohammadi, line 53: I wouldn't complain if I saw a `Notation:` section here : )

torchtune/models/flamingo/_component_builders.py
```diff
@@ -0,0 +1,257 @@
...
+    MultiHeadAttention,
+    TransformerDecoder,
+    TransformerSelfAttentionLayer,
+    Fp32LayerNorm
+)
+
+
+
+"""
+Component builders for the Flamingo model and it's constituant models.
```
SalmanMohammadi, line 25:
```suggestion
Component builders for the Flamingo model and its constituent models.
```

torchtune/models/flamingo/_component_builders.py
```diff
@@ -0,0 +1,257 @@
...
+    # projection parameters
+    num_layers_projection: int,
+    decoder_embed_dim: int,
+    # image parameters
+    tile_size: int,
+    max_num_tiles: int = 4,
+    in_channels: int = 3,
+    ) -> FlamingoEncoder:
+    """
+    Build the encoder associated with the CLIP image model with an additonal
```
SalmanMohammadi, line 51:
```suggestion
    Build the encoder associated with the Flamingo image model with an additonal
```

torchtune/models/flamingo/_component_builders.py
```diff
@@ -0,0 +1,257 @@
...
+    num_special_tokens: int,
+    num_heads: int,
+    num_kv_heads: int,
+    embed_dim: int,
+    max_seq_len: int,
+    rope_base: int = 500000.0,
+    intermediate_dim: Optional[int] = None,
+) -> TransformerDecoder:
+    """
+    Build the decoder associated with the Llama3 model with additonal fused
```
SalmanMohammadi, line 159: I'm once more slightly confused by the model naming in the docstrings, which is likely due to my lack of knowledge here. This is a builder for a flamingo decoder where the language model component is llama3?

pbontrager: Here we're using flamingo to refer to the general architecture and not the specific implementation of the original flamingo model from DeepMind. So to modernize the approach we're using llama3 and CLIP instead. This will allow us to test our deep fusion setup.

tests/torchtune/models/flamingo/test_flamingo_encoder.py
```diff
@@ -0,0 +1,130 @@
...
+            self.n_imgs
+            * self.num_tiles
+            * vision_transformer.clip.get_image_tokens_per_tile(),
+            transformer_config["decoder_embed_dim"],
+        )
+        assert (
+            output.shape == expected_shape
+        ), f"Expected shape {expected_shape}, but got {output.shape}"
+
+        assert_expected(output.mean(), torch.tensor(-10.0646), atol=1e-3, rtol=1e-3)
```
SalmanMohammadi, line 73: where did these magical numbers come from? : )

tests/torchtune/modules/test_vision_transformer.py
```diff
@@ -173,13 +176,15 @@ def test_vision_transformer_return_hidden_layers(self, transformer_config):
                 hidden_layer.shape == expected_shape_hidden_layers
             ), f"Expected shape {expected_shape_hidden_layers}, but got {hidden_layer.shape=}"
 
+        # Target based off of reference nn.TransformerEncoderLayer implementation
         assert_expected(
             torch.stack(hidden_layers, dim=-1).mean(),
-            torch.tensor(6.6938),
+            torch.tensor(8.3112),
```
SalmanMohammadi, line 182: Wizadry!!

SalmanMohammadi: I see now it's based on `nn.TransformerEncoderLayer`. Do you have a gist or something to point to by any chance?

torchtune/modules/model_fusion/_fusion.py
```diff
@@ -71,19 +71,21 @@ def _state_dict_hook(self, state_dict, *args, **kwargs):
         """
         keys = list(state_dict.keys())
         for key in keys:
-            if key.startswith("layer"):
-                new_key = key.replace("layer.", "")
+            local_key = key[len(prefix) :]
+            if local_key.startswith("layer"):
```
tarun292, line 75: I think this should be:
```
if local_key.startswith("layer.layer"):
```

torchtune/models/flamingo/_encoder.py
```diff
@@ -0,0 +1,121 @@
...
+        self.layers = _get_clones(layer, num_layers)
+        self.output = output
+        self.num_hidden = num_hidden_inputs
+
+    def forward(
+        self, x: Tensor, hidden_states: Optional[List[Tensor]] = None
+    ) -> Tensor:
+        """
+        Args:
+            x (Tensor): input tensor with shape [b x i x t x e x d]
```
SalmanMohammadi, line 45: sorry this hurt my tiny brain
```suggestion
            x (Tensor): input tensor with shape [b * i * t * e * d]
```

torchtune/models/flamingo/_encoder.py
```diff
@@ -0,0 +1,121 @@
...
+            hidden_states (Optional[List[Tensor]]): list of hidden states
+                from the encoder. Each hidden state has the same shape as x.
+
+        Returns:
+            Tensor: output tensor of a sequence of embedings [b x s x d]
```
SalmanMohammadi, line 50:
```suggestion
            Tensor: output tensor of a sequence of embedings [b * s * d]
```

torchtune/models/clip/_component_builders.py
```diff
@@ -99,3 +127,12 @@ def clip_vision_encoder(
         embed_dim=embed_dim,
         in_channels=in_channels,
     )
+
+
+def clip_mlp(in_dim: int, out_dim: int, hidden_dim: int, activation: nn.Module, quantize_base: bool = False) -> FeedForward:
+    """
+    Build the MLP layer associated with the clip model.
```
ebsmothers, line 134: Not *the* CLIP model, right? I feel like we can be a bit more explicit here

torchtune/modules/transformer.py
```diff
@@ -338,6 +339,7 @@ def __init__(
         self.num_heads = num_heads
         self.head_dim = head_dim
         self.causal_mask = None
+        self.cur_pos = None
```
ebsmothers, line 342: Can you explain what this is about?

torchtune/models/flamingo/_component_builders.py
```diff
@@ -0,0 +1,262 @@
...
+                k_proj=nn.Linear(embed_dim, num_kv_heads * head_dim, bias=False),
+                v_proj=nn.Linear(embed_dim, num_kv_heads * head_dim, bias=False),
+                output_proj=nn.Linear(embed_dim, embed_dim, bias=False),
+                q_norm=RMSNorm(dim=head_dim, eps=1e-05),
+                k_norm=RMSNorm(dim=head_dim, eps=1e-05),
+                pos_embeddings=None,
+                max_seq_len=max_seq_len,
+                is_causal=False,
+                attn_dropout=0.0,
+            )
```
ebsmothers, line 236: nbd yet but come LoRA time might want a builder for this kinda stuff

ebsmothers (commented):
Mostly a bunch of nits for now, will give a more thorough pass tomorrow though

torchtune/models/flamingo/_encoder.py
```diff
@@ -0,0 +1,125 @@
...
+class FlamingoEncoder(nn.Module):
+    """Vision encoder model for Flamingo. This combines a pretrained
+    vision encoder with a learnable projection head. The projection head
+    is converted to a fusion module and supports fusion utils.
+
+    Args:
+        clip (nn.Module): CLIP encoder vision model
+        projection_head (nn.Module): projection_head that takes embeddings
+            with dimension encoder_dim as input and outputs embeddings of
+            size decoder_dim.
```
ebsmothers, line 92: nbd to leave these both as general nn.Modules but they do have very specific signatures that make it hard to plug in any old nn.Module. Maybe point to the relevant CLIP and projection classes as an example or something

torchtune/models/flamingo/_transform.py
```diff
@@ -0,0 +1,209 @@
...
+        """
+        return self.tokenizer.tokenize_messages(
+            messages=message,
+            add_eos=add_eos,
+        )
+
+    def __call__(self, sample: Mapping[str, Any]) -> Mapping[str, Any]:
+        """
+        Apply image decoding and transformations to the "images" field in the sample
+        and tokenizization to the "messages" field in the sample. Also returns the
```
ebsmothers, line 189:
```suggestion
        and tokenization to the "messages" field in the sample. Also returns the
```

torchtune/models/flamingo/_encoder.py
```diff
@@ -0,0 +1,125 @@
...
+from typing import List, Optional
+
+import torch
+
+from torch import nn, Tensor
+from torchtune.modules.model_fusion import register_fusion_module
+from torchtune.modules.transformer import _get_clones
+
+
+class FlamingoProjectionHead(nn.Module):
```
RdoubleA, line 16: nit: is Head the right term? I would think Head is something that attached on top of the hidden states of a transformer and not a full model. I thought we had referred to this as an Adapter?

torchtune/models/flamingo/_encoder.py
```diff
@@ -0,0 +1,125 @@
...
+        self,
+        layer: nn.Module,
+        num_layers: int,
+        output: nn.Module,
+        num_hidden_inputs: int = 0,
+    ) -> None:
+        super().__init__()
+        self.layers = _get_clones(layer, num_layers)
+        self.output = output
+        self.num_hidden = num_hidden_inputs
```
RdoubleA, line 38: nit: use the same name

torchtune/models/flamingo/_transform.py
```diff
@@ -0,0 +1,209 @@
...
+
+            - Task-specific templates to gear models for a particular task that it will expect after training
+            - Model-specific templates that are required whenever the model is prompted, such as the [INST]
+              tags in Llama2 and in Mistral
+            - Community standardized templates, such as :class:`~torchtune.data.ChatMLTemplate`
+
+            The extra text will still get tokenized as normal text, not as special tokens. Default is None.
+
+    Examples:
+        >>> model_transform = FlamingoTransform("/path/to/tokenizer.model", tile_size=256)
```
ebsmothers, line 56: nit: I think you also need to provide patch_size in this example

torchtune/models/flamingo/_encoder.py
```diff
@@ -0,0 +1,125 @@
...
+
+import torch
+
+from torch import nn, Tensor
+from torchtune.modules.model_fusion import register_fusion_module
+from torchtune.modules.transformer import _get_clones
+
+
+class FlamingoProjectionHead(nn.Module):
+    """Projection transformer to adapt the output of a
```
RdoubleA, line 17: Would like to see more details here, specifically on how this is used to map from encoder hidden dim to the decoder hidden dim in the cross attention layer

ebsmothers (approved):
Sorry all I did was correct typos. But that just means the design is solid through and through, I have no major complaints there. Please do address my old comments too though. Preemptive stamp so that you're unblocked

torchtune/models/flamingo/_encoder.py
```diff
@@ -0,0 +1,125 @@
...
+        output (nn.Module): Output linear layer. Input dim is
+            (num_hidden + 1) * encoder_dim and output is decoder_dim.
+        num_hidden_inputs (int): Number of expected hidden state inputs
+    """
+
+    def __init__(
+        self,
+        layer: nn.Module,
+        num_layers: int,
+        output: nn.Module,
```
RdoubleA, line 32: should we just hardcode this to nn.Linear? I'm wondering if it makes more sense for a user to configure encoder_dim -> decoder_dim rather than the output module directly

torchtune/models/flamingo/_encoder.py
```diff
@@ -0,0 +1,125 @@
...
+        num_hidden_inputs: int = 0,
+    ) -> None:
+        super().__init__()
+        self.layers = _get_clones(layer, num_layers)
+        self.output = output
+        self.num_hidden = num_hidden_inputs
+
+    def forward(
+        self,
+        x: Tensor,
```
RdoubleA, line 42: nit: we've been using `torch.Tensor` everywhere

torchtune/models/flamingo/_encoder.py
```diff
@@ -0,0 +1,125 @@
...
+        x = x.view(bsz * imgs, tiles * embeds, dim)
+        for layers in self.layers:
+            x = layers(x)
+        x = x.view(bsz, imgs, tiles, embeds, dim)
+
+        # interleave hidden states and cat with x
+        if self.num_hidden > 0:
+            hidden_states = torch.stack(hidden_states, dim=-1)
+            hidden_states = hidden_states.view(bsz, imgs, tiles, embeds, -1)
+            x = torch.cat([x, hidden_states], dim=-1)
```
RdoubleA, line 75: is this a residual connection? If so, should note it in the docstring that the encoder hidden states are passed through like this

torchtune/models/flamingo/_encoder.py
```diff
@@ -0,0 +1,125 @@
...
+        # shape [b x s x d]
+        x = self.output(x).reshape(bsz, imgs * tiles * embeds, -1)
+
+        return x
+
+
+class FlamingoEncoder(nn.Module):
+    """Vision encoder model for Flamingo. This combines a pretrained
+    vision encoder with a learnable projection head. The projection head
+    is converted to a fusion module and supports fusion utils.
```
RdoubleA, line 86: what would this practically mean for a user?

torchtune/models/flamingo/_encoder.py
```diff
@@ -0,0 +1,125 @@
...
+
+    Args:
+        clip (nn.Module): CLIP encoder vision model
```
RdoubleA, line 89: can this ever be non-clip? does it make sense to just call this `vision_encoder`?

torchtune/models/flamingo/_transform.py
```diff
@@ -0,0 +1,209 @@
...
+class FlamingoTransform(ModelTokenizer, Transform):
+    """
+    This transform combines the transforms for the different modalities of Flamingo. It
+    is made up of the following transforms:
+    - :class:`torchtune.models.llama3.Llama3Tokenizer`
+    - :class:`torchtune.models.clip.CLIPImageTransform`
+    - :class:`torchtune.modules.transforms.VisionCrossAttentionMask`
+
+    This transform can be used as a drop-in replacement for tokenizers in recipes and generation
+    but handles additional transformations from the `__call__` method.
```
RdoubleA, line 26:
```suggestion
    but additionally handles image transformations from the `__call__` method.
```

torchtune/models/flamingo/_transform.py
RdoubleA, line 189:
```suggestion
        and tokenization to the "messages" field in the sample. Also returns the
```

torchtune/models/flamingo/_transform.py
```diff
@@ -0,0 +1,209 @@
...
+
+    def __call__(self, sample: Mapping[str, Any]) -> Mapping[str, Any]:
+        """
+        Apply image decoding and transformations to the "images" field in the sample
+        and tokenizization to the "messages" field in the sample. Also returns the
+        encoder mask.
+
+        Args:
+            sample (Mapping[str, Any]): A sample with a "tokens", "mask",
+                "encoder_input" and "encoder_mask" field to feed directly into the model.
```
RdoubleA, line 194: we don't expect encoder_input, encoder_mask to already be in sample, we should expect "images" though

torchtune/models/llama3/_tokenizer.py
```diff
@@ -25,7 +25,7 @@
     "<|eom_id|>": 128008,
     "<|eot_id|>": 128009,
     "<|python_tag|>": 128010,
-    "<|image|>": 128011,
+    "<|image|>": 128256,
```
RdoubleA, line 28: have we verified that this the official image token id...? When I first added this it was 128011

torchtune/modules/transformer.py
```diff
@@ -338,6 +339,7 @@ def __init__(
         self.num_heads = num_heads
         self.head_dim = head_dim
         self.causal_mask = None
+        self.pos = None
```
RdoubleA, line 342: is this for generation? if so, a quick comment would be helpful

torchtune/modules/transforms/_transforms.py
```diff
@@ -127,16 +127,15 @@ def __call__(self, sample: Mapping[str, Any]) -> Mapping[str, Any]:
                     (n_tiles, c, h, w) each.
 
         Returns:
-            Mapping[str, Any]: updated sample with the following keys:
-                - encoder_mask (List[torch.Tensor]): list of masks with shape (text_seq_len, image_seq_len),
-                    where length of list == number of images in sample
-                - tokens (List[int]): original tokens
-                - images (List[torch.Tensor]): original images
+            Mapping[str, Any]: sample with a new key encoder_mask, with a mask per image with shape
```
RdoubleA, line 130: why are we packaging multiple keys into encoder_input? also, you mention encoder_mask here but use encoder_input below. Would also be good to keep the bullets about tokens and images

torchtune/models/llama3/_tokenizer.py
pbontrager: Long story, but both numbers were supposed to be correct. One for finetuning and one for inference, but now it's just the one.

torchtune/models/flamingo/_transform.py
pbontrager: We do expect encoder_input to be in sample by this point, look at FlamingoTransform. I agree there is some questionable generalization here, but I think we should address that when it comes up. In summary, to allow unpacking of any arbitrary input for the encoder, we treat it as a dictionary.

torchtune/models/flamingo/_encoder.py
pbontrager: The Flamingo Encoder is a one off encoder, so it doesn't need to generalize at all and can make a lot of concrete assumptions. It's not much different than a component builder, but it's a nn.Module.

torchtune/models/flamingo/_encoder.py
pbontrager: I lost that vote :(

torchtune/models/flamingo/_encoder.py
pbontrager: I can update this, I was just following our typical pattern but this would never need to be customized by a user anyway.

torchtune/models/flamingo/_encoder.py
pbontrager: A projection heads have been around a long time and can vary a lot in architecture. The main point here is that it's learning a projection from the pre-trained encoder to the pretrained decoder.

torchtune/models/clip/_component_builders.py
```diff
@@ -23,51 +31,73 @@ def clip_vision_encoder(
...
+            output_proj=nn.Linear(embed_dim, embed_dim, bias=attn_bias),
+            pos_embeddings=None,
+            attn_dropout=0.0,
+            is_causal=False,
+    )
+    mlp = clip_mlp(
+        in_dim=embed_dim,
+        hidden_dim=4 * embed_dim,
+        out_dim=embed_dim,
+        activation=activation(),
```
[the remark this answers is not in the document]
pbontrager: What would this be called? A Class?

torchtune/modules/transformer.py
```diff
@@ -293,6 +291,8 @@ class TransformerDecoder(nn.Module):
             before final MLP.
         output (nn.Linear): Callable that applies a linear transformation to the output of
             the decoder.
+        num_layers (Optional[int]): Number of Transformer Decoder layers, only define when
+            layers is not a list.
```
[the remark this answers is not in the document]
pbontrager: This was in the TransformerLayer refactor. I left it in to support both because otherwise the builders get a lot less clean. But We could update that in the future.

torchtune/models/flamingo/_encoder.py
pbontrager: This being inside of the flamingo folder it's not really meant to be reused for other purposes. I can just set the type to be the specific modules.

torchtune/modules/transformer.py
pbontrager: This tracks position for the TransformerDecoder since the layers also track it during decoding. This is already obsolete as Salaman is updating how cacheing handles position. But I'll leave this in until that is updated. This is basically a fix for the previous update around input_pos and kvcache.

torchtune/models/clip/_component_builders.py
pbontrager: You mean CLIP ViT? I feel this is inline with naming for all our other model specific sub builders.

tests/torchtune/models/flamingo/test_flamingo_decoder.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+class TestFlamingoVisionEncoder:
+    def test_flamingo_text_decoder_initialization(self, decoder_config):
+        # Attempt to instantiate the Flamingo text decoder
+        try:
+            decoder = flamingo_decoder(**decoder_config)
+            print("Flamingo text decoder instantiated successfully.")
+        except Exception as e:
+            pytest.fail(f"Failed to instantiate Flamingo text decoder: {str(e)}")
```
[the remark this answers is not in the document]
pbontrager: It was a placeholder that's been udpated.

torchtune/models/clip/_model_builders.py
```diff
@@ -1,6 +1,6 @@
 from torchtune.models.clip._transforms import CLIPImageTransform
 
-def _clip_vit_224_transform():
+def clip_vit_224_transform():
```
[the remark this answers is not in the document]
pbontrager: There is zero reason for a builder to exist and not be public. It's not a utility but a model. The only reason it's not add to the docs yet is because there should be a model builder as well, not just the transform.

Issue #1311, opened 2024-08-12
add max_seq_len to configs


felipemello1:
In this PR (https://github.com/pytorch/torchtune/pull/1234) we changed where max_seq_len should be defined in the config.

Before, it was defined in dataset.max_seq_len, and then passed to the tokenizer:

datasets._instruct.py
```
        tokens, mask = self._tokenizer.tokenize_messages(
            messages, max_seq_len=self.max_seq_len
        )
```

But now, max_seq_len needs to be defined in the tokenizer.max_seq_len
```
 tokens, mask = self._tokenizer.tokenize_messages(
            messages,
        )
```

Task:
1) Update recipes/configs, so that this option is exposed, otherwise the training loop breaks and its not clear what the user should do (ValueError: seq_len (9274) of input tensor should be smaller than max_seq_len (8192));

e.g.
llama3.yaml
```
tokenizer:
	max_seq_len=None

```
2) remove max_seq_len from dataset args, since its not used. Another option is to raise some error if users pass dataset.max_seq_len instead. Code pointer of one of the datasets: https://github.com/RdoubleA/torchtune/blob/5c7246e05b56729324625dee8a886727cfefeac0/torchtune/datasets/_instruct.py#L56

3) add a better error message to let users know that they can update tokenizer.max_seq_len to avoid this issue. Pointer: https://github.com/RdoubleA/torchtune/blob/288ff4435b0cf17325b5c3b112f6859a6cdf0ea2/torchtune/modules/attention.py#L165;

cc: @RdoubleA

joecummings:
@felipemello1  is this done?

PR #1363, opened 2024-08-18
Adds max_seq_len to recipes and updates unit tests

thomasjpfan:
#### Context
- [ ] add a new feature
- [ ] fix a bug
- [x] other (Expose new feature in recipe yaml)

Addes task 1 of https://github.com/pytorch/torchtune/issues/1311

#### Changelog
This PR adds `max_seq_len` to the recipes and adds unit tests for loading the config

#### Test plan

### DIFF
--- a/recipes/configs/eleuther_evaluation.yaml
+++ b/recipes/configs/eleuther_evaluation.yaml
@@ -21,6 +21,7 @@ checkpointer:
 tokenizer:
   _component_: torchtune.models.llama2.llama2_tokenizer
   path: /tmp/Llama-2-7b-hf/tokenizer.model
+  max_seq_len: null
 
 # Environment
 device: cuda
--- a/tests/torchtune/config/test_instantiate.py
+++ b/tests/torchtune/config/test_instantiate.py
@@ -4,7 +4,6 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
-import os
 from pathlib import Path
 from textwrap import dedent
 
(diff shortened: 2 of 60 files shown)


recipes/configs/eleuther_evaluation.yaml
```diff
@@ -21,6 +21,7 @@ checkpointer:
 tokenizer:
   _component_: torchtune.models.llama2.llama2_tokenizer
   path: /tmp/Llama-2-7b-hf/tokenizer.model
+  max_seq_len: null
```
RdoubleA, line 24: `max_seq_length` is already defined below and is used instead of the one with the tokenizer. To be consistent, let's keep the max_seq_len here in the tokenizer, set it to 4096, remove the other one, and update the eleuther_eval recipe to use the tokenizer's max_seq_len. You simply would need to update the line here: https://github.com/pytorch/torchtune/blob/f9f75bb563ecae371492a9d49da4a9f514c081b3/recipes/eleuther_eval.py#L246

to use the tokenizer max_seq_len: `self._cfg.tokenizer.max_seq_len`

RdoubleA (commented):
Sorry for the delay, appreciate you doing this, this certainly improves the usability of all our configs. A few things:
- It seems like the mistral and phi3 configs were missed entirely? Any reason for that?
- For `mistral/7B_full_ppo_low_memory.yaml` in particular, we'd probably want to change this line: https://github.com/pytorch/torchtune/blob/f9f75bb563ecae371492a9d49da4a9f514c081b3/recipes/configs/mistral/7B_full_ppo_low_memory.yaml#L36 to be consistent with the classifier, where max_seq_len=32768 (cc @SalmanMohammadi)

SalmanMohammadi:
>     * For `mistral/7B_full_ppo_low_memory.yaml` in particular, we'd probably want to change this line: https://github.com/pytorch/torchtune/blob/f9f75bb563ecae371492a9d49da4a9f514c081b3/recipes/configs/mistral/7B_full_ppo_low_memory.yaml#L36
>        to be consistent with the classifier, where max_seq_len=32768 (cc @SalmanMohammadi)

Text completion datasets don't defer to the tokenizer for max seq length truncation right? If this is the case it's fine to leave as null - unless I'm missing something re why it should be consistent with the classifier?
In any case this is more of a recipe/task hyperparam than it is about your models (max_seq_length will be in [64, 2048] or however big you can fit). If you're doing summarization tasks you might want it to be longer, if you're doing single-turn chat you might want it to be shorter, etc.

RdoubleA:
Hi @thomasjpfan, let us know if you still plan to carry this through or if we should wrap it up. Appreciate your work so far!

thomasjpfan:
> It seems like the mistral and phi3 configs were missed entirely? Any reason for that?

I had a quick script to automate the new keyword argument and missed mistral and phi3

From https://github.com/pytorch/torchtune/pull/1363#issuecomment-2307063732, I kept the tokenizer in `mistral/7B_full_ppo_low_memory.yaml` to `max_seq_len: null`.

tests/torchtune/config/test_instantiate.py
```diff
@@ -74,3 +78,53 @@ def test_instantiate(self, config, module):
         del config.test.dim
         actual = instantiate(config.test, 3)
         assert self.get_dim(actual) == 3
+
+
+ASSETS = Path(__file__).parent.parent.parent / "assets"
+
+
+class TestInstantiateTokenizer:
```
RdoubleA, line 86: Appreciate you adding this test, but actually don't think this is necessary. This is simply testing that an object can be instantiated and has the correct parameters, but this is already covered by the other tests in this directory

RdoubleA (commented):
Looks great! Just one comment and it's ready to go

tests/torchtune/config/test_instantiate.py
thomasjpfan: I added the test to mostly check for the null case, which we couldn't do with `RMSNorm`.

In [`4eeb1fd` (#1363)](https://github.com/pytorch/torchtune/pull/1363/commits/4eeb1fd99d6e54335ce492b7023233b764a2e6a1), I simplified the test to only check the null case.

RdoubleA (approved):

PR #1366, opened 2024-08-19
Add system prompt

RdoubleA:
#### Context
- [ ] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

Enable adding an optional system prompt in our built-in message transforms. Expose this parameter in our dataset builders. Any user dataset that uses our common message transforms can leverage this feature.

#### Test plan

#### UX

### DIFF
--- a/torchtune/data/_messages.py
+++ b/torchtune/data/_messages.py
@@ -118,7 +118,7 @@ class Message:
 
 class InputOutputToMessages(Transform):
     """
-    Message transform class that converts a sample with "input" and "output" fields,
+    Message transform class that converts a single sample with "input" and "output" fields,
     (or equivalent fields specified in column_map) to user and assistant messages,
     respectively. This is useful for datasets that have two columns, one containing
     the user prompt and the other containing the model response.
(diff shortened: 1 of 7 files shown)


torchtune/data/_messages.py
```diff
@@ -129,13 +129,19 @@ class InputOutputToMessages(Transform):
...
+        system_prompt (Optional[str]): if specified, prepend a system message. This can
+            serve as instructions to guide the model response. Default is None.
     """
 
     def __init__(
-        self, train_on_input: bool = False, column_map: Optional[Dict[str, str]] = None
+        self,
+        train_on_input: bool = False,
+        column_map: Optional[Dict[str, str]] = None,
+        system_prompt: Optional[str] = None,
```
joecummings, line 140: Need some way to convey that this will either add or override the given system prompt:

```new_system_prompt```?

File: torchtune/datasets/_grammar.py [hunk 2]
```diff
@@ -19,6 +19,7 @@ def grammar_dataset(
     source: str = "liweili/c4_200m",
     column_map: Optional[Dict[str, str]] = None,
     train_on_input: bool = False,
+    system_prompt: Optional[str] = None,
```
joecummings, line 22: I think these need to be updated to or at least the documentation stating that this will override any system prompt?

RdoubleA: oi, missed these

joecummings (approved):

Issue #1352, opened 2024-08-15
create import protection for torchvision


felipemello1:
in utils, it imports torchvision even if the recipe doesnt need it. We should protect this import, in case the user hasnt installed it and doesnt need it.

joecummings:
What's your implementation plan @felipemello1 ?

felipemello1:
This is the error:
```
  File "/data/users/felipemello/torchtune/torchtune/data/__init__.py", line 7, in <module>
    from torchtune.data._chat_formats import (
  File "/data/users/felipemello/torchtune/torchtune/data/_chat_formats.py", line 10, in <module>
    from torchtune.data._messages import Message, Role
  File "/data/users/felipemello/torchtune/torchtune/data/_messages.py", line 9, in <module>
    from torchtune.modules.transforms import Transform
  File "/data/users/felipemello/torchtune/torchtune/modules/transforms/__init__.py", line 17, in <module>
    from torchtune.modules.transforms.vision_utils.resize_with_pad import (  # noqa
  File "/data/users/felipemello/torchtune/torchtune/modules/transforms/vision_utils/resize_with_pad.py", line 14, in <module>
    import torchvision
ModuleNotFoundError: No module named 'torchvision'
```

It happens because in [transforms.__init__](https://github.com/pytorch/torchtune/blob/main/torchtune/modules/transforms/__init__.py) we expose vision transforms.

A simple solution is to not import chat transforms from the same place as vision transforms. This can be done by renaming /vision_utils to /vision_transforms

And putting the transforms.py (chat related) under /text_transforms

cc: @RdoubleA @pbontrager

joecummings:
I agree this needs to be better thought out how we structure our imports here. However another stopgap we could do is the following:

```
try:
	import torchvision
except ImportError:
	log.warning("Attempted to import torchvision. Feel free to ignore if you are not trying to do multimodal training. Otherwise, please install ``torchvision`` package via conda or pip")
```

This would immediately allow users to use our package without the pre-requisite of another heavy PyTorch library.

I acknowledge it's hackiness, but in my mind this is a bad starting experience and it's not immediately clear ATM why someone would need torchvision when we don't (yet) support MM training.

Thoughts @felipemello1 @pbontrager ?

felipemello1:
wouldnt you have to add it in all files that import torchvision? If you want a quick solution, I think that moving these functions from transforms/init --> transforms/vision_transforms/init will be best.

joecummings:
[...]

PR #1367, opened 2024-08-19
[STOPGAP] Remove public APIs for vision utils

joecummings:
#### Context
- [ ] add a new feature
- [x] fix a bug
- [ ] other (please add here)

Related to #1352 but does not actually close it.

#### Changelog
* Remove all functions within `vision_utils` from `modules/transforms/__init__.py`
* Update all references to these functions 
* Remove functions from documentation

#### Test plan

In addition, I did the following test to ensure that I did not need torchvision installed in order to run basic stuff.

```
1. conda create fresh_env
2. conda activate fresh_env
3. pip install torch
4. pip install -e .
5. tune ls
# Works!
```

#### UX

### DIFF
--- a/docs/source/api_ref_modules.rst
+++ b/docs/source/api_ref_modules.rst
@@ -103,11 +103,6 @@ Functions used for preprocessing images.
    :nosignatures:
 
     transforms.Transform
-    transforms.get_canvas_best_fit
-    transforms.get_inscribed_size
-    transforms.resize_with_pad
-    transforms.tile_crop
-    transforms.find_supported_resolutions
     transforms.VisionCrossAttentionMask
 
 Reinforcement Learning From Human Feedback (RLHF)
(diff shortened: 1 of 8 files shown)


felipemello1 (approved):

pbontrager (changes requested):
On board with the import path changes. Though I think these should remain in the api docs.

felipemello1:
I think that just migrating them to vision_utils/__init__.py would be better, and then "from transforms/vision_utils import X". But not the end of the world to keep as is

joecummings:
> On board with the import path changes. Though I think these should remain in the api docs.

It's still directly importable for models within the library, but I think this is in line with our original objectives around APIs: Keep them private by default until a need arises. This was code submitted by a third party and it's helpful for building CLIP, but I see no reason to publicize these for direct use by torchtune users.

pbontrager (approved):

PR #1370, opened 2024-08-19
Undo commenting out NF4/bitsandbytes tests

SalmanMohammadi:
#### Context
- [ ] add a new feature
- [x] fix a bug
- [ ] other (please add here)

My bad. I must've forgot to un-comment this when I was testing locally on MPS, and it landed in #1005. It was just such an exciting time.

#### Changelog

Un-comment the test file. I've also added a check to skip if bnb isn't installed. 

#### Test plan

#### UX


SalmanMohammadi:
> I've also added a check to skip if bnb isn't installed.

This is mostly because it was annoying me but I realise it may not be happening for many people, so don't mind removing it.

joecummings (approved):

PR #1371, opened 2024-08-19
Fix dtype error during quantized generation

### LINKED ISSUE #1349: RuntimeError: Index put requires the source and destination dtypes match, got BFloat16 for the destination and Float for the source.
Hi,

I am trying to apply the generate recipe on a quantized llama 3.1 8B model but run into the following error:
```
...
  File "/home/mreso/torchtune/torchtune/modules/attention.py", line 211, in forward
    k, v = self.kv_cache.update(input_pos, k, v)
  File "/home/mreso/torchtune/torchtune/modules/kv_cache.py", line 69, in update
    k_out[:, :, input_pos] = k_val
RuntimeError: Index put requires the source and destination dtypes match, got BFloat16 for the destination and Float for the source.
```
I looked a bit through the code of torchtune and torchao and found that the linear layers like k_proj are wrapped with Int8DynActInt4WeightLinear which has a precision argument that seems to get used [here](https://github.com/pytorch/ao/blob/582b6d490d03fffd19388c4f88ef7682a82dba34/torchao/quantization/GPTQ.py#L842) to determine the precision of the computation. Its default value is float32 and it seems to be unaffected buy the model.to(dtype) call [here](https://github.com/pytorch/torchtune/blob/34162c9b2b24e6e555f6a06074dc0c34431a5aa8/recipes/generate.py#L73).
```
> /home/mreso/torchtune/generate.py(77)_setup_model()
-> model.load_state_dict(model_state_dict)
(Pdb) print(model.layers[0].attn.q_proj.precision)
torch.float32
```
Thus the value thats gets used to updated the kv cache is float32 instead of bloat16.
Trying to set the precision in the configuration only results in this error:
```
  File "generate.py", line 74, in _setup_model
    model = self._quantizer.quantize(model)
  File "/home/mreso/.conda/envs/tune/lib/python3.10/site-packages/torchao/quantization/GPTQ.py", line 1014, in quantize
    state_dict = self._create_quantized_state_dict(model)
  File "/home/mreso/.conda/envs/tune/lib/python3.10/site-packages/torch/utils/_contextlib.py", line 116, in decorate_context
    return func(*args, **kwargs)
  File "/home/mreso/.conda/envs/tune/lib/python3.10/site-packages/torchao/quantization/GPTQ.py", line 989, in _create_quantized_state_dict
    weight.to(self.precision),
RuntimeError: Invalid device string: 'bfloat16'
```
Not sure if there is a way to set the precision correctly from the config.
Happy to create a PR to create a dtype for the precision from the string or something.

**Steps to reproduce**

```
tune run quantize --config quantize.yaml
```
quantize.yaml:
```
# Config for QuantizationRecipe in quantize.py
#
# To launch, run the following command from root torchtune directory:
#    tune run quantize --config quantization

#
# Model arguments
model:

checkpointer:
  _component_: torchtune.utils.FullModelMetaCheckpointer
  checkpoint_dir: Meta-Llama-3-8B-Instruct/
  checkpoint_files: [
    meta_model_0.pt
  ]
  recipe_checkpoint: null
  output_dir: ./quantized
  model_type: LLAMA3

device: cuda
dtype: bf16
seed: 1234
max_seq_len: 2048

quantizer:
  _component_: torchtune.utils.quantization.Int8DynActInt4WeightQuantizer
  groupsize: 256
```

```
tune run generate --config generate.yaml
```

generate.yaml:
```
# Config for running the InferenceRecipe in generate.py to generate output from an LLM
#
# To launch, run the following command from root torchtune directory:
#    tune run generate --config generation

# Model arguments
model:
  max_seq_len: 2048

checkpointer:
  _component_: torchtune.utils.FullModelTorchTuneCheckpointer
  checkpoint_dir: quantized/
  checkpoint_files: [
    meta_model_0-8da4w.pt
  ]
  output_dir: model-output
  model_type: LLAMA3

device: cuda
dtype: bf16

seed: 1234

# Tokenizer arguments
tokenizer:
  path: /home/mreso/.cache/huggingface/hub/models--meta-llama--Meta-Llama-3.1-8B-Instruct/snapshots/8c22764a7e3675c50d4c7c9a4edb474456022b16/original/tokenizer.model

# Generation arguments; defaults taken from gpt-fast
prompt: 'Amanda: I baked  cookies. Do you want some?\nJerry: Sure \nAmanda: I will bring you tomorrow :-)'
# instruct_template: torchtune.data.SummarizeTemplate
#chat_format: null
max_new_tokens: 300
temperature: 0.6 # 0.8 and 0.6 are popular values to try
top_k: 300
enable_kv_cache: True

quantizer:
  _component_: torchtune.utils.quantization.Int8DynActInt4WeightQuantizer
  groupsize: 256
#  precision: bfloat16 # uncommenting this results into second error
```

**Full Logs**
Full error stack trace:
```
INFO:torchtune.utils.logging:Starting compilation to improve generation performance ...
  File "/home/mreso/.conda/envs/tune/bin/tune", line 8, in <module>
    sys.exit(main())
  File "/home/mreso/torchtune/torchtune/_cli/tune.py", line 49, in main
    parser.run(args)
  File "/home/mreso/torchtune/torchtune/_cli/tune.py", line 43, in run
    args.func(args)
  File "/home/mreso/torchtune/torchtune/_cli/run.py", line 179, in _run_cmd
    self._run_single_device(args)
  File "/home/mreso/torchtune/torchtune/_cli/run.py", line 93, in _run_single_device
    runpy.run_path(str(args.recipe), run_name="__main__")
  File "/home/mreso/.conda/envs/tune/lib/python3.10/runpy.py", line 289, in run_path
    return _run_module_code(code, init_globals, run_name,
  File "/home/mreso/.conda/envs/tune/lib/python3.10/runpy.py", line 96, in _run_module_code
    _run_code(code, mod_globals, init_globals,
  File "/home/mreso/.conda/envs/tune/lib/python3.10/runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "/home/mreso/torchtune/recipes/generate.py", line 204, in <module>
    sys.exit(main())
  File "/home/mreso/torchtune/torchtune/config/_parse.py", line 50, in wrapper
    sys.exit(recipe_main(conf))
  File "/home/mreso/torchtune/recipes/generate.py", line 200, in main
    recipe.generate(cfg=cfg)
  File "/home/mreso/.conda/envs/tune/lib/python3.10/site-packages/torch/utils/_contextlib.py", line 116, in decorate_context
    return func(*args, **kwargs)
  File "/home/mreso/torchtune/recipes/generate.py", line 151, in generate
    _ = utils.generate(
  File "/home/mreso/.conda/envs/tune/lib/python3.10/site-packages/torch/utils/_contextlib.py", line 116, in decorate_context
    return func(*args, **kwargs)
  File "/home/mreso/torchtune/torchtune/utils/_generation.py", line 122, in generate
    tokens = generate_next_token(
  File "/home/mreso/torchtune/torchtune/utils/_generation.py", line 47, in generate_next_token
    logits = model(x, input_pos=input_pos)[:, -1]
  File "/home/mreso/.conda/envs/tune/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1736, in _wrapped_call_impl
    return self._call_impl(*args, **kwargs)
  File "/home/mreso/.conda/envs/tune/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1747, in _call_impl
    return forward_call(*args, **kwargs)
  File "/home/mreso/torchtune/torchtune/modules/transformer.py", line 243, in forward
    h = layer(h, mask=mask, input_pos=input_pos)
  File "/home/mreso/.conda/envs/tune/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1736, in _wrapped_call_impl
    return self._call_impl(*args, **kwargs)
  File "/home/mreso/.conda/envs/tune/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1747, in _call_impl
    return forward_call(*args, **kwargs)
  File "/home/mreso/torchtune/torchtune/modules/transformer.py", line 72, in forward
    attn_out = self.attn(self.sa_norm(x), mask=mask, input_pos=input_pos)
  File "/home/mreso/.conda/envs/tune/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1736, in _wrapped_call_impl
    return self._call_impl(*args, **kwargs)
  File "/home/mreso/.conda/envs/tune/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1747, in _call_impl
    return forward_call(*args, **kwargs)
  File "/home/mreso/torchtune/torchtune/modules/attention.py", line 211, in forward
    k, v = self.kv_cache.update(input_pos, k, v)
  File "/home/mreso/torchtune/torchtune/modules/kv_cache.py", line 69, in update
    k_out[:, :, input_pos] = k_val
RuntimeError: Index put requires the source and destination dtypes match, got BFloat16 for the destination and Float for the source.
```

mreso:
This PR fixes #1349 RuntimeError: Index put requires the source and destination dtypes match, got BFloat16 for the destination and Float for the source

Repro:
See #1349 for repro
Cmd:
```
$tune run generate --config generate_quant.yaml
```
Output:
```
INFO:torchtune.utils.logging:Running InferenceRecipe with resolved config:

checkpointer:
  _component_: torchtune.utils.FullModelTorchTuneCheckpointer
  checkpoint_dir: quantized/
  checkpoint_files:
  - meta_model_0-8da4w.pt
  model_type: LLAMA3
  output_dir: model-output
device: cuda
dtype: bf16
enable_kv_cache: true
max_new_tokens: 300
model:
  max_seq_len: 2048
prompt: 'Amanda: I baked  cookies. Do you want some?\nJerry: Sure \nAmanda: I will
  bring you tomorrow :-)'
quantizer:
  _component_: torchtune.utils.quantization.Int8DynActInt4WeightQuantizer
  groupsize: 256
seed: 1234
temperature: 0.6
tokenizer:
  path: /home/mreso/.cache/huggingface/hub/models--meta-llama--Meta-Llama-3.1-8B-Instruct/snapshots/8c22764a7e3675c50d4c7c9a4edb474456022b16/original/tokenizer.model
top_k: 300

DEBUG:torchtune.utils.logging:Setting manual seed to local seed 1234. Local seed is seed + rank = 1234 + 0
linear: layers.0.attn.q_proj, in=4096, out=4096
linear: layers.0.attn.k_proj, in=4096, out=1024
linear: layers.0.attn.v_proj, in=4096, out=1024
linear: layers.0.attn.output_proj, in=4096, out=4096
....
linear: layers.31.mlp.w1, in=4096, out=14336
linear: layers.31.mlp.w2, in=14336, out=4096
linear: layers.31.mlp.w3, in=4096, out=14336
linear: output, in=4096, out=128256
INFO:torchtune.utils.logging:Model is initialized with precision torch.bfloat16.
INFO:torchtune.utils.logging:Starting compilation to improve generation performance ...
INFO:torchtune.utils.logging:Warmup run for quantized model takes: 33.09 sec
INFO:torchtune.utils.logging:Amanda: I baked  cookies. Do you want some?\nJerry: Sure \nAmanda: I will bring you tomorrow :-) \nJerry: What are you talking about? I want them now\nAmanda: I said tomorrow, you can't just eat them all by yourself, that's not fair. (She laughs.) \nJerry: I'm starving\nAmanda: Go ask someone else, I'm not going to be your cookie m
other \nJerry: That's not fair\nAmanda: I know, but it's not my responsibility to feed you all day \nJerry: Fine, I'll just ask someone else. \nAmanda: And don't eat all the cookies, they're for sharing, remember?\nAmanda: I'm going to go, bye \nJerry: Wait, come back, I'll pay you, I'll buy you more cookies\nAmanda: I don't need your money, I don't need your
cookies, I'll just eat them myself, bye \nJerry: You're really mean\nAmanda: I'm not mean, I'm just firm, that's all\nAmanda: I'm going to go, bye \nJerry: Wait, come back, I'll get you an ice cream, you like ice cream, don't you?\nAmanda: No, I don't want your ice cream, I don't want your cookies, I don't want your money, I'll just go, bye \nJerry: You're rea
lly mean, I'm going to tell everyone, you're mean\nAmanda
INFO:torchtune.utils.logging:Time for inference: 47.79 sec total, 6.28 tokens/sec
INFO:torchtune.utils.logging:Bandwidth achieved: 61.40 GB/s
INFO:torchtune.utils.logging:Memory used: 22.97 GB
```

#### Context
- [ ] add a new feature
- [X] fix a bug
- [ ] other (please add here)

#### Changelog

- Cast k,v,q to kv_cache dtype
- Clone token before feeding it to generate_next token (avoids torch.compile/CUDAGraph overwritten ssieu
- Remove size parameter in kvcache as its incompatible with torch.compile

#### Test plan

- [X] include relevant commands and any other artifacts in this summary (pastes of loss curves, eval results, etc.)

#### UX

- [X] I did not change any public API;

### DIFF
--- a/torchtune/modules/attention.py
+++ b/torchtune/modules/attention.py
@@ -211,7 +211,7 @@ class MultiHeadAttention(nn.Module):
         s_y = y.shape[1]
 
         if self.kv_cache and input_pos is None:
-            cache_size = self.kv_cache.size
+            cache_size = self.kv_cache.k_cache.dim(2)
             input_pos = torch.arange(cache_size, cache_size + s_y, device=x.device)
 
         # q has shape [b, s_x, num_heads * head_dim]
@@ -264,6 +264,7 @@ class MultiHeadAttention(nn.Module):
         # Update key-value cache
         if self.kv_cache is not None:
             k, v = self.kv_cache.update(input_pos, k, v)
+            q = q.to(self.kv_cache.dtype)
 
         # shape: [b, 1, s, s]
         if mask is not None:
--- a/torchtune/modules/kv_cache.py
+++ b/torchtune/modules/kv_cache.py
@@ -40,8 +40,8 @@ class KVCache(nn.Module):
         self.register_buffer(
             "v_cache", torch.zeros(cache_shape, dtype=dtype), persistent=False
         )
-        self.size = 0
         self.batch_size = batch_size
+        self.dtype = dtype
 
     def reset(self) -> None:
         """Reset the cache to zero."""
@@ -64,11 +64,10 @@ class KVCache(nn.Module):
             Tuple[Tensor, Tensor]: Updated KV cache with key first
         """
         assert input_pos.shape[0] == k_val.shape[2]
-        self.size = input_pos.max().item() + 1
 
         k_out = self.k_cache
         v_out = self.v_cache
-        k_out[:, :, input_pos] = k_val
-        v_out[:, :, input_pos] = v_val
+        k_out[:, :, input_pos] = k_val.to(self.dtype)
+        v_out[:, :, input_pos] = v_val.to(self.dtype)
 
         return k_out, v_out
--- a/torchtune/modules/kv_cache.py
+++ b/torchtune/modules/kv_cache.py
@@ -42,6 +42,7 @@ class KVCache(nn.Module):
         )
         self.batch_size = batch_size
         self.dtype = dtype
+        self.size = torch.tensor(0, dtype=dtype)
 
     def reset(self) -> None:
         """Reset the cache to zero."""
@@ -64,6 +65,8 @@ class KVCache(nn.Module):
             Tuple[Tensor, Tensor]: Updated KV cache with key first
         """
         assert input_pos.shape[0] == k_val.shape[2]
+
+        self.size = input_pos.max()
 
         k_out = self.k_cache
         v_out = self.v_cache
--- a/torchtune/utils/_generation.py
+++ b/torchtune/utils/_generation.py
@@ -159,7 +159,7 @@ def generate(
         tokens = custom_generate_next_token(
             model,
             input_pos=curr_input_pos,
-            x=tokens,
+            x=tokens.clone(),
             temperature=temperature,
             top_k=top_k,
         )


torchtune/utils/_generation.py
```diff
@@ -159,7 +159,7 @@ def generate(
         tokens = custom_generate_next_token(
             model,
             input_pos=curr_input_pos,
-            x=tokens,
```
felipemello1, line 162: Can you help me understand why we need to do .clone? I worry about performance issues

torchtune/modules/kv_cache.py
```diff
@@ -64,11 +64,10 @@ def update(
             Tuple[Tensor, Tensor]: Updated KV cache with key first
         """
         assert input_pos.shape[0] == k_val.shape[2]
-        self.size = input_pos.max().item() + 1
 
         k_out = self.k_cache
         v_out = self.v_cache
-        k_out[:, :, input_pos] = k_val
-        v_out[:, :, input_pos] = v_val
```
felipemello1, line 72: question: We need to do it because the base model has one dtype, e.g. float32, but the cache is in fp16, for example, is that right?

after we save the k_val and v_val in fp16, we have to return the entire cache to be used by the attention. Dont we have to upcast it back to float32, since the model is in float32? I dont see it happening anywhere.

torchtune/modules/attention.py
```diff
@@ -211,7 +211,7 @@ def forward(
         s_y = y.shape[1]
 
         if self.kv_cache and input_pos is None:
-            cache_size = self.kv_cache.size
+            cache_size = self.kv_cache.k_cache.dim(2)
```
felipemello1, line 214: This is great! Much better than doing .item(). Can you keep self.size inside of the cache and do this change there? I think its a bit safer and cleaner.

FYI, a bit related to this PR: https://github.com/pytorch/torchtune/pull/1364

felipemello1 (commented):
Thanks for the PR! Very clean changes. I left a few comments

torchtune/utils/_generation.py
mreso: Sure, without it we're running into this error:
```
RuntimeError: Error: accessing tensor output of CUDAGraphs that has been overwritten by a subsequent run. Stack trace: File "/home/mreso/torchtune/torchtune/utils/_generation.py", line 48, in generate_next_token
    return sample(logits, temperature, top_k)
  File "/home/mreso/torchtune/torchtune/utils/_generation.py", line 34, in sample
    return multinomial_sample_one(probs)
  File "/home/mreso/torchtune/torchtune/utils/_generation.py", line 16, in multinomial_sample_one
    return torch.argmax(probs / q, dim=-1, keepdim=True).to(dtype=torch.int). To prevent overwriting, clone the tensor outside of torch.compile() or call torch.compiler.cudagraph_mark_step_begin() before each model invocation.
```
Could not get this to work with torch.compiler.cudagraph_mark_step_begin() in the first try so I opted for cloning. Can retry to make cudagraph_mark_step_begin() work,  if you think performance will take a big hit.

torchtune/modules/kv_cache.py
mreso: The dtype in the yaml config is actually bfloat16 so the model is cast to that dtype [here](https://github.com/pytorch/torchtune/blob/2f8ed7a34457eaddce59a58e8c37dd51d6ba65f8/recipes/generate.py#L73). Its actually the quantization component wrapping the linear layers that returns float32, so we need to explicitly downcast to the same dtype as the model. See [here ](https://github.com/pytorch/torchtune/issues/1349#issuecomment-2297562155) why this returns float32. I think we would need to upstream changes to torchao to change this.

Alternatively, we could do the downcast right after computing q,k,v but kv cache already had the dtype given during initialization. Otherwise we either need derive the dtype for q,k,v from the model parameters or change the interface for the attention layer creation to accept a dtype.

torchtune/modules/attention.py
mreso: Yes, good idea. I now just removed the item(), so size will be a tensor which can still be used in the same way as the scalar before.

torchtune/modules/kv_cache.py
felipemello1: >Its actually the quantization component wrapping the linear layers that returns float32

@mreso, thanks for explaining it! Lets try to fix the root cause then. I dont think it is the intended behavior of the quantizer to return float32 if the input is float16.[ I took a look at the wrapper ](https://github.com/pytorch/ao/blob/227d4bfece6bb00f0365642ad1d7afd509079ba4/torchao/quantization/GPTQ.py#L893)and it says that:

"precision: precision of input and output. e.g. torch.float32 means input
activation is float32 and output is float32.
scales_precision: precision of per group scale."

If you have bandwidth, would you mind doing a quick test with the quantizer dtype inputs? If the inputs is indeed fp16, and precision=fp16 (and scales_precision), and it still output fp32, then I will ping the torchao folks.

mreso: Here is a quick test:
```
import torch
import torch.nn as nn
from torchtune.utils.quantization import Int8DynActInt4WeightQuantizer


class Foo(nn.Module):
    def __init__(self):
        super().__init__()
        self.linear = nn.Linear(256, 256, bias=False)
        self.linear2 = nn.Linear(256, 256, bias=False)

    def forward(self, x):
        x = self.linear(x)
        x = self.linear2(x)
        return x



quantizer_f32 = Int8DynActInt4WeightQuantizer()

foo = Foo()
foo = quantizer_f32.quantize(foo)
foo = foo.to(dtype=torch.bfloat16)

# foo = torch.compile(foo)

x_bf16 = torch.rand(1,256, dtype=torch.bfloat16)
print(f"{x_bf16.dtype=}")
y=foo(x_bf16)
print(f"{y.dtype=}")

x_f32 = torch.rand(1,256, dtype=torch.float32)
print(f"{x_f32.dtype=}")
y=foo(x_f32)
print(f"{y.dtype=}")

quantizer_bf16 = Int8DynActInt4WeightQuantizer(precision=torch.bfloat16)

bar = Foo()
bar = quantizer_bf16.quantize(bar)
bar = bar.to(dtype=torch.bfloat16)

# bar = torch.compile(bar)

x = torch.rand(1,256, dtype=torch.bfloat16)
print(f"{x.dtype=}")
y=bar(x)
print(f"{y.dtype=}")
```
As the docs state the output's precision depends on quantizer.precision. So for bfloat16 in it should be bfloat16 out:
```
linear: linear, in=256, out=256
linear: linear2, in=256, out=256
x_bf16.dtype=torch.bfloat16
y.dtype=torch.float32
x_f32.dtype=torch.float32
y.dtype=torch.float32
linear: linear, in=256, out=256
linear: linear2, in=256, out=256
x.dtype=torch.bfloat16
y.dtype=torch.bfloat16
```
This brings us back to two problems described in the original issue:
1. We currently can not set the precision with the current implementation of ao + torchtune as precision has to be a dtype but torchtune only passes the string read from the config.
2. If we somehow set precision to the correct dtype as e.g. [described here](https://github.com/pytorch/torchtune/issues/1349#issuecomment-2294317411) AND enable torch.compile (which the generate recipe does) we end up with the inductor error.
```
  File "/home/mreso/torchtune/test_quantizer.py", line 47, in <module>
    y=bar(x)
 ...
  File "/tmp/torchinductor_mreso/ji/cjib4exg2swki3pwhhi7cgq5qh7l6yriypsqejd3wdcajrx55zrc.py", line 69, in call
    extern_kernels.mm(buf6, reinterpret_tensor(buf8, (256, 256), (1, 256), 0), out=buf9)
RuntimeError: Expected out tensor to have dtype float, but got c10::BFloat16 instead
```
So if we don't want to solve the problem by casting we need to fix the inductor issue on the ao side.

felipemello1: Nice!

1. one is easy to fix. We can just use [this mapping](https://github.com/pytorch/torchtune/blob/34162c9b2b24e6e555f6a06074dc0c34431a5aa8/torchtune/utils/precision.py#L17) to go from string to torch.bfloat16


2. This seems to need attention

@msaroufim, can you take a look at this issue? 

**TLDR:**
using compile + Int8DynActInt4WeightQuantizer with precision=bf16 breaks.
```
quantizer_bf16 = Int8DynActInt4WeightQuantizer(precision=torch.bfloat16)

bar = Foo()
bar = quantizer_bf16.quantize(bar)
bar = bar.to(dtype=torch.bfloat16)
bar = torch.compile(bar)

x = torch.rand(1,256, dtype=torch.bfloat16)
y=bar(x)
```

raises the error:
```
raceback (most recent call last):
  File "/home/mreso/torchtune/test_quantizer.py", line 47, in <module>
    y=bar(x)
 ...
  File "/tmp/torchinductor_mreso/ji/cjib4exg2swki3pwhhi7cgq5qh7l6yriypsqejd3wdcajrx55zrc.py", line 69, in call
    extern_kernels.mm(buf6, reinterpret_tensor(buf8, (256, 256), (1, 256), 0), out=buf9)
RuntimeError: Expected out tensor to have dtype float, but got c10::BFloat16 instead
```

If we dont add ```precision=torch.bfloat16```, then for an fp16 input, we get fp32 output.

msaroufim: @jerryzh168 any reason why we're not using `quantize_()` here instead?

felipemello1: For reference, here is where the component is defined in the config: https://github.com/pytorch/torchtune/blob/main/recipes/configs/quantization.yaml#L27

felipemello1: And here is where its instantiated: https://github.com/pytorch/torchtune/blob/9e65fa9ae67ce09e953cb68b5093d42cb6b31310/recipes/quantize.py#L51

jerryzh168: @msaroufim there was some issues with loading the tensor subclass quantized weights in `generate.py` before (with map_location="cpu": https://github.com/pytorch/torchtune/blob/f9f75bb563ecae371492a9d49da4a9f514c081b3/torchtune/utils/_checkpointing/_checkpointer_utils.py#L110-L111), things might have changed though. Also we need to clarify the plan for QAT flow I think. cc @andrewor14 is the recent tensor subclass refactor tested in torchtune?

mreso: @msaroufim @jerryzh168 @felipemello1 is there any short term solution to unblock us here? If its the original issue or the PR?

PR #1382, opened 2024-08-20
When profile_memory, also export and save the snapshot.pickle for lora_finetune_single_device.py

janeyx99:
#### Context
- [x] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

#### Changelog
* enable the snapshot.pickle to be exported when profile_memory is True.
* the above required adding attributes to self for recipes/lora_finetune_single_device.py
* needing to export the snapshot.pickle _does_ mean that it takes a little longer to export.

Previously, you only had the html from the profiler, which is not interactive:
image

Now, you get that in addition to:
image

#### Test plan

#### UX

### DIFF
--- a/recipes/lora_finetune_single_device.py
+++ b/recipes/lora_finetune_single_device.py
@@ -328,10 +328,11 @@ class LoRAFinetuneRecipeSingleDevice(FTRecipeInterface):
 
         log.info(f" Profiler config after instantiation: {profiler_cfg}")
 
-        self.profiler_wait_steps = profiler_cfg["wait_steps"]
-        self.profiler_warmup_steps = profiler_cfg["warmup_steps"]
-        self.profiler_active_steps = profiler_cfg["active_steps"]
-        self.profiler_profile_memory = profiler_cfg["profile_memory"]
+        self.profiler_profile_memory = profiler_cfg.get("profile_memory", False)
+        if profiler_cfg["enabled"]:
+            self.profiler_wait_steps = profiler_cfg["wait_steps"]
+            self.profiler_warmup_steps = profiler_cfg["warmup_steps"]
+            self.profiler_active_steps = profiler_cfg["active_steps"]
 
         return profiler
 
@@ -584,7 +585,12 @@ class LoRAFinetuneRecipeSingleDevice(FTRecipeInterface):
                     ):
                         break
 
-                    if curr_epoch == 0 and self.profiler_profile_memory and idx == self.profiler_wait_steps + self.profiler_warmup_steps:
+                    # Start tracking CUDA memory for active steps for just the first epoch
+                    if (
+                        curr_epoch == 0
+                        and self.profiler_profile_memory
+                        and idx == self.profiler_wait_steps + self.profiler_warmup_steps
+                    ):
                         torch.cuda.memory._record_memory_history()
 
                     batch = {k: v.to(self._device) for k, v in batch.items()}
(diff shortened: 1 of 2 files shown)


ebsmothers (approved):
Thanks for adding this! Can you fix the linter complaints? After that this should be good to go

recipes/lora_finetune_single_device.py
```diff
@@ -579,6 +584,9 @@ def train(self) -> None:
                     ):
                         break
```
felipemello1, line 586:
```suggestion
                    # Memory profiling
                    if curr_epoch == 0 and self.profiler_profile_memory and idx == (self.profiler_wait_steps + self.profiler_warmup_steps):
                        torch.cuda.memory._record_memory_history()
```

recipes/lora_finetune_single_device.py
```diff
@@ -579,6 +584,9 @@ def train(self) -> None:
...
 
+                    if curr_epoch == 0 and self.profiler_profile_memory and idx == self.profiler_wait_steps + self.profiler_warmup_steps:
+                        torch.cuda.memory._record_memory_history()
```
felipemello1, line 588: The issue is if in the future we support other backend, like intel. It makes me think that this should be a function, instead of being hardcoded as "torch.cuda"

cc: @ebsmothers

recipes/lora_finetune_single_device.py
```diff
@@ -626,6 +634,9 @@ def train(self) -> None:
                         num_tokens = 0
                         t0 = time.perf_counter()
 
+                    if curr_epoch == 0 and self.profiler_profile_memory and idx == self.profiler_wait_steps + self.profiler_warmup_steps + self.profiler_active_steps:
```
felipemello1, line 637:
```suggestion
                    # Stop memory profiling
                    if curr_epoch == 0 and self.profiler_profile_memory and idx == self.profiler_wait_steps + self.profiler_warmup_steps + self.profiler_active_steps:
```

felipemello1 (commented):
@ebsmothers, wdyt? Were were missing the memory snapshot. My only concern is hardcoding torch.cuda if we plan to support xpu

felipemello1:
should we just delete the HTML to save time, since it takes a while to save? Or is it useful?

recipes/lora_finetune_single_device.py
ebsmothers: > The issue is if in the future we support other backend, like intel. It makes me think that this should be a function, instead of being hardcoded as "torch.cuda"
> 
> cc: @ebsmothers

I think it's OK, idk that memory snapshot would be supported on XPU anyways. We can do a check on device + profile_memory combo just to be safe though

recipes/lora_finetune_single_device.py
```diff
@@ -328,6 +328,11 @@ def _setup_profiler(
 
         log.info(f" Profiler config after instantiation: {profiler_cfg}")
 
+        self.profiler_wait_steps = profiler_cfg["wait_steps"]
```
ebsmothers, line 331: This seems to be throwing an error in our recipe tests, can you just quickly inspect `profiler_cfg` after running e.g.

```
pytest tests/recipes/test_lora_finetune_single_device.py -m integration_test -k 'test_loss'
```

I thought these fields would be defined based on [this](https://github.com/pytorch/torchtune/blob/2f8ed7a34457eaddce59a58e8c37dd51d6ba65f8/torchtune/utils/_profiler.py#L370), but seems like something weird is happening here.

janeyx99: Ah, it's because profiling isn't enabled for tests, so it returns here https://github.com/pytorch/torchtune/blob/2f8ed7a34457eaddce59a58e8c37dd51d6ba65f8/torchtune/utils/_profiler.py#L262

janeyx99:
> should we just delete the HTML to save time, since it takes a while to save? Or is it useful?

Writing the html is very light so it doesn't make a tangible difference, I'm not sure if people are using it today so it'd be bad to just remove it if there were actual users.

PR #1386, opened 2024-08-21
expose rope scale_factor to the builder

felipemello1:
#### Context
- [ ] add a new feature
- [ ] fix a bug
- [x] other (please add here)

scaling factors of RoPE are hardcoded. This PR exposes them. Even though all 4 factors are exposed in the RoPE class, in the builder, I only expose the scaling_factor, since this is the main one that would matter when instantiating a model.

#### Change log
- added the values to unit test (previously hardcoded in the function)
- exposed only scaling_factor to the model builder
- exposed all scaling factors in rope init

#### Test plan
image


torchtune/models/llama3_1/_position_embeddings.py
```diff
@@ -118,10 +143,16 @@ def forward(self, x: Tensor, *, input_pos: Optional[Tensor] = None) -> Tensor:
...
+
         # TODO: remove once our distributed recipes are on FSDP2
         if not self.is_cache_built:
             with torch.device(x.device):
-                self.rope_init()
+                self.rope_init(
+                    scale_factor=self.scale_factor,
+                    low_freq_factor=self.low_freq_factor,
+                    high_freq_factor=self.high_freq_factor,
+                    old_context_len=self.old_context_len,
```
felipemello1, line 154: You may ask: "Why pass the variables to init_rope, if you can just use self.<var_name>" inside of rope_init. I think that this give the user a bit more control if they want to test different factors without having to instantiate a new class. I am happy to change it if you prefer.

torchtune/models/llama3_1/_position_embeddings.py
```diff
@@ -21,38 +21,63 @@ class Llama3ScaledRoPE(nn.Module):
...
         dim (int): Embedding dimension. This is usually set to the dim of each
             head in the attention module computed as ````embed_dim`` // ``num_heads````
         max_seq_len (int): Maximum expected sequence length for the
             model, if exceeded the cached freqs will be recomputed
         base (int): The base for the geometric progression used to compute
             the rotation angles
+        scale_factor (int): scaling factor for scaling. Default: 8
+        low_freq_factor (int): low frequency factor for scaling. Default: 1
+        high_freq_factor (int): high frequency factor for scaling. Default: 4
+        old_context_len (int): old context length for scaling. Default: 8192
```
RdoubleA, line 37:
```suggestion
        old_context_len (int): old context length for scaling theta. Default: 8192
```

RdoubleA (approved):
Looks good to me, just minor comments

torchtune/models/llama3_1/_position_embeddings.py
kartikayk: Actually what I want to ask is when will we remove this function and rally behind @joecummings RFC on scaled RoPE? :) I think this was a bit of a last min bandaid for the July launch. Do we have plans of updating? Or can we track an issue around this?

cc: @ebsmothers

ebsmothers: > You may ask: "Why pass the variables to init_rope, if you can just use self.<var_name>" inside of rope_init. I think that this give the user a bit more control if they want to test different factors without having to instantiate a new class. I am happy to change it if you prefer.

I think that's fine, but can also make them optional and default to `self.` when not passed

> Actually what I want to ask is when will we remove this function and rally behind @joecummings RFC on scaled RoPE? :) I think this was a bit of a last min bandaid for the July launch. Do we have plans of updating? Or can we track an issue around this?

Did we get to a clear conclusion on that RFC? If so I agree we should go ahead with it, I believe even in the RFC @joecummings mentioned

> It'll be pretty easy and valuable to define a task for contributors to add NTK-by-parts, PI, etc. to the repo so users can play around with different scaling methods.

So we can definitely start going down this path, but probably need to at least do the initial refactor to support parametrizing `scaling_module` ourselves. Though I don't think we have to block this PR on that; this seems reasonable as a standalone change to add until we do that refactor.

PR #1390, opened 2024-08-22
chunked cross entropy

felipemello1:
#### Context
- [X] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

Mechanism: Chunked cross entropy reduces memory by processing and upcasting one chunk at a time. This allows the memory to be released after the computation. If we dont chunk, then compiling model+loss together or independently has no difference, as long as we compile the upcasting with the loss, e.g.:

```
@torch.compile()
def _cross_entropy_loss(logits, float):
	return cross_entropy_loss(logits.float(), labels)
```

It is important to notice that chunking must happen BEFORE the linear output, or the memory is not released:
chunking after the linear output layer (bad):
image

chunking before the linear output layer (This pr):
image

When compiling model and loss independently, if the upcast doesnt happen inside of the compiled Cross Entropy, chunking is less effective
image

The other impact of this PR is removing logits manipulation:
```
```
image

torch.compile is much better in 2.5 vs 2.4
image

At least for this config, no huge gains in memory for num_chunks > 4. However, we dont see any performance degradation using num_chunks=16, so this was the chosen value.
image

edit: when running smaller models, there is a performance degradation (~10% TPS). I changed the number to num_chunk=8, and there is almost no TPS reduction, and memory savings is still very high.

### IMPACT
For llama 8B, seq_len=8192, if activation checkpointing is FALSE, then:

toks/sec:
compiling model+loss together == chunked + compile model + loss separetely

however, if checkpointing is TRUE, then:
compiling model+loss together >> chunked + compile model + loss separetely

In every scenario, chunked saves about 2.5GB of memory per batch with max_seq_len=8192.
For bsz=2, this means 5GB saved.

Therefore, it seems an issue with compile + AC. If we solve that, we can compile model + loss separately, use chunked loss and have the best of all worlds: low compile time, low memory, high toks/sec.
image

edit: we decided to compile the model per layer. This doesnt match the speed of compile(model + loss), but get us closer to that, while having a much faster compile time (~40s vs 10min for Evan).

Overall:
- We reduced peak memory and should see higher impact as bsz and seq length grows;
- We removed an extra copy of logits, reducing memory further;
- We greatly improved compile time by only having to compile the model, instead of the loss_step (in my machine: 200s -> 20s);

### Things to be aware of
Compiling the upcasting + CE changes the loss by a small amount. According to the compile team, this is expected. We would have to be aware about it in future parity checks. This is independent of chunked cross entropy.

Edit: when training for 200 steps, there is no difference in loss up to the 2nd decimal point.

### Design

We have two main options: setter vs wrapper

1) setter (this PR): Directly changes the TransformerDecoder, and adds to it:

```
def set_num_output_chunks(self, num_output_chunks: int) -> None:
        """Use to save memory in combination with ChunkedCrossEntropy (TODO: add link to ChunkedCrossEntropy)"""
        self.num_output_chunks = num_output_chunks

def forward(...):

	if self.num_output_chunks > 0:
		# chunk for ChunkedCrossEntropyLoss, saving memory
		output = [
			self.output(chunk) for chunk in h.chunk(self.num_output_chunks, dim=1)
		]
	else:
		output = self.output(h)
```

The set_num_output_chunks is called in the recipe:
```
self.num_output_chunks = getattr(self._loss_fn, "num_output_chunks", 0)
model.set_num_output_chunks(self.num_output_chunks)
```

If no one intentionally calls set_num_output_chunks, the behavior of transformer is unchanged, and will work as expected during inference.

pros:
- The contract of the transformer is clear: the output may be a list. Nothing is hidden from the user.
- It is simple: no need to worry about checkpointing/FSDP wrapping/other functions
Cons:
- Transformer decoder is a bit more polluted

2) Wrapper: wrap the output layer with the chunk logic
```
class ChunkedOutputWrapper(nn.Module):
    def __init__(self, output_layer, num_output_chunks: int = 16):
        super(ChunkedOutputWrapper, self).__init__()
        self.output_layer = output_layer
        self.num_output_chunks = num_output_chunks

    def forward(self, x):
        return [self.output_layer(chunk) for chunk in x.chunk(self.num_output_chunks, dim=-2)]

    def state_dict(self, destination=None, prefix='', keep_vars=False):
        return self.output_layer.state_dict(destination=destination, prefix=prefix + 'output_layer.', keep_vars=keep_vars)

    def load_state_dict(self, state_dict, strict=True):
        # Adjust keys in state_dict
        adjusted_state_dict = {key[len('output_layer.'):]: value for key, value in state_dict.items() if key.startswith('output_layer.')}
        return self.output_layer.load_state_dict(adjusted_state_dict, strict=strict)
```

call it in the recipe:
```
self.num_output_chunks = getattr(self._loss_fn, "num_output_chunks", 0)
if hasattr(model, 'output') and self.num_output_chunks > 0:
            model.output = ChunkedOutputWrapper(model.output, num_output_chunks=self.num_output_chunks)
```

Pros:
- No extra lines in the TransformerDecoder

cons:
- Having to worry about any other logic that checks if the layer is a LoRA layer
- Getting checkpoint right
- User cant see what is happening in the transformer. They would have to realize that a wrapper was applied after the instantiation.

Results:
Apparently compile has some issues with the wrapper. It also causes higher reserved memory (prob due to graph break?). Maybe these can be solved?
image
"

Other alternatives that were ruled out:
- Adding it to the builders: This would require changing many files and add extra logic not only in the builder, but also in the lora module and configs.

#### Changelog
- Added Chunked Cross Entropy class
- Added if/else to TransformerDecoder to do chunk if self.num_output_chunks > 0, else keep normal behavior
- Changed compile to compile only the model, and not the loss_step
- Changed the label + logit off by one operation to be more efficient
- Moved .float() from transformer and added it to the chunked cross entropy

TODO:
[ ] add unit tests after reviewers are ok with the changes.
[ ] consider removing the step loss, since we wont compile it with the model anymore. 
[ ] run 100+ steps for every recipe
[ ] add srt
[ ] improve docstring

#### Test plan

llama lora single
image

```
tune run lora_finetune_single_device --config llama3_1/8B_qlora_single_device optimizer_in_bwd=False enable_activation_checkpointing=True loss=torchtune.modules.loss.CEWithChunkedOutputLoss  optimizer._component_=torch.optim.AdamW compile=True dataset.packed=False dataset.train_on_input=True tokenizer.max_seq_len=2048 metric_logger=torchtune.utils.metric_logging.WandBLogger metric_logger.project=chunked_CE log_every_n_steps=1 log_peak_memory_stats=True gradient_accumulation_steps=1 epochs=1 batch_size=16 max_steps_per_epoch=20 metric_logger.name=8b__vanilaCE__lora__AC checkpointer.output_dir=/tmp/Meta-Llama-3.1-8B-Instruct/8b__vanilaCE__lora__AC
```

llama full single
image

```
tune run full_finetune_single_device --config llama3_1/8B_full_single_device optimizer_in_bwd=False enable_activation_checkpointing=True loss=torchtune.modules.loss.CEWithChunkedOutputLoss optimizer._component_=bitsandbytes.optim.PagedAdamW8bit compile=True dataset.packed=False dataset.train_on_input=True tokenizer.max_seq_len=2048 metric_logger=torchtune.utils.metric_logging.WandBLogger metric_logger.project=chunked_CE log_every_n_steps=1 log_peak_memory_stats=True gradient_accumulation_steps=1 epochs=1 batch_size=16 max_steps_per_epoch=20 metric_logger.name=8b__chunkedCE__full__AC checkpointer.output_dir=/tmp/Meta-Llama-3.1-8B-Instruct/8b__chunkedCE__full__AC
```

qwen distributed lora
image

```
tune run --nnodes 1 --nproc_per_node 8 full_finetune_distributed --config qwen2/0.5B_full optimizer_in_bwd=False enable_activation_checkpointing=True loss=torchtune.modules.loss.CEWithChunkedOutputLoss optimizer._component_=torch.optim.AdamW compile=True dataset.packed=False dataset.train_on_input=True tokenizer.max_seq_len=2048 metric_logger=torchtune.utils.metric_logging.WandBLogger metric_logger.project=chunked_CE log_every_n_steps=1 log_peak_memory_stats=True gradient_accumulation_steps=1 epochs=1 batch_size=32 max_steps_per_epoch=20 metric_logger.name=qwen__chunkedCE__nrpoc8_full__AC checkpointer.output_dir=/tmp/Meta-Llama-3.1-8B-Instruct/qwen__chunkedCE__nrpoc8_full__AC
```

gemma distributed full
image

```
tune run --nnodes 1 --nproc_per_node 8 lora_finetune_distributed --config gemma/2B_lora optimizer_in_bwd=False enable_activation_checkpointing=True loss=torchtune.modules.loss.CEWithChunkedOutputLoss optimizer._component_=torch.optim.AdamW compile=True dataset.packed=False dataset.train_on_input=True tokenizer.max_seq_len=2048 metric_logger=torchtune.utils.metric_logging.WandBLogger metric_logger.project=chunked_CE log_every_n_steps=1 log_peak_memory_stats=True gradient_accumulation_steps=1 epochs=1 batch_size=16 max_steps_per_epoch=20 metric_logger.name=gemma2b__chunkedCE__nrpoc8_lora__AC checkpointer.output_dir=/tmp/Meta-Llama-3.1-8B-Instruct/gemma2b__chunkedCE__nrpoc8_lora__AC
```

QWEN QAT

```
tune run --nnodes 1 --nproc_per_node 8 qat_distributed --config qwen2/0.5B_full optimizer_in_bwd=False enable_activation_checkpointing=True loss=torchtune.modules.loss.CEWithChunkedOutputLoss optimizer._component_=torch.optim.AdamW compile=True dataset.packed=False dataset.train_on_input=True tokenizer.max_seq_len=2048 metric_logger=torchtune.utils.metric_logging.WandBLogger metric_logger.project=chunked_CE log_every_n_steps=1 log_peak_memory_stats=True gradient_accumulation_steps=1 epochs=1 batch_size=32 max_steps_per_epoch=20 metric_logger.name=qwen__chunkedCE__nrpoc8_full__AC checkpointer.output_dir=/tmp/Meta-Llama-3.1-8B-Instruct/qwen__chunkedCE__nrpoc8_full__AC quantizer._component_=torchtune.utils.quantization.Int8DynActInt4WeightQATQuantizer quantizer.groupsize=256
``` 

image

#### UX


gau-nernst:
Just curious, do you have data for "compile step loss + chunk CE"?

> We greatly improved compile time by only having to compile the model, instead of the loss_step

How much longer was it for you? I didn't notice any significant difference on my machine when I added the compile step loss.

> We removed an extra copy of logits, reducing memory further

Yea I have always thought the logits slicing is a bit iffy (HF is doing this too :laughing:). Technically due to this, the number of tokens seen is `seq_len-1` since we remove 1 token from loss calculation. Still think that right shift the labels should be done in the dataset instead, but it would be a separate discussion.

torchtune/modules/transformer.py
```diff
@@ -460,8 +465,14 @@ def forward(
         # shape: [b, s, d]
         h = self.norm(h)
 
-        # shape: [b, s, out_dim] - out_dim is usually the vocab size
-        output = self.output(h).float()
+        # shape: [b, seq_len//num_chunks, out_dim] - out_dim is usually the vocab size
+        if self.num_output_chunks > 0:
+            # chunk for ChunkedCrossEntropyLoss, saving memory
```
awgu, line 470: My hypothesis was that by chunking before the linear, the output chunks are not views into the same underlying tensor allowing earlier ones to be freed and have their memory reused for later ones.

I wonder if we can check the memory snapshot to confirm that if not already, as it might be nice to understand the constraint on the modeling side (namely, having to chunk the _input_ to the final output projection rather than chunking the output itself).

awgu (commented):
exciting!

felipemello1:
@gau-nernst 
> Just curious, do you have data for "compile step loss + chunk CE"?

Its a mess. It consumes more memory and takes 10+ min to do 5 steps, because it keeps recompiling on every step.
image

> How much longer was it for you?

~1.5min
image

> Still think that right shift the labels should be done in the dataset instead

I see your point, and it makes sense. But I guess that having it exposed in the recipe gives more flexibility to the user.

felipemello1:
@awgu 
> I wonder if we can check the memory snapshot to confirm that if not already, as it might be nice to understand the constraint on the modeling side (namely, having to chunk the input to the final output projection rather than chunking the output itself).

I will get it for you. If we can chunk after the linear, that would be beautiful

felipemello1:
One update: when training for longer, compiling loss + model consumes more memory, but it is faster. Working with @weifengpy to see if we can make compile work with chunked + compile together.


image

gau-nernst:
I think two important things to understand is
1. How compiling model+loss together gives such a significant speedup.
2. Where are the memory savings come from. I believe what Andrew said is probably right. Also try to see if memory savings are still there if chunking is done after.

I try to reproduce 1. in a standalone script with Linear (LM head) + CE but couldn't reproduce the speedup. Perhaps something else is happening here :eyes: 

Another question. Is there a need to upcast logits after LM head to FP32? From what I understand, CE will upcast logits to FP32 anyway internally, both in eager and compile mode.

felipemello1:
> Another question. Is there a need to upcast logits after LM head to FP32? From what I understand, CE will upcast logits to FP32 anyway internally, both in eager and compile mode.

I dont know :/ . Note: compiling the upcast with the CE has a small difference in the loss.

>I try to reproduce 1. in a standalone script with Linear (LM head) + CE but couldn't reproduce the speedup. Perhaps something else is happening here 👀

Yeah, i would love to have the best of all worlds: speed up + fast compile + lowest memory

felipemello1:
Ok, I confirmed it: If activation checkpointing is FALSE, then:

toks/sec
compiling model+loss together == chunked + compile model + loss separetely

however, if checkpointing is TRUE, then
compiling model+loss together >> chunked + compile model + loss separetely

In every scenario, chunked saves about 3GB of memory per batch with max_seq_len=8192.
For bsz=2, this means 6GB saved.

image

felipemello1:
@gau-nernst 
> From what I understand, CE will upcast logits to FP32 anyway internally, both in eager and compile mode.

Loss is different if i remove .float()

recipes/lora_finetune_single_device.py
```diff
@@ -537,18 +546,28 @@ def save_checkpoint(self, epoch: int) -> None:
...
-        # Shift so that tokens < n predict n
-        logits = logits[..., :-1, :].contiguous()
-        labels = labels[..., 1:].contiguous()
-        logits = logits.transpose(1, 2)
+
+        # Shift labels to compute loss
+        # equivalent to doing labels[..., 1:] and logits[..., :-1, :]
+        # But this way we dont need to slice the logits. We just add an ignore index to labels.
+        labels = torch.hstack(
+            (labels[..., 1:], self.ignore_labels_cache[: labels.shape[0]])
```
kartikayk, line 562: Sorry if I'm being dense but isn't labels.shape[0] always going to be batch_size which is also what the first dim of the cache is? If so, why do we need to slice?

recipes/lora_finetune_single_device.py
```diff
@@ -537,18 +546,28 @@ def save_checkpoint(self, epoch: int) -> None:
...
+        )
+        if not isinstance(logits, list):
```
kartikayk, line 564: Ah :(

torchtune/modules/loss/chunked_cross_entropy.py
```diff
@@ -0,0 +1,55 @@
...
+import torch
+
+
+class ChunkedCrossEntropyLoss(torch.nn.Module):
+    """
+    TODO: explain why it saves memory
+    """
+
+    def __init__(self, num_output_chunks: int = 16, ignore_index: int = -100):
+        super(ChunkedCrossEntropyLoss, self).__init__()
```
kartikayk, line 19: why not ```super().__init__()```? I think thats how we do this across the code base?

torchtune/modules/loss/chunked_cross_entropy.py
```diff
@@ -0,0 +1,55 @@
...
+        self.num_output_chunks = num_output_chunks
+        self.ignore_index = ignore_index
+        self.cross_entropy_loss = torch.nn.CrossEntropyLoss(
+            reduction="sum", ignore_index=self.ignore_index
```
kartikayk, line 23: Why the change in reduction? I think previously we were using the default?

torchtune/modules/transformer.py
```diff
@@ -460,8 +465,14 @@ def forward(
...
+            output = [
+                self.output(chunk) for chunk in h.chunk(self.num_output_chunks, dim=1)
+            ]
+        else:
+            output = self.output(h)
```
kartikayk, line 475: Sorry if we already discussed this, but why move the float call from here to the recipe? Easy for recipe writers to trip up and doesnt seem like its conceptually different?

torchtune/modules/transformer.py
kartikayk: This has been bugging me for a while and I still don't understand this. Where would they be freed though? We pass the entire list of chunks to the loss? Or did I misunderstand your point?

recipes/lora_finetune_single_device.py
felipemello1: I believe that we need to for the last batch of the dataloader, which may be of a different size.

Eg: you have 10 samples, but bsz=3. Then, unless we have drop_last=True in the dataloader, the last batch will have size=1.

torchtune/modules/loss/chunked_cross_entropy.py
felipemello1: The default is mean. I can do the mean, sum the means, and divide by the number of chunks. It should be equivalent and probably a bit faster

gau-nernst: I think your current way is more correct. When there are different paddings across samples in a batch (i.e. number of ignore_index is different), taking the mean of (mean of each chunk) != (total loss of non_padded tokens / no. of non_padded_tokens)

torchtune/modules/transformer.py
felipemello1: its not the the chunks are fred. If we split AFTER, then we have chunks + whole output. When we split BEFORE, we have only the chunks.

The memory savings is because, when we have chunks, the extra memory by upcasting is at most of size of one chunk, instead of doubling the whole output.

Ideally, if we split AFTER, pytorch should learn to delete the whole output that we have no need for anymore. But this doesnt happen.

image
image

torchtune/modules/transformer.py
felipemello1: thats true! I can put the float back here. It doesnt need to be in the recipe.

torchtune/modules/loss/chunked_cross_entropy.py
felipemello1: oh, right! I would have to do a few more operations for it to be numerically the same. Thats why i computed the sum instead of the mena.

gau-nernst:
Some interesting triton kernels from Liger-Kernel, might be interesting to test them too https://github.com/linkedin/Liger-Kernel
- Cross entropy: fused forward and backward. Drop-in replacement by using `liger_kernel.transformers.cross_entropy.LigerCrossEntropyLoss`
- Fused linear cross entropy: also implement pre-LM-head chunking logic. Integration will be harder.

torchtune/modules/loss/ce_chunked_output_loss.py
```diff
@@ -0,0 +1,82 @@
...
+
+import torch
+import torch.nn.functional as F
+
+
+class CEWithChunkedOutputLoss(torch.nn.Module):
+    """
+    CE with chunked outputs that saves memory by only upcasting one chunk at a time.
+
+    Since the model is trained with fp16, before running CE, we have to upcast
```
ebsmothers, line 17: should be this, right? Alternatively can just keep it general and say something like "16-bit precision" to cover all cases 
```suggestion
    Since the model is trained with bf16, before running CE, we have to upcast
```

torchtune/modules/loss/ce_chunked_output_loss.py
```diff
@@ -0,0 +1,82 @@
...
+    it to fp32 for better accuracy and stability. When upcasting happens, the memory usage doubles.
+    Models like llama3 have large vocabulary size and, therefore, have a large output
+    result (bsz, num_tokens, vocab_size). If we chunk on the token level, you can still compute
+    the cross entropy normally, but upcasting only one chunk at a time saves considerable memory.
+
+    The CE and upcasting have to be compiled together for better performance.
+    Also, compiling CE always yields great memory gains.
```
ebsmothers, line 24: nit: I would leave absolutes like this out of the docstring (fine to say something like "When using this class, we recommend compiling either `_compute_cross_entropy` or the entire class for reduced memory")

recipes/qat_distributed.py
```diff
@@ -235,7 +237,22 @@ def setup(self, cfg: DictConfig) -> None:
             else None,
         )
 
+        # initialize loss
         self._loss_fn = config.instantiate(cfg.loss)
+        backend = os.environ.get("TORCH_COMPILE_BACKEND", "inductor")
+        if self._loss_fn.__class__.__name__ == "CEWithChunkedOutputLoss":
```
ebsmothers, line 243: Why not just use `isinstance` here?

recipes/qat_distributed.py
```diff
@@ -235,7 +237,22 @@ def setup(self, cfg: DictConfig) -> None:
...
+            # set num_output_chunks for model
+            self._model.set_num_output_chunks(self._loss_fn.num_output_chunks)
+            if self._model_compile:
+                log.info("Compiling loss with torch.compile...")
+                self._loss_fn._compute_cross_entropy = torch.compile(
+                    self._loss_fn._compute_cross_entropy, backend=backend
+                )
```
ebsmothers, line 250: Could be good to add a quick comment here around why we compile this method specifically (I know you already explain in detail in the class, but this is more visible)

recipes/lora_finetune_single_device.py
ebsmothers: Yeah no way around this one unfortunately. But it's worth it for the memory wins 😃

recipes/full_finetune_distributed.py
```diff
@@ -202,6 +203,7 @@ def setup(self, cfg: DictConfig) -> None:
 
         checkpoint_dict = self.load_checkpoint(cfg_checkpointer=cfg.checkpointer)
 
+        self._model_compile = cfg.get("compile", False)
```
ebsmothers, line 206: We should definitely land distributed FFT model compile as a fast follow here; until then the behavior will be confusing to folks

ebsmothers (approved):
A handful of small comments (for whatever reason I left them all in the QAT recipe, but ofc applicable to all recipes). But no major concerns from me here. Really excited to get this in and realize these memory savings!

recipes/qat_distributed.py
felipemello1: either I would have to import the class just for this check, or i could use modules.loss.CEWithChunkedOutputLoss, since modules is already imported. However, for some reason i was getting error "loss is not in torchtune.modules". So it seemed to be the easies way to avoid an extra import

ebsmothers: Hmm weird. It's definitely OK to import from modules in the recipe and imo a little bit cleaner (even if it does require the extra import). We already import from mouldes in our LoRA recipes for various PEFT stuff though we don't in full finetune recipes. Anyways no strong preference here, definitely not a blocker

felipemello1:
warning: DO NOT MERGE - need to fix some regression tests

felipemello1:
@SLR722, this changed all default configs. So before it lands in genie, you would have to update the recipes.

PR #1403, opened 2024-08-23
Update quantization to use tensor subclasses

andrewor14:
**Summary:** In torchao, we are migrating our quantization flows from module swap to tensor subclasses. The existing `Int8DynActInt4WeightQuantizer` will be deprecated in the near future in favor of quantizing using the `quantize_` API, so we should do the same in torchtune. This quantizer is currently only used by QAT, which also recently migrated to a tensor subclass implementation.

This also changes the eval script slightly since models quantized through the torchao tensor subclasses are expected to be loaded with `assign=True` ([see this test](https://github.com/pytorch/ao/blob/9a56e80cb6070599701b8f5f587bd8187c8dccb4/test/quantization/test_quant_api.py#L610)). We should load the model similarly in torchtune.

**Test Plan:**

Quantized and evaluated the base Llama3-8B model on 1 A100 GPU:

```
CUDA_VISIBLE_DEVICES=1 tune run quantize --config recipes/configs/quantization.yaml \
    model._component_=torchtune.models.llama3.llama3_8b \
    checkpointer._component_=torchtune.training.FullModelMetaCheckpointer \
    checkpointer.checkpoint_dir=/tmp/Meta-Llama-3-8B-Instruct/original \
    checkpointer.output_dir=/tmp/Meta-Llama-3-8B-Instruct/original \
    checkpointer.checkpoint_files=[consolidated.00.pth] \
    checkpointer.model_type=LLAMA3

CUDA_VISIBLE_DEVICES=1 tune run eleuther_eval --config eleuther_evaluation \
    model._component_=torchtune.models.llama3.llama3_8b \
    checkpointer._component_=torchtune.training.FullModelTorchTuneCheckpointer \
    checkpointer.checkpoint_dir=/tmp/Meta-Llama-3-8B-Instruct/original \
    checkpointer.output_dir=/tmp/Meta-Llama-3-8B-Instruct/original \
    checkpointer.checkpoint_files=[consolidated-8da4w.pt] \
    checkpointer.model_type=LLAMA3 \
    tokenizer._component_=torchtune.models.llama3.llama3_tokenizer \
    quantizer._component_=torchtune.training.quantization.Int8DynActInt4WeightQuantizer
```

**Reviewers:** ebsmothers, kartikayk, RdoubleA

**Subscribers:** ebsmothers, kartikayk, RdoubleA

### DIFF
--- a/recipes/eleuther_eval.py
+++ b/recipes/eleuther_eval.py
@@ -222,11 +222,15 @@ class EleutherEvalRecipe(EvalRecipeInterface):
     ) -> nn.Module:
         with training.set_default_dtype(self._dtype), self._device:
             model = config.instantiate(model_cfg)
+
         if self._quantization_mode is not None:
             model = self._quantizer.quantize(model)
             model = model.to(device=self._device, dtype=self._dtype)
-
-        model.load_state_dict(model_state_dict)
+            for k, v in model_state_dict.items():
+                model_state_dict[k] = v.to(self._device)
+            model.load_state_dict(model_state_dict, assign=True)
+        else:
+            model.load_state_dict(model_state_dict)
 
         # Put model in eval mode.
         # Note: This will not disable the dropout applied in SDPA,
--- a/torchtune/training/quantization.py
+++ b/torchtune/training/quantization.py
@@ -6,8 +6,23 @@
 
 from typing import Callable, Optional
 
+from torchao.quantization import int8_dynamic_activation_int4_weight, quantize_
+from torchao.quantization.prototype.qat import (
+    disable_8da4w_fake_quant,
+    enable_8da4w_fake_quant,
+    Int8DynActInt4WeightQATQuantizer,
+)
+from torchao.quantization.prototype.qat._module_swap_api import (
+    disable_8da4w_fake_quant_module_swap,
+    enable_8da4w_fake_quant_module_swap,
+    Int8DynActInt4WeightQATQuantizerModuleSwap,
+)
+
+
 __all__ = [
     "get_quantizer_mode",
+    "Int8DynActInt4WeightQuantizer",
+    "Int8DynActInt4WeightQATQuantizer",
 ]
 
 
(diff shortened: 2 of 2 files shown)


torchtune/utils/quantization.py
```diff
@@ -22,7 +22,19 @@
 
 
 if TORCH_VERSION_AFTER_2_3:
-    from torchao.quantization.quant_api import Int8DynActInt4WeightQuantizer
+    from torchao.quantization.quant_api import (
```
kartikayk, line 25: @ebsmothers Can you remind me what our principles are around version guards? now that 2.4 is launched do we just claim we work with stable and remove such guards? or what's the downside of doing this?

torchtune/utils/quantization.py
```diff
@@ -22,7 +22,19 @@
...
+        int8_dynamic_activation_int4_weight,
+        quantize_,
+    )
+
+    class Int8DynActInt4WeightQuantizer:
```
kartikayk, line 30: I know it's unrelated to this PR, but can I request a doc string here which explains this a bit more. Pointing back to ao docs is fine too

torchtune/utils/quantization.py
ebsmothers: Yeah we can remove, our assumption is that users are at least on the latest stable version of PyTorch (so at this moment 2.4)

jerryzh168, line 25: @andrewor14 nit: we can import from `torchao.quantization` directly

torchtune/utils/quantization.py
```diff
@@ -6,39 +6,51 @@
 
 from typing import Callable, Optional
 
-# importing TORCH_VERSION_AFTER_2_3 because `Int8DynActInt4WeightQuantizer`
-# is only available after 2.3 so we have to guard the pytorch versions to decide
-# the list of supported quantizers
-from torchao.utils import TORCH_VERSION_AFTER_2_3, TORCH_VERSION_AFTER_2_4
```
andrewor14, line 12: cc @msaroufim @ebsmothers Just wanted to confirm this is OK

ebsmothers: Sorry to ask for the 100th time, but TORCH_VERSION_AFTER_2_4 returns True if and only if the PyTorch version is >= 2.4, right? If so we can remove both of these, since we assume everyone is on at least latest stable PyTorch

ebsmothers (commented):
Overall the changes look reasonable to me. My main question is whether there is any logic that's BC breaking (i.e. can I run this as is on ao's 0.3.1, latest nightly, and anything in between?)

torchtune/utils/quantization.py
andrewor14: Mark can confirm but yes I think `TORCH_VERSION_AFTER_2_4` means >= 2.4

andrewor14:
> Overall the changes look reasonable to me. My main question is whether there is any logic that's BC breaking (i.e. can I run this as is on ao's 0.3.1, latest nightly, and anything in between?)

Good question. I haven't tested an older version of ao, but this change is needed to run the full end-to-end flow after the QAT subclass refactor. If there are concerns about breaking BC, maybe we should wait until we upgrade to torchao 0.5.0? (expected maybe mid-September)

torchtune/utils/quantization.py
felipemello1: I believe it is true only in torchao nightlies. This was fixed. Before it was >2.4

andrewor14:
Hi @ebsmothers, I think this is ready from my side. I saw that the torchao version is actually removed from pyproject.toml. Do we expect torchtune to only work with the latest version? Either way I tested the quantization and eval with both the module swap QAT (old torchao version) and the tensor subclass QAT (latest torchao version) and both work. Please take another look. Thanks.

ebsmothers:
@andrewor14 thanks for the ping. Yeah currently we will install whatever the latest stable version of torchao is. This is what we do in our CI and what we recommend our users to do as well (ofc we also support nightlies). So yes, the latest version of torchao should be sufficient here.

ebsmothers (approved):
This looks good to me, thanks! Once CI is green I think we're good to merge

gau-nernst:
Just curious, does this work with FSDP2 recipes? Trying out INT8 mixed-precision with torchtune and I saw that in distributed recipes, load state dict uses `assign=True`, which will override the tensor subclass.

https://github.com/pytorch/torchtune/blob/ee343e61804f9942b2bd48243552bf17b5d0d553/torchtune/training/_distributed.py#L346

This is because we call torchao's `quantize_()` before loading state dict. @andrewor14 Did you check that the subclass ops are actually called?

One fix for this is to add DTensor support for `quantize_()` API (https://github.com/pytorch/ao/issues/803), which allows us to do `quantize_()` after fully_shard() and load state dict.

PR #1405, opened 2024-08-24
[WIP] Fix batched generation with KV Cache

### LINKED ISSUE #1250: Fix generation for bsz > 1
Our modules only work with generation under two conditions: batch_size = 1 or every single sample in a batch has the same length. The main culprit is this line of code: https://github.com/pytorch/torchtune/blob/288ff4435b0cf17325b5c3b112f6859a6cdf0ea2/torchtune/modules/transformer.py#L167

For a batch that looks like the following:
```
My, name, is, Joe
Hello, world, <PAD>, <PAD>
Bye, <PAD>, <PAD>, <PAD>
```

A proper mask would look like:
```
1 0 0 0
1 1 0 0 
1 1 1 0
1 1 1 1

1 0 0 0
1 1 0 0
0 0 0 0
0 0 0 0

1 0 0 0
0 0 0 0
0 0 0 0
0 0 0 0
```

of size [b x s x s], which is [3 x 4 x 4]

This will be a fairly involved change that touches several utils and modules. The general changes needed will be:
* Delete causal mask from KV cache, instead opting for this to come in the `mask` param
* Update generate utils to pass a mask into its call to `model.forward()`
* Modify eleuther eval recipe to construct proper causal mask to pass to 

***

This was originally found and reported by @iankur

joecummings:
#### Context
- [ ] add a new feature
- [x] fix a bug
- [ ] other (please add here)

Closes #1250

#### Test plan

> This is super ugly and we should probably be doing some prefilling, but I just wanted to get this working.

### DIFF
--- a/recipes/eleuther_eval.py
+++ b/recipes/eleuther_eval.py
@@ -13,7 +13,6 @@ import torch
 from omegaconf import DictConfig
 
 from torch import nn
-from torch.nn.utils.rnn import pad_sequence
 
 from torchtune import config, utils
 from torchtune.modules import TransformerDecoder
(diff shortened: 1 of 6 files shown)


joecummings:
I don't think this needs to get in before I'm back next Friday, but if someone wants to commandeer, feel free.

SalmanMohammadi:
Thanks for putting this up. Will keep the ship steady while you're away and try integrate into PPO. 

enjoy your time off : )

SalmanMohammadi:
https://github.com/pytorch/torchtune/pull/1424

SalmanMohammadi:
Closing after landing #1424

PR #1415, opened 2024-08-26
force uninstall of stable torchao on nightly CI jobs

ebsmothers:
Right now our install commands always install torchao==0.3.1 regardless of whether we install PyTorch stable or nightly. 

As a result, our CI is (and has been) red: there is a change in core nightlies that breaks ao 0.3.1. 

I claim we should actually just remove ao from our pyproject.toml to give the flexibility to either (a) install a pinned version when using stable PyTorch, or (b) install ao nightlies when using PyTorch nightlies. This might be a bit controversial though, so until then.. 

This PR adds a hack to force uninstall torchao in our failing CI jobs so that nightly torchao can be installed properly. This should make our CI green again.

Test plan: I hacked in a trigger on PRs for the nightly recipe test workflow, so that signal should be green if this works. Will attach a screenshot here once it finishes

### DIFF
--- a/.github/workflows/recipe_test_nightly.yaml
+++ b/.github/workflows/recipe_test_nightly.yaml
@@ -5,6 +5,7 @@ on:
     # Runs at midnight every day
     - cron:  '0 0 * * *'
   workflow_dispatch:
+  pull_request:
 
 concurrency:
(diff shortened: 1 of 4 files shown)


RdoubleA (approved):
can also do `pip install --force-reinstall`. just tested this and it does indeed install nightly over the stable

ebsmothers:
Closing this in favor of #1491 (and other stuff that has been landed already)

Issue #1289, opened 2024-08-08
Change self.compile in recipes


SalmanMohammadi:
nit: `self._is_model_compiled` or similar is a bit more clear

_Originally posted by @RdoubleA in https://github.com/pytorch/torchtune/pull/1223#discussion_r1710027730_

SalmanMohammadi:
closing in favour of #1426

PR #1419, opened 2024-08-27
Add per-layer compile support to recipes

yf225:
Enabling per-layer compile for our single-device LoRA, single-device full finetune, and FSDP2 LoRA recipes. FSDP2 full finetune will be done in a follow-up.

## Results

All recipes were run with three different configurations: (1) per-layer compile (this PR), (2) full-model compile (i.e. `compile=True` on main), (3) no compile. 

#### WandB results

Screenshot 2024-08-27 at 10 40 51 PM

#### Repro

```
TORCHINDUCTOR_FORCE_DISABLE_CACHES=1 CUDA_VISIBLE_DEVICES=0 tune run lora_finetune_single_device \
--config llama3/8B_qlora_single_device model.lora_rank=16 optimizer=bitsandbytes.optim.AdamW8bit \
 gradient_accumulation_steps=4 tokenizer.max_seq_len=2048 max_steps_per_epoch=100 \
 metric_logger=torchtune.utils.metric_logging.WandBLogger metric_logger.project=pr-1419 compile=True \
 log_peak_memory_stats=True metric_logger.name=qlora-per-layer-compile
```

#### WandB results
Screenshot 2024-08-27 at 10 41 40 PM

#### Repro

```
TORCHINDUCTOR_FORCE_DISABLE_CACHES=1 CUDA_VISIBLE_DEVICES=1 tune run \
lora_finetune_single_device --config llama3/8B_lora_single_device model.lora_rank=16 \
 optimizer=bitsandbytes.optim.AdamW8bit gradient_accumulation_steps=4 tokenizer.max_seq_len=2048 \
 max_steps_per_epoch=100 model.lora_attn_modules=['q_proj','k_proj','v_proj','output_proj'] \
 model.apply_lora_to_mlp=True metric_logger=torchtune.utils.metric_logging.WandBLogger \
metric_logger.project=pr-1419 compile=True log_peak_memory_stats=True metric_logger.name=lora-per-layer-compile
```

#### WandB results
Screenshot 2024-08-27 at 10 42 12 PM

#### Repro

```
TORCHINDUCTOR_FORCE_DISABLE_CACHES=1 CUDA_VISIBLE_DEVICES=2 \
tune run full_finetune_single_device --config llama3/8B_full_single_device \
dataset=torchtune.datasets.alpaca_cleaned_dataset optimizer=bitsandbytes.optim.AdamW8bit \
gradient_accumulation_steps=4 tokenizer.max_seq_len=2048 max_steps_per_epoch=100 epochs=1 \
optimizer_in_bwd=False metric_logger=torchtune.utils.metric_logging.WandBLogger metric_logger.project=pr-1419 \
metric_logger.name=fft-per-layer-compile compile=True log_peak_memory_stats=True
```

### LoRA FSDP2

Repro only to ensure things aren't broken, will work on perf in a follow-up

#### Repro
First change dev/llama2/70B_qlora_fsdp2.yaml recipe to [this](https://gist.github.com/ebsmothers/43ae789c5d3db397ebc5ffbb932486b6), then run

```
tune run --nproc_per_node 2 lora_finetune_fsdp2 --config llama2/70B_qlora max_steps_per_epoch=10
1|10|Loss: 1.7545080184936523: 100%|█████████████| 10/10 [06:03<00:00, 36.29s/it]
```

Also consolidated E2E time over 100 steps and compile time for QLoRA, LoRA, and full finetunes of Llama3 8B on single device. Note that compile on main currently OOMs due to some kind of memory leak, so we don't have E2E times for the models compiled on main. 

Screenshot 2024-08-27 at 10 47 31 PM

### DIFF
--- a/recipes/dev/lora_finetune_fsdp2.py
+++ b/recipes/dev/lora_finetune_fsdp2.py
@@ -208,6 +208,9 @@ class LoRAFinetuneRecipeDistributed(FTRecipeInterface):
             self._metric_logger.log_config(cfg)
 
         checkpoint_dict = self.load_checkpoint(cfg_checkpointer=cfg.checkpointer)
+        self._model_compile = cfg.compile
+        self._per_layer_compile = cfg.per_layer_compile
+        assert self._model_compile and self._per_layer_compile
 
         self._model = self._setup_model(
             cfg_model=cfg.model,
@@ -230,6 +233,10 @@ class LoRAFinetuneRecipeDistributed(FTRecipeInterface):
         )
 
         self._loss_fn = config.instantiate(cfg.loss)
+        if self._model_compile and self._per_layer_compile:
+            log.info("Compiling loss with torch.compile...")
+            backend = os.environ.get("TORCH_COMPILE_BACKEND", "inductor")
+            self._loss_fn = torch.compile(self._loss_fn, backend=backend)
 
         # sampler and dataloader depend on the tokenizer and loss_fn and should be
         # setup after all of these are setup
@@ -298,6 +305,13 @@ class LoRAFinetuneRecipeDistributed(FTRecipeInterface):
 
         self.adapter_params = get_adapter_params(model)
         set_trainable_params(model, self.adapter_params)
+
+        print(f"model: {model}")
+        if self._model_compile and self._per_layer_compile:
+            backend = os.environ.get("TORCH_COMPILE_BACKEND", "inductor")
+            for m in reversed(list(model.modules())):
+                if isinstance(m, modules.TransformerSelfAttentionLayer):
+                    m.compile(backend=backend)
 
         if enable_activation_checkpointing:
             utils.set_activation_checkpointing(
--- a/recipes/dev/lora_finetune_fsdp2.py
+++ b/recipes/dev/lora_finetune_fsdp2.py
@@ -209,8 +209,6 @@ class LoRAFinetuneRecipeDistributed(FTRecipeInterface):
 
         checkpoint_dict = self.load_checkpoint(cfg_checkpointer=cfg.checkpointer)
         self._model_compile = cfg.compile
-        self._per_layer_compile = cfg.per_layer_compile
-        assert self._model_compile and self._per_layer_compile
 
         self._model = self._setup_model(
             cfg_model=cfg.model,
@@ -233,7 +231,7 @@ class LoRAFinetuneRecipeDistributed(FTRecipeInterface):
         )
 
         self._loss_fn = config.instantiate(cfg.loss)
-        if self._model_compile and self._per_layer_compile:
+        if self._model_compile:
             log.info("Compiling loss with torch.compile...")
             backend = os.environ.get("TORCH_COMPILE_BACKEND", "inductor")
             self._loss_fn = torch.compile(self._loss_fn, backend=backend)
@@ -306,7 +304,8 @@ class LoRAFinetuneRecipeDistributed(FTRecipeInterface):
         self.adapter_params = get_adapter_params(model)
         set_trainable_params(model, self.adapter_params)
 
-        if self._model_compile and self._per_layer_compile:
+        if self._model_compile:
+            log.info("Compiling model layers with torch.compile...")
             backend = os.environ.get("TORCH_COMPILE_BACKEND", "inductor")
             for m in reversed(list(model.modules())):
                 if isinstance(m, modules.TransformerSelfAttentionLayer):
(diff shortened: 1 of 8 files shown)


felipemello1 (commented):
Thanks, Will!

a couple of thoughts: I personally dont like that we have two flags for compiling, and would prefer to avoid cases like:
```
compile: False
per_layer_compile: True
```

IMO, we should either:
1) If compile is true, have per_layer_compile by default
2) make compile a nested config, like:
```
compile:
	per_layer: True
	loss: True
```

I prefer 1, as I dont think that the user needs this level of control through configs.

Second:
Do you think you could test it for our distributed recipe that already uses FSDP2?

SalmanMohammadi:
Would this also enable optimizer in bwd 🤝 compile?

recipes/dev/lora_finetune_fsdp2.py
```diff
@@ -230,6 +231,10 @@ def setup(self, cfg: DictConfig) -> None:
         )
 
         self._loss_fn = config.instantiate(cfg.loss)
+        if self._model_compile:
+            log.info("Compiling loss with torch.compile...")
+            backend = os.environ.get("TORCH_COMPILE_BACKEND", "inductor")
+            self._loss_fn = torch.compile(self._loss_fn, backend=backend)
```
felipemello1, line 237: lets leave the loss compile out of this recipe. I can add it to the chunked CE, since it needs some extra logic.

recipes/full_finetune_single_device.py
```diff
@@ -222,6 +222,10 @@ def setup(self, cfg: DictConfig) -> None:
         )
 
         self._loss_fn = config.instantiate(cfg.loss)
+        if self._model_compile:
+            log.info("Compiling loss with torch.compile...")
+            backend = os.environ.get("TORCH_COMPILE_BACKEND", "inductor")
+            self._loss_fn = torch.compile(self._loss_fn, backend=backend)
```
felipemello1, line 228: chunked CE can only have compile on the CE + upcasting part. If the chunking is compiled with it, it loses the benefit :/

I think we can leave the loss compile outside of the PR, if chunked CE will be the default

gau-nernst:
Came across this PR. @yf225 Am I seeing memory leak for "qlora-compile-main"? If you are using torch nightly after 20240824, might be the same as what I'm seeing here https://github.com/pytorch/pytorch/issues/134642

ebsmothers:
> Came across this PR. @yf225 Am I seeing memory leak for "qlora-compile-main"? If you are using torch nightly after 20240824, might be the same as what I'm seeing here [pytorch/pytorch#134642](https://github.com/pytorch/pytorch/issues/134642)

@gau-nernst yeah same here. I'm not sure what's going on because I've been compiling models with no problems for a while. But as of a couple days ago I also started seeing what looks like a memory leak. Attaching a memory viz I ran on a full finetune; seems like a bunch of memory is not getting freed after each step.

455259770_1065769865167420_9145965881265563120_n

recipes/dev/lora_finetune_fsdp2.py
```diff
@@ -230,6 +231,10 @@ def setup(self, cfg: DictConfig) -> None:
         )
 
         self._loss_fn = config.instantiate(cfg.loss)
+        if self._model_compile:
```
felipemello1, line 234: nit: should we rename "model_compile" to just "compile", since this command will control all loss compile and flexattention compile?

felipemello1: not a priority, i am fine to keep as it for now, and maybe add a todo to rename

felipemello1 (approved):

recipes/dev/lora_finetune_fsdp2.py
ebsmothers: Yeah I think our naming here is not great. Let's do in a follow-up; rather than add a todo in the code I may just create an issue.

SalmanMohammadi: [#](https://github.com/pytorch/torchtune/issues/1289)

ebsmothers: #1426

yf225:
@gau-nernst @ebsmothers For the full-model compile memory leak issue, this should now be resolved by reverting https://github.com/pytorch/pytorch/pull/134272.

In this PR we've also switched to using per-layer compile (https://github.com/pytorch/torchtune/pull/1419), which also won't have the memory leak issue.

Issue #1421, opened 2024-08-27
[Break BC][RFC] utils directory refactor


RdoubleA:
Collecting all the discussions around this here from many private channels. @ebsmothers has already written up an excellent quick RFC on this topic (#1414) but I wanted to formalize the discussion here.

## Problem
`torchtune/utils/` is a massive directory with a wide spectrum of helpers, from things like `shard_model` for distributed training to `get_logger` to retrieve Python's built-in logger... this is not great. It's served as a "i don't know where else to put this function" type of folder but this is leading to tangible problems:
- Hinders discoverability of code since a lot of important functionality for running our recipes is found under utils, which is not the first place users will look nor is it specific enough to know how to navigate
- Complicates our dependency graph, leading to many cycles that are created as we continue to inflate the directory. Utils depends on models and modules, but also everything else in the library depends on utils, so it's very easy to create cycles.

Ideally, the utils directory remains a miscellaneous collection of helpers that any other directory in the library can take a dependency on, but this isn't the case with folders like `utils/_checkpointing` and `utils/_distributed.py` depending on `models/` and `modules/`

Here is the current dependency graph (A -> B means B depends on A). Notice how easy it is to create a cycle.
![image](https://github.com/user-attachments/assets/57a05717-8cdb-4ed2-bc16-6662e873192a)
If `data` took a dependency on `utils`, you would get a cycle. `modules` and `models` can never take a dependency on utils otherwise there's a clear cycle. `modules` and `models` cannot take a dependency on `config` or `dataset` because that would create a cycle. This is not scalable and is currently blocking important features, like #1193, and adding unnecessary tech debt, like deprecate being placed in `data` instead of `utils` to avoid a cycle (see #1286)

## Approach
We need a way to restructure our dependency graph. However, this would induce a massive refactor and would undoubtedly break BC. So there's a couple of options.
1. Rename `utils/_checkpointing` -> `utils/checkpointing` and remove `checkpointing` imports from `utils/__init__.py`. Import from `utils/checkpointing` directly. This removes the utils-models dependency and stops a common cause of cycles. This is the easiest approach, but will still require updating all our configs, will break BC, does not address the utils-modules dependency, nor the core problem of utils being bloated
2. Move all utils that depend on other directories to new folder: `training`. Keep utils as a directory that **does not take a dependency on any other directory**. This fundamentally restructures our dependency graph and will prevent further cycles.
[...]

Issue #1261, opened 2024-08-05
Organize our collation utils


SalmanMohammadi:
see #1005 for some context. From @ebsmothers  (and @joecummings)@

> I don't love our collate utilities rn. In an ideal world I want two simple abstractions: right_padded_collate and left_padded_collate, and any further padding built up on top of that. I don't think this version should exist at all (especially with its heavy dictionary usage)

SalmanMohammadi:
I guess the idea would be to have a general utility which applies a choice of padding function to a dict?

SalmanMohammadi:
I'm going to bump this because I'd like to right-pad outputs from a dataset which are of the form:

```
{'tokens': [920,
  508,
  10110,
  13047,
  5786,
  1371,
  12566,
  592,
  2750,
  10110,
  278,
  615,
  ...],
 'labels': 1}
```

i.e. for a classification task. Similar to `left_padded_collate` I only want to right-pad the prompt tokens.

SalmanMohammadi:
One proposal could be:

We define `left_padded_collate` + `right_padded_collate` with the signatures
```python

def left_pad_sequence(torch.Tensor) -> torch.Tensor:
    ...

def right_pad_sequence(torch.Tensor) -> torch.Tensor:
    ...
```

Then, we have task-specific pad functionality:
```python
def padded_collate_sft(
    batch: List[Dict[str, List[int]]],
    pad_fn: Callable[List[Any], torch.Tensor] = right_pad_sequence
    padding_idx: int = 0,
    ignore_idx: int = CROSS_ENTROPY_IGNORE_IDX
) -> Dict[str, torch.Tensor]:
    ...

def padded_collate_dpo(
    ...
):
    ...
```
i.e. largely the same but paramterize the pad sequence. This won't be exposed at the config level - it's just for recipe-writers and hackers.
Then, if we wanted to be really abusive, we could do something like:
```python
from torchtune.., import right_padded_collate
from operator import itemgetter
from torch.utils.data import DataLoader
dl = DataLoader(ds, batch_size=4,
                collate_fn=lambda batch: 
                (right_padded_collate(map(itemgetter("tokens"), batch)), 
                 torch.Tensor(list(map(itemgetter("labels"), batch)))))
```

for classifier recipes, or even:
```python
from torchtune.modules.rlhf.collate import left_padded_collate
from operator import itemgetter
from torch.utils.data import DataLoader
dl = DataLoader(ds, batch_size=4,
[...]

PR #1422, opened 2024-08-27
Move collate to data

RdoubleA:
#### Context
- [ ] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

This should've happened a long time ago. Collaters are fundamentally a data helper and belong in `torchtune/data`. This needs to happen regardless of the approach taken in #1421. This is also discussed in #1261. 

#### Changelog
Move collate.py and update all callsites.

#### Test plan

#### UX


tests/torchtune/data/test_collate.py
```diff
@@ -8,7 +8,7 @@
 
 import torch
 
-from torchtune.utils.collate import padded_collate
+from torchtune.data._collate import padded_collate
```
SalmanMohammadi, line 11: super nit: not sure if you need the full namespace
```suggestion
from torchtune.data import padded_collate
```

docs/source/tutorials/datasets.rst
```diff
@@ -97,7 +97,7 @@ on Hugging Face's `documentation. <https://huggingface.co/docs/datasets/en/loadi
 Setting max sequence length
 ---------------------------
 
-The default collator, :func:`~torchtune.utils.padded_collate`, used in all
+The default collator, :func:`~torchtune.padded_collate`, used in all
```
SalmanMohammadi, line 100:
```suggestion
The default collator, :func:`~torchtune.data.padded_collate`, used in all
```

SalmanMohammadi:
Looks like there's a circular import in `collate.py`?

tests/torchtune/data/test_collate.py
RdoubleA: I am totally burned by all the circular imports, so I'm going to be explicit in most cases

PR #1424, opened 2024-08-28
[RFC] Batched inference 🤝 KV-cache 🤝 compile

SalmanMohammadi:
#### Context
- [x] add a new feature
- [x] fix a bug
- [ ] other (please add here)

Closes https://github.com/pytorch/torchtune/issues/1250.

#### Changelog

Currently, we have two separate generation utils, one in `torchtune.generation`, and one in `torchtune.rlhf._generation`. 
- `torchtune.generation` is used for single-device, single-sample generation through the `generate` recipe, and supports iterative decoding with KV-cacheing, and compile. This makes it fast and efficient.
- `torchtune.rlhf._generation` is primarily used for the batched generation step in PPO, and it allows for generation with variable-length samples in a batch. This is about all  it does currently.

This PR combines the above functionality into a single unified generation module, pairing batched generation with the speedups of KV-cacheing and compile.

#### Batched inference

To correctly perform inference on batched, variable length samples, we must first pad all sequences to the same length. It doesn't make a huge amount of sense to right-pad like we usually do for SFT, since we're generating new tokens to concatenate to the prompts. Sequences are therefore left padded:
```python
prompts = [
    [1, 2, 3],
    [1, 2, 3, 4, 5],
    [1]
]
prompts_padded = left_pad_sequence(..., pad_id=0)
prompts_padded
torch.tensor(
            [[0, 0, 0, 1, 2, 3], 
             [0, 1, 2, 3, 4, 5], 
             [0, 0, 0, 0, 0, 1]]
)
```
Now, when we generate tokens, the completed prompt-generation sequences will be left-padded together. However, we don't actually want the model to attend to these padding tokens - they're invalid. How can we do this?

1) Causal masks

Attention masks are a way of ensuring the model ignores tokens we don't want it to pay attention to. A common example is for next-token prediction tasks - for each token in a sequence, attention masks ensure only preceeding tokens are used in attention calculation and not any future tokens. For SDPA, we typically use boolean masks where True indicates participation in attention, and False otherwise.

image

When we have invalid tokens in the prompt, we not only want to exclude all future tokens from the attention calculation, but also any invalid tokens in the sequence so far. If we take the first element of our unpadded-prompt earlier, we have:

```python
prompt = [1, 2, 3]
mask = [
    [True, False, False],
    [True, True, False],
    [True, True, True]
]
```

However, with padding, this becomes:

```python
prompt = [0, 0, 0, 1, 2, 3]
mask = [
    [True, False, False, False, False, False],
    [False, True, False, False, False, False],
    [False, False, True, False, False, False],
    [False, False, False, True, False, False],
    [False, False, False, True, True, False],
    [False, False, False, True, True, True],
]
```

Our original causal mask is now a block of a full causal mask for the whole sequence. For the batched case, the appropriate mask for each element is simply stacked i.e. shape `[bsz, seq_len, seq_len]`.

2) Position IDs

Position IDs are used to inform the model of the relative position in a sequence of each token. Taking my example from above:
```python
prompt = [1, 2, 3]
position_ids = [0, 1, 2]
```
tells the model that the first token is in position 0, the second in position 1, etc. Now, when we have padded inputs, these position IDs are right-shifted so that the first valid token still remains in position 0:
```python
prompt = [0, 0, 0, 1, 2, 3]
position_ids = [0, 0, 0, 0, 1, 2]
```

Make sense? The batched case is exactly the same, but, you know, batched. 

3) How does this work with KV-cacheing?

There are two relatively small changes we need to make here to support KV-cacheing. Firstly, note that when setting caches up, we create emtpy key and value tensors with sequence lengths `model.max_seq_len`.When using inference with KV-cacheing enabled, we pre-fill the first `prompt_length` positions in the cache with the key-value tensors for the prompt so far. Then, rather than recomputing the keys/values for the whole sequence when generating every new token, we only do it for one token at a time and reuse the cached values. This also means caches are updated every time we generate a new token - first the position at `prompt_length + 1`, then `prompt_length + 2`, and so on, until `prompt_length + max_generated_tokens`. 

Tying this into above:

We don't actually need to change how we we grab position IDs from the usual incremental decoding - just grab the position IDs at the current position for all batch elements!

We do, however, need to ensure we provide the correct shape of causal masks. If we work our way backwards, the full causal mask for a prompt + completion will be of shape `(bsz, prompt_length + max_generated_tokens, model_max_seq_len`). To convince yourself of this fact, consider the full mask again for the example prompt above, with `max_generated_tokens=2`, and `model_max_seq_len=9`:

```python
prompt = [0, 0, 0, 1, 2, 3]
mask = tensor(
     [[[ True, False, False, False, False, False, False, False, False],
         [False,  True, False, False, False, False, False, False, False],
         [False, False,  True, False, False, False, False, False, False],
         [False, False, False,  True, False, False, False, False, False],
         [False, False, False,  True,  True, False, False, False, False],
         [False, False, False,  True,  True,  True, False, False, False],
         [False, False, False,  True,  True,  True,  True, False, False],
         [False, False, False,  True,  True,  True,  True,  True, False]]])
``` 

It's pretty much the same mask we showed above, but now we've extended it out to an additional `max_generated_tokens` rows, and up to`model_max_seq_length` columns. It makes sense from here that during prefill step, we're just using the first `prompt_length` rows for this mask:

```python
tensor([[[ True, False, False, False, False, False, False, False, False],
         [False,  True, False, False, False, False, False, False, False],
         [False, False,  True, False, False, False, False, False, False],
         [False, False, False,  True, False, False, False, False, False],
         [False, False, False,  True,  True, False, False, False, False],
         [False, False, False,  True,  True,  True, False, False, False]]])
```

Then for every subsequent step, we're just grabbing the corresponding row. For the first generated token, this will be the 7th row of the full causal mask for the whole sequence:
```python
tensor([[False, False, False,  True,  True,  True,  True,  False, False]])
```
Great! Make sense so far? Seeing the Matrix yet? You can see how we can be kind-of efficient about this and pre-generate the whole causal mask for the entire expected sequence, and all the position IDs, and just correctly index at each step. This is what we're doing here.

4) Cache positions

One last thing. Recall that our position IDs are shifted to account for padding. Currently, we use position IDs to index and update KV-caches (this will be changed with #1449 when the cache tracks its own position internally). However, this doesn't make sense when position IDs are shifted, since they no longer line up neatly with the key/value tensors - KV cache positions should be unaffected by padding, i.e. the 0th cache position should correspond to the key/values for the 0th element in the sequence, even if it's an invalid token - otherwise we're stacking all the keys/values for invalid tokens into a single cache position. This is a nightmare for indexing and general sane reasoning. 

Concretely, our cache positions are always just `torch.arange(0, model_max_seq_len)`, and indexed accordingly for prefill and subsequent token generation steps (`[:prompt_length]`, and `[curr_pos]`, respectively). Hopefully it's clear that we don't need to use separate cache positions when there are no padding tokens in the input.

Did I say two changes earlier? I probably meant more.

#### Caveats

Currently, in order to use batched inference with KV-cacheing, we must use `model.setup_caches(batch_size=batch_size)`, which sets up caches with the model's `max_seq_len`. This is not efficient for models with long context lengths (or even any scenario when we know how many tokens we want to generate, and where `max_generated_tokens << model.max_seq_len`. This means using batched generation with KV-cacheing will consume a huge amount of memory scaling with the model's context length and the desired batch size. This will be addressed in #1449.

#### very quick results

Will provide example outputs from `generate.py` - but also look at my tests pls

quick results - more comprehensive testing to come.
(bsz=4 varying-length sequences, max_generated_tokens = 256) on a 2080 super)
```bash
on main - rlhf.generate_with_logits - no cache or compile:
Time for inference: 28.48 sec total, 35.96 tokens/sec
Bandwidth achieved: 79.52 GB/s
Memory used: 2.79 GB
---
this branch - generation.generate w/cache+compile:
Time for inference: 1.73 sec total, 590.29 tokens/sec
Bandwidth achieved: 1493.16 GB/s
Memory used: 2.68 GB
---
on main - utils.generate w/cache+compile (generates nonesense):
Time for inference: 2.68 sec total, 381.58 tokens/sec
Bandwidth achieved: 1407.27 GB/s
Memory used: 3.85 GB
```

#### Test plan

#### UX


torchtune/generation/_generation.py
joecummings, line 1: It's hard to parse the exact changes you made here b/c the entire file was moved (thanks Github).

Would you mind highlighting those changes in the description?

recipes/generate.py
```diff
@@ -161,15 +164,17 @@ def generate(self, cfg: DictConfig):
...
-        generated_tokens = utils.generate(
+        generated_tokens, _ = generate(
             model=self._model,
             prompt=prompt,
             max_generated_tokens=cfg.max_new_tokens,
+            pad_id=self._tokenizer.pad_id,
             temperature=cfg.temperature,
             top_k=cfg.top_k,
-            stop_tokens=self._tokenizer.stop_tokens,
+            # stop_tokens=self._tokenizer.stop_tokens,
```
joecummings, line 174: Why remove this?

torchtune/generation/_generation.py
```diff
@@ -0,0 +1,300 @@
...
+    mask = torch.tril(
+        torch.ones(seq_len, target_seq_len, device=padding_mask.device, dtype=bool),
+        diagonal=0,
+    ).repeat(bsz, 1, 1)
+    mask.narrow(2, 0, seq_len).mul_(padding_mask[:, None, :].expand(-1, seq_len, -1))
+    mask.diagonal(dim1=1, dim2=2).copy_(True)
+    return mask
+
+
+def get_position_ids_from_padding_masks(
```
joecummings, line 141: masks -> mask

torchtune/modules/transformer.py
```diff
@@ -43,14 +43,18 @@ def __init__(
...
         """Setup key value caches for attention calculation.
 
         Args:
             batch_size (int): batch size for the caches.
             dtype (torch.dtype): dtype for the caches.
+            max_seq_len (Optional[int]): maximum sequence length for the caches. Default None,
+                in which case ``model.max_seq_len`` is used.
         """
-        self.attn.setup_cache(batch_size, dtype)
+        self.attn.setup_cache(batch_size, dtype, max_seq_len=max_seq_len)
```
joecummings, line 57: Probably need a check to make sure that max_seq_len < self.max_seq_len

joecummings (commented):
BIG MOVES

We can leave evals for a follow-up.

torchtune/modules/transformer.py
SalmanMohammadi: no longer doing this in this PR

torchtune/modules/transformer.py
```diff
@@ -89,6 +91,10 @@ def forward(
                 of each token relative to its sample when packed, shape [b x s].
                 During inference, this indicates the position of the current token.
                 If none, assume the index of the token is its position id. Default is None.
+            cache_pos (Optional[torch.Tensor]): Optional tensor which contains the cache positions
+                of each token, used during inference. This is useful when ``input_ids`` are
+                right-shifted to account for padding tokens. Default is None, in which case
+                ``input_pos`` is used (if specified).
```
ebsmothers, line 97: I'm confused by cache_pos; is this just a stopgap until #1449? If so why not consolidate the changes across the two PRs a bit?

torchtune/modules/transformer.py
```diff
@@ -453,9 +466,24 @@ def forward(
...
+            shorter prompts are left-padded and position ids are correspondingly right-shifted,
+            thus positional ids should be of shape ``[b, padded_prompt_length]``.
+            This is because we will need to retrieve the positional embeddings for each input id.
+            In the subsequent steps, if the model has been setup with KV-caches, ``input_pos`` will contain
+            the position(s) of the current token(s) ``torch.tensor([padded_prompt_length])``. Otherwise,
+            ``input_pos`` will contain all the position ids up to the current token.
+
+            In the case above when ``input_pos`` are right-shifted due to padding, ``cache_pos``
+            should be used to correctly update KV-caches, where ``cache_pos`` is ``torch.arange(prompt_length)``
+            during the first pre-fill step, and ``torch.tensor([prompt_length])`` for subsequent steps.
```
ebsmothers, line 484: Similar comment here. Also I know that some of this is inevitable, but I notice that our transformer class is getting a bit bloated and a lot of the bloat comes from generation-related functionality. I think some of that is inevitable but still would like to keep it to a minimum as much as possible. I think something like this is useful for someone doing generation, but for those who are not it's a lot of extra stuff to sort through to get to the stuff they care about.

torchtune/generation/_generation.py
```diff
@@ -4,31 +4,29 @@
...
-
-    Args:
-        logits (torch.Tensor): logits from which to sample
-        temperature (float): value to scale the predicted logits by, default 1.0.
-        top_k (Optional[int]): If specified, we prune the sampling to only token ids within the top_k probabilities
-
-    Returns:
-        torch.Tensor: sampled token id
-    """
+    """Generic sample from a probability distribution."""
```
ebsmothers, line 29: Where'd the arg list etc go?

torchtune/generation/_generation.py
```diff
@@ -40,21 +38,54 @@ def sample(
...
+            with shape [bsz x seq_length].
+        mask (Optional[torch.Tensor]): attention mask with shape [bsz x seq_length x seq_length],
+            default None.
+        cache_pos (Optional[torch.Tensor]): tensor which contains the cache positions
+            of each token, used during inference. This is useful when ``input_ids`` are
+            right-shifted to account for padding tokens. Default is None, in which case
+            ``input_pos`` is used (if specified).
+        temperature (float): value to scale the predicted logits by, default 1.0.
+        top_k (Optional[int]): Top-k value to use for sampling, default None.
+        q (Optional[torch.Tensor]): randomly sampled tensor for softmax sampling trick.
```
ebsmothers, line 72: Maybe point to a reference here

torchtune/generation/_generation.py
```diff
@@ -160,31 +318,46 @@ def generate(
...
         else:
-            curr_input_pos = input_pos[: curr_pos + 1]
             tokens = generated_tokens.clone()
+            curr_input_pos = input_pos[:, : curr_pos + 1]
+            curr_masks = masks[:, : curr_pos + 1, : curr_pos + 1]
 
-        tokens = custom_generate_next_token(
+        q = torch.empty(
+            (bsz, model.tok_embeddings.num_embeddings), device=prompt.device
+        ).exponential_(1, generator=rng)
```
ebsmothers, line 333: I feel like we've now just given a couple equivalent default definitions of this in different places which adds to the burden of understanding the code (but lmk if I'm missing something important here)

torchtune/generation/_generation.py
```diff
@@ -69,86 +100,213 @@ def update_stop_tokens_tracker(
...
+        input_pos = torch.arange(
+            0, total_response_length, device=generated_tokens.device
+        ).unsqueeze(0)
+
+    del padding_masks
+
+    if incremental_decoding:
+        curr_masks = masks[:, :prompt_length]
+    else:
+        curr_masks = masks[:, :prompt_length, :prompt_length]
```
ebsmothers, line 270: Sorry I'm having a hard part following this. Can you add some comments here explaining what's happening?

ebsmothers:
> However, with padding, this becomes:

> prompt = [0, 0, 0, 1, 2, 3]
> mask = [
>     [True, False, False, False, False, False],
>     [False, True, False, False, False, False],
>     [True, False, True, False, False, False],
>     [True, False, False, True, False, False],
>     [True, False, False, True, True, False],
>     [True, False, False, True, True, True],
> ]

nit: mask[2:, 0] should be False 


Also one high-level question: what is the plan with respect to this and generate_v2 in #1563? Are we gonna consolidate to a single implementation, and if so when?

torchtune/modules/transformer.py
SalmanMohammadi: I generally agree here. I don't think I mess with things too much here, but in #1449 at the least, I'd like to lift a lot of the generation logic out of the transformer/layer/attention. Currently we do a bit of hand holding like creating input pos and masks if they're not there. I'd find things much cleaner and simpler if all the generation logic was front loaded to the generate fn or recipe.

torchtune/generation/_generation.py
SalmanMohammadi: The main reason we have to define it here is to make generate_next_token compile-friendly when using rng, but also allow generate_next_token to work independently from this kind-of-hacky logic i.e. you can do all this without having to sample outside of generate_next_token. 

I agree it makes things confusing and we could drop the default defn inside generate_next_token so it's a required arg.

torchtune/modules/transformer.py
SalmanMohammadi: We're landing both now so I'll update in 1449 probably.

torchtune/generation/_generation.py
```diff
@@ -4,28 +4,34 @@
...
 
 
-def multinomial_sample_one(probs: torch.Tensor) -> torch.Tensor:
+def multinomial_sample_one(
+    probs: torch.Tensor, q: Optional[torch.Tensor] = None
+) -> torch.Tensor:
     """Samples from a multinomial distribution."""
-    q = torch.empty_like(probs).exponential_(1)
+
+    q = torch.empty_like(probs).exponential_(1) if q is None else q
```
joecummings, line 18: Maybe I'm being dumb - why do we need q passed in if we define it here?

torchtune/generation/_generation.py
```diff
@@ -69,85 +109,227 @@ def update_stop_tokens_tracker(
...
+    if target_seq_len < seq_len:
+        raise AssertionError(
+            "target_seq_len cannot be shorter than the sequence length of the padding mask."
+        )
+
+    mask = torch.tril(
+        torch.ones(seq_len, target_seq_len, device=padding_mask.device, dtype=bool),
+        diagonal=0,
+    ).repeat(bsz, 1, 1)
+    mask.narrow(2, 0, seq_len).mul_(padding_mask[:, None, :].expand(-1, seq_len, -1))
```
joecummings, line 156: neat

torchtune/generation/_generation.py
```diff
@@ -69,85 +109,227 @@ def update_stop_tokens_tracker(
...
     *,
     max_generated_tokens: int,
+    pad_id: int = 0,
     temperature: float = 1.0,
     top_k: Optional[int] = None,
     stop_tokens: Optional[List[int]] = None,
+    rng: Optional[torch.Generator] = None,
     custom_generate_next_token: Optional[Callable] = None,
-) -> List[List[int]]:
+):
```
joecummings, line 196: return type?

torchtune/generation/_generation.py
```diff
@@ -69,85 +109,227 @@ def update_stop_tokens_tracker(
...
+                max_seq_len,
+                dtype=torch.bool,
+                device=prompt.device,
+            )
+        ).unsqueeze(0)
+        input_pos = torch.arange(
+            0, total_response_length, device=generated_tokens.device
+        ).unsqueeze(0)
+
+    del padding_masks
```
joecummings, line 286: Just curious - Do we know how much memory this saves?

torchtune/generation/_generation.py
joecummings: Another dumb q - why does it need to be defined every time within this loop?

joecummings (approved):
Overall looks good, but some questions mainly around the use of `q`

torchtune/generation/_generation.py
SalmanMohammadi: For use with PPO when we define it outside of `generate_next_token` so we aren't passing in the rng object which breaks compile.

SalmanMohammadi: this seems generally confusing so maybe it's better to always define it in generate and pass it through to generate_next_token.

torchtune/generation/_generation.py
SalmanMohammadi: in the best case 1 byte * bsz * model_max_seq_len, so like ~10MB with Llama3.1 and bsz=100. will remove

torchtune/generation/_generation.py
SalmanMohammadi: Similar to how you'd re-generate it every time in `sample` you have a new random distribution to sample from, right?

Issue #1396, opened 2024-08-22
Add filter_fn to text_completion_dataset


RdoubleA:
Unstructured text corpuses are often raw and have many empty lines. Adding the ability to apply a filter_fn similar to `SFTDataset` [here](https://github.com/pytorch/torchtune/blob/f9f75bb563ecae371492a9d49da4a9f514c081b3/torchtune/datasets/_sft.py#L102) would be very useful. See #1394 as an example.

AnuravModak:
Heyy, if you are looking for external contribution on this, do let me know ! Will be happy to take this up and try to solve it!

joecummings:
> Heyy, if you are looking for external contribution on this, do let me know ! Will be happy to take this up and try to solve it!

@AnuravModak - We would love to accept a PR from you on this work! Please tag @RdoubleA  and me for review.

AnuravModak:
> hey @AnuravModak, just checking in! let us know if your plans change

Heyy @RdoubleA , sorry for the delay! right now i am working on my first draft, will raise a PR today.

AnuravModak:
Hi @RdoubleA , i have raised a PR: https://github.com/pytorch/torchtune/pull/1429. Please review and provide feedback!

PR #1429, opened 2024-08-28
Added filter_fn to text_completion_dataset v1.0

AnuravModak:
Towards #1396 

I have made changes to include a filter function to remove empty lines from raw text before tokenization, similar to the implementation in `_sft.py`.

The updates involve adding two new parameters to the TextCompletionDataset class:

custom_filter: bool = False
filter_fn: Optional[Callable] = None
These additions address the need for a filter function to remove empty lines, thereby saving memory before tokenization.

Usage:

If custom_filter is set to True, users can provide their own filter_fn.
If custom_filter is False, the default filter function (default_filter) will automatically remove empty lines from raw text.
The filter_fn is applied in the _prepare_sample method before tokenization.

Addition of that flag adds a lot more flexibility because personally, I’ve often encountered situations where internal filtration was necessary, but there were also cases where I needed to adjust the raw corpus according to specific requirements.

If the additional flag custom_filter is adding unnecessary complexity and you feel its unwanted let me know i will remove it and update it in the PR with a new commit!

@RdoubleA kindly review and advise!

.

### DIFF
--- a/torchtune/datasets/_text_completion.py
+++ b/torchtune/datasets/_text_completion.py
@@ -4,7 +4,7 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
-from typing import Any, Dict, List, Mapping, Optional, Union
+from typing import Any, Dict, List, Mapping, Optional, Union, Callable
 
 from datasets import load_dataset
 from torch.utils.data import Dataset
@@ -44,6 +44,8 @@ class TextCompletionDataset(Dataset):
         column: str = "text",
         max_seq_len: Optional[int] = None,
         add_eos: bool = True,
+        custom_filter: bool=False,
+        filter_fn: Optional[Callable] = None,
         **load_dataset_kwargs: Dict[str, Any],
     ) -> None:
         self._tokenizer = tokenizer
@@ -51,6 +53,13 @@ class TextCompletionDataset(Dataset):
         self.max_seq_len = max_seq_len
         self._column = column
         self.add_eos = add_eos
+
+        # Default filter function to remove empty lines
+        def default_filter(text:str) -> str:
+            return "\n".join([line for line in text.splitlines() if line.strip()])
+        
+        # Set the filter function
+        self.filter_fn=filter_fn if custom_filter else default_filter
 
     def __len__(self):
         return len(self._data)
@@ -61,6 +70,11 @@ class TextCompletionDataset(Dataset):
 
     def _prepare_sample(self, sample: Mapping[str, Any]) -> Dict[str, List[int]]:
         prompt = sample[self._column]
+        
+        # Apply the filter function to the text data
+        if self.filter_fn:
+            prompt = self.filter_fn(prompt)
+
         tokens = self._tokenizer.encode(text=prompt, add_bos=True, add_eos=self.add_eos)
 
         # Truncate if needed, but don't coerce EOS id
@@ -81,6 +95,8 @@ def text_completion_dataset(
     add_eos: bool = True,
     packed: bool = False,
     split_across_pack: bool = True,
+    custom_filter=False,
+    filter_fn: Optional[Callable] = None,
     **load_dataset_kwargs: Dict[str, Any],
 ) -> Union[TextCompletionDataset, PackedDataset]:
     """
@@ -143,6 +159,8 @@ def text_completion_dataset(
         column=column,
         max_seq_len=max_seq_len,
         add_eos=add_eos,
+        filter_fn=filter_fn,
+        custom_filter=custom_filter,
         **load_dataset_kwargs,
     )
     return (
--- a/torchtune/datasets/_text_completion.py
+++ b/torchtune/datasets/_text_completion.py
@@ -33,6 +33,9 @@ class TextCompletionDataset(Dataset):
             Default is None, disabling truncation. We recommend setting this to the highest you can fit in memory
             and is supported by the model. For example, llama2-7B supports up to 4096 for sequence length.
         add_eos (bool): Whether to add an EOS token to the end of the sequence. Default is True.
+        filter_fn (Optional[Callable]): callable used to filter the dataset prior to any pre-processing. See
+            the Hugging Face `docs <https://huggingface.co/docs/datasets/v2.20.0/process#select-and-filter>`_ for more
+            details.
         **load_dataset_kwargs (Dict[str, Any]): additional keyword arguments to pass to ``load_dataset``,
             such as ``data_files`` or ``split``.
     """
--- a/torchtune/datasets/_text_completion.py
+++ b/torchtune/datasets/_text_completion.py
@@ -45,6 +45,7 @@ class TextCompletionDataset(Dataset):
         max_seq_len: Optional[int] = None,
         add_eos: bool = True,
         custom_filter: bool=False,
+        skip_filter: bool = False,
         filter_fn: Optional[Callable] = None,
         **load_dataset_kwargs: Dict[str, Any],
     ) -> None:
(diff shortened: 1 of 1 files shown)


AnuravModak:
Also, in the second commit i have added one more feature where if user does not want to use any filter function (be it default or custom) then simply by marking skip_filter as True it can be achieved!

Let me know which one works best!

torchtune/datasets/_text_completion.py
```diff
@@ -61,6 +74,11 @@ def __getitem__(self, index: int) -> Dict[str, List[int]]:
 
     def _prepare_sample(self, sample: Mapping[str, Any]) -> Dict[str, List[int]]:
         prompt = sample[self._column]
+
+        # Apply the filter function to the text data
+        if self.filter_fn:
```
RdoubleA, line 79: let's apply the filter_fn the same way as in `_sft.py`:
https://github.com/pytorch/torchtune/blob/f95a9f8cccb61e55faacdeb2efdb30a8dd3d33f8/torchtune/datasets/_sft.py#L110
it's consistent with how HF handles filtering and makes the contract for the filter_fn very clear (returning a bool for whether the data sample should be kept or not)

torchtune/datasets/_text_completion.py
```diff
@@ -44,6 +44,9 @@ def __init__(
         column: str = "text",
         max_seq_len: Optional[int] = None,
         add_eos: bool = True,
+        custom_filter: bool=False,
+        skip_filter: bool = False,
+        filter_fn: Optional[Callable] = None,
```
RdoubleA, line 49: while I see that this may provide more flexibility, as a new user it might be overwhelming to see three filtering parameters and understand how they work together. I would advise going with the simplest approach first of just passing in an optional filter_fn and only apply it if it is specified.

On if there should be a default filter, that's an interesting point. Personally, I would lean towards having the default be none, so we're not removing data samples without the user knowing and forces them to have the explicit intention of modifying their data. In the dataset builders themselves for specific datasets it's more reasonable to set `filter_fn` to something like what you defined because we already know that this dataset needs some preprocessing.

RdoubleA (commented):
Thanks for putting this together. Left a few comments.

AnuravModak:
Hi, your suggestions are noted!

I will update this PR in next 8-12 hrs as its midnight in India. Will update soon as per your comments.



Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________
From: Rafi Ayub ***@***.***>
Sent: Thursday, August 29, 2024 12:04:13 AM
To: pytorch/torchtune ***@***.***>
Cc: Anurav Modak ***@***.***>; Mention ***@***.***>
Subject: [External] : Re: [pytorch/torchtune] Added filter_fn to text_completion_dataset v1.0 (PR #1429)


@RdoubleA commented on this pull request.

Thanks for putting this together. Left a few comments.

________________________________

In torchtune/datasets/_text_completion.py<https://urldefense.com/v3/__https://github.com/pytorch/torchtune/pull/1429*discussion_r1735124340__;Iw!!ACWV5N9M2RV99hQ!POAvh14cz8PUqygKPBrB3Az_gltDrmRcUqp7AavfgCiLMJZgjvFZV1ZpqXMalpgdDT099Mo2b8aC6_EUoXQBVTAh_pQ$>:

> @@ -61,6 +74,11 @@ def __getitem__(self, index: int) -> Dict[str, List[int]]:

     def _prepare_sample(self, sample: Mapping[str, Any]) -> Dict[str, List[int]]:
         prompt = sample[self._column]
+
+        # Apply the filter function to the text data
+        if self.filter_fn:


let's apply the filter_fn the same way as in _sft.py:
https://github.com/pytorch/torchtune/blob/f95a9f8cccb61e55faacdeb2efdb30a8dd3d33f8/torchtune/datasets/_sft.py#L110<https://urldefense.com/v3/__https://github.com/pytorch/torchtune/blob/f95a9f8cccb61e55faacdeb2efdb30a8dd3d33f8/torchtune/datasets/_sft.py*L110__;Iw!!ACWV5N9M2RV99hQ!POAvh14cz8PUqygKPBrB3Az_gltDrmRcUqp7AavfgCiLMJZgjvFZV1ZpqXMalpgdDT099Mo2b8aC6_EUoXQBhFZln6Y$>
it's consistent with how HF handles filtering and makes the contract for the filter_fn very clear (returning a bool for whether the data sample should be kept or not)

________________________________

In torchtune/datasets/_text_completion.py<https://urldefense.com/v3/__https://github.com/pytorch/torchtune/pull/1429*discussion_r1735128678__;Iw!!ACWV5N9M2RV99hQ!POAvh14cz8PUqygKPBrB3Az_gltDrmRcUqp7AavfgCiLMJZgjvFZV1ZpqXMalpgdDT099Mo2b8aC6_EUoXQB7S2dgV8$>:

> @@ -44,6 +44,9 @@ def __init__(
         column: str = "text",
         add_eos: bool = True,
+        custom_filter: bool=False,
+        skip_filter: bool = False,
+        filter_fn: Optional[Callable] = None,


while I see that this may provide more flexibility, as a new user it might be overwhelming to see three filtering parameters and understand how they work together. I would advise going with the simplest approach first of just passing in an optional filter_fn and only apply it if it is specified.

On if there should be a default filter, that's an interesting point. Personally, I would lean towards having the default be none, so we're not removing data samples without the user knowing and forces them to have the explicit intention of modifying their data. In the dataset builders themselves for specific datasets it's more reasonable to set filter_fn to something like what you defined because we already know that this dataset needs some preprocessing.

AnuravModak:
Hi @RdoubleA , i have made the changes as per your comments and the third commit is as per the changes   `torchtune/datasets/_sft.py`. Also i have removed the code inside `_prepare_sample`  as per your requirement to keep it consistent.

RdoubleA (approved):

RdoubleA:
Looks great! Do you mind signing the CLA?

AnuravModak:
> Looks great! Do you mind signing the CLA?

So currently i have raised this PR from my github account which is also verified by Oracle (where i work basically) , will it create any conflict?

AnuravModak:
> Looks great! Do you mind signing the CLA?

Heyy, I have done it, kindly check and advise!

RdoubleA:
Looks like there's still a lint error, you'll just need to run `pre-commit run --all-files`

AnuravModak:
[like]  Anurav Modak reacted to your message:
________________________________
From: Rafi Ayub ***@***.***>
Sent: Thursday, August 29, 2024 5:05:24 PM
To: pytorch/torchtune ***@***.***>
Cc: Anurav Modak ***@***.***>; Mention ***@***.***>
Subject: [External] : Re: [pytorch/torchtune] Added filter_fn to text_completion_dataset v1.0 (PR #1429)


Looks like there's still a lint error, you'll just need to run pre-commit run --all-files

AnuravModak:
Hey @RdoubleA , I have run the precommit script on this! Let me know if anything else is required from my side!

RdoubleA:
@AnuravModak you'll need to add a docstring for the new parameter in `TextCompletionDataset` and `text_completion_dataset`. You can just use the same docstring from SFTDataset for filter_fn:
```
filter_fn (Optional[Callable]): callable used to filter the dataset prior to any pre-processing. See
      the Hugging Face `docs <https://huggingface.co/docs/datasets/v2.20.0/process#select-and-filter>`_ for more
      details.
```

Please make sure the pre-commit passes after you add it

torchtune/datasets/_text_completion.py
```diff
@@ -52,6 +56,9 @@ def __init__(
         self._column = column
         self.add_eos = add_eos
 
+        if filter_fn:
```
joecummings, line 59: Super nit at the 11th hour, I'm sorry:

```suggestion
        if filter_fn is not None:
```

RdoubleA:
Thanks for all your help @AnuravModak !

AnuravModak:
> Thanks for all your help @AnuravModak !



> Thanks for all your help @AnuravModak !

No worries! Let me know if I can help with anything else, just tag me in or assign me directly will look into it !

PR #1433, opened 2024-08-28
Cast tags parameter from ListConfig to list for CometLogger

qqlabs:
#### Context
- [ ] add a new feature
- [x] fix a bug
- [ ] other (please add here)

When we specify "tags" in configs, the list is passed into CometLogger as a ListConfig type. 

Example YAML:
`metric_logger:
  _component_: torchtune.utils.metric_logging.CometLogger
  project: ${project_name}
  experiment_name: ${xp_name}
  tags: [
    lora, 
    llama2_7B, 
    text_completion, 
    distributed,
    memory_profiling
  ]`

comet_ml.ExperimentConfig expects the tags to be a normal python list and throws the following error right now:
```
COMET ERROR: Experiment.add_tags: parameter 'tags' must be of type(s) 'list' but 'ListConfig' was given
COMET WARNING: Failed to add tag(s) ['lora', 'llama2_7B', 'text_completion', 'distributed', 'memory_profiling'] to the experiment
```

#### Changelog
- Check if "tags" is of type ListConfig, cast to list if it is before passing to comet_ml

#### Test plan

I installed a local version of torchtune with this fix and tags are now getting created for my experiments.
image

Let me know if I need to create a publicly visible comet project of the runs.

#### UX

### DIFF
--- a/torchtune/utils/metric_logging.py
+++ b/torchtune/utils/metric_logging.py
@@ -11,7 +11,7 @@ from pathlib import Path
 from typing import Any, Dict, List, Mapping, Optional, Union
 
 from numpy import ndarray
-from omegaconf import DictConfig, OmegaConf
+from omegaconf import DictConfig, ListConfig, OmegaConf
 from torch import Tensor
 
 from torchtune.utils import get_logger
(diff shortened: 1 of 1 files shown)


RdoubleA:
Hm, this is interesting. I had thought OmegaConf would instantiate dict/list parameters as their primitive types and not DictConfig/ListConfig. A more permanent fix would be to make sure that this is the case.

Approving to unblock, but this needs a closer look in our config.instantiate to make sure we get lists when we expect lists.

You may need to run `pre-commit` to fix the lint, fyi

RdoubleA (approved):

ebsmothers (changes requested):
Sorry for the last-minute request changes, but I think I have a fix at the root here: #1450. @qqlabs can you patch this and see if it works for you? (If not, or if there are other issues with my approach, I am happy to land this PR)

PR #1443, opened 2024-08-28
Streaming offloading in (q)lora single device

janeyx99:
#### Context
- [x] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

#### Changelog

Add a streaming activations offloading API to the lora finetune single device script and relevant recipes. All changes are in the newly added context manager, which streams activations to CPU and back with heuristical synchronization. The one thing that may be weird: for peak performance, we don't want to offload the last output computation of TransformerDecoder, because it does very little compute on super large memory. So offloading is actually detrimental, due to the time to offload being much longer than the compute for overlapping. So in this PR, I disable offloading by wrapping a no-op context manager around the code. This shouldn't have an impact on any existing recipe even if no saved tensor hooks were applied (as it still does AC), but I can gate the logic on only when activation_offloading is enabled.

This change depends on https://github.com/pytorch/pytorch/pull/134728 to make sure the post node hooks actually run.
For offloading to work with NF4Tensor, we also need to modify AO to give NF4 more overloads. There will be a PR coming up for that.

#### Test plan

I have run this on llama3 with seq len 8k, bs2 to observe memory savings with <1% perf hit per step. The command: 

```
tune run lora_finetune_single_device --config llama3/8B_qlora_single_device \
optimizer_in_bwd=False \
optimizer._component_=bitsandbytes.optim.PagedAdamW8bit \
compile=False \
model.lora_attn_modules="[q_proj,v_proj]" \
model.apply_lora_to_mlp=False \
model.apply_lora_to_output=False \
model.lora_rank=8 \
model.lora_alpha=16 \
dataset.source=Yukang/LongAlpaca-12k \
dataset.packed=False \
dataset.split=train[:10%] \
dataset.train_on_input=True \
tokenizer.max_seq_len=8192 \
metric_logger=torchtune.utils.metric_logging.StdoutLogger \
metric_logger.project=recipe_profiling \
log_every_n_steps=1 \
log_peak_memory_stats=True \
max_steps_per_epoch=4 \
epochs=1 \
batch_size=2 \
metric_logger.name=llama3__qlora__seqlen_8192__act_ckpt_True__act_off_True__bs2 \
profiler.enabled=True \
profiler.with_stack=True \
profiler.wait_steps=0 \
profiler.warmup_steps=2 \
profiler.active_steps=2 \
profiler.num_cycles=1
```

#### UX

With activations offloading, we use only 10.1GiB peak memory for batchsize 2 seqlen 8k compared to 13.2GiB for batchsize 1. 

image

Runtime, with offloading compared to without was only slower 20ms for a step of 23s, meaning <1% slowdown for llama3-8B. This is because almost all the comm is overlapped. 
[INTERNAL ONLY] https://interncache-all.fbcdn.net/manifold/perfetto-artifacts/tree/ui/index.html#!/?url=https://interncache-all.fbcdn.net/manifold/perfetto_internal_traces/tree/shared_trace/janeyx_731fbe41-9354-4f14-b308-83166146c376_qlora_AC_stream_AO_noop_bs2_llama3_4steps.json.gz

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -6,7 +6,7 @@ ifneq ($(EXAMPLES_PATTERN),)
 endif
 
 # You can set these variables from the command line.
-SPHINXOPTS    = -W -j auto $(EXAMPLES_PATTERN_OPTS)
+SPHINXOPTS    = -W -j auto $(EXAMPLES_PATTERN_OPTS) -T -v
 SPHINXBUILD   = sphinx-build
 SPHINXPROJ    = torchtune
 SOURCEDIR     = source
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -6,7 +6,7 @@ ifneq ($(EXAMPLES_PATTERN),)
 endif
 
 # You can set these variables from the command line.
-SPHINXOPTS    = -W -j auto $(EXAMPLES_PATTERN_OPTS) -T -v
+SPHINXOPTS    = -W -j auto $(EXAMPLES_PATTERN_OPTS)
 SPHINXBUILD   = sphinx-build
 SPHINXPROJ    = torchtune
 SOURCEDIR     = source
--- a/docs/source/tutorials/memory_optimizations.rst
+++ b/docs/source/tutorials/memory_optimizations.rst
@@ -99,16 +99,18 @@ for more details about how this is implemented through saved_tensors_hooks.
 
 This setting is especially helpful for larger batch sizes, or longer context lengths when you're memory constrained.
 However, these savings in memory can come at the cost of training speed (i.e. tokens per-second), as it takes runtime
-and resources to move Tensors from GPU to CPU and back. The implementation in torchtune uses multiple CUDA streams
-in order to overlap the extra communication with the computation to hide the extra runtime. As the communication
-workload is variable depending on the number and size of tensors being offloaded, it is common to not offload every
-single activation. In fact, once can use offloading in conjunction with activations checkpointing, where all
-activations will either be recomputed later in the backward or brought back from the CPU.
+and resources to move Tensors from GPU to CPU and back. The implementation in torchtune has the ``offload_with_streams``
+option to use multiple CUDA streams in order to overlap the extra communication with the computation to hide the extra
+runtime. As the communication workload is variable depending on the number and size of tensors being offloaded, it is
+common to not offload every single activation. In fact, once can use offloading in conjunction with activations
+checkpointing, where all activations will either be recomputed later in the backward or brought back from the CPU.
 
 *Sounds great! How do I use it?*
 
 To enable activation offloading, use the ``enable_activation_offloading`` config entry or flag
-in our lora finetuning single device recipe, e.g. ``enable_activation_offloading=True``.
+in our lora finetuning single device recipe, e.g. ``enable_activation_offloading=True``. To allow
+usage of streams, make sure you are on a torch version later than PyTorch 2.5.0.dev20240907 and
+specify ``offload_with_streams=True``.
 
 .. _glossary_grad_accm:
 
--- a/docs/source/tutorials/memory_optimizations.rst
+++ b/docs/source/tutorials/memory_optimizations.rst
@@ -83,6 +83,33 @@ and in most cases training can slow-down quite a bit as a result of this activat
 To enable activation checkpointing, use the ``enable_activation_checkpointing`` config entry or flag
 in any of our recipes, e.g. ``enable_activation_checkpointing=True``.
 
+.. _glossary_act_off:
+
+Activation Offloading
+---------------------
+
+*What's going on here?*
+
+You may have just read about activation checkpointing! Similar to checkpointing, offloading is a memory
+efficiency technique that allows saving GPU VRAM by temporarily moving activations to CPU and bringing
+them back when needed in the backward pass.
+
+See `PyTorch autograd hook tutorial <https://pytorch.org/tutorials/intermediate/autograd_saved_tensors_hooks_tutorial.html#saving-tensors-to-cpu>`_
+for more details about how this is implemented through saved_tensors_hooks.
+
+This setting is especially helpful for larger batch sizes, or longer context lengths when you're memory constrained.
+However, these savings in memory can come at the cost of training speed (i.e. tokens per-second), as it takes runtime
+and resources to move Tensors from GPU to CPU and back. The implementation in torchtune uses multiple CUDA streams
+in order to overlap the extra communication with the computation to hide the extra runtime. As the communication
+workload is variable depending on the number and size of tensors being offloaded, it is common to not offload every
+single activation. In fact, once can use offloading in conjunction with activations checkpointing, where all
+activations will either be recomputed later in the backward or brought back from the CPU.
+
+*Sounds great! How do I use it?*
+
+To enable activation offloading, use the ``enable_activation_offloading`` config entry or flag
+in our lora finetuning single device recipe, e.g. ``enable_activation_offloading=True``.
+
 .. _glossary_grad_accm:
 
 Gradient Accumulation
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -25,6 +25,7 @@ dependencies = [
     "numpy<=1.26.4", # Pin here until https://github.com/tensorflow/tensorboard/issues/6869 is addressed
     "tqdm",
     "omegaconf",
+    "psutil",
 
 ]
 dynamic = ["version"]
--- a/recipes/lora_finetune_single_device.py
+++ b/recipes/lora_finetune_single_device.py
@@ -31,8 +31,12 @@ from torchtune.modules.peft import (
     validate_missing_and_unexpected_for_lora,
 )
 from torchtune.recipe_interfaces import FTRecipeInterface
-from torchtune.training import DummyProfiler, PROFILER_KEY
-from torchtune.training._activation_offloading import NoOpManager, OffloadActivations
+from torchtune.training import (
+    DummyProfiler,
+    NoOpManager,
+    OffloadActivations,
+    PROFILER_KEY,
+)
 from tqdm import tqdm
 
 log = utils.get_logger("DEBUG")
--- a/recipes/lora_finetune_single_device.py
+++ b/recipes/lora_finetune_single_device.py
@@ -57,7 +57,7 @@ class LoRAFinetuneRecipeSingleDevice(FTRecipeInterface):
 
         - Activation Offloading. This can be controlled using the ``enable_activation_offloading``
             flag. Activation offloading is a technique similar to activations checkpointing that helps
-            reduce the memory footprint to prevent OOMs and enable bigger batches. Where activations
+            reduce the memory footprint to prevent OOMs on CUDA and enable bigger batches. Where activations
             checkpointing drops the activation in the forward to recompute it later in the backward,
             activations offloading will drop the activation in the forward to the CPU and bring it
             back during the backward pass. As always, there is a tradeoff--these savings in memory can
@@ -118,6 +118,7 @@ class LoRAFinetuneRecipeSingleDevice(FTRecipeInterface):
     Raises:
         ValueError: If ``dtype`` is set to fp16.
         RuntimeError: If ``dtype`` is set to bf16 and the hardware does not support bf16.
+        RuntimeError: If ``enable_activation_offloading`` is True and device is not CUDA.
 
     """
 
@@ -155,6 +156,13 @@ class LoRAFinetuneRecipeSingleDevice(FTRecipeInterface):
         self._save_adapter_weights_only = cfg.get("save_adapter_weights_only", False)
         self._gradient_accumulation_steps = cfg.gradient_accumulation_steps
         self._clip_grad_norm = cfg.get("clip_grad_norm", None)
+        self._enable_activation_offloading = cfg.get(
+            "enable_activation_offloading", False
+        )
+        if self._enable_activation_offloading and self._device.type != "cuda":
+            raise RuntimeError(
+                "enable_activation_offloading should only be enabled for training on CUDA"
+            )
 
     def load_checkpoint(self, cfg_checkpointer: DictConfig) -> Dict[str, Any]:
         """
@@ -239,7 +247,7 @@ class LoRAFinetuneRecipeSingleDevice(FTRecipeInterface):
         self._model = self._setup_model(
             cfg_model=cfg.model,
             enable_activation_checkpointing=cfg.enable_activation_checkpointing,
-            enable_activation_offloading=cfg.get("enable_activation_offloading", False),
+            enable_activation_offloading=self._enable_activation_offloading,
             compile_model=cfg.compile,
             base_model_state_dict=checkpoint_dict[training.MODEL_KEY],
             lora_weights_state_dict=(
--- a/recipes/lora_finetune_single_device.py
+++ b/recipes/lora_finetune_single_device.py
@@ -62,10 +62,10 @@ class LoRAFinetuneRecipeSingleDevice(FTRecipeInterface):
             activations offloading will drop the activation in the forward to the CPU and bring it
             back during the backward pass. As always, there is a tradeoff--these savings in memory can
             come at the cost of training performance and CPU resources. To recover some runtime cost,
-            specify ``offload_with_streams: True`` to enable offloading on a different stream to permit
-            overlapping with the computation. This option is currently only available on PyTorch nightly
-            version 2.5.0.dev20240907 or later. Activation offloading can be used in conjunction with
-            activation checkpointing.
+            we've added an option to enable offloading on a different stream to permit overlapping with
+            the computation. This option is currently only available on PyTorch nightly 2.5.0.dev20240907
+            or later and will be enabled by default if an acceptable torch version is found. Activation
+            offloading can be used in conjunction with activation checkpointing.
 
         - Precision. Full fp32 and bf16 training are supported. Precision is controlled using the ``dtype``
             flag. When ``dtype=bf16``, all activations, gradients and optimizer states are in bfloat16. In
@@ -240,7 +240,6 @@ class LoRAFinetuneRecipeSingleDevice(FTRecipeInterface):
             cfg_model=cfg.model,
             enable_activation_checkpointing=cfg.enable_activation_checkpointing,
             enable_activation_offloading=cfg.get("enable_activation_offloading", False),
-            offload_with_streams=cfg.get("offload_with_streams", False),
             compile_model=cfg.compile,
             base_model_state_dict=checkpoint_dict[training.MODEL_KEY],
             lora_weights_state_dict=(
@@ -387,7 +386,6 @@ class LoRAFinetuneRecipeSingleDevice(FTRecipeInterface):
         cfg_model: DictConfig,
         enable_activation_checkpointing: bool,
         enable_activation_offloading: bool,
-        offload_with_streams: bool,
         compile_model: bool,
         base_model_state_dict: Dict[str, Any],
         lora_weights_state_dict: Optional[Dict[str, Any]] = None,
@@ -443,9 +441,7 @@ class LoRAFinetuneRecipeSingleDevice(FTRecipeInterface):
 
         self.activations_handling_ctx = contextlib.nullcontext()
         if enable_activation_offloading:
-            self.activations_handling_ctx = OffloadActivations(
-                use_streams=offload_with_streams
-            )
+            self.activations_handling_ctx = OffloadActivations()
 
             # Below is our hack to disable offloading the last output Linear in every
             # step, as the cost for offloading the activation and then soon after bringing
--- a/recipes/lora_finetune_single_device.py
+++ b/recipes/lora_finetune_single_device.py
@@ -4,6 +4,7 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
+import contextlib
 import sys
 import time
 
@@ -31,7 +32,7 @@ from torchtune.modules.peft import (
 )
 from torchtune.recipe_interfaces import FTRecipeInterface
 from torchtune.training import DummyProfiler, PROFILER_KEY
-
+from torchtune.training._activation_offloading import NoOpManager, OffloadActivations
 from tqdm import tqdm
 
 log = utils.get_logger("DEBUG")
@@ -43,12 +44,21 @@ class LoRAFinetuneRecipeSingleDevice(FTRecipeInterface):
     for single GPU training. Training on CPU is not supported.
 
     Features:
-        - Activation Checkpointing. This can be controlled using the ``activation_checkpointing``
+        - Activation Checkpointing. This can be controlled using the ``enable_activation_checkpointing``
             flag. Activation checkpointing helps reduce the memory footprint since we no longer keep
             activations in memory and instead recompute them during the backward pass. This is especially
             helpful for larger batch sizes when you're memory constrained. But these savings in memory
             come at the cost of training performance. In most cases training can slow-down quite a bit as
             a result of this activation recomputation.
+
+        - Activation Offloading. This can be controlled using the ``enable_activation_offloading``
+            flag. Activation offloading is a technique similar to activations checkpointing that helps
+            reduce the memory footprint to prevent OOMs and enable bigger batches. Where activations
+            checkpointing drops the activation in the forward to recompute it later in the backward,
+            activations offloading will drop the activation in the forward to the CPU and bring it
+            back during the backward pass. As always, there is a tradeoff--these savings in memory can
+            come at the cost of training performance and CPU resources. Activation offloading
+            can be used in conjunction with activation checkpointing.
 
         - Precision. Full fp32 and bf16 training are supported. Precision is controlled using the ``dtype``
             flag. When ``dtype=bf16``, all activations, gradients and optimizer states are in bfloat16. In
@@ -222,6 +232,8 @@ class LoRAFinetuneRecipeSingleDevice(FTRecipeInterface):
         self._model = self._setup_model(
             cfg_model=cfg.model,
             enable_activation_checkpointing=cfg.enable_activation_checkpointing,
+            enable_activation_offloading=cfg.get("enable_activation_offloading", False),
+            offload_with_streams=cfg.get("offload_with_streams", False),
             compile_model=cfg.compile,
             base_model_state_dict=checkpoint_dict[training.MODEL_KEY],
             lora_weights_state_dict=(
@@ -367,6 +379,8 @@ class LoRAFinetuneRecipeSingleDevice(FTRecipeInterface):
         self,
         cfg_model: DictConfig,
         enable_activation_checkpointing: bool,
+        enable_activation_offloading: bool,
+        offload_with_streams: bool,
         compile_model: bool,
         base_model_state_dict: Dict[str, Any],
         lora_weights_state_dict: Optional[Dict[str, Any]] = None,
@@ -419,6 +433,22 @@ class LoRAFinetuneRecipeSingleDevice(FTRecipeInterface):
         training.validate_expected_param_dtype(
             self.adapter_params.items(), dtype=self._dtype
         )
+
+        self.activations_handling_ctx = contextlib.nullcontext()
+        if enable_activation_offloading:
+            self.activations_handling_ctx = OffloadActivations(
+                use_streams=offload_with_streams
+            )
+
+            # Below is our hack to disable offloading the last output Linear in every
+            # step, as the cost for offloading the activation and then soon after bringing
+            # it back is expensive. Moreover, due to heuristics in our streaming API,
+            # we actually use more memory if we offload it as it interferes with chunkedCE.
+            noop_ctx = NoOpManager()
+            model.output.register_forward_pre_hook(lambda *args: noop_ctx.__enter__())
+            model.output.register_forward_hook(
+                lambda *args: noop_ctx.__exit__(), always_call=True
+            )
 
         log.info(f"Model is initialized with precision {self._dtype}.")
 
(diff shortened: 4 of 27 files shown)


torchtune/modules/transformer.py
```diff
@@ -466,16 +466,21 @@ def forward(
...
-            # Used with CEWithChunkedOutputLoss. Need to set num_output_chunks in the recipe,
-            # before calling forward. Upcasting it done inside of the loss function.
-            output = [
-                self.output(chunk) for chunk in h.chunk(self.num_output_chunks, dim=1)
-            ]
-        else:
-            # shape: [b, seq_len, out_dim]
-            output = self.output(h).float()
+        # Disable activations offloading locally
+        from torchtune.utils._activations_offloading.offload_ctx_mgr import NoOpManager
```
felipemello1, line 470: :O 

is there anyway that we can do it outside of the transformer?

for example:
`NoOpWrapper(model.output)`

I think that out transformer class it becoming way too crazy haha

felipemello1: or in the context manager ignore a module if module in list_ignore_module

torchtune/utils/_activations_offloading/offload_ctx_mgr.py
```diff
@@ -0,0 +1,260 @@
...
+
+    Enable the memory efficiency technique of activation offloading, where activations bigger than
+    min_offload_size bytes will be offloaded to CPU in the forward and brought back in the backward.
+    This is in contrast to maintaining the activation on GPU VRAM throughout the program.
+
+    This manager uses one additional CUDA stream to handle the communication between CUDA and CPU,
+    which is intended to overlap with the default computation stream to improve runtime. We designed
+    synchronization with a few heuristics for optimizing the tradeoff between runtime vs memory usage.
+
+    max_fwd_stash_size: int = 5. The maximum size of the forward stash, or the maximum number of
```
felipemello1, line 23: can we follow the pattern of our docstrings:

```
Args:
     param (param type): description

Returns:
     param type: descriptions

Example (if it makes sense):
    >>> your_example(a=a,
    >>>  b=b)
```

torchtune/utils/_activations_offloading/offload_ctx_mgr.py
```diff
@@ -0,0 +1,260 @@
...
+                f"max_fwd_stash_size should be at least 1 but is {max_fwd_stash_size}"
+            )
+        self.max_fwd_stash_size = max_fwd_stash_size
+        self.bwd_tensor_stash = {}  # tensor_id => activation
+        self.bwd_ev_stash = {}  # tensor_id => ev0
+        self.curr_graph_id = None
+        self.curr_autograd_node = None
+
+        # platform util functions
+        def verify_sufficient_virtual_memory():
```
felipemello1, line 77: is there a motivation to define so many functions within __init__? i.e. why not do it as a method of the class or as another function outside of the class?

torchtune/utils/_activations_offloading/offload_ctx_mgr.py
```diff
@@ -0,0 +1,260 @@
...
+            # query for basic tensor info
+            num_bytes = get_num_bytes_tensor(activation)
+            tensor_id = get_tensor_id()
+
+            # only offload hefty bois
+            if num_bytes >= self.min_tensor_size_bytes:
+                # First, sync back and dereference previously offloaded tensors
+                # as the offloading should be done sufficiently long ago.
+                for k in [x for x in self.fwd_stash.keys()]:
+                    if k <= tensor_id - self.max_fwd_stash_size:
```
felipemello1, line 120: as a rule of thumb, if we can replace "x", "k" and "ev" with meaningful var names, i think it is worth to do so

recipes/lora_finetune_single_device.py
```diff
@@ -42,13 +44,21 @@ class LoRAFinetuneRecipeSingleDevice(FTRecipeInterface):
...
             come at the cost of training performance. In most cases training can slow-down quite a bit as
             a result of this activation recomputation.
 
+        - Activation Offloading. This can be controlled using the ``enable_activation_offloading``
+            flag. Activation offloading helps reduce the memory footprint since we no longer keep
+            activations in memory and instead offload them to CPU and bring them back during the
+            backward pass. This is especially helpful for larger batch sizes when you're memory
+            constrained. But these savings in memory come at the cost of training performance and
+            CPU resources. In most cases, activation offloading is most effective when used in
+            conjunction with activation checkpointing.
```
felipemello1, line 60: > "In most cases, activation offloading is most effective when used in conjunction with activation checkpointing." 

nit: can we add 1 sentence here explaining why? I think it may be worth to add to the very beginning that it can be used independently of checkpointing.

recipes/lora_finetune_single_device.py
```diff
@@ -239,7 +253,7 @@ def setup(self, cfg: DictConfig) -> None:
         if self._loss_fn.__class__.__name__ == "CEWithChunkedOutputLoss":
             # set num_output_chunks for model
             self._model.set_num_output_chunks(self._loss_fn.num_output_chunks)
-            if self._model_compile:
+            if True:  # self._model_compile:
```
felipemello1, line 256: i guess you forgot it here :P

recipes/lora_finetune_single_device.py
```diff
@@ -570,8 +584,11 @@ def _loss_step(self, batch: Dict[str, torch.Tensor]) -> torch.Tensor:
         mask = batch.get("mask", None)  # shape [b, s, s]
         input_pos = batch.get("input_pos", None)  # shape [b, s]
 
-        # run model
-        logits = self._model(tokens, mask=mask, input_pos=input_pos)
+        if self.enable_activation_offloading:
+            with OffloadActivations():
+                logits = self._model(tokens, mask=mask, input_pos=input_pos)
```
felipemello1, line 589: no strong opinion, but i wonder if this can be cleaner (remove if/else).

For example, one idea could be:
```
with OffloadActivations(is_enabled=enable_activation_offload):
        logits = self._model(tokens, mask=mask, input_pos=input_pos)
```

felipemello1 (commented):
Thanks for the PR! Very exciting results!!

I did a first pass, but I dont think that you should make changes just yet, until we are sure that streaming is the way to go. I can do some testing on my end this week, but we need to check the trade-off vs compile. If compile + no offloading > no compile + offloading, we may need to go back to the drawing board.

However, since the PR is up, maybe this would make it easier for folks from pytorch to investigate streaming support? cc: @weifengpy 

On a second note, if we do proceed with streaming, I would like to brainstorm a bit about how we can avoid touching the transformer class.

thanks again!

felipemello1:
> > If compile + no offloading > no compile + offloading, we may need to go back to the drawing board.
> 
> What does > mean here? Are we talking runtime? memory usage? both?

memory wise, i dont think that compile can be better. So probably the correct comparison should be:

"if compile + some offloading method that works with compile > no compile + stream offloading, we may need to go back to the drawing board."

What do you think? E.g. if using tensor hooks is 15% slower than stream, but compile makes it 2x faster, and saves more memory on top of that, then it may be better to use tensor hooks.

weifengpy:
> However, since the PR is up, maybe this would make it easier for folks from pytorch to investigate streaming support? cc: @weifengpy

Hey Felipe, Do you mean torch.compile + streaming? that would be another Will Feng

felipemello1:
Sorry! I meant @yf225

yf225:
I believe with per-layer compile, compile + offloading should functionally work (since we are doing offloading outside of the compiled graph). But yes performance testing is still needed.

recipes/lora_finetune_single_device.py
```diff
@@ -570,8 +588,19 @@ def _loss_step(self, batch: Dict[str, torch.Tensor]) -> torch.Tensor:
...
+        maybe_offload_ctx = contextlib.nullcontext()
+        if self.enable_activation_offloading:
+            maybe_offload_ctx = OffloadActivations(use_streams=True)
+            noop_ctx = NoOpManager()
+            self._model.output.register_forward_pre_hook(
+                lambda *args: noop_ctx.__enter__()
+            )
+            self._model.output.register_forward_hook(
+                lambda *args: noop_ctx.__exit__(), always_call=True
+            )
```
ebsmothers, line 600: Just so I understand, this is our hack to ensure that output activations do not get offloaded, right? (Nice hack though.) Is this just cause those are actually too hefty, or are there other reasons for this (e.g. interaction with our chunked output layer)?

torchtune/utils/_activations_offloading/offload_ctx_mgr.py
```diff
@@ -0,0 +1,278 @@
...
+                    node = torch._C._current_autograd_node()
+                    prev_node_ids = []
+
+                    # Init graph id if it was uninitialized
+                    if not self.curr_graph_id:
+                        self.curr_graph_id = graph_id
+
+                    # If we're on a new node, mark prev node's tensors to be freed later
+                    if (
+                        graph_id == self.curr_graph_id
```
ebsmothers, line 211: Maybe this is just cause I don't understand what `_current_graph_task_id` represents, but when would equality not hold here?

torchtune/utils/_activations_offloading/offload_ctx_mgr.py
```diff
@@ -0,0 +1,278 @@
...
+            assert (
+                unpack_tensor_id in self.tracker
+            ), f"untracked tensor with id {unpack_tensor_id}"
+
+            maybe_gpu_tensor, modified = self.tracker[unpack_tensor_id]
+            if modified:
+                if not use_streams:
+                    gpu_tensor = maybe_gpu_tensor.to(device="cuda", non_blocking=True)
+                    maybe_gpu_tensor = gpu_tensor
+                else:
```
ebsmothers, line 199: More of an organizational thing, but I wonder if it's clearer to just have separate `unpack_tensor_{streams, no_streams}` methods since the logic for the two cases is mostly disjoint anyways.

torchtune/utils/_activations_offloading/offload_ctx_mgr.py
```diff
@@ -0,0 +1,278 @@
...
+                        wait_and_del_remaining_references
+                    )
+
+                if self.is_first_forward_pass:
+                    self.is_first_forward_pass = False
+                    if self.use_pin_memory:
+                        verify_sufficient_virtual_memory()
+
+                self.is_first_backward_call = False
+                self.is_first_forward_call = True
```
ebsmothers, line 188: OK yeah now I'm more confused lol. What do each of these three `is_first_*` variables represent?

torchtune/utils/_activations_offloading/offload_ctx_mgr.py
```diff
@@ -0,0 +1,278 @@
...
+    min_offload_size: int = 1024. The minimum number of bytes a Tensor must be in order to qualify
+        for offloading. If the tensor is too small, we do not want to waste bandwidth and resources
+        moving it to CPU and back. Default: 1024 bytes.
+    """
+
+    def __init__(
+        self,
+        use_pin_memory: bool = True,
+        use_streams: bool = True,
+        max_fwd_stash_size: int = 2,
```
ebsmothers, line 46: default doesn't match what's in the docstring

torchtune/utils/_activations_offloading/offload_ctx_mgr.py
```diff
@@ -0,0 +1,278 @@
...
+                        graph_id == self.curr_graph_id
+                        and self.curr_autograd_node != node
+                    ):
+                        self.curr_autograd_node = node
+                        prev_node_ids = [id for id in self.bwd_tensor_stash.keys()]
+
+                    brought_back_from_cpu = True
+                    if unpack_tensor_id in self.fwd_stash:
+                        maybe_gpu_tensor = self.fwd_stash[unpack_tensor_id][0]
+                        brought_back_from_cpu = False
```
ebsmothers, line 220: Sorry I don't fully follow this; do we need to keep on CPU in this case cause it's one of the last `max_fwd_stash_size` tensors?

ebsmothers (commented):
This is awesome! Mostly just have a bunch of noob questions 😃

recipes/lora_finetune_single_device.py
janeyx99: The compute for them is super small compared to offloading them (cuz they're chunky), and they're used pretty much right away so we'd waste time moving them to CPU and waiting for them to come back. I'll add a comment for this

torchtune/utils/_activations_offloading/offload_ctx_mgr.py
```diff
@@ -0,0 +1,278 @@
...
+        )  # tensor_id => (new_tensor, if_modified)  ---> track what saved/offloaded tensors are where
+        self.tensor_id: int = 0
+        self.is_first_forward_call = True
+        self.is_first_backward_call = True
+        self.is_first_forward_pass = True
+
+        # managing cpu memory
+        self.use_pin_memory: bool = use_pin_memory
+        self.virtual_memory_safe_pct = (
+            0.60  # we should not exceed this percentage of memory
```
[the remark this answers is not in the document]
janeyx99: Yea, this is arbitrary + was Less's idea. I guess it would print fair warnings until someone's RAM becomes busted lol

torchtune/utils/_activations_offloading/offload_ctx_mgr.py
janeyx99: It's cuz everything gets accessed from the __init__, so everything got defined here for linearity. I don't have any real preference though. Is there a reason to move them out? If so, where would be best? (In the class but as a class method or completely out of the class.)

torchtune/utils/_activations_offloading/offload_ctx_mgr.py
```diff
@@ -0,0 +1,278 @@
...
+        def get_num_bytes_tensor(x: torch.Tensor) -> int:
+            # get the number of bytes in a tensor, for memory management purposes
+            return (
+                x.element_size() * x.nelement()
+            )  # x.element_size() * x._base_storage().nbytes()
+
+        # -------- core pack / unpack work --------
+        def pack_tensor(activation: torch.Tensor) -> int:
+            # activations are passed in during forward pass - from here we take over and return a unique id
+            if self.is_first_forward_call:
```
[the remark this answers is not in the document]
janeyx99: So a super contrived fwd could be thought of as a collection of ops, aka a collection of nodes:

```
def fwd(z):
    a = z * z
    b = sin(a)
    c = b + 1
```

In every step, you can imagine the fwd gets called once. So the first forward pass is the first step overall in training. And the first forward call is the first operation in each fwd, so here, it would be the `a = z * z`.

torchtune/utils/_activations_offloading/offload_ctx_mgr.py
janeyx99: answered in prev comment

torchtune/utils/_activations_offloading/offload_ctx_mgr.py
janeyx99: Different backwards could maybe be occurring at the same time--this makes sure that we're tracking for the right graph. I don't know many details on this and followed @albanD's advice here.

torchtune/utils/_activations_offloading/offload_ctx_mgr.py
janeyx99: Sure, I can move it out.

torchtune/utils/_activations_offloading/offload_ctx_mgr.py
janeyx99: We don't need to keep it on CPU, but we don't have preemption to know whether it's one of the last in the fwd, so we'd already have kicked off the moving to CPU. Here, we're just taking advantage of the fact that the fwd_stash must be 1 large, and we're telling the bwd to just pull from the stash if there's still something in it instead of waiting for the tensor to be pulled from CPU to GPU.

torchtune/utils/_activations_offloading/offload_ctx_mgr.py
janeyx99: Oh actually, I know why everything's in the init lol. It's cuz `pack` can't take in any variables other than the tensor --> so we can't make that a class method that takes in `self`. And technically we could move it to the global space and have `OffloadActivations` write a bunch of metadata to global, but that doesn't seem better.

janeyx99: we _could_ go and delete all the self.'s though lol

felipemello1:
I see some conflicts. Please ping me after they are solved and I will review the PR again, so we can land it.

torchtune/training/_activation_offloading.py
```diff
@@ -0,0 +1,308 @@
...
+    runtime. We designed synchronization with a few heuristics for optimizing the tradeoff between
+    runtime vs memory usage.
+
+    Args:
+        use_pin_memory (bool): Whether or not the offloaded Tensor will be placed in pinned
+            memory on the CPU. Pinned memory allows the Tensor to be moved back onto GPU more quickly
+            but is a limited resource. Default: True.
+
+        use_streams (bool): Whether or not to use streams for performance optimization where
+            the communications get overlapped with the computation. Default: True.
```
felipemello1, line 32: we should probably add some safe guards here for torch 2.4. Is that true @ebsmothers, or people cant use nightlies with 2.4?

felipemello1 (commented):
first pass. I will read more carefully later

docs/source/tutorials/memory_optimizations.rst
```diff
@@ -83,6 +83,33 @@ and in most cases training can slow-down quite a bit as a result of this activat
...
+See `PyTorch autograd hook tutorial <https://pytorch.org/tutorials/intermediate/autograd_saved_tensors_hooks_tutorial.html#saving-tensors-to-cpu>`_
+for more details about how this is implemented through saved_tensors_hooks.
+
+This setting is especially helpful for larger batch sizes, or longer context lengths when you're memory constrained.
+However, these savings in memory can come at the cost of training speed (i.e. tokens per-second), as it takes runtime
+and resources to move Tensors from GPU to CPU and back. The implementation in torchtune uses multiple CUDA streams
+in order to overlap the extra communication with the computation to hide the extra runtime. As the communication
+workload is variable depending on the number and size of tensors being offloaded, it is common to not offload every
+single activation. In fact, once can use offloading in conjunction with activations checkpointing, where all
+activations will either be recomputed later in the backward or brought back from the CPU.
```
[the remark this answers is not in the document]
janeyx99: What do you mean by "this" in the first sentence? Activations offloading works when AC is false as well, it's just super slow.

docs/source/tutorials/memory_optimizations.rst
```diff
@@ -83,6 +83,33 @@ and in most cases training can slow-down quite a bit as a result of this activat
...
+
+*Sounds great! How do I use it?*
+
+To enable activation offloading, use the ``enable_activation_offloading`` config entry or flag
+in our lora finetuning single device recipe, e.g. ``enable_activation_offloading=True``.
```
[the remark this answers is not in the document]
janeyx99: Yea I was thinking we should mention lora finetuning single device for now in particular as it's only enabled for those. It would be important to widen this when more support is added --> I'm happy to do it.

In distributed, it should be the same command, but I haven't tested at all, so did not want to promise anything for that.

recipes/lora_finetune_single_device.py
```diff
@@ -42,13 +43,22 @@ class LoRAFinetuneRecipeSingleDevice(FTRecipeInterface):
     for single GPU training. Training on CPU is not supported.
 
     Features:
-        - Activation Checkpointing. This can be controlled using the ``activation_checkpointing``
+        - Activation Checkpointing. This can be controlled using the ``enable_activation_checkpointing``
```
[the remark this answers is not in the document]
janeyx99: oh...but today the recipes take enable_activation_checkpointing and not activation_checkpointing, from what I see in the files? I was just intending to fix a typo here--I didn't think there would be BC issues. If there are BC issues I agree it would be better to not change.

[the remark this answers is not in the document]
felipemello1: sorry, my bad :P

docs/source/tutorials/memory_optimizations.rst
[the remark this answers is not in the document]
felipemello1: i remember trying to use only offloading, with AC=False, and it broke. Maybe its not the case anymore?

docs/source/tutorials/memory_optimizations.rst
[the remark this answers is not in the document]
felipemello1: >In distributed, it should be the same command

I thought that for FSDP we use fsdp_cpu_offload and let fsdp do the offloading. Unless you think it makes sense to have two different offloading strategies

recipes/lora_finetune_single_device.py
```diff
@@ -415,6 +429,22 @@ def _setup_model(
...
+            )
+
+            # Below is our hack to disable offloading the last output Linear in every
+            # step, as the cost for offloading the activation and then soon after bringing
+            # it back is expensive. Moreover, due to heuristics in our streaming API,
+            # we actually use more memory if we offload it as it interferes with chunkedCE.
+            noop_ctx = NoOpManager()
+            model.output.register_forward_pre_hook(lambda *args: noop_ctx.__enter__())
+            model.output.register_forward_hook(
+                lambda *args: noop_ctx.__exit__(), always_call=True
```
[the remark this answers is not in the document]
janeyx99: Ah yea, I agree we need to test these. The hooks do only work with module (since they're module hooks) so that would require some design...e.g., we enable adding hooks to this class you added, or we make it an nn.Module still but exclude it from FSDP/other wrappers. Or something else 😛

torchtune/training/_activation_offloading.py
janeyx99: 100% I think it'd have to be 2.5 too, since the change needed is in nightlies and not even 2.4.

ebsmothers: Oh yeah let's definitely add a version check then. What will the behavior be if someone tries to run on 2.4? Will it fully error out? And is it only a concern when `use_streams=True`?

felipemello1: my understanding is that use_streams=False works in 2.4, so i dont see why it should error. A warning plus defaulting to use_streams=False sounds reasonable to me

recipes/lora_finetune_single_device.py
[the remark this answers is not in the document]
ebsmothers: Oh yeah this tied embedding business is actually a pretty important point. As a hack we can maybe do something like `if hasattr(model, "output") and isinstance(model.output, nn.Module)` to gate the registration of these hooks? 

Would have to think more about whether we can get away with making `TiedLinear` an nn.Module though.. half the reason we're doing things this way is to avoid the breaking of references for tied weights when FSDP calls `to_empty`. Is there an easy way to exclude an nn.Module from all FSDP's usual hook registration etc?

Anyways before I get too far down that rabbit hole, I'd propose just going with the hack (provided it works). It's dumb, simple, and explicit, which I like.

docs/source/tutorials/memory_optimizations.rst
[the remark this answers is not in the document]
ebsmothers: FSDP's CPU offload will just offload parameters, gradients, and optimizer states though, right? Not activations. Do we expect that the two will work together? I can't immediately think of any reason why they wouldn't but maybe there's something obvious I'm missing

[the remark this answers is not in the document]
felipemello1: oh, i missed that FSDP doesnt offload activations. Thats nice! I thought it was a solved issue for FSDP and we were just improving single device

docs/Makefile
```diff
@@ -6,7 +6,7 @@ ifneq ($(EXAMPLES_PATTERN),)
 endif
 
 # You can set these variables from the command line.
-SPHINXOPTS    = -W -j auto $(EXAMPLES_PATTERN_OPTS)
+SPHINXOPTS    = -W -j auto $(EXAMPLES_PATTERN_OPTS) -T -v
```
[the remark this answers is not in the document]
janeyx99: something to allow verbose to debug the docs, i'm moving this to a diff PR to land, but wanted to test what errors we get as I no longer get local errors

torchtune/training/_activation_offloading.py
janeyx99: Yea, use_streams=False will work with 2.4. It's only with use_streams=True where we should error. Is there already another feature gated by torch version that I should copy?

And yes, I will default to False for use_streams for this PR

docs/source/tutorials/memory_optimizations.rst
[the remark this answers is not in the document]
janeyx99: They theoretically should work together but who knows man, we can't say so til we test! I'm suspicious that there will be unforeseen results with intranode comms, so the scope of this PR is just single device.

docs/source/tutorials/memory_optimizations.rst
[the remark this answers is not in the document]
janeyx99: Yea it shouldn't break! It should just be reaaaaally slow

pyproject.toml
```diff
@@ -25,6 +25,7 @@ dependencies = [
     "numpy<=1.26.4", # Pin here until https://github.com/tensorflow/tensorboard/issues/6869 is addressed
     "tqdm",
     "omegaconf",
+    "psutil",
```
janeyx99, line 28: @ebsmothers is this okay? This is a new requirement as we use psutil to check cpu RAM usage and warn on too much usage.

janeyx99: FYI: psutil does not pull in other deps!

torchtune/utils/_activations_offloading/offload_ctx_mgr.py
janeyx99: done!

recipes/lora_finetune_single_device.py
janeyx99: done

torchtune/utils/_activations_offloading/offload_ctx_mgr.py
```diff
@@ -0,0 +1,278 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
```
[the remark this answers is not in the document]
janeyx99: done :)

torchtune/training/_activation_offloading.py
```diff
@@ -81,6 +81,10 @@ def __init__(
 
         # for streaming
         if use_streams:
+            if torch.__version__ < "2.5.0.dev20240907":
+                raise RuntimeError(
+                    "OffloadActivations with use_streams=True requires PyTorch 2.5.0.dev20240907 or later."
```
felipemello1, line 86: How do you guys feel about use_treams=True being the default, and if version<2.5, we set it to false and raise a warning (not error)?

janeyx99: I think this depends on when this feature will be released in torchtune--if that release will depend on 2.5, then I think defaulting to True makes sense. Otherwise, I think it would not be great to auto-fallback to a slower method when the user may be expecting something else.

Another alternative: we have use_streams default to None, and when unspecified, we check which torch version the running script is using, and we choose True or False depending on that.

felipemello1: i think that our stable now will use 2.5

recipes/lora_finetune_single_device.py
```diff
@@ -222,6 +239,8 @@ def setup(self, cfg: DictConfig) -> None:
         self._model = self._setup_model(
             cfg_model=cfg.model,
             enable_activation_checkpointing=cfg.enable_activation_checkpointing,
+            enable_activation_offloading=cfg.get("enable_activation_offloading", False),
+            offload_with_streams=cfg.get("offload_with_streams", False),
```
felipemello1, line 243: I think that this is the last thing to approve the PR: I dont like having two arguments for offloading. What do you think about making enable_activation_offloading have 3 options?

False / "with_stream" / "without_stream"

or something like that?

ebsmothers: One q here: is there any case that I wouldn't wanna use streams provided I'm on a sufficiently recent PyTorch version? If there are no feature gaps I'm inclined to just have a single `enable_activation_offloading` flag that will run with streams if possible and otherwise run the vanilla version.

torchtune/training/_activation_offloading.py
ebsmothers: > Another alternative: we have use_streams default to None, and when unspecified, we check which torch version the running script is using, and we choose True or False depending on that.

Lol did not see this before making my other comment, but yeah I agree with this.

ebsmothers (approved):
I think the only remaining thing here is sorting out versioning/how we expose streaming vs. non-streaming. But that's a minor point (and seems like we're in agreement on it anyways). Other than that this looks great. Can't wait to have this available for people to try out!

recipes/lora_finetune_single_device.py
janeyx99: We haven't tested streams use on other hardware or with distributed or with other shape models, so I was hesitant to make it the default. Since we're only landing the lora finetuning recipe on single device with this PR, it is fine to just not include that flag option for now.

I've removed it but it would be p easy to add it back if required in the future.

torchtune/training/_activation_offloading.py
```diff
@@ -0,0 +1,331 @@
...
+                            self.s0.wait_event(ev)
+                            del self.fwd_stash[id]
+                        else:
+                            break
+
+                    # Sync in, offload, and add an event to sync back later
+                    self.s1.wait_stream(self.s0)
+
+                stream = self.s1 if self.use_streams else self.s0
+                with torch.cuda.stream(stream):
```
janeyx99, line 166: the following code is also new @ebsmothers @felipemello1 

We add a similar version check for ao. This one is less "essential" than the torch check because someone will only run into this if they play around with AC and start offloading NF4Tensors. I've landed the torchao counterpart here https://github.com/pytorch/ao/pull/881.

PR #1445, opened 2024-08-29
Reduce compile time for single-device and multi-device recipes

yf225:
#### Context
- [ ] add a new feature
- [ ] fix a bug
- [x] other (improve compile time)

Improvements in compile time (on my A100 machine):
Llama3 8B QLoRA single-device: 83s -> 47s (**-43%**)
Llama3 8B full-finetune single-device: 29s -> 20s (**-31%**)
Llama3 70B QLoRA FSDP2: 104s -> 62s (**-40%**)

#### Changelog
- No nested FSDP wrapping for LoRALinear, to reduce some recompiles.
- mark_dynamic for each layer's input `h`, to avoid dynamic sequence length induced recompiles.
- mark_dynamic for loss_fn inputs (`logits` and `labels`), to avoid dynamic sequence length induced recompiles.

#### Test plan

TODO (mentioned by Evan):
- we have CI for our recipes to make sure loss values don't change, we could add something like this to the recipe test: https://fburl.com/pusfwz1n, since we don't have it for our FSDP2 recipe yet
- How about just a stripped-down version of my test from https://github.com/pytorch/torchtune/pull/1419? Basically run for 1-2 models and show that memory and perf look good; I guess you already have compile time covered. Just paste in the memory and toks/sec into the PR desc.

#### UX

### DIFF
--- a/recipes/lora_finetune_single_device.py
+++ b/recipes/lora_finetune_single_device.py
@@ -554,6 +554,8 @@ class LoRAFinetuneRecipeSingleDevice(FTRecipeInterface):
         labels = labels[..., 1:].contiguous()
         logits = logits.transpose(1, 2)
         # Compute loss
+        torch._dynamo.mark_dynamic(logits, 2)
+        torch._dynamo.mark_dynamic(labels, 1)
         loss = self._loss_fn(logits, labels)
         # free logits otherwise it peaks backward memory
         del logits
--- a/torchtune/modules/transformer.py
+++ b/torchtune/modules/transformer.py
@@ -449,6 +449,7 @@ class TransformerDecoder(nn.Module):
             if i in self.output_hidden_states:
                 hidden.append(h)
             # shape: [b, s, d]
+            torch._dynamo.mark_dynamic(h, 1)
             h = layer(
                 h,
                 mask=mask,
--- a/recipes/configs/dev/llama2/70B_qlora_fsdp2.yaml
+++ b/recipes/configs/dev/llama2/70B_qlora_fsdp2.yaml
@@ -15,46 +15,64 @@
 
 # Model Arguments
 model:
-  _component_: torchtune.models.llama2.qlora_llama2_70b
+  _component_: torchtune.models.llama3.qlora_llama3_70b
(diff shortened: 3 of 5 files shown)


torchtune/modules/transformer.py
```diff
@@ -449,6 +449,7 @@ def forward(
             if i in self.output_hidden_states:
                 hidden.append(h)
             # shape: [b, s, d]
+            torch._dynamo.mark_dynamic(h, 1)
```
ebsmothers, line 452: Noob question on where we apply `mark_dynamic`: is it here because this is the entry point for per-layer compile? (And separately would there be any benefit to compiling the token embeddings? I guess not?)

And similarly on logits and labels: because they are the entry point to the compiled loss?

yf225: > is it here because this is the entry point for per-layer compile?

> And similarly on logits and labels: because they are the entry point to the compiled loss?

yes to both :)

> separately would there be any benefit to compiling the token embeddings? I guess not?

was thinking about it as well, and I just quickly tested it - seems no perf benefit from compiling the token embeddings :(

recipes/lora_finetune_single_device.py
```diff
@@ -554,6 +554,8 @@ def _loss_step(self, batch: Dict[str, torch.Tensor]) -> torch.Tensor:
         labels = labels[..., 1:].contiguous()
         logits = logits.transpose(1, 2)
         # Compute loss
+        torch._dynamo.mark_dynamic(logits, 2)
+        torch._dynamo.mark_dynamic(labels, 1)
```
felipemello1, line 558: does this matter if we are not compiling loss_step anymore?

felipemello1:
@yf225 , i have just landed the chunked CE. In that PR, i also changed the shift logic:

```
# Shift so that tokens < n predict n
labels = labels[..., 1:].contiguous()
```

and added chunking to the transformer + CE loss. I would be glad to help with testing the dynamic flags with you, if you need help.

recipes/lora_finetune_single_device.py
yf225: yes this influence how `self._loss_fn` is compiled (currently for per-layer compile, we also compile the loss_fn)

PR #1449, opened 2024-08-29
[RFC] Adding overrides for max cache seq length

SalmanMohammadi:
#### Context
- [x] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

#1364 

#### Changelog

This PR:
- Adds support for overriding the maximum sequence length used when setting up KV-caches. 
- Adds support for correctly setting up caches for self-attention, cross-attention, and fusion layers by exposing `encoder` and `decoder` `max_seq_len` args. These arguments are exposed to the top-level transformer class (i.e. `TransformerDecoder,` `DeepFusionModel`, so that the API for all models remains the same `model.setup_caches(bsz, dtype, encoder_max_seq_len, decoder_max_seq_len)`
- We also remove the use of `input_pos` to update and retrieve from the KV-cache. Instead, the KV cache tracks its own position.

image

#### Test plan

#### UX


SalmanMohammadi:
What's up with the eleuther eval tests? They're passing locally for me.

torchtune/modules/model_fusion/_fusion.py
```diff
@@ -293,27 +314,46 @@ class DeepFusionModel(nn.Module):
...
     Args:
-        decoder (TransformerDecoder): decoder module
+        decoder (torchtune.modules.TransformerDecoder): decoder module
         encoder (nn.Module): encoder module
     """
 
     def __init__(
         self,
-        decoder: TransformerDecoder,
+        decoder: torchtune.modules.TransformerDecoder,
```
SalmanMohammadi, line 323: had to do this due to circular imports, open to something more sane looking

torchtune/modules/attention.py
```diff
@@ -282,7 +281,7 @@ def forward(
 
             # Update key-value cache
             if self.kv_cache is not None:
```
SalmanMohammadi, line 283: What would be a sensible way to only use the cache for generation? For PPO, I need to run two forward passes with the same model, all under `no_grad`
1) During generation, with cache enabled
2) During trajectory estimation, using the same model with LoRA adapters disabled, but **without KV cacheing**

Then, we obviously don't want to use KV cacheing when `requires_grad=True` during the loss calculations.
My current solution is to wrap all generation code into `torch.inference_mode()`. Then, we check here for `if x.is_inference()` to only use the cache during inference, and not in any other case. 

Currently this is broken for compile (see issue here https://github.com/pytorch/pytorch/issues/135439#issuecomment-2337454492), but hopefully a fix by 2.5. Like most other places in the repo, could we version check for compile + generation?

@pbontrager @joecummings

SalmanMohammadi: On reflection this isn't important for this PR and we can address in a follow up when it is needed for PPO.

torchtune/modules/model_fusion/_fusion.py
```diff
@@ -3,12 +3,13 @@
 #
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
+from __future__ import annotations
```
ebsmothers, line 6: Is this also cause of circular imports?

torchtune/modules/model_fusion/_fusion.py
ebsmothers: I'm confused, what's the circular import here? Seems to me like TransformerDecoder should not be depending on fusion, right?

torchtune/modules/transformer.py
```diff
@@ -554,47 +623,86 @@ def __init__(
...
         Args:
             batch_size (int): batch size for the caches.
             dtype (torch.dtype): dtype for the caches.
+            encoder_max_seq_len (Optional[int]): maximum encoder cache sequence length.
+            decoder_max_seq_len (Optional[int]): maximum decoder cache sequence length.
         """
+        has_encoder_layers = any(
+            isinstance(l, TransformerCrossAttentionLayer) or isinstance(l, FusionLayer)
+            for l in self.layers
+        )
```
ebsmothers, line 660: So it might be too much of a sledgehammer to do it this way, but in my mind it's also more explicit about the condition we actually care about.
```suggestion
        has_encoder_layers = any(
            isinstance(m, TransformerCrossAttentionLayer) for m in self.modules()
        )
```

torchtune/modules/transformer.py
```diff
@@ -657,18 +765,26 @@ def forward(
...
-            # in most cases input_pos_len should be 1
-            mask = self.causal_mask[None, input_pos]
+        if (
+            self.encoder_caches_are_enabled
+            or self.decoder_caches_are_enabled
+            and input_pos is None
+        ):
+            raise ValueError(
+                "KV-caches are setup for inference mode, input positions must be provided!"
+            )
```
ebsmothers, line 787: At this point I would offload all of this (plus the seq len validation just above) into `self._validate_inputs` or something. Either way should do this check prior to calling token embeddings though

torchtune/utils/_generation.py
```diff
@@ -114,13 +118,50 @@ def generate(
...
+    # if key value caches are enabled, we can incrementally decode
+    incremental_decoding = (
+        model.encoder_caches_are_enabled or model.decoder_caches_are_enabled
+    )
+
+    # setup encoder+/decoder masks for caches
+    mask, encoder_mask = None, None
+    curr_mask, curr_encoder_mask = None, None
+    if model.encoder_caches_are_enabled:
+        # generation for bsz=1 isn't supported right now anyway, broadcasting is fine here
```
ebsmothers, line 130: bsz > 1?

torchtune/utils/_generation.py
```diff
@@ -114,13 +118,50 @@ def generate(
...
     input_pos = torch.arange(0, model.max_seq_len, device=prompt.device)
     tokens = generate_next_token(
         model,
+        mask=mask[:, :prompt_length] if mask is not None else None,
+        encoder_mask=(
+            encoder_mask[:, :prompt_length]
+            if encoder_mask is not None
+            else None
+            if encoder_mask is not None
+            else None
```
ebsmothers, line 163: 😕

ebsmothers (commented):
Overall it looks reasonable to me, will leave it to @pbontrager or @joecummings for the final sign-off

SalmanMohammadi:
Thanks so much for the reviews @ebsmothers. Will address once #1424 lands and things are merged.

joecummings (approved):
No concerns, but pls address @ebsmothers comments.

PR #1450, opened 2024-08-29
Cast DictConfig -> dict in instantiate

### LINKED ISSUE #1444: See if config.instantiate should always instantiate dict/list as primitive types
In some cases arguments passed to `_component_` get instantiated as `ListConfig` or `DictConfig` when really we want the primitive `list` or `dict`. Investigate why this is happening, and if it should always default to instantiating the primitive type.

More context: https://github.com/pytorch/torchtune/pull/1433#issuecomment-2316404063

ebsmothers:
Fixes #1444 

Tested via 
```
tune run lora_finetune_single_device --config llama2/7B_lora_single_device gradient_accumulation_steps=1
...
1|855|Loss: 0.6323381066322327:   3%|████████████▊                                                                                                                                                                                                                                                                                              | 855/25880 [04:01<2:03:21,  3.38it/s]
```

Logged `type(lora_modules)` [here](https://github.com/pytorch/torchtune/blob/4fba6cd44c8f24fe4ea3a2cf3a9c03e12fd9f847/torchtune/models/llama2/_component_builders.py#L85). 

On main: `<class 'omegaconf.listconfig.ListConfig'>`
On this PR: `<class 'list'>`

### DIFF
--- a/torchtune/config/_instantiate.py
+++ b/torchtune/config/_instantiate.py
@@ -20,7 +20,7 @@ def _create_component(
     return _component_(*args, **kwargs)
 
 
-def _instantiate_node(node: DictConfig, *args: Tuple[Any, ...]):
+def _instantiate_node(node: Dict[str, Any], *args: Tuple[Any, ...]):
     """
     Creates the object specified in _component_ field with provided positional args
     and kwargs already merged. Raises an InstantiationError if _component_ is not specified.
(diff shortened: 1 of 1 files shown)


torchtune/config/_instantiate.py
```diff
@@ -103,4 +103,4 @@ def instantiate(
     # Resolve all interpolations, or references to other fields within the same config
     OmegaConf.resolve(config)
 
-    return _instantiate_node(config, *args)
+    return _instantiate_node(OmegaConf.to_object(config), *args)
```
RdoubleA, line 106: we should update the typing of `_instantiate_node` since now you're passing in a dict instead of DictConfig

RdoubleA (approved):
good thing we have the config instantiation tests for all our configs, otherwise I would be very nervous

Issue #1407, opened 2024-08-25
[Feature request] Add grad norm monitoring/logging


gau-nernst:
Personally I have found that monitoring grad norm is useful to understand stability of training. It is also useful to set an appropriate clipping value (though I don't think torchtune supports grad norm clipping atm?).

Some considerations:
- Calculate grad norm may slow down training -> mitigation: only calculate grad norm at logging step
- If there is grad norm clipping, we have grad norm for free, don't need to recompute (https://pytorch.org/docs/stable/generated/torch.nn.utils.clip_grad_norm_.html)
- How to calculate grad norm in FSDP? (we can limit this feature to single device first)

ebsmothers:
@gau-nernst thanks for this suggestion. Actually we had a similar discussion a few months back in #897, maybe I was too much of a stickler about it at that time 😅. In addition to the comments I left there, I agree with your point on not slowing down training. I think your proposal to only calculate at logging step is reasonable, but in practice I think many of our configs set `log_every_n_steps=1` so in that case it doesn't do much for us anyways.

One alternative to using `nn.utils.clip_grad_norm` is to provide a bit more flexibility in the metrics we log in general. There is a world where we allow passing a list of callables that return any custom metrics that a user may want to compute and update the log dict accordingly somewhere around [here](https://github.com/pytorch/torchtune/blob/9f37c73453042c5b07f81c4f82e882365d594669/recipes/lora_finetune_single_device.py#L625). The main question is how we actually design this, but I can imagine the signature of each callable being optional loss, optional model, and optional optimizer, which should be reasonably flexible (I don't wanna get into per-batch vs per-step vs per-epoch logging though cause that's a whole other can of worms). The downside of this is that (a) we don't actually clip the grad norm (though technically I guess the user can do it themselves by calling that API) and (b) it might be a bit overengineered.

Otherwise there is the `nn.utils.clip_grad_norm` route. My main question here is around how we enable optional grad norm logging (so we don't have to slow down training) and/or clipping nicely from the config. Maybe something like `clip_grad_norm: Optional[float]` and `log_grad_norm: bool`?

Personally I am open to either of these approaches, would be interested to hear your thoughts on the pros and cons here as well.

Agree that doing this properly for FSDP will need a bit more thought (I assume we would want the norm across all ranks and not per-rank? Also I believe if we use ). But fine to punt it for now.

gau-nernst:
[...]

PR #1451, opened 2024-08-29
Add grad norm logging single device

lindawangg:
#### Context
- [x] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

- Addresses #1407 

#### Changelog
- Adds optional clip grad norm
- Default logs grad norm is `clip_grad_norm` is set

#### Test plan

```
tune run lora_finetune_single_device --config llama2/7B_lora_single_device max_steps_per_epoch=100 clip_grad_norm=100 metric_logger._component_=torchtune.utils.metric_logging.TensorBoardLogger checkpointer.output_dir=/tmp/llama2_w_grad_norm
```
Did not observe any changes to overall compute time or loss.
* orange is with `log_grad_norm=True`, blue is baseline
image
image
image
image

```
tune run full_finetune_single_device --config llama2/7B_full_low_memory metric_logger._component_=torchtune.utils.metric_logging.TensorBoardLogger checkpointer.output_dir=/tmp/llama2_w_grad_norm_full optimizer_in_bwd=False clip_grad_norm=100
```
Observed no increase in batch/s. Verified clip_grad_norm clips gradients (in orange)
* green is baseline, purple is clip_grad_norm='inf', orange is clip_grad_norm=100
image
image
image

#### UX


recipes/lora_finetune_single_device.py
```diff
@@ -132,6 +136,10 @@ def __init__(self, cfg: DictConfig) -> None:
         self._resume_from_checkpoint = cfg.resume_from_checkpoint
         self._save_adapter_weights_only = cfg.get("save_adapter_weights_only", False)
         self._gradient_accumulation_steps = cfg.gradient_accumulation_steps
+        self._clip_grad_norm = cfg.get("clip_grad_norm", None)
+        self._log_grad_norm = cfg.get("log_grad_norm", False)
```
lindawangg, line 140: We don't really need `log_grad_norm`. If `clip_grad_norm` is set, we can log grad norm by default. Computing this each step doesn't increase the time for single device.

ebsmothers: Yeah based on your results I agree, seems like no negative perf impact to just logging by default. Can you also make the same changes in full_finetune_single_device.py and confirm that no negative perf impact there? (I think we should enable for both recipes in one go.)

Distributed may be a different story if we have to sync, but we can consider that out of scope for this PR.

If the results look similar for full finetune, I agree it makes sense to go ahead and remove the `log_grad_norm` config. I would even bias towards just always logging grad norm (that way you don't have to awkwardly set it to `'inf'` to turn on logging, imo that's a bit unintuitive).

ebsmothers (commented):
This looks good! Apart from my comment around enabling for full finetune, we can also run a version with `clip_grad_norm` set to some finite value to ensure that we see it reflected correctly in the logs.

ebsmothers (approved):
This looks great, thank you!

PR #1453, opened 2024-08-29
[EZ] Use tokenizer max_seq_len in text_completion

RdoubleA:
#### Context
- [ ] add a new feature
- [x] fix a bug
- [ ] other (please add here)

As titled. We use max_seq_len from the tokenizer in all our datasets instead of an explicit parameter

#### Test plan
Tested with `ds = text_completion_dataset(tokenizer=tokenizer, source="allenai/c4", column="text", data_dir="realnewslike", packed=False, split="train")`

#### UX


felipemello1 (approved):

SalmanMohammadi:
Think you gotta move `max_seq_len=null` from dataset to tokenizer in the Mistral 7B PPO config.

PR #1460, opened 2024-08-29
Move seed from utils to training

RdoubleA:
#### Context
- [ ] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

Addresses #1441.

#### Changelog

#### Test plan

#### UX


docs/source/api_ref_training.rst
```diff
@@ -38,6 +38,16 @@ Utilities for working in a reduced precision setting.
...
 
+
+Miscellaneous
+-------------
+
+.. autosummary::
+    :toctree: generated/
+    :nosignatures:
+
+    set_seed
```
SalmanMohammadi, line 49: Why does this need to be a public API?

SalmanMohammadi (approved):

joecummings (changes requested):
This may be one of those rare cases in which it belongs actually in a common utils, rather than training.

My reasoning here is that we need this set_seed for generation and eval, as well, which are not training. Therefore, it would be confusing to import from trainiing.

RdoubleA:
> My reasoning here is that we need this set_seed for generation and eval, as well, which are not training. Therefore, it would be confusing to import from trainiing.

A few of the utils APIs fall under this category, like set_device, set_dtype, etc. But I don't have strong opinions either way, if we leave set_seed in utils then I can close this PR.

joecummings:
> > My reasoning here is that we need this set_seed for generation and eval, as well, which are not training. Therefore, it would be confusing to import from trainiing.
> 
> A few of the utils APIs fall under this category, like set_device, set_dtype, etc. But I don't have strong opinions either way, if we leave set_seed in utils then I can close this PR.

Yeah I'd prefer to leave those in utils or rename to common or something. not sure.

SalmanMohammadi:
There's `torchune.modules.common_utils.py`

joecummings:
> There's `torchune.modules.common_utils.py`

Nah, I don't think these should live nested in modules/. Let's keep it where it is for now and we can discuss if we need a more descriptive name than utils.

joecummings (approved):
not happy about it, but unblocking for now :/

PR #1461, opened 2024-08-29
Move distributed from utils to training

RdoubleA:
#### Context
- [ ] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

Addresses #1442.

#### Changelog

#### Test plan

#### UX


ebsmothers (approved):
Once CI is green this looks good

torchtune/training/_distributed.py
```diff
@@ -280,7 +280,7 @@ def lora_wrap_fsdp(module: nn.Module, recurse: bool, **kwargs):
 
 
 def load_from_full_model_state_dict(
-    model: "FSDPModule",
+    model: "FSDPModule",  # noqa
```
felipemello1, line 283: this seems a bit off, since we only import FSDPModule a few lines later. So it means that it may be something that is not an FSDPModule. Is that right?

Not related to this PR, but maybe the qa shouldnt be supressed.

felipemello1 (approved):
tests are running still. Approving to unblock. Left just one comment.

torchtune/utils/__init__.py
```diff
@@ -5,23 +5,6 @@
 # LICENSE file in the root directory of this source tree.
 
 from ._device import get_device
-from ._distributed import (  # noqa
-    contains_fsdp,
-    FSDPPolicyType,
```
felipemello1, line 10: some of the removed imports are still in __all__

felipemello1 (changes requested):

torchtune/utils/__init__.py
RdoubleA: ugh, too many merge conflicts

felipemello1 (approved):

felipemello1:
approving to unblock

torchtune/training/_distributed.py
RdoubleA: Probably because the import line is not at the top so the linter doesn't like this. I'll leave the noqa for now and we can address in a follow up

Issue #1249, opened 2024-08-01
Does torchtune support for traditional text-generation tasks?


sherlcok314159:
I wonder whether torchtune can support traditional tasks such as translation or more general text generation tasks which have a input and output column. I have read the datasets doc at [here](https://pytorch.org/torchtune/stable/api_ref_datasets.html). But it looks like we all need to format it by InstructTemplate. How can I do traditional text-generation tasks by Llama3.1-8B? Any help is welcome. I am new to torchtune.

SalmanMohammadi:
Hey @sherlcok314159. Thanks for your interest in torchtune and for raising this!

It sounds like you're interested in a general next-token-prediction task? Could you possibly provide a link to a dataset you're interested in fine-tuning on?

It sounds like the [text completion dataset](https://docs-preview.pytorch.org/pytorch/torchtune/1147/tutorials/datasets.html#custom-unstructured-text-corpus) could be suitable for your needs, which enables the use of unstructured datasets (and, optionally, the use of sample packing!). 

Let me know if that's pointed you in the right direction - very happy to help more.

sherlcok314159:
Thanks for your quick response. For example, I want to let LLMs to predict the sentiment of movie reviews. https://huggingface.co/datasets/stanfordnlp/sst2. Generally speaking, we can have one or two inputs, and a label which can be a class (classification) or a sequence (generation).

SalmanMohammadi:
Cool!!

So, for classification tasks we support classification models out of the box (e.g. [here's a llama2 model for [reward modelling](https://github.com/pytorch/torchtune/blob/288ff4435b0cf17325b5c3b112f6859a6cdf0ea2/torchtune/models/llama2/_model_builders.py#L279) which can be generalised through the `num_classes` arg). 

You'd then need to add functionality for a classification dataset. @ebsmothers' has far more intelligent things to say than I do: answer [here](https://github.com/pytorch/torchtune/issues/1124#issuecomment-2197037012). To quote:

> I assume you'd need to feed in labels from your dataset in that case (instead of just using shifted tokens as we currently do). 
[...]

PR #1463, opened 2024-08-30
Re-organizing collation utils

### LINKED ISSUE #1261: Organize our collation utils

SalmanMohammadi:
#### Context
- [x] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

closes #1261, ref #1249

#### Changelog
This PR:
- Renames our current collation utility which is used in our SFT recipes to `padded_collate_sft`, to indicate that it is task-specific.
- Changes our current left-padding collation utilities to a generic `left_pad_sequence_function`. 
- Introduces a generic `padded_collate`, which can be flexibly configured with a choice of base padding function to suit tasks not covered by `padded_collate_sft`. This is demonstrated through the PPO recipe currently, but the tests for the function also show how it could be used for e.g. classification tasks.

#### Test plan

#### UX


torchtune/data/_collate.py
```diff
@@ -12,7 +12,133 @@
...
+        batch_first=batch_first,
+        padding_value=padding_value,
+    ).flip(dims=[1])
+
+
 def padded_collate(
+    batch: List[Dict[str, List[int]]],
+    *,
+    pad_fn: Callable,
+    keys_to_pad: List[str],
```
SalmanMohammadi, line 60: I'm not 100% about how this should be paramterized and very open to suggestions. The current setup means all `keys_to_pad` will be collated with `pad_fn`, even if they don't need padding, which shouldn't have any effect when the key doesn't need padding e.g. in the case of classification tasks, but it's not super intuitive.

SalmanMohammadi: After looking back on this, it looks silly. I've simplified.

recipes/eleuther_eval.py
```diff
@@ -112,15 +111,13 @@ def tok_batch_encode(
         tokenized_text = [self.tok_encode(x) for x in text]
 
         # pad left
-        x = pad_sequence(
+        x = left_pad_sequence(
             [
                 torch.tensor(x[::-1]) for x in tokenized_text
```
SalmanMohammadi, line 116: whoops gotta fix

torchtune/data/_collate.py
```diff
@@ -12,7 +12,136 @@
...
+                              [-10,   4,   5,   6,   7],
+                              [  8,   9,  10,  11,  12]])
+        }
+    """
+    if not isinstance(keys_to_pad, list) or not keys_to_pad:
+        raise ValueError(
+            f"keys_to_pad should be a list of strings with at least one element, but found {keys_to_pad}!"
+        )
+    if isinstance(padding_idx, dict):
+        if isinstance(padding_idx, dict) and not set(padding_idx.keys()) == set(
```
RdoubleA, line 118: do you need to check for dict twice?

torchtune/data/_collate.py
```diff
@@ -12,7 +12,136 @@
...
+            )
+        if not set(keys_to_pad) <= set(batch[0].keys()):
+            raise ValueError(
+                "keys_to_pad should be a subset of keys in the batch, but found "
+                f"{set(keys_to_pad)} and {set(batch[0].keys())}, respectively."
+            )
+
+    batch_keys = [k for k in batch[0].keys() if k not in keys_to_pad]
+    output_dict = {k: torch.tensor([x[k] for x in batch]) for k in batch_keys}
+    for k in keys_to_pad:
```
RdoubleA, line 133: maybe should convert to a set to avoid accidental duplicates

torchtune/data/_collate.py
```diff
@@ -12,7 +12,136 @@
...
+                f"padding_idx was provided as a dictionary, but the keys ({padding_idx.keys()}) "
+                f"are not the same as keys_to_pad ({keys_to_pad})"
+            )
+        if not set(keys_to_pad) <= set(batch[0].keys()):
+            raise ValueError(
+                "keys_to_pad should be a subset of keys in the batch, but found "
+                f"{set(keys_to_pad)} and {set(batch[0].keys())}, respectively."
+            )
+
+    batch_keys = [k for k in batch[0].keys() if k not in keys_to_pad]
```
RdoubleA, line 131: a quick comment explaining that this is taking out keys that won't be padded would be good

torchtune/data/_collate.py
```diff
@@ -12,7 +12,136 @@
...
+        map(lambda x: torch.flip(x, dims=[0]), sequences),
+        batch_first=batch_first,
+        padding_value=padding_value,
+    ).flip(dims=[1])
+
+
 def padded_collate(
+    batch: List[Dict[str, List[int]]],
+    *,
+    pad_fn: Callable,
```
RdoubleA, line 59: I'm still on the fence about this, it feels like a case of generalizing too early. In PPO you are only using left pad. I imagine right pad would also be a common case, but why not just specify with a string "left" or "right"?

RdoubleA (approved):
beautiful

Issue #1326, opened 2024-08-13
Add validation for `model_transform` param in `SFTDataset`


joecummings:
In the docs, we mention that the model_transform is expected to return "tokens" and "mask". 

We should add better validation that these are the returned values and blame the transform. Right now, we just get a key error.

andrewldesousa:
@joecummings hey can you assign this one to me if still available? i can fix this one in the coming days

RdoubleA:
sounds good @andrewldesousa, thanks for taking this up! let us know when you're ready for one of us to look at the changes

PR #1470, opened 2024-09-02
SFT dataset - model transform validation

andrewldesousa:
#### Context
- [ ] add a new feature
- [ ] fix a bug
- [x] other (improve engineering practices and raising value errors for bad transforms)

https://github.com/pytorch/torchtune/issues/1326

#### Changelog
Raise value error and unit tests for this code path

#### Test plan

#### UX

### DIFF
--- a/tests/torchtune/datasets/test_sft_dataset.py
+++ b/tests/torchtune/datasets/test_sft_dataset.py
@@ -20,6 +20,13 @@ class ToDummyMessages(Transform):
         dialogue = sample["dialogue"]
         messages = [Message.from_dict(d) for d in dialogue]
         return {"messages": messages}
+    
+
+class DummyTokenizerInvalidModelTransform(DummyTokenizer):
+    def __call__(self, sample: Mapping[str, Any]) -> Mapping[str, Any]:
+        sample = super().__call__(sample)
+        del sample["tokens"]
+        return sample
 
 
 class TestSFTDataset:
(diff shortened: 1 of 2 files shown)


andrewldesousa:
cc: @joecummings @RdoubleA

felipemello1:
hey @andrewldesousa , thanks for the PR! A few questions:

1. Is this the only dataset that needs it? For example, why not add it to the instructdataset https://github.com/pytorch/torchtune/blob/71ea1d236af1821611c1faed57e6ea7f3c7f1ea6/torchtune/datasets/_instruct.py#L122

2. The if condition may break if the output is not an iterable (you are checking if "key" in X). Should we first confirm that the output is a dictionary?

3. In the error message, to help with debugging, should we output the keys found?

andrewldesousa:
> hey @andrewldesousa , thanks for the PR! A few questions:
> 
> 1. Is this the only dataset that needs it?
> 2. The if condition may break if the output is not an iterable (you are checking if "key" in X). Should we first confirm that the output is a dictionary?
> 3. In the error message, to help with debugging, should we output the keys found?

1. For validating model transform, yes. There is a potential opportunity to validate the same keys returned by PackedDataset if that is valuable to do, but i think that should be considered a separate issue if it is even an issue at all.
2. I think _either approach here is viable_, but I think that 1) the same issue occurs in the message_transform call above and through various parts of the code base. it might add more friction to developers to here to type check every similar situation but of course this is open to discussion. 2) its fine to rely on users here to use the SFTDataset correctly and implement valid transforms which adhere to the Transform spec given there is fairly clear documentation for what is expected of the user. 3) the datasets in this repos that will leverage this class are should be protected against this issue from code review and testing

felipemello1 (approved):
lgtm

RdoubleA (approved):
Thanks for covering this and adding a test. Just need to fix lint and we can merge

PR #1476, opened 2024-09-03
Only merge model weights in LoRA recipe when `save_adapter_weights_only=False`

SalmanMohammadi:
#### Context
- [ ] add a new feature
- [x] fix a bug
- [ ] other (please add here)

#### Changelog

When saving LoRA checkpoints, we only need to copy and merge the full model state dict when `save_adapter_weights_only=False`. For distributed recipes, this change will only skip the merging logic.

I've also added tests for resuming recipe state with LoRA DPO.

#### Test plan

#### UX


tests/recipes/test_lora_dpo_single_device.py
```diff
@@ -0,0 +1,182 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
```
ebsmothers, line 1: You're a hero

SalmanMohammadi: I was just afraid someone would ask me to run the recipe manually, so this felt like less effort.

recipes/lora_finetune_single_device.py
```diff
@@ -527,24 +529,34 @@ def save_checkpoint(self, epoch: int) -> None:
...
-        )
-        ckpt_dict.update({training.MODEL_KEY: merged_state_dict})
+            ckpt_dict.update({training.MODEL_KEY: merged_state_dict})
+        else:
+            # No need to merge state dict if we're only saving adapter weights
+            adapter_state_dict = {
+                k: v
+                for k, v in self._model.state_dict().items()
+                if adapter_key_filter(k)
+            }
```
ebsmothers, line 559: Doesn't this still materialize the entire state dict though?

recipes/lora_finetune_distributed.py
```diff
@@ -628,12 +628,13 @@ def save_checkpoint(
             checkpoint_dict.update({training.ADAPTER_KEY: adapter_state_dict})
 
             # merge the adapter weights and base weights to create the model checkpoint
-            merged_state_dict = get_merged_lora_ckpt(
-                cpu_state_dict,
-                rank=self._lora_rank,
-                alpha=self._lora_alpha,
-            )
-            checkpoint_dict.update({training.MODEL_KEY: merged_state_dict})
+            if not self._save_adapter_weights_only:
```
ebsmothers, line 631: I think @pbontrager had looked at something with this previously, but we still create the full state dict on CPU in this case then, right? To actually reduce the memory footprint to just the adapter weights we'd need to change the call to `get_full_model_state_dict` above (btw it's ok to say "let's do this in a follow-up")

ebsmothers (approved):
A couple questions around whether we can get around materializing the full model state dict when `save_adapter_weights_only=True`. (I think it'll be trickier for the distributed recipe, but would be nice to sort out for single-device.) Other than that looks good!

PR #1482, opened 2024-09-03
Expose ``add_eot`` param to ``Llama3Tokenizer.tokenize_messages``

### LINKED ISSUE #1475: tokenize_messages doesn't expose `tokenize_end` flag from tokenize_message
tokenize_messages calls tokenize_message for each message but doesn't expose these options. During generation, you want tokenize_end = False for the assistant message since the assistant output would go before the end of the message. For tokenize_messages we should either have a separate flag to control tokenize_end for the last message, or combine add_eos and tokenize_end into a single flag that controls the end tokens that would be used for generation.

joecummings:
#### Context
- [ ] add a new feature
- [x] fix a bug
- [ ] other (please add here)

Closes #1475

#### Changelog
* Adds a param to tokenize_messages called ``add_eot`` for "add end-of-turn" token
* Adds a test to ensure that this token is *not* appended if ``add_eot=False``
* Added docstring examples for what this looks like in Llama3

#### Test plan

#### UX

### DIFF
--- a/torchtune/models/llama3/_tokenizer.py
+++ b/torchtune/models/llama3/_tokenizer.py
@@ -241,13 +241,29 @@ class Llama3Tokenizer(ModelTokenizer, Transform):
         self,
         messages: List[Message],
         add_eos: bool = True,
+        add_eot: bool = True,
     ) -> Tuple[List[int], List[bool]]:
         """
         Tokenize a list of messages into a list of token ids and masks.
 
         Args:
             messages (List[Message]): The list of messages to tokenize.
-            add_eos (bool): Wether to add the tokenizer's eos_id. Default True.
+            add_eos (bool): Add the tokenizer's ``eos_id`` (end-of-sequence ID). Default True.
+            add_eot (bool): Add the tokenizer's ``eot_id`` (end-of-turn ID) to final assistant message. Default True.
+
+        Examples:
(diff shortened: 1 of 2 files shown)


torchtune/models/llama3/_tokenizer.py
```diff
@@ -241,13 +241,29 @@ def tokenize_messages(
...
+        add_eot: bool = True,
     ) -> Tuple[List[int], List[bool]]:
         """
         Tokenize a list of messages into a list of token ids and masks.
 
         Args:
             messages (List[Message]): The list of messages to tokenize.
-            add_eos (bool): Wether to add the tokenizer's eos_id. Default True.
+            add_eos (bool): Add the tokenizer's ``eos_id`` (end-of-sequence ID). Default True.
+            add_eot (bool): Add the tokenizer's ``eot_id`` (end-of-turn ID) to final assistant message. Default True.
```
RdoubleA, line 252: would be good to explain what scenarios to use this parameter in (i.e., for generation you would set this to False, in most other cases it should be True)

RdoubleA (approved):

torchtune/models/llama3/_tokenizer.py
joecummings: Excellent point.

PR #1489, opened 2024-09-04
Add back support for QAT module swap flow

andrewor14:
**Summary:** In torchao, QAT was recently refactored to use tensor subclasses instead of module swap. However, tensor subclass implementation doesn't work with all kinds of distribution strategies yet, e.g. FSDP1 and DDP. As a fallback, users may still wish to use the old module swap flow.

**Test Plan:**
```
CUDA_VISIBLE_DEVICES=2,3,4,5,6,7 tune run --nnodes 1 --nproc_per_node 6 qat_distributed --config llama3/8B_qat_full \
    quantizer._component_=torchtune.training.quantization.Int8DynActInt4WeightQATQuantizerModuleSwap
```


ebsmothers (approved):
Once CI is green I think this should be good

PR #1491, opened 2024-09-04
Get nightly jobs passing

ebsmothers:
Remove usage of deprecated version checks we were importing from ao and drop Python 3.8 support

Test plan:

Screenshot 2024-09-04 at 12 44 28 PM


RdoubleA (approved):
What's the motivation for removing 3.8 support? Is it because PyTorch 2.4 is 3.9 and above?

ebsmothers:
> What's the motivation for removing 3.8 support? Is it because PyTorch 2.4 is 3.9 and above?

Ya also ao doesn't support it anymore so when we try to install their nightlies on 3.8 it really screws things up

PR #1494, opened 2024-09-04
Make final EOT tag optional for Llama3 tokenizer

joecummings:
#### Context
- [ ] add a new feature
- [x] fix a bug
- [ ] other (please add here)

#### Changelog

#### Test plan

#### UX

### DIFF
--- a/torchtune/models/llama3/_tokenizer.py
+++ b/torchtune/models/llama3/_tokenizer.py
@@ -212,23 +212,24 @@ class Llama3Tokenizer(ModelTokenizer, Transform):
     def tokenize_message(
         self,
         message: Message,
-        tokenize_header: bool = True,
-        tokenize_end: bool = True,
+        *,
+        add_start_tokens: bool = True,
+        add_end_tokens: bool = True,
     ) -> List[int]:
         """
         Tokenize a message into a list of token ids.
 
         Args:
             message (Message): The message to tokenize.
-            tokenize_header (bool): Whether to prepend a tokenized header to the message.
-            tokenize_end (bool): Whether to append eot or eom id at the end of the message.
+            add_start_tokens (bool): Whether to prepend a tokenized header to the message. Default is True.
+            add_end_tokens (bool): Whether to append eot or eom id at the end of the message. Default is True.
 
         Returns:
             List[int]: The list of token ids.
         """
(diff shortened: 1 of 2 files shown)


torchtune/models/llama3/_tokenizer.py
```diff
@@ -268,9 +282,22 @@ def tokenize_messages(
             if self.max_seq_len and len(tokens) >= self.max_seq_len:
                 break
 
+        # Remove final eot if add_final_eot is False
+        # and last message was from the assistant
+        if not add_final_eot:
```
joecummings, line 287: Not sure how much I like this...

joecummings: What if there are a bunch of assistant messages in the list? Or what if the assistant message comes before the user message?

pbontrager: In what case would you ask for an assistant response where the last message wasn't assistant? Also if there are multiple "prefilled" assistant responses, it would still only care about the last one as that's the one that begins the model's response during decoding.

torchtune/models/llama3/_tokenizer.py
```diff
@@ -268,9 +282,22 @@ def tokenize_messages(
...
+            if message.role == "assistant":
+                tokens = tokens[:-1]
```
pbontrager, line 289: Why are we doing it this way instead of using the tokenize_message tokenize_end flag? This would still break of message eot=False. Or if a new additional end token gets added in the future.

torchtune/models/llama3/_tokenizer.py
```diff
@@ -226,28 +226,42 @@ def tokenize_message(
...
         messages: List[Message],
         add_eos: bool = True,
+        add_final_eot: bool = True,
     ) -> Tuple[List[int], List[bool]]:
         """
         Tokenize a list of messages into a list of token ids and masks.
 
         Args:
             messages (List[Message]): The list of messages to tokenize.
-            add_eos (bool): Wether to add the tokenizer's eos_id. Default True.
```
pbontrager, line 250: Why not just reuse the tokenize_message "tokenize_end" flag and account for all ending tokens? You either want to continue from the middle of a message or you want a completed sequence.

pbontrager (approved):
Thanks!

Issue #1412, opened 2024-08-25
Full finetune recipe not working with FSDP2 CPU offload


ebsmothers:
As in the title.. I spent a bit of time debugging it but haven't figured out the cause yet. E.g. running

```
tune run --nproc_per_node 2 full_finetune_distributed --config llama2/7B_full fsdp_cpu_offload=True
```

gives the error

```
RuntimeError: attempting to assign a gradient with device type 'cpu' to a tensor with device type 'cuda'. Please ensure that the gradient and the tensor are on the same device
```

We aren't passing the CPU offload policy to the top-level sharding of the model [here](https://github.com/pytorch/torchtune/blob/9f37c73453042c5b07f81c4f82e882365d594669/torchtune/utils/_distributed.py#L592), which also needs to be fixed. But even after this change, the error persists. Probably missing something obvious here..

Once we fix, we need to add a test for CPU offload as well to catch this for next time.

ebsmothers:
cc @weifengpy is there extra initialization logic we need to do for proper CPU offloading support? I know in the LoRA recipe you explicitly call `to_empty` on CPU for LoRA params, but thought that was only necessary if we aren't loading those params from the state dict

weifengpy:
> is there extra initialization logic we need to do for proper CPU offloading support? 

I do not expect extra init logic. if we set `fully_shard(cpu_offload=)`, FSDP2 will move parameters to cpu during `state dict loading`. users do not need to call `to_empty` or `.cpu()` or anything. https://github.com/pytorch/pytorch/blob/ed86ac2f25e7ae5d4e432e1c85837e83f4bd17a6/torch/distributed/_composable/fsdp/_fsdp_param.py#L720

btw: I tested your command and it works on my machine, `tune run --nproc_per_node 2 full_finetune_distributed --config llama2/7B_full fsdp_cpu_offload=True`. verified `model.parameters()` are on cpu at `optimizer.step`

if you want to debug at local, if you know the fqn for the problematic param, we can discuss

ebsmothers:
@weifengpy thanks for taking a look. After forward, all params are on CUDA, not CPU. After that the error occurs during backward though (specifically during FSDP's [post_backward](https://github.com/pytorch/pytorch/blob/5612408735ad078ee07309d17b84c34decaf519b/torch/distributed/_composable/fsdp/_fsdp_param_group.py#L339)). Btw when you tested were you on PyTorch nightlies?

Delaunay:
[...]

PR #1495, opened 2024-09-04
[FSDP2] full finetune: move state dict to cpu when cpu offloading

weifengpy:
#### Context
- [x] add a new feature
- [x] fix a bug
- [ ] other (please add here)

resolve: https://github.com/pytorch/torchtune/issues/1412

full finetune with cpu offload: `tune run --nproc_per_node 2 full_finetune_distributed --config llama2/7B_full fsdp_cpu_offload=True`

full finetune without cpu offload: `tune run --nproc_per_node 2 full_finetune_distributed --config llama2/7B_full fsdp_cpu_offload=False`

lora: `tune run --nproc_per_node 2 lora_finetune_fsdp2 --config llama2/7B_lora`

qlora: `tune run --nproc_per_node 2 lora_finetune_fsdp2 --config llama2/7B_qlora`

when cpu offloading, we can move state dict to cpu to avoid peaking memory. As the snapshot shows, peak memory dropped from 7GB to 3GB: 

memory behavior before
Screenshot 2024-09-04 at 3 29 52 PM

memory behavior after
Screenshot 2024-09-04 at 3 31 29 PM

### DIFF
--- a/torchtune/training/_distributed.py
+++ b/torchtune/training/_distributed.py
@@ -285,6 +285,7 @@ def load_from_full_model_state_dict(
     device: torch.device,
     is_rank_zero: bool,
     strict: bool = False,
+    cpu_offload: bool = False,
 ):
     """
     Converting full state dict into a sharded state dict
@@ -338,6 +339,8 @@ def load_from_full_model_state_dict(
                 sharded_meta_param.device_mesh,
                 sharded_meta_param.placements,
             )
+        if cpu_offload:
+            sharded_tensor = sharded_tensor.cpu()
         sharded_sd[param_name] = nn.Parameter(sharded_tensor)
     # choose `assign=True` since we cannot call `copy_` on meta tensor
     return model.load_state_dict(sharded_sd, strict=strict, assign=True)
@@ -584,4 +587,4 @@ def shard_model(
             fully_shard(m, **fsdp_kwargs)
 
     # Finally shard the entire model to account for any stragglers
-    fully_shard(model)
+    fully_shard(model, **fsdp_kwargs)
--- a/torchtune/training/_distributed.py
+++ b/torchtune/training/_distributed.py
@@ -8,7 +8,7 @@
 import logging
 import os
 from itertools import chain
-from typing import Any, Callable, cast, Dict, List, Set, Tuple, Type
+from typing import Any, Callable, cast, Dict, List, Set, Tuple, Type, Optional
 
 import torch
 import torch.distributed as dist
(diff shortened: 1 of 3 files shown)


torchtune/training/_distributed.py
```diff
@@ -584,4 +587,4 @@ def shard_model(
             fully_shard(m, **fsdp_kwargs)
 
     # Finally shard the entire model to account for any stragglers
-    fully_shard(model)
```
weifengpy, line 587: this seems to be dropping `**fsdp_kwargs` by accident? it prevents cpu_offloading for root model

torchtune/training/_distributed.py
```diff
@@ -338,6 +339,8 @@ def load_from_full_model_state_dict(
                 sharded_meta_param.device_mesh,
                 sharded_meta_param.placements,
             )
+        if cpu_offload:
+            sharded_tensor = sharded_tensor.cpu()
```
weifengpy, line 343: `sharded_tensor` have `device=cuda` because `distribute_tensor`/DTensor requires NCCL. For cpu offloading, we can move DTensor to `device=cpu` afterwards to avoid peaking memory

weifengpy:
this torchtune side fix alone can resolve the problem, because we move state dict to cpu when `fsdp_cpu_offload=True`

pytorch side fix covers cases when state dict are on gpu

torchtune/training/_distributed.py
ebsmothers: Yeah I had meant to fix this, thanks for adding it here!

ebsmothers:
Thank you @weifengpy for figuring this out and landing the fix!

ebsmothers (approved):

ebsmothers:
Btw @weifengpy I assume for save_checkpoint we will need to do the inverse operation, right? Move from CPU back to current device prior to resharding?

weifengpy:
> Btw @weifengpy I assume for save_checkpoint we will need to do the inverse operation, right? Move from CPU back to current device prior to resharding?

you're right. I need to update the PR to cover save_checkpoint

torchtune/training/_distributed.py
awgu, line 343: This change makes sense to me. I am not sure if we should support the user trying to load a GPU state dict into an FSDP module that has CPU offloading enabled. We can provide a better error message though.

weifengpy: got you. this reminds me of my overdued BE to check cpu device in lazy_init when cpu offloading is enabled

PR #1496, opened 2024-09-05
hotifx - remove param from input of rope_init

felipemello1:
#### Context
- [ ] add a new feature
- [x] fix a bug
- [ ] other (please add here)

In https://github.com/pytorch/torchtune/pull/1386 i introduced a bug by requiring parameters in rope_init,i .e. 
```
def rope_init(self, scaling_factor):
	foo = baa(scaling_factor)
```

 instead of:
```
def rope_init(self): 
	foo = baa(self.scaling_factor)
```

When trying to run distributed, it broke, because distributed calls rope_init before compile. This PR fixes it by removing inputs from rope_init.

#### Test plan
ran ```tune run --nproc_per_node 4 full_finetune_distributed --config llama3_1/8B_full``` without errors

### DIFF
--- a/torchtune/models/llama3_1/_position_embeddings.py
+++ b/torchtune/models/llama3_1/_position_embeddings.py
@@ -65,6 +65,10 @@ class Llama3ScaledRoPE(nn.Module):
         self.rope_init()
 
     def rope_init(self):
+        """
+        Warning: this is called in recipes before torch.compile,
+        so that the cache is built in advance.
+        """
         freqs = 1.0 / (
             self.base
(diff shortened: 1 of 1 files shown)


joecummings (approved):

felipemello1:
@ebsmothers, do our regression tests call the recipes? It should have failed if we were using the distributed full recipe: https://github.com/pytorch/torchtune/blob/d31649e75e51cf15dbdca004288c32550e8f7f8d/recipes/full_finetune_distributed.py#L428C32-L428C41

PR #1499, opened 2024-09-05
Split q and kv in MHA

pbontrager:
#### Context
- [ ] add a new feature
- [x] fix a bug
- [ ] other (please add here)

During decoding, it's not guaranteed that there will be a second input for cross attention. For MHA, this translates to no y input. The kv values are static in that case and retrieved from the cache while only the q value is new (the new values being generated by the decoder's previous output). To handle this efficiently, we split the kv code from the q code so that we can optionally skip kv code when no y input is provided.

#### Changelog
- Split q and kv code, place kv code in a conditional dependent on y
- Since y is now optional for SA and CA, we alway route y to kv. Update SelfAttentionLayer to pass in MHA x and y.
		- This looks like attn(x, x) for self attention.
No qkv lines were changed, just moved.
#### Test plan

#### UX


torchtune/modules/attention.py
```diff
@@ -190,8 +191,11 @@ def forward(
                 During inference, this indicates the position of the current token.
                 If none, assume the index of the token is its position id. Default is None.
 
+        Raises:
+            ValueError: If no `y` input and `kv_cache` is not enabled.
+
         Returns:
-            Tensor: output tensor with attention applied
+            torch.Tensor: output tensor with attention applied
```
ebsmothers, line 198: 👀

torchtune/modules/kv_cache.py
```diff
@@ -64,7 +65,7 @@ def update(
             Tuple[torch.Tensor, torch.Tensor]: Updated KV cache with key first
         """
         assert input_pos.shape[0] == k_val.shape[2]
-        self.size = input_pos.max().item() + 1
+        self.size = input_pos.max() + 1
```
ebsmothers, line 68: Is this just to get rid of a sync? Or was this actually broken before

torchtune/modules/attention.py
```diff
@@ -176,8 +176,9 @@ def forward(
     ) -> torch.Tensor:
         """
         Args:
-            x (torch.Tensor): input tensor with shape [b x s_x x d]
-            y (Optional[torch.Tensor]): second input tensor for cross attention with shape [b x s_y x d]
+            x (torch.Tensor): input tensor with shape [b x s_x x d] for the query
+            y (Optional[torch.Tensor]): second input tensor for key and value with shape [b x s_y x d].
+                Optional only with kv_cache enabled.
```
ebsmothers, line 181: nit: this wording is confusing. Also can we explicit say something like "for self-attention, set y=x" somewhere?

torchtune/modules/attention.py
```diff
@@ -205,65 +209,78 @@ def forward(
...
-        if self.kv_cache is not None:
-            k, v = self.kv_cache.update(input_pos, k, v)
+        if y is None:
+            if self.kv_cache is None:
+                raise ValueError(
+                    "Must provide y input or use kv_cache to enable streaming decoding"
+                )
+            k = self.kv_cache.k_cache
+            v = self.kv_cache.v_cache
+        else:
```
ebsmothers, line 246: add a comment to say that we are doing K and V projections and optionally applying RoPE here since K and V are not cached (basically something to offset the hit to readability by splitting Q logic from K and V logic)

ebsmothers (approved):
A couple small comments, otherwise looks good

torchtune/modules/kv_cache.py
pbontrager: Just to get rid of the sync and match what's being done in TransformerDecoder

PR #1501, opened 2024-09-05
Added Inference Option to Tokenizer/Transform

pbontrager:
#### Context
- [ ] add a new feature
- [x] fix a bug
- [ ] other (please add here)

Some transforms behave differently during inference than during training. For example, when we tokenize_messages, during inference we want add_end_tokens=False so that the model can "respond" to the prompt before end tokens are added. This flag allows for transforms to modify behavior depending on the situation.

#### Changelog
 - Added inference=False as a new parameter to every tokenizer/transform/template __call__
 - Updated tokenize_messages to `add_end_tokens= not inference` for every tokenizer
 - Updated VisionCrossAttentionMask to match kv_cache size if inference=True.
 
#### Test plan

#### UX


joecummings (commented):
tests?

torchtune/data/_prompt_templates.py
```diff
@@ -87,13 +89,16 @@ def __init__(
     ):
         self.template = template
 
-    def __call__(self, messages: List[Message]) -> List[Message]:
+    def __call__(
+        self, messages: List[Message], inference: bool = False
```
RdoubleA, line 93: this isn't being used anywhere?

torchtune/models/clip/_transform.py
```diff
@@ -144,13 +144,16 @@ def __init__(
         self.tile_crop = tile_crop
         self.pad_tile_size = max_num_tiles if pad_max_tiles else None
 
-    def __call__(self, sample: Mapping[str, Any]) -> Mapping[str, Any]:
+    def __call__(
+        self, sample: Mapping[str, Any], inference: bool = False
```
RdoubleA, line 148: same here

torchtune/data/_prompt_templates.py
```diff
@@ -25,13 +25,15 @@ class PromptTemplateInterface(Protocol):
...
         messages: List[Message],
+        inference: bool = False,
     ) -> List[Message]:
         """
         Format each role's message(s) according to the prompt template
 
         Args:
             messages (List[Message]): a single conversation, structured as a list
                 of :class:`~torchtune.data.Message` objects
+            inference (bool): Whether the template is being used for inference or not.
```
RdoubleA, line 36: should describe what this practically changes

torchtune/modules/transforms/_transforms.py
```diff
@@ -159,8 +176,10 @@ def __call__(self, sample: Mapping[str, Any]) -> Mapping[str, Any]:
             # Mask will be block of 1s at the corresponding interval in the text.
             # It is not a causal block because all the image tokens correspond
             # to a single image, so text tokens attend to all the image's tokens
-            mask = torch.zeros(text_seq_len, image_seq_len, dtype=torch.bool)
-            mask[start:end, :] = True
+            text_size = text_seq_len if not inference else self.max_seq_len
```
RdoubleA, line 179: need to check if max_seq_len is not None

RdoubleA:
should update respective tests with inference=True to make sure we get expected behavior

torchtune/data/_prompt_templates.py
```diff
@@ -151,13 +156,15 @@ class ChatMLTemplate(PromptTemplateInterface):
...
         messages: List[Message],
+        inference: bool = False,
     ) -> List[Message]:
         """
         Format user, assistant, and system messages with appropriate tags.
 
         Args:
             messages (List[Message]): a single conversation, structured as a list
                 of `Message` objects
+            inference (bool): Whether the template is being used for inference or not.
```
RdoubleA, line 167: probably not needed for this PR, but this is a good opportunity to use the inference flag for the empty assistant message logic below

pbontrager:
I updated all of the current tests to tokenizer/transform `__call__` method

RdoubleA (approved):

PR #1502, opened 2024-09-05
Adding Classification Use Case

qqlabs:
initial attempt at getting classification, validation, adapter checkpointing, and some memory optimization working

#### Context
- [ ] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

#### Changelog

#### Test plan

#### UX


recipes/lora_finetune_classification_distributed.py
```diff
@@ -0,0 +1,991 @@
...
+        self._optimizer = self._setup_optimizer(
+            cfg_optimizer=cfg.optimizer,
+            opt_state_dict=checkpoint_dict[training.OPT_KEY]
+            if self._resume_from_checkpoint
+            else None,
+        )
+
+        self._loss_fn = config.instantiate(cfg.loss)
+
+        ############################
```
qqlabs, line 257: 2 dataloaders right now so that we can do train and validation

recipes/lora_finetune_classification_distributed.py
```diff
@@ -0,0 +1,991 @@
...
+            )
+
+            # Load both the base model weights and (if available) the adapter weights. Both
+            # of this should happen only on Rank 0
+
+            ############################
+            # Since this is classification drop the output weights from base llama model
+            # Otherwise load_state_dict would have key size mismatch
+            ############################
+            del base_model_state_dict["output.weight"]
```
qqlabs, line 443: We can't directly load in the base llama model weights since they don't match the newly replaced mlp layer for the classification version. Current hack is to drop the weights before loading into the classification model.

recipes/lora_finetune_classification_distributed.py
```diff
@@ -0,0 +1,991 @@
...
+        ############################
+        dataloader = DataLoader(
+            dataset=ds,
+            batch_size=batch_size,
+            sampler=sampler,
+            collate_fn=(
+                partial(
+                    padded_collate,
+                    padding_idx=self._tokenizer.pad_id,
+                    # ignore_idx=self._loss_fn.ignore_index,
```
qqlabs, line 573: BCEWithLogitsLoss doesn't have ignore_index

recipes/lora_finetune_classification_distributed.py
```diff
@@ -0,0 +1,991 @@
...
+
+        # zero out the gradients before starting training
+        self._optimizer.zero_grad()
+
+        # Initialize tokens count and running loss (for grad accumulation)
+        t0 = time.perf_counter()
+        running_loss = 0
+        num_tokens = 0
+
+        ############################
```
qqlabs, line 687: Validation only works at end of epoch

recipes/lora_finetune_classification_distributed.py
```diff
@@ -0,0 +1,991 @@
...
+                num_tokens += tokens.numel()
+                labels = labels.to(self._device)
+                mask = mask.to(self._device) if mask is not None else None
+                input_pos = (
+                    input_pos.to(self._device) if input_pos is not None else None
+                )
+
+                logits = self._model(tokens, mask=mask, input_pos=input_pos)
+
+                ############################
```
qqlabs, line 737: added step here where we pick the output for the eos token. This should ideally be built into the model itself - not sure what's the best way of doing that.

recipes/lora_finetune_classification_distributed.py
```diff
@@ -0,0 +1,991 @@
...
+                            log_dict,
+                            step=self.global_step,
+                        )
+
+                    # Reset running stats for the next step
+                    running_loss = 0
+                    num_tokens = 0
+                    t0 = time.perf_counter()
+
+                    ############################
```
qqlabs, line 815: free up memory at end of training step

recipes/lora_finetune_classification_distributed.py
```diff
@@ -0,0 +1,991 @@
...
+            # if (self.global_step % self._validate_every_n_steps != 0):
+            self.validation(curr_epoch)
+
+            self.epochs_run += 1
+            print("Saving checkpoint")
+            self.save_checkpoint(epoch=curr_epoch)
+
+        self._profiler.stop()
+
+    ############################
```
qqlabs, line 869: validation loop

torchtune/datasets/_prompt_classification.py
```diff
@@ -0,0 +1,183 @@
...
+
+    def __init__(
+        self,
+        tokenizer: ModelTokenizer,
+        source: str,
+        from_disk: bool = False,
+        text_column: str = "text",
+        label_column: str = "label",
+        split: str = "train",
+        classes: Optional[List[Any]] = None,
```
qqlabs, line 48: class/label mapping currently passed in as a an ordered list. Not sure what's the best practice for this (do we store a separate label mapping file somewhere?)

torchtune/models/convert_weights.py
```diff
@@ -246,6 +246,29 @@ def tune_to_peft_adapter_config(
     return adapter_config
 
 
+# When we only save adapter weights, we need to convert back from peft format to tune format
```
qqlabs, line 249: I use this when trying to load adapter config weights back. Will add sample code for what I got working later

recipes/lora_finetune_classification_distributed.py
```diff
@@ -0,0 +1,991 @@
...
+        if self._dtype == torch.bfloat16:
+            model = model.to(torch.bfloat16)
+
+        # LoRA hyper-params needed for merging weights while saving checkpoints
+        self._lora_rank = cfg_model.lora_rank
+        self._lora_alpha = cfg_model.lora_alpha
+
+        # Note: this needs to be set before wrapping with FSDP
+        self.adapter_params = get_adapter_params(model)
+        set_trainable_params(model, self.adapter_params)
```
qqlabs, line 462: do we also need to set the final mlp layer of the base model to be trainable?

recipes/lora_finetune_classification_distributed.py
felipemello1, line 1: Thanks for the PR! This is not a review yet, but just food for thought: We have been having conversations for a while about removing the output head from the LLM forward, and running it as a second step.

Do you think that if this was implement, we would still need a different recipe? For example:

```
hidden_state = model(inputs)
output = your_output_head(hidden_state)
loss = output(logits, label)
```

If we could make it work without adding a bunch of if/else everywhere in the recipe, then we could leverage the existing ones instead of 2x the number of recipes in the library

qqlabs: to clarify, did you mean `loss = _loss_fn(output, label)` for the last line?

I think this could work to separate out the model forward logic for the classification use case (where we need to find the last token and pick the hidden state for it). Another option is to build the forward logic into the model itself like how huggingface transformers are implemented.

The other big change for classification/prompt completion use cases is the logic for processing the labels. The chat/instruct use cases shift all the logits by 1 to do next token prediction for the labels. Prompt completion/classification tasks have the label defined in the dataset and processed by the dataloader. We would need to figure out how to pull out that label processing logic as well to avoid duplicating recipes.

SalmanMohammadi:
@qqlabs check https://pytorch.org/torchtune/main/generated/torchtune.training.update_state_dict_for_classifier.html#torchtune.training.update_state_dict_for_classifier which might help your troubles on checkpointing. Essentially you just need to call this function once before any LoRA state dict validation and it should work. There's an example in the PPO recipe for non-LoRA state dict loading.

recipes/lora_finetune_classification_distributed.py
felipemello1: > to clarify, did you mean loss = _loss_fn(output, label) for the last line?

yes! Thanks!

> We would need to figure out how to pull out that label processing logic as well to avoid duplicating recipes

That sounds manageable. It could be in the dataloader or some utility function. 

I guess one 3rd issue is the checkpoint. I am not sure if this would require something different for classification vs generation.

Assuming you want to follow through, if you could adapt an existing generation recipe to support both generation/classification by making these changes, i think it would be easy to approve and a huge contribution to torchtune.

As a proof-of-concept, its fine to have some if/else and not be polished. And if it looks good, then we can brainstorm about how to polish it. What do you think?

SalmanMohammadi: FYI there are a lot of changes coming to generation in #1424 which might be worth holding off for. I'm actually not sure inference for seq->label should be in the generation recipe which is beefed up for next-token prediction tasks + kv cacheing + compile. Not really stuff we need for classification tasks right?

I think the code we need for prediction here is pretty minimal - grab a prompt, run the model, then extract the logits at the last valid token (i.e. eos token) and activation function/argmax on those.

SalmanMohammadi:
I also haven't looked too closely at how you're predicting the labels. AFAIK the standard way (which is also what HF does) is to grab the logits from the model at the last valid token in the prompt - this will usually be the EOS token if it exists, or the last token otherwise, then use your activation functions / argmaxing etc on the logits for that token(s). I can point you to an example of this in the codebase - we actually used to have a separate utility just for this but it was lost in a refactor.

qqlabs:
> I also haven't looked too closely at how you're predicting the labels. AFAIK the standard way (which is also what HF does) is to grab the logits from the model at the last valid token in the prompt - this will usually be the EOS token if it exists, or the last token otherwise, then use your activation functions / argmaxing etc on the logits for that token(s). I can point you to an example of this in the codebase - we actually used to have a separate utility just for this but it was lost in a refactor.

Great, this is what I did [here](https://github.com/pytorch/torchtune/pull/1502#discussion_r1746141492)! This is currently copied from HF's implementation, but using the token before the eos token instead of the padding token which HF does.

recipes/lora_finetune_classification_distributed.py
qqlabs: Thanks for all the feedback! I have some immediate use cases I need to get working, so will be focused on getting a working proof of concept that can replicate my current finetuned model that I trained with huggingface. Once we reach parity, we can know that the logic is correct and start refactoring into a more generalized design.

My current plan is:
1. Create an evaluate recipe for classification so that we have some way to batch deploy trained models. I'll use this to check whether my current models are reaching parity with my baseline relevance models.
2. Fix training logic for classification. My validation metrics are showing increasing validation loss and very low validation accuracy right now, so I think I have a bug somewhere. Trying to hunt this down.
3. Make checkpoint loadable by HF from_pretrained. There are some format conversion issues that I had to adjust to get from_pretrained working. I'm also getting different logits for some reason when running eval for torchtune loaded checkpoint vs HF loaded checkpoint, so more investigation needed here.

Once the above are good to go, then can start refactoring to cleaner code design.

torchtune/models/convert_weights.py
SalmanMohammadi: As in, resuming training in torchtune using adapter weights saved from a previous torchtune LoRA run? How come that wasn't working?

torchtune/models/llama2/_model_builders.py
```diff
@@ -369,6 +368,83 @@ def lora_llama2_reward_7b(
         quantize_base=quantize_base,
     )
 
+# Add llama multi classification model builders
```
SalmanMohammadi, line 371: I definitely want to make it easier to use classification models in our codebase, this is high-up on my TODO list - your comment about replacing reward models is on-point. Soon you should just be able to use wrap any existing model builder without having to re-define builders.

recipes/lora_finetune_classification_distributed.py
SalmanMohammadi: Yes - excellent catch. This is done through the `apply_lora_to_output` arg

recipes/lora_finetune_classification_distributed.py
SalmanMohammadi: Sounds good! I think arriving at a sane baseline first of all is a good idea.

> Make checkpoint loadable by HF from_pretrained. There are some format conversion issues that I had to adjust to get from_pretrained working.
I'm curious about this - what changes did you have to make?

> I'm also getting different logits for some reason when running eval for torchtune loaded checkpoint vs HF loaded checkpoint, so more investigation needed here.

If you have any code to share here I'd be happy to look. I've tried matching logits between HF before, but also couldn't get it to match. There are several differences in the underlying transformer modelling code, and quite a lot of differences between generation. I haven't found this to be an issue yet, though. When testing against HF baselines, a good sanity check is to use less granular comparisons e.g. similar loss curves/similar evaluation metrics across identical setups.

qqlabs:
> @qqlabs check https://pytorch.org/torchtune/main/generated/torchtune.training.update_state_dict_for_classifier.html#torchtune.training.update_state_dict_for_classifier which might help your troubles on checkpointing. Essentially you just need to call this function once before any LoRA state dict validation and it should work. There's an example in the PPO recipe for non-LoRA state dict loading.

@SalmanMohammadi I just realized this function is a bit problematic for lora classification. From my understanding, the lora adapter values are [added](https://github.com/qqlabs/torchtune/blob/5a85048d4518c012466936a4f4807b2766daba15/torchtune/modules/peft/_utils.py#L250) to the corresponding layers when merge the adapter into the base model. This would mess up the output layer since we're replacing the base model's output layer with random weights before merging in lora adapter values - thus messing up the output weights post training right? We need to keep track of the original output weights, which isn't done if we only save the adapter weights during training now.

SalmanMohammadi:
I'm sorry I'm not 100% sure I follow. If we're fine-tuning a language model for a classification task, here's roughly what I think should be happening:
1) We should have a corresponding classifier model, which is identical to the base model, but the output projection layer is changed.
2) Now, when we want to finetune our base model, we load all the weights in, aside from the output projection, which remains randomly initialized as you pointed out. From here on out we have no further need of the base model state dict.
3) In full fine-tuning, we'll fine-tune the whole output projection weights, and in LoRA we'll fine-tune just an adapter on top of the randomly initialized weights.

Does that sound roughly right to you? I'm curious what you mean by keeping track of the original output weights.

You've made me think of something important here though. Adapter based fine-tuning might not be suitable for the output projection which is randomly initialized. This means we might want to do something like adapter-based fine-tuning for the rest of the model, don't apply an adapter to the output projection, and just unfreeze the final layer so it trains from scratch. 

I've been doing some LoRA classification model fine-tuning, but I'm going from pertained classifier -> finetuned classifier so I haven't hit the same edges cases here. This is really useful.

qqlabs:
> 4. If LoRA fine-tuning, we merge the weights back in post training (if we've not set save adapters only). This maybe isn't great if we're merging trained output projection adapters with a randomly initialized output head.

This is the part that I'm worried about. I think it's ok right now if we have the save_adapter_weights_only set to false since we would merge the weights with the trained model while calling save_checkpoint and can load the whole model back in as is without looking at the adapter weights.

The issue is when we only save adapter weights as checkpoints since we lose the "random initialized output weights" that the adapter weights were trained on.

Current steps to load the model with only adapter checkpoints:
1. Download original llama2_7b base model weights
2. Load the base model weights.
3. We don't have the random initialized weights that we used during finetuning which are associated with the adapter checkpoint. This means we can only call update_state_dict_for_classifier which creates a new random set of weights for the output layer.
4. Merge the adapter weights with base model weights. At this step, we're basically adding a random number to each of the output layer weights of the adapter.

I think a finetuned llama model with a completely random output layer would produce random results. I'm seeing exactly random accuracy (~25% for my 4 class classification) with my eval script with this setup.

I did have the full model checkpointed for one of my runs and I am able to reproduce something closer to the validation metrics that I was seeing during my training process (which has some other bug since I'm getting like 9% accuracy...).

Overall, we cannot only save the adapter weights for classification tasks - we need to also save the randomly initialized output weights (or trained output weights) somewhere (could be part of the adapter weights).

SalmanMohammadi:
Can you try set ` apply_lora_to_output: False`, and then as a quick hack add, re-enable `requires_grad=True` on the output weight parameters? This should mean we freezer base model weights -> setup adapter params (not for the output weight) -> unfreeze output weight -> train. 

This won't work with `save_adapter_weights_only`. I think longer term for classification tasks, we need some sensible solution which either trains the output weight from scratch, or, if you're not starting from randomly initialized weights (finetuning an existing classifier), only then adding an adapter.

PR #1504, opened 2024-09-05
Images in Messages

joecummings:
### Background

No snappy intro, I'm just tired.

This all started in considering the inference UX for multimodal models. The ideal state would be something like this in the config:

```yaml
prompt:
	- system: You are a helpful assistant. 
	- user:
		- image: https://en.wikipedia.org/wiki/Wikipedia:Images#/media/File:Spinifex_Pigeon_0A2A1585.jpg
		- text: What's in this image? 
```

Then the user can call the generate recipe:

```bash
python recipes/generate.py --config recipes/configs/multimodal_generation.yaml
This is an image of a bird sitting on a tree. 
```

**But, what is this getting at?**

The concept of a "Message" should contain all the information needed for an interaction with a model INCLUDING images. Before, images were split off from Messages and a placeholder was put in the Message object so the model knew where to inject the content:

```python
message_content = [
	{"type": "image"},
	{"type": "text", "content": "What is in this image?"},
]

images = ["https://en.wikipedia.org/wiki/Wikipedia:Images#/media/File:Spinifex_Pigeon_0A2A1585.jpg"]
```

After these changes, the model now has ALL the information it needs within the message content itself:

```python
message_content = [
	{"type": "image", "content":  <PIL.Image.Image>},
	{"type": "text", "content": "What is in this image?"},
]
```

The core of the change is relatively simple; however, we already backed the split images work into all our multimodal stuff, so that all had to be reworked as well. See below for a full changelog.

### Changelog

* Added "content" field for PIL type in message content
* Added a `get_media` function to easily get all the image content from a Message
* Updated the Flamingo transform to use this new functionality
* Added a function to load a PIL image from local drive or remote source
* Added a new function to format content with images properly
* Update Cauldron and Llava Instruct datasets with new changes
* Tests!!! Including uploading an asset to test if it can actually load an image properly

### Testing

```bash
(joe-torchtune) [jrcummings@devvm050.nha0 ~/projects/joe-torchtune (3fb44c161)]$ python -m pytest tests/torchtune/
...
 =============================================== 467 passed, 5 skipped, 8 warnings in 119.40s (0:01:59) ===============================================
```

### FAQs

**Why did you make it possible to load from remote or local sources?** This is a feature usually demonstrated for inference like in [Phi3 Vision ](https://huggingface.co/microsoft/Phi-3.5-vision-instruct#loading-the-model-locally) and [Qwen2 vision](https://huggingface.co/Qwen/Qwen2-VL-7B-Instruct#quickstart). We don't want to force users to have to download the image to device to start, so this handles both. 

**Why not convert to PIL image within Message?** ~I wanted to do this originally; however, Message is a class used for text and multimodal models. Right now, it's possible to train and run inference without downloading extra things needed for multimodal like PIL and torchvision. This keeps the library lightweight. In keeping with that, if we make the specific multimodal model transforms load and use PIL, we keep this usage separate. I'd be open to changing if PIL and Multimodal becomes so important that it's used everywhere, but right now it's not.~  I did it, sue me.


torchtune/data/_messages.py
```diff
@@ -59,6 +59,7 @@ def __init__(
         eot: bool = True,
     ):
         self.role = role
+        # What is this?
```
RdoubleA, line 62: lets people just pass in a string and still get the dictionary format

RdoubleA (commented):
This would mean that model transform just takes in a `List[Message]` directly instead of a sample dictionary, or do we still plan to pass in `sample["messages"]` to model transforms?

torchtune/datasets/multimodal/_llava_instruct.py
```diff
@@ -89,6 +89,8 @@ def __call__(self, sample: Mapping[str, Any]) -> Mapping[str, Any]:
                     role="system", content=self.new_system_prompt, masked=True, eot=True
                 )
             )
+        
+        # Add in image stuffs / load from file
```
RdoubleA, line 93: I am leaning towards handling image loading in the model transform and keeping a URL here, so we don't need to pass in a heavy PIL image everywhere Messages is used. We only load in the image when it's absolutely needed. For tokenize_messages, messages would be more lightweight since it won't contain the actual image.

ebsmothers: Re passing around PIL images, do you mean that there are implications for memory to doing something like this? Cause if we are just passing stuff by reference it shouldn't be an issue, right? 

If there are a bunch of operations that we want to apply to messages that'd be more convenient to have the file path for, that'd be one thing. But it seems to me like most (nontrivial) functions we're gonna call on Messages containing images would be those that act on the image and not the path. 

In that case I'd be inclined to have things as a PIL because it's the raw datatype of an image. It's analogous to the raw string we have in the content field of a Message with text type. I think {"text" -> str, "image" -> PIL} is a lot more natural for a user than {"text" -> str, "image" -> Path}. But lmk if I'm missing the point here

joecummings: See comment in FAQs. Generally, I agree, but I'm making the tradeoff for "weight" of dependencies.

RdoubleA: > Right now, it's possible to train and run inference without downloading extra things needed for multimodal like PIL and torchvision

I think Joe's point here is quite valid. I agree having it PIL image by default is more intuitive but as of now multimodal is not prevalent enough as a use case.

tests/assets/dog_on_skateboard.jpg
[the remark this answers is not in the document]
joecummings: "Requests changes"

😭

torchtune/datasets/multimodal/_the_cauldron.py
```diff
@@ -91,7 +91,10 @@ def __call__(self, sample: Mapping[str, Any]) -> Mapping[str, Any]:
                 Message(
                     role="user",
                     content=[
-                        {"type": "image"},
+                        {
+                            "type": "image",
+                            "content": sample[self._column_map["images"]],
```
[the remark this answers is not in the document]
joecummings: I mean it can - but it should be consistent

RdoubleA:
> No snappy intro, I'm just tired.
I feel that

torchtune/datasets/multimodal/_the_cauldron.py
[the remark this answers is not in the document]
joecummings: Damn, if this is coming in as a PIL image, then it might make sense to go ahead with the changes to load image into Message as a PIL image

torchtune/models/clip/_transform.py
```diff
@@ -24,6 +25,36 @@
 logger = logging.getLogger(__name__)
 
 
+def _load_image(image_path: str) -> Image.Image:
```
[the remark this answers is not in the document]
joecummings: Great question. Really it belongs in modules/vision/utils.py but that whole thing needs a refactor, which I don't want to handle in this PR. 

For now, just leaving it in CLIP

tests/torchtune/datasets/multimodal/test_llava_instruct_dataset.py
```diff
@@ -55,7 +55,8 @@ def test_label_no_masking(self, load_dataset, tokenizer):
             model_transform=tokenizer,
             train_on_input=True,
         )
-        input, labels, images = ds[0]["tokens"], ds[0]["labels"], ds[0]["images"]
+
+        input, labels = ds[0]["tokens"], ds[0]["labels"]
```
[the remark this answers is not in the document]
joecummings: Not sure I follow. Our tokenizers  (model transforms) take in images through the messages, but they do not return them afterwards.

Looking at Flamingo transform for this.

torchtune/datasets/multimodal/_llava_instruct.py
```diff
@@ -89,15 +90,23 @@ def __call__(self, sample: Mapping[str, Any]) -> Mapping[str, Any]:
...
             role = role_map[message["from"]]
+            content = message["value"]
             if role == "system" and self.new_system_prompt is not None:
                 continue
-            content = split_text_by_image_tag(message["value"], "<image>")
+            if role == "user":
+                content = format_content_with_images(
+                    content,
+                    image_tag="<image>",
+                    images=[sample[self._column_map["image"]]],
```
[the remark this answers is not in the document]
joecummings: Great catch

torchtune/datasets/multimodal/_the_cauldron.py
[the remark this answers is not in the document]
RdoubleA: ha... thinking about it a bit more, I suppose there's no actual memory hit. if you're using multimodal datasets you will certainly need to load the image at some point. then again, changing my mind after five minutes hurts my pride just a little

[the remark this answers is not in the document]
joecummings: I don't really want a message to be ```Union[Path, str, PIL.Image]```

That seems like a recipe for confusion.

tests/torchtune/datasets/multimodal/test_llava_instruct_dataset.py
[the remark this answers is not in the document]
RdoubleA: we're using `DummyTokenizer` as the substitute for model transform here. Really it should be called `DummyModelTransform`. My point is we should test that images are passed through from the dataset to messages and are processed correctly. Right now with these changes we are not checking images at all, only the image special tokens.

torchtune/datasets/multimodal/_the_cauldron.py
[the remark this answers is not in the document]
joecummings: ALRIGHT F IT, here's the plan:

I was a big ol dumb dumb. This stuff is gunna go. Messages will now hold PIL images. This will be loosely typed in Messages so that we don't need to actually import PIL and Messages can stay free of that dependency hell. Then, the onus for providing a proper PIL image to the Message class is on the dataset builder and the generation recipe. For generation, we could just have a separate one for multimodal or require PIL for everything. For dataset builder, everything is in the multimodal folder., which protects us from imports (thank for the forward thinking @RdoubleA)

This makes the most logical sense IMO. 

data entry points (datasets, generation) load in the PIL image -> feed to message in proper format -> message is used by model transform however it wants. 

Okay? Okay.

torchtune/datasets/multimodal/_llava_instruct.py
ebsmothers: OK yeah that's fine with me

torchtune/datasets/multimodal/_the_cauldron.py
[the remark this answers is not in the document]
ebsmothers: Idk what just happened here but I'm on board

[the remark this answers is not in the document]
RdoubleA: > ALRIGHT F IT

you right, I'm on board

torchtune/datasets/multimodal/_llava_instruct.py
```diff
@@ -89,22 +94,34 @@ def __call__(self, sample: Mapping[str, Any]) -> Mapping[str, Any]:
...
-        return {"messages": messages, "images": sample[self._column_map["image"]]}
+        return {"messages": messages}
 
 
 # TODO: point to Flamingo model transform as an example
 def llava_instruct_dataset(
     model_transform: Transform,
     *,
     source: str = "liuhaotian/LLaVA-Instruct-150K",
+    images_dir: str = "coco/",
```
joecummings, line 124: I would love some input here. Right now this is a string b/c I'm anticipating usage from a config. However, the proper typing would be something like Path (Pathlib) b/c it would be much more tolerant to errant forward-slashes and plays better with combination of sub paths. 

But then, I'll have to do some conversion on the backside to do that. 

Thoughts? @RdoubleA @ebsmothers

RdoubleA: If we convert to Path on the backside but still maintain a string parameter, that should handle any trailing/incorrect slashes right? If so I would do that

torchtune/data/_utils.py
```diff
@@ -53,7 +90,7 @@ def split_text_by_image_tag(content: str, image_tag: str) -> List[Dict[str, str]
                 "role": "system" | "user" | "assistant",
                 "content":
                     [
-                        {"type": "image"},
+                        {"type": "image", "content": "path/to/image1.png"},
```
RdoubleA, line 93: this needs to be PIL image

torchtune/data/_utils.py
```diff
@@ -63,22 +100,41 @@ def split_text_by_image_tag(content: str, image_tag: str) -> List[Dict[str, str]
...
-        List[Dict[str, str]]: list of dictionaries to be used in the ``Message`` content field
+    Raises:
+        ValueError: If the number of images does not match the number of image tags in the content
 
     Example:
-        >>> content = split_text_by_image_tag("<image>hello <image>world", "<image>")
+        >>> content = format_content_with_media(
+        ...     "<|image|>hello <|image|>world",
+        ...     image_tag="<|image|>",
+        ...     images=[<PIL.Image.Image>, <PIL.Image.Image>"]
```
RdoubleA, line 112:
```suggestion
        ...     images=[<PIL.Image.Image>, <PIL.Image.Image>]
```

torchtune/data/_utils.py
```diff
@@ -42,7 +43,43 @@ def truncate(
...
+    # Open the local image as a PIL image
+    try:
+        image = Image.open(image_path)
+    except Exception as e:
+        raise ValueError(f"Failed to open image as PIL Image from {image_path}") from e
+
+    return image
+
+
+def format_content_with_images(
```
RdoubleA, line 80: nit: maybe the name `insert_images_in_content` is more clear on what the function is doing

torchtune/data/_utils.py
```diff
@@ -42,7 +43,43 @@ def truncate(
...
+    content: str, *, image_tag: str, images: List["PIL.Image.Image"]
+) -> List[Dict[str, str]]:
     """
     Given a raw text string, split by the specified ``image_tag``
     and form into list of dictionaries to be used in the ``Message`` content
```
RdoubleA, line 85: nit: while you're here, could you describe what happens if image tag is not found in the content string? IIRC, it's just a no-op

torchtune/datasets/multimodal/_llava_instruct.py
```diff
@@ -54,6 +55,8 @@ class LlavaInstructToMessages(Transform):
         new_system_prompt (Optional[str]): if specified, prepend a system message. This can
             serve as instructions to guide the model response. Setting this will OVERRIDE any system
             messages already present in the dataset. Default is None.
+        images_dir (str): path to the directory containing the images. User is expected to download the COCO dataset.
```
RdoubleA, line 58: not sure if we should have a default. this should be explicitly defined by the user, otherwise they may use the builder/transform and forget to specify this and get confused when it break

torchtune/datasets/multimodal/_llava_instruct.py
```diff
@@ -89,22 +94,34 @@ def __call__(self, sample: Mapping[str, Any]) -> Mapping[str, Any]:
...
             role = role_map[message["from"]]
+            content = message["value"]
             if role == "system" and self.new_system_prompt is not None:
                 continue
-            content = split_text_by_image_tag(message["value"], "<image>")
+            if role == "user":
+                pil_image = load_image(
+                    self.images_dir + sample[self._column_map["image"]]
+                )
+                content = format_content_with_images(
```
RdoubleA, line 108: this reads really nicely now

torchtune/data/_utils.py
```diff
@@ -42,7 +43,43 @@ def truncate(
...
+
+    Raises:
+        ValueError: If the image cannot be loaded from remote source.
+        ValueError: If the image cannot be opened as a PIL.Image.
+
+    Returns:
+        PIL.Image.Image: The loaded image.
+    """
+    # Hackily import PIL to avoid burdensome import in the main module
+    # TODO: Fix this
```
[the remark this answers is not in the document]
joecummings: It's only prevalent if doing MM. Right now, you can completely run all our text stuff without the need to import torchvision, PIL, etc...

I'd love to keep it that way.

torchtune/data/_utils.py
joecummings: True, but it also does the splitting?

RdoubleA: splitting so we can insert the images in the correct location :) but, your call

torchtune/data/_utils.py
joecummings: <tiny>u need to be a pil image</tiny>😡

torchtune/data/_utils.py
joecummings: So, if image_tag and images are both none, it's a no-op. But otherwise, it'll error b/c if there's images, it expects an image_tag.

torchtune/datasets/multimodal/_llava_instruct.py
joecummings: I like the default for something like this where it's for a very specific dataset.

RdoubleA (approved):
heroic effort, thanks for this. I have no other concerns

pbontrager (approved):
I left one comment on a potential change but otherwise it looks good to go.

torchtune/data/_messages.py
```diff
@@ -48,28 +46,40 @@ class Message:
...
 
+    def _convert_to_list_of_dict(self, content) -> List[Dict[str, Any]]:
+        """User is currently allowed to pass in a string for text-only content.
+        This ensures that the content is formatted as a list of dictionaries."""
+        if isinstance(content, str):
+            return [{"type": "text", "content": content}]
+
+        assert isinstance(
+            content, list
+        ), f"content must be of type List[Dict[str, Any]], got {content}"
```
ebsmothers, line 79: I know we don't wanna do it here, but is there somewhere that we can/should validate that we have a PIL image when type="image"?

torchtune/datasets/multimodal/_llava_instruct.py
```diff
@@ -89,22 +94,35 @@ def __call__(self, sample: Mapping[str, Any]) -> Mapping[str, Any]:
...
+
+        # Add in image stuffs / load from file
         for message in sample[self._column_map["conversations"]]:
             role = role_map[message["from"]]
+            content = message["value"]
             if role == "system" and self.new_system_prompt is not None:
                 continue
-            content = split_text_by_image_tag(message["value"], "<image>")
+            if role == "user":
+                image_path = sample[self._column_map["image"]]
```
ebsmothers, line 105: So this will be present in every sample in the dataset?

torchtune/data/_utils.py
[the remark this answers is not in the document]
ebsmothers: What's the fix here? Honestly I think this is an OK solution

torchtune/data/_utils.py
```diff
@@ -63,22 +113,41 @@ def split_text_by_image_tag(content: str, image_tag: str) -> List[Dict[str, str]
...
-    Returns:
-        List[Dict[str, str]]: list of dictionaries to be used in the ``Message`` content field
-
-    Example:
-        >>> content = split_text_by_image_tag("<image>hello <image>world", "<image>")
+    Raises:
+        ValueError: If the number of images does not match the number of image tags in the content
+
+    Examples:
+        >>> content = format_content_with_media(
```
ebsmothers, line 122:
```suggestion
        >>> content = format_content_with_images(
```

torchtune/data/_utils.py
```diff
@@ -42,18 +44,66 @@ def truncate(
...
+        try:
+            image_loc = request.urlopen(image_loc)
+        except Exception as e:
+            raise ValueError(f"Failed to load image from {image_loc}") from e
+
+    # Open the local image as a PIL image
+    try:
+        image = Image.open(image_loc)
+    except Exception as e:
+        raise ValueError(f"Failed to open image as PIL Image from {image_loc}") from e
```
ebsmothers, line 88: nit: are these really ValueErrors? (Or are you doing the thing I also do and lazily using `ValueError` as a catch-all for things that are not actually ValueErrors)

torchtune/data/_utils.py
```diff
@@ -42,18 +44,66 @@ def truncate(
...
+
+    Returns:
+        PIL.Image.Image: The loaded image.
+    """
+    # Hackily import PIL to avoid burdensome import in the main module
+    # TODO: Fix this
+    from PIL import Image
+
+    # If pointing to remote source, try to load to local
+    if isinstance(image_loc, str) and image_loc.startswith("http"):
```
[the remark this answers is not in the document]
joecummings: How? I can parse it, but urllib.parse parses it even if it's not a valid URL. 

Limiting to http and https is restricting, sure, but we can relax when needed.

torchtune/data/_messages.py
joecummings: validate_messages?

torchtune/datasets/multimodal/_llava_instruct.py
joecummings: yup

torchtune/data/_utils.py
joecummings: WRT the first one, no it could be several different errors. Trying to call a remote URL always has plenty of exception types. I could just leave the error as is and let the user deal with it, but that seems messy.

WRT the second value error, more often than not, it will be a ValueError as the provided value is unable to be opened by the PIL Image interface. 

Do you want me to remove the exceptions with error messages and just let it ride all the way to the user?

torchtune/data/_utils.py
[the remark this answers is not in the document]
joecummings: Would be putting multimodal stuff in a subfolder likely.

torchtune/data/_messages.py
joecummings: Argh, I cannot do this without a nested import of PIL.

torchtune/data/_utils.py
ebsmothers: OK thanks for the explanation, in that case I think what you have is fine

torchtune/data/_messages.py
ebsmothers: OK fine to leave it as is then

ebsmothers (approved):

Issue #676, opened 2024-04-10
Install gcc >= 9 to support torch.compile testing with inductor backend


rohan-varma:
- CI workers need to install gcc >= 9 as torch.compile needs at least this GCC to compile w/inductor backend. We currently work around this by setting an env var to tell torch compile to use `aot_eager` for its backend, but users will use the default inductor backend generally so we should test this.

ebsmothers:
This should've been closed after #1522 landed

PR #1508, opened 2024-09-06
[testing] remove compile backend override from recipe test

ebsmothers:
Check if this breaks due to gcc version issue as described in #676 

On CPU: same compile error with or without installing dev-toolset 10 (see [here](https://github.com/pytorch/torchtune/actions/runs/10732174739?pr=1508))

On GPU runner:
            runs-on: linux.g5.12xlarge.nvidia.gpu
            gpu-arch-type: "cuda"
            gpu-arch-version: "12.1"
Same error on both nightly and stable versions (both with dev-toolset 10 installed)


ebsmothers:
Closing this after #1522

PR #1515, opened 2024-09-06
utils refactor clean-up

RdoubleA:
#### Context
- [ ] add a new feature
- [x] fix a bug
- [ ] other (please add here)

Few follow-up items after #1421 was completed. Some helpers were in the wrong locatino to avoid circular dependencies. We moved those back and remove some pesky dependencies between modules that pollute our dependency graph.
- Removes the config <> training dependency. config should not depend on anything in training, only recipes should depend on training. Instead of checking for rank zero without `log_config`, do it in distribtued recipes instead.
- Moves `_get_prompt_template` to `data/` so `config/` does not depend on any other directory other than `utils/`
- Moves `deprecated` from `data/` to `utils/`, a more appropriate location.

#### Changelog

#### Test plan

#### UX


torchtune/config/_utils.py
```diff
@@ -12,28 +12,17 @@
...
     """
-    Logs the resolved config (merged YAML file and CLI overrides) to rank zero.
+    Logs the resolved config (merged YAML file and CLI overrides).
 
     Args:
         recipe_name (str): name of the recipe to display
         cfg (DictConfig): parsed config object
     """
-    # Log the config only on rank 0
-    _, rank = get_world_size_and_rank()
```
ebsmothers, line 33: rank = torch.distributed.get_rank() if torch.distributed.is_available() and torch.distributed.is_initialized() else 0

felipemello1 (approved):

recipes/dev/lora_finetune_fsdp2.py
```diff
@@ -718,9 +718,7 @@ def recipe_main(cfg: DictConfig) -> None:
     os.environ["TORCH_NCCL_AVOID_RECORD_STREAMS"] = "1"
     init_process_group(backend="gloo" if cfg.device == "cpu" else "nccl")
 
-    _, rank = training.get_world_size_and_rank()
-    if rank == 0:
-        config.log_config(recipe_name="LoRAFinetuneRecipeDistributed", cfg=cfg)
+    config.log_config(recipe_name="LoRAFinetuneRecipeDistributed", cfg=cfg)
```
felipemello1: i think you missed adding the rank here

tests/torchtune/config/test_config_utils.py
```diff
@@ -128,12 +128,37 @@ def test_log_config(self, capsys):
...
+                )
+
+            # Clear the stream
+            stream.truncate(0)
+            stream.seek(0)
+
+            # Make sure all other ranks do not log anything
+            with mock.patch(
+                "torchtune.config._utils.dist.get_rank",
+                return_value=1,
```
felipemello1, line 157: i thought that you removed it because logging doesnt check for rank anymore

ebsmothers (approved):

PR #1519, opened 2024-09-06
no new cycles

RdoubleA:
no new cycles, no new cycles no no no

Removes the last dependency on training from other modules. See #1421 

## Test plan
o______________o


ebsmothers (approved):
pls god no more

PR #1522, opened 2024-09-07
Make our compile tests actually work

ebsmothers:
Make our recipe tests with `compile=True` not a no-op. Even better yet make them pass. I don't believe that #676 is relevant anymore so we can close that once this lands.

The one problem is that this still would not have caught #1498, since the full finetune test case does still pass (it's possible there was something specifically about the fused optimizer, cause otherwise the run is fairly similar to our test configs). There is also the issue of the compiled version of LoRA not failing for 10+ iterations; this also won't be caught by our current testing. To deal with this we probably need longer-running compile tests (ideally with perf numbers) as part of our (currently very limited) regression test suite. We will file a testing mega-issue soon and this will be included.

Test plan:

```
======== 37 passed, 467 deselected, 8 warnings in 619.62s (0:10:19) =========
```

Also added an assertion to verify that all recipe tests in CI are now using inductor backend (see [this commit](https://github.com/pytorch/torchtune/pull/1522/commits/d97de19118806c398b3894f5e8b2dbeddc3eabad)). As demonstrated in the below screenshot, the CI is green on this commit, which demonstrates that our CI is now running on the inductor backend as it should.

Screenshot 2024-09-07 at 3 23 03 PM

### DIFF
--- a/tests/recipes/test_full_finetune_single_device.py
+++ b/tests/recipes/test_full_finetune_single_device.py
@@ -74,9 +74,6 @@ class TestFullFinetuneSingleDeviceRecipe:
         ckpt_dir = ckpt_path.parent
         log_file = gen_log_file_name(tmpdir)
 
-        # To workaround https://github.com/pytorch/torchtune/issues/676
-        if compile:
-            os.environ["TORCH_COMPILE_BACKEND"] = "aot_eager"
         cmd = f"""
         tune run full_finetune_single_device \
             --config {config} \
@@ -99,8 +96,13 @@ class TestFullFinetuneSingleDeviceRecipe:
         with pytest.raises(SystemExit, match=""):
             runpy.run_path(TUNE_PATH, run_name="__main__")
 
+        # Make sure to clear compile state in between tests
+        if compile:
+            torch._dynamo.reset()
+
         loss_values = get_loss_values_from_metric_logger(log_file)
         expected_loss_values = self._fetch_expected_loss_values(model_type)
+
         torch.testing.assert_close(
             loss_values, expected_loss_values, rtol=1e-4, atol=1e-4
         )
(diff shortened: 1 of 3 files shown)


tests/recipes/test_full_finetune_single_device.py
```diff
@@ -74,9 +74,6 @@ def test_loss(self, compile, config, model_type, ckpt_type, tmpdir, monkeypatch)
         ckpt_dir = ckpt_path.parent
         log_file = gen_log_file_name(tmpdir)
 
-        # To workaround https://github.com/pytorch/torchtune/issues/676
-        if compile:
-            os.environ["TORCH_COMPILE_BACKEND"] = "aot_eager"
```
felipemello1, line 79: do we enforce inductor as our backend anywhere? if not, should we?

felipemello1: I jsut saw that you tested it with a commit, but maybe we could enforce it here to use with inductor backend? Or add inductor to some defaults.py, and always use the default_compile_backend if defaults.py?

felipemello1 (approved):

tests/recipes/test_full_finetune_single_device.py
ebsmothers: > do we enforce inductor as our backend anywhere? if not, should we?

We may still wanna allow users the flexibility to change it, but in my other PR I added inductor as the default in our compile utilities [here](https://github.com/pytorch/torchtune/blob/c169bcdddaeee94fb4b599c1f083f94ee1a76143/torchtune/training/_compile.py#L38). 

> I jsut saw that you tested it with a commit, but maybe we could enforce it here to use with inductor backend? Or add inductor to some defaults.py, and always use the default_compile_backend if defaults.py? Just trying to prevent some silent error in case, for some reasons, the default of the CI machine is not inductor

I'm ambivalent about putting it in the actual test. Strictly speaking it may be safest to explicitly set it, but in order for this kind of silent error we would need (a) to remove (or change) the default in the compile utils and (b) the CI runner to have its own custom environment variables defined somehow. Based on that it's low-risk imo, but lmk if you disagree and I can add it

PR #1527, opened 2024-09-09
WIP --- [memory] replace _copy with nn.ModuleList

felipemello1:
#### Context
- [ ] add a new feature
- [x] fix a bug
- [ ] other (please add here)

TLDR: this PR saves 2GiB for llama 3_1 by not having 32 (num_layers) identical copies of the positional embedding.

Currently all of our model builders pass a single Layer to the TransformerDecoder, which is copied `num_layers` times. Inside of the layer, we have the buffer for RoPE embeddings.

How much memory?
The cache for llama 3_1 is size (max_seq_len, att_head_dim//2, 2) --> (131000, 64, 2)
It is stored in fp32
There are 32 layers

total memory: (131000 * 64* 2) * (4) * (32) / 1024 / 1024 / 1024 = 2GiB

if we have only a single cache, we dont need to multiple by 32, and it only takes 0.12GiB, which is what the profiling shows

image

#### Changelog
- Transformer accept module list
- Update in all configs to receive a module list
- Replace GemmaTransformerDecoder with TiedTransformerDecoder ---> WHY ARE WE DOING IT IN THE SAME PR? because i will have to test it, and i didnt want to test twice :/

#### Test plan
ran at least one config of each model on at least one recipe (lora, full, qat. Didnt include rlhf). The curves are main vs this PR, using the same command.

base config
```
tune run --nproc_per_node 4 full_finetune_distributed --config gemma/2B_full \
fsdp_cpu_offload=False \
compile=True \
dataset.packed=False \
dataset.split=train \
tokenizer.max_seq_len=4096 \
metric_logger=torchtune.training.metric_logging.WandBLogger \
metric_logger.project=profiling \
metric_logger.tags=[my_experiment_name] \
log_every_n_steps=1 \
log_peak_memory_stats=True \
max_steps_per_epoch=30 \
epochs=1 \
batch_size=4 \
optimizer_in_bwd=True
```

"config_name": "llama3_1/8B_full_single_device",
"recipe_type": "full_finetune_single_device",
image

"config_name": "llama3/8B_qlora_single_device",
"recipe_type": "lora_finetune_single_device",
image

"config_name": "llama2/7B_lora",
"recipe_type": "lora_finetune_distributed",
image

"config_name": "gemma/2B_qlora_single_device",
"recipe_type": "lora_finetune_single_device",
image

"config_name": "gemma/2B_full",
"recipe_type": "full_finetune_distributed",
image

"config_name": "llama3/8B_qat_full",
"recipe_type": "qat_distributed",
image

"config_name": "qwen2/1.5B_lora_single_device",
"recipe_type": "lora_finetune_single_device",
image

"config_name": "qwen2/1.5B_lora",
"recipe_type": "lora_finetune_distributed",
image

"config_name": "mistral/7B_lora",
"recipe_type": "lora_finetune_distributed",
image


felipemello1:
@pbontrager , i didnt make the change for flamingo, but we should replace _copy there with ModuleList, IMO

felipemello1:
@joecummings @ebsmothers , since we are saving 2GiB, and we are not upcasting the rope embeddings anywhere, i am not sure if the positional embedding is being done with float32. We should test it.

torchtune/models/gemma/_component_builders.py
```diff
@@ -12,12 +12,14 @@
...
     MultiHeadAttention,
     FeedForward,
+    TiedLinear,
     FrozenNF4Linear,
     RotaryPositionalEmbeddings,
     TransformerSelfAttentionLayer,
 )
 from torchtune.models.gemma.rms_norm import GemmaRMSNorm
-from torchtune.models.gemma.transformer import GemmaTransformerDecoder
+from torchtune.models.gemma.gemma_norm_embedding import GemmaNormEmbeddings
```
felipemello1, line 21: we put together the embeddings + norm that gemma requires, so that we can delete the GemmaTransformerDecoder

PR #1529, opened 2024-09-10
[RFC] Dataset documentation overhaul

RdoubleA:
Yet another documentation update for datasets!

There are A LOT of new concepts and modules that we've recently added that are COMPLETELY ABSENT from our documentation (excluding API ref). Here are some of the recent changes that are not mentioned anywhere:
- Prompt templating happens in the tokenizer
- Setting max sequence length happens in the tokenizer
- Hey we have sample packing, how do I use it (this is kinda mentioned but very briefly)
- What is a Message
- What is a message transform and how do I use it
- What is a model transform and how do I use it
- What is SFTDataset
- How do I use PreferenceDataset
- How do I use TextCompletionDataset
- How do I handle images in my custom dataset
- What is the difference base tokenizer and a model tokenizer
- How do I create a custom model tokenizer
- How do I customize a tokenizer's special tokens

...the list goes on. Here, I put up a rough outline of where these concepts should be addressed, so I'm looking for feedback on:
- How/where these topics should go and be organized
- Anything else that I'm missing

I've created a new General Usage category in our sidebar. The intention is for pages under here to serve as a more detailed, casual discussion of our APIs and general concepts. API ref remains for concrete parameter descriptions, tutorials are for e2e workflows, and deep dives are for highly technical investigation into a singular concept. Some of these may belong in tutorials or deep dives, but I would like to push for something in between raw API ref and tutorials that is easy to quickly reference and search.

image

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -137,6 +137,24 @@ torchtune tutorials.
 .. toctree::
    :glob:
    :maxdepth: 1
+   :caption: General Usage
+   :hidden:
+
+   usage/tokenizers
+   usage/messages
+   usage/sample_packing
+   usage/model_transforms
+   usage/message_transforms
+   usage/instruct_datasets
+   usage/chat_datasets
+   usage/multimodal_datasets
+   usage/preference_datasets
+   usage/text_completion_datasets
+   usage/custom_datasets
+
+.. toctree::
+   :glob:
+   :maxdepth: 1
    :caption: API Reference
    :hidden:
 
--- a/docs/source/usage/message_transforms.rst
+++ b/docs/source/usage/message_transforms.rst
@@ -1,0 +1,30 @@
+.. _message_transforms_usage_label:
+
+==================
+Message Transforms
+==================
+
+Discuss the concept of message transforms and how they should be used. This is the core piece
+that a user needs to be most familiar with to fine-tune on custom data.
+
+Common message transforms
+-------------------------
+
+Overview of common dataset formats (like ShareGPT, OpenAI, Instruct data) and the message
+transforms to use for these.
+
+Converting text-data to Messages
+--------------------------------
+
+Show how example text-only data should be converted into a list of Messages.
+
+
+Converting text and image data to Messages
+------------------------------------------
+
+Show how example text and image data should be converted into a list of Messages.
+
+Creating a custom message transform
+-----------------------------------
+
+Walk through how one would design their own message transform and principles they should follow.
--- a/docs/source/usage/chat_datasets.rst
+++ b/docs/source/usage/chat_datasets.rst
@@ -1,0 +1,22 @@
+.. _chat_dataset_usage_label:
+
+=============
+Chat Datasets
+=============
+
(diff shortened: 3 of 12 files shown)


ebsmothers (commented):
Overall I think this looks good. A few miscellaneous comments in no particular order:

(1) We should consider giving a separate section for multimodal. There is already some soft division here (multimodal_datasets and model_transforms), but from the way the pages are structured it's not clear. I think it'll make it easier for people to find what they want if we have a clear "multimodal" section.

(2) One thing we want to emphasize in a fairly visible place is that torchtune has its own intermediate representation which allows for pretty general support of tokenization and image transforms as part of the dataset. So users need to understand that all dataset flow happens in basically two stages: (1) get data from the raw format into Message format, and (2) once in Message format, tokenize and apply any other transformations to return a dictionary the model can parse. Imo this kinda motivates all the other lower-level details about model transforms, message transforms, etc.

(3) Kinda related to (2), but the majority of our datasets used will be instruct, chat, or multimodal, all of which follow the SFTDataset flow. It might be good to outline the "general" flow in a torchtune dataset somewhere in a nice diagram to give a general mental model of how everything needs to fit together. Basically I would like to see an overview page "understanding datasets in torchtune" that's some combo if (2) and (3). (Can add a caveat on packed/preference/completion datasets and link to separate pages)

(4) Do we wanna add an item on collators? I feel like this is one area of potential customization that people get tripped up on as well.

(5) I still feel like model transform + tokenizer is a potential area of confusion. I think splitting multimodal out may help some here, but even so there are a lot of layers (base tokenizer, model tokenizer, builder). I would lead that page with a very clear "if you want to do X, see this section" framing. Otherwise it can get overwhelming quickly.

docs/source/index.rst
```diff
@@ -134,6 +134,24 @@ torchtune tutorials.
    deep_dives/recipe_deepdive
    deep_dives/wandb_logging
 
+.. toctree::
+   :glob:
+   :maxdepth: 1
+   :caption: General Usage
```
svekars, line 140: Thinking there could be a better caption than "General Usage". Some otions:
* Concepts and Operations
* Torchtune Basics
* User Guide
* HowTos 

Also I think it should probably go after Finetuning recipes?

docs/source/usage/message_transforms.rst
```diff
@@ -0,0 +1,30 @@
+.. _message_transforms_usage_label:
+
+==================
```
svekars, line 3: The whole PR looks like a good start. One thing I've noticed is that you have a lot of headings in each file. Make sure you have enough content under each of those or remove them and just right paragraphs of text conveying the idea you've expressed in the heading.

svekars (commented):
It looks like a good start.

docs/source/index.rst
RdoubleA: Yes I like Torchtune Basics or just Basics. and that location sounds better

docs/source/usage/message_transforms.rst
RdoubleA: definitely - will keep this in mind as I fill out each section

PR #1530, opened 2024-09-10
[Draft]Support Optimizer-in-the-backward

mori360:
#### Context
- [x] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

#### Changelog
In FullFinetuneDistributed, switch self._optimizer to self._optimizer_in_bwd that runs in backwards
Running in the backwards could save the peak memory cost during `loss.backward()`
By the local testing with llama2/7B_full, peak memory is optimized from 31.5GB to 21.2GB, saves 32.6%

#### Test plan

#### UX


recipes/full_finetune_distributed.py
```diff
@@ -670,7 +714,13 @@ def train(self) -> None:
                         time_per_step = time.perf_counter() - t0
                         log_dict = {
                             "loss": loss_to_log,
-                            "lr": self._optimizer.param_groups[0]["lr"],
+                            "lr": (
```
mori360, line 717: "lr" here are all the same

PR #1539, opened 2024-09-11
Add single device KD recipe

lindawangg:
#### Context
- [x] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

#### Changelog
* KD recipe (`knowledge_distillation_single_device.py`) is similar to `lora_finetune_single_device.py`. Main differences are:
  * adds kd loss, currently just `ForwardKLLoss` (in `kd_losses.py`), to CE loss
  * adds teacher model inference to get logits
* KD config: `knowledge_distillation_single_device.yaml`

#### Test plan

```
CUDA_VISIBLE_DEVICES=0 tune run knowledge_distillation_single_device --config qwen2/knowledge_distillation_single_device
```
##### Llama3.1 KD Training
Legend: KD Llama3.1 student (blue), LoRA Llama3.1 student (orange)
imageimageimage

##### Qwen2 KD Training
Legend: KD Qwen2 0.5B student (green), LoRA Qwen2 0.5B student (grey), LoRA Qwen2 1.5B teacher (blue)
imageimageimage

##### Llama3.1 Eval Results
image

##### Qwen2 Eval Results
image

#### UX


recipes/configs/llama3_1/kd_single_device.yaml
```diff
@@ -0,0 +1,110 @@
+# Config for single device knowledge distillation in kd_single_device.py
+# using a teacher and student model
+#
+# This config assumes that you've run the following commands before launching
+# this run:
+#   tune download meta-llama/Meta-Llama-3.1-8B-Instruct --output-dir /tmp/Meta-Llama-3.1-8B-Instruct
+#   tune run lora_finetune_single_device --config llama3_1/8B_lora_single_device
```
ebsmothers, line 7: Can you add a bit more detail here? Tbh when I first looked at it I thought you had accidentally just copy-pasted a `tune run` command from another config 😅 . Maybe just add a couple statements explicitly saying something like "Run this to download the model: {tune download ...}. You will then need to fine-tune the teacher model, you can do this with {tune run...}"

recipes/configs/llama3_1/kd_single_device.yaml
```diff
@@ -0,0 +1,110 @@
...
+    consolidated.00.pth
+  ]
+  recipe_checkpoint: null
+  output_dir: /tmp/kd_model/
+  model_type: LLAMA3
+
+# Teacher checkpoint
+teacher_checkpointer:
+  _component_: torchtune.training.FullModelMetaCheckpointer
+  checkpoint_dir: /tmp/Meta-Llama-3.1-8B-Instruct/lora_finetuned_single_device_epoch_1/
```
ebsmothers, line 59: Should make sure that this directory matches the output of whatever command you give for LoRA finetuning at the top of this file

recipes/configs/llama3_1/kd_single_device.yaml
```diff
@@ -0,0 +1,110 @@
...
+
+kd_loss:
+  _component_: torchtune.modules.loss.ForwardKLWithChunkedOutputLoss
+kd_ratio: 0.5
+
+# Training
+epochs: 1
+max_steps_per_epoch: null
+gradient_accumulation_steps: 16
+compile: False
```
ebsmothers, line 97: Out of curiosity, did you try with compile yet?

torchtune/modules/loss/kd_losses.py
```diff
@@ -0,0 +1,119 @@
...
+
+import torch
+import torch.nn.functional as F
+
+
+class ForwardKLLoss(torch.nn.Module):
+    """
+    The Kullback-Leibler divergence loss for valid indexes.
+    Implementation of https://github.com/jongwooko/distillm/blob/master/distillm/losses.py.
+    """
```
ebsmothers, line 17: I'm surprised the linter didn't yell at you for this, can we add args (well I guess just single arg) with typehints here?

ebsmothers: Also two nits on the link: (1) don't include the period at the end (makes it not clickable), and (2) replace master with a specific commit hash (in case things change in the future)

torchtune/modules/loss/kd_losses.py
```diff
@@ -0,0 +1,119 @@
...
+                (batch_size*num_tokens, vocab_size).
+            labels (torch.Tensor): Ground truth labels of shape
+                (batch_size, vocab_size).
+
+        Returns:
+            torch.Tensor: KL divergence loss of shape (1,).
+        """
+
+        teacher_prob = F.softmax(teacher_logits, dim=-1, dtype=torch.float32)
+        inf_mask = torch.isinf(student_logits)
```
ebsmothers, line 43: Noob q: why would student logits be infinite? Does it mean there is some numerical issue? (I know it's in the original implementation, just curious about the rationale)

torchtune/modules/loss/kd_losses.py
```diff
@@ -0,0 +1,119 @@
...
+class ForwardKLWithChunkedOutputLoss(torch.nn.Module):
+    """
+    Forward KL with chunked outputs that saves memory by only upcasting one chunk at a time.
+
+    Since the model is trained with bf16, before computing KL divergence, we have to upcast
+    it to fp32 for better accuracy and stability. When upcasting happens, the memory usage doubles.
+    Models like llama3 have large vocabulary size and, therefore, have a large output
+    result (bsz, num_tokens, vocab_size). If we chunk on the token level, you can still compute
+    the cross entropy normally, but upcasting only one chunk at a time saves considerable memory.
+    """
```
ebsmothers, line 62: Similar comment here about init args

tests/torchtune/modules/loss/test_kd_losses.py
```diff
@@ -0,0 +1,56 @@
...
+
+        # vanilla FKL
+        fkl_loss = ForwardKLLoss(ignore_index=ignore_index)
+        logits = logits.reshape(-1, logits.size(-1))
+        teacher_logits = teacher_logits.reshape(-1, teacher_logits.size(-1))
+        labels = labels.reshape(-1)
+        standard_loss = fkl_loss(logits, teacher_logits, labels)
+
+        # Assert
+        assert_expected(chunked_loss, standard_loss, rtol=1e-2, atol=1e-2)
```
ebsmothers, line 56: Would also be good to run the loss from jongwooko repo with identical sets of values, use that to determine the expected value, then compare both chunked and standard losses to that (that way we know that we have numerical parity with a reference implementation)

recipes/configs/llama3_1/kd_single_device.yaml
lindawangg: Yea tested that compile works (in pink)
image

tests/torchtune/modules/loss/test_kd_losses.py
lindawangg: Good idea. I couldn't find any tests in the repo, so I randomly generated the logits and ran it through the distillm implementation.

torchtune/modules/loss/kd_losses.py
lindawangg: The teacher_logits could be infinite too. I believe the original implementation only considered that the student logits could be infinite because that's the model that's training. The `inf` in the student logits would cause the torch.sum part to be `inf`.

recipes/kd_single_device.py
```diff
@@ -0,0 +1,804 @@
...
+                # For CEWithChunkedOutputLoss, if we compile the entire class
+                # we lose the benefits from the chunked loss.
+                # Therefore, we only compile the cross entropy function + upcasting
+                self._loss_fn.compute_cross_entropy = torch.compile(
+                    self._loss_fn.compute_cross_entropy, backend=backend
+                )
+        else:
+            if self._model_compile:
+                log.info("Compiling loss with torch.compile...")
+                self._loss_fn = torch.compile(self._loss_fn, backend=backend)
```
ebsmothers, line 283: Oh we have some utilities for this now, you can try to use those instead. See usage in our LoRA single-device recipe [here](https://github.com/pytorch/torchtune/blob/4fbe7b2d4956b3790c51d7a255c0040cf5c38fad/recipes/lora_finetune_single_device.py#L247-L253). (If they don't work for KD out of the box lmk, we can refactor as needed)

recipes/kd_single_device.py
```diff
@@ -0,0 +1,804 @@
...
+        self.adapter_params = get_adapter_params(model)
+        self._is_dora = any(["magnitude" in k for k in self.adapter_params.keys()])
+        set_trainable_params(model, self.adapter_params)
+
+        if compile_model:
+            log.info("Compiling model layers with torch.compile...")
+            backend = os.environ.get("TORCH_COMPILE_BACKEND", "inductor")
+            for m in reversed(list(model.modules())):
+                if isinstance(m, modules.transformer.TransformerSelfAttentionLayer):
+                    m.compile(backend=backend)
```
ebsmothers, line 420: Similar comment here about using the compile utilities

recipes/kd_single_device.py
```diff
@@ -0,0 +1,804 @@
...
+        )
+
+        log.info(f"Model is initialized with precision {self._dtype}.")
+
+        if self._device.type == "cuda":
+            memory_stats = training.get_memory_stats(device=self._device)
+            training.log_memory_stats(memory_stats)
+        return model
+
+    def _setup_teacher_model(
```
ebsmothers, line 463: Wonder if it's worth it to also compile the teacher model when compile=True?

recipes/kd_single_device.py
ebsmothers: Can also try compiling KD loss function

recipes/kd_single_device.py
```diff
@@ -0,0 +1,804 @@
...
+                                self._model.parameters(),
+                                max_norm=float(self._clip_grad_norm),
+                            )
+                        self._optimizer.step()
+                        self._optimizer.zero_grad(set_to_none=True)
+                        self._lr_scheduler.step()
+                        # Update the number of steps when the weights are updated
+                        self.global_step += 1
+
+                        loss_to_log = running_loss.item()
```
ebsmothers, line 727: Nit: I'm not positive, but it might be slightly slower to do things this way. Each `.item()` will cause a sync and you could probably get away with calling just `running_class_loss.item()` and `running_kd_loss.item()` and figuring out `loss_to_log` from that (I think).

ebsmothers (commented):
OK I took another, more thorough pass, and this looks great! I left a handful more comments but after that there are no real concerns from me

recipes/kd_single_device.py
lindawangg: I got an error when trying to compile the teacher model. I'm using torch.no_grad when inferencing the teacher model to save memory consumption. However, it seems that torch.no_grad isn't compatible with torch.compile (https://github.com/pytorch/pytorch/issues/100241).

recipes/kd_single_device.py
lindawangg: Calling .item() twice instead of 3x helps slightly. I don't see much of a difference: 
image

I removed `running_loss` and added computing from class and kd loss.

recipes/configs/llama3_1/kd_single_device.yaml
lindawangg: The loss curves between compile and non-compile are fairly similar.
imageimage

I do see some non-determinism during eval. The losses are all around 1.2, but there's slight differences in eval. Also interesting that fine-tuning on alpaca dataset actually hurts performances on all benchmarks but Truthful QA.
image

joecummings (commented):
Super nit / discussion point:

kd -> distillation. I think it's much easier to understand at a glance what it is. I didn't understand the abbreviation KD until I read through the PR description.

lindawangg:
> Super nit / discussion point:
> 
> kd -> distillation. I think it's much easier to understand at a glance what it is. I didn't understand the abbreviation KD until I read through the PR description.

We could rename from `kd_single_device` to `knowledge_distillation_single_device`. Distillation might be confusing since there's many types.

ebsmothers:
> We could rename from kd_single_device to knowledge_distillation_single_device. Distillation might be confusing since there's many types.

Yeah this sounds good to me. I agree that it's nice to just be explicit in the recipe name so it's obvious what the recipe is doing (even if the name is a bit longer as a result).

tests/recipes/test_kd_single_device.py
```diff
@@ -0,0 +1,308 @@
...
+            teacher_checkpointer.output_dir={tmpdir} \
+            teacher_checkpointer.model_type={model_type.upper()} \
+            tokenizer._component_=torchtune.models.llama3.llama3_tokenizer \
+            tokenizer.path='{tokenizer_path}' \
+            tokenizer.prompt_template=null \
+            ~tokenizer.merges_file \
+            metric_logger._component_=torchtune.training.metric_logging.DiskLogger \
+            metric_logger.filename={log_file} \
+            compile={compile} \
+            kd_loss._component_=torchtune.modules.loss.ForwardKLWithChunkedOutputLoss \
```
ebsmothers, line 91: Do you actually need this override? This should be the default in the config, right?

lindawangg: Removed.

ebsmothers (approved):
Thank you for adding this!

PR #1544, opened 2024-09-11
fix 3.1 rope init for compile

ebsmothers:
Currently our Llama 3.1 doesn't work with compile because we init the RoPE cache in the first forward. 

For our single-device recipe we can init the cache as part of `__init__` no problem, but for our distributed recipe we load on meta device. In that case we cannot init the cache in `__init__` because the RoPE scaling factors require data to exist in intermediate tensors (which they won't on meta device). So for our distributed recipes currently initialize RoPE cache directly in the recipe [here](https://github.com/pytorch/torchtune/blob/b9f2cafa70898888b72f839f1eec0ae37660a0a0/recipes/lora_finetune_distributed.py#L461-L462) after sharding the model.

So we just need a way to skip calling `rope_init` from a meta device context. Actually we've solved this problem hackily before, see e.g. [here](https://github.com/pytorch/torchtune/blob/44d829e3a28a7e663a3089524afae3d8d1946c63/torchtune/modules/peft/lora.py#L92-L93). Well this change is basically identical to that, but this time `rope_init` always gets called. Then once the frequency is constructed, we break out early if it's on meta device and rely on the recipe to manually call after FSDP sharding.

## Test plan

Apart from green CI, checking that loss curves are the same on a few different configs..

### Single-device LoRA 8B
Screenshot 2024-09-11 at 1 56 51 PM

### Distributed LoRA 8B
Screenshot 2024-09-11 at 1 56 34 PM

### Single-device FFT 8B
Screenshot 2024-09-11 at 1 56 11 PM

### DIFF
--- a/torchtune/models/llama3_1/_position_embeddings.py
+++ b/torchtune/models/llama3_1/_position_embeddings.py
@@ -56,6 +56,7 @@ class Llama3ScaledRoPE(nn.Module):
         self.low_freq_factor = low_freq_factor
         self.high_freq_factor = high_freq_factor
         self.old_context_len = old_context_len
+        self.is_cache_built = False
         self.rope_init()
 
     def rope_init(self):
@@ -82,6 +83,7 @@ class Llama3ScaledRoPE(nn.Module):
         )
         self.register_buffer("theta", theta, persistent=False)
         self.build_rope_cache(self.max_seq_len)
+        self.is_cache_built = True
 
     def build_rope_cache(self, max_seq_len: int = 4096) -> None:
         # Create position indexes `[0, 1, ..., max_seq_len - 1]`
@@ -145,7 +147,15 @@ class Llama3ScaledRoPE(nn.Module):
             - s: sequence length
             - n_h: num heads
             - h_d: head dim
+
+        Raises:
+            RuntimeError: if RoPE cache is not initialized prior to forward call
         """
+
+        if not self.is_cache_built:
+            raise RuntimeError(
+                "RoPE cache is not built. Please call rope_init() first."
+            )
 
         # input tensor has shape [b, s, n_h, h_d]
         seq_len = x.size(1)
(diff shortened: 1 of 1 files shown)


joecummings (approved):
Neaet!

torchtune/models/llama3_1/_position_embeddings.py
```diff
@@ -146,12 +147,15 @@ def forward(
...
+        Raises:
+            RuntimeError: if RoPE cache is not initialized prior to forward call
         """
 
-        # TODO: remove once our distributed recipes are on FSDP2
         if not self.is_cache_built:
-            with torch.device(x.device):
-                self.rope_init()
+            raise RuntimeError(
+                "RoPE cache is not built. Please call rope_init() first."
```
felipemello1, line 157: this feels a bit inconvenient. Do we have to raise the error?

I also think that we should update the other positional embedding to follow the same pattern. They dont have is_cache_built. In the recipes, we should add a check:

if hasattr(m, init_rope) and m.is_cache_bult

currently it is just:
if hasattr(m, init_rope)

ebsmothers: > this feels a bit inconvenient. Do we have to raise the error?

I think it's safest. Was talking with @joecummings about this a bit; previously we had an issue where RoPE was not getting initialized and so there were garbage values left over from the `to_empty` call. This was extremely hard to debug (Joe can tell you all about that one). 

> I also think that we should update the other positional embedding to follow the same pattern.

Updating the other one is fine but I might save that for now since it has a larger blast radius and isn't blocking anything like this is.

> if hasattr(m, init_rope) and m.is_cache_bult

I think it'd be `not m.is_cache_built`, right? Either way I don't think there's any real cost to calling it twice. But I plan to move this init logic into a common util soon, so we can make the change as part of that move.

PR #1547, opened 2024-09-11
remove TiedEmbeddingTransformerDecoder from qwen

felipemello1:
#### Context
- [ ] add a new feature
- [ ] fix a bug
- [x] other (please add here)

we dont need TiedEmbeddingTransformerDecoder if we pass output_proj as a lambda

#### Changelog
- pass output_proj as lambda to the model TransformerDecoder
- deprecate  TiedEmbeddingTransformerDecoder, as qwen is the only model using it. Gemma has its own transformer, and will be deprecated in a follow up PR

#### Test plan
running a regression test before/after: to be updated


RdoubleA (approved):
Great to see this. I remember the lambda has some weird interactions with FSDP in the past but that may not be the case with FSDP2. As long as you're able to test on a distributed recipe and get identical loss, I have no concerns. Stamping to unblock

torchtune/modules/transformer.py
```diff
@@ -10,8 +10,8 @@
 import torch.nn.functional as F
 from torch import nn
 from torchtune.modules import MultiHeadAttention
-
```
felipemello1, line 13: PS: I forgot to update the typehint of the TransformerDecoder, saying that output can now be a callable. To avoid rerunning tests, this will come in a followin gemma PR

ebsmothers (commented):
This looks great. One other request for a sanity check prior to landing: please make sure that you're able to save the checkpoint then resuming training again properly (especially for a distributed run). Other than that no concerns from me!

torchtune/modules/__init__.py
```diff
@@ -32,7 +33,7 @@
     "KVCache",
     "RotaryPositionalEmbeddings",
     "RMSNorm",
-    "Fp32LayerNorm",
+    "TiedLinear" "Fp32LayerNorm",
```
joecummings, line 36: formatting weird?

torchtune/modules/transformer.py
```diff
@@ -516,6 +516,10 @@ def forward(
         return output
 
 
+@deprecated(
```
joecummings, line 519: Tell them to use TransformerDecoder WITH TiedLinear.

joecummings (approved):
two nits

PR #1548, opened 2024-09-11
Update Full Finetune for MM

pbontrager:
#### Context
- [x] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

Generalize full_finetune_single_device.py to be compatible with Flamingo. The changes in the recipe are small but required a number of bug fixes for multimodal datasets and fusion models. This PR is a grab all of bug fixes that were required to get the recipe working with Flamingo.

#### Test plan

#### UX


torchtune/modules/model_fusion/_fusion_utils.py
```diff
@@ -37,3 +37,33 @@ def fusion_params(self) -> List[str]:
...
+        model (nn.Module): Instance of model class containing some
+            fusion params.
+
+    Returns:
+        Dict[str, nn.Parameter]: the subset of model's state dict containing
+            only adapter parameters.
+
+    """
+    fusion_params = {}
+    for k, v in model.named_modules():
```
RdoubleA, line 58: could you make the for loop variables more descriptive? it's hard to follow what's happening

torchtune/modules/model_fusion/_fusion_utils.py
```diff
@@ -37,3 +37,33 @@ def fusion_params(self) -> List[str]:
...
+        if hasattr(v, "fusion_params") and callable(v.fusion_params):
+            current_fusion_params = v.fusion_params()
+            for n, p in v.named_parameters(recurse=True):
+                if n in current_fusion_params:
+                    full_key = f"{k}.{n}" if k else n
+                    fusion_params.update({full_key: p})
+                    current_fusion_params.remove(n)
+            assert (
+                current_fusion_params == []
+            ), f"Fusion params {current_adapter_params} not converted"
```
RdoubleA, line 68: maybe "Fusion params retrieved but not found in model's named parameters?"

torchtune/utils/_device.py
```diff
@@ -123,3 +123,26 @@ def get_device(device: Optional[str] = None) -> torch.device:
...
+    Raises:
+        AttributeError: if batch dict contains anything other than tensors
+    """
+    for k, v in batch.items():
+        if isinstance(v, dict):
+            batch_to_device(v, device)
+        elif isinstance(v, torch.Tensor):
+            batch[k] = v.to(device)
+        else:
+            raise AttributeError(
```
RdoubleA, line 146: This should be a ValueError or TypeError

RdoubleA (commented):
I know we need to expose `collate_fn` for multimodal, but it feels odd that users  need to remember to set a collater (or even know what that is) if they want to train multimodal. can we do an alternative approach, like a multimodal flag, or infer it from the dataset?

recipes/full_finetune_single_device.py
```diff
@@ -240,10 +242,12 @@ def setup(self, cfg: DictConfig) -> None:
 
         # sampler and dataloader depend on the tokenizer and loss_fn and should be
         # setup after both of these are initialized
+        collate_name = cfg.get("collate_fn", "torchtune.data.padded_collate_sft")
```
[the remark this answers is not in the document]
pbontrager: This is a good point. I'd be interested in compiling a list of guildelines for recipes and deciding on what are standard is going forward.

docs/source/api_ref_datasets.rst
```diff
@@ -34,8 +34,8 @@ Multimodal datasets
     :toctree: generated/
     :nosignatures:
 
-    llava_instruct_dataset
-    the_cauldron_dataset
+    multimodal.llava_instruct_dataset
```
joecummings, line 37: thank u

recipes/full_finetune_single_device.py
```diff
@@ -452,14 +462,12 @@ def _setup_data(
...
-                padded_collate_sft,
+                collate_fn,
                 padding_idx=self._tokenizer.pad_id,
                 ignore_idx=self._loss_fn.ignore_index,
             )
             if not packed
-            else partial(
-                padded_collate_packed,
-            ),
+            else padded_collate_packed,
```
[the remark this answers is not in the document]
pbontrager: I actually don't like this if/else construct here. I think after this we need to do an update and standardization of our padding function offerings.

torchtune/modules/model_fusion/_fusion_utils.py
```diff
@@ -37,3 +37,33 @@ def fusion_params(self) -> List[str]:
         return [k for k, v in self.named_parameters()]
 
     module.fusion_params = functools.partial(fusion_params, module)
+
+
+def get_fusion_params(model: nn.Module) -> Dict[str, nn.Parameter]:
```
joecummings, line 42: What's the purpose of this function? Why would you need it?

torchtune/datasets/multimodal/_the_cauldron.py
```diff
@@ -85,18 +84,21 @@ def __init__(
...
+        # Convert to messages
         messages = []
-        for message in sample[self._column_map["texts"]]:
+        for i, message in enumerate(sample[self._column_map["texts"]]):
+            user_content = [{"type": "text", "content": message["user"]}]
+            if i == 0:
+                user_content = img_content + user_content
             messages.append(
                 Message(
                     role="user",
```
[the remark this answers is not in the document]
pbontrager: Every "message" here is a user/message pair

joecummings (commented):
+ 1 to all Rafi's comments

torchtune/modules/model_fusion/_fusion_utils.py
pbontrager: This is a direct port of the get_peft_params function, so I'd prefer to keep both of those implementations in sync for now.

torchtune/modules/model_fusion/_fusion_utils.py
pbontrager: It allows you to control which parameters you want to freeze. It's similar to get_peft_params. For an example of this, look in the DeepFusionModel __init__ for an example.

joecummings (commented):
How does this deal with the kv cache changes? 

(it doesnt)

torchtune/datasets/__init__.py
```diff
@@ -43,8 +42,7 @@
     "ConcatDataset",
     "wikitext_dataset",
     "PreferenceDataset",
-    "the_cauldron_dataset",
     "SFTDataset",
     "hh_rlhf_helpful_dataset",
-    "llava_instruct_dataset",
+    "multimodal",
```
ebsmothers, line 47: Why not just expose the APIs directly in datasets/multimodal/__init__.py? Feels a bit weird to list a folder as a public API like this

torchtune/models/flamingo/_transform.py
```diff
@@ -100,6 +100,7 @@ def __init__(
         self.stop_tokens = self.tokenizer.stop_tokens
         self.max_seq_len = max_seq_len
         self.prompt_template = prompt_template
+        self.pad_id = self.tokenizer.pad_id
```
ebsmothers, line 103: I know there's not a clear way around it, but I don't love that we have to do this. It'll be a gotcha for anyone trying to write their own transforms

torchtune/modules/model_fusion/_fusion.py
```diff
@@ -269,8 +271,8 @@ def forward(self, input: torch.Tensor) -> torch.Tensor:
         # [batch_size x seq_length x embed_dim]
         out = self._fused_embed(bs, seq_len)
         mask = mask.unsqueeze(-1).expand(bs, seq_len, self.dim)
-        out.masked_scatter_(mask, embeds)
-        out.masked_scatter_(~mask, fusion_embeds)
+        out = out.masked_scatter(mask, embeds)
+        out = out.masked_scatter(~mask, fusion_embeds)
```
ebsmothers, line 275: Is this cause we need to track the grads?

torchtune/modules/model_fusion/_fusion.py
```diff
@@ -315,17 +317,45 @@ class DeepFusionModel(nn.Module):
...
+                f"encoder.{n}" for n, p in self.encoder.named_parameters()
+            }
+        if decoder_trainable:
+            trainable_params |= {
+                f"decoder.{n}" for n, p in self.decoder.named_parameters()
+            }
+        if fusion_trainable:
+            trainable_params |= set(get_fusion_params(self))
+        else:
+            trainable_params -= set(get_fusion_params(self))
```
ebsmothers, line 351: This assumes a default state of (decoder_trainable, encoder_trainable, fusion_trainable) = (False, False, True), right? It's a bit unintuitive to me that we remove only the fusion params explicitly but not the other ones

torchtune/modules/model_fusion/_fusion_utils.py
ebsmothers: The proliferation of my laziness..

torchtune/modules/transformer.py
```diff
@@ -261,22 +261,27 @@ def forward(
 
         # A mask of tokens (x) with no encoder_input
         skip_mask = self._skip_mask(encoder_mask)
+        if encoder_mask is not None:
+            # TODO: remove after PyTorch 2.5 is released
+            # This unmasks the skipped rows to avoid NaNs in SPDA Softmax backward
```
ebsmothers, line 266:
```suggestion
            # This unmasks the skipped rows to avoid NaNs in SDPA Softmax backward
```

ebsmothers: Also is this todo only related to encoder_mask? Or can we remove the usage of skip_mask altogether when 2.5 is released? I had assumed it was the latter

recipes/full_finetune_single_device.py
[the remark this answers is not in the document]
ebsmothers: Yeah I agree, we have a bunch of collate functions now and they are all added in a very adhoc fashion. I think @RdoubleA's comments still hold though.. the change in how we use collate_fn can potentially cause problems for people. Also I hate to say it but all our usage of `_get_component_from_path` is starting to feel like we are just hacking in a registry.. (not saying we shouldn't do it here btw, I actually think it's the best approach rn)

recipes/full_finetune_single_device.py
```diff
@@ -493,14 +501,10 @@ def save_checkpoint(self, epoch: int) -> None:
...
-        tokens, labels = batch["tokens"], batch["labels"]
-        # Get the attention mask and position ids from the dataset if they
-        # exist. Currently, only sample packing in PackedDataset returns these
-        mask = batch.get("mask", None)  # shape [b, s, s]
-        input_pos = batch.get("input_pos", None)  # shape [b, s]
+        # Shape [b, s], needed for the loss not the model
+        labels = batch.pop("labels")
 
-        logits = self._model(tokens, mask=mask, input_pos=input_pos)
+        logits = self._model(**batch)
```
ebsmothers, line 507: Now that we're fully in dictionary land here, we should think about doing some key validation down the line (or at least giving very clear documentation on what the expected fields are)

torchtune/datasets/multimodal/_the_cauldron.py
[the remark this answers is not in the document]
ebsmothers: Just commenting here for convenience, but is the example given in the docstring actually correct? Seems to me like it isn't

ebsmothers (commented):
No huge concerns from me, just a bunch of small comments

torchtune/models/flamingo/_transform.py
pbontrager: This is a way around this, but it involves inheriting from the tokenizer

torchtune/modules/model_fusion/_fusion.py
pbontrager: It's because using the inplace masked_scatter was giving bad gradients

torchtune/modules/model_fusion/_fusion.py
pbontrager: It's because fusion params are not their own separate module but a part of the encoder and decoder. The encoder = pretrained_encoder + fusion_params and decoder = pretrained_decoder + fusion_params. So you add the encoder and/or decoder. Then you can either remove all fusion_params if fusion_trainable = False or you add them all, since if the encoder/decoder is missing you might be missing some.

torchtune/modules/transformer.py
pbontrager: We still need skip_mask, we just don't need this extra step of updating the encoder_mask. The update would make skip_mask optional as you wouldn't have to worry about NaN's but you still get different behavior as you're not masking out the ffwd or output in attention.

recipes/full_finetune_single_device.py
pbontrager: We don't strictly need the dictionary unpacking here since we have a standard input for TransformerDecoder, but I think this is much cleaner and makes future changes easier.

torchtune/datasets/__init__.py
pbontrager: I tried not including multimodal in this init and the import path stopped working

RdoubleA (approved):

torchtune/modules/transformer.py
```diff
@@ -260,22 +260,27 @@ def forward(
 
         # A mask of tokens (x) with no encoder_input
         skip_mask = self._skip_mask(encoder_mask)
+        if encoder_mask is not None:
+            # TODO: remove after PyTorch 2.5 is released
+            # This unmasks the skipped rows to avoid NaNs in SDPA Softmax backward
+            # This doesn't affect the output since outputs are masked out later
+            encoder_mask = encoder_mask.masked_fill(skip_mask, True)
```
mmehdig, line 267: This seems to be disabling/canceling the cross-attention masking. Do we still need this?

PR #1552, opened 2024-09-12
Integrate INT8 mixed-precision from torchao 0.5

gau-nernst:
#### Context
- [x] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

Recent INT8 mixed-precision work in torchao shows very **promising results**.
- Single device A100 -> ~40% speedup
- Single device 4090 -> ~70% speedup (consumer 4000 series GPUs have unusual speedup, which is nice)
- Works with FSDP2 (tested with torchtitan)

**Known major limitations**
- Requires `torch.compile()` to enjoy speedup (to codegen efficient dynamic quantization code)
- Input sizes should not vary too much, since it will trigger autotune for triton INT8 matmul kernel -> this only works well for PackedDataset w/ FlexAttention, since `seq_len` is static.

See https://github.com/pytorch/ao/tree/v0.5.0/torchao/prototype/quantized_training#int8-mixed-precision for more details.

For now, I only added the code to show the necessary changes. I'm open to suggestions on how to expose this in torchtune. One idea from mine:
- Add a global config flag `int8_mixed_precision` (similar to `compile` flag). This will be a boolean
- Handle it inside `_setup_model()` -> repeated code for each recipe

Some concerns:
- It's possible to customize INT8 mixed-precision via `Int8MixedPrecisionTrainingConfig` (see [doc](https://github.com/pytorch/ao/tree/v0.5.0/torchao/prototype/quantized_training#int8-mixed-precision)). Should we expose it to torchtune's users? From my testing, the default config works well. There might be more knobs to customize in the future too.
- Ability to extend to other torchao's subclasses? e.g. Float8 and NF4 (right now they don't use `quantize_()` API, though they can be re-implemented to do so).

These concerns can be addressed in the future I think, when torchao's training subclasses become more mature/stable.

#### Changelog

Integrate INT8 mixed-precision from torchao 0.5

#### Test plan

#### UX

**Llama3.1-8B single device A100** 40% speedup. torch=2.5.0.dev20240911, torchao=0.5.0

> tune run full_finetune_single_device --config llama3_1/8B_full_single_device dataset.packed=True tokenizer.max_seq_len=8192 optimizer=torch.optim.AdamW optimizer.fused=True optimizer_in_bwd=False compile=True metric_logger=torchtune.training.metric_logging.WandBLogger log_peak_memory_stats=True

Screenshot from 2024-09-12 10-23-02

(TODO - a small model on 4090 to show 70% speedup)

### DIFF
--- a/recipes/full_finetune_single_device.py
+++ b/recipes/full_finetune_single_device.py
@@ -372,6 +372,11 @@ class FullFinetuneRecipeSingleDevice(FTRecipeInterface):
             memory_stats = training.get_memory_stats(device=self._device)
             training.log_memory_stats(memory_stats)
 
+        from torchao.prototype.quantized_training import int8_mixed_precision_training
+        from torchao.quantization import quantize_
+
+        quantize_(model.layers, int8_mixed_precision_training())
+
         return model
 
     def _setup_optimizer(


torchtune/training/quantization.py
```diff
@@ -74,6 +82,45 @@ def quantize(self, model):
...
+
+    NOTE: due to the limitations of the current implementation, the following
+    requirements must be satisfied to enjoy speedup:
+
+    1. Must use ``torch.compile()`` (set ``compile=True``).
+    2. Inputs to the model must not be too dynamic e.g. input sequence length changes
+    for every batch.
+
+    To satisfy (2), you can use :class:`~torchtune.datasets.PackedDataset` (set
+    ``dataset.packed=True``), which ensures input tokens always have fixed length.
```
felipemello1, line 98: I believe that the user also has to set tokenizer.max_seq_len. @RdoubleA , can you confirm?

torchtune/training/quantization.py
```diff
@@ -74,6 +82,45 @@ def quantize(self, model):
 ] = enable_8da4w_fake_quant_module_swap
 
 
+class Int8MixedPrecisionTrainingQuantizer:
+    """Apply INT8 mixed-precision training. During training, weights and activations
```
felipemello1, line 86: Great docstring. Since its a heavy topic, would you mind adding:

1) The expected benefits and tradeoffs? E.g. the user may be confused if this is for memory and if it impacts loss.
2) Args (argtype): arg comment
3) Alink to learn more about int8 mixed precision, if there is one?

torchtune/training/quantization.py
felipemello1, line 1: n00b question: would it be nice to add some sort of import guard to tell the user they need torchao > N for this? Torchao is not a requirement anymore

cc: @ebsmothers

recipes/full_finetune_single_device.py
```diff
@@ -360,6 +362,10 @@ def _setup_model(
                 model, auto_wrap_policy={modules.TransformerSelfAttentionLayer}
             )
 
+        if quantizer_cfg is not None:
+            quantizer = config.instantiate(quantizer_cfg)
```
felipemello1, line 366: it might be worth adding a logger here. Something like:

f"Quantizing the model for mixed precision training with precision {quantizer.precision}"

torchtune/training/quantization.py
```diff
@@ -74,6 +82,45 @@ def quantize(self, model):
...
+    ) -> None:
+        self._config = Int8MixedPrecisionTrainingConfig(
+            output=output,
+            grad_input=grad_input,
+            grad_weight=grad_weight,
+        )
+
+    def prepare(self, model: nn.Module) -> nn.Module:
+        # don't apply INT8 mixed-precision training to LM head
+        # since speed is slightly lower.
```
felipemello1, line 115: how about the input embedding? I just dont know if models with embeddings tied to the output head (qwen, gemma) would have issues with it

recipes/full_finetune_single_device.py
```diff
@@ -212,6 +212,7 @@ def setup(self, cfg: DictConfig) -> None:
             enable_activation_checkpointing=cfg.enable_activation_checkpointing,
             compile_model=self._compile,
             model_state_dict=ckpt_dict[training.MODEL_KEY],
+            quantizer_cfg=cfg.get("quantizer", None),
```
felipemello1, line 215: This seems to follow the same pattern as QAT. I think it makes sense, unless we have reason to worry that the user may be confused about it.

```
# QAT arguments
quantizer:
  _component_: torchtune.training.quantization.Int8DynActInt4WeightQATQuantizer
  groupsize: 256
```

I think that we should add a "enabled" flag, like we use for profiler. That way, we can add it to every config, but have enabled=False as default (or True, if we have torchao as a dependency again at some point).

Then, you would have to do:

```if quantizer_cfg is not None and quantizer_cfg.enabled:
	quantizer = config.instantiate(quantizer_cfg)
	model = quantizer.prepare(model)```

**Review by felipemello1 — COMMENTED** [review]
Its very simple and it looks great! thanks for the PR!

My two cents:
1. We need some tests to make sure it works with compile, AC, offloading (not landed yet), optimizer in backward (i guess?), etc
2. The configs can be updated in bulk using something like this dummy script: https://gist.github.com/felipemello1/5f2002433c6da3a21f33d6cdf82e702a

Let me know if you want me to help with any of these

File: torchtune/training/quantization.py (hunk 1 above)
Comment (gau-nernst) [reply torchtune/training/quantization.py:98]: Yes, `tokenizer.max_seq_len` must be set too. An error will be raised and ask to set `tokenizer.max_seq_len` when it is not set while `dataset.packed=True` is set. I will add it for completeness.

**Comment by felipemello1** [comment]
Btw, does this work with older GPUs? Does it work on CPU? Maybe we need something like this:

```
_SUPPORTS_INT8_MIXED_PRECISION = (
    torch_version_ge("2.5.0")
    and torch.cuda.is_available()
    and torch.cuda.get_device_capability() >= (7, 5)
)
```
https://github.com/pytorch/torchtune/blob/ee343e61804f9942b2bd48243552bf17b5d0d553/torchtune/modules/attention_utils.py#L22

File: torchtune/training/quantization.py (hunk 4 above)
Comment (gau-nernst) [reply torchtune/training/quantization.py:115]: This only speedups matmuls, so there is no point to applying this to input embedding anyway.
So only applying INT8 mixed-precision training to transformer layers is the safest.

File: torchtune/training/quantization.py
Comment (felipemello1) [reply torchtune/training/quantization.py:1]: does this work with older GPUs? Does it work on CPU? Maybe we need something like this:

```
_SUPPORTS_INT8_MIXED_PRECISION = (
    torch_version_ge("2.5.0")
    and torch.cuda.is_available()
    and torch.cuda.get_device_capability() >= (7, 5)
)
```
https://github.com/pytorch/torchtune/blob/ee343e61804f9942b2bd48243552bf17b5d0d553/torchtune/modules/attention_utils.py#L22

Comment (gau-nernst) [reply torchtune/training/quantization.py:1]: Good question! I use Triton for this so it will probably run on any GPUs Triton supports (same as torch.compile). Though I think only Ampere (sm80) and above has INT8 tensor cores. To be safe, I think we just guard for sm80 and above.

CPU is not supported. Technically it is possible, but I didn't add it in torchao since I can't reliably test it / see it useful.

This works with PyTorch 2.4 (though FlexAttention requires PyTorch 2.5 🤣).

Comment (gau-nernst) [reply torchtune/training/quantization.py:1]: Btw doesn't QAT also need some kind of guards like this? 🤔

**Comment by felipemello1** [comment]
@RdoubleA , do you have any thoughts about the flex attention issue on 4090?

>Note: I can't test with 4090 since FlexAttention errors out on 4090
triton.runtime.errors.OutOfResources: out of resource: shared memory
It's pretty strange since it works fine for another repo of mine 🤔.

File: tests/recipes/test_full_finetune_single_device.py [hunk 6]
```diff
@@ -263,3 +264,89 @@ def test_gradient_accumulation(self, tmpdir, monkeypatch):
...
+            checkpointer.checkpoint_files=[{ckpt_path}]\
+            checkpointer.output_dir={tmpdir} \
+            checkpointer.model_type={model_type.upper()} \
+            tokenizer.path='{tokenizer_path}' \
+            tokenizer.prompt_template=null \
+            tokenizer.max_seq_len=4096 \
+            dataset.packed=True \
+            metric_logger.filename={log_file_int8mp} \
+            compile=True \
+            quantizer._component=quantizer._component_={quantizer} \
```
felipemello1, line 337:
```suggestion
            quantizer._component_={quantizer}\
```

tests/recipes/test_full_finetune_single_device.py
```diff
@@ -263,3 +264,71 @@ def test_gradient_accumulation(self, tmpdir, monkeypatch):
...
+            "seed=9",
+            "epochs=1",
+            "max_steps_per_epoch=5",
+            "optimizer=torch.optim.AdamW",
+            "optimizer_in_bwd=False",
+            "compile=True",
+        ] + dummy_alpaca_dataset_config()
+
+    @pytest.mark.integration_test
+    def test_smoke(self, tmpdir, monkeypatch):
```
gau-nernst, line 281: couldn't make speed test pass, probably because the model is too small in CI. Anyway, speed benchmarks don't seem to be reliable without a controlled setup. So changed this to smoke test i.e. make sure it works.

torchtune/training/quantization.py
```diff
@@ -127,6 +152,80 @@ def quantize(self, model):
...
+        )
+
+    def prepare(self, model: nn.Module) -> nn.Module:
+        quantize_fn = int8_mixed_precision_training(self._config)
+
+        # custom filter_fn to work with torchtune's peft
+        def filter_fn(module: nn.Module, name: str) -> bool:
+            if isinstance(module, nn.Linear):
+                return not (name.endswith(".lora_a") or name.endswith(".lora_b"))
+
```
felipemello1, line 213:
```suggestion
TODO: LoRALinear, DoRALinear from some constant variable e.g. PEFT_MODULES, so its future-proof
```

torchtune/training/quantization.py
```diff
@@ -127,6 +152,80 @@ def quantize(self, model):
...
+            if isinstance(module, (LoRALinear, DoRALinear)):
```
felipemello1, line 214:
```suggestion
            # TODO: LoRALinear, DoRALinear from some constant variable e.g. PEFT_MODULES, so its future-proof
            if isinstance(module, (LoRALinear, DoRALinear)):
```

torchtune/training/quantization.py
```diff
@@ -127,6 +152,80 @@ def quantize(self, model):
...
+
+            return False
+
+        # Don't apply INT8 mixed-precision training to LM head since end2end speedup
+        # will be slightly worse. There are also possible issues with tied word
+        # embeddings.
+        if isinstance(model, TransformerDecoder):
+            quantize_(model.layers, quantize_fn, filter_fn=filter_fn)
+        else:
+            quantize_(model, quantize_fn, filter_fn=filter_fn)
```
felipemello1, line 225: @pbontrager , any recommendations here for multimodal? I am thinking that we could do something like this:

if isinstance(model, TransformerDecoder) or isinstance(model, VisionTransformer):
    quantize_(model.layers, quantize_fn, filter_fn=filter_fn)
else:
    quantize_(model, quantize_fn, filter_fn=filter_fn)

tests/recipes/test_full_finetune_single_device.py
felipemello1: I believe that its still not passing in the CI because number of ranks is larger than number of gpus available. I think we need this flag `@gpu_test(gpu_count=1)`

https://github.com/pytorch/torchtune/blob/0a3762d058fd9d860606a3a6bbebf71ac593ab5c/tests/recipes/test_full_finetune_distributed.py#L63C5-L63C27

torchtune/training/quantization.py
felipemello1: how bad is the performance if flex attention if off? I am thinking if this should be part of the requirement. If so, then we can just include this import guard in the check: https://github.com/pytorch/torchtune/blob/0a3762d058fd9d860606a3a6bbebf71ac593ab5c/torchtune/utils/_import_guard.py#L11

Maybe thats a bad idea, we can raise a warning instead: Use compile and dataset.packed for best results.

torchtune/training/quantization.py
```diff
@@ -29,6 +32,27 @@
...
 
+from torchtune.modules import TransformerDecoder
+from torchtune.modules.low_precision._utils import _get_torchao_version
+from torchtune.modules.peft import DoRALinear, LoRALinear
+from torchtune.utils._version import torch_version_ge
+
+
+_TORCHAO_VERSION, _ = _get_torchao_version()
+
+_SUPPORTS_INT8_MIXED_PRECISION_TRAINING = (
```
felipemello1, line 43: how bad is the performance if flex attention if off? I am thinking if this should be part of the requirement. If so, then we can just include this import guard in the check: https://github.com/pytorch/torchtune/blob/0a3762d058fd9d860606a3a6bbebf71ac593ab5c/torchtune/utils/_import_guard.py#L11

Maybe thats a bad idea, we can raise a warning instead: Use compile and dataset.packed for best results.

torchtune/training/quantization.py
```diff
@@ -127,6 +152,80 @@ def quantize(self, model):
...
+    To satisfy (2), you can use :class:`~torchtune.datasets.PackedDataset` (set
+    ``dataset.packed=True`` and ``tokenizer.max_seq_len`` to a desired value.), which
+    ensures input tokens always have fixed length.
+    """
+
+    def __init__(
+        self,
+        output: bool = True,
+        grad_input: bool = True,
+        grad_weight: bool = True,
```
felipemello1, line 192: I dont think that there is a way around us always having our own wrapper, since we need it to deal with LoRA/other details. Therefore, I think that we should add "enabled" here, and have config like this:

```
mixed_precision:
	_component_: torchtune.training.quantization.Int8MixedPrecisionTraining
	arg1: True
	arg2: True
	enabled: False
```

That way the configs can advertise this functionality.

I also think that we should rename quantizer to "mixed_precision" or "mixed_precision_training", so its self-explanatory.

Let me know if you need help with making the changes or if you disagree with the ideas above.

thanks again for the PR! @ebsmothers might want to leave a few comments too
For me, after these changes and the test pass, i am good to merge.

tests/recipes/test_full_finetune_single_device.py
```diff
@@ -263,3 +264,71 @@ def test_gradient_accumulation(self, tmpdir, monkeypatch):
...
+        model_config = [
+            "model._component_=torchtune.models.llama3.llama3",
+            "model.vocab_size=128_256",
+            "model.num_layers=2",
+            "model.num_heads=2",
+            "model.num_kv_heads=2",
+            "model.embed_dim=256",
+            "model.max_seq_len=1024",
+        ]
+        dummy_model = llama3(
```
felipemello1, line 299: i love this dummy_model idea

tests/recipes/test_full_finetune_single_device.py
```diff
@@ -263,3 +264,71 @@ def test_gradient_accumulation(self, tmpdir, monkeypatch):
...
+            max_seq_len=1024,
+        )
+        ckpt_dir = tmpdir / "ckpt_dir"
+        ckpt_dir.mkdir()
+        torch.save(dummy_model.state_dict(), ckpt_dir / "model.pt")
+
+        # set dataset.packed=True to have fixed input seq len
+        cmd = f"""
+        tune run full_finetune_single_device \
+            --config llama3/8B_full_single_device \
```
felipemello1, line 314: since there may be issues around lora, can we duplicate this function and add it for QLoRA and/or LoRA too?

tests/recipes/test_full_finetune_single_device.py
ebsmothers: Yeah looks like that plus a version check which currently fails on even the GPU runners. Can add something like `@pytest.mark.skipif(not _SUPPORTS_INT8_MIXED_PRECISION_TRAINING)`

ebsmothers: But agreed on not doing speed test in here, correctness should be the main focus. If you're able to get reproducible loss values that would be a good sanity check. Down the line I'd like to add some performance testing in our currently-very-underutilized `tests/regression_tests` but that's a bigger effort that we shouldn't block on here.

tests/recipes/test_full_finetune_single_device.py
```diff
@@ -263,3 +264,71 @@ def test_gradient_accumulation(self, tmpdir, monkeypatch):
...
+    @pytest.mark.integration_test
+    def test_smoke(self, tmpdir, monkeypatch):
+        model_type = "llama3"
+        ckpt_component = CKPT_COMPONENT_MAP["tune"]
+        tokenizer_path = Path(TOKENIZER_PATHS[model_type])
+        log_file = gen_log_file_name(tmpdir)
+
+        # MODEL_TEST_CONFIGS["llama3"] doesn't work with FlexAttention
+        # because some dims are not multiple of 128
+        # create a dummy model and save state_dict so it works with torchtune
```
ebsmothers, line 289: Oh no... we should update the dummy model

torchtune/training/quantization.py
ebsmothers: Sorry I completely missed this convo previously. Actually now that we've asked users to install torchao manually I am kinda taking a similar stance there to what we have with PyTorch: people should always be running on the latest stable version. So actually I claim that the first two lines of the `_SUPPORTS_INT8_MIXED_PRECISION_TRAINING` check are not strictly necessary (fine to keep them in though, I don't have a strong preference). But anyways that should hopefully answer the question about why we don't have similar such guards for QAT

torchtune/training/quantization.py
```diff
@@ -127,6 +152,80 @@ def quantize(self, model):
...
+            grad_weight=grad_weight,
+        )
+
+    def prepare(self, model: nn.Module) -> nn.Module:
+        quantize_fn = int8_mixed_precision_training(self._config)
+
+        # custom filter_fn to work with torchtune's peft
+        def filter_fn(module: nn.Module, name: str) -> bool:
+            if isinstance(module, nn.Linear):
+                return not (name.endswith(".lora_a") or name.endswith(".lora_b"))
```
ebsmothers, line 212: Just curious.. why do we not apply on LoRA weights? Is it just because these are the only trainable weights and the model quality degradation is not worth the perf improvement we'd get in this case?

torchtune/training/quantization.py
```diff
@@ -127,6 +152,80 @@ def quantize(self, model):
...
+
+    Raises:
+        RuntimeError: If runtime requirements for INT8 mixed-precision training are not met.
+
+    NOTE: Due to the limitations of the current implementation, the following
+    requirements must be satisfied to enjoy the expected speedup:
+
+    1. Must use ``torch.compile()`` (set ``compile=True``).
+    2. Inputs to the model must not be too dynamic. For example, when input tokens
+    length changes for every batch, you won't see the expected speedup.
```
ebsmothers, line 181: Should we warn in the recipe about one or both of these? (The compile one at least seems fairly important). Also just out of curiosity, what exactly does "too dynamic" mean? I know you mention in the summary about triggering autotune, under what conditions does this occur?

torchtune/training/quantization.py
```diff
@@ -127,6 +152,80 @@ def quantize(self, model):
...
+    minimal accuracy loss. If convergence is an issue, please refer to torchao
+    documentation below.
+
+    For more details, as well as details about arguments of this quantizer, please refer to
+    https://github.com/pytorch/ao/tree/main/torchao/prototype/quantized_training#int8-mixed-precision
+
+    Args:
+        output (bool): whether to apply INT8 mixed-precision for calculating output.
+        grad_input (bool): whether to apply INT8 mixed-precision for calculating grad_input.
+        grad_weight (bool): whether to apply INT8 mixed-precision for calculating grad_weight.
```
ebsmothers, line 171: nit: add defaults here

torchtune/training/quantization.py
ebsmothers: Yeah personally I like the change to rename `quantizer` -> `mixed_precision` (unless the idea is to later support full low-precision training using the same API, in which case it wouldn't make sense). I am on the fence about the inclusion of the `enabled` flag.. it's nice because it allows us to expose directly in our configs without actually enabling it by default. At the same time, I don't love the pattern of defining dummy objects with corresponding thin wrappers that optionally return them (unless it's absolutely necessary, as is kinda the case with some of our context-manager-based features). Personally I am OK with just adding something like this to all our configs if we wanna advertise this feature:

```
# Set to torchtune.training.Int8MixedPrecisionTrainingQuantizer to use int8 mixed precision
mixed_precision: null
```

recipes/full_finetune_single_device.py
```diff
@@ -364,6 +366,11 @@ def _setup_model(
                 model, auto_wrap_policy={modules.TransformerSelfAttentionLayer}
             )
 
+        if quantizer_cfg is not None:
```
ebsmothers, line 369:
```suggestion
        if quantizer_cfg is not None and quantizer_cfg.get("enabled", False):
```

ebsmothers (commented):
Apologies for the delay in getting to this PR. Overall it looks good, I don't have any major concerns. One question based on the PR summary: what's the issue with `training.load_from_full_model_state_dict()`? If that's the only blocker for enabling this with FSDP2 recipes we should push to figure it out. Otherwise I tried to address some of your UX questions from the summary with my inline comments, but lmk if things are still unclear. It'll be great to have this supported in torchtune!

torchtune/training/quantization.py
gau-nernst: The weight should be sufficiently large for the INT8 matmul speedup to outweigh the dynamic quantization overhead. I think it could speedup the LoRA path too if LoRA rank is large enough. But again, the LoRA path computation is tiny compared to computation of base weight, so I think it's not really worth the efforts to apply it to LoRA weights.

torchtune/training/quantization.py
gau-nernst: The autotune is triggered every time input shapes and layouts change https://github.com/pytorch/ao/blob/9ce7ebbadc2433dc2935b77b4ada0e4d9e54fcb7/torchao/prototype/quantized_training/int8_mm.py#L52

Input sequence length changes for every batch will be considered "too dynamic".
Input sequence length only has a few possible values, like 256, 512, 1024, is dynamic, but still ok, since we can just autotune on these fixed set of values.

gau-nernst:
@ebsmothers Regarding FSDP2 issue, when loading state dict, the current logic is to shard the pre-trained weights and load it with `assign=True`

https://github.com/pytorch/torchtune/blob/7af77c7b7f2de4bc11d739a3c6bb272bccd499a3/torchtune/training/_distributed.py#L346

Since my implementation depends on tensor subclass (swap `nn.Parameter`), `assign=True` will replace my tensor subclass with the ordinary tensor -> no effect. Possibly a reason why NF4 has a special logic inside `load_from_full_model_state_dict()` too. (btw I think this would affect QAT tensor subclass, but didn't check myself. Andrew is changing QAT back to module swap anyway, so might not be too important).

Another option is to do tensor subclass after `fully_shard()`. However, this interferes with FSDP2 logic, since I think FSDP2 assigns its own privates state to `nn.Parameter` as well as some links between `nn.Module` <-> `nn.Parameter`?

For a quick hack, I suppose we could add special handling logic inside `load_from_full_model_state_dict()` just like NF4, but considering that my work is still pretty prototype/experimental (i.e. can change anytime), I'm not too inclined to do it.

In summary, the problem stems from the fact that we need to shard and convert sharded tensors from pre-trained weights to its correct tensor subclass in the model (to use `assign=True`).
- I think it's possible to NOT use `assign=True`, so that we will trigger the copy logic instead, which will preserve the existing tensor subclass inside the model. But this will be slower (due to extra copy) + incur extra memory usage at model load time (need to materialize model weights + state_dict weight).
- To combat the extra memory usage mentioned above, perhaps we could shard the state dict using CPU process group instead (e.g. mpi/gloo)? (just speculating, not even sure if it's possible in a simple way. and will likely be slower).

The suggested workaround above is still not exactly ideal (bad in many ways). So yea, I guess I will leave it for distributed wizards :smile:

torchtune/training/quantization.py
gau-nernst: Yea I agree `quantizer` name right now can sound confusing, since the weight is actually not quantized (well QAT also does not actually quantize the weight :smile:)

Regarding `mixed_precision` name, I'm thinking about what other training recipes out there that would fall under this category (because it would be weird if we have `mixed_precision` but only 1 option). There are FP8 and FP16/BF16 mixed-precision training, but they follow a totally different UX
- **FP8**: the current FP8 training recipes, in torchao or transformer engine, is actually mixed-precision training, since master weights is still in high precision, although people usually don't say it is mixed-precision. Currently FP8 UX is to do module swap (but I think for FP8 all-gather, FP8 also does tensor subclass). I think Vasiliy is having plans to update FP8 UX, not exactly sure. But anyway this is an implementation detail which can be hidden away from torchtune users. (also someone should add FP8 to torchtune :smile:)
- **BF16** (w/ FP32 master weights): for single-GPU, it will be using `torch.autocast(dtype=torch.bfloat16)`. For FSDP2, it would be using `MixedPrecisionPolicy(param_dtype=torch.bfloat16)`. This is very different UX (and mechanism) from INT8 and FP8, and would be hard to unify the torchtune user-facing API (and not sure if torchtune wants to add this option? Btw there can be small gaps between full BF16 training (current torchtune) and BF16 mixed-precision training, especially when LR is small. Would take some efforts for someone to investigate this :sweat_smile:).
- **FP16** (w/ FP32 master weights): probably won't consider this due to complexity with gradient scaling + finetune BF16 models with FP16 is a [bad idea](https://github.com/huggingface/transformers/issues/23918).

There are also MX FP6/FP4 in the new NVIDIA GPUs, but that's too far in the future. Considering that at least we can have INT8 and FP8 mixed-precision training, using the `mixed_precision` name seems reasonable!

felipemello1: > also someone should add FP8 to torchtune 

Would it be too different from int8? Or can we leverage the same API and replace the quantization module?
BTW, does FP8 work with FSDP? In this case, is there anything that makes it special, since int8 seems to not be compatible?

torchtune/training/quantization.py
felipemello1: Can we maybe use mark_dynamic to avoid recompilation if dataset.packed is false?

torchtune/training/quantization.py
gau-nernst: @felipemello1 Currently torchao API is different https://github.com/pytorch/ao/tree/main/torchao/float8. But it can be hidden away by torchtune (e.g. a separate Quantizer class, or some other name like MixedPrecisionTrainer class)

Both FP8 (by FP8 folks) and INT8 (by me) support FSDP2. The issue is with how torchtune handles distributed weight loading, which is not trivial to handle. FP8 module swap might not have this problem though.

torchtune/training/quantization.py
gau-nernst: The autotune logic is done by Triton (since this is a hand-written Triton kernel). `mark_dynamic` is for torch.compile(). It won't do anything in this case.

felipemello1: n00b question, beyond this PR's scope, do you know if torch.compile can be used on a triton module? And if so, then maybe mark_dynamic be used?

gau-nernst: From what I understand, torch.compile() will treat a triton kernel as a "black box". Even if it can "see inside" the triton kernel, it can't possibly bypass/change the autotune logic. If we want the kernel to work with more "dynamicism", the solution would be to write custom autotune logic (i.e. when to trigger a re-tuning), which is not exactly simple.

weifengpy:
> FSDP2 assigns its own privates state to nn.Parameter as well as some links between nn.Module <-> nn.Parameter?

exactly. FSDP2 init interal state by visiting `model.parameters()`. We want that parameters to be `Int8QuantizedTrainingLinearWeight` at the time of `fully_shard(model)`. so we can trigger fsdp_pre_all_gather and fsdp_post_all_gather

> I think it's possible to NOT use assign=True

for experimenting purpose, it's totally up to you to try out `assign=True`. If we want to landing this seriously, we might need to do `assign=True` to keep consistency with NF4

to reduce memory, what do you think of calling `model.load_state_dict(strict=False)` tensor by tensor? this way we only materialize one tensor on cuda at a time. You can use DTensor API (NCCL) without worry about creating gloo/mpi backend. The downside is slowness

> the state_dict (for weight checkpoint) will contain tensor subclass too

for landing purpose, we usually upcast weight to high precision and then saving model state dict in plain tensor

weifengpy:
cc @mori360 for INT8 + distributed state dict

gau-nernst:
@ebsmothers @weifengpy I'm not too keen on adding the hack (i.e. manually shard original weight and construct tensor subclass from it) inside `load_from_full_model_state_dict()` like NF4. I think this is an internal implementation that downstream users like torchtune should not use/rely on it, as it can change in the future, either because of new features or some kind of refactoring (I think NF4 might have this problem too: It is hard to refactor/change NF4 dtype in torchao now).

If we add more training recipes that are implemented with tensor subclass, such as FP8 with FP8 all-gather, torchtune would need to again manually add more logic in `load_from_full_model_state_dict()` too.

Also, just to clarify, this PR will use `Int8MixedPrecisionTrainingLinearWeight` from [here](https://github.com/pytorch/ao/blob/9e2a2536d56626e59618a8932e2d1e160f7f76ca/torchao/prototype/quantized_training/int8_mixed_precision.py). Weight is kept at original precision, and it is quantized dynamically to use INT8 tensor cores, just like FP8 training. So when saving state dict, it's just the matter of unwrapping the tensor subclass. The weight itself is not quantized.

Actually this can be implemented as module-swap too. Since it doesn't implement quantized all-gather yet (only modify forward and backward pass), module-swap will save troubles for torchtune -> state dict is still pure tensor. (though iirc, module-swap was slower than tensor subclass implementation due to some torch.compile quirks). And to clarify, today, INT8 mixed-precision training in torchao is only implemented as tensor subclass, though I did briefly try module-swap implementation in the past.

weifengpy:
> @ebsmothers @weifengpy I'm not too keen on adding the hack (i.e. manually shard original weight and construct tensor subclass from it) inside `load_from_full_model_state_dict()` like NF4

totally agree that float8 and INT8 can be done in a cleaner way. NF4 is unique in 3 ways : 1) NF4 has double-quantization  that requires computation over global tensors. quantize-then-chunk is much easier than chunk-then-quantize, 2) NF4 is static quantizaiton and thus lost high precision tensor, 3) NF4 dispatch had a hack on `copy_` back in the old days to quickly unblock single device QLoRA

> when saving state dict, it's just the matter of unwrapping the tensor subclass. The weight itself is not quantized.
> module-swap was slower than tensor subclass implementation due to some torch.compile quirks

great insights!

ebsmothers:
Thanks @gau-nernst and @weifengpy for the helpful discussion here. I guess we don't need to block this PR on figuring out the plan for enabling in distributed recipes. But for the sake of my understanding, let me summarize the options that have been discussed:

1) Perform `load_state_dict` without `assign=True`
	a) Optionally shard with CPU process group
2) Call `load_state_dict(strict=False)` one tensor at a time
3) Manually shard original weight and build tensor subclass inside `load_from_full_model_state_dict`
4) Enable module-swap int8 mixed precision training

It seems like (1)+(a) is unclear, and (3) is not easy to generalize. Also I imagine (4) is a nontrivial effort. So then it is either the vanilla version of (1) or (2). 

It seems to me like (2) should be similar to the memory and speed of the current `assign=True` usage, but it is a bit awkward. Then the question for me is whether the doubled memory of (1) would actually be the peak memory for the entire run -- if so I think we shouldn't do it. My very basic math is that this would be the case whenever `memory(model weights) > max(memory(grads + optimizer states + activations))`, is that right? If so I guess it could hold for LoRA + low-precision optimizer at smaller sequence lengths (but as mentioned in a comment elsewhere, maybe LoRA is not the primary use case here).

gau-nernst:
@ebsmothers (4) is actually somewhat trivial, but I think it's best to be done in torchao instead of torchtune. Maybe I will add module-swap UX to torchao first before continuing with this PR (more about torchao problem, I'm not sure if supporting 2 UX flows is the wise thing to do. But perhaps since it's experimental, having more UX options is good to evaluate which flow is better for downstream users like torchtune).

gau-nernst:
@felipemello1 @ebsmothers With pytorch/ao#1179, we can revisit this PR again. Do you mind resolve previous conversations if they are indeed resolved, so it's less cluttered? And i can address the unresolved ones.

I have updated the PR description. You can see it for new changes + benchmarks.

In summary:
1. pytorch/ao#1179 adds module-swap UX -> torchtune will use the module swap UX -> resolve the issues with (1) state dict (everything is still plain tensors) and (2) distributed training (shard and load pre-trained weights)
    - This will only be available for the next torchao release (0.7) or from nightly/source.
2. Expose this as `mixed_precision` flag in the config file, with `enabled` option. Once we are ok with this, I will add it to the rest of the configs (if you want)
    - Some alternative names: `low_bit_linear/matmul` -> a bit more precise than `mixed_precision`, since it refers to the fact that it uses low-bit for matmul (and the output is still BF16). I think nowadays "mixed precision" can mean many things, thus might be a bit vague. (Just raising some ideas/options to think about. No strong preference about the naming)
    - `torchtune.training.quantization.Int8MixedPrecisionTrainingQuantizer` might not be the right name/place to put this? Perhaps `torchtune.training.mixed_precision` (or whatever we gonna call the flag from the previous pointer). To give a better picture, in the future we can do float8 in a similar manner. Again, no strong preference, I can follow whatever you all decide.
3. This is no longer compatible with LoRA/QLoRA (previously tensor-subclass is compatible with LoRA, but not QLoRA). This is because `LoRALinear` hard-code `F.linear(x, self.weight)` for the base weight, making it less flexible to modify this op. See https://github.com/pytorch/torchtune/pull/1931#issuecomment-2453454525 for more detailed explanation and my proposed solution (this would be for future PR)

weifengpy:
great work, @gau-nernst! 

@mori360 this module swap UX and torchtune example should help with torchtitan recipe as well.

torchtune/training/quantization.py
```diff
@@ -43,6 +48,22 @@
         Int8DynActInt4WeightQATQuantizer,
     )
 
+from torchtune.utils._version import torch_version_ge
+
+
+_SUPPORTS_INT8_MIXED_PRECISION_TRAINING = (
+    torch_version_ge("2.4.0")
+    and Version(torchao.__version__) >= Version("0.7")
```
felipemello1, line 56: this didnt work for me

image

torchtune/training/quantization.py
```diff
@@ -144,6 +166,84 @@ def quantize(self, model):
...
+        self._config = Int8MixedPrecisionTrainingConfig(
+            output=output,
+            grad_input=grad_input,
+            grad_weight=grad_weight,
+        )
+
+    def prepare(self, model: nn.Module) -> nn.Module:
+        # we use module-swap implementation so that the state_dict remains plain tensors,
+        # as well as better FSDP compatibility in torchtune.
+        quantize_fn = int8_mixed_precision_training(self._config, module_swap=True)
```
felipemello1, line 223: got error

quantize_fn = int8_mixed_precision_training(self._config, module_swap=True)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: int8_mixed_precision_training() got an unexpected keyword argument 'module_swap'

torchtune/training/quantization.py
```diff
@@ -144,6 +166,84 @@ def quantize(self, model):
...
+                " a CUDA-capable device with compute capability >= 8.0"
+            )
+
+        self._config = Int8MixedPrecisionTrainingConfig(
+            output=output,
+            grad_input=grad_input,
+            grad_weight=grad_weight,
+        )
+
+    def prepare(self, model: nn.Module) -> nn.Module:
```
felipemello1, line 220: not working with llama 3b. Here is the TiedLinear implementation: https://github.com/pytorch/torchtune/blob/e1caa9f82fea24d728f9b244a9dd1957f5ed7465/torchtune/modules/tied_linear.py#L28

the docstring explains it, but tldr: TiedLinear is not a nn.module, so that fsdp doesnt try to shard it. Then the Linear implementation is an nn.Module, so that we can add hooks to it, so it works with act offloading.

```
  File "/data/users/felipemello/torchtune/torchtune/modules/transformer.py", line 649, in forward
    output = self.chunked_output(h)
             ^^^^^^^^^^^^^^^^^^^^^^
  File "/home/felipemello/.conda/envs/torchtune/lib/python3.11/site-packages/torch/_dynamo/eval_frame.py", line 721, in _fn
    return fn(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^
  File "/data/users/felipemello/torchtune/torchtune/modules/transformer.py", line 501, in chunked_output
    return [
           ^
  File "/data/users/felipemello/torchtune/torchtune/modules/transformer.py", line 502, in <listcomp>
    self.output(chunk)
  File "/data/users/felipemello/torchtune/torchtune/modules/tied_linear.py", line 66, in __call__
    return self.linear(x, self.tied_module.weight)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/felipemello/.conda/envs/torchtune/lib/python3.11/site-packages/torch/nn/modules/module.py", line 1740, in _wrapped_call_impl
    return self._call_impl(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/felipemello/.conda/envs/torchtune/lib/python3.11/site-packages/torch/nn/modules/module.py", line 1751, in _call_impl
    return forward_call(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/data/users/felipemello/torchtune/torchtune/modules/tied_linear.py", line 27, in forward
    return F.linear(x, weight)
           ^^^^^^^^^^^^^^^^^^^
RuntimeError: std::bad_alloc
```

I printed the x and weight and the weight
```
torch.Size([5, 256, 3072]) torch.bfloat16
torch.Size([128256, 3072]) torch.bfloat16
```

cfg
```
tune run full_finetune_single_device --config llama3_2/8B_full_single_device \
optimizer_in_bwd=False \
enable_activation_checkpointing=False \
enable_activation_offloading=False \
optimizer._component_=torch.optim.AdamW \
compile=True \
dataset.packed=True \
dataset.split=train[:5%] \
tokenizer.max_seq_len=2048 \
log_every_n_steps=1 \
log_peak_memory_stats=True \
max_steps_per_epoch=15 \
epochs=1 \
batch_size=1 \
loss=torchtune.modules.loss.CEWithChunkedOutputLoss \
mixed_precision._component_=torchtune.training.quantization.Int8MixedPrecisionTrainingQuantizer \
mixed_precision.enabled=True
```

update: not working with llama 8b either

```
  File "/data/users/felipemello/torchtune/torchtune/modules/transformer.py", line 649, in forward
    output = self.chunked_output(h)
             ^^^^^^^^^^^^^^^^^^^^^^
  File "/home/felipemello/.conda/envs/torchtune/lib/python3.11/site-packages/torch/_dynamo/eval_frame.py", line 721, in _fn
    return fn(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^
  File "/data/users/felipemello/torchtune/torchtune/modules/transformer.py", line 501, in chunked_output
    return [
           ^
  File "/data/users/felipemello/torchtune/torchtune/modules/transformer.py", line 502, in <listcomp>
    self.output(chunk)
  File "/home/felipemello/.conda/envs/torchtune/lib/python3.11/site-packages/torch/nn/modules/module.py", line 1740, in _wrapped_call_impl
    return self._call_impl(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/felipemello/.conda/envs/torchtune/lib/python3.11/site-packages/torch/nn/modules/module.py", line 1751, in _call_impl
    return forward_call(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/felipemello/.conda/envs/torchtune/lib/python3.11/site-packages/torch/nn/modules/linear.py", line 125, in forward
    return F.linear(input, self.weight, self.bias)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
RuntimeError: std::bad_alloc
```

```
tune run full_finetune_single_device --config llama3_1/8B_full_single_device \
optimizer_in_bwd=False \
enable_activation_offloading=False \
optimizer._component_=torch.optim.AdamW \
compile=True \
dataset.packed=True \
dataset.split=train[:5%] \
tokenizer.max_seq_len=2048 \
log_every_n_steps=1 \
log_peak_memory_stats=True \
max_steps_per_epoch=15 \
epochs=1 \
batch_size=1 \
loss=torchtune.modules.loss.CEWithChunkedOutputLoss \
mixed_precision._component_=torchtune.training.quantization.Int8MixedPrecisionTrainingQuantizer \
mixed_precision.enabled=True
```

maybe its my machine? will test a bit more

torchtune/training/quantization.py
```diff
@@ -144,6 +166,84 @@ def quantize(self, model):
 ] = enable_8da4w_fake_quant_module_swap
 
 
+class Int8MixedPrecisionTrainingQuantizer:
```
felipemello1, line 169: Here is my understanding:

1) torchao has a nice api that does _quantize(model, config)
2) This pr creates a wrapper specific for int8. The wrapper is needed for two reasons:
	i) ignora lora_a/lora_b
	ii) we dont do nested configs

3) this means that if we want to support fp8. bitnet, or any other torchao technique, we have to create a custom wrapper, instead of doing _quantize(model, torchao_config)

---- 
IMO, this makes a lot of sense if:
a) every torchao technique interacts differently with torchtune, e.g. one doesnt work with offloading, another needs some extra work for ckpt, etc, and we cant solve it with a config parser
b) if there are realistically only a couple or two quantization methods we will use from torchao (e.g. int8 and fp8)

But if thats not the case, then we should probably avoid having a custom torchtune wrapper per torchao technique.

I guess we had a similar situation with OptimizerCPUOffload. We ended up giving up on a custom torchtune wrapper and just instantiated directly from the config.

----
In summary, instead of "Int8MixedPrecisionTrainingQuantizer", should we have a generalist torchtune "PrecisionTrainingQuantizer", just to handle LoRA, etc?

recipes/configs/llama3_1/8B_full.yaml
```diff
@@ -83,6 +83,11 @@ output_dir: /tmp/full-llama3.1-finetune
 log_every_n_steps: 1
 log_peak_memory_stats: True
 
+# mixed precision (disabled)
+mixed_precision:
+  _component_: torchtune.training.quantization.Int8MixedPrecisionTrainingQuantizer
+  enabled: false
```
felipemello1, line 89: thanks for adding "enabled"!

felipemello1:
@gau-nernst , can you help me understand why memory is not changed? Do we do a bf16->int8->bf16 projection  inplace?

torchtune/training/quantization.py
gau-nernst: Regarding `std::bad_alloc`, I think it's an issue with recent pytorch nightlies. ao CI has been seeing it too pytorch/ao#1229. Currently CI pins to 20241101 https://github.com/pytorch/ao/blob/b2642fb33e360ffe478fe19665b1c4efd80537c6/.github/workflows/regression_test.yml#L64. Can you try with this version?

I don't exactly recall, but I think usually I had to kill the first run (it kinda hangs?), then the subsequent run is fast. Didn't think much of it. I will try to reproduce it again in a fresh environment to see if my memory serves me right.

gau-nernst:
@felipemello1 

> can you help me understand why memory is not changed?

Memory is not expected to change. In this scheme, weights and activations are quantized dynamically to INT8 (with scaling, so that we can do INT8 matmul), then the result is scaled back to BF16. Activations and weights are still in BF16 throughout the model. This is the same strategy as the current `torchao.float8` (in fact, in some bad cases, `torchao.float8` can consume more memory in FSDP2 due to some autograd shenanigans).

To reduce memory, either weight or activations must stay in low-precision. Using only low-bit weight for training is a bit challenging (not impossible, but there will be convergence/accuracy issues). There are some research works on using low-bit activations (FP8/INT8), but we don't have that in torchao yet. e.g.
- FP8 activations: https://arxiv.org/abs/2410.19313
- INT8 activations: https://arxiv.org/abs/2403.12422

torchtune/training/quantization.py
gau-nernst: Apparently I need to add `.dev` https://github.com/pytorch/ao/blob/2ba1a61fe1244560325b5051b5d3c10044553be0/torchao/utils.py#L529-L532

I was installing torchao from source, which has version number as `0.7.0+gitxxxxx`, and the comparison works as expected. Will push the change.

torchtune/training/quantization.py
gau-nernst: ok I additionally set `quantize_(set_inductor_config=False)` to avoid exhaustive torch.compile tuning. Can you try again? It seemed fast to me (<2min). Worked with both Llama3.1-8B and Llama3.2-3B. `torch==2.5.1+cu121` and `torchao==0.7.0.dev20241112+cu121`

The command

> `tune run full_finetune_single_device --config llama3_2/3B_full_single_device dataset.packed=True tokenizer.max_seq_len=8192 optimizer=torch.optim.AdamW optimizer.fused=True optimizer_in_bwd=False compile=True metric_logger=torchtune.training.metric_logging.WandBLogger log_peak_memory_stats=True mixed_precision._component_=torchtune.training.quantization.Int8MixedPrecisionTrainingQuantizer mixed_precision.enabled=True`

torchtune/training/quantization.py
gau-nernst: I think `PrecisionTrainingQuantizer` is not needed atm. We can revisit it in the future once there are more things like this in the future (like you said, realistically I can only think of int8 and fp8 for now, but it may change)

felipemello1:
I tested it on multiple other models. Main issues:

1) For LoRA/QLoRA, there is no WPS improvement, which is different from your results
image

2) compile and packed are requisites. We should add it to the recipe __init__ and raise an error if mixed_precision.enabled=True but compile=False or packed=False.
image

3) Lets also add the changes to lora_distributed, if you dont midn

Lets address these and I am good to merge. I can edit all configs in a follow up PR :)

Thanks again @gau-nernst  :)

gau-nernst:
1. This PR does not support LoRA/QLoRA. Because now we use module-swap to inject INT8 matmul, we can't replace `LoRALinear` with the custom INT8 Linear module. Possible solutions
  a. Make base weight of `LoRALinear` a separate `nn.Linear` module. I did it here to demonstrate https://github.com/pytorch/torchtune/compare/main...gau-nernst:qlora (you only need to look at `torchtune/modules/peft/lora.py`). The QLoRA result in this PR description is done using this branch (as stated). However, Evan has some reservations as outlined here https://github.com/pytorch/torchtune/pull/1931#pullrequestreview-2439451976
  b. Another option is to propagate the INT8 flag to `LoRALinear`, and `LoRALinear` will directly call `torchao.prototype.quantized_training.int8_mixed_precision._Int8MixedPrecisionTrainingLinearFunction.apply()`. This is also not exactly ideal. Open to other options.
2. ok
3. Since this PR doesn't support LoRA/QLoRA, this is not necessary.

felipemello1:
sorry, I realized that you mentioned it about LoRA, but I didnt connect the dots. Let me talk to Evan and get back to you. Maybe we can land it for full finetuning only, since this is working well and we can follow up about LoRA. I will reply here on Monday.

ebsmothers:
Sorry for the delay here @gau-nernst and @felipemello1. After looking through the code again I think this is clean enough to move forward with (subject to Felipe's comment), even if we do only support full finetune. 

I need to do a bit more research to have a more informed opinion on the options laid out for handling LoRA. Regarding option (a): I took a look at the implementation in https://github.com/pytorch/torchtune/compare/main...gau-nernst:qlora and while I am still averse to state dict hooks it is quite clean. 

Separately I am still trying to understand pros and cons of module-swap vs tensor subclass -- I know tensor subclass needs custom handling for state dict, are there drawbacks to module-swap (I previously there were potentially challenges with FSDP, but actually it seems like that's not the case)? Also the point raised in https://github.com/pytorch/ao/pull/1179 about composability is not clear to me, isn't this a function of one quantization method (NF4) using tensor subclass and another (int8) using module swap? If I look at https://github.com/pytorch/ao/issues/987 it claims that composability of tensor subclass is better (at least with other tensor subclasses). Lmk if I'm missing the point here though.

Again, I don't want to block on any of this because unfortunately even the pretty trivial change to `LoRALinear` on your fork would break BC of checkpoints and may have other non-obvious implications for our various utilities and recipes that I am not yet aware of. 

So let's move forward with this for full finetune only (I will give a proper review shortly) and separately we can create an RFC to discuss pros and cons of a possible LoRA redesign.

gau-nernst:
@ebsmothers Thanks for your feedback. I will add "raise an error if mixed_precision.enabled=True but compile=False or packed=False", as commented by Felipe, soon.

Regarding module-swap vs tensor subclass, after all they are just different ways to inject custom logic to an existing model. Currently, we have 2 general use cases:

1. **Use quantized weights** (e.g. NF4, INT4/INT8). Only tensor subclass can do this (since we change storage of the weight). More generally, this can be considered any kind of custom parametrization (e.g. `torch.nn.utils.parameterize` - though this uses a different approach)
2. **Modify `F.linear()`**. Usually this goes hand-in-hand with (1) for inference of quantized models, since we want to call custom/specialized kernels for quantized weights. However, in cases where we don't use quantized weight (i.e. only modify `F.linear()`, weights are plain tensors), it's possible to use module-swap. e.g. QAT, INT8 training introduced here, `torchao.float8`.
    - Also want to repeat again that INT8 and FP8 training keeps **weights in FP32/BF16**, and quantize them dynamically during training. This is because we need high precision for weight udpate.

| Use case | Tensor subclass | Module swap |
|--------|--------|--------|
| Quantized weight | x |  |
| Modify `F.linear()` | x | x |

In terms of "composability", I don't think there is a clear answer about which one is better, because it largely depends on what kind of functionalities being composed together. The point about "composability" that I mentioned in pytorch/ao#1179 is specifically about custom weight-only quantization + low-bit compute. e.g.
- **Quantized weight** e.g. NF4 -> delegate to tensor subclass
- **Low-bit compute** e.g. INT8/FP8 matmul -> delegate to module-swap

This is possible since for weight-only quant, we don't have custom `F.linear()` kernel (just dequant then normal matmul). Hence, we can replace the normal matmul with the fast INT8/FP8 matmul.

Of course it's also possible to use INT8-matmul-subclass + NF4-subclass: INT8-matmul-subclass can hold NF4-subclass internally (similar to how dynamic activation quant is implemented in torchao). However, this can be a bit messy: INT8-matmul-subclass may need to be aware of the NF4-subclass implementation, multiple level of indirection is hard to debug, and possibly many others. So at least for this use case of "weight-only quant + low-bit compute", delegate one function to subclass and another to module-swap seems clearer.

Hope it somewhat clarifies the options presented here (and hope that it didn't confuse you further :sweat_smile:)

recipes/lora_finetune_single_device.py
```diff
@@ -441,6 +443,15 @@ def _setup_model(
                 model, auto_wrap_policy={modules.TransformerSelfAttentionLayer}
             )
 
+        if mixed_precision_cfg is not None and mixed_precision_cfg.get(
```
ebsmothers, line 446: We should revert the changes to the LoRA recipe, right?

torchtune/training/quantization.py
```diff
@@ -43,6 +48,22 @@
         Int8DynActInt4WeightQATQuantizer,
     )
 
+from torchtune.utils._version import torch_version_ge
+
+
+_SUPPORTS_INT8_MIXED_PRECISION_TRAINING = (
+    torch_version_ge("2.4.0")
```
ebsmothers, line 55: nit: can remove this check now (was probably more relevant when you first opened the PR)

torchtune/training/quantization.py
```diff
@@ -144,6 +166,86 @@ def quantize(self, model):
...
+        # custom filter_fn to work with torchtune's peft
+        def filter_fn(module: nn.Module, name: str) -> bool:
+            if isinstance(module, nn.Linear):
+                # skip LoRA adapters since they are too small, so the speedup will not
+                # outweight quantization overhead.
+                # also skip LM head since end2end speedup is slightly worse.
+                # there are also possible issues with tied word embeddings.
+                if (
+                    name.endswith(".lora_a")
+                    or name.endswith(".lora_b")
```
ebsmothers, line 234: Can probably remove this too, right?

ebsmothers (commented):
@gau-nernst @felipemello1 are we aiming to get this in for our next release? I left a few comments but I think (modulo a final round of testing, which I'm happy to help with) the remaining gap (at least for full finetune) to land should be pretty small.

gau-nernst:
I addressed all of the remaining comments. Please take another look if I missed out anything. Thank you.

Also added `mixed_precision` entry to all full-finetune and full-finetune-distributed configs. Did a few test runs and everything seems fine.

ebsmothers (commented):
Hi @gau-nernst I took a look through the code and no concerns from me, everything looks good. I think @felipemello1 landed a bunch of changes today that will cause merge conflicts on this PR, if it's too much of a headache just let us know and he can do the merge then push to your PR. Last thing: I was trying to test out the perf gains one final time before stamping but can't seem to find a good PyTorch nightly to run on (seems like lots of issues with nightlies lately). Lmk if you have a good stable env that I can test with (until then I will blindly try out various nightlies 😅 )

gau-nernst:
@ebsmothers PyTorch 2.5 should work. I have been testing with PyTorch 2.5. Nightly is not needed.

I can confirm that INT8 mixed-precision does not play well with bnb 8-bit AdamW. Loss will explode shortly after training starts. I'm not sure why. Switching to the plain PyTorch AdamW seems to solve the issue.

In terms of speed, indeed batch size must be sufficiently large to see speedup. I ran a few runs

| bsize x seq_len = num_tokens | BF16 speed | INT8-mp speed | Speedup | Screenshot
|--------|--------|--------|--------|--------|
| 2 x 2048 = 4,096 | 3400 tok/s | 3500 tok/s | +3% | [screenshot](https://github.com/user-attachments/assets/5fa9b3c6-b954-4805-9519-7da03b97ade4)
| 4 x 2048 = 8,192 | 3700 tok/s | 4400 tok/s | +19% | [screenshot](https://github.com/user-attachments/assets/9769d624-3ebe-4e7d-a696-bd5880038852)
| 8 x 2048 = 16,384 | 3900 tok/s | 4900 tok/s | +26% | [screenshot](https://github.com/user-attachments/assets/8eddba0d-495b-4d2d-971a-83c924da904f)
| 16 x 2048 = 32,768 | 4000 tok/s | 5100 tok/s | +28% | [screenshot](https://github.com/user-attachments/assets/7effdf8f-7cff-4293-8fc9-41832fbfec03)
| 4 x 8192 = 32,768 | 4100 tok/s | 5300 tok/s | +29% | [screenshot](https://github.com/user-attachments/assets/f8c178d0-3bde-4a89-9639-7b49bebc25b4)

We probably should add these 2 notes somewhere
1. Divergence when used together with 8-bit optimizers
2. Require >=4x2048 tokens/batch to see speedup

Perhaps you can advise me an appropriate place to put these notes?

ebsmothers:
@gau-nernst thanks for the detailed analysis. On a new env with PyTorch 2.5 and ao nightlies I see similar results to you for batch size 8, max seq len 2048. I have a few more comments -- the first one should be handled in this PR but the others are more for follow-ups.

1) Regarding where to put the notes on usage with 8-bit optimizers and the need for a large number of tokens in the batch, what do you think about a utility in `quantization.py`? Then we can pass the different config fields we need and give warnings (or potentially even throw an error) as needed. This can also include compile and packed btw. I think validation of the optimizer is a bit tricky (e.g. I don't think there is a non-hacky way to check that an arbitrary optimizer path is for an 8-bit optimizer, and we need to validate before instantiating the optimizer). However, given the loss divergence I have half a mind to explicitly require this to be used with `torch.optim.AdamW` only, and to raise an error otherwise. I know this is fairly restrictive, so open to a discussion on whether you think this is the right choice. (Alternatively we can modify the optimizer and raise a warning, but that feels a bit too sneaky to me.) For tokens/batch I would just log a warning if it's below some specified threshold, but leave it up to the user whether they want to do anything. This should be easy to infer from config (assuming it holds independent of the # of devices).

2) There are some other pieces that could require more detailed education too. Beyond the two that you mention, I think we should e.g. suggest people not to use this with Llama 1B models (at least when I ran it with those I saw slower training, probably due to the dominance of embedding and output projection layers). In terms of where to put something like this, we already have our [memory optimizations overview docs page](https://pytorch.org/torchtune/main/tutorials/memory_optimizations.html), and while this doesn't fit in there, I actually think it could be worthy of its own standalone docs page. There's a lot we can do here to educate folks, both in terms of the underlying implementation and on recommendations for usage. Depending on how far you want to take it, there's a world where we convert this into a blog (this is similar to what we did with e.g. our [knowledge distillation recipe](https://pytorch.org/blog/llama-into-torchtune/)).

3) Finally, one thing that makes me a bit nervous is that even with AdamW we see a bigger delta in loss curves than I would like (at least around 1e-1 on some early iterations). The good news is that in our test runs it does eventually converge, but this is a behavior I want us to be very explicit with people about. (Again this kinda ties into (2) around ensuring people know how to use the API correctly and what they should expect.)

Anyways, none of this is to push back too much on the PR. I think once (1) is addressed we should land it. It's an awesome feature that most other finetuning frameworks don't have, I'm really excited to support it. But given that we are one of the first to have it, it's up to us to set the right expectations for the community on how to use it (hence the importance of (2) and (3)).

gau-nernst:
@ebsmothers 

1. How about always add a warning inside `torchtune.training.quantization.Int8MixedPrecisionTrainingQuantizer`? Either in `__init__()` or in `prepare()`. This should warn about both 8-bit optimizer (for convergence) and large batch size (for speedup). `compile` and `packed` are currently checked at the recipe level, though it may make sense to consolidate all the checks within the quantizer class itself (then we possibly need to pass the whole `config` object to the quantizer, and the quantizer can check if the config makes sense / raise warning/error). 
  1.1 Like you have pointed out, checking for the optimizer is hard. We can check for the common ones (e.g. from bnb and torchao, since those are used in torchtune), which might be fine for 90% of the users, but it won't cover everything. Do you think if we log a warning, user will read it and it won't give a bad experience? versus straight-out error
2. In terms of education, I have this over at ao doc: https://github.com/pytorch/ao/tree/main/torchao/prototype/quantized_training#int8-mixed-precision-training, which is also referred to in `Int8MixedPrecisionTrainingQuantizer` docstring. You all had the "first user experience" with this in torchtune, so perhaps you can feedback to me what needs to be clearer/what kind of info is useful to users, and I'm happy to write a more detailed version catered to torchtune. You can decide where to put this to give the best visibility, I don't really mmind.
3. Yes, agree.

Also, regarding 1B, I tested it and indeed A100 does not see significant speedup, but on my personal GPU, the speedup is still observed at 4x2048.

> tune run full_finetune_single_device --config llama3_2/1B_full_single_device optimizer=torchao.prototype.low_bit_optim.AdamW8bit compile=True tokenizer.max_seq_len=2048 dataset.packed=True enable_activation_checkpointing=True metric_logger=torchtune.training.metric_logging.WandBLogger metric_logger.project=int8mp-1B metric_logger.name=int8mp_8bit-adamw_8x2048 mixed_precision.enabled=True batch_size=8

On A100

| bsize x seq_len = num_tokens | BF16 speed | INT8-mp speed | Speedup | Screenshot |
|--------|--------|--------|--------|--------|
| 4 x 2048 = 8,192 | 20,000 tok/s | 20,000 tok/s | +0.0% | [screenshot](https://github.com/user-attachments/assets/f9c2653e-6174-4613-9489-088f19d3b00a) |
| 8 x 2048 = 16,384 | 21,000 tok/s | 22,000 tok/s | +4.8% | [screenshot](https://github.com/user-attachments/assets/ea3574d4-43f7-48e0-ac43-d58c05ad05d7) |
| 16 x 2048 = 32,768 | 22,000 tok/s | 23,000 tok/s | +4.5% | [screenshot](https://github.com/user-attachments/assets/f5c92eff-aea3-40e4-a3bd-08028e6b3977) | 
| 32 x 2048 = 65,536 | 22,000 tok/s | 22,000 tok/s | +0.0% | [screenshot](https://github.com/user-attachments/assets/d10f9e45-7d6d-48c5-901e-d5898751cb8f)

On 4070Ti SUPER

| bsize x seq_len = num_tokens | BF16 speed | INT8-mp speed | Speedup | Screenshot |
|--------|--------|--------|--------|--------|
| 4 x 2048 = 8,192 | 7,000 tok/s | 10,000 tok/s | +43% | [screenshot](https://github.com/user-attachments/assets/adf21209-0814-4666-a4b0-a27375145ff8) |
| 8 x 2048 = 16,384 | 7,000 tok/s | 11,000 tok/s | +57% | [screenshot](https://github.com/user-attachments/assets/e8d19b58-0fd3-4fb8-9ed7-bfd75f55d87e)

My GPU OOM with larger batch sizes (only 16GB 😢). 4090 (24GB VRAM) should have the same speedup characteristic. This makes sense because INT8 tensor cores in consumer cards are rated at 4x faster than BF16/FP16 tensor cores, while enterprise GPUs are only rated at 2x speedup. But it may indicate that there is room for improvement on A100.

Anyway, this is to show that the speedup is different across GPUs. Perhaps once we work on (2), we can do more extensive end2end benchmarks across model sizes and GPUs.

ebsmothers:
@gau-nernst on (1): this sounds good to me. The one minor point I would change if possible is to pass each field explicitly to a utility like `validate_env_for_mp_training` or something (instead of passing the entire config). The args will be a bit of a hodge-podge (packed, optimizer path, etc), but that way we have an explicit contract on which fields we're validating against (and it can be seen directly from the callsite). Regarding the optimizer, I am fine to log a warning instead of raise an error. But yeah in general it's hard to comprehensively know all 8-bit optimizers -- hence why I think it's safest to use an allowlist-type approach (since we then say explicitly which ones are supported). Ultimately either way is OK as long as the process for inferring that someone is using a low-precision optimizer isn't too too hacky.

Re (2): yes, I think the ao readme page is great here. I think if we wanted to expand upon it in our torchtune docs, there are two potential directions we could go: lower-level (i.e. getting into the underlying implementation in more detail) or more user-focused (i.e. discussing different feature interactions, giving perf numbers, running experiments, etc.). If we combine what you already have with one or both of these, it'd make a nice standalone page on our live docs. But again we can figure out the details here in a follow-up. So yeah the ao documentation is already really clear in terms of usage, it's just a matter of seeing if there's a nice narrative we can provide around the feature to get people even more aware and interested to try it.

Thanks for the benchmarking results! This is interesting and not something I would have expected. And totally agree -- this is exactly the kind of thing that would be great to discuss in more detail in (2).

felipemello1:
> though it may make sense to consolidate all the checks within the quantizer class itself

I like that :)

PR #1568, opened 2024-09-12
[hotfix] Fix flex attention test

RdoubleA:
#### Context
- [ ] add a new feature
- [x] fix a bug
- [ ] other (please add here)

Flex attention test was failing on main because of the last minute change to `compile_friendly_flex_attention`. This broke the mocking in the tests. We quickly update to mock flex attention correctly.

#### Test plan
`pytest tests/torchtune/modules/test_attention_utils.py`


joecummings (approved):
merge merge merge when green

PR #1571, opened 2024-09-12
Instruct and chat datasets docs

RdoubleA:
#### Context
- [ ] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

Details around how to set up custom instruct and chat datasets and the expected format for each of these. As per #1529.

#### Test plan
doc build


docs/source/basics/instruct_datasets.rst
```diff
@@ -0,0 +1,233 @@
...
+
+.. code-block:: text
+
+    |  input          |  output          |
+    |-----------------|------------------|
+    | "user prompt"   | "model response" |
+
+The user input column is converted to a :class:`~torchtune.data.Message` with ``role="user"`` and the assistant
+output column is converted to a a :class:`~torchtune.data.Message` with ``role="assistant"``. These are then pass into
+the model tokenizer's ``tokenize_messages`` to encode and add appropriate model-specific special tokens.
```
ebsmothers, line 31: I think this part could be clearer. Why not just give an explicit example of each stage? Like you have the raw data format already, which is great. Then show how the same example would look in the Message representation, and finally the format after the messages have been tokenized. But as a first-time reader I see "tokenize_messages" and "model-specific special tokens" and I'm like wth is this

RdoubleA: no no to goal is to obscure further

docs/source/basics/chat_datasets.rst
```diff
@@ -0,0 +1,192 @@
+.. _chat_dataset_usage_label:
```
ebsmothers, line 1: Did we ever figure out how to get a nested toc tree? If not, is there a reason we're calling this section "Basics"? I would personally go with "Data and Datasets" or something like that

RdoubleA: once I get everything in I can make a pass and organize everything better, perhaps with @svekars 's help

docs/source/basics/instruct_datasets.rst
```diff
@@ -0,0 +1,262 @@
...
+and the assistant prompt is in another column.
+
+.. code-block:: text
+
+    |  input          |  output          |
+    |-----------------|------------------|
+    | "user prompt"   | "model response" |
+
+The user input column is converted to a :class:`~torchtune.data.Message` with ``role="user"`` and the assistant
+output column is converted to a a :class:`~torchtune.data.Message` with ``role="assistant"``.
```
ebsmothers, line 30:
```suggestion
output column is converted to a :class:`~torchtune.data.Message` with ``role="assistant"``.
```

docs/source/basics/instruct_datasets.rst
```diff
@@ -0,0 +1,262 @@
...
+        split="train"
+    )
+
+.. code-block:: yaml
+
+    # Tokenizer is passed into the dataset in the recipe
+    dataset:
+      _component_: torchtune.datasets.instruct_dataset
+      source: liweili/c4_200m
+      split: train
```
ebsmothers, line 88: tokenizer isn't in here

docs/source/basics/instruct_datasets.rst
```diff
@@ -0,0 +1,262 @@
...
+
+This will use the default column names "input" and "output". To change the column names, use the ``column_map`` argument.
```
ebsmothers, line 90: Maybe link to the section on this?

docs/source/basics/instruct_datasets.rst
```diff
@@ -0,0 +1,262 @@
...
+      data_files: data/my_data.json
+      split: train
+
+.. _column_map:
+
+Renaming columns
+----------------
+
+If the default column names are "input", "output" and you need to change them to something else,
+such as "prompt", "response" for example, define ``column_map`` as such:
```
ebsmothers, line 127: Give an explicit of what the raw dataset looks like in this case. The keys and values in this mapping feel backward to me and I think it may be a point of confusion for others

docs/source/basics/instruct_datasets.rst
```diff
@@ -0,0 +1,262 @@
...
+    )
+
+.. code-block:: yaml
+
+    # Tokenizer is passed into the dataset in the recipe
+    dataset:
+      _component_: torchtune.datasets.instruct_dataset
+      source: liweili/c4_200m
+      split: train
+      train_on_input: True
```
ebsmothers, line 182: Maybe print out the results of a sample and show that masking occurs in the positions of the user text

docs/source/basics/instruct_datasets.rst
```diff
@@ -0,0 +1,262 @@
...
+
+    from torchtune.models.gemma import gemma_tokenizer
+    from torchtune.datasets import instruct_dataset
+
+    g_tokenizer = gemma_tokenizer("/tmp/gemma-7b/tokenizer.model")
+    ds = instruct_dataset(
+        tokenizer=g_tokenizer,
+        source="liweili/c4_200m",
+        split="train",
+        new_system_prompt="You are a friendly AI assistant",
```
ebsmothers, line 202: Give a before and after here as well (can just be in terms of Messages or something)

docs/source/basics/chat_datasets.rst
```diff
@@ -0,0 +1,192 @@
...
+=============
+Chat Datasets
+=============
+
+Chat datasets involve multi-turn conversations (multiple back-and-forths) between user and assistant.
+This is more structured than freeform text association that models are typically pre-trained with,
+where they learn to simply predict the next token instead of responding accurately to the user.
+
+The primary entry point for fine-tuning with chat datasets in torchtune is the :func:`~torchtune.datasets.chat_dataset`
+builder. This lets you specify a local or Hugging Face dataset that follows the chat data format
```
SalmanMohammadi, line 12: For example, to use the X chat dataset from HuggingFace, use the following config...

^^ ideally one which needs just enough customisation to show the user how the parameters work - I think this would give sufficient info to a fair amount of our users already without needing to deepdive. I think Evan's comments on the other PR are good - lead with examples and detail for people who perhaps don't grok just from examples

docs/source/basics/instruct_datasets.rst
```diff
@@ -0,0 +1,262 @@
...
+      split: train
+      new_system_prompt: You are a friendly AI assistant
+
+.. _instruct_template:
+
+Instruct templates
+------------------
+
+Typically for instruct datasets, you will want to add a :class:`~torchtune.data.PromptTemplate` to provide task-relevant
+information. For example, for a grammar correction task, we may want to use a prompt template like :class:`~torchtune.data.GrammarErrorCorrectionTemplate`
```
ebsmothers, line 220: God the docstrings on these partials are horrible

docs/source/basics/chat_datasets.rst
```diff
@@ -0,0 +1,192 @@
...
+
+.. code-block:: python
+
+    from torchtune.models.phi3 import phi3_mini_tokenizer
+
+    p_tokenizer = phi3_mini_tokenizer("/tmp/Phi-3-mini-4k-instruct/tokenizer.model")
+    tokens, mask = p_tokenizer.tokenize_messages(messages)
+    print(tokens)
+    # [1, 32010, 29871, 13, 29984, 29896, 32007, 29871, 13, 32001, 29871, 13, 29909, 29896, 32007, 29871, 13]
+    print(p_tokenizer.decode(tokens))
```
SalmanMohammadi, line 49: can you keep the role headers in the decode call to make things clearer?

docs/source/basics/chat_datasets.rst
```diff
@@ -0,0 +1,192 @@
...
+    # '\nQ1 \n \nA1 \n'
+
+As an example, you can see the schema of the `SlimOrca dataset <https://huggingface.co/datasets/Open-Orca/SlimOrca-Dedup>`_.
+
+Specifying conversation style
+-----------------------------
+
+The structure of the conversation in the raw dataset can vary widely with different role names, different fields indicating
+where the message content name, and other ways. There are a few standardized formats that are common across many datasets.
```
SalmanMohammadi, line 58: where the message content name? 🤔🤔

also remove "other ways"

docs/source/basics/chat_datasets.rst
```diff
@@ -0,0 +1,192 @@
...
+
+The list of messages is tokenized by the model tokenizer with the appropriate model-specific special tokens added
+(such as beginning-of-sequence, end-of-sequence, and others).
+
+.. code-block:: python
+
+    from torchtune.models.phi3 import phi3_mini_tokenizer
+
+    p_tokenizer = phi3_mini_tokenizer("/tmp/Phi-3-mini-4k-instruct/tokenizer.model")
+    tokens, mask = p_tokenizer.tokenize_messages(messages)
```
felipemello1, line 46: nit: maybe the messages should be within the same python code

docs/source/basics/chat_datasets.rst
```diff
@@ -0,0 +1,192 @@
...
+        "messages": [
+            {
+                "role": "system" | "user" | "assistant",
+                "content": <message>,
+            },
+            ...
+        ]
+    }
+
+If your dataset does not fit one of the above conversation styles, then you will need to create a custom message transform.
```
SalmanMohammadi, line 104: going to link somewhere else for this at some point?

docs/source/basics/chat_datasets.rst
```diff
@@ -0,0 +1,192 @@
...
+
+Loading chat datasets from Hugging Face
```
SalmanMohammadi, line 106: not sure if you need this whole subsection if you just lead the page with a nice example

docs/source/basics/chat_datasets.rst
```diff
@@ -0,0 +1,192 @@
...
+      source: Open-Orca/SlimOrca-Dedup
+      conversation_column: conversations
+      conversation_style: sharegpt
+      split: train
+
+
+Loading local and remote chat datasets
+--------------------------------------
+
+To load in a local or remote dataset via https that has conversational data, you need to additionally specify the ``data_files`` and ``split``
```
felipemello1, line 140: Can you add code here showing how to create a dummy "data/my_data.json"? I dont know where "split" should go

Also not sure what it should like when its a .csv or .txt

docs/source/basics/chat_datasets.rst
```diff
@@ -0,0 +1,192 @@
...
+    # Tokenizer is passed into the dataset in the recipe
+    dataset:
+      _component_: torchtune.datasets.chat_dataset
+      source: Open-Orca/SlimOrca-Dedup
+      conversation_column: conversations
+      conversation_style: sharegpt
+      split: train
+
+
+Loading local and remote chat datasets
```
SalmanMohammadi, line 137: Think this section needs to be a lil beefier. Show the first couple lines of an example CSV, even just using `head mydata.csv -n 5` or whatever the command is, showing where it lives and how to setup the path correctly, etc etc. Feel like we get a lot of Q's on this and this is a good opportunity to be ice cold crystal clear

docs/source/basics/chat_datasets.rst
```diff
@@ -0,0 +1,192 @@
...
+``"json"``
+^^^^^^^^^^
+The associated message transform is :class:`~torchtune.data.JSONToMessages`. The expected format is:
+
+.. code-block:: python
+
+    {
+        "messages": [
+            {
+                "role": "system" | "user" | "assistant",
```
felipemello1, line 97: just a thought: it feels a bit bloated that we need a whole new class to rename human -> user, gpt -> assistant

docs/source/basics/instruct_datasets.rst
```diff
@@ -0,0 +1,262 @@
...
+      column_map:
+        input: prompt
+        output: response
+
+.. _train_on_input:
+
+Training on user input
+----------------------
+
+By default, user prompts are masked from the loss. To override this behavior, set ``train_on_input=True``.
```
felipemello1, line 160: I dont think that this section is necessary. Maybe leave train_on_input=False on others, and add this line explaining the arg

docs/source/basics/instruct_datasets.rst
```diff
@@ -0,0 +1,262 @@
...
+      source: liweili/c4_200m
+      split: train
+      train_on_input: True
+
+.. _system_prompt:
+
+Adding system prompts
+---------------------
+
+By specifying a system prompt, you will prepend a system :class:`~torchtune.data.Message` to each sample
```
felipemello1, line 189: I dont think that this section is necessary. Having "new_system_prompt" in the other examples should be enough

docs/source/basics/instruct_datasets.rst
felipemello1, line 220: Do we know if users use it? If i remember it, under the hood is just a prompt. do we need a class for it? It gives me an (overwhelming) feeling that datasets are complicated and have so many options and classes

docs/source/basics/instruct_datasets.rst
```diff
@@ -0,0 +1,262 @@
...
+
+.. _instruct_template:
+
+Instruct templates
+------------------
+
+Typically for instruct datasets, you will want to add a :class:`~torchtune.data.PromptTemplate` to provide task-relevant
+information. For example, for a grammar correction task, we may want to use a prompt template like :class:`~torchtune.data.GrammarErrorCorrectionTemplate`
+to structure each of our samples. Prompt templates are passed into the tokenizer and automatically applied to the dataset
+you are fine-tuning on.
```
ebsmothers, line 222: Would link to the other page for more details on prompt templates

docs/source/basics/chat_datasets.rst
```diff
@@ -0,0 +1,192 @@
+.. _chat_dataset_usage_label:
+
+=============
+Chat Datasets
+=============
+
+Chat datasets involve multi-turn conversations (multiple back-and-forths) between user and assistant.
```
ebsmothers, line 7: Give an example of a chat

docs/source/basics/chat_datasets.rst
```diff
@@ -0,0 +1,192 @@
...
+
+Chat datasets typically have a single column named "conversations" or "messages" that contains a list of messages on a single topic
+per sample. The list of messages could include a system prompt, multiple turns between user and assistant, and tool calls/returns.
+
+.. code-block:: text
+
+    |  conversations                         |
+    |----------------------------------------|
+    | [{"role": "user", "content": Q1},      |
+    |  {"role": "assistant", "content": A1}] |
```
ebsmothers, line 26: Maybe give a slightly less contrived example so that the back-and-forth and the sequential nature of the chat is clearer?

docs/source/basics/chat_datasets.rst
```diff
@@ -0,0 +1,192 @@
...
+    from torchtune.models.phi3 import phi3_mini_tokenizer
+
+    p_tokenizer = phi3_mini_tokenizer("/tmp/Phi-3-mini-4k-instruct/tokenizer.model")
+    tokens, mask = p_tokenizer.tokenize_messages(messages)
+    print(tokens)
+    # [1, 32010, 29871, 13, 29984, 29896, 32007, 29871, 13, 32001, 29871, 13, 29909, 29896, 32007, 29871, 13]
+    print(p_tokenizer.decode(tokens))
+    # '\nQ1 \n \nA1 \n'
+
+As an example, you can see the schema of the `SlimOrca dataset <https://huggingface.co/datasets/Open-Orca/SlimOrca-Dedup>`_.
```
ebsmothers, line 52: Feels like this should be closer to the top of this section

felipemello1 (commented):
Thanks Rafi! It is well written, but i will be a parrot and repeat myself:

IMO, a measurement of how good a tutorial is is how short it is. I think that there is an opportunity to do end2end with a monolithic example and almost no words paragraphs:

<python>
<create dummy data>
<instantiate dataset with a bunch of args, e.g. train_on_input, new_prompt, replace_columns, so the user can understand everything in one go>
<call dataset>
<show outputs / tags / tokens>
</python>

felipemello1 (commented):
Thanks Rafi! It is well written, but i will be a parrot and repeat myself:

IMO, a measurement of how good a tutorial is is how short it is. I think that there is an opportunity to do end2end with a monolithic example and almost no words paragraphs:

<python>
<create dummy data>
<instantiate dataset with a bunch of args, e.g. train_on_input, new_prompt, replace_columns, so the user can understand everything in one go>
<call dataset>
<show outputs / tags / tokens>
</python>

docs/source/basics/chat_datasets.rst
ebsmothers, line 58: This first sentence doesn't make a lot of sense

docs/source/basics/chat_datasets.rst
```diff
@@ -0,0 +1,192 @@
...
+
+.. code-block:: python
+
+    [
+        {
+            "role": "system" | "user" | "assistant" | "ipython",
+            "content": <message>,
+        },
+        ...
+    ]
```
ebsmothers, line 70: I'm a bit on the fence about this way of representing the different chat formats. While it's very general and the most concise I think it takes a second of staring at it to understand. Anyways I don't have a real ask to change this rn, but we should think about whether it makes sense to give explicit examples here as well.

docs/source/basics/chat_datasets.rst
ebsmothers, line 106: This is a bit out of order from how you do things on the instruct dataset page, right? There you lead with loading the dataset which I think is nicer. I assume you're doing it here because `conversation_style` hasn't been defined yet. But another way to frame it is just "let's load a dataset.. what are all these different fields?" then use that to motivate the exposition of `conversation_style`. Personally I think that's better cause you give people a working code snippet right from the outset

docs/source/basics/chat_datasets.rst
```diff
@@ -0,0 +1,192 @@
...
+Adding system prompts
+---------------------
+
+You can set a system prompt for your dataset similarly to :func:`~torchtune.datasets.instruct_dataset`. See :ref:`system_prompt` for more info.
+
+Chat templates
+--------------
+
+Chat templates are defined the same way as instruct templates in :func:`~torchtune.datasets.instruct_dataset`. See :ref:`instruct_template` for more info.
+
```
ebsmothers, line 189: Some of these might merit a tiny bit more explanation beyond just the link to the instruct docs. E.g. for system prompts: is it the first message of the chat or does it precede each user message? I guess chat templates is kind of a 1:1 mapping with our implementation, but I can imagine that being confusing for folks who are used to more of a separation between how chat and instruct data is prepared

docs/source/basics/instruct_datasets.rst
RdoubleA: will update those classes to just be builder functions at some point

docs/source/basics/instruct_datasets.rst
RdoubleA: removed this section as per Felipe's suggestion

docs/source/basics/chat_datasets.rst
RdoubleA: yes, after the messages docs lands

RdoubleA:
> are you mostly done with the changes for me to emulate with the preference/text completion docs?

Yes, this should be the bulk of it

docs/source/basics/chat_datasets.rst
```diff
@@ -0,0 +1,248 @@
...
+builder. This lets you specify a local or Hugging Face dataset that follows the chat data format
+directly from the config and train your LLM on it.
+
+Example chat dataset
+--------------------
+
+.. code-block:: bash
+
+    head data/my_data.json -n 22
+    # [
```
SalmanMohammadi, line 31: hashes for why?

docs/source/basics/chat_datasets.rst
```diff
@@ -0,0 +1,248 @@
...
+
+    dataset:
+      _component_: torchtune.datasets.chat_dataset
+      source: json
+      data_files: data/my_data.json
+      split: train
+      conversation_column: conversations
+      conversation_style: sharegpt
+      train_on_input: True
+      new_system_prompt: You are an AI assistant.
```
SalmanMohammadi, line 98: this is missing from the above example? or no?

docs/source/basics/chat_datasets.rst
```diff
@@ -0,0 +1,248 @@
...
+    # Tokenizer is passed into the dataset in the recipe
+    dataset:
+      _component_: torchtune.datasets.chat_dataset
+      source: Open-Orca/SlimOrca-Dedup
+      conversation_column: conversations
+      conversation_style: sharegpt
+      split: train
+
+
+Loading local and remote chat datasets
```
SalmanMohammadi, line 148: is this not already covered above?

docs/source/basics/chat_datasets.rst
```diff
@@ -0,0 +1,248 @@
...
+
+    [
+        {
+            "role": "system" | "user" | "assistant" | "ipython",
+            "content": <message>,
+        },
+        ...
+    ]
+
+``"sharegpt"``
```
SalmanMohammadi, line 199: show that we need conversation_style: sharegpt in config for this

docs/source/basics/chat_datasets.rst
```diff
@@ -0,0 +1,248 @@
...
+        "messages": [
+            {
+                "role": "system" | "user" | "assistant",
+                "content": <message>,
+            },
+            ...
+        ]
+    }
+
+If your dataset does not fit one of the above conversation styles, then you will need to create a custom message transform.
```
SalmanMohammadi, line 231: can you point to a custom message transform for chat data in the codebase? alpaca?

docs/source/basics/chat_datasets.rst
```diff
@@ -0,0 +1,248 @@
...
+
+
+Renaming columns
+----------------
+
+You can remap column names similarly to :func:`~torchtune.datasets.instruct_dataset`. See :ref:`column_map` for more info.
```
SalmanMohammadi, line 237: is this right? column map here here is just for `conversations` right? I think it'd be far simpler to show that with a small example

docs/source/basics/chat_datasets.rst
RdoubleA: probably better to just show the json file directly

docs/source/basics/chat_datasets.rst
RdoubleA: yeah I guess. was thinking it's worth having a direct section on it, vs just implicit in the example.

docs/source/basics/chat_datasets.rst
RdoubleA: hmm I don't think we have one...

docs/source/basics/chat_datasets.rst
RdoubleA: I think I need to refine the column handling in chat dataset. but yeah it's actually a bit different you're right

docs/source/basics/chat_datasets.rst
RdoubleA: ah, let me remove this. Mistral doesn't use system prompts

pbontrager (approved):
Thanks!

pbontrager (approved):
Thanks!

PR #1575, opened 2024-09-13
add fused = true to adam, except pagedAdam

felipemello1:
#### Context
- [x] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

update all optimizers to use fused, since its faster

#### Test plan
image


SalmanMohammadi:
So the only reason `fused=True` isn't default is because it's relatively new and torch wants to give some break in time?

SalmanMohammadi (approved):

PR #1585, opened 2024-09-14
nits in memory optimizations doc

ebsmothers:
Just fixing a couple commands and rewording things slightly in a couple places


SalmanMohammadi (approved):
these PRs are hit after hit after hit

Issue #1581, opened 2024-09-13
Generate Command phi3 Error


sgupta1007:
I have used  command  t`une run generate --config custom_quantization.yaml prompt='Explain some topic'`

Config custom_quantization.yaml
```
model:
  _component_: torchtune.models.phi3.qlora_phi3_mini

checkpointer:
  _component_: torchtune.utils.FullModelHFCheckpointer
  checkpoint_dir: //fine_tuned_phi/
  checkpoint_files: [
    hf_model_0001_0.pt,hf_model_0002_0.pt,adapter_0.pt
  ]
  model_type: PHI3
  output_dir: /fine_tuned_legal_phi/

device: cuda
dtype: bf16
seed: 1234

quantizer:
  _component_: torchtune.utils.quantization.Int4WeightOnlyQuantizer
  groupsize: 256

```

Error Flagged 
KeyError: 'PHI3``

sgupta1007:
How to generate inference using  phi3 finetune model fine tune through  qlora

felipemello1:
I believe it should be "model_type: PHI3_MINI"

https://github.com/pytorch/torchtune/blob/4fbe7b2d4956b3790c51d7a255c0040cf5c38fad/recipes/configs/phi3/mini_lora.yaml#L46

sgupta1007:
model type change resolved this error but lead to `FullModelHFCheckpointer.load_checkpoint() got an unexpected keyword argument 'weights_only'` error

felipemello1:
@joecummings , have you seen this before?

sgupta1007:
I am not able to use generate without quantization.

I will try to explain my approach for generation :

```
1. Perform phi3 qlora finetuning on 1 epoch
2. Supply the adapter and models weights to checkpointer files in config file 
3. Keep the model component as torchtune.models.phi3.qlora_phi3_mini. 
4. Run Generation Command tune run generate --config custom_quantization.yaml prompt='Explain some topic'
```

apthagowda97:
@sgupta1007  as adapter is already merged why we need to give adapter and model weights??

apthagowda97:
```
model:
  _component_: torchtune.models.llama3_1.llama3_1_8b

checkpointer:
  _component_: torchtune.utils.FullModelHFCheckpointer
  checkpoint_dir: /path/output/
  checkpoint_files: [
    hf_model_0001_0.pt,
    hf_model_0002_0.pt,
    hf_model_0003_0.pt,
[...]

PR #1603, opened 2024-09-16
Set max cache seq len in generate recipe

SalmanMohammadi:
#### Context
- [ ] add a new feature
- [x] fix a bug
- [ ] other (please add here)

#1581 

#### Changelog

We only need a cache size of `prompt_len + max_generated_tokens` for the generate recipe. This utilizes #1449 to significantly reduce unnecessary memory consumption.

Tested on a 4090 with Llama3.1 8B:
On this branch 
```bash
INFO:torchtune.utils._logging:Time for inference: 4.57 sec total, 25.39 tokens/sec
INFO:torchtune.utils._logging:Bandwidth achieved: 413.58 GB/s
INFO:torchtune.utils._logging:Memory used: 16.43 GB
```
on main:
```
torch.OutOfMemoryError: CUDA out of memory. Tried to allocate 1024.00 MiB. GPU 0 has a total capacity of 23.64 GiB of which 62.81 MiB is free. 
Process 539192 has 23.57 GiB memory in use. Of the allocated memory 23.02 GiB is allocated by PyTorch, and 121.49 MiB is reserved by PyTorch but unallocated. 
If reserved but unallocated memory is large try setting PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True to avoid fragmentation.  See documentation for Memory Management  (https://pytorch.org/docs/stable/notes/cuda.html#environment-variables)
```

#### Test plan

#### UX

### DIFF
--- a/recipes/generate.py
+++ b/recipes/generate.py
@@ -55,9 +55,9 @@ class InferenceRecipe:
         self._model = self._setup_model(
             model_cfg=cfg.model,
             model_state_dict=ckpt_dict[training.MODEL_KEY],
-            enable_kv_cache=cfg.enable_kv_cache,
         )
         self._tokenizer = config.instantiate(cfg.tokenizer)
+        self._enable_kv_cache = cfg.enable_kv_cache
 
     def _setup_model(
         self,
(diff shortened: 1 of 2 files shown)


recipes/generate.py
```diff
@@ -142,6 +137,15 @@ def generate(self, cfg: DictConfig):
 
         custom_generate_next_token = None
 
+        # Ensure the cache is setup on the right device, with only as many tokens as we need
+        if cfg.enable_kv_cache:
+            with self._device:
+                self._model.setup_caches(
+                    batch_size=1,
+                    dtype=self._dtype,
+                    decoder_max_seq_len=prompt.numel() + cfg.max_new_tokens,
```
joecummings, line 146: This is brilliant

joecummings (approved):
zoom zoom

Issue #1291, opened 2024-08-08
Fix IPOLoss in DPO recipe - IPOLoss uses average logprobs instead of summed


SalmanMohammadi:
See [this issue](https://github.com/huggingface/trl/issues/1677) and https://huggingface.co/blog/pref-tuning

The DPO recipe should pass the average, not summed logprobs into the IPOLoss - similar to SimPO.

PR #1615, opened 2024-09-17
remove ipo loss + small fixed

felipemello1:
#### Context
- [ ] add a new feature
- [ ] fix a bug
- [x] other (please add here)

IPO is currently not working properly (https://github.com/pytorch/torchtune/issues/12910) and needs more testing. This PR removes it before our new release

Also small fixes were done. Major one was removing loss parameters from DPO, so its easy to switch losses with diferent parameters in the config.


torchtune/rlhf/loss/dpo.py
```diff
@@ -160,87 +160,6 @@ def forward(
...
-class IPOLoss(nn.Module):
-    r"""
-    Identity Preference Optimization (IPO) Loss module: https://arxiv.org/abs/2310.12036.
-    Intuition from the paper:
-
-        (Given a policy :math:`\pi` and reference policy, :math:`\pi_{\text{ref}}`)
-
-        IPO learns from preferences dataset simply by regressing the gap between log-likelihood ratios
-
-        :math:`\log \bigg(\frac{(\pi(\text{chosen})}{\pi(\text{rejected})}\bigg)` and :math:`\log \bigg(\frac{\pi_{\text{ref}}(\text{chosen})}{\pi_{\text{ref}}(\text{rejected})} \bigg)`
```
SalmanMohammadi, line 172: my beautiful latex : (

SalmanMohammadi (approved):

torchtune/rlhf/loss/dpo.py
felipemello1: Its not a goodbye. Its just a "until next time".

Thanks for all the reviews and help with this @SalmanMohammadi !

kashif:
one thing about IPO was that we needed was to average over the length of the completions rather than just summing over the log-probs over the seq see: https://github.com/huggingface/trl/blob/main/trl/trainer/dpo_trainer.py#L1416-L1417

SalmanMohammadi:
> one thing about IPO was that we needed to average over the length of the completions rather than just summing over the log-probs over the seq see: https://github.com/huggingface/trl/blob/main/trl/trainer/dpo_trainer.py#L1416-L1417

Hey @kashif. Yep you're totally right. We actually had an issue for this (https://github.com/pytorch/torchtune/issues/1291), but didn't get round to it, and since we just put out a new release we didn't want to ship a version of the IPOLoss which didn't have this fix. 

If you'd be interested in contributing this fix for at all I can restore the loss on main now that the release is out.

kashif:
no worries! i'll let someone else take a shot at it! Already have a bunch on my plate haha but yeah if no one takes it, I can do it too...

SalmanMohammadi:
> no worries! i'll let someone else take a shot at it! Already have a bunch on my plate haha but yeah if no one takes it, I can do it too...

No sweat at all. It's a pretty straightforward fix since we already do something similar with the SimPO loss so hopefully it can make it to the next release.

kashif:
we need to get your sweet latex back in!

PR #1622, opened 2024-09-19
Fix kv-cacheing and bsz > 1 in eval recipe

### LINKED ISSUE #1600: Fix KV-cacheing + bsz > 1 with eval recipe
Currently, the eleuther recipe won't actually work with bsz>1 and KV-cacheing because some batches recieved by `_model_generate` are smaller than the batch size specified. In the code, there's this comment:

```python
        # Setup caches for a given batch size
        # Technically this is not necessary, but it's a good way to ensure that
        # the caches won't error on a different batch size. In addition, caches
        # are not needed for a regular model call, so we just setup here
```
Which also alludes to this, but it makes repeated calls to `setup_caches` to no avail, because there's currently no way to completely tear down caches. `reset_caches` just zeros them out.

cc @joecummings

SalmanMohammadi:
#### Context
- [ ] add a new feature
- [x] fix a bug
- [ ] other (please add here)

closes #1600

#### Changelog
* The eval harness will return `batch_size` specified prompts until the last batch, where it will return a smaller number of samples.
* This breaks our static KV-cacheing which is only setup for a single batch size. 
* To address this, we can manually pad out batches, and then remove the padded samples after generation.
* This could be wasteful for large batch sizes, and where the last batch may be mostly empty. We can kind of mitigate this by padding out with `self._tokenizer.eos_id` so we correctly trigger early stopping in generation.
* Expected test values are changed because `limit` has changed.

* Since swapping between KV-cacheing-inference and regular inference isn't supported yet, I've guarded against using generation tasks with non-generation tasks. This will be addressed in a follow up.
* I've also moved setup cache logic outside of `_model_generate` - it only needs to be called once if we're expecting a generation task.

Thanks to @baberabb for the advice here.

#### Test plan

#### UX

### DIFF
--- a/recipes/eleuther_eval.py
+++ b/recipes/eleuther_eval.py
@@ -181,6 +181,9 @@ class EleutherEvalRecipe(EvalRecipeInterface):
         self._quantization_mode = training.get_quantizer_mode(self._quantizer)
         self._enable_kv_cache = self._cfg.get("enable_kv_cache", True)
 
+        self._batch_size = self._cfg.batch_size
+        self._max_seq_length = self._cfg.get("max_seq_length", 4096)
+
         training.set_seed(seed=self._cfg.seed)
 
         checkpointer = config.instantiate(self._cfg.checkpointer)
@@ -236,8 +239,8 @@ class EleutherEvalRecipe(EvalRecipeInterface):
             self._model,
             self._tokenizer,
             device=self._device,
-            max_seq_length=self._cfg.max_seq_length,
-            batch_size=self._cfg.batch_size,
+            max_seq_length=self._max_seq_length,
+            batch_size=self._batch_size,
             dtype=self._dtype,
         )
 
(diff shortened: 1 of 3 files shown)


recipes/eleuther_eval.py
```diff
@@ -130,20 +123,11 @@ def _model_generate(
...
-        # the caches won't error on a different batch size. In addition, caches
-        # are not needed for a regular model call, so we just setup here
-        if self.enable_kv_cache:
-            with context.device:
-                self._model.setup_caches(batch_size=curr_batch_size, dtype=self._dtype)
+        maybe_padded_context = torch.nn.functional.pad(
+            context,
+            (0, 0, 0, self._batch_size - curr_batch_size),
+            value=self._tokenizer.eos_id,
+        )
```
ebsmothers, line 130: nit: add a comment explaining this cause otherwise it's not super obvious

PR #1623, opened 2024-09-19
Adding generic preference dataset builder

SalmanMohammadi:
#### Context
- [x] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

I was writing the docs for the preference datasets but noticed we don't have a generic builder to point to, so, here we are.

#### Changelog
* I've added a `preference_dataset` builder which uses `ChosenToRejectedMessages`. I don't think it's worth adding the stack exchange message format as an option at this time. 
* I've also exposed the HH RLHF dataset builder in our docs.

#### Test plan

#### UX

### DIFF
--- a/tests/torchtune/datasets/test_preference_dataset.py
+++ b/tests/torchtune/datasets/test_preference_dataset.py
@@ -8,10 +8,11 @@ from typing import Any, Mapping
 from unittest import mock
 
 import pytest
+from tests.common import ASSETS
 from tests.test_utils import DummyTokenizer
 from torchtune.data import Message
 from torchtune.data._common import CROSS_ENTROPY_IGNORE_IDX
-from torchtune.datasets._preference import PreferenceDataset
+from torchtune.datasets._preference import preference_dataset, PreferenceDataset
 from torchtune.modules.transforms import Transform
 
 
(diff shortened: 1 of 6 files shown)


tests/torchtune/datasets/test_preference_dataset.py
```diff
@@ -111,3 +112,46 @@ def test_get_item(self, mock_load_dataset, dialogue, expected):
         prompt, label = ds[0]["rejected_input_ids"], ds[0]["rejected_labels"]
         assert prompt == expected_rejected_tokens
         assert label == expected_rejected_labels
+
+    def test_load_local_json(self):
```
RdoubleA, line 116: nice, we hoenstly need more of these local dataset tests

torchtune/datasets/_preference.py
```diff
@@ -137,3 +140,136 @@ def _prepare_sample(self, sample: Mapping[str, Any]) -> Dict[str, List[int]]:
...
+                        "role": "user"
+                    },
+                    { "content": "Fix the hole.", "role": "assistant" }
+                ],
+                "rejected": [
+                    {
+                        "content": "What do I do when I have a hole in my trousers?",
+                        "role": "user"
+                    },
+                    { "content": "Take them off.", "role": "assistant" }
```
RdoubleA, line 239: 🫢 🫢 🫢

torchtune/datasets/_preference.py
```diff
@@ -137,3 +140,136 @@ def _prepare_sample(self, sample: Mapping[str, Any]) -> Dict[str, List[int]]:
...
+    >>> from torchtune.datasets import preference_dataset
+    >>> dataset = preference_dataset(
+    ...     tokenizer=tokenizer,
+    ...     source="json",
+    ...     data_files=my_preference_dataset.json,
+    ...     train_on_input=False,
+    ...     split="train",
+    >>> )
+    >>> tokenizer.decode(dataset[0]["chosen_input_ids"], skip_special_tokens=True)
+    What do I do when I have a hole in my trousers?Fix the hole.
```
RdoubleA, line 255: We should fix this trim whitespace issue at some point...

RdoubleA (approved):
do you want to also enable packing for preference datasets? should be easy to test and there's nothing blocking it

torchtune/datasets/_preference.py
SalmanMohammadi: actually renders for me as `'user\n\nWhat do I do when I have a hole in my trousers?assistant\n\nTake them off.'` with llama3 tokenizer and `skip_special_tokens=True`, and `'[INST] What do I do when I have a hole in my trousers?  [/INST] Take them off.'` with llama2, but I wrote it this way for consistency with the other examples.

SalmanMohammadi:
> do you want to also enable packing for preference datasets? should be easy to test and there's nothing blocking it

I'm actually pretty out of the loop on packed datasets, sorry. TLDR how to?

RdoubleA:
> I'm actually pretty out of the loop on packed datasets, sorry. TLDR how to?

just emulate this and that's all you need, then expose the packed parameter: https://github.com/pytorch/torchtune/blob/c5db813ce0473db090a4f1f6b450f559acac58e5/torchtune/datasets/_instruct.py#L257

then just run a DPO config with preference_dataset and use the override `dataset.packed=True tokenizer.max_seq_len=4096` and see if it packs successfully and trains. You can hardcode num packs to make it pack faster

PR #1628, opened 2024-09-19
Update full finetune distributed for MM

pbontrager:
#### Context
- [x] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

Updated fullfinetune_distributed to match the changes from #1548 . This adds gradient norm clipping, a configurable collate function, and batch unpacking for the model.

#### Test plan

#### UX

### DIFF
--- a/recipes/full_finetune_distributed.py
+++ b/recipes/full_finetune_distributed.py
@@ -20,7 +20,8 @@ from torch.distributed import destroy_process_group, init_process_group
 from torch.optim import Optimizer
 from torch.utils.data import DataLoader, DistributedSampler
 from torchtune import config, modules, training, utils
-from torchtune.data import padded_collate_packed, padded_collate_sft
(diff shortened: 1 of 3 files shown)


recipes/full_finetune_distributed.py
```diff
@@ -131,6 +137,7 @@ def __init__(self, cfg: DictConfig) -> None:
         self.total_epochs = cfg.epochs
         self.max_steps_per_epoch = cfg.max_steps_per_epoch
         self.global_step = 0
+        self._clip_grad_norm = cfg.get("clip_grad_norm", None)
```
ebsmothers, line 140: Can we add a test case similar to [this](https://github.com/pytorch/torchtune/blob/9f329a261fce1935b40029914e38ee31d952c50a/tests/recipes/test_full_finetune_single_device.py#L49)?

ebsmothers (approved):
One comment, otherwise looks good

PR #1637, opened 2024-09-20
Make ``q`` optional in ``sample``

joecummings:
#### Context
- [ ] add a new feature
- [ ] fix a bug
- [x] other (please add here)

User should not have to specify ``q`` in order to use the ``sample`` method. Therefore, we can provide the default. 

TODO: Look into whether setting the RNG is actually necessary or if a global Generator works fine, too.

#### Changelog
* Set default for ``q``

#### Test plan

```
>>> from torchtune.generation import sample
>>> import torch
>>> compiled_sample = torch.compile(sample, fullgraph=True, backend="aot_eager")
>>> logits = torch.empty(3, 3).uniform_(0, 1)
>>> compiled_sample(logits)
tensor([[1],
        [2],
        [0]], dtype=torch.int32)
>>> sample(logits)
tensor([[1],
        [2],
        [0]], dtype=torch.int32)
```

#### UX

### DIFF
--- a/tests/torchtune/generation/test_generation.py
+++ b/tests/torchtune/generation/test_generation.py
@@ -224,7 +224,7 @@ class TestGenerate:
         q = torch.empty(
             (1, 2000),
         ).exponential_(1)
-        token = sample(logits, q, temperature=1, top_k=1)
+        token = sample(logits, temperature=1, top_k=1, q=q)
         assert token.item() == 100
 
     @pytest.mark.parametrize(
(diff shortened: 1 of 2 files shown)


tests/torchtune/generation/test_generation.py
```diff
@@ -224,7 +224,7 @@ def test_sample_consistency(self):
         q = torch.empty(
             (1, 2000),
         ).exponential_(1)
-        token = sample(logits, q, temperature=1, top_k=1)
+        token = sample(logits, temperature=1, top_k=1, q=q)
```
pbontrager, line 227: Can we remove q here all together?

pbontrager (approved):
Please test this with all recipes that explicitly use "q". Otherwise looks good.

PR #1640, opened 2024-09-20
[BUG] Fix basic tokenizer no special tokens

joecummings:
#### Context
- [ ] add a new feature
- [x] fix a bug
- [ ] other (please add here)

**Goal**: Toggle `add_bos` and `add_eos` for Llama2 tokenizer.
**Problem**: This is not configurable from the Llama2 tokenizer `tokenize_messages`, which is now the default API.
**Root cause**: The underlying method used by the Llama2 tokenizer is `tokenize_messages_no_special_tokens`, which requires you to pass in `bos_id` and `eos_id` and always applies them. 

> **Side quest**: Turns out the masking for bos_id and eos_id using this function was INCORRECT (it would always apply the mask from the message instead of the proper mask for the special tokens. I fixed this.

**Solution**: I made the `bos_id` and `eos_id` optional for users to pass in. If they passed it in, I apply them. Otherwise, the method will not add the bos/eos tokens. 

#### Changelog
* Make `bos_id` and `eos_id` optional. If they are specified, then I do add them. Otherwise, I don't
* Update all corresponding tests
* Update docstring for tokenize_messages_no_special_tokens

#### Test plan

#### UX

### DIFF
--- a/tests/torchtune/models/gemma/test_gemma_tokenizer.py
+++ b/tests/torchtune/models/gemma/test_gemma_tokenizer.py
@@ -240,6 +240,6 @@ class TestGemmaTokenizer:
             2,
         ]
         # Mask user, unmask assistant, add EOS token
-        expected_mask = [True] * 75 + [False] * 124 + [True]
+        expected_mask = [True] * 75 + [False] * 125
         assert expected_tokens == tokens
         assert expected_mask == mask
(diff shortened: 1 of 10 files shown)


tests/torchtune/models/gemma/test_gemma_tokenizer.py
```diff
@@ -239,6 +239,7 @@ def test_tokenize_messages(self, tokenizer):
             4,
             2,
         ]
-        expected_mask = [True] * 75 + [False] * 125
+        # Mask user, unmask assistant, add EOS token
```
joecummings, line 242: These were WRONG

ebsmothers (approved):
Great work

PR #1651, opened 2024-09-22
add BlockMask to batch_to_device

felipemello1:
#### Context
- [ ] add a new feature
- [x] fix a bug
- [ ] other (please add here)

batch_to_device only checks dict and tensors. When input is BlockMask (packed=True), it triggers the RaiseError condition.

#### Changelog

- Improve error message
- Add BlockMask as one of the isinstance checks
- Add file _import_guard.py to check for flex attention, since we have two different files checking it, and we should have only one source of truth

#### Test plan

After this PR, the code below runs well
`tune run full_finetune_single_device --config llama3_1/8B_full_single_device dataset.packed=True tokenizer.max_seq_len=4096`

If I remove the check for BlockMask, the error message is improved:
```
ValueError: To use batch_to_device, all elements in the batch must be a dict or Tensor.
Got key "mask" with value of type <class 'torch.nn.attention.flex_attention.BlockMask'>
```

### DIFF
--- a/torchtune/utils/_device.py
+++ b/torchtune/utils/_device.py
@@ -154,5 +154,5 @@ def batch_to_device(batch: dict, device: torch.device) -> None:
         else:
             raise ValueError(
                 f"""To use batch_to_device, all elements in the batch must be a dict or Tensor.
-Got key {k} with value {v} of type {type(v)}"""
+Got key "{k}" with value of type {type(v)}"""
             )
--- a/torchtune/utils/_device.py
+++ b/torchtune/utils/_device.py
@@ -9,6 +9,13 @@ from typing import Optional
 
 import torch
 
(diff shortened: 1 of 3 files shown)


torchtune/utils/_device.py
```diff
@@ -142,7 +149,10 @@ def batch_to_device(batch: dict, device: torch.device) -> None:
             batch_to_device(v, device)
         elif isinstance(v, torch.Tensor):
             batch[k] = v.to(device)
+        elif _SUPPORTS_FLEX_ATTENTION and isinstance(v, BlockMask):
```
RdoubleA, line 152: Could you just combine this case and the above with `or`?

RdoubleA (approved):

PR #1656, opened 2024-09-23
Skip entire header for llama3 decode

RdoubleA:
#### Context
- [ ] add a new feature
- [x] fix a bug
- [ ] other (please add here)

With `skip_special_tokens=True`, the llama3 tokenizer still included the role name because it is part of the header but is not a special token. This PR filters out all tokens between start_header_id and end_header_id. Bonus: also expose `skip_special_tokens` in Phi3MiniTokenizer for consistency.

Before:
```
user

I can see the sun. But even if I cannot see the sun, I know that it exists.assistant

And to know that the sun is there - that is living.
```

After:
```
I can see the sun. But even if I cannot see the sun, I know that it exists.And to know that the sun is there - that is living.
```

#### Test plan

#### UX


torchtune/models/llama3/_tokenizer.py
```diff
@@ -166,11 +166,34 @@ def decode(
...
+            mask = True
+            idx = 0
+            while idx < len(token_ids):
+                if token_ids[idx] == self.start_header_id:
+                    # Start masking out at beginning of header
+                    mask = False
+                elif token_ids[idx] == self.end_header_id:
+                    # Mask out end header id and "\n\n" after it, then reset mask
+                    not_header[idx] = False
+                    mask = True
```
felipemello1, line 180: So after we see "end_header_id", mask is set to True, and will never be False again, is that right? If so, can we just exist the while loop? Or we need to keep checking because of sample packing, for example?

torchtune/models/llama3/_tokenizer.py
```diff
@@ -166,11 +166,34 @@ def decode(
...
-        return self.tt_model.decode(
-            token_ids,
-            truncate_at_eos=truncate_at_eos,
-            skip_special_tokens=skip_special_tokens,
-        )
+        if skip_special_tokens:
+            not_header = [True] * len(token_ids)
+            mask = True
+            idx = 0
+            while idx < len(token_ids):
```
felipemello1, line 173: This feels like it would be more efficient with numpy arrays. What sucks is going list -> array -> list

But you could do something like:

```
indices_start = get_index_start_header(token_ids, start_header_id)
indices_end = get_index_start_header(token_ids, start_header_id)

for idx_start, index_end zip(indices_start, indices_end):
     mask[idx_start:index_end+1] = False
```

that way the python for-loop is over number of header ids, and not over indices

torchtune/models/llama3/_tokenizer.py
```diff
@@ -166,11 +166,34 @@ def decode(
         Returns:
             str: The decoded string.
         """
-        return self.tt_model.decode(
-            token_ids,
-            truncate_at_eos=truncate_at_eos,
-            skip_special_tokens=skip_special_tokens,
-        )
+        if skip_special_tokens:
```
felipemello1, line 169: will this test pass? Probably not well formulated, but i want to see if after every encode/decode we are adding \n\n

`text = decode(encode(text), skip_special_tokens=True)`

torchtune/models/llama3/_tokenizer.py
```diff
@@ -166,11 +166,34 @@ def decode(
...
+                    # Mask out end header id and "\n\n" after it, then reset mask
+                    not_header[idx] = False
+                    mask = True
+                    idx += 1
+                    continue
+                not_header[idx] = mask
+                idx += 1
+
+            return self.tt_model.decode(
+                [token_ids[i] for i, m in enumerate(not_header) if m],
```
felipemello1, line 187: I suggested defining token_ids = [token_ids[i] for i, m in enumerate(not_header) if m]

and having a single return outside, so we can remove the if/else and duplicated code

felipemello1 (commented):
some comments and questions

torchtune/models/llama3/_tokenizer.py
felipemello1, line 169: I think that a string here is necessary to explain why you need this extra logic for "skip_special_tokens" if decode already has this flag. In other words: why do we skip special tokens tokens in two different places? Is there a more elegant way to solve this, like adding the special token to the tt_model?

torchtune/models/llama3/_tokenizer.py
```diff
@@ -166,11 +182,18 @@ def decode(
...
+        # This is because removing all special tokens does not remove the role and
+        # whitespace added from the special tokens, i.e., the "user" and "\n\n" in
+        # "<|start_header_id|>user<|end_header_id|>\n\n"
+        decoded_string = self.tt_model.decode(
+            token_ids=token_ids,
             truncate_at_eos=truncate_at_eos,
-            skip_special_tokens=skip_special_tokens,
+            skip_special_tokens=False,
         )
+        if skip_special_tokens:
```
ebsmothers, line 194: So I understand that the goal here is just to fix `skip_special_tokens` for Llama3 tokenizer decode, but it seems to me like we are doing something very unexpected here with `skip_special_tokens`. (a) we have it defined on the base tokenizer and it's basically a no-op now, and (b) we are now inconsistent on whether this needs to be defined on the `ModelTokenizer` and `BaseTokenizer`. If it is a function of `tokenize_messages` on the `ModelTokenizer` moreso than the `BaseTokenizer`, maybe we should update the Protocol along with other callsites?

torchtune/models/llama3/_tokenizer.py
```diff
@@ -166,11 +182,18 @@ def decode(
         Returns:
             str: The decoded string.
         """
-        return self.tt_model.decode(
-            token_ids,
+        # We will remove special tokens manually via regex on the decoded string.
+        # This is because removing all special tokens does not remove the role and
+        # whitespace added from the special tokens, i.e., the "user" and "\n\n" in
+        # "<|start_header_id|>user<|end_header_id|>\n\n"
```
ebsmothers, line 188: I would maybe move this comment up to where you define `self._special_token_regex` and `self._special_token_header_regex`

RdoubleA: this is where it actually happens, so makes more sense to keep it here? no strong opinions

ebsmothers: Yeah I feel the same, fine to keep it here then

ebsmothers (approved):
Looks like the PR is out of sync with the internal diff? Otherwise it looks good to me, stamping so you're unblocked

PR #1660, opened 2024-09-23
Update validation, remove slicing logic from classes

joecummings:
#### Context
- [ ] add a new feature
- [x] fix a bug
- [ ] other (please add here)

#### Changelog
*

#### Test plan

#### UX

### DIFF
--- a/torchtune/modules/transformer.py
+++ b/torchtune/modules/transformer.py
@@ -429,20 +429,13 @@ class TransformerDecoder(nn.Module):
                 decoder_max_seq_len=self.decoder_max_cache_seq_len,
             )
 
-    def encoder_caches_are_enabled(self) -> bool:
-        """Checks if there are any :class:`~torchtune.modules.TransformerCrossAttentionLayer`,
-        or :class:`~torchtune.modules.fusion.FusionLayer` layers which have cache enabled.
-        """
-        return self.encoder_max_cache_seq_len is not None
-
-    def decoder_caches_are_enabled(self) -> bool:
-        """Check if there are any :class:`~torchtune.modules.TransformerCrossAttentionLayer`
-        layers which have cache enabled."""
-        return self.decoder_max_cache_seq_len is not None
+    def caches_are_enabled(self) -> bool:
+        """Check if the key value caches are setup."""
+        return self.layers[0].cache_enabled
 
     def reset_caches(self):
         """Reset the key value caches."""
-        if not (self.encoder_caches_are_enabled() or self.decoder_caches_are_enabled()):
+        if not self.caches_are_enabled():
             raise RuntimeError(
                 "Key value caches are not setup. Call ``setup_caches()`` first."
             )
@@ -476,6 +469,7 @@ class TransformerDecoder(nn.Module):
         self,
         seq_len: int,
         mask: Optional[torch.Tensor] = None,
+        encoder_input: Optional[torch.Tensor] = None,
         encoder_mask: Optional[torch.Tensor] = None,
         input_pos: Optional[torch.Tensor] = None,
     ):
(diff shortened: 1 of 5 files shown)


torchtune/modules/transformer.py
```diff
@@ -429,20 +429,13 @@ def setup_caches(
...
-        """
-        return self.encoder_max_cache_seq_len is not None
-
-    def decoder_caches_are_enabled(self) -> bool:
-        """Check if there are any :class:`~torchtune.modules.TransformerCrossAttentionLayer`
-        layers which have cache enabled."""
-        return self.decoder_max_cache_seq_len is not None
+    def caches_are_enabled(self) -> bool:
+        """Check if the key value caches are setup."""
+        return self.layers[0].cache_enabled
```
felipemello1, line 434: I see that in the old logic we checked encoder/decoder. It seems that this is not necessary, and we can always check layer 0

is it enough? It makes sense to me, just double checking

torchtune/modules/transformer.py
```diff
@@ -429,20 +429,13 @@ def setup_caches(
...
-        or :class:`~torchtune.modules.fusion.FusionLayer` layers which have cache enabled.
-        """
-        return self.encoder_max_cache_seq_len is not None
-
-    def decoder_caches_are_enabled(self) -> bool:
-        """Check if there are any :class:`~torchtune.modules.TransformerCrossAttentionLayer`
-        layers which have cache enabled."""
-        return self.decoder_max_cache_seq_len is not None
+    def caches_are_enabled(self) -> bool:
+        """Check if the key value caches are setup."""
```
felipemello1, line 433: nit: fine as it is, but maybe add context about how/when this is used. Something like: "useful during inference to xyz". Feel free to ignore

torchtune/modules/transformer.py
```diff
@@ -500,26 +495,23 @@ def _validate_inputs(
...
                 )
 
-        if self.encoder_caches_are_enabled():
-            if encoder_mask is None:
+            if encoder_input is None and encoder_mask:
                 raise ValueError(
-                    "KV-caches for cross-attention/fusion layers are setup for inference mode, causal masks must be provided!"
-                    " Use the `encoder_mask` arg to provide a causal mask."
+                    "KV-caches for cross-attention/fusion layers are setup for inference mode and you seem to be using"
+                    " encoder_input, causal masks must be provided! Use the `encoder_mask` arg to provide a causal mask."
```
felipemello1, line 508: nit: prob a period instead of comma would be better

torchtune/modules/transformer.py
```diff
@@ -500,26 +495,23 @@ def _validate_inputs(
...
+        if self.caches_are_enabled():
             if mask is None:
                 raise ValueError(
                     "KV-caches for self-attention layers are setup for inference mode, causal masks must be provided!"
                     " Use the `mask` arg to provide a causal mask."
                 )
 
-        if self.encoder_caches_are_enabled():
-            if encoder_mask is None:
+            if encoder_input is None and encoder_mask:
```
felipemello1, line 505: in there some good default for encoder_mask, like causal mask? If not, just mark as resolved

felipemello1 (commented):
Minor nits. I worry about the logic for checking if cache is enabled. I dont think it is robust enough.

I would like to see a test confirming that these errors are properly raised. Do you think its possible?

torchtune/modules/transformer.py
joecummings: I plan on removing / re-working most of this validation logic to have better defaults, but that will be a follow-up PR.

felipemello1 (approved):
stamping to unblock, but IMO adding a small test before merging would be much better

PR #1661, opened 2024-09-24
LoRA Builders for MM

pbontrager:
#### Context
- [x] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

#### Changelog
- CLIP lora encoder + lora mlp + lora attention
- Flamingo lora encoder + decoder + projection head
- Llama 3.1 updated shared lora util
- TODO: update lora recipes to match full finetune updates

#### Test plan

#### UX

### DIFF
--- a/recipes/lora_finetune_distributed.py
+++ b/recipes/lora_finetune_distributed.py
@@ -20,7 +20,8 @@ from torch.distributed import destroy_process_group, init_process_group
 from torch.optim import Optimizer
 from torch.utils.data import DataLoader, DistributedSampler
 from torchtune import config, modules, training, utils
-from torchtune.data import padded_collate_packed, padded_collate_sft
+from torchtune.config._utils import _get_component_from_path
+from torchtune.data import padded_collate_packed
 from torchtune.datasets import ConcatDataset
 from torchtune.modules.peft import (
     DoRALinear,
@@ -94,6 +95,10 @@ class LoRAFinetuneRecipeDistributed(FTRecipeInterface):
 
         - Logging. Terminal, Disk, WandB and TensorBoard are all supported.
 
+        - Gradient Clipping. Gradient clipping is supported using the ``clip_grad_norm`` flag. By default,
+            ``clip_grad_norm`` is set to ``None``. If you only want to log the grad norm, you can set
+            ``clip_grad_norm='inf'``.
+
     For a full list of example configs for this recipe, run ``tune ls`` on the command line. Each config
     has example commands for how to kick-off training.
 
@@ -104,6 +109,7 @@ class LoRAFinetuneRecipeDistributed(FTRecipeInterface):
         ValueError: If ``dtype`` is set to fp16.
         ValueError: If world_size is 1
         RuntimeError: If ``dtype`` is set to bf16 and the hardware does not support bf16.
+        RuntimeError: If ``left_pad_sequence`` is set as the data collator.
     """
 
     def __init__(self, cfg: DictConfig) -> None:
@@ -136,6 +142,7 @@ class LoRAFinetuneRecipeDistributed(FTRecipeInterface):
         self.total_epochs = cfg.epochs
         self.max_steps_per_epoch = cfg.max_steps_per_epoch
         self.global_step = 0
+        self._clip_grad_norm = cfg.get("clip_grad_norm", None)
 
         self._save_adapter_weights_only = cfg.get("save_adapter_weights_only", False)
         self._resume_from_checkpoint = cfg.resume_from_checkpoint
@@ -257,10 +264,12 @@ class LoRAFinetuneRecipeDistributed(FTRecipeInterface):
 
         # sampler and dataloader depend on the tokenizer and loss_fn and should be
         # setup after all of these are setup
+        collate_name = cfg.get("collate_fn", "torchtune.data.padded_collate_sft")
         self._sampler, self._dataloader = self._setup_data(
             cfg_dataset=cfg.dataset,
             shuffle=cfg.shuffle,
             batch_size=cfg.batch_size,
+            collate_fn=collate_name,
         )
 
         # Finally update the recipe state which can only be correctly set after all of the
@@ -535,6 +544,7 @@ class LoRAFinetuneRecipeDistributed(FTRecipeInterface):
         cfg_dataset: DictConfig,
         shuffle: bool,
         batch_size: int,
+        collate_fn: str,
     ) -> Tuple[DistributedSampler, DataLoader]:
         """
         All data related setup happens here. Currently this recipe only supports the
@@ -545,14 +555,19 @@ class LoRAFinetuneRecipeDistributed(FTRecipeInterface):
 
         if isinstance(cfg_dataset, ListConfig):
             datasets = [
-                config.instantiate(single_cfg_dataset, tokenizer=self._tokenizer)
+                config.instantiate(single_cfg_dataset, self._tokenizer)
                 for single_cfg_dataset in cfg_dataset
             ]
             ds = ConcatDataset(datasets=datasets)
             packed = False
         else:
-            ds = config.instantiate(cfg_dataset, tokenizer=self._tokenizer)
+            ds = config.instantiate(cfg_dataset, self._tokenizer)
             packed = cfg_dataset.get("packed", False)
+
+        # Instantiate collate_fn
+        if "left_pad_sequence" in collate_fn:
+            raise RuntimeError("left_pad_sequence collator is only for inference.")
+        collate_fn = _get_component_from_path(collate_fn)
 
         sampler = DistributedSampler(
             ds, num_replicas=world_size, rank=rank, shuffle=shuffle, seed=0
@@ -565,14 +580,12 @@ class LoRAFinetuneRecipeDistributed(FTRecipeInterface):
             # dropping last avoids shape issues with compile + flex attention
             drop_last=cfg_dataset.get("drop_last", True),
             collate_fn=partial(
-                padded_collate_sft,
+                collate_fn,
                 padding_idx=self._tokenizer.pad_id,
                 ignore_idx=self._loss_fn.ignore_index,
             )
             if not packed
-            else partial(
-                padded_collate_packed,
-            ),
+            else padded_collate_packed,
         )
 
         if self._is_rank_zero:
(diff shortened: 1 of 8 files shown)


torchtune/models/clip/_component_builders.py
```diff
@@ -140,3 +152,334 @@ def clip_mlp(in_dim: int, out_dim: int, hidden_dim: int, activation: nn.Module,
...
+            )
+    else:
+        mlp = clip_mlp(
+            in_dim=embed_dim,
+            hidden_dim=4 * embed_dim,
+            out_dim=embed_dim,
+            activation=activation(),
+            quantize_base=quantize_base,
+        )
+    transformer_layer = TransformerSelfAttentionLayer(
```
felipemello1, line 269: do you mind using ModuleList like in llama 3.1, so we can avoid using _copy?

torchtune/models/flamingo/_component_builders.py
```diff
@@ -197,21 +527,34 @@ def flamingo_decoder(
     for idx in range(1, num_layers + 1):
 
         # Self attention layers for text decoder
-        rope = RotaryPositionalEmbeddings(dim=head_dim, max_seq_len=max_seq_len, base=rope_base)
-        self_attn = MultiHeadAttention(
+        rope = Llama3ScaledRoPE(dim=head_dim, max_seq_len=max_seq_len, base=rope_base)
```
felipemello1, line 530: rope should be instantiated only once, ouside of the for loop, to avoid copies and extra memory

torchtune/models/clip/_component_builders.py
pbontrager: I'd rather leave this for a future depreciation of the layer option

felipemello1: does it use rope embeddings? If so, it will consume more memory. Either way works for me

recipes/lora_finetune_single_device.py
```diff
@@ -282,10 +284,12 @@ def setup(self, cfg: DictConfig) -> None:
 
         # Dataloader depends on the tokenizer and loss_fn and should be
         # setup after all of these are setup
+        collate_name = cfg.get("collate_fn", "torchtune.data.padded_collate_sft")
```
joecummings, line 287: Why is padded_collate_sft the default? I thought that didn't work

torchtune/models/clip/_component_builders.py
joecummings: Do it @pbontrager

recipes/lora_finetune_single_device.py
pbontrager: That's the standard collate for finetuning text models.

recipes/lora_finetune_distributed.py
```diff
@@ -545,15 +555,20 @@ def _setup_data(
 
         if isinstance(cfg_dataset, ListConfig):
             datasets = [
-                config.instantiate(single_cfg_dataset, tokenizer=self._tokenizer)
+                config.instantiate(single_cfg_dataset, self._tokenizer)
```
ebsmothers, line 558: This doesn't need to be keyword?

torchtune/models/clip/_component_builders.py
```diff
@@ -140,3 +152,334 @@ def clip_mlp(in_dim: int, out_dim: int, hidden_dim: int, activation: nn.Module,
...
+    num_layers: int,
+    num_heads: int,
+    activation: Callable = nn.SiLU,
+    cls_output_dim: int = 512,
+    attn_bias: bool = True,
+    out_indices: Optional[List[int]] = None,
+    output_cls_projection: bool = False,
+    max_num_tiles: int = 4,
+    in_channels: int = 3,
+    intermediate_act: torch.nn.Module = torch.nn.SiLU(),
```
ebsmothers, line 178: Looking at L171 and L178 makes me sad

torchtune/models/clip/_component_builders.py
```diff
@@ -140,3 +152,334 @@ def clip_mlp(in_dim: int, out_dim: int, hidden_dim: int, activation: nn.Module,
...
     return FeedForward(gate_proj=gate_proj, down_proj=down_proj, up_proj=None, activation=activation)
+
+
+# ------------------ LoRA CLIP ------------------
+
+
+def lora_clip_vision_encoder(
+    lora_modules: List[LORA_ATTN_MODULES],
+    apply_lora_to_mlp: bool = False,
+    apply_lora_to_output: bool = False,
```
ebsmothers, line 163: Do we just need to pass this for consistency even though it's a no-op? Would maybe raise an error if it's set to true or something

torchtune/models/flamingo/_encoder.py
```diff
@@ -19,22 +18,20 @@ class FlamingoProjectionHead(nn.Module):
...
         self,
-        layer: nn.Module,
-        num_layers: int,
+        layers: nn.Module,
         output: nn.Module,
         num_hidden_inputs: int = 0,
     ) -> None:
         super().__init__()
-        self.layers = _get_clones(layer, num_layers)
+        self.layers = nn.ModuleList(layers)
```
ebsmothers, line 34: Shouldn't layers be `List[nn.Module]` type then?

torchtune/models/flamingo/_component_builders.py
```diff
@@ -264,3 +238,507 @@ def flamingo_decoder(
...
+    return FlamingoProjectionHead(
+        layers=layers,
+        output=nn.Linear(proj_in, decoder_embed_dim),
+        num_hidden_inputs=num_hidden_inputs
+    )
+
+# ------------------ LoRA Flamingo ------------------
+
+
+class LoRATrainable(Enum):
```
ebsmothers, line 315: Where is this used?

torchtune/models/flamingo/_component_builders.py
```diff
@@ -264,3 +238,507 @@ def flamingo_decoder(
...
+    }
+
+    # clip encoder
+    clip_options = {
+        "tile_size": tile_size,
+        "patch_size": patch_size,
+        "embed_dim": clip_embed_dim,
+        "num_layers": clip_num_layers,
+        "num_heads": num_heads,
+        "activation": nn.GELU,
```
ebsmothers, line 414: Maybe it's deliberate but seems like this doesn't match the default in the CLIP builder

torchtune/models/clip/_component_builders.py
felipemello1: i like consistency, agree with raising an error saying that it is a no-op/not supported. We do it for all tied embedding models that dont support apply_lora_to_output

torchtune/models/flamingo/_component_builders.py
```diff
@@ -264,3 +238,507 @@ def flamingo_decoder(
...
+        projection_head = lora_flamingo_projection_head(**projection_options, **lora_options)
+    else:
+        projection_head = flamingo_projection_head(**projection_options)
+
+    return FlamingoEncoder(clip=clip, projection_head=projection_head)
+
+
+def lora_flamingo_decoder(
+    decoder_lora: bool,
+    fusion_lora: bool,
```
ebsmothers, line 444: not used?

torchtune/models/flamingo/_encoder.py
felipemello1: please look at the transformer, that supports all cases (List[nn.Module], nn.ModuleList, nn.module: https://github.com/pytorch/torchtune/blob/30b8519e86d2094879576e16ab1f8d21a6f6f9f7/torchtune/modules/transformer.py#L353

I am fine if in flamingo it only supports nn.ModuleList though, but i prefer the consistency

torchtune/models/flamingo/_component_builders.py
```diff
@@ -264,3 +238,507 @@ def flamingo_decoder(
...
+                lora_dropout=lora_dropout,
+                use_dora=use_dora,
+            )
+        else:
+            mlp = llama3_mlp(dim=embed_dim, hidden_dim=hidden_dim, quantize_base=quantize_base)
+        decoder_layer = TransformerSelfAttentionLayer(
+            attn=self_attn,
+            mlp=mlp,
+            sa_norm=RMSNorm(dim=embed_dim, eps=1e-5),
+            mlp_norm=RMSNorm(dim=embed_dim, eps=1e-5),
```
ebsmothers, line 554: super nit: use the same format for eps (below it's `1e-05`)

torchtune/models/flamingo/_component_builders.py
```diff
@@ -264,3 +238,507 @@ def flamingo_decoder(
...
+    *,
+    # projection head parameters
+    num_layers: int,
+    num_heads: int,
+    decoder_embed_dim: int,
+    clip_embed_dim: int,
+    num_hidden_inputs: int,
+    # LoRA args
+    apply_lora_to_mlp: bool,
+    apply_lora_to_output: bool,
```
ebsmothers, line 644: Any particular reason these are now keyword-only args as opposed to how we have them elsewhere?

torchtune/models/flamingo/_component_builders.py
```diff
@@ -264,3 +238,507 @@ def flamingo_decoder(
...
+            attn=self_attn,
+            mlp=mlp,
+            sa_norm=RMSNorm(dim=embed_dim, eps=1e-5),
+            mlp_norm=RMSNorm(dim=embed_dim, eps=1e-5),
+        )
+
+        # cross attention layers, mixing text and vision,
+        # placed every `fusion_interval` layers
+        if idx % fusion_interval == 0:
+            attn = lora_llama3_attention(
```
ebsmothers, line 560: Could maybe use partials to reduce the duplicative code here? But nbd either way

torchtune/models/flamingo/_component_builders.py
```diff
@@ -264,3 +238,507 @@ def flamingo_decoder(
...
+                    lora_dropout=lora_dropout,
+                    use_dora=use_dora,
+                )
+            else:
+                mlp = llama3_mlp(dim=embed_dim, hidden_dim=hidden_dim, quantize_base=quantize_base)
+            xattn_layer = TransformerCrossAttentionLayer(
+                attn=attn,
+                mlp=mlp,
+                ca_norm=RMSNorm(dim=embed_dim),
+                mlp_norm=RMSNorm(dim=embed_dim),
```
ebsmothers, line 594: Is it deliberate to have different eps here vs in self-attention layers?

ebsmothers (approved):

recipes/lora_finetune_distributed.py
pbontrager: Some datasets call it tokenizer and others call it transforms

torchtune/models/clip/_component_builders.py
pbontrager: What can't they be true?

torchtune/models/flamingo/_component_builders.py
pbontrager: Deliberate

torchtune/models/flamingo/_component_builders.py
pbontrager: will be used in builders

PR #1664, opened 2024-09-24
Readme Updates

ebsmothers:
Long-overdue updates to our README. Changes include:

1) Updating memory and tokens/sec numbers to report for Llama3.1 instead of Llama2.

2) Providing a (n imo) clearer version of our table of supported recipes

3) Generally reorganizing things. Now the flow is

- Intro
  - Models
  - Recipes
  - Memory efficiency and perf
- Install
- Getting Started
  - Downloading models
  - Running recipes
  - Modifying configs
- Llama3 and 3.1
- Community
  - Community Contributions
- Acknowledgments
- License

4)  Updating the intro bullet points slightly for clarity.. now the first three more explicitly outline the three tables that follow
5) Removing the design principles section

A bunch of other small/cosmetic changes, but these are the main ones


README.md
```diff
@@ -192,17 +129,24 @@ options:
 ...
 ```
 
-You can also install the latest and greatest torchtune has to offer by [installing a nightly build](https://pytorch.org/torchtune/main/install.html).
+Alternatively, you can install a nightly build of torchtune to gain access to the latest features not yet available in the stable release.
+
+```bash
+pip install --pre torchtune --extra-index-url https://download.pytorch.org/whl/nightly/cpu
```
felipemello1, line 135: should we only recommend torchtune nightlies with torch/torchao/torchvision nightlies? In other words, is it possible that someone will install stable torchtune, nightlies torch, and have issues?

README.md
```diff
@@ -192,17 +129,24 @@ options:
...
 ## Get Started
 
-To get started with fine-tuning your first LLM with torchtune, see our tutorial on [fine-tuning Llama2 7B](https://pytorch.org/torchtune/main/tutorials/first_finetune_tutorial.html). Our [end-to-end workflow](https://pytorch.org/torchtune/main/tutorials/e2e_flow.html) tutorial will show you how to evaluate, quantize and run inference with this model. The rest of this section will provide a quick overview of these steps with Llama2.
+To get started with torchtune, see our [Fine-Tune Your First LLM Tutorial](https://pytorch.org/torchtune/main/tutorials/first_finetune_tutorial.html). Our [end-to-end workflow](https://pytorch.org/torchtune/main/tutorials/e2e_flow.html) tutorial will show you how to evaluate, quantize and run inference with this model. The rest of this section will provide a quick overview of these steps with Llama3.
+
+If you have a more custom workflow or need additional information on torchtune components and recipes, please check out our documentation page
 
-&nbsp;
 
 ### Downloading a model
```
felipemello1, line 152: downloading a model recommends llama 3 instead of 3.1

README.md
```diff
@@ -278,36 +222,87 @@ Check out `tune --help` for all possible CLI commands and options. For more info
 
 &nbsp;
 
-## Design Principles
+---
+
+## Llama3 and Llama3.1
```
felipemello1, line 232: why llama 3? I think its fine to just use 3.1. This may get stable soon, so may keeping it just as "llama" is better

README.md
ebsmothers: Yeah this is a good point. Will update

README.md
```diff
@@ -278,36 +222,87 @@ Check out `tune --help` for all possible CLI commands and options. For more info
...
+
+torchtune supports fine-tuning for the Llama3/Llama3.1 8B, 70B, and 405B size models. You can fine-tune the 8B model with LoRA, QLoRA and full fine-tunes on one or more GPUs. You can also fine-tune the 70B model with QLoRA on a single device or LoRA and full-finetunes on multiple devices. Finally, you can fine-tune the 405B model on a single node with QLoRA. For all the details, take a look at our [tutorial](https://pytorch.org/torchtune/main/tutorials/llama3.html).
```
felipemello1, line 234: imo, less is more. No need to specify 8b/70b/405b in 3 different lines. Something like this should be good

"torchtune supports fine-tuning for the Llama3/Llama3.1 8B, 70B, and 405B size models with LoRA, QLoRA and full fine-tunes on one or more GPUs, depending on the model size. For all the details, take a look at our [tutorial](https://pytorch.org/torchtune/main/tutorials/llama3.html)."

README.md
ebsmothers: Yeah I will drop 3, probably will keep 3.1 in just to be explicit (otherwise it's not clear what we're referring to). One thing is that the tutorial in the docs is still for Llama3, not 3.1. So strictly speaking it's not 100% correct, but maybe that's a minor point

RdoubleA:
I think we absolutely need a section in getting started on running on a custom dataset, maybe even listing out the datset types we support and pointing to docs for more details.

I don't see any mention of multimodal except in the install for why we depend on torchvision, we should be highlighting this somewhere

README.md
```diff
@@ -278,36 +222,87 @@ Check out `tune --help` for all possible CLI commands and options. For more info
...
-torchtune embodies PyTorch’s design philosophy [[details](https://pytorch.org/docs/stable/community/design.html)], especially "usability over everything else".
+```bash
+tune run --nproc_per_node 2 lora_finetune_distributed --config llama3_1/8B_lora
+```
 
-### Native PyTorch
+LoRA 70B
+
+Note that the download command for the Meta-Llama3 70B model slightly differs from download commands for the 8B models. This is because we use the HuggingFace [safetensor](https://huggingface.co/docs/safetensors/en/index) model format to load the model. To download the 70B model, run
+```bash
```
felipemello1, line 277: i dont think that this warning is really necessary. But we should add a link to our models.api, that talks about downloading every model, not only llama. IMO, llama here should just be an example, not a full tutorial to download every llama model.

README.md
```diff
@@ -278,36 +222,87 @@ Check out `tune --help` for all possible CLI commands and options. For more info
...
+tune download meta-llama/Meta-Llama-3.1-70b --hf-token <> --output-dir /tmp/Meta-Llama-3.1-70b --ignore-patterns "original/consolidated*"
```
felipemello1, line 278: Not sure if i fully follow why for llama 8b we just share tune run, but for 70b, we have also download instructions here.

README.md
```diff
@@ -278,36 +222,87 @@ Check out `tune --help` for all possible CLI commands and options. For more info
...
-torchtune is a native-PyTorch library. While we provide integrations with the surrounding ecosystem (e.g. Hugging Face Datasets, EleutherAI Eval Harness), all of the core functionality is written in PyTorch.
+Then, a finetune can be kicked off:
 
-### Simplicity and Extensibility
+```bash
+tune run --nproc_per_node 8 lora_finetune_distributed --config llama3_1/70B_lora.yaml
+```
 
-torchtune is designed to be easy to understand, use and extend.
+You can find a full list of all our Llama3 configs [here](recipes/configs/llama3) and Llama3.1 configs [here.](recipes/configs/llama3_1)
```
felipemello1, line 287: this is nice, but i dont think that we should share only about llama. Maybe shoutout llama 3.1. and replace llama3 with just recipes/configs.

README.md
```diff
@@ -278,36 +222,87 @@ Check out `tune --help` for all possible CLI commands and options. For more info
...
+- [Weights & Biases](https://wandb.ai/site) for [logging](https://pytorch.org/torchtune/main/deep_dives/wandb_logging.html) metrics and checkpoints, and tracking training progress
+- [Comet](https://www.comet.com/site/) as another option for [logging](https://pytorch.org/torchtune/main/deep_dives/comet_logging.html)
+- [ExecuTorch](https://pytorch.org/executorch-overview) for [on-device inference](https://github.com/pytorch/executorch/tree/main/examples/models/llama2#optional-finetuning) using fine-tuned models
+- [bitsandbytes](https://huggingface.co/docs/bitsandbytes/main/en/index) for low memory optimizers for our [single-device recipes](recipes/configs/llama2/7B_full_low_memory.yaml)
+- [PEFT](https://github.com/huggingface/peft) for continued fine-tuning or inference with torchtune models in the Hugging Face ecosystem
 
 &nbsp;
 
-## Community Contributions
+### Community Contributions
```
felipemello1, line 308: are we keeping this? I remember we had a discussion about it a while ago

ebsmothers:
> I think we absolutely need a section in getting started on running on a custom dataset, maybe even listing out the datset types we support and pointing to docs for more details.
> 
> I don't see any mention of multimodal except in the install for why we depend on torchvision, we should be highlighting this somewhere

I would punt that one to you and/or @pbontrager. The goal of these changes is just to get us from 6 months behind to approximately present-day. And imo custom datasets can get hairy quickly, I think the readme should be simple and to-the-point, leveraging our live docs as a reference. E.g. even axolotl's readme basically only has two sentences on datasets and just points to their docs.

README.md
ebsmothers: Yeah I planned to keep it for now

README.md
```diff
@@ -42,113 +34,63 @@ torchtune currently supports the following models.
...
-|:-:|:-:|:-:|:-:|:-:|:-:|
-| 1 x RTX 4090 |     QLoRA **         |  Llama2-7B      |    Batch Size = 4, Seq Length = 2048   | 12.3 GB | 3155  |
-| 1 x RTX 4090 |     LoRA          |  Llama2-7B      |    Batch Size = 4, Seq Length = 2048   | 21.3 GB | 2582  |
-| 2 x RTX 4090 |     LoRA          |  Llama2-7B      |    Batch Size = 4, Seq Length = 2048   | 16.2 GB | 2768  |
-| 1 x RTX 4090 |   Full finetune *   |  Llama2-7B      |    Batch Size = 4, Seq Length = 2048   | 24.1 GB | 702  |
-| 4 x RTX 4090 |   Full finetune   |  Llama2-7B      |    Batch Size = 4, Seq Length = 2048   | 24.1 GB | 1388  |
-| 8 x A100     |     LoRA          |  Llama2-70B     |    Batch Size = 4, Seq Length = 4096   | 26.4 GB | 3384  |
-| 8 x A100     |   Full Finetune *   |  Llama2-70B     |    Batch Size = 4, Seq Length = 4096   | 70.4 GB | 2032  |
+| Fine-tuning method                          | Devices | Recipe  | Example config(s) |
+|:-:|:-:|:-:|:-:|
```
felipemello1, line 56: I think its worth adding another column with [Speed optimized, memory optimized], or it may not be clear for the user why examples are duplicated and numbers are different

felipemello1: probably worth also adding bsz and max_seq_len

ebsmothers: On the second point, I followed your advice and use fixed bsz and max_seq_len for all examples. I call this out in the note just above the table. I left out the speed vs memory optimized column because it's really only applicable to like 50% of the rows, so it might be redundant. Tried to instead include the hardware type to delineate that one can be run in a more memory-constrained environment

README.md
```diff
@@ -192,17 +139,16 @@ options:
...
 &nbsp;
 
 ---
 
 ## Get Started
 
-To get started with fine-tuning your first LLM with torchtune, see our tutorial on [fine-tuning Llama2 7B](https://pytorch.org/torchtune/main/tutorials/first_finetune_tutorial.html). Our [end-to-end workflow](https://pytorch.org/torchtune/main/tutorials/e2e_flow.html) tutorial will show you how to evaluate, quantize and run inference with this model. The rest of this section will provide a quick overview of these steps with Llama2.
+To get started with torchtune, see our [Fine-Tune Your First LLM Tutorial](https://pytorch.org/torchtune/main/tutorials/first_finetune_tutorial.html). Our [end-to-end workflow](https://pytorch.org/torchtune/main/tutorials/e2e_flow.html) tutorial will show you how to evaluate, quantize and run inference with this model. The rest of this section will provide a quick overview of these steps with Llama3.
+
+If you have a more custom workflow or need additional information on torchtune components and recipes, please check out our documentation page
```
ebsmothers, line 150: TODO: update

joecummings (commented):
Controversial, but I would be in favor of dropping the full section on Llama 3 and Llama 3.1. We already say what models we support and can highlight new model updates at the top of the README in recent updates section

joecummings (approved):
meh

Issue #1540, opened 2024-09-11
Custom imports are broken


pbontrager:
If you make your own custom dataset.py or model.py and include the import path in the config, these imports will fail because `tune run` always sets the current working directory to torchtune. Since all of torchtune imports are available anywhere once installed, `tune run` should use the launch location as the cwd.

Potential complication - since recipes are not in the cwd, we'll have to test how recipe work being launched from a different directory.

RdoubleA:
Only single device was patched. distributed recipes still do not detect modules from cwd. potential issue: https://github.com/pytorch/pytorch/issues/77737#issuecomment-1402217701

PR #1667, opened 2024-09-24
Multimodal dataset builder + docs

RdoubleA:
#### Context
- [x] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

There is currently no way to load a custom multimodal dataset. The current workflow is users will write their own transform and builder - which is already a big lift - but on top of that our custom component imports don't even work (#1540). So basically, it's impossible to use anything other than llava instruct and cauldron. This sucks.

I added a quick `multimodal_chat_dataset` builder. It assumes conversational data with images, much like llava instruct and ShareGPT4V. To keep changes minimal, I only support ShareGPT format. We can expand to other chat formats or to instruct data later.

This unfortunately required me to generalize the LlavaInstructToMessages (cc @joecummings, I agree we shouldn't generalize too early but in this case I think it's warranted). Again, to keep changes minimal, I simply merged LlavaInstructToMessages with ShareGPTToMessages. There were identical in handling text, the only difference was the image loading logic.

With this builder, now I finally have something to point to and discuss in the multimodal dataset docs, and users have some avenue to finetune multimodal on their own data. A bit of extra work, but worthwhile imo.

#### Changelog
* Merge ShareGPTToMessages and LlavaInstructToMessages
* Merge any relevant tests
* Add `multimodal_chat_dataset` builder
* Add doc page on multimodal datasets

#### Test plan

#### UX

### DIFF
--- a/docs/source/basics/multimodal_datasets.rst
+++ b/docs/source/basics/multimodal_datasets.rst
@@ -4,7 +4,8 @@
 Multimodal Datasets
 ===================
 
-Multimodal datasets include both text and image data used to train VLMs. torchtune currently supports multimodal chat datasets.
+Multimodal datasets include more than one data modality, e.g. text + image, and can be used to train transformer-based models.
+torchtune currently only supports multimodal text+image chat datasets for Vision-Language Models (VLMs).
 
 The primary entry point for fine-tuning with multimodal datasets in torchtune is the :func:`~torchtune.datasets.multimodal.multimodal_chat_dataset`
 builder. This lets you specify a local or Hugging Face dataset that follows the multimodal chat data format
@@ -15,7 +16,9 @@ directly from the config and train your VLM on it.
 Example multimodal dataset
 --------------------------
 
-Here is an example of a multimodal chat dataset for a visual question-answering task.
+Here is an example of a multimodal chat dataset for a visual question-answering task. Note that there is a placeholder
+in the text, ``"<image>"`` for where to place the image tokens. This will get replaced by the image special token
+``<|image|>`` in the example below.
 
 .. code-block:: python
 
@@ -69,7 +72,7 @@ Here is an example of a multimodal chat dataset for a visual question-answering 
 
 .. code-block:: yaml
 
-    # In config
+    # In config - model_transforms takes the place of the tokenizer
     model_transform:
       _component_: torchtune.models.flamingo
       path: /tmp/Meta-Llama-3-8B-Instruct/original/tokenizer.model
@@ -102,6 +105,8 @@ and the user-assistant conversations are in another column.
     |  {"from": "gpt", "value": "A1"}]   |              |
 
 As an example, you can see the schema of the `ShareGPT4V dataset <https://huggingface.co/datasets/Lin-Chen/ShareGPT4V>`_.
+
+Currently, :func:`~torchtune.datasets.multimodal.multimodal_chat_dataset` only supports a single image path per conversation sample.
 
 
 Loading multimodal datasets from Hugging Face
--- a/docs/source/basics/multimodal_datasets.rst
+++ b/docs/source/basics/multimodal_datasets.rst
@@ -1,0 +1,285 @@
+.. _multimodal_dataset_usage_label:
+
+===================
+Multimodal Datasets
+===================
+
+Multimodal datasets include both text and image data used to train VLMs. torchtune currently supports multimodal chat datasets.
+
+The primary entry point for fine-tuning with multimodal datasets in torchtune is the :func:`~torchtune.datasets.multimodal.multimodal_chat_dataset`
+builder. This lets you specify a local or Hugging Face dataset that follows the multimodal chat data format
+directly from the config and train your VLM on it.
+
+.. _example_multimodal:
+
+Example multimodal dataset
+--------------------------
+
+Here is an example of a multimodal chat dataset for a visual question-answering task.
+
+.. code-block:: python
+
+    # data/my_data.json
+    [
+        {
+            "dialogue": [
+                {
+                    "from": "human",
+                    "value": "<image>What time is it on the clock?",
+                },
+                {
+                    "from": "gpt",
+                    "value": "It is 10:00 AM.",
+                },
+            ],
+            "image_path": "images/clock.jpg",
+        },
+        ...,
+    ]
+
+.. code-block:: python
+
+    from torchtune.models.flamingo import FlamingoTransform
+    from torchtune.datasets.multimodal import multimodal_chat_dataset
+
+    transform = FlamingoTransform(
+        path="/tmp/Meta-Llama-3-8B-Instruct/original/tokenizer.model",
+        prompt_template="torchtune.data.QuestionAnswerTemplate",
+        max_seq_len=8192,
+        tile_size=224,
+        patch_size=14,
+    )
+    ds = multimodal_chat_dataset(
+        model_transform=model_transform,
+        source="json",
+        data_files="data/my_data.json",
+        column_map={
+            "dialogue": "conversations",
+            "image_path": "image",
+        },
+        image_dir="/home/user/dataset/",  # /home/user/dataset/images/clock.jpg
+        image_tag="<image>",
+        split="train",
+    )
+    tokenized_dict = ds[0]
+    print(transform.decode(tokenized_dict["tokens"], skip_special_tokens=False))
+    # '<|begin_of_text|><|start_header_id|>user<|end_header_id|>\n\nQuestion:<|image|>What time is it on the clock?Answer:<|eot_id|><|start_header_id|>assistant<|end_header_id|>\n\nIt is 10:00AM.<|eot_id|>'
+    print(tokenized_dict["encoder_input"]["images"][0].shape)  # (num_tiles, num_channels, tile_height, tile_width)
+    # torch.Size([4, 3, 224, 224])
+
+.. code-block:: yaml
+
+    # In config
+    model_transform:
+      _component_: torchtune.models.flamingo
+      path: /tmp/Meta-Llama-3-8B-Instruct/original/tokenizer.model
+      prompt_template: torchtune.data.QuestionAnswerTemplate
+      max_seq_len: 8192
+
+    dataset:
+      _component_: torchtune.datasets.multimodal.multimodal_chat_dataset
+      source: json
+      data_files: data/my_data.json
+      split: train
+      column_map:
+        dialogue: conversations
+        image_path: image
+      image_dir: /home/user/dataset/
+      image_tag: "<image>"
+      split: train
+
+Multimodal dataset format
+-------------------------
+
+Multimodal datasets are currently expected to follow the :ref:`sharegpt` chat format, where the image paths are in one column
+and the user-assistant conversations are in another column.
+
+.. code-block:: text
+
+    |  conversations                     | image        |
+    |------------------------------------|--------------|
+    | [{"from": "human", "value": "Q1"}, | images/1.jpg |
+    |  {"from": "gpt", "value": "A1"}]   |              |
+
+As an example, you can see the schema of the `ShareGPT4V dataset <https://huggingface.co/datasets/Lin-Chen/ShareGPT4V>`_.
+
+
+Loading multimodal datasets from Hugging Face
+---------------------------------------------
+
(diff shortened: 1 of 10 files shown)


docs/source/basics/multimodal_datasets.rst
```diff
@@ -0,0 +1,285 @@
+.. _multimodal_dataset_usage_label:
+
+===================
+Multimodal Datasets
+===================
+
+Multimodal datasets include both text and image data used to train VLMs. torchtune currently supports multimodal chat datasets.
```
felipemello1, line 7:
```suggestion
Multimodal datasets include more than one data modality, e.g. text + image + speech, and can be used to train transformers. Torchtune currently only supports multimodal chat datasets for Vision-Language Models (VLMs).
```

docs/source/basics/multimodal_datasets.rst
```diff
@@ -0,0 +1,285 @@
...
+The primary entry point for fine-tuning with multimodal datasets in torchtune is the :func:`~torchtune.datasets.multimodal.multimodal_chat_dataset`
+builder. This lets you specify a local or Hugging Face dataset that follows the multimodal chat data format
+directly from the config and train your VLM on it.
+
+.. _example_multimodal:
+
+Example multimodal dataset
+--------------------------
+
+Here is an example of a multimodal chat dataset for a visual question-answering task.
```
felipemello1, line 18: I would add something like: "Notice how <image> is a place holder for the file provided in image_path, so we can know where to insert the image tokens".

I also think that the example should contain more than one image, so people know how to gerenalize it.

docs/source/basics/multimodal_datasets.rst
```diff
@@ -0,0 +1,285 @@
...
+    # torch.Size([4, 3, 224, 224])
+
+.. code-block:: yaml
+
+    # In config
+    model_transform:
+      _component_: torchtune.models.flamingo
+      path: /tmp/Meta-Llama-3-8B-Instruct/original/tokenizer.model
+      prompt_template: torchtune.data.QuestionAnswerTemplate
+      max_seq_len: 8192
```
felipemello1, line 77: oh, this is confusing, because in every other config, this is done in the tokenizer. Maybe we should explicitly say that model_transform takes the place of the tokenizer. A better solution might be to rename every config.tokenizer as model_transform, but i am not sure if i even like it.

felipemello1: dumb question: Why wouldnt "FlamingoTransform" just be "FlamingoTokenizer"? we are tokenizing now images and text

docs/source/basics/multimodal_datasets.rst
```diff
@@ -0,0 +1,285 @@
...
+
+    dataset:
+      _component_: torchtune.datasets.multimodal.multimodal_chat_dataset
+      source: json
+      data_files: data/my_data.json
```
felipemello1, line 82: can we add a dummy 2-example-file of what my_data.json should look like

docs/source/basics/multimodal_datasets.rst
```diff
@@ -0,0 +1,285 @@
...
+      split: train
+      column_map:
+        dialogue: conversations
+        image_path: image
+      image_dir: /home/user/dataset/
+      image_tag: "<image>"
```
felipemello1, line 88: noob question: will this ever break if the user adds <image> to their prompt, or <image> is a special token and the tokenizer understands if it comes from the user or the ChatML?

docs/source/basics/multimodal_datasets.rst
```diff
@@ -0,0 +1,285 @@
...
+and the user-assistant conversations are in another column.
+
+.. code-block:: text
+
+    |  conversations                     | image        |
+    |------------------------------------|--------------|
+    | [{"from": "human", "value": "Q1"}, | images/1.jpg |
+    |  {"from": "gpt", "value": "A1"}]   |              |
+
+As an example, you can see the schema of the `ShareGPT4V dataset <https://huggingface.co/datasets/Lin-Chen/ShareGPT4V>`_.
```
felipemello1, line 104: it seems that it forces single image per row, is that right? If so, can we make it explicit that this is a constraint of this format? Otherwise, can we add an example with more than 1 image?

docs/source/basics/multimodal_datasets.rst
```diff
@@ -0,0 +1,285 @@
...
+
+
+Loading multimodal datasets from Hugging Face
+---------------------------------------------
+
+You simply need to pass in the dataset repo name to ``source``, which is then passed into Hugging Face's ``load_dataset``.
+For most datasets, you will also need to specify the ``split`` and/or the subset via ``name``.
```
felipemello1, line 111: nice! super useful!

docs/source/basics/multimodal_datasets.rst
```diff
@@ -0,0 +1,285 @@
...
+                    {"type": "image", "content": pil_image},
+                    {"type": "text", "content": msg["value"]},
+                ]
+            )
+    print(user_message.contains_media)
+    # True
+    print(user_message.get_media())
+    # [<PIL.Image.Image>]
+    print(user_message.text_content)
+    # What time is it on the clock?
```
felipemello1, line 213: very nice!

torchtune/data/_messages.py
```diff
@@ -335,14 +336,22 @@ class ShareGPTToMessages(Transform):
         ]
 
     Args:
-        train_on_input (bool): whether the prompt should remain unmasked. Default: False
+        train_on_input (bool): whether the prompt should remain unmasked. For multimodal datasets, ``train_on_input``
```
felipemello1, line 339: can we add why this is always false?

torchtune/data/_messages.py
```diff
@@ -335,14 +336,22 @@ class ShareGPTToMessages(Transform):
...
+            is always False and this value is ignored. Default: False
         column_map (Optional[Dict[str, str]]): a mapping from the expected columns ("conversations")
             to the new column names in the dataset. Key should be "conversations" and value should
             be the new column name. If None, keep the default "conversations".
             Default is None.
         new_system_prompt (Optional[str]): if specified, prepend a system message. This can
             serve as instructions to guide the model response. Setting this will OVERRIDE any system
             messages already present in the dataset. Default is None.
+        image_dir (Optional[Path]): path to the directory containing the images that is prepended to all image
```
felipemello1, line 348: nit: fine to ignore, maybe add an example?

torchtune/data/_messages.py
```diff
@@ -335,14 +336,22 @@ class ShareGPTToMessages(Transform):
...
+            paths in the dataset. If None, assume images are available in current working directory or are located
+            on a remote url. For text-only,leave as None. Default is None.
+        image_tag (Optional[str]): placeholder tags in the text content of each message to be replaced by dictionaries
```
felipemello1, line 351: nit: maybe add an example? Based on this alone, i have no clue of what you are talking about

torchtune/data/_messages.py
```diff
@@ -384,12 +397,41 @@ def __call__(self, sample: Mapping[str, Any]) -> Mapping[str, Any]:
...
+                            {"type": "image", "content": pil_image},
+                            {"type": "text", "content": content},
+                        ]
+                    else:
+                        content = format_content_with_images(
+                            content,
+                            image_tag=self.image_tag,
+                            images=[pil_image],
+                        )
+                    image_loaded = True
```
felipemello1, line 430: what if there is more than one image per sample? wouldnt this break?

torchtune/data/_messages.py
```diff
@@ -384,12 +397,41 @@ def __call__(self, sample: Mapping[str, Any]) -> Mapping[str, Any]:
...
+        image_loaded = False
         for message in sample[self._column_map["conversations"]]:
             role = role_map[message["from"]]
+            content = message["value"]
             if role == "system" and self.new_system_prompt is not None:
                 continue
-            content = message["value"]
-            masked = (role != "assistant") and (not self.train_on_input)
+            if role == "user":
+                if is_multimodal and not image_loaded:
```
felipemello1, line 413: I dont understand this. After we load one image, then if there is some other message with role=="user", we will just ignore it? Shouldnt we raise an error instead of silently ignoring it?

torchtune/data/_messages.py
```diff
@@ -384,12 +397,41 @@ def __call__(self, sample: Mapping[str, Any]) -> Mapping[str, Any]:
...
+                        ]
+                    else:
+                        content = format_content_with_images(
+                            content,
+                            image_tag=self.image_tag,
+                            images=[pil_image],
+                        )
+                    image_loaded = True
+
+            masked = (role != "assistant") and (
```
felipemello1, line 432: dont fully follow it, could you please add a comment?

torchtune/datasets/multimodal/_llava_instruct.py
```diff
@@ -5,125 +5,19 @@
...
-                        images=[pil_image],
-                    )
-                    image_loaded = True
-            masked = role != "assistant"
-            messages.append(Message(role=role, content=content, masked=masked))
-
-        return {"messages": messages}
-
-
 # TODO: point to Flamingo model transform as an example
```
felipemello1, line 15: do we need to address this TODO?

torchtune/datasets/multimodal/_multimodal.py
```diff
@@ -0,0 +1,172 @@
...
+    *,
+    source: str,
+    column_map: Optional[Dict[str, str]] = None,
+    new_system_prompt: Optional[str] = None,
+    image_tag: Optional[str] = None,
+    image_dir: Optional[str] = None,
+    **load_dataset_kwargs: Dict[str, Any],
+) -> SFTDataset:
+    """
+    Configure a text+image dataset with conversations between user and model assistant.
```
felipemello1, line 26: nit: It bothers me a bit that its called multimodal, but it actually means text+image. Not asking for changes, just sharing

torchtune/datasets/multimodal/_multimodal.py
```diff
@@ -0,0 +1,172 @@
...
+    as an alternative to :class:`~torchtune.datasets.SFTDataset`, as it is made to be config friendly.
+
+    The dataset is expected to follow the ShareGPT format:
+
+    .. code-block:: text
+
+        |  conversations                     | image        |
+        |------------------------------------|--------------|
+        | [{"from": "human", "value": "Q1"}, | images/1.jpg |
+        |  {"from": "gpt", "value": "A1"}]   |              |
```
felipemello1, line 38: same as before, if we only support one image, can we make it explicit? otherwise, can we have an example with 2 images?

felipemello1 (approved):
very nice! thanks for putting this up. I didnt have time to finish reviewin "multimodal_chat_dataset". Approving to unblock.

torchtune/datasets/multimodal/_multimodal.py
joecummings: I concur.

docs/source/basics/multimodal_datasets.rst
RdoubleA: that's shown above?

docs/source/basics/multimodal_datasets.rst
RdoubleA: yes it will leade to unexpected behaviors, the burden is on the user for setting this correctly.. I don't know if there's an easy way to check for those edge cases.

docs/source/basics/multimodal_datasets.rst
RdoubleA: yes, will state that it's constrained to single image per row for now

torchtune/data/_messages.py
RdoubleA: I'm not sure really, I think it's because it's not functionally useful? @pbontrager

torchtune/data/_messages.py
RdoubleA: this will only load and append image tokens for the first user message. all other messages, including other user messages, will be added as normal.

torchtune/datasets/multimodal/_llava_instruct.py
RdoubleA: after certain events have transpired

torchtune/data/_messages.py
felipemello1: Do you mind making this comment in the code?

PR #1669, opened 2024-09-25
Add changes to support MM eval

joecummings:
#### Context
- [x] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

#### Changelog
*

#### Test plan

#### UX

### DIFF
--- a/.github/workflows/recipe_test.yaml
+++ b/.github/workflows/recipe_test.yaml
@@ -41,7 +41,7 @@ jobs:
         run: |
           python -m pip install torch torchvision torchao
           python -m pip install -e ".[dev]"
-          python -m pip install lm-eval==0.4.*
+          python -m pip install git+https://github.com/EleutherAI/lm-evaluation-harness.git@fb963f0f0a5b28b69763590bb59676072cf43a01
       - name: Run recipe tests with coverage
         run: pytest tests -m integration_test --cov=. --cov-report=xml --durations=20 -vv
       - name: Upload Coverage to Codecov
--- a/recipes/eleuther_eval.py
+++ b/recipes/eleuther_eval.py
@@ -40,7 +40,13 @@ if not lm_eval_version >= "0.4.2":
     )
 
 from lm_eval.evaluator import evaluate, get_task_list
-from lm_eval.models.hf_vlms import HFMultimodalLM
+
+# User doesn't have to have nightlies installed, they just won't be able
+# to use the multimodal model
+try:
+    from lm_eval.models.hf_vlms import HFMultimodalLM
+except ImportError as e:
+    pass
 from lm_eval.models.huggingface import HFLM
 from lm_eval.tasks import get_task_dict, TaskManager
 from lm_eval.utils import make_table
@@ -190,7 +196,7 @@ class _VLMEvalWrapper(HFMultimodalLM):
         tok_batch = padded_collate_tiled_images_and_mask(
             all_encoded_messages,
             pad_direction="left",
-            pad_max_images=7,
+            pad_max_images=self._max_images_per_sample,
         )
         utils.batch_to_device(tok_batch, self.device)
 
@@ -230,7 +236,8 @@ class _VLMEvalWrapper(HFMultimodalLM):
                 self.model.setup_caches(
                     batch_size=1,
                     dtype=self._dtype,
-                    encoder_max_seq_len=6404 * self._max_images_per_sample,
+                    encoder_max_seq_len=self.model_transform.image_seq_len
+                    * self._max_images_per_sample,
                     decoder_max_seq_len=self.max_length,
                 )
             causal_mask = torch.tril(
--- a/recipes/eleuther_eval.py
+++ b/recipes/eleuther_eval.py
@@ -34,7 +34,7 @@ try:
 except ImportError:
     print(
         "You must install the EleutherAI Eval Harness to run this recipe. "
-        "Please install with `pip install lm_eval>=0.4.4`"
+        "Please install with `pip install lm_eval>=0.4.2`"
     )
     sys.exit(1)
 
@@ -42,7 +42,7 @@ lm_eval_version = importlib.metadata.version("lm_eval")
 if not lm_eval_version >= "0.4.2":
     print(
         "You must install the EleutherAI Eval Harness >= v0.4.2 to run this recipe. "
-        "Please install with `pip install lm_eval>=0.4.4`"
+        "Please install with `pip install lm_eval>=0.4.2`"
     )
     sys.exit(1)
 
@@ -76,7 +76,7 @@ class _VLMEvalWrapper(HFMultimodalLM):
         transform (Transform): The transform (tokenizer) to use for preprocessing.
         device (torch.device): The device to use.
         max_seq_length (int): The maximum sequence length.
-        batch_size (int): The batch size per GPU.
+        batch_size (int): The batch size.
         dtype (torch.dtype): dtype for the model caches during generation.
         enable_kv_cache (bool): Whether to enable KV cache for generation.
         image_tag (str): The string to use for the image token. Default is "<image>", which
@@ -163,10 +163,8 @@ class _VLMEvalWrapper(HFMultimodalLM):
         return True
 
     def tok_encode(self, string, **kwargs) -> List[int]:
-        # Note on add_bos flag: setting to False as this gives better results, for example
-        # +1% on truthfulqa_mc2 with a LoRA finetune. lit-gpt also sets this to False,
-        # see https://github.com/Lightning-AI/lit-gpt/blob/main/eval/lm_eval_harness.py#
-        # L66, though notably fast-gpt does the opposite
+        # This is only used to get a number of tokens for use in sorting samples in dataset
+        # These values will not actually be used for eval
         return self._transform.tokenizer.encode(string, add_bos=False, add_eos=False)
 
     def tok_decode(self, tokens, skip_special_tokens=True) -> str:
--- a/recipes/eleuther_eval.py
+++ b/recipes/eleuther_eval.py
@@ -506,6 +506,11 @@ class EleutherEvalRecipe(EvalRecipeInterface):
 
         # Initialize tokenizer/transform
         model_transform = config.instantiate(cfg.tokenizer)
+        max_seq_len = (
+            model_transform.max_seq_len
+            if model_transform.max_seq_len is not None
+            else 4096  # default max_seq_len to 4096
+        )
 
         # Finally, we setup the actual EvalWrapper class
         if isinstance(model, DeepFusionModel):
@@ -521,7 +526,7 @@ class EleutherEvalRecipe(EvalRecipeInterface):
             model,
             model_transform,
             device=self.device,
-            max_seq_length=model_transform.max_seq_len,
+            max_seq_length=max_seq_len,
             batch_size=self.batch_size,
             dtype=self.dtype,
             enable_kv_cache=self.enable_kv_cache,
--- a/recipes/eleuther_eval.py
+++ b/recipes/eleuther_eval.py
@@ -79,7 +79,7 @@ class _VLMEvalWrapper(HFMultimodalLM):
         batch_size (int): The batch size per GPU.
         dtype (torch.dtype): dtype for the model caches during generation.
         enable_kv_cache (bool): Whether to enable KV cache for generation.
-        image_str (str): The string to use for the image token. Default is "<image>", which
+        image_tag (str): The string to use for the image token. Default is "<image>", which
             is the default used by the MMMU dataset.
         max_images_per_sample (int): The maximum number of images per sample. Defaults to
             the max number of images in MMMU.
@@ -95,7 +95,7 @@ class _VLMEvalWrapper(HFMultimodalLM):
         batch_size: int = 8,
         dtype: torch.dtype = torch.bfloat16,
         enable_kv_cache: bool = True,
-        image_str: str = "<image>",
+        image_tag: str = "<image>",
         max_images_per_sample: int = 7,
     ):
         self._model = model
@@ -105,11 +105,13 @@ class _VLMEvalWrapper(HFMultimodalLM):
         self._batch_size = batch_size
         self._dtype = dtype
         self._enable_kv_cache = enable_kv_cache
-        self._image_str = image_str
+        self._image_tag = image_tag
         self._max_images_per_sample = max_images_per_sample
 
     @property
     def model(self):
+        # Not actually changing the dtype here, just adding it as a
+        # property on the model
         self._model.dtype = self._dtype
         return self._model
 
@@ -123,6 +125,7 @@ class _VLMEvalWrapper(HFMultimodalLM):
 
     @property
     def cache_hook(self):
+        # Dummy class to appease the Harness
         class DummyCacheHook:
             def __init__(self):
                 self.add_partial = lambda x, y, z: True
@@ -195,7 +198,7 @@ class _VLMEvalWrapper(HFMultimodalLM):
             # Construct the messages
             messages = []
             content = format_content_with_images(
-                text, image_tag=self._image_str, images=proper_images
+                text, image_tag=self._image_tag, images=proper_images
             )
             messages.append(Message(role="user", content=content))
             messages.append(Message(role="assistant", content=""))
@@ -236,8 +239,8 @@ class _VLMEvalWrapper(HFMultimodalLM):
 
         if bsz > 1:
             raise ValueError(
-                f"Got a batch size of '{bsz}'. Batch size > 1 is not supported for "
-                "generation. See https://github.com/pytorch/torchtune/issues/1250 for more info."
+                f"Got a batch size of '{bsz}'. Batch size > 1 is not yet supported for "
+                "multimodal generation."
             )
 
         # 1. Setup caches for a given batch size
--- a/recipes/eleuther_eval.py
+++ b/recipes/eleuther_eval.py
@@ -4,39 +4,278 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
+import importlib.metadata
 import sys
 import time
 
-from typing import Any, Dict, List, Tuple, Union
+from typing import Dict, List, Tuple, Union
+
+import PIL
 
 import torch
 from omegaconf import DictConfig
 
-from torch import nn
-from torchtune import config, generation, training, utils
-from torchtune.data import left_pad_sequence
+from torchtune import config, training, utils
+from torchtune.data import (
+    format_content_with_images,
+    left_pad_sequence,
+    Message,
+    padded_collate_tiled_images_and_mask,
+)
+from torchtune.generation import generate, sample
 from torchtune.modules import TransformerDecoder
+from torchtune.modules.model_fusion import DeepFusionModel
 from torchtune.modules.tokenizers import ModelTokenizer
+from torchtune.modules.transforms import Transform
 from torchtune.recipe_interfaces import EvalRecipeInterface
 
-logger = utils.get_logger("DEBUG")
-
 try:
-    import lm_eval
-    from lm_eval.evaluator import evaluate, get_task_list
-    from lm_eval.models.huggingface import HFLM
-    from lm_eval.tasks import get_task_dict, TaskManager
-    from lm_eval.utils import make_table
-except ImportError:
-    logger.error(
-        "Recipe requires EleutherAI Eval Harness v0.4. Please install with `pip install lm_eval==0.4.*`"
+    lm_eval_version = importlib.metadata.version("lm_eval")
+except importlib.metadata.PackageNotFoundError:
+    lm_eval_version = ""
+
+if not lm_eval_version >= "0.4.2":
+    raise ImportError(
+        "lm_eval version must be >= 0.4.2. Please install lm_eval >= 0.4.2."
     )
-    sys.exit(1)
+
+from lm_eval.evaluator import evaluate, get_task_list
+from lm_eval.models.hf_vlms import HFMultimodalLM
+from lm_eval.models.huggingface import HFLM
+from lm_eval.tasks import get_task_dict, TaskManager
+from lm_eval.utils import make_table
 
 
-class _EvalWrapper(HFLM):
+class _VLMEvalWrapper(HFMultimodalLM):
     """An EvalWrapper for EleutherAI's eval harness based on gpt-fast's
     EvalWrapper: https://github.com/pytorch-labs/gpt-fast/blob/main/eval.py.
+
+    Note:
+        This is ONLY for vision-language models.
+
+    Args:
+        model (DeepFusionModel): The VLM to evaluate.
+        transform (Transform): The transform (tokenizer) to use for preprocessing.
+        device (torch.device): The device to use.
+        max_seq_length (int): The maximum sequence length.
+        batch_size (int): The batch size per GPU.
+        dtype (torch.dtype): dtype for the model caches during generation.
+        enable_kv_cache (bool): Whether to enable KV cache for generation.
+        image_str (str): The string to use for the image token. Default is "<image>", which
+            is the default used by the MMMU dataset.
+        max_images_per_sample (int): The maximum number of images per sample. Defaults to
+            the max number of images in MMMU.
+    """
+
+    def __init__(
+        self,
+        model: DeepFusionModel,
+        transform: Transform,
+        *,
+        device: torch.device,
+        max_seq_length: int = 4096,
+        batch_size: int = 8,
+        dtype: torch.dtype = torch.bfloat16,
+        enable_kv_cache: bool = True,
+        image_str: str = "<image>",
+        max_images_per_sample: int = 7,
+    ):
+        self._model = model
+        self._transform = transform
+        self._device = device
+        self._max_seq_length = max_seq_length
+        self._batch_size = batch_size
+        self._dtype = dtype
+        self._enable_kv_cache = enable_kv_cache
+        self._image_str = image_str
+        self._max_images_per_sample = max_images_per_sample
+
+    @property
+    def model(self):
+        self._model.dtype = self._dtype
+        return self._model
+
+    @property
+    def model_transform(self):
+        return self._transform
+
+    @property
+    def device(self):
+        return self._device
+
+    @property
+    def cache_hook(self):
+        class DummyCacheHook:
+            def __init__(self):
+                self.add_partial = lambda x, y, z: True
+
+        return DummyCacheHook()
+
+    @property
+    def rank(self):
+        # Hardcoded for now b/c we only support single GPU eval
+        return 0
+
+    @property
+    def world_size(self):
+        # Hardcoded for now b/c we only support single GPU eval
+        return 1
+
+    @property
+    def batch_size(self):
+        return self._batch_size
+
+    @property
+    def eos_token_id(self):
+        return self._transform.tokenizer.eos_id
+
+    @property
+    def eot_token_id(self):
+        return self._transform.tokenizer.eot_id
+
+    @property
+    def max_length(self):
+        return self._max_seq_length
+
+    @property
+    def truncation(self):
+        return True
+
+    def tok_encode(self, string, **kwargs) -> List[int]:
+        # Note on add_bos flag: setting to False as this gives better results, for example
+        # +1% on truthfulqa_mc2 with a LoRA finetune. lit-gpt also sets this to False,
+        # see https://github.com/Lightning-AI/lit-gpt/blob/main/eval/lm_eval_harness.py#
+        # L66, though notably fast-gpt does the opposite
+        return self._transform.tokenizer.encode(string, add_bos=False, add_eos=False)
+
+    def tok_decode(self, tokens, skip_special_tokens=True) -> str:
+        if isinstance(tokens, int):
+            tokens = [tokens]
+        return self._transform.tokenizer.decode(
+            tokens, skip_special_tokens=skip_special_tokens
+        )
+
+    def tok_batch_multimodal_encode(
+        self,
+        all_texts: List[str],
+        all_images: List[List[PIL.Image.Image]],
+        *args,
+        **kwargs,
+    ):
+        # Eleuther already parses out the text and images, so we just need to get
+        # it into a Message format for our tokenizer
+        all_encoded_messages = []
+
+        for text, images in zip(all_texts, all_images):
+            # Ensure images are all RGB
+            proper_images = []
+            for image in images:
+                if image.mode != "RGB":
+                    image = image.convert("RGB")
+                proper_images.append(image)
+
+            # Construct the messages
+            messages = []
+            content = format_content_with_images(
+                text, image_tag=self._image_str, images=proper_images
+            )
+            messages.append(Message(role="user", content=content))
+            messages.append(Message(role="assistant", content=""))
+
+            # Transform the messages
+            tok_batch = self.model_transform({"messages": messages}, inference=True)
+            all_encoded_messages.append(tok_batch)
+
+        # Pad the encoded messages
+        tok_batch = padded_collate_tiled_images_and_mask(
+            all_encoded_messages,
+            pad_direction="left",
+            pad_max_images=7,
+        )
+        utils.batch_to_device(tok_batch, self.device)
+
+        # Convert the batch to the format expected by the HF
+        tok_batch["input_ids"] = tok_batch.pop("tokens")
+        return tok_batch
+
+    @torch.inference_mode()
+    def _model_multimodal_generate(
+        self,
+        batch: Dict[str, torch.Tensor],
+        max_length: int,
+        stop: List[str],
+        **generation_kwargs,
+    ):
+        prompt = batch.pop("input_ids")
+        bsz, seq_len = prompt.shape
+
+        temperature = generation_kwargs.get("temperature", 0.0)
+        do_sample = generation_kwargs.get("do_sample", False)
+        if do_sample or temperature != 0.0:
+            raise RuntimeError(
+                "Any decoding strategy other than greedy is not supported."
+            )
+
+        if bsz > 1:
+            raise ValueError(
+                f"Got a batch size of '{bsz}'. Batch size > 1 is not supported for "
+                "generation. See https://github.com/pytorch/torchtune/issues/1250 for more info."
+            )
+
+        # 1. Setup caches for a given batch size
+        with self.device:
+            if self.model.caches_are_enabled():
+                self.model.reset_caches()
+            else:
+                self.model.setup_caches(
+                    batch_size=1,
+                    dtype=self._dtype,
+                    encoder_max_seq_len=6404 * self._max_images_per_sample,
+                    decoder_max_seq_len=self.max_length,
+                )
+            causal_mask = torch.tril(
+                torch.ones(
+                    size=(self.max_length, self.max_length),
+                    dtype=torch.bool,
+                )
+            )
+            input_pos = torch.arange(self.max_length)
+
+        batch["input_pos"] = input_pos[None, :seq_len]
+        batch["mask"] = causal_mask[None, :seq_len]
+
+        # 4. Prefill step
+        generated_tokens = []
+        logits = self.model(prompt, **batch)[:, -1]
+        token = sample(logits, temperature=0.0, top_k=None)
+        generated_tokens.append(token.item())
+
+        cache_mask = batch["encoder_mask"][:, -1:]
+
+        # 5. Continue generating
+        for _ in range(max_length):
+            if token.item() in self.model_transform.stop_tokens:
+                break
+            logits = self.model(
+                token,
+                mask=causal_mask[None, seq_len, None, :],
+                encoder_input=None,
+                encoder_mask=cache_mask,
+                input_pos=input_pos[None, seq_len],
+            )[:, -1]
+            token = sample(logits, temperature=0.0, top_k=None)
+            generated_tokens.append(token.item())
+            seq_len += 1
+
+        return torch.tensor(generated_tokens, dtype=torch.int32).unsqueeze(0)
+
+
+class _LLMEvalWrapper(HFLM):
+    """An EvalWrapper for EleutherAI's eval harness based on gpt-fast's
+    EvalWrapper: https://github.com/pytorch-labs/gpt-fast/blob/main/eval.py.
+
+    Note:
+        This is for text-only decoder models.
 
     Args:
         model (TransformerDecoder): The model to evaluate.
@@ -46,6 +285,7 @@ class _EvalWrapper(HFLM):
         max_seq_length (int): The maximum sequence length to use.
         batch_size (int): The batch size per GPU to use.
         dtype (torch.dtype): dtype for the model caches during generation.
+        enable_kv_cache (bool): Whether to enable KV cache for generation.
     """
 
     def __init__(
@@ -57,13 +297,16 @@ class _EvalWrapper(HFLM):
         max_seq_length: int = 4096,
         batch_size: int = 8,
         dtype: torch.dtype = torch.float32,
+        enable_kv_cache: bool = True,
     ):
+        # TODO (@joecummings): Remove this init function so we don't load in extraneous stuff
         super().__init__(pretrained="gpt2", device=str(device))
         self._model = model
         self._tokenizer = tokenizer
         self._max_seq_length = max_seq_length
         self._batch_size = batch_size
         self._dtype = dtype
+        self._enable_kv_cache = enable_kv_cache
 
     @property
     def model(self):
@@ -88,6 +331,10 @@ class _EvalWrapper(HFLM):
     @property
     def device(self):
         return self._device
+
+    @property
+    def enable_kv_cache(self):
+        return self._enable_kv_cache
 
     def tok_encode(self, text: str, **kwargs) -> List[int]:
         # Note on add_bos flag: setting to False as this gives better results, for example
@@ -119,9 +366,22 @@ class _EvalWrapper(HFLM):
     def _model_call(self, inps: torch.Tensor, **kwargs) -> torch.Tensor:
         return self._model(inps)
 
+    @torch.inference_mode()
     def _model_generate(
         self, context: torch.Tensor, **generation_kwargs
     ) -> torch.Tensor:
+
+        # Setup KV caches OR reset them if they're already set up
+        if self.model.caches_are_enabled():
+            self.model.reset_caches()
+        else:
+            with self.device:
+                self.model.setup_caches(
+                    batch_size=self.batch_size,
+                    dtype=self._dtype,
+                    decoder_max_seq_len=self._max_seq_length,
+                )
+
         curr_batch_size = context.size(0)
 
         # if we've recieved fewer than self._batch_size samples in the current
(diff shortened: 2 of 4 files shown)


recipes/eleuther_eval.py
```diff
@@ -4,40 +4,279 @@
...
 try:
-    import lm_eval
-    from lm_eval.evaluator import evaluate, get_task_list
-    from lm_eval.models.huggingface import HFLM
-    from lm_eval.tasks import get_task_dict, TaskManager
-    from lm_eval.utils import make_table
-except ImportError:
-    logger.error(
-        "Recipe requires EleutherAI Eval Harness v0.4. Please install with `pip install lm_eval==0.4.*`"
+    lm_eval_version = importlib.metadata.version("lm_eval")
```
joecummings, line 33: Instead of checking via importing legacy functions, we actually check the version. 

This is the correct way lol

recipes/eleuther_eval.py
```diff
@@ -4,40 +4,279 @@
...
+        self,
+        model: DeepFusionModel,
+        transform: Transform,
+        *,
+        device: torch.device,
+        max_seq_length: int = 4096,
+        batch_size: int = 8,
+        dtype: torch.dtype = torch.bfloat16,
+        enable_kv_cache: bool = True,
+        image_str: str = "<image>",
```
joecummings, line 80: Not going to make image_st and max_images_per_sample configurable from the config ATM b/c we only support MMMU and these are the features necessary for MMMU.

recipes/eleuther_eval.py
```diff
@@ -4,40 +4,279 @@
...
+    @property
+    def model_transform(self):
+        return self._transform
+
+    @property
+    def device(self):
+        return self._device
+
+    @property
+    def cache_hook(self):
```
joecummings, line 107: Have to do this to make the harness happy.

recipes/eleuther_eval.py
```diff
@@ -4,40 +4,279 @@
...
+        **kwargs,
+    ):
+        # Eleuther already parses out the text and images, so we just need to get
+        # it into a Message format for our tokenizer
+        all_encoded_messages = []
+
+        for text, images in zip(all_texts, all_images):
+            # Ensure images are all RGB
+            proper_images = []
+            for image in images:
```
joecummings, line 172: Not all the images are in RGB format so we need to convert them

recipes/eleuther_eval.py
```diff
@@ -4,40 +4,279 @@
...
+                    image = image.convert("RGB")
+                proper_images.append(image)
+
+            # Construct the messages
+            messages = []
+            content = format_content_with_images(
+                text, image_tag=self._image_str, images=proper_images
+            )
+            messages.append(Message(role="user", content=content))
+            messages.append(Message(role="assistant", content=""))
```
joecummings, line 183: Append assistant message to kick start generation.

recipes/eleuther_eval.py
```diff
@@ -4,40 +4,279 @@
...
+
+        # 1. Setup caches for a given batch size
+        with self.device:
+            if self.model.caches_are_enabled():
+                self.model.reset_caches()
+            else:
+                self.model.setup_caches(
+                    batch_size=1,
+                    dtype=self._dtype,
+                    encoder_max_seq_len=6404 * self._max_images_per_sample,
```
joecummings, line 233: Will need to grab the encoder_image_length instead of defaulting to 6404.

recipes/eleuther_eval.py
```diff
@@ -163,101 +420,84 @@ class EleutherEvalRecipe(EvalRecipeInterface):
...
-
-        model_eval_wrapper = _EvalWrapper(
-            self._model,
-            self._tokenizer,
-            device=self._device,
-            max_seq_length=self._max_seq_length,
-            batch_size=self._batch_size,
-            dtype=self._dtype,
+        # Finally, we setup the actual EvalWrapper class
+        eleuther_model_wrapper = (
```
joecummings, line 484: per our discussion earlier, this is simplified @felipemello1

recipes/eleuther_eval.py
```diff
@@ -266,35 +506,31 @@ def evaluate(self) -> None:
...
-            limit=self._limit,
+            limit=self.limit,
         )
+        t1 = time.time() - t0
 
-        logger.info(f"Eval completed in {time.time() - t1:.02f} seconds.")
-
+        # Log metrics
+        self.logger.info(f"Eval completed in {t1:.02f} seconds.")
+        self.logger.info(
```
joecummings, line 521: I added some memory logging.

tests/recipes/test_eleuther_eval.py
```diff
@@ -22,21 +22,21 @@ class TestEleutherEval:
     @pytest.mark.parametrize(
         "eval_name, expected_acc, bsz",
         [
-            ("truthfulqa_gen", 0.1, 8),
+            ("truthfulqa_gen", 0.1, 4),
```
joecummings, line 25: No need to do bsz 8, that's just slow as hell.

tests/recipes/test_eleuther_eval.py
```diff
@@ -122,3 +122,40 @@ def test_eval_recipe_errors_without_lm_eval(self, caplog, monkeypatch, tmpdir):
 
         err_log = caplog.messages[-1]
         assert "Recipe requires EleutherAI Eval Harness v0.4" in err_log
+
+    @pytest.mark.integration_test
+    def test_eval_recipe_errors_with_generate_until_and_mc_tasks(
```
joecummings, line 127: more tests good, yes?

recipes/eleuther_eval.py
```diff
@@ -4,40 +4,279 @@
...
-except ImportError:
-    logger.error(
-        "Recipe requires EleutherAI Eval Harness v0.4. Please install with `pip install lm_eval==0.4.*`"
+    lm_eval_version = importlib.metadata.version("lm_eval")
+except importlib.metadata.PackageNotFoundError:
+    lm_eval_version = ""
+
+if not lm_eval_version >= "0.4.2":
+    raise ImportError(
+        "lm_eval version must be >= 0.4.2. Please install lm_eval >= 0.4.2."
```
felipemello1, line 39: fyi, i had to create this import_guard.py for another PR, because i was using it in multiple files. I wonder if it would make sense to move this there. My first intuition is that we should NOT, as keeping this logic closer to the code is better, unless you are using it somewhere else too: https://github.com/pytorch/torchtune/blob/main/torchtune/utils/_import_guard.py

joecummings: This is good to keep in mind - my guess is that we will end up utilizing version and package guards much more frequently as our recipes expand.

recipes/eleuther_eval.py
```diff
@@ -4,40 +4,279 @@
...
+    Note:
+        This is ONLY for vision-language models.
+
+    Args:
+        model (DeepFusionModel): The VLM to evaluate.
+        transform (Transform): The transform (tokenizer) to use for preprocessing.
+        device (torch.device): The device to use.
+        max_seq_length (int): The maximum sequence length.
+        batch_size (int): The batch size per GPU.
+        dtype (torch.dtype): dtype for the model caches during generation.
```
felipemello1, line 62: if this is only for the model cache, and not the model itself, can we make it explicit?

recipes/eleuther_eval.py
```diff
@@ -4,40 +4,279 @@
...
+        enable_kv_cache (bool): Whether to enable KV cache for generation.
+        image_str (str): The string to use for the image token. Default is "<image>", which
```
felipemello1, line 64: i believe Rafi was using the term "tag" in his PR. We should be consistent on the vocab/var name

recipes/eleuther_eval.py
```diff
@@ -4,40 +4,279 @@
...
+        model: DeepFusionModel,
+        transform: Transform,
+        *,
+        device: torch.device,
+        max_seq_length: int = 4096,
+        batch_size: int = 8,
+        dtype: torch.dtype = torch.bfloat16,
+        enable_kv_cache: bool = True,
+        image_str: str = "<image>",
+        max_images_per_sample: int = 7,
```
felipemello1, line 81: 7 feels odd. Can we leave it empty and infer it from the dataset?

recipes/eleuther_eval.py
```diff
@@ -4,40 +4,279 @@
...
+        self._batch_size = batch_size
+        self._dtype = dtype
+        self._enable_kv_cache = enable_kv_cache
+        self._image_str = image_str
+        self._max_images_per_sample = max_images_per_sample
+
+    @property
+    def model(self):
+        self._model.dtype = self._dtype
+        return self._model
```
felipemello1, line 96: It feels wrong that we silently change the dtype of the model here. Explicit is better than implicit.

recipes/eleuther_eval.py
```diff
@@ -4,40 +4,279 @@
...
+        return self._device
+
+    @property
+    def cache_hook(self):
+        class DummyCacheHook:
+            def __init__(self):
+                self.add_partial = lambda x, y, z: True
+
+        return DummyCacheHook()
+
```
felipemello1, line 113: small docstring  please. Not sure whats going on

recipes/eleuther_eval.py
joecummings: So what's weird here is we're not actually changing the dtype. We already loaded the model in as the specified dtype. However, in the code somewhere, the harness check the `dtype` property of the model, which we don't explicitly set. So we're setting it here. No conversion is happening.

recipes/eleuther_eval.py
```diff
@@ -4,40 +4,279 @@
...
+    @property
+    def eot_token_id(self):
+        return self._transform.tokenizer.eot_id
+
+    @property
+    def max_length(self):
+        return self._max_seq_length
+
+    @property
+    def truncation(self):
```
felipemello1, line 147: probably "is_truncated" would be better

recipes/eleuther_eval.py
```diff
@@ -4,40 +4,279 @@
...
+    @property
+    def eos_token_id(self):
+        return self._transform.tokenizer.eos_id
+
+    @property
+    def eot_token_id(self):
+        return self._transform.tokenizer.eot_id
+
+    @property
+    def max_length(self):
```
felipemello1, line 143: probably "max_seq_length" would be better

recipes/eleuther_eval.py
```diff
@@ -4,40 +4,286 @@
...
+        return self._max_seq_length
+
+    @property
+    def truncation(self):
+        return True
+
+    def tok_encode(self, string, **kwargs) -> List[int]:
```
felipemello1, line 150: should probably be just "encode", to keep the same api for everything

recipes/eleuther_eval.py
```diff
@@ -4,40 +4,286 @@
...
+        # Note on add_bos flag: setting to False as this gives better results, for example
+        # +1% on truthfulqa_mc2 with a LoRA finetune. lit-gpt also sets this to False,
+        # see https://github.com/Lightning-AI/lit-gpt/blob/main/eval/lm_eval_harness.py#
+        # L66, though notably fast-gpt does the opposite
+        return self._transform.tokenizer.encode(string, add_bos=False, add_eos=False)
```
felipemello1, line 155: kwargs are not passed. Should we do args and kwargs?

recipes/eleuther_eval.py
```diff
@@ -4,40 +4,286 @@
...
+
+    def tok_decode(self, tokens, skip_special_tokens=True) -> str:
```
felipemello1, line 157: should probably be just "decode", to keep the same api for everything. Should we add args/kwargs?

recipes/eleuther_eval.py
```diff
@@ -4,40 +4,286 @@
...
+        if isinstance(tokens, int):
+            tokens = [tokens]
+        return self._transform.tokenizer.decode(
+            tokens, skip_special_tokens=skip_special_tokens
+        )
+
+    def tok_batch_multimodal_encode(
```
felipemello1, line 164: should probably be batch_encode (or whatever is the common way of saying it)

recipes/eleuther_eval.py
```diff
@@ -4,40 +4,286 @@
...
+        self,
+        all_texts: List[str],
+        all_images: List[List[PIL.Image.Image]],
```
felipemello1, line 167: nit: fine to keep "all_", but it feels weird. I think it should be the same as encode and reduce the number of bumps on the way because of different arg names

recipes/eleuther_eval.py
```diff
@@ -4,40 +4,286 @@
...
+        *args,
+        **kwargs,
+    ):
+        # Eleuther already parses out the text and images, so we just need to get
+        # it into a Message format for our tokenizer
+        all_encoded_messages = []
+
+        for text, images in zip(all_texts, all_images):
```
felipemello1, line 175: we should probably add a test here to check if they are lists. If not, maybe we should make them a list in the case bsz=1

recipes/eleuther_eval.py
```diff
@@ -4,40 +4,286 @@
...
+            # Ensure images are all RGB
+            proper_images = []
+            for image in images:
+                if image.mode != "RGB":
+                    image = image.convert("RGB")
```
felipemello1, line 180: our image transform takes care of it, I believe. Is it the job of the eval to do it?

recipes/eleuther_eval.py
felipemello1: is it fine for the context here to be empty, or should it be defined somewhere?

recipes/eleuther_eval.py
```diff
@@ -4,40 +4,286 @@
...
+            messages.append(Message(role="user", content=content))
+            messages.append(Message(role="assistant", content=""))
+
+            # Transform the messages
+            tok_batch = self.model_transform({"messages": messages}, inference=True)
+            all_encoded_messages.append(tok_batch)
+
+        # Pad the encoded messages
+        tok_batch = padded_collate_tiled_images_and_mask(
+            all_encoded_messages,
```
felipemello1, line 197: nit: feel free to ignore. Not a big fan of "all_" and "tok_", but i think it may help the reader to understand its not a single sample?

recipes/eleuther_eval.py
```diff
@@ -4,40 +4,286 @@
...
+            pad_max_images=self._max_images_per_sample,
+        )
+        utils.batch_to_device(tok_batch, self.device)
+
+        # Convert the batch to the format expected by the HF
+        tok_batch["input_ids"] = tok_batch.pop("tokens")
+        return tok_batch
+
+    @torch.inference_mode()
+    def _model_multimodal_generate(
```
felipemello1, line 208: this is a bit ugly haha. Do we have to include "multimodal"? could it just be _generate?

recipes/eleuther_eval.py
```diff
@@ -4,40 +4,286 @@
...
+        **generation_kwargs,
+    ):
+        prompt = batch.pop("input_ids")
+        bsz, seq_len = prompt.shape
+
+        temperature = generation_kwargs.get("temperature", 0.0)
+        do_sample = generation_kwargs.get("do_sample", False)
+        if do_sample or temperature != 0.0:
+            raise RuntimeError(
+                "Any decoding strategy other than greedy is not supported."
```
felipemello1, line 222: :O People almost never use temp=0, right? Is there an easy fix? If not, maybe make it a bit more friendly and instruct the user to set temperature=0 and do_sample=False, while we are working to enable it.

recipes/eleuther_eval.py
```diff
@@ -4,40 +4,286 @@
...
+            )
+
+        if bsz > 1:
+            raise ValueError(
+                f"Got a batch size of '{bsz}'. Batch size > 1 is not supported for "
+                "generation. See https://github.com/pytorch/torchtune/issues/1250 for more info."
```
felipemello1, line 234: this is also frustrating for the user. Maybe make a bit more friendly, indicating that we are working on it.

recipes/eleuther_eval.py
joecummings: Not in charge of this one.

recipes/eleuther_eval.py
joecummings: Not in charge of this one.

recipes/eleuther_eval.py
joecummings: Not in charge of this one.

recipes/eleuther_eval.py
joecummings: Not in charge of this one.

recipes/eleuther_eval.py
joecummings: Not in charge of this one :(

recipes/eleuther_eval.py
joecummings: This is pretty standard for evals.

recipes/eleuther_eval.py
```diff
@@ -4,40 +4,286 @@
...
+                    * self._max_images_per_sample,
+                    decoder_max_seq_len=self.max_length,
+                )
+            causal_mask = torch.tril(
+                torch.ones(
+                    size=(self.max_length, self.max_length),
+                    dtype=torch.bool,
+                )
+            )
+            input_pos = torch.arange(self.max_length)
```
felipemello1, line 255: noob question: is this right? dont images need some special masking?

recipes/eleuther_eval.py
```diff
@@ -4,40 +4,286 @@
...
+        generated_tokens = []
+        logits = self.model(prompt, **batch)[:, -1]
+        token = sample(logits, temperature=0.0, top_k=None)
+        generated_tokens.append(token.item())
+
+        cache_mask = batch["encoder_mask"][:, -1:]
+
+        # 5. Continue generating
+        for _ in range(max_length):
+            if token.item() in self.model_transform.stop_tokens:
```
felipemello1, line 270: prob not a big deal in this case, but i dont like .item(). I wonder if theres a way around it or if we need it

felipemello1: e.g. 

```
stop_tokens = torch.tensor([1,2,3])
token = torch.tensor([2])

torch.any(token == stop_tokens)
```

recipes/eleuther_eval.py
```diff
@@ -4,40 +4,286 @@
...
+                break
+            logits = self.model(
+                token,
+                mask=causal_mask[None, seq_len, None, :],
+                encoder_input=None,
+                encoder_mask=cache_mask,
+                input_pos=input_pos[None, seq_len],
+            )[:, -1]
```
felipemello1, line 278: not for this PR: this looks like we could be generating only the logit for the last token, and avoid extra computation/memory, when kv_cache==True

recipes/eleuther_eval.py
```diff
@@ -4,40 +4,286 @@
...
+            token = sample(logits, temperature=0.0, top_k=None)
+            generated_tokens.append(token.item())
```
felipemello1, line 280: no need to do .item() here if we will return a tensor later anyway. Can we just do stack or concat?

recipes/eleuther_eval.py
```diff
@@ -163,101 +427,84 @@ class EleutherEvalRecipe(EvalRecipeInterface):
...
-        if self._quantization_mode is None:
+        self.device = utils.get_device(device=cfg.device)
+        self.dtype = training.get_dtype(dtype=cfg.dtype, device=self.device)
+        self.logger = utils.get_logger(cfg.get("log_level", "info"))
+        training.set_seed(seed=cfg.seed)
+
+        # Eval specific variables
+        self.limit = cfg.limit
+        self.tasks = list(cfg.tasks)
+        self.max_seq_length = cfg.max_seq_length
```
felipemello1, line 455: we use "max_seq_len" everywhere else, i believe

recipes/eleuther_eval.py
felipemello1: maybe worth logging?

tests/recipes/test_eleuther_eval.py
felipemello1: honestly, if it doesnt impact the test, can we do 1 and reduce test time?

tests/recipes/test_eleuther_eval.py
```diff
@@ -22,21 +22,21 @@ class TestEleutherEval:
...
-        self, capsys, monkeypatch, tmpdir, eval_name, expected_acc, bsz
+        self, caplog, monkeypatch, tmpdir, eval_name, expected_acc, bsz
     ):
         ckpt = "llama2_tune"
         ckpt_path = Path(CKPT_MODEL_PATHS[ckpt])
         ckpt_dir = ckpt_path.parent
 
         # explicitly setting limit to an odd number here to ensure generation tasks
-        # work with KV-cacheing + bsz > 1 - we'll recieve batches of size 8, 8, 5
+        # work with KV-cacheing + bsz > 1 - we'll recieve batches of size 4, 4, 3
```
felipemello1, line 39:
```suggestion
        # work with KV-cacheing + bsz > 1 - we'll receive batches of size 4, 4, 3
```

recipes/eleuther_eval.py
felipemello1: I am a bit confused about this error, since this PR apparently is about enabling bsz>1

joecummings: What PR? This one is just for enabling Multimodal.

recipes/eleuther_eval.py
felipemello1: can we add this as a comment?

felipemello1:
please, if you make changes, and they are applicable to the other class, we should probably keep them in sync

recipes/eleuther_eval.py
joecummings: No, the content should be empty.

recipes/eleuther_eval.py
felipemello1: I guess i dont understand the tests enough, but I think i saw bsz=4 in the tests. Is that for text only?

recipes/eleuther_eval.py
joecummings: Not sure if possible to infer from the dataset ATM.  I'd prefer to handle extra datasets once MMMU is no longer the only supported dataset in Eleuther.

recipes/eleuther_eval.py
joecummings: Is `batch_texts` and `batch_images` better?

recipes/eleuther_eval.py
joecummings: Not an issue - they are definitely lists even if bsz  = 1

recipes/eleuther_eval.py
joecummings: When I remove this, our transform does not work. Worth investigating if our transform *should* take care of this but I can tell you right now it doesn't seem to be.

recipes/eleuther_eval.py
joecummings: cc @pbontrager ??

recipes/eleuther_eval.py
felipemello1: @pbontrager

recipes/eleuther_eval.py
joecummings: Not sure I understand - this is only grabbing the last token logits.

recipes/eleuther_eval.py
joecummings: Ahh yes we can.

recipes/eleuther_eval.py
RdoubleA: So to properly prompt the model for generation we need to
- Make sure we trail with an empty assistant message
- Set inference to true on tokenize messages

This is hard to remember both... not for this PR, but we should see if there's a more intuitive way to do this

.github/workflows/recipe_test.yaml
```diff
@@ -41,7 +41,7 @@ jobs:
         run: |
           python -m pip install torch torchvision torchao
           python -m pip install -e ".[dev]"
-          python -m pip install lm-eval==0.4.*
+          python -m pip install git+https://github.com/EleutherAI/lm-evaluation-harness.git@fb963f0f0a5b28b69763590bb59676072cf43a01
```
ebsmothers, line 44: Do we plan to add something on this to our readme or something?

recipes/eleuther_eval.py
```diff
@@ -4,40 +4,298 @@
...
+        "You must install the EleutherAI Eval Harness to run this recipe. "
+        "Please install with `pip install lm_eval>=0.4.4`"
+    )
+    sys.exit(1)
+
+lm_eval_version = importlib.metadata.version("lm_eval")
+if not lm_eval_version >= "0.4.2":
+    print(
+        "You must install the EleutherAI Eval Harness >= v0.4.2 to run this recipe. "
+        "Please install with `pip install lm_eval>=0.4.4`"
```
ebsmothers, line 45: Why not raise? Also version numbers don't match

.github/workflows/recipe_test.yaml
joecummings: README would be ideal

recipes/eleuther_eval.py
joecummings: yeah that's just for text

.github/workflows/recipe_test.yaml
ebsmothers: U gonna add?

recipes/eleuther_eval.py
joecummings: Hard to catch raises in the tests. 

I want to investigate this further, but it's a follow-up PR.

tests/recipes/test_eleuther_eval.py
ebsmothers: I though @SalmanMohammadi added a test very similar to this already? Also test isn't added?

recipes/eleuther_eval.py
felipemello1: right, you dont need all of the logits, yet, you calculate all of them. Inside of the transformer, you can do something like:

output = output_projection(h[..., -1])

so we project only the last token, which is the only one you need.

recipes/eleuther_eval.py
felipemello1: your call, just a nit

recipes/eleuther_eval.py
ebsmothers: OK fair enough. The version numbers though? What's with 0.4.2/0.4.4 mismatch

recipes/eleuther_eval.py
felipemello1: can we add a TODO then? I feel that this may cause silent performance degradation when we have a dataset with max>7, and it may be hard to debug

recipes/eleuther_eval.py
joecummings: Yeah definitely agree. Should have to be that hard to know what's going to give proper generation.

recipes/eleuther_eval.py
```diff
@@ -4,40 +4,301 @@
...
 
+    Note:
+        This is ONLY for vision-language models.
+
+    Args:
+        model (DeepFusionModel): The VLM to evaluate.
+        transform (Transform): The transform (tokenizer) to use for preprocessing.
+        device (torch.device): The device to use.
+        max_seq_length (int): The maximum sequence length.
+        batch_size (int): The batch size per GPU.
```
ebsmothers, line 79: nit: if we only support single GPU this is just the batch size

recipes/eleuther_eval.py
```diff
@@ -4,40 +4,301 @@
...
+
+    @property
+    def truncation(self):
+        return True
+
+    def tok_encode(self, string, **kwargs) -> List[int]:
+        # Note on add_bos flag: setting to False as this gives better results, for example
+        # +1% on truthfulqa_mc2 with a LoRA finetune. lit-gpt also sets this to False,
+        # see https://github.com/Lightning-AI/lit-gpt/blob/main/eval/lm_eval_harness.py#
+        # L66, though notably fast-gpt does the opposite
```
ebsmothers, line 169: Is this still applicable for multimodal?

tests/recipes/test_eleuther_eval.py
joecummings: I've decided to be a bad person and add it in a follow-up.

recipes/eleuther_eval.py
joecummings: Seems intrusive no?

recipes/eleuther_eval.py
ebsmothers: In general `Transform` may not have `tokenizer` field, right? May wanna add a check or something

recipes/eleuther_eval.py
pbontrager: Yeah, this should be in the model transform, I think we enforce 3 channels now but not rgb vs bgr which might be happening here.

recipes/eleuther_eval.py
pbontrager: encoder mask should take care of this. Also, since all of the images are prepended, they're all visible by all of the tokens following the system prompt.

recipes/eleuther_eval.py
joecummings: Fixed.

.github/workflows/recipe_test.yaml
joecummings: no u

recipes/eleuther_eval.py
```diff
@@ -4,40 +4,301 @@
...
+                if image.mode != "RGB":
+                    image = image.convert("RGB")
+                proper_images.append(image)
+
+            # Construct the messages
+            messages = []
+            content = format_content_with_images(
+                text, image_tag=self._image_tag, images=proper_images
+            )
+            messages.append(Message(role="user", content=content))
```
ebsmothers, line 201: Dumb q: how do we support system messages?

recipes/eleuther_eval.py
ebsmothers: Is there a specific blocker here? At least looking at it naively I don't get why we can't just pass nonzero temperature to sample below

recipes/eleuther_eval.py
joecummings: MMMU does not have system messages so we do not support system messages right now.

recipes/eleuther_eval.py
```diff
@@ -4,40 +4,299 @@
...
+                )
+            )
+            input_pos = torch.arange(self.max_length)
+
+        batch["input_pos"] = input_pos[None, :seq_len]
+        batch["mask"] = causal_mask[None, :seq_len]
+
+        # 4. Prefill step
+        generated_tokens = []
+        logits = self.model(prompt, **batch)[:, -1]
```
ebsmothers, line 269: Just wondering: why do we pop prompt earlier just to get shape and then reinsert it here? Seems unintuitive

recipes/eleuther_eval.py
ebsmothers: Yeah we've had similar discussions elsewhere about cases where you want to extract e.g. CLS token without projecting every element of the sequence. Worth thinking about but def not for this PR

recipes/eleuther_eval.py
```diff
@@ -57,13 +317,16 @@ def __init__(
         max_seq_length: int = 4096,
         batch_size: int = 8,
         dtype: torch.dtype = torch.float32,
+        enable_kv_cache: bool = True,
     ):
+        # TODO (@joecummings): Remove this init function so we don't load in extraneous stuff
```
ebsmothers, line 322: Sorry what does this mean? Are some of these fields unused?

recipes/eleuther_eval.py
```diff
@@ -163,101 +440,84 @@ class EleutherEvalRecipe(EvalRecipeInterface):
...
-            model = config.instantiate(model_cfg)
-
-        if self._quantization_mode is not None:
-            model = self._quantizer.quantize(model)
-            model = model.to(device=self._device, dtype=self._dtype)
-            for k, v in model_state_dict.items():
-                model_state_dict[k] = v.to(self._device)
-            model.load_state_dict(model_state_dict, assign=True)
-        else:
-            model.load_state_dict(model_state_dict)
```
ebsmothers, line 225: Was this changed deliberately? Based on #1403 I think we want to load state dict with assign=True when quantization is enabled

recipes/eleuther_eval.py
SalmanMohammadi: I think these properties are overriding the parent class from the harness, right?

recipes/eleuther_eval.py
```diff
@@ -4,40 +4,299 @@
...
+        bsz, seq_len = prompt.shape
+
+        temperature = generation_kwargs.get("temperature", 0.0)
+        do_sample = generation_kwargs.get("do_sample", False)
+        if do_sample or temperature != 0.0:
+            raise RuntimeError(
+                "Any decoding strategy other than greedy is not supported."
+            )
+
+        if bsz > 1:
```
SalmanMohammadi, line 238: Bit out of the loop here, what needs to change to make this happen?

recipes/eleuther_eval.py
```diff
@@ -4,40 +4,299 @@
...
+        # 1. Setup caches for a given batch size
+        with self.device:
+            if self.model.caches_are_enabled():
+                self.model.reset_caches()
+            else:
+                self.model.setup_caches(
+                    batch_size=1,
+                    dtype=self._dtype,
+                    encoder_max_seq_len=self.model_transform.image_seq_len
+                    * self._max_images_per_sample,
```
SalmanMohammadi, line 253: nit
```suggestion
                    encoder_max_seq_len= (self.model_transform.image_seq_len
                    * self._max_images_per_sample),
```

SalmanMohammadi: unless this turns it into a tuple or smth

recipes/eleuther_eval.py
SalmanMohammadi: https://pytorch.org/docs/stable/generated/torch.isin.html

recipes/eleuther_eval.py
SalmanMohammadi: Maybe worth adding to our beefier generate utils for sweet sweet perf @felipemello1 would be interested in your thoughts on how we can do this non-intrusively (happy to chat on Discord or a separate issue)

recipes/eleuther_eval.py
```diff
@@ -163,101 +440,84 @@ class EleutherEvalRecipe(EvalRecipeInterface):
 
     Features:
         - Single GPU evaluation. Multi-GPU evaluation is currently not supported.
-        - Loading model in fp32 or bf16. Fp16 is currently not supported.
-
-    We recommend launching evaluation using the tune CLI:
+        - Quantization and torch.compile (for text-only models) is supported.
```
SalmanMohammadi, line 443: Are you planning to add in compile support?

.github/workflows/recipe_test.yaml
SalmanMohammadi: Sorry, why are we pinning to a commit here?

recipes/eleuther_eval.py
SalmanMohammadi: I think it's also used below to initialize the model, no?

SalmanMohammadi:
very very nice

tests/recipes/test_eleuther_eval.py
SalmanMohammadi: I can add this back in when I fix this.

recipes/eleuther_eval.py
SalmanMohammadi: Worth just adding Phil's comment in the code?

recipes/eleuther_eval.py
```diff
@@ -4,40 +4,299 @@
...
+                "Any decoding strategy other than greedy is not supported."
+            )
+
+        if bsz > 1:
+            raise ValueError(
+                f"Got a batch size of '{bsz}'. Batch size > 1 is not yet supported for "
+                "multimodal generation."
+            )
+
+        # 1. Setup caches for a given batch size
```
SalmanMohammadi, line 244: We're not actually doing this rn

recipes/eleuther_eval.py
```diff
@@ -4,40 +4,299 @@
...
+        with self.device:
+            if self.model.caches_are_enabled():
+                self.model.reset_caches()
+            else:
+                self.model.setup_caches(
```
SalmanMohammadi, line 249: What if `self._enable_kv_cache=False`?

recipes/eleuther_eval.py
joecummings: A multimodal transform will

recipes/eleuther_eval.py
joecummings: Mirrors our generate recipe where the tokens used in the forward pass are passed in positionally and the batch is unrolled via double asterisk.

recipes/eleuther_eval.py
joecummings: Uhmmmmm, turns out we actually load in a copy of a GPT-2 model (not that large so es okay) when we do this call. We overwrite anything that would affect generation, but it's extra memory overhead we shouldn't need.

recipes/eleuther_eval.py
```diff
@@ -119,9 +386,22 @@ def tok_decode(self, tokens: Union[List[int], int], **kwargs) -> str:
...
+    @torch.inference_mode()
     def _model_generate(
         self, context: torch.Tensor, **generation_kwargs
     ) -> torch.Tensor:
+
+        # Setup KV caches OR reset them if they're already set up
+        if self.model.caches_are_enabled():
+            self.model.reset_caches()
+        else:
+            with self.device:
```
SalmanMohammadi, line 398: same comment above about`self._enable_kv_cache`

recipes/eleuther_eval.py
joecummings: Yeah good catch. I'm actually considering turning off this option since using the kv cache is strictly faster. 

For text-only models, this was not a huge deal, but for multimodal models, things are getting slowwwwww. I'll drop a note that this is what we're doing.

joecummings:
> What's a guy gotta do to see some outputs round here?

```
|                 Tasks                 |Version|Filter|n-shot|Metric|   |Value |   |Stderr|
|---------------------------------------|------:|------|------|------|---|-----:|---|-----:|
|mmmu_val                               |      0|none  |      |acc   |↑  |0.3233|±  |0.0271|
| - Art and Design                      |      0|none  |      |acc   |↑  |0.2000|±  |0.0656|
|  - Art                                |      0|none  |None  |acc   |↑  |0.1000|±  |0.1000|
|  - Art Theory                         |      0|none  |None  |acc   |↑  |0.2000|±  |0.1333|
|  - Design                             |      0|none  |None  |acc   |↑  |0.2000|±  |0.1333|
|  - Music                              |      0|none  |None  |acc   |↑  |0.3000|±  |0.1528|
| - Business                            |      0|none  |      |acc   |↑  |0.3200|±  |0.0680|
|  - Accounting                         |      0|none  |None  |acc   |↑  |0.3000|±  |0.1528|
|  - Economics                          |      0|none  |None  |acc   |↑  |0.3000|±  |0.1528|
|  - Finance                            |      0|none  |None  |acc   |↑  |0.2000|±  |0.1333|
|  - Manage                             |      0|none  |None  |acc   |↑  |0.5000|±  |0.1667|
|  - Marketing                          |      0|none  |None  |acc   |↑  |0.3000|±  |0.1528|
| - Health and Medicine                 |      0|none  |      |acc   |↑  |0.3600|±  |0.0706|
|  - Basic Medical Science              |      0|none  |None  |acc   |↑  |0.3000|±  |0.1528|
|  - Clinical Medicine                  |      0|none  |None  |acc   |↑  |0.5000|±  |0.1667|
|  - Diagnostics and Laboratory Medicine|      0|none  |None  |acc   |↑  |0.3000|±  |0.1528|
|  - Pharmacy                           |      0|none  |None  |acc   |↑  |0.3000|±  |0.1528|
|  - Public Health                      |      0|none  |None  |acc   |↑  |0.4000|±  |0.1633|
| - Humanities and Social Science       |      0|none  |      |acc   |↑  |0.4000|±  |0.0791|
|  - History                            |      0|none  |None  |acc   |↑  |0.2000|±  |0.1333|
|  - Literature                         |      0|none  |None  |acc   |↑  |0.5000|±  |0.1667|
|  - Psychology                         |      0|none  |None  |acc   |↑  |0.4000|±  |0.1633|
|  - Sociology                          |      0|none  |None  |acc   |↑  |0.5000|±  |0.1667|
| - Science                             |      0|none  |      |acc   |↑  |0.3000|±  |0.0585|
|  - Biology                            |      0|none  |None  |acc   |↑  |0.2000|±  |0.1333|
|  - Chemistry                          |      0|none  |None  |acc   |↑  |0.0000|±  |0.0000|
|  - Geography                          |      0|none  |None  |acc   |↑  |0.2000|±  |0.1333|
|  - Math                               |      0|none  |None  |acc   |↑  |0.4000|±  |0.1633|
|  - Physics                            |      0|none  |None  |acc   |↑  |0.7000|±  |0.1528|
| - Tech and Engineering                |      0|none  |      |acc   |↑  |0.3429|±  |0.0583|
|  - Agriculture                        |      0|none  |None  |acc   |↑  |0.2000|±  |0.1333|
|  - Architecture and Engineering       |      0|none  |None  |acc   |↑  |0.4000|±  |0.1633|
|  - Computer Science                   |      0|none  |None  |acc   |↑  |0.4000|±  |0.1633|
|  - Electronics                        |      0|none  |None  |acc   |↑  |0.3000|±  |0.1528|
|  - Energy and Power                   |      0|none  |None  |acc   |↑  |0.4000|±  |0.1633|
|  - Materials                          |      0|none  |None  |acc   |↑  |0.2000|±  |0.1333|
|  - Mechanical Engineering             |      0|none  |None  |acc   |↑  |0.5000|±  |0.1667|
```

recipes/eleuther_eval.py
joecummings: Yes definitely.

recipes/eleuther_eval.py
joecummings: So currently we have no utils that support batching for Fusion models (encoder_input and encoder_mask is never accounted for). My plan would be to add it here first, then upstream it to the generate function.

SalmanMohammadi (approved):

recipes/eleuther_eval.py
joecummings: Yeah.

recipes/eleuther_eval.py
RdoubleA: the transform should have the encode method directly, no?

but yeah in general, I'm leaning towards just calling them both tokenizers now since it still takes in a list of messages...

recipes/eleuther_eval.py
RdoubleA: agree with keeping this consistent

recipes/eleuther_eval.py
joecummings: Makes sense - we should have a unified language going forward and do a s/ replace globally at some point.

WRT this specific issue, no I don't think transforms are guaranteed to have the encode method. And the encode output is NOT being used as input to the model - it's just for sorting lengths of inputs on the backend for Eleuther.

recipes/eleuther_eval.py
SalmanMohammadi: Take out compile for now?

recipes/eleuther_eval.py
ebsmothers: How big? If nontrivial let's def file a follow-up task (I assume we can just load in some dummy empty model instead?)

ebsmothers (approved):

PR #1670, opened 2024-09-25
Initial multimodal compile support

ebsmothers:
For multimodal models we are currently unable to compile both the encoder and the decoder due to some shape inference from the encoder bleeding into the decoder (specifically the seq len in the RoPE embeddings). This is a short-term fix to at least enable compiling the decoder. Additionally, it adds support for compiling cross-attention layers.

Test plan:

Run

```
tune run lora_finetune_single_device --config llama3_1/8B_lora_single_device metric_logger=torchtune.training.metric_logging.WandBLogger metric_logger.project=compile-test metric_logger.name=compile-off gradient_accumulation_steps=1
```

and 

```
tune run lora_finetune_single_device --config llama3_1/8B_lora_single_device metric_logger=torchtune.training.metric_logging.WandBLogger metric_logger.project=compile-test metric_logger.name=compile-on compile=True gradient_accumulation_steps=1
```

to confirm we still get compile speedups. Results:

Screenshot 2024-09-24 at 8 16 35 PM

### DIFF
--- a/torchtune/training/_compile.py
+++ b/torchtune/training/_compile.py
@@ -40,11 +40,12 @@ def compile_model(
 
     """
     backend = os.environ.get("TORCH_COMPILE_BACKEND", "inductor")
-    model_to_compile = model if isinstance(model, TransformerDecoder) else model.decoder
+    if isinstance(model, DeepFusionModel):
+        model = model.decoder
     if torch_version_ge("2.5.0"):
         if verbose:
             log.info("Compiling model layers with torch.compile...")
-        for m in reversed(list(model_to_compile.modules())):
+        for m in reversed(list(model.modules())):
             if isinstance(m, TransformerSelfAttentionLayer) or isinstance(
                 m, TransformerCrossAttentionLayer
             ):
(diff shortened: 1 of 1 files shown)


torchtune/training/_compile.py
```diff
@@ -5,38 +5,49 @@
...
+        model (Union[TransformerDecoder, DeepFusionModel]): A model to compile.
+            Can be a TransformerDecoder or DeepFusionModel; in the latter case only
+            the model's decoder will be compiled.
         verbose (bool): Whether to log compile info. Default: True
     Returns:
         None
 
     """
     backend = os.environ.get("TORCH_COMPILE_BACKEND", "inductor")
+    model_to_compile = model if isinstance(model, TransformerDecoder) else model.decoder
```
pbontrager, line 43: nit: I'd rather just keep it as model with a patch "if isinstance(DeepFusionModel): model = model.decoder"

pbontrager (approved):
One comment, but looks good.

torchtune/training/_compile.py
RdoubleA: The inverse logic of checking if Deep Fusion model seems tighter

felipemello1:
n00b question: can we still compile the encoder if we just add disable.compile to the positional embeddings of the encoder?

ebsmothers:
> n00b question: can we still compile the encoder if we just add disable.compile to the positional embeddings of the encoder?

No because the encoder does not even have position embeddings (I know..)

PR #1674, opened 2024-09-25
Use image_seq_len to determine proper cache size in generate_v2

joecummings:


recipes/dev/generate_v2.py
```diff
@@ -133,9 +133,7 @@ def generate(self, cfg: DictConfig):
                 batch_size=1,
                 dtype=self._dtype,
                 encoder_max_seq_len=(
-                    model_inputs["encoder_mask"][0].size(1)
-                    if is_multimodal_input
-                    else None
+                    self.model_transform.image_seq_len if is_multimodal_input else None
```
ebsmothers, line 136: Rn this is only supported for `FlamingoTransform`, right? Should probably assert or something

ebsmothers (approved):

PR #1678, opened 2024-09-25
convert rgba to rgb

ebsmothers:
For datasets with RGBA samples we will error out with shape mismatch since we assume 3 channels elsewhere. This change will convert them to RGB up front


RdoubleA (approved):

vancoykendall:
@ebsmothers Why don't all non-`RGB` images get converted? https://github.com/pytorch/torchtune/issues/1975

ebsmothers:
@vancoykendall thanks for raising this. Just commented on the issue

PR #1681, opened 2024-09-25
Do not run CI on forked repos

janeyx99:
#### Context
- [ ] add a new feature
- [ ] fix a bug
- [x] other (please add here)

https://github.com/janeyx99/torchtune/actions/runs/11004925000 Every night I get 2 spammy emails telling me my CI doesn't work, but like duh it doesn't work on my fork. Thanks @felipemello1  for explaining the problem, and here's the PR to gate CI behind this repo! 

FYI this experience is not very fun for OSS contributors because CI will run on their forks, and they will either get errors, or compute will be wasted on running the periodic CI. 

#### Changelog
* Mentioned above

#### Test plan

#### UX

### DIFF
--- a/.github/workflows/build_docs.yaml
+++ b/.github/workflows/build_docs.yaml
@@ -21,6 +21,7 @@ defaults:
 
 jobs:
   build_docs:
+    if: github.repository_owner == 'pytorch'
     runs-on: ubuntu-latest
     strategy:
       matrix:
@@ -57,7 +58,7 @@ jobs:
   doc-preview:
     runs-on: [self-hosted, linux.2xlarge]
     needs: build_docs
-    if: ${{ github.event_name == 'pull_request' }}
+    if: ${{ github.repository_owner == 'pytorch' && github.event_name == 'pull_request' }}
     steps:
       - name: Checkout
         uses: actions/checkout@v3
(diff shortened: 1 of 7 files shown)


felipemello1:
Thanks, @janeyx99! You are a champ! Do you know the best way to test it? Just let the CI run?

janeyx99:
@felipemello1 Yes, nothing should be changed in torchtune CI (and I should stop getting emails).

.github/workflows/build_docs.yaml
```diff
@@ -82,7 +83,7 @@ jobs:
   upload:
     runs-on: ubuntu-latest
     needs: build_docs
-    if: github.repository == 'pytorch/torchtune' && github.event_name == 'push' && (github.ref == 'refs/heads/main' || startsWith(github.ref, 'refs/tags/v'))
+    if: github.repository_owner == 'pytorch' && github.event_name == 'push' && (github.ref == 'refs/heads/main' || startsWith(github.ref, 'refs/tags/v'))
```
ebsmothers, line 86: I'm curious about this one in particular.. why isn't `github.repository == 'pytorch/torchtune'` sufficient? My assumption is that forks would have a separate repository name, is that not the case?

ebsmothers (approved):
This is why I was getting all those emails?! Seriously though.. thank you for fixing this

.github/workflows/build_docs.yaml
janeyx99: it totally was sufficient, i just changed it for consistency so someone looking can see that it’s the same across the workflows

PR #1693, opened 2024-09-26
Fix Mistral tokenizer if no eos_token is added

devxinvestor:
#### Context
- [ ] add a new feature
- [x] fix a bug
- [ ] other (please add here)

#### Changelog
* Added an add_bos and add_eos attribute to ModelTokenizer to reference in tokenize_messages_no_special_tokens. If not special condition is met, than add_bos and add_eos are set to false, and the tokenized_messages and mask are treated accordingly. 

#### Test plan

#### UX


RdoubleA:
Hey @devxinvestor, thanks for opening this PR. I believe the change should occur in the Mistral tokenizer itself rather than in `tokenize_messages_no_special_tokens`. In that function, BOS and EOS are controlled by passing in the BOS/EOS id from the caller. If they are not passed in, then it is treated as `add_eos=False`. In Mistral, it seems like `tokenize_messages` just adds the EOS by default without any control from the user. Perhaps we can just add a flag to `tokenize_messages` here:
https://github.com/pytorch/torchtune/blob/57dc6f32398967ff59dd585cbb1c4a4b2d9b6ebe/torchtune/models/mistral/_tokenizer.py#L125 

and just pass in the EOS id only if this flag is true?
https://github.com/pytorch/torchtune/blob/57dc6f32398967ff59dd585cbb1c4a4b2d9b6ebe/torchtune/models/mistral/_tokenizer.py#L170

krammnic:
Probably we can close this as it was fixed in #1806

ebsmothers:
Yeah agreed @krammnic I'll close this one. Thanks for your PR @devxinvestor!

PR #1695, opened 2024-09-26
Resizable image positional embeddings

felipemello1:
#### Context
- [x] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

When loading state dict for image models, the positional embeddings may have different shape. This PR allows the reshaping of the embeddings to match the target desired shape that the model was initialized with.

TLDR of the steps:
1) permute shapes of the input embedding, so that the shapes you are permuting are the last two in the tensor
2) pass to F.interpolate the shape of your instantiated embedding
3) now your input embedding have the same shape as the instantiated one

Unit tests and docstrings should help understand the numbers

#### Changelog
added resizing to match fairs implementation here:
1) https://github.com/fairinternal/internal-llama-models/blob/c14773ebe252064fa88d7f613d3bee1757480ae7/models/llama3/reference_impl/multimodal/model.py#L411
2) https://github.com/fairinternal/internal-llama-models/blob/c14773ebe252064fa88d7f613d3bee1757480ae7/models/llama3/reference_impl/multimodal/model.py#L829
3) added rst for clip

#### Test plan
- New unit tests
- Checkpoint loads for instruct and pretraining
- NOT ABLE TO ACTUALLY RUN AND SEE LOSS, GIVEN MY SDPA ISSUE


recipes/configs/llama3_2_vision/11B_full_single_device_pretrained.yaml
```diff
@@ -0,0 +1,104 @@
+# Config for single device full finetuning in full_finetune_single_device.py
```
ebsmothers, line 1: What's the purpose of adding this? Just so that we support the non-instruct version of the model? I'm a bit confused cause I thought one big diff with instruct-tuned vs not is the extra trainable special tokens on the text size, which this PR doesn't address

felipemello1: this is just for testing. I would remove it before the PR is ready

felipemello1: regarding the special token, thats a question for @pbontrager . I am not sure.

tests/torchtune/models/clip/test_pos_embedding_interpolation.py
```diff
@@ -0,0 +1,159 @@
...
+from torchtune.models.clip._position_embeddings import (
+    TiledTokenPositionalEmbedding,
+    TilePositionalEmbedding,
+)
+
+# generated comparing vs fairinternal/internal-llama-models
+tile_pos_emb_test_cases = [
+    {
+        "tgt_num_tiles": 1,
+        # [max_num_tiles, max_num_tiles, -1, embed_dim] -> (2, 2, 2, 3)
```
ebsmothers, line 21: Sorry I don't fully follow these comments in each test case

torchtune/models/clip/_position_embeddings.py
```diff
@@ -100,23 +100,314 @@ def __init__(
 
         self.gate = nn.Parameter(torch.zeros(1))
 
+        self._register_load_state_dict_pre_hook(self._load_state_dict_hook)
+
+    def _load_state_dict_hook(
```
ebsmothers, line 105: I don't see any test case for this, which arguably is the important part

torchtune/models/clip/_position_embeddings.py
```diff
@@ -100,23 +100,314 @@ def __init__(
...
+
+        Raises:
+            ValueError: if loaded local or global embedding n_tokens_per_tile is not derived
+                from a squared grid.
+            ValueError: if after interpolation, the shape of the loaded local embedding
+                is not compatible with the current embedding.
+            ValueError: if after interpolation, the shape of the loaded global embedding
+                is not compatible with the current embedding.
+        """
+        # local
```
ebsmothers, line 134: remove (unless you meant to add something else here)

torchtune/models/clip/_position_embeddings.py
```diff
@@ -100,23 +100,314 @@ def __init__(
...
+            if math.sqrt(inpt_n_tokens_per_tile - 1) % 1 != 0:
+                raise ValueError(
+                    f"Loaded local positional embedding has shape {inpt_n_tokens_per_tile=}, "
+                    f"which indicates a grid_size that is not squared. This is currently not supported."
+                )
+
+            # instantiated pos emb
+            (
+                tgt_max_num_tiles_x,
+                tgt_max_num_tiles_y,
```
ebsmothers, line 185: These will never differ, right? Maybe just use `tgt_max_num_tiles, _` instead?

torchtune/models/clip/_position_embeddings.py
```diff
@@ -100,23 +100,314 @@ def __init__(
...
+        Args:
+            inpt_pos_embed (torch.Tensor): The position embeddings tensor to be resized. It
+                has shape [n_tokens_per_tile, emb_dim], where the first token is the CLS token
+                and n_tokens_per_tile = patch_grid_size**2 + 1.
+            tgt_patch_grid_size (int): The target size of each patch grid, i.e.,
+                the square root of the number of tokens per tile, excluding the class token.
+
+        Returns:
+            torch.Tensor: The resized position embeddings tensor of shape
+                [tgt_patch_grid_size, dim].
```
ebsmothers, line 228: Is this right? Shouldn't it be `[n_tokens_per_tile, emb_dim]`, but where `n_tokens_per_tile = (tgt_patch_grid_size ** 2) + 1`?

torchtune/models/clip/_position_embeddings.py
```diff
@@ -100,23 +100,314 @@ def __init__(
...
+            >>> import math
+            >>> inpt_pos_embed = torch.randn((10*10+1, 64))  # Example input tensor
+            >>> tgt_patch_grid_size = 20  # Target number of tokens per tile
+            >>> resized_pos_embed = _resize_local_position_embedding(inpt_pos_embed, tgt_patch_grid_size)
+            >>> print(resized_pos_embed.shape)
+            torch.Size([20*20+1, 64])
+        """
+        # inverse n_tokens_per_tile = patch_grid_size**2 + 1, where +1 is the cls token
+        inpt_n_tokens_per_tile, inpt_embed_dim = inpt_pos_embed.shape
+        inpt_patch_grid_size = int(math.sqrt(inpt_n_tokens_per_tile - 1))
```
ebsmothers, line 241: Kind of a nit, but why do you compute it in the method for local pos embeddings and pass it to the method for global pos embeddings?

torchtune/models/clip/_position_embeddings.py
```diff
@@ -100,23 +100,314 @@ def __init__(
...
+                inpt_global_pos_embed.shape
+                != self.global_token_positional_embedding.shape
+            ):
+                raise ValueError(
+                    f"Loaded global positional embedding has shape {inpt_global_pos_embed.shape}, "
+                    f"after interpolation. Expected shape {self.global_token_positional_embedding.shape}."
+                )
+
+    @staticmethod
+    def _resize_local_position_embedding(
```
ebsmothers, line 211: Dumb q: do we wanna wrap these methods in `no_grad`? (Maybe it's moot as they're currently used)

torchtune/models/clip/_position_embeddings.py
```diff
@@ -100,23 +100,314 @@ def __init__(
...
+        # we also add a batch dim to the tensor, since F.interpolate expects it
+        inpt_pos_embed = inpt_pos_embed.reshape(
+            1, inpt_patch_grid_size, inpt_patch_grid_size, -1
+        ).permute(0, 3, 1, 2)
+
+        inpt_pos_embed = F.interpolate(
+            inpt_pos_embed,
+            size=[tgt_patch_grid_size, tgt_patch_grid_size],
+            mode="bilinear",
+            align_corners=True,
```
ebsmothers, line 261: I assume this is to match the reference implementation? Might be worth adding a comment or something, otherwise it's not clear to me why we do this

torchtune/models/clip/_position_embeddings.py
```diff
@@ -100,23 +100,314 @@ def __init__(
...
+            embed_dim,
+        )
+
+        # combine max_num_tiles and patch_grid_size into one dimension
+        pos_embed = pos_embed.permute(0, 2, 1, 3, 4).contiguous()
+        pos_embed = pos_embed.reshape(
+            max_num_tiles_x * inpt_patch_grid_size,
+            max_num_tiles_y * inpt_patch_grid_size,
+            embed_dim,
+        )
```
ebsmothers, line 340: Do we really need all 3 of these?

torchtune/models/clip/_position_embeddings.py
felipemello1: I thought about it, but i thought of making it explicit so the user can understand what is in this dimension. What do you think?

tests/torchtune/models/clip/test_pos_embedding_interpolation.py
felipemello1: just trying to help the reader understand the dimensions provided. I can make it better. -1 is because the actual pos embedding has dim=1 there, but when i created the tests, i created with 2.

torchtune/models/clip/_position_embeddings.py
felipemello1: it mostly calls the other functions. But it makes sense to create a unit test for it.

torchtune/models/clip/_position_embeddings.py
felipemello1: just trying to help the reader understand that this is the local portion, and later there is the global portion. But maybe not useful at all. Will remove

torchtune/models/clip/_position_embeddings.py
felipemello1: will make it consistent. I dont think that there is a reason

torchtune/models/clip/_position_embeddings.py
felipemello1: it shouldnt do any harm. Will add

torchtune/models/clip/_position_embeddings.py
felipemello1: i think so. I dont see other way to get the same shape/order. I will ask metamate :P

torchtune/models/clip/_position_embeddings.py
ebsmothers: Yeah no strong preference tbh, it's fine as is too. But might just add a comment somewhere, I think the usage of just `tgt_max_num_tiles_x` in L193 made me take a second glance

torchtune/models/clip/_position_embeddings.py
felipemello1: took a closer look, in both methods i am passing "tgt_patch_grid_size"

```
inpt_local_pos_embed = self._resize_local_position_embedding(
                inpt_local_pos_embed,
                tgt_patch_grid_size=int(math.sqrt(tgt_n_tokens_per_tile - 1)),
            )

inpt_global_pos_embed = self._resize_global_position_embedding(
                global_pos_embed=inpt_global_pos_embed,
                tgt_max_num_tiles=tgt_max_num_tiles_x,
                tgt_patch_grid_size=int(math.sqrt(tgt_n_tokens_per_tile - 1)),
            )
```

torchtune/models/clip/_position_embeddings.py
```diff
@@ -100,23 +100,322 @@ def __init__(
 
         self.gate = nn.Parameter(torch.zeros(1))
 
+        self._register_load_state_dict_pre_hook(self._load_state_dict_hook)
```
pbontrager, line 103: Add a TODO here to switch to the public method after 2.5 is stable

torchtune/models/clip/_position_embeddings.py
```diff
@@ -166,16 +464,127 @@ def __init__(
         )
         self.gate = nn.Parameter(torch.zeros(1))
 
+        # Register load hook to interpolate positional embeddings
+        self._register_load_state_dict_pre_hook(self._load_state_dict_hook)
```
pbontrager, line 468: Same comment for TODO

pbontrager (approved):
Looks good! Please add the TODOs from the comments and an additional one to support resizing for the none-tiled positional embedding.

PR #1706, opened 2024-09-27
MPS support

### LINKED ISSUE #1215: Support for MPS?
I was wondering if torchtune could support apple silicon. It looks like all the examples are for CUDA.

SalmanMohammadi:
#### Context
- [x] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

closes #1215

#### Changelog

Added MPS support. Tested with `generate_v2`, `eleuther_eval`, `lora_finetune_single_device`

#### Test plan

#### UX

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -59,6 +59,7 @@ Note that the above flags can be combined with other pytest flags, so e.g. `pyte
 > For contributors who are developing on MPS, you may find that several of our tests fail due to precision differences falling outside of numerical tolerances on our tests. As of 27/09/2024 the following tests will fail for this reason:
 <details>
   <summary>Failed Tests</summary>
+
   ```bash
   FAILED tests/torchtune/generation/test_generation.py::TestGenerate::test_stop_tokens_batched[prompt_tokens_batched-generation_model_no_kv_cache] - assert False
   FAILED tests/torchtune/generation/test_generation.py::TestGenerate::test_stop_tokens_batched[prompt_tokens_batched-generation_model_kv_cache_batched] - assert False
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -55,6 +55,33 @@ Whenever running tests in torchtune, favor using the command line flags as much 
 
 Note that the above flags can be combined with other pytest flags, so e.g. `pytest tests -m integration_test -k 'test_loss'` will run only recipe tests matching the substring `test_loss`.
 
+> [!NOTE]
+> For contributors who are developing on MPS, you may find that several of our tests fail due to precision differences falling outside of numerical tolerances on our tests. As of 27/09/2024 the following tests will fail for this reason:
+<details>
+  <summary>Failed Tests</summary>
+  ```bash
+  FAILED tests/torchtune/generation/test_generation.py::TestGenerate::test_stop_tokens_batched[prompt_tokens_batched-generation_model_no_kv_cache] - assert False
+  FAILED tests/torchtune/generation/test_generation.py::TestGenerate::test_stop_tokens_batched[prompt_tokens_batched-generation_model_kv_cache_batched] - assert False
+  FAILED tests/torchtune/generation/test_generation.py::TestGenerate::test_stop_tokens_batched[prompt_tokens_batched_left_padded-generation_model_no_kv_cache] - assert False
+  FAILED tests/torchtune/generation/test_generation.py::TestGenerate::test_stop_tokens_batched[prompt_tokens_batched_left_padded-generation_model_kv_cache_batched] - assert False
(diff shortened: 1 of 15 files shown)


torchtune/generation/_generation.py
```diff
@@ -325,6 +321,11 @@ def generate(
 
     # keeps track at a high level if we've already hit a stop token in a sequence so we can early stop
     stop_token_reached = torch.zeros(bsz, dtype=torch.bool, device=prompt.device)
+    stop_tokens = (
```
joecummings, line 324: Any reason?

torchtune/generation/_generation.py
```diff
@@ -325,6 +321,11 @@ def generate(
...
+        torch.tensor(stop_tokens, device=prompt.device, dtype=tokens.dtype)
```
SalmanMohammadi, line 325: Not sure why but outputs from the model on my device were `int32`. The prompt is `int64` and `torch.isin` was complaining when `stop_tokens.dtype != tokens.dtype` later on. I wasn't ready for further investigation.

torchtune/generation/_generation.py
SalmanMohammadi: see below

torchtune/generation/_generation.py
joecummings: carry on

joecummings (approved):

torchtune/generation/_generation.py
SalmanMohammadi:
```python
import torch
for dvc in ["cpu", "cuda"]:
    print(dvc)
    device = torch.device("cuda")
    x = torch.arange(10, device = device, dtype=torch.int64)
    y = torch.arange(10, device = device, dtype=torch.int32)
    torch.isin(x, y)
```
So fine on CPU/CUDA
BUT this seems like expected behaviour on MPS? There's actually a test for this in core
https://github.com/pytorch/pytorch/blob/e5a57932f06746a7f6779a97b4e11394b614fa81/test/test_mps.py#L8512

SalmanMohammadi: yeah there's a check for it in the MPS implementation https://github.com/pytorch/pytorch/blob/e5a57932f06746a7f6779a97b4e11394b614fa81/aten/src/ATen/native/mps/operations/TensorCompare.mm#L297

SalmanMohammadi: Maybe because the underlying MPS call doesn't support different datatypes?
https://developer.apple.com/documentation/metalperformanceshadersgraph/mpsgraph/equal(_:_:name:)?language=objc

SalmanMohammadi: This is probably it and there's NO WAY I'm loading up XCode to try it out. I hope you are contented.
https://xkcd.com/356/

CONTRIBUTING.md
```diff
@@ -55,6 +55,34 @@ Whenever running tests in torchtune, favor using the command line flags as much
 
 Note that the above flags can be combined with other pytest flags, so e.g. `pytest tests -m integration_test -k 'test_loss'` will run only recipe tests matching the substring `test_loss`.
 
+> [!NOTE]
+> For contributors who are developing on MPS, you may find that several of our tests fail due to precision differences falling outside of numerical tolerances on our tests. As of 27/09/2024 the following tests will fail for this reason:
+<details>
+  <summary>Failed Tests</summary>
```
ebsmothers, line 61: Commented in Discord, but why not just (a) add a new method in test_utils.py, (b) decorate each of these test cases with it, and (c) point to an issue tracking all of them in case someone wants to help debug any individual one?

torchtune/modules/kv_cache.py
```diff
@@ -102,7 +102,7 @@ def update(
         k_out = self.k_cache
         v_out = self.v_cache
 
-        k_out.index_copy_(2, cache_pos, k_val)
-        v_out.index_copy_(2, cache_pos, v_val)
+        k_out[:, :, cache_pos] = k_val
+        v_out[:, :, cache_pos] = v_val
```
ebsmothers, line 106: Does this create an extra copy or anything? (Not a blocker, just curious)

ebsmothers (commented):
Would be nice to have an actual gate on those failing unit tests with a pointer to a proper issue for now (otherwise it bloats up our CONTRIBUTING.md a lot). Other than that if you've run the generation recipe (yes just that one and not every single recipe), this looks good to me

torchtune/modules/kv_cache.py
SalmanMohammadi: I'm not 100% sure. I drew inspiration from here https://github.com/huggingface/transformers/blob/2e24ee4dfa39cc0bc264b89edbccc373c8337086/src/transformers/cache_utils.py#L1168

ebsmothers (approved):
It’s beautiful

PR #1726, opened 2024-10-01
QLoRA with bias + Llama 3.2 Vision QLoRA configs

ebsmothers:
After opening https://github.com/pytorch/ao/issues/979 on torchao, it was pointed out to me that I was overcomplicating things.. we can just keep the bias in bf16, which is apparently a pretty standard thing to do ([ref]([maartengrootendorst.com/blog/quantization#weights-and-biases](https://www.maartengrootendorst.com/blog/quantization/#weights-and-biases))). 

So this PR does exactly that.. just let bias stay in the higher precision for our `FrozenNF4Linear`, `LoRALinear`, and `DoRALinear` when we set `quantize_base=True`.

## Test plan

Added test cases for LoRALinear and FrozenNF4Linear with bias=True. 

Fun fact I discovered while writing the LoRALinear test: if x, weight, bias are all bf16 then `F.linear(x, weight, bias) != F.linear(x, weight) + bias` (repro [here](https://gist.github.com/ebsmothers/912cd77132fc6c31cc1972bc10029bbf)). So I left that test case out.

E2E tests: 

```
tune run lora_finetune_single_device --config llama3_2_vision/11B_qlora_single_device \
metric_logger=torchtune.training.metric_logging.WandBLogger metric_logger.project=qlora-32-vision \
metric_logger.name=single-device max_steps_per_epoch=500
```

and

```
tune run --nproc_per_node 2 lora_finetune_distributed --config llama3_2_vision/11B_qlora \
metric_logger=torchtune.training.metric_logging.WandBLogger metric_logger.project=qlora-32-vision \
metric_logger.name=distributed max_steps_per_epoch=500
```

Loss curves from both these runs (compared to analogous LoRA runs as a baseline):

Screenshot 2024-10-23 at 5 29 21 PM

Also run with DoRA applied to MLP and output layers as a sanity check:

```
tune run lora_finetune_single_device --config llama3_2_vision/11B_qlora_single_device \
metric_logger=torchtune.training.metric_logging.WandBLogger metric_logger.project=qlora-32-vision \
metric_logger.name=single-device max_steps_per_epoch=5 model.apply_lora_to_mlp=True model.apply_lora_to_output=True model.use_dora=True max_steps_per_epoch=5 gradient_accumulation_steps=1
...
```


recipes/configs/llama3_2_vision/11B_lora.yaml
```diff
@@ -81,7 +81,7 @@ enable_activation_offloading: False
 dtype: bf16
 
 # Logging
-output_dir: /tmp/full-llama3.2-vision-finetune
+output_dir: /tmp/lora-llama3.2-vision-finetune
```
joecummings, line 84: whoops

torchtune/modules/low_precision/nf4_linear.py
```diff
@@ -18,27 +18,29 @@ class FrozenNF4Linear(nn.Linear):
...
     and is meant to be used as the base Linear layer for modeling
     use cases such as QLoRA where base model parameters are frozen.
-    NOTE: biases are currently not supported.
 
     Args:
         in_dim (int): input dimension
         out_dim (int): output dimension
         device (Optional[torch.device]): device to use for the underlying weight. If ``None``, uses the default
             device given by `torch.get_default_device()`.
+        bias (bool): whether to include bias in the original linear layer. Default: False
```
pbontrager, line 27: nit: remove "original"

tests/torchtune/modules/low_precision/test_nf4_linear.py
```diff
@@ -59,9 +55,10 @@ def test_state_dict(self, dtype):
         assert isinstance(state_dict["weight"], NF4Tensor)
 
     @pytest.mark.parametrize("dtype", [torch.bfloat16, torch.float32])
-    def test_output_dtype(self, dtype):
+    @pytest.mark.parametrize("bias", [True, False])
```
joecummings, line 58: What is the point of adding bias to this test? The dtype isn't changing and you're only checking the dtype?

torchtune/modules/low_precision/nf4_linear.py
```diff
@@ -18,27 +18,29 @@ class FrozenNF4Linear(nn.Linear):
...
+        **kwargs,
     ):
-        if "bias" in kwargs and kwargs.pop("bias"):
-            raise RuntimeError("FrozenNF4Linear does not currently support biases!")
-
-        super().__init__(in_dim, out_dim, device=device, bias=False, **kwargs)
+        super().__init__(in_dim, out_dim, device=device, bias=bias, **kwargs)
         self.weight.requires_grad_(False)
+        if self.bias is not None:
+            self.bias.requires_grad_(False)
```
pbontrager, line 43: Why is requires_grad false here? Is nf4 always false?

tests/torchtune/modules/peft/test_lora.py
```diff
@@ -61,19 +61,22 @@ def lora_linear(self, in_dim, out_dim) -> LoRALinear:
...
+            with training.set_default_dtype(dtype):
+                qlora_linear = LoRALinear(
+                    in_dim=512,
+                    out_dim=512,
+                    rank=RANK,
+                    alpha=ALPHA,
+                    use_bias=use_bias,
+                    quantize_base=True,
+                )
+                # fixed_init_model(qlora_linear, dtype=torch.bfloat16)
```
joecummings, line 75: Did you mean to comment this out?

torchtune/modules/low_precision/nf4_linear.py
```diff
@@ -18,27 +18,29 @@ class FrozenNF4Linear(nn.Linear):
...
+        out_dim: int,
+        device: Optional[torch.device] = None,
+        bias: bool = False,
+        **kwargs,
     ):
-        if "bias" in kwargs and kwargs.pop("bias"):
-            raise RuntimeError("FrozenNF4Linear does not currently support biases!")
-
-        super().__init__(in_dim, out_dim, device=device, bias=False, **kwargs)
+        super().__init__(in_dim, out_dim, device=device, bias=bias, **kwargs)
```
pbontrager, line 40: What is this super doing? Is it safe to assume it won't do anything with the bias we don't want?

joecummings (approved):
Brilliant work. The kind of work they sing about in old minstrel songs. 

One question.

torchtune/modules/peft/lora.py
```diff
@@ -123,6 +119,8 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
         """
         if self._quantize_base:
             out = linear_nf4(input=x, weight=self.weight)
+            if self.use_bias:
```
pbontrager, line 122: Why do we use torchao linear_nf4 here and not our own?

recipes/configs/llama3_2_vision/11B_qlora.yaml
```diff
@@ -0,0 +1,88 @@
...
+# You can add specific overrides through the command line. For example
+# to override the checkpointer directory while launching training:
+#   tune run --nproc_per_node 2 lora_finetune_distributed --config llama3_2_vision/11B_qlora checkpointer.checkpoint_dir=<YOUR_CHECKPOINT_DIR>
+#
+# This config works best when the model is being fine-tuned on 2+ GPUs.
+# For single device QLoRA finetuning please use 11B_qlora_single_device.yaml
+
+# Model arguments
+model:
+  _component_: torchtune.models.llama3_2_vision.qlora_llama3_2_vision_11b
```
pbontrager, line 19: Can you confirm that this is identical to lora except for this line? Whenever you do a merge you should re-check that assumption.

pbontrager (commented):
Thank you Evan for adding this! Added mostly questions, but overall looks good. Could you also run a test with lora applied to mlp and output as well, since those also have bias values?

tests/torchtune/modules/low_precision/test_nf4_linear.py
ebsmothers: Agreed it's pretty trivial but I'd like to at least build `FrozenNF4Linear` with bias somewhere in our unit tests, and the overhead of this unit test is tiny

tests/torchtune/modules/peft/test_lora.py
ebsmothers: Oh yeah oops, lemme update

torchtune/modules/low_precision/nf4_linear.py
ebsmothers: It's the same as before, just passing bias explicitly to nn.Linear init

pbontrager: Why doesn't this create an extra parameter in the parent class? Does it create it and then we override those same variables in our init? I feel it would be safer to inherit from nn.Module

pbontrager: Okay, nvm, I misunderstood what's going on in the init here.

torchtune/modules/low_precision/nf4_linear.py
```diff
@@ -18,27 +18,29 @@ class FrozenNF4Linear(nn.Linear):
...
         self.weight.requires_grad_(False)
+        if self.bias is not None:
+            self.bias.requires_grad_(False)
         self.nf4_weight = to_nf4(self.weight)
```
pbontrager, line 44: nit: do we want this "self." here?

torchtune/modules/peft/lora.py
ebsmothers: Not sure I follow.. we don't have our own linear_nf4. Even our `FrozenNF4Linear` is just a wrapper around their `linear_nf4`

torchtune/modules/low_precision/nf4_linear.py
ebsmothers: Yeah good point, maybe not strictly necessary

recipes/configs/llama3_2_vision/11B_qlora.yaml
ebsmothers: Yep, did this

torchtune/modules/peft/lora.py
pbontrager: I meant, why do we not reuse the FrozenNF4Linear class here so we don't have to define this bias solution in multiple places?

ebsmothers: Discussed a bit offline. It's a nice idea but going to punt on it for now

torchtune/models/llama3_2_vision/_component_builders.py
```diff
@@ -432,11 +447,22 @@ def lora_llama3_2_vision_encoder(
...
+    encoder = Llama3VisionEncoder(clip=clip, projection_head=projection_head)
+
+    if quantize_base:
+        # For QLoRA, we reparametrize 4-bit tensors to bf16, and offload to CPU on the fly
+        # so as to not increase peak memory
+        encoder._register_state_dict_hook(
+            partial(reparametrize_as_dtype_state_dict_post_hook, offload_to_cpu=True)
+        )
+
+    return encoder
```
ebsmothers, line 465: A bunch of miscellaneous linter changes in this file, this is the only one of substance

torchtune/modules/model_fusion/_fusion.py
```diff
@@ -232,10 +232,11 @@ def _load_state_dict_hook(self, state_dict, prefix, *args, **kwargs):
         """Apply extra "embedding" prefix to the state_dict key to
         account for the FusionEmbedding wrapping.
         """
-        key = prefix + "weight"
-        new_key = prefix + "embedding.weight"
-        state_dict[new_key] = state_dict[key]
-        del state_dict[key]
+        if state_dict:
```
ebsmothers, line 235: @pbontrager this is my hack to support DoRA. Lmk if any concerns

pbontrager: I saw why you do this. This seems like a general safety check that would be good to have in all of the load_state_dict hooks as this case could come up anytime "strict=False". I'll approve this but could you add this change here or in a followup in the other load hooks?

PR #1727, opened 2024-10-01
Add HF Checkpoint Format Support for Llama Vision

pbontrager:
#### Context
- [x] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

Added hf_to_tune and tune_to_hf mappings.

#### Changelog
- updated the llama3_2_vision/_convert_weights.py to include the mapping
- updated the _checkpointer.py to call these

#### Test plan
- Tested running 1 epoch of size 1 with the hf checkpoint format `tune run full_finetune_single_device --config llama3_2_vision/11B_full_single_device max_steps_per_epoch=1`
- Verified meta_to_tune(meta_checkpoint) == hf_to_tune(hf_checkpoint)
- Verified tune_to_hf(hf_to_tune(hf_checkpoint)) == hf_checkpoint
- Verified tune_to_meta(meta_to_tune(meta_checkpoint)) == meta_checkpoint
- HF training run showed comparable loss curve to Meta


torchtune/models/llama3_2_vision/_convert_weights.py
```diff
@@ -171,3 +237,188 @@ def llama3_vision_tune_to_meta(
...
+    cross_attention_layers: Optional[List[int]] = None,
+    # Vision Encoder Paramters
+    encoder_dim: int = 1280,
+    tile_size: int = 448,
+    num_tiles: int = 4,
+    supported_aspect_ratios: List[Tuple[int, int]] = None,
+) -> Dict[str, torch.Tensor]:
+    """
+    Convertor from HF state dict to torchtune state dict. This handles:
+    - Updateing the cross attention layer numbers
```
RdoubleA, line 258:
```suggestion
    - Updating the cross attention layer numbers
```

torchtune/models/llama3_2_vision/_convert_weights.py
```diff
@@ -171,3 +237,188 @@ def llama3_vision_tune_to_meta(
...
+            elif new_key == "decoder.tok_embeddings.weight":
+                learned_embedding = "decoder.tok_embeddings.fusion_embedding.weight"
+                converted_state_dict[learned_embedding] = value[vocab_size:]
+                value = value[:vocab_size]
+        elif "vision_model" in key:
+            if (
+                "tile_pos_embed.embedding" in new_key
+                or "global_token_positional_embedding" in new_key
+            ):
+                # WARNING
```
RdoubleA, line 307: ?

torchtune/models/llama3_2_vision/_convert_weights.py
```diff
@@ -171,3 +237,188 @@ def llama3_vision_tune_to_meta(
...
+    cross_attention_layers: Optional[List[int]] = None,
+    # Vision Encoder Paramters
+    encoder_dim: int = 1280,
+    tile_size: int = 448,
+    num_tiles: int = 4,
+    supported_aspect_ratios: List[Tuple[int, int]] = None,
+) -> Dict[str, torch.Tensor]:
+    """
+    Convertor from Tune state dict to HF state dict. This handles:
+    - Updateing the cross attention layer numbers
```
RdoubleA, line 346:
```suggestion
    - Updating the cross attention layer numbers
```

torchtune/training/checkpointing/_checkpointer.py
```diff
@@ -537,8 +555,28 @@ def save_checkpoint(
...
+                )
+
+                text_config = self._config.get("text_config", {})
+                vision_config = self._config.get("vision_config", {})
+                state_dict[training.MODEL_KEY] = llama3_vision_tune_to_hf(
+                    state_dict[training.MODEL_KEY],
+                    num_heads=text_config["num_attention_heads"],
+                    num_kv_heads=text_config["num_key_value_heads"],
+                    dim=text_config["hidden_size"],
+                    head_dim=text_config.get("head_dim", None),
```
RdoubleA, line 569: why do some of these have defaults for `.get` and some do not have defaults? could `text_config` and `vision_config` ever be empty?

pbontrager: I don't know, I was following the base implementation and all lists defaulted to None.

torchtune/models/llama3_2_vision/_convert_weights.py
```diff
@@ -171,3 +237,188 @@ def llama3_vision_tune_to_meta(
...
+            .reshape((head_dim * n_heads), dim)
+        )
+
+    for key, value in state_dict.items():
+        if "rotary_emb.inv_freq" in key:  # Skip loading the position embeddings
+            continue
+        new_key = get_mapped_key(key, _FROM_HF)
+        if "language_model" in key:
+            if "layers" in key:  # Update layer numbers
+                layer = int(key.split(".")[3])
```
ebsmothers, line 282: Not the end of the world, but I feel like half the reason we added the placeholders in e.g. `_FROM_HF` was so that we could avoid stuff like this (e.g. [here](https://github.com/pytorch/torchtune/blob/10b02e0e358952205ab1f6eb7fcbf24b99d4b7fd/torchtune/models/convert_weights.py#L48-L65)) and keep the `hf_to_tune` methods simpler. I know the multimodal models are more involved, so maybe there's no way around it. But still, all of this is a bit hard to parse

torchtune/training/checkpointing/_checkpointer.py
```diff
@@ -467,8 +467,26 @@ def load_checkpoint(self) -> Dict[str, Any]:
                 tie_word_embeddings=self._config["tie_word_embeddings"],
             )
         elif self._model_type == ModelType.LLAMA3_VISION:
-            raise NotImplementedError(
-                "Safe tensors support for LLaMA3 Vision will be added soon. Pleaee use the FullModelMetaCheckpointer for now."
+            from torchtune.models.llama3_2_vision._convert_weights import (
+                llama3_vision_hf_to_tune,
```
ebsmothers, line 471: What's the reason for gating the import here? Due to transitive torchvision dep or something?

torchtune/models/llama3_2_vision/_convert_weights.py
pbontrager: There are a number of challenges here that our setup doesn't handle out of the box and I'm not quite sure if there's a good generalization. The layer numbering scheme for cross_attention is different between ours and theirs. There are also a number of instances where a single one of their keys maps to two of our keys. We could solve the latter generally, but I'm not sure about the former, we might need some layer mapping too. If we add more MM models, we might want to go down that route.

ebsmothers (commented):
Re testing, can you add a check that running with Meta checkpointer and HF checkpointer give the same loss curves?

torchtune/models/llama3_2_vision/_convert_weights.py
ebsmothers: Yeah please update this comment

torchtune/models/llama3_2_vision/_convert_weights.py
```diff
@@ -171,3 +237,188 @@ def llama3_vision_tune_to_meta(
...
+                    key_lst[2] = str(new_layer)
+                new_key = ".".join(key_lst)
+            if "q_proj" in key and "cross_attn" not in key:
+                value = _permute(value, num_heads)
+            elif "k_proj" in key and "cross_attn" not in key:
+                value = _permute(value, num_kv_heads)
+            elif new_key == "decoder.tok_embeddings.weight":
+                learned_embedding = "decoder.tok_embeddings.fusion_embedding.weight"
+                converted_state_dict[learned_embedding] = value[vocab_size:]
+                value = value[:vocab_size]
```
ebsmothers, line 301: Please add a comment for this

torchtune/models/llama3_2_vision/_convert_weights.py
pbontrager: Forgot to finish it.

ebsmothers (approved):
Couple small comments, the main one is that I'd like to see loss curves for the two checkpointers as a sanity check. Stamping to unblock

PR #1729, opened 2024-10-01
Torchtune config for pruned 12 layers, 80% MLP (aligned to group size 256)

ifed-ucsd:
Summary: Torchtune config for pruned 12 layers, 80% MLP (aligned to group size 256)

Differential Revision: D63700598


ebsmothers:
Hi @ifed-ucsd what does this particular builder correspond to? If it's something more experimental (or even just a non-canonical version of Llama 3.2) I'd recommend to put it in a downstream codebase rather than in our core Llama 3.2 builders.

PR #1733, opened 2024-10-01
config changes

felipemello1:
#### Context
- [ ] add a new feature
- [x] fix a bug
- [ ] other (please add here)

#### Changelog
3 mains fixes:
1) llama 3.1 KD config loading wrong checkpoint
2) qwen2 config loading wrong checkpoint
3) llama 70b repeated fused argument + warning that it doesnt work on torch stable
4) llama 405B-lora warning to not use pytorch stable

extra:
added compile flag for llama. I should have done it in another PR, sorry/not sorry.

#### Test plan
ran the configs and saved ckpt

### DIFF
--- a/recipes/configs/llama3/70B_full.yaml
+++ b/recipes/configs/llama3/70B_full.yaml
@@ -99,7 +99,7 @@ device: cuda
 enable_activation_checkpointing: True
 custom_sharded_layers: ['tok_embeddings', 'output']
 fsdp_cpu_offload: True
-compile: False # set it to True for better memory and speed
+compile: False # set it to True for better memory and performance
 
 # Reduced precision
 dtype: bf16
--- a/recipes/configs/llama3/70B_full.yaml
+++ b/recipes/configs/llama3/70B_full.yaml
@@ -14,6 +14,12 @@
 #   tune run --nproc_per_node 8 full_finetune_distributed --config llama3/70B_full checkpointer.checkpoint_dir=<YOUR_CHECKPOINT_DIR>
 #
 # This config is only tested on an 8xA100 machine.
+#
+# !!!!!!!!!!!!!
+# !!!!!!!!!!!!!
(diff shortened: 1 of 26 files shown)


recipes/configs/llama3/70B_full.yaml
```diff
@@ -14,6 +14,12 @@
 #   tune run --nproc_per_node 8 full_finetune_distributed --config llama3/70B_full checkpointer.checkpoint_dir=<YOUR_CHECKPOINT_DIR>
 #
 # This config is only tested on an 8xA100 machine.
+#
+# !!!!!!!!!!!!!
+# !!!!!!!!!!!!!
```
RdoubleA, line 19: nit: I would take the !!!! out personally

recipes/configs/qwen2/knowledge_distillation_single_device.yaml
```diff
@@ -44,9 +44,9 @@ checkpointer:
...
 teacher_checkpointer:
   _component_: torchtune.training.FullModelHFCheckpointer
-  checkpoint_dir: /tmp/Qwen2-1.5B-Instruct-lora-finetune
+  checkpoint_dir: /tmp/Qwen2-1.5B-Instruct
   checkpoint_files: [
-    hf_model_0001_0.pt
+    model.safetensors
   ]
   recipe_checkpoint: null
   output_dir: /tmp/Qwen2-1.5B-Instruct-lora-finetune
```
RdoubleA, line 52: should change output dir here

RdoubleA (approved):

recipes/configs/llama3/70B_full.yaml
felipemello1: i am afraid that users would ignore it without the "!", and not understand why it crashes

recipes/configs/qwen2/knowledge_distillation_single_device.yaml
felipemello1: i thought about it. Indeed in all of our configs we use the same output dir. I will make the change

PR #1737, opened 2024-10-01
Support Optimizer-in-the-backward

mori360:
#### Context
- [x] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

Enable Optimizer-in-the-backward for full_finetune_distributed

#### Changelog

- Update full_finetune_distributed for enabling Optimizer-in-the-backward
- Update test_full_finetune_distributed with `_optimizer_in_bwd` config
- updated test_distributed to test running with/without optimized_in_the_backward, and performance after saving-loading state_dict.

#### Test plan

- Test running with optimizer_in_the_backward: `tune run --nproc_per_node 2 full_finetune_distributed --config llama2/7B_full fsdp_cpu_offload=False max_steps_per_epoch=2 optimizer_in_bwd=True`
- Test running optimizer_in_the_backward with resume_from_checkpoint: ` tune run --nproc_per_node 2 full_finetune_distributed --config llama2/7B_full fsdp_cpu_offload=False max_steps_per_epoch=2 epochs=10 optimizer_in_bwd=True  resume_from_checkpoint=True checkpointer.recipe_checkpoint=/tmp/Llama-2-7b-hf/recipe_state.pt checkpointer.checkpoint_files=[hf_model_0001_1.pt,hf_model_0002_1.pt]`
- Verify that running with Optimizer-in-the-backward could have the same loss, model_state_dict and optimizer_state_dict, model after saving and loading could also have the same: `pytest tests/torchtune/training/test_distributed.py -k test_optimizer_in_backward`

Memory cost analysis:
With each layer gradient cost 193MB memory, the origin(left) case has the peak memory at the 31th layer with accumulation of 193MB memory times 30.
The right case with optimizer-in-the-backward frees these memory during backward, gets lower peak memory.
memory compare

Training time and loss analysis:
training time and loss

### DIFF
--- a/recipes/configs/mistral/evaluation.yaml
+++ b/recipes/configs/mistral/evaluation.yaml
@@ -1,0 +1,41 @@
+# Config for EleutherEvalRecipe in eleuther_eval.py
+#
+# To launch, run the following command:
+#    tune run eleuther_eval --config mistral/evaluation
+
+# Model Arguments
+model:
+  _component_: torchtune.models.mistral.mistral_7b
+
+# Checkpointer
+checkpointer:
+  _component_: torchtune.training.FullModelHFCheckpointer
+  checkpoint_dir: /tmp/Mistral-7B-v0.1/
+  checkpoint_files: [
+    pytorch_model-00001-of-00002.bin,
+    pytorch_model-00002-of-00002.bin
+  ]
+  output_dir: /tmp/Mistral-7B-v0.1/
+  model_type: MISTRAL
+resume_from_checkpoint: False
+
+# Tokenizer
+tokenizer:
+  _component_: torchtune.models.mistral.mistral_tokenizer
+  path: /tmp/Mistral-7B-v0.1/tokenizer.model
+  max_seq_len: null
+
+# Environment
+device: cuda
+dtype: bf16
+seed: 1234 # It is not recommended to change this seed, b/c it matches EleutherAI's default seed
+
+# EleutherAI specific eval args
+tasks: ["truthfulqa_mc2"]
+limit: null
+max_seq_length: 4096
+batch_size: 8
+enable_kv_cache: True
+
+# Quantization specific args
+quantizer: null
--- a/torchtune/_recipe_registry.py
+++ b/torchtune/_recipe_registry.py
@@ -336,6 +336,10 @@ _ALL_RECIPES = [
                 name="phi3/evaluation",
                 file_path="phi3/evaluation.yaml",
             ),
+            Config(
+                name="mistral/evaluation",
+                file_path="mistral/evaluation.yaml",
+            ),
         ],
         supports_distributed=False,
     ),


recipes/full_finetune_distributed.py
```diff
@@ -456,19 +458,56 @@ def _is_layer_fqn(s: str) -> bool:
...
+                )
 
-        if self._is_rank_zero:
-            log.info("Optimizer is initialized.")
-        return optimizer
+            if self._is_rank_zero:
+                log.info("Optimizer is initialized.")
+            return optimizer
+        else:
+            optim_dict = {
```
weifengpy, line 479: I assume the code for "else" branch is shared with single device recipe, except `load_from_full_optimizer_state_dict` ? could we have a util so it's shared between single device and full finetuning?

recipes/full_finetune_distributed.py
```diff
@@ -549,19 +588,25 @@ def save_checkpoint(
             device=self._device,
         )
 
-        if intermediate_checkpoint:
+        if intermediate_checkpoint and not self._optimizer_in_bwd:
```
weifengpy, line 591: if-else `not self._optimizer_in_bwd` inside `if intermediate_checkpoint` ?

recipes/full_finetune_distributed.py
```diff
@@ -681,7 +727,13 @@ def train(self) -> None:
                         time_per_step = time.perf_counter() - t0
                         log_dict = {
                             "loss": loss_to_log,
-                            "lr": self._optimizer.param_groups[0]["lr"],
+                            "lr": (
+                                list(self._optim_ckpt_wrapper.values())[0].param_groups[
```
weifengpy, line 731: make this a util function? probably better to assert on "lr" is the same across param_groups

tests/recipes/test_full_finetune_distributed.py
```diff
@@ -42,6 +42,7 @@ def _get_test_config_overrides(self):
             "optimizer.lr=2e-5",
             "log_every_n_steps=1",
             "clip_grad_norm=100",
+            "optimizer_in_bwd=False",
```
weifengpy, line 45: why need to set it explicitly? is optimizer_in_bwd default to True or False?

weifengpy:
could we draw loss curves in weights & bias to showcase numerics are the same with/without optimizer-in-the-backward?

recipes/full_finetune_distributed.py
mori360: Call the same API from OptimizerInBackwardWrapper now

recipes/full_finetune_distributed.py
```diff
@@ -593,7 +638,8 @@ def train(self) -> None:
         _, rank = training.get_world_size_and_rank()
 
         # zero out the gradients before starting training
-        self._optimizer.zero_grad()
+        if not self._optimizer_in_bwd:
+            self._optimizer.zero_grad()
```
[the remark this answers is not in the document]
mori360: Zero out optim_ckpt_wrapper for optimizer-in-the-backward now

recipes/full_finetune_distributed.py
mori360: add get_lr_scheduler to check whether all the lr_scheduler are the same and return it

mori360:
> could we draw loss curves in weights & bias to showcase numerics are the same with/without optimizer-in-the-backward?

The loss curves have been added in the comments section of the third column on the right-hand side table.

recipes/full_finetune_distributed.py
```diff
@@ -456,19 +458,55 @@ def _is_layer_fqn(s: str) -> bool:
...
+                model=self._model, optim_dict=optim_dict
+            )
+            # Create a wrapper for checkpoint save/load of optimizer states when running in backward.
+            self._optim_ckpt_wrapper = training.create_optim_in_bwd_wrapper(
+                model=self._model, optim_dict=optim_dict
             )
+            # Load optimizer states for each param.
+            if opt_state_dict is not None:
+                for param in opt_state_dict.keys():
+                    torch.distributed.breakpoint()
```
ebsmothers, line 496: remove

recipes/full_finetune_distributed.py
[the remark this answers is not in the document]
ebsmothers: Yeah it's a bit interesting, we actually don't currently do this in the single-device recipe (though maybe we should)

recipes/full_finetune_distributed.py
```diff
@@ -456,19 +458,55 @@ def _is_layer_fqn(s: str) -> bool:
...
+        optimizer_in_bwd: bool = False,
+        opt_state_dict: Optional[Dict[str, Any]] = None,
     ) -> Optimizer:
-        optimizer = config.instantiate(cfg_optimizer, self._model.parameters())
-        if opt_state_dict:
-            training.load_from_full_optimizer_state_dict(
-                optimizer,
-                opt_state_dict,
-                self._device,
+        if not optimizer_in_bwd:
```
ebsmothers, line 466: This is a nit, but let's switch the order of this if/else so that it's `if optimizer_in_bwd`. Then it's closer to what's in the single-device recipe and easier to compare between the two

recipes/full_finetune_distributed.py
```diff
@@ -456,19 +458,55 @@ def _is_layer_fqn(s: str) -> bool:
...
+                model=self._model, optim_dict=optim_dict
             )
+            # Load optimizer states for each param.
+            if opt_state_dict is not None:
+                for param in opt_state_dict.keys():
+                    torch.distributed.breakpoint()
+                    if param not in self._optim_ckpt_wrapper.state_dict():
+                        raise RuntimeError(
+                            f"Trying to load optimizer state for unexpected param {param}"
+                        )
```
ebsmothers, line 500: Let's try to be more explicit about the error being related to optimizer-in-backward here. You can use the [single-device version](https://github.com/pytorch/torchtune/blob/5de500147e57d4d35e296faba68b296c03a1a2a0/recipes/full_finetune_single_device.py#L418-L421) as a reference here

recipes/full_finetune_distributed.py
```diff
@@ -657,12 +704,18 @@ def train(self) -> None:
...
-                        grad_norm = torch.nn.utils.clip_grad_norm_(
-                            self._model.parameters(),
-                            max_norm=float(self._clip_grad_norm),
-                        )
-                    self._optimizer.step()
-                    self._optimizer.zero_grad(set_to_none=True)
+                        if self._optimizer_in_bwd:
+                            raise NotImplementedError(
+                                "Gradient clipping is not supported after optimizer-in-the-backward."
+                            )
```
ebsmothers, line 710: Is this actually due to optimizer in backward, or something else? I don't think we have such a check in the single-device recipe

recipes/full_finetune_distributed.py
```diff
@@ -722,6 +775,27 @@ def train(self) -> None:
 
         self._profiler.stop()
 
+    def get_lr_scheduler(self) -> float:
```
ebsmothers, line 778: If we are gonna have this as a utility I would take it out of this recipe since it's equally applicable to the single-device case. Also I wouldn't call it `get_lr_scheduler`, since we're really getting the current lr, not the scheduler itself.

tests/torchtune/training/test_distributed.py
```diff
@@ -523,3 +523,223 @@ def _broadcast_full_state_dict(self, full_sd):
             result.append(None)
         torch.distributed.broadcast_object_list(result, src=0)
         return result[0]
+
+    @gpu_test(gpu_count=2)
+    @pytest.mark.skipif(
+        version.parse(torch.__version__).base_version < "2.5.0",
+        reason="torch >= 2.5 required",
+    )
+    def test_optimizer_in_backward(self):
```
ebsmothers, line 532: I'm wondering why we create a whole new test in here rather than adding a distributed test case in the existing [TestOptimInBackward test case](https://github.com/pytorch/torchtune/blob/5de500147e57d4d35e296faba68b296c03a1a2a0/tests/torchtune/training/test_memory.py#L91) (or if we want to inherit from `FSDPTest`, a new distributed version of that class in the same file). Because testing optimizer-in-backward as part of a class that is otherwise meant to test our fully_shard + state dict save and load logic feels a bit strange to me.

recipes/full_finetune_distributed.py
mori360: We have the initial zero out at single device at:
https://github.com/pytorch/torchtune/blob/665ab3f0c5a68f024c4ce0c512138bf8aa062dff/recipes/full_finetune_single_device.py#L591-L593

recipes/full_finetune_distributed.py
mori360: Single_device get lr with the assumption that [all optimizers have the same LR.](https://github.com/pytorch/torchtune/blob/5de500147e57d4d35e296faba68b296c03a1a2a0/recipes/full_finetune_single_device.py#L659)

Shall we apply the logic here to single device, rename to `get_lr` and move to `/torchtune/utils`?
or just taking the same assumption as single device?

recipes/full_finetune_distributed.py
mori360: It's due to optimizer-in-backward.
optimizer-in-backward calls `.step()` and `.zero_grad()` during `loss.backward()`, thus `grad`s are None at the time after to call `torch.nn.utils.clip_grad_norm_`, could not have gradient clipping successfully.
Single_device has the same issue that if `optimizer_in_bwd=True` and `clip_grad_norm=True`

tests/torchtune/training/test_distributed.py
mori360: The test here wants to proof that models running with optimzier-in-backward could have the same performance as running without.
The state_dict saving and loading want to test optimzier-in-backward's wrapper which is a bit different to the traditional optim.
There are not covered in `TestOptimInBackward`.

ebsmothers: If we want to do such an end-to-end test, maybe we can add a recipe test case instead? See e.g. test_full_finetune_distributed.py [here](https://github.com/pytorch/torchtune/blob/50d3ef13e709665b126904dcb7bae8e9679411f4/tests/recipes/test_full_finetune_distributed.py#L55-L66): with this test case you can run the full end-to-end recipe on a small test model and set `optimizer_in_bwd=True` directly from the config

recipes/full_finetune_distributed.py
ebsmothers: Technically I think this was also previously the assumption in this recipe, right? Since we just log the LR from the first param group. So we should be able to maintain the same logic for both

PR #1739, opened 2024-10-02
Fix CLIP pos embedding interpolation to work on DTensors

ebsmothers:
Currently our state dict hooks to resize CLIP positional embeddings do not work with FSDP2. This is because some of the ops used by interpolate (e.g. `aten.upsample_bilinear2d.default`) are not natively defined on `DTensor`, and as can be seen [here](https://github.com/pytorch/torchtune/blob/fc0249d6c88e0988dcce2ce9bbaf26b828794cd6/torchtune/training/_distributed.py#L337-L346), we shard the state dict tensors into DTensors before calling `load_state_dict` (where the interpolation hook is triggered).

Until these ops are defined on DTensors, we need to do something like this. This PR adds logic to the CLIP positional embedding state dict hooks to handle the case that tensors in the state dict are already sharded. For each relevant tensor, we:

1) Check if the state dict tensor is a DTensor
2) If it is, we gather via `.full_tensor()`. If it isn't, we don't do anything.
3) Call interpolate (along with various reshapes) on the vanilla Tensor from (2)
4) If the original state dict tensor was a DTensor, we call `distribute_tensor` using the original mesh and placements. Otherwise we don't do anything.

Test plan:

```
pytest tests/torchtune/models/clip/test_pos_embedding_interpolation.py
...
========== 15 passed in 0.66s ============
```

Compare loss curves on recipe before and after fix on base model. First, update 11B_lora_single_device.yaml to match [this](https://gist.github.com/ebsmothers/b9ab22658e8cd7ed27d8d550b2c4cd5f)

Before fix (on main):

```
tune run lora_finetune_single_device --config llama3_2_vision/11B_lora_single_device max_steps_per_epoch=100 \
metric_logger=torchtune.training.metric_logging.WandBLogger metric_logger.project=test-1739 \
metric_logger.name=pre-fix
```

After fix (this PR)

```
tune run lora_finetune_single_device --config llama3_2_vision/11B_lora_single_device max_steps_per_epoch=100 \
metric_logger=torchtune.training.metric_logging.WandBLogger metric_logger.project=test-1739 \
metric_logger.name=post-fix
```

Screenshot 2024-10-02 at 10 15 15 AM

Test that distributed recipe now runs: change 11B_lora config to use base model: [here](https://gist.github.com/ebsmothers/f96bbc7ee80c0b195db4892db608c094). Then run:

```
tune run --nproc_per_node 2 lora_finetune_distributed --config llama3_2_vision/11B_lora
...
1|2|Loss: 1.3108230829238892:   0%|                                                                                                                         | 2/10359 [00:57<82:38:19, 28.72s/it]
```

### DIFF
--- a/torchtune/models/clip/_position_embeddings.py
+++ b/torchtune/models/clip/_position_embeddings.py
@@ -139,7 +139,7 @@ class TiledTokenPositionalEmbedding(nn.Module):
             prefix + "local_token_positional_embedding"
         )
         local_device = inpt_local_pos_embed.device
-        if isinstance(self.local_token_positional_embedding, DTensor):
+        if isinstance(inpt_local_pos_embed, DTensor):
             inpt_local_pos_embed = inpt_local_pos_embed.full_tensor()
 
         if inpt_local_pos_embed is not None:
@@ -189,7 +189,7 @@ class TiledTokenPositionalEmbedding(nn.Module):
             prefix + "global_token_positional_embedding"
         )
         global_device = inpt_global_pos_embed.device
-        if isinstance(self.global_token_positional_embedding, DTensor):
+        if isinstance(inpt_global_pos_embed, DTensor):
             inpt_global_pos_embed = inpt_global_pos_embed.full_tensor()
 
         if inpt_global_pos_embed is not None:
@@ -521,7 +521,7 @@ class TilePositionalEmbedding(nn.Module):
 
         embedding = state_dict.get(prefix + "embedding")
         device = embedding.device
-        if isinstance(self.embedding, DTensor):
+        if isinstance(embedding, DTensor):
             embedding = embedding.full_tensor()
         if embedding is not None:
 
--- a/torchtune/models/clip/_position_embeddings.py
+++ b/torchtune/models/clip/_position_embeddings.py
@@ -140,7 +140,10 @@ class TiledTokenPositionalEmbedding(nn.Module):
         )
         local_device = inpt_local_pos_embed.device
         if isinstance(inpt_local_pos_embed, DTensor):
+            local_embed_is_sharded = True
             inpt_local_pos_embed = inpt_local_pos_embed.full_tensor()
+        else:
+            local_embed_is_sharded = False
 
         if inpt_local_pos_embed is not None:
 
@@ -164,7 +167,7 @@ class TiledTokenPositionalEmbedding(nn.Module):
                 tgt_patch_grid_size=int(math.sqrt(tgt_n_tokens_per_tile - 1)),
             )
 
-            if isinstance(inpt_local_pos_embed, DTensor):
+            if local_embed_is_sharded:
                 inpt_local_pos_embed = distribute_tensor(
                     inpt_local_pos_embed,
                     device_mesh=self.local_token_positional_embedding.device_mesh,
@@ -190,7 +193,10 @@ class TiledTokenPositionalEmbedding(nn.Module):
         )
         global_device = inpt_global_pos_embed.device
         if isinstance(inpt_global_pos_embed, DTensor):
+            global_embed_is_sharded = True
             inpt_global_pos_embed = inpt_global_pos_embed.full_tensor()
+        else:
+            global_embed_is_sharded = False
 
         if inpt_global_pos_embed is not None:
 
@@ -218,7 +224,7 @@ class TiledTokenPositionalEmbedding(nn.Module):
                 tgt_patch_grid_size=int(math.sqrt(tgt_n_tokens_per_tile - 1)),
             )
 
-            if isinstance(inpt_global_pos_embed, DTensor):
+            if global_embed_is_sharded:
                 inpt_global_pos_embed = distribute_tensor(
                     inpt_global_pos_embed,
                     device_mesh=self.global_token_positional_embedding.device_mesh,
@@ -522,7 +528,11 @@ class TilePositionalEmbedding(nn.Module):
         embedding = state_dict.get(prefix + "embedding")
         device = embedding.device
         if isinstance(embedding, DTensor):
+            embedding_is_sharded = True
             embedding = embedding.full_tensor()
+        else:
+            embedding_is_sharded = False
+
         if embedding is not None:
 
             # ckpt pos emb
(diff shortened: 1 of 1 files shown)


torchtune/models/clip/_position_embeddings.py
```diff
@@ -137,6 +138,10 @@ def _load_state_dict_hook(
         inpt_local_pos_embed = state_dict.get(
             prefix + "local_token_positional_embedding"
         )
+        local_device = inpt_local_pos_embed.device
+        if isinstance(inpt_local_pos_embed, DTensor):
+            inpt_local_pos_embed = inpt_local_pos_embed.full_tensor()
```
felipemello1, line 143: delete "local_device" since its not used

put this inside of the "if inpt_local_pos_embed is not None:"

torchtune/models/clip/_position_embeddings.py
```diff
@@ -159,6 +164,13 @@ def _load_state_dict_hook(
                 tgt_patch_grid_size=int(math.sqrt(tgt_n_tokens_per_tile - 1)),
             )
 
+            if isinstance(inpt_local_pos_embed, DTensor):
```
felipemello1, line 167: inpt_local_pos_embed is not a DTensor, because its full now

need to chech if self.local_token_positional_embedding is DTensor. We should probably do the same in the previous check (there you check if inpt_local_pos_embed is DTensor)

torchtune/models/clip/_position_embeddings.py
```diff
@@ -137,6 +138,10 @@ def _load_state_dict_hook(
         inpt_local_pos_embed = state_dict.get(
             prefix + "local_token_positional_embedding"
         )
+        local_device = inpt_local_pos_embed.device
+        if isinstance(inpt_local_pos_embed, DTensor):
+            inpt_local_pos_embed = inpt_local_pos_embed.full_tensor()
+
         if inpt_local_pos_embed is not None:
```
felipemello1, line 145: we should probably add here: 

if inpt_local_pos_embed is not None and inpt_local_pos_embed.shape != self. local_token_positional_embedding

But testing becomes a bit trickier. Maybe for now its better to not add it until testing is completed

torchtune/models/clip/_position_embeddings.py
```diff
@@ -497,7 +520,9 @@ def _load_state_dict_hook(
         """
 
         embedding = state_dict.get(prefix + "embedding")
-
+        device = embedding.device
+        if isinstance(embedding, DTensor):
+            embedding = embedding.full_tensor()
```
felipemello1, line 525: same comments as above

felipemello1 (commented):
to test it, please either use the pretrained version, since it has a different shape for local and global. The other pos embedding is the same, or change the max_num_tiles and patch size in the configs.

torchtune/models/clip/_position_embeddings.py
ebsmothers: Yeah will leave out for now

felipemello1: i think we should add before the PR is finalized though. What do you think?

torchtune/models/clip/_position_embeddings.py
```diff
@@ -137,8 +138,20 @@ def _load_state_dict_hook(
...
         if inpt_local_pos_embed is not None:
 
+            # We can only apply F.interpolate to vanilla tensors, not DTensors
+            # If pos embeds are a DTensor, we gather the full tensor, apply
+            # interpolate, and then reshard after
+            if isinstance(inpt_local_pos_embed, DTensor):
+                local_embed_is_sharded = True
+                local_embed_device_mesh = inpt_local_pos_embed.device_mesh
+                local_embed_placements = inpt_local_pos_embed.placements
+                inpt_local_pos_embed = inpt_local_pos_embed.full_tensor()
```
felipemello1, line 151: i thought that we had to use the device_mech and placements from self.local_token_positional_embedding

ebsmothers: I think they should be the same, which is why it was working before. But in my mind this is the more "correct" thing to do. We apply some operation to a DTensor, then restore it to its original state after the fact

torchtune/models/clip/_position_embeddings.py
ebsmothers: Why do you say testing becomes trickier?

ebsmothers: After thinking about it more, I don't think we should add it. The DTensor fix resolves the issue and no need to add extra logic on top of this that was not present before

torchtune/models/clip/_position_embeddings.py
RdoubleA: will this increase chances of OOMs, or is the pos_embed small enough where this is not a concern?

RdoubleA (approved):

felipemello1:
thanks Evan! I am a bit confused: I am not sure how single device is a good test, since it doesnt require FSDP

I think that the best test is to try to run the pretrained Model on single device, save the interpolated embeddings (i sahred them with you), run it in distributed and confirm that the interpolated embeddings are the same produced in single device

torchtune/models/clip/_position_embeddings.py
ebsmothers: @RdoubleA well as of today it doesn't work so I guess the current chance of OOM is NaN. This is supposed to be a no-op for single device (hence wrapping everything in `isinstance(..., DTensor)` checks), so no memory implications there.

ebsmothers:
@felipemello1 the purpose of single device is to show that we haven't broken anything there. I can run the test you're describing as well

felipemello1:
the unit tests should cover the single device. I added this to the functions, where i saved with single device, and loaded with distributed. It passed. We are good to merge

```
# torch.save(inpt_local_pos_embed, "inpt_local_pos_embed.pt")
inpt_local_pos_embed_single = torch.load("inpt_local_pos_embed.pt", map_location=torch.device('cpu'))
inpt_local_pos_embed_single = inpt_local_pos_embed_single.to(inpt_local_pos_embed.device)
assert torch.allclose(inpt_local_pos_embed, inpt_local_pos_embed_single, rtol=1e-03, atol=1e-03)
print("passed local")

# torch.save(inpt_global_pos_embed, "inpt_global_pos_embed.pt")
inpt_global_pos_embed_single = torch.load("inpt_global_pos_embed.pt", map_location=torch.device('cpu'))
inpt_global_pos_embed_single = inpt_global_pos_embed_single.to(inpt_global_pos_embed.device)
assert torch.allclose(inpt_global_pos_embed, inpt_global_pos_embed_single, rtol=1e-03, atol=1e-03)
print("passed global")
```

PR #1743, opened 2024-10-02
Move urllib pin to Github workflow

joecummings:
Pin urllib for recipe test in the Github workflow script instead of in dev dependencies. 

**Why?**: Follows the pattern we use for lm-eval. If we want to pin in dev dependencies, then we should move the eleuther download to the dev dependencies, as well.


ebsmothers:
Discussed offline, let's leave in dev dependencies for now

PR #1745, opened 2024-10-02
Assert LLama Vision Image size divides by 14

pbontrager:
Image size must be divisible by ViT patches in the CLIP encoder, 14.

### DIFF
--- a/torchtune/models/llama3_2_vision/_model_builders.py
+++ b/torchtune/models/llama3_2_vision/_model_builders.py
@@ -41,6 +41,7 @@ def llama3_2_vision_11b(
     Returns:
         DeepFusionModel: Instantiation of the Llama 3.2 Vision 11B model
     """
+    assert image_size % 14 == 0, "image_size must be divisible by 14, the encode patch size"
     encoder = llama3_2_vision_encoder(
         patch_size=14,
         num_heads=16,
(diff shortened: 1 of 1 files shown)


joecummings:
Add a test.

torchtune/models/llama3_2_vision/_model_builders.py
```diff
@@ -159,6 +160,7 @@ def lora_llama3_2_vision_11b(
         a subset of the attention projections in each layer.
 
     """
+    assert image_size % 14 == 0, "image_size must be divisible by 14, the encode patch size"
```
RdoubleA, line 163: Can you make this a ValueError instead?

torchtune/models/llama3_2_vision/_model_builders.py
```diff
@@ -41,6 +41,7 @@ def llama3_2_vision_11b(
     Returns:
         DeepFusionModel: Instantiation of the Llama 3.2 Vision 11B model
     """
+    assert image_size % 14 == 0, "image_size must be divisible by 14, the encode patch size"
```
felipemello1, line 44: did you mean encoder or encode?

Also, should it be ValueError, or assert? I think that normally we prefer to raise ValueError

felipemello1: just saw Rafi's comment :P

felipemello1 (approved):

RdoubleA (changes requested):
just need to address comments and it's good to go

PR #1756, opened 2024-10-06
Create CITATION.cff

joecummings:
👀

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -1,0 +1,9 @@
+cff-version: 1.2.0
+title: "torchtune: the PyTorch native finetuning library"
+message: "If you use this software, please cite it as below."
+type: software
+authors:
+  - given-names: "torchtune maintainers and contributors"
+url: "https//github.com/pytorch/torchtune"
+license: "BSD-3-Clause"
+date-released: "2024-04-14"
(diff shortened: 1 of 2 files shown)


CITATION.cff
```diff
@@ -0,0 +1,9 @@
+cff-version: 1.2.0
+title: "torchtune: the PyTorch native finetuning library"
```
SalmanMohammadi, line 2:
```suggestion
title: "torchtune: a PyTorch native finetuning library"
```

joecummings: You don't like being THE finetuning library?? 🥹

joecummings:
> Should we also add a section in the README?

added

SalmanMohammadi:
"a finetuning library" bestows us with a modest sagaciousness

Think we just need to link to the section in the readme heading (https://github.com/pytorch/torchtune/blob/a8a64ec6a99a6ea2be4fdaf0cd5797b03a2567cf/README.md?plain=1#L10)

SalmanMohammadi (approved):

CITATION.cff
ebsmothers: Or we can split the difference and say "PyTorch's finetuning library"

PR #1759, opened 2024-10-07
Update run.py to use runpy.run_path instead of runpy.run_module

psoulos:
In order to run a custom recipe from a local file, I needed to change this line.

#### Context
- [ ] add a new feature
- [X] fix a bug
- [ ] other (please add here)

#### Changelog
*

#### Test plan

#### UX

### DIFF
--- a/torchtune/_cli/run.py
+++ b/torchtune/_cli/run.py
@@ -102,7 +102,7 @@ For a list of all possible recipes, run `tune ls`."""
             runpy.run_path(str(args.recipe), run_name="__main__")
         else:
             # custom recipes are specified as a relative module dot path
-            runpy.run_module(str(args.recipe), run_name="__main__")
+            runpy.run_path(str(args.recipe), run_name="__main__")
 
     def _is_distributed_args(self, args: argparse.Namespace):
         """Check if the user is trying to run a distributed recipe."""


joecummings (commented):
I think this is correct, but @RdoubleA can you take a look?

joecummings:
@psoulos Thanks for the PR! Would you mind signing the CLA so we can kick off our CI?

RdoubleA:
Hi @psoulos, I've actually just fixed this so please install the latest nightlies (available by tomorrow) and try with the standard file path with slashes again, let me know if there's still issues.

psoulos:
@RdoubleA I did not install the latest nightlies in order to not mess with my environment, but I did manually apply the change from #1760. I can confirm that using the dotpath with the current release version or the full path with the changes from #1760 works as expected. The only issue I ran into is that I placed my_custom_recipe.py in recipes/ which throws a ModuleNotFoundError because of recipes/__init__.py. It might be worth documenting that custom recipes should not be placed in recipes/

RdoubleA:
> It might be worth documenting that custom recipes should not be placed in recipes/

Absolutely, this is on my radar. Thanks for testing and verifying both scenarios, nothing else needed from you. We can close this PR. See you around~

PR #1760, opened 2024-10-07
Custom recipes use slash path

RdoubleA:
#### Context
- [ ] add a new feature
- [x] fix a bug
- [ ] other (please add here)

Custom imports/recipes were fixed by running as modules, but this required users to specify custom recipes as dotpaths, i.e., `tune run path.to.my.recipe` which isn't consistent (as discussed in #1759). We do the conversion under the hood for them here so they can still specify standard file paths with slashes.

#### Test plan
both of these now run, and they rely on custom dataset builders located outside of torchtune repo:
`tune run --nproc_per_node 8 recipe/sft --config config/sft_llama3_2_1B.yaml`
`tune run --nproc_per_node 8 recipe/sft.py --config config/sft_llama3_2_1B.yaml`

### DIFF
--- a/torchtune/_cli/run.py
+++ b/torchtune/_cli/run.py
@@ -123,6 +123,18 @@ For a list of all possible recipes, run `tune ls`."""
             if recipe.name == recipe_str:
                 return recipe
 
+    def _convert_to_dotpath(self, recipe_path: str) -> str:
+        """Convert a custom recipe path to a dot path that can be run as a module.
+
+        Args:
+            recipe_path (str): The path of the recipe.
+
+        Returns:
+            The dot path of the recipe.
+        """
+        filepath, _ = os.path.splitext(recipe_path)
+        return filepath.replace("/", ".")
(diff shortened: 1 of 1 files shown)


acisseJZhong (approved):
thanks Rafi for the fix. Does this also work with customized configs inside torchtune repo?

RdoubleA:
> Does this also work with customized configs inside torchtune repo?

can't run any custom recipes from instead the recipes directory because of `ModuleNotFoundError: The torchtune recipes directory isn't a package and you should not import anything from here.`, but I personally think this is ok

joecummings:
> > Does this also work with customized configs inside torchtune repo?
> 
> can't run any custom recipes from instead the recipes directory because of `ModuleNotFoundError: The torchtune recipes directory isn't a package and you should not import anything from here.`, but I personally think this is ok

Wait, so will it still work if the user adds their recipe to the registry?

PR #1761, opened 2024-10-07
Set drop_last to always True

### LINKED ISSUE #1754: drop_last does not work with `ConcatDataset`
`drop_last` is expected to be specified with the dataset config object, but this errors out when you use multiple datasets and cfg_dataset becomes a ListConfig: https://github.com/pytorch/torchtune/blob/main/recipes/full_finetune_distributed.py#L511

Possible fixes:
- Always use `drop_last=True`
- Explicitly check for ListConfig, if so just use True

RdoubleA:
#### Context
- [ ] add a new feature
- [x] fix a bug
- [ ] other (please add here)

Closes #1754. Instead of specifying drop_last with the dataset config (which will break when passed into the dataset builder), just hardcode to True. No configs expose this. We can make it configurable down the line if users request it.

#### Test plan
CI

### DIFF
--- a/recipes/full_finetune_distributed.py
+++ b/recipes/full_finetune_distributed.py
@@ -248,6 +248,8 @@ class FullFinetuneRecipeDistributed(FTRecipeInterface):
             shuffle=cfg.shuffle,
             batch_size=cfg.batch_size,
             collate_fn=collate_name,
+            # dropping last avoids shape issues with compile + flex attention
+            drop_last=cfg.get("drop_last", True),
         )
 
         # Finally update the recipe state which can only be correctly set after all of the
@@ -476,6 +478,7 @@ class FullFinetuneRecipeDistributed(FTRecipeInterface):
         shuffle: bool,
         batch_size: int,
         collate_fn: str,
+        drop_last: bool,
     ) -> Tuple[DistributedSampler, DataLoader]:
         """
         All data related setup happens here. Currently this recipe only supports the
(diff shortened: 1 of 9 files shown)


acisseJZhong (approved):

ebsmothers (changes requested):
Is this the right fix? Personally I find a top-level config field called `drop_last` to be pretty unclear. Why not just infer `drop_last` in the existing if/else for `ConcatDataset` here? https://github.com/pytorch/torchtune/blob/main/recipes/full_finetune_distributed.py#L487-L496

E.g. 

```
  # same stuff as before
  drop_last = any([single_cfg_dataset.get("drop_last", True) for single_cfg_dataset in cfg_dataset])
else:
  # same stuff as before
  drop_last = cfg_dataset.get("drop_last", True)
```

RdoubleA:
drop_last being in the dataset config does not make sense, as dataset points to a component and all heaths should be related to that component/builder. drop_last is not in any of our dataset builders. I think having it top level is less confusing in that sense.

Frankly, we should have a more configurable dataloader or a section in the config for dataloader args, as this is becoming more customized.

ebsmothers:
OK yeah it is a dataloader config more than a dataset config so I do see your point there. Mainly I just do not want us to wind up with a hodgepodge of a bunch of random top-level configs, and `drop_last` is too in the weeds for the majority of users to care about (it is pretty much the exact definition of an edge case).

joecummings (commented):
Rather than trying to conceive of edge cases that users might not even care about, can we just default drop_last to True, document it, and then *if* users want to have this be configurable, we can add a config. 

Either way it's a one line change. For most cases, the last batch should be dropped.

RdoubleA:
Agreed

ebsmothers (approved):

Issue #1675, opened 2024-09-25
[RFC] Supporting KV-cache toggling


SalmanMohammadi:
### Problem

Currently, when we use `model.setup_caches()`, KV-caches are always updated for every subsequent forward pass on the model. We have valid use cases for using `model.setup_caches()`, but then not updating the KV-cache on forward passes. The most immediate use-case we have for this is in the eval recipe (see https://github.com/pytorch/torchtune/issues/1621). When specifying multiple tasks in the recipe:
- If one task is a generation task and another a log-likelihood task **and**
- The generation task is evaluated before the log-likelihood task **and**
- KV-cacheing is enabled

Then, KV-cacheing will still be enabled for the log-likelihood task which is incorrect behaviour for a number of reasons. In this example, we have two forward passes occurring, one with KV-cacheing enabled, one with KV-cacheing disabled.

Another use case is for my work on improving our RLHF offerings.
1) For the current PPO recipe, the overall structure looks something like:
```python
batch = batch.to(device)
with torch.no_grad():
    ...
    completions = policy_model.generate(...)

loss = self._ppo_step(completions, policy_model, ...)
loss.backward()
optimizer.step()
```
This isn't 1:1 but you get the gist. Here, we have two kinds of forward passes: one under `torch.no_grad()` with KV-cacheing, and one with `requires_grad=True` and without KV-cacheing (infact, we should never really have the case where `requires_grad=True` + KV-cacheing, but this part is kind of relevant for compile).

2) For a LoRA PPO recipe I've been working on, the structure is a bit different:

```python
batch = batch.to(device)
with torch.no_grad():
    ...
    completions = policy_model.generate(...)
    with torchtune.modules.peft.disable_adapter(policy_model):
        ref_logits = policy_model(completions)

loss = self._ppo_step(completions, policy_model, ...)
loss.backward()
optimizer.step()
```

So, now three kinds of forward passes: two under `torch.no_grad()`: one with KV-cacheing and one without, and one with `requires_grad=True` and without KV-cacheing.

We want to support these use-cases in a compile friendly manner.

#### 1) Context managing KV-cacheing

To me this feels like the most user-friendly solution. We define a context manager like:
```python
@contextlib.contextmanager
def enable_kv_cache(model):
    if not model.caches_are_enabled():
        raise ValueError()
    for layer in model.layers:
        layer.attn.cache_enabled = True
    try:
        yield
    finally:
        for layer in model.layers:
            layer.attn.cache_enabled = False
[...]

PR #1763, opened 2024-10-07
Toggling KV-caches

### LINKED ISSUE #1621: Fix eval recipe for consecutive generation and non-generation tasks
Currently, if we specify multiple tasks for the eval recipe, and one of the tasks is a generation task which uses KV-cacheing, then the cache is still enabled for non-generation tasks. The recipe will then error out here.

We should try having some check when using cacheing in addition to `caches_are_enabled`, which also checks that we're in e.g. inference mode.

EDIT: This is generally an issue for our models and will also become apparent for LoRA PPO.

SalmanMohammadi:
#### Context
- [x] add a new feature
- [x] fix a bug
- [ ] other (please add here)

closes #1621 
RFC here #1675 

---
# Multimodal eval results

## On main

```bash
root@736fb59b1bb9:~/torchtune#  tune run eleuther_eval --config llama3_2_vision/evaluation limit=5 max_seq_length=2048
Running EleutherEvalRecipe with resolved config:

batch_size: 1
checkpointer:
  _component_: torchtune.training.FullModelMetaCheckpointer
  checkpoint_dir: /tmp/Llama-3.2-11B-Vision-Instruct/original
  checkpoint_files:
  - consolidated.pth
  model_type: LLAMA3_VISION
  output_dir: ./
device: cuda
dtype: bf16
enable_kv_cache: true
limit: 5
log_level: INFO
max_seq_length: 2048
model:
  _component_: torchtune.models.llama3_2_vision.llama3_2_vision_11b
quantizer: null
seed: 1234
tasks:
- mmmu_val_science
tokenizer:
  _component_: torchtune.models.llama3_2_vision.llama3_2_vision_transform
  max_seq_len: 8192
  path: /tmp/Llama-3.2-11B-Vision-Instruct/original/tokenizer.model

2024-10-09:15:51:36,147 INFO     [_logging.py:101] Running EleutherEvalRecipe with resolved config:

batch_size: 1
checkpointer:
  _component_: torchtune.training.FullModelMetaCheckpointer
  checkpoint_dir: /tmp/Llama-3.2-11B-Vision-Instruct/original
  checkpoint_files:
  - consolidated.pth
  model_type: LLAMA3_VISION
  output_dir: ./
device: cuda
dtype: bf16
enable_kv_cache: true
limit: 5
log_level: INFO
max_seq_length: 2048
model:
  _component_: torchtune.models.llama3_2_vision.llama3_2_vision_11b
quantizer: null
seed: 1234
tasks:
- mmmu_val_science
tokenizer:
  _component_: torchtune.models.llama3_2_vision.llama3_2_vision_transform
  max_seq_len: 8192
  path: /tmp/Llama-3.2-11B-Vision-Instruct/original/tokenizer.model

Model is initialized with precision torch.bfloat16.
2024-10-09:15:51:39,457 INFO     [eleuther_eval.py:500] Model is initialized with precision torch.bfloat16.
Running evaluation on the following tasks: ['mmmu_val_science']
2024-10-09:15:51:51,298 INFO     [eleuther_eval.py:549] Running evaluation on the following tasks: ['mmmu_val_science']
2024-10-09:15:51:51,302 INFO     [task.py:415] Building contexts for mmmu_val_biology on rank 0...
100%|█████████████████████████████████████████████████████████████████████████████████████████████████| 5/5 [00:00<00:00, 13001.56it/s]
2024-10-09:15:51:51,338 INFO     [task.py:415] Building contexts for mmmu_val_chemistry on rank 0...
100%|█████████████████████████████████████████████████████████████████████████████████████████████████| 5/5 [00:00<00:00, 20846.44it/s]
2024-10-09:15:51:51,342 INFO     [task.py:415] Building contexts for mmmu_val_geography on rank 0...
100%|█████████████████████████████████████████████████████████████████████████████████████████████████| 5/5 [00:00<00:00, 20846.44it/s]
2024-10-09:15:51:51,359 INFO     [task.py:415] Building contexts for mmmu_val_math on rank 0...
100%|█████████████████████████████████████████████████████████████████████████████████████████████████| 5/5 [00:00<00:00, 14246.96it/s]
2024-10-09:15:51:51,377 INFO     [task.py:415] Building contexts for mmmu_val_physics on rank 0...
100%|█████████████████████████████████████████████████████████████████████████████████████████████████| 5/5 [00:00<00:00, 21822.60it/s]
2024-10-09:15:51:51,392 INFO     [evaluator.py:489] Running generate_until requests
Running generate_until requests with text+image input: 100%|███████████████████████████████████████████| 25/25 [04:34<00:00, 10.98s/it]
Eval completed in 274.73 seconds.
2024-10-09:15:56:26,028 INFO     [eleuther_eval.py:558] Eval completed in 274.73 seconds.
Max memory allocated: 32.86 GB
2024-10-09:15:56:26,028 INFO     [eleuther_eval.py:559] Max memory allocated: 32.86 GB


|   Tasks    |Version|Filter|n-shot|Metric|   |Value|   |Stderr|
|------------|------:|------|------|------|---|----:|---|-----:|
|Science     |      0|none  |      |acc   |↑  | 0.32|±  |0.0938|
| - Biology  |      0|none  |None  |acc   |↑  | 0.20|±  |0.2000|
| - Chemistry|      0|none  |None  |acc   |↑  | 0.00|±  |0.0000|
| - Geography|      0|none  |None  |acc   |↑  | 0.40|±  |0.2449|
| - Math     |      0|none  |None  |acc   |↑  | 0.40|±  |0.2449|
| - Physics  |      0|none  |None  |acc   |↑  | 0.60|±  |0.2449|


2024-10-09:15:56:26,086 INFO     [eleuther_eval.py:563] 

|   Tasks    |Version|Filter|n-shot|Metric|   |Value|   |Stderr|
|------------|------:|------|------|------|---|----:|---|-----:|
|Science     |      0|none  |      |acc   |↑  | 0.32|±  |0.0938|
| - Biology  |      0|none  |None  |acc   |↑  | 0.20|±  |0.2000|
| - Chemistry|      0|none  |None  |acc   |↑  | 0.00|±  |0.0000|
| - Geography|      0|none  |None  |acc   |↑  | 0.40|±  |0.2449|
| - Math     |      0|none  |None  |acc   |↑  | 0.40|±  |0.2449|
| - Physics  |      0|none  |None  |acc   |↑  | 0.60|±  |0.2449|


```
---

## On this branch

```bash
root@736fb59b1bb9:~/torchtune#  tune run eleuther_eval --config llama3_2_vision/evaluation limit=5 max_seq_length=2048
Running EleutherEvalRecipe with resolved config:

batch_size: 1
checkpointer:
  _component_: torchtune.training.FullModelMetaCheckpointer
  checkpoint_dir: /tmp/Llama-3.2-11B-Vision-Instruct/original
  checkpoint_files:
  - consolidated.pth
  model_type: LLAMA3_VISION
  output_dir: ./
device: cuda
dtype: bf16
enable_kv_cache: true
limit: 5
log_level: INFO
max_seq_length: 2048
model:
  _component_: torchtune.models.llama3_2_vision.llama3_2_vision_11b
quantizer: null
seed: 1234
tasks:
- mmmu_val_science
tokenizer:
  _component_: torchtune.models.llama3_2_vision.llama3_2_vision_transform
  max_seq_len: 8192
  path: /tmp/Llama-3.2-11B-Vision-Instruct/original/tokenizer.model

2024-10-09:15:45:53,771 INFO     [_logging.py:101] Running EleutherEvalRecipe with resolved config:

batch_size: 1
checkpointer:
  _component_: torchtune.training.FullModelMetaCheckpointer
  checkpoint_dir: /tmp/Llama-3.2-11B-Vision-Instruct/original
  checkpoint_files:
  - consolidated.pth
  model_type: LLAMA3_VISION
  output_dir: ./
device: cuda
dtype: bf16
enable_kv_cache: true
limit: 5
log_level: INFO
max_seq_length: 2048
model:
  _component_: torchtune.models.llama3_2_vision.llama3_2_vision_11b
quantizer: null
seed: 1234
tasks:
- mmmu_val_science
tokenizer:
  _component_: torchtune.models.llama3_2_vision.llama3_2_vision_transform
  max_seq_len: 8192
  path: /tmp/Llama-3.2-11B-Vision-Instruct/original/tokenizer.model

Model is initialized with precision torch.bfloat16.
2024-10-09:15:45:57,532 INFO     [eleuther_eval.py:500] Model is initialized with precision torch.bfloat16.
Running evaluation on the following tasks: ['mmmu_val_science']
2024-10-09:15:46:10,590 INFO     [eleuther_eval.py:549] Running evaluation on the following tasks: ['mmmu_val_science']
2024-10-09:15:46:10,594 INFO     [task.py:415] Building contexts for mmmu_val_biology on rank 0...
100%|█████████████████████████████████████████████████████████████████████████████████████████████████| 5/5 [00:00<00:00, 12336.19it/s]
2024-10-09:15:46:10,631 INFO     [task.py:415] Building contexts for mmmu_val_chemistry on rank 0...
100%|█████████████████████████████████████████████████████████████████████████████████████████████████| 5/5 [00:00<00:00, 20301.57it/s]
2024-10-09:15:46:10,635 INFO     [task.py:415] Building contexts for mmmu_val_geography on rank 0...
100%|█████████████████████████████████████████████████████████████████████████████████████████████████| 5/5 [00:00<00:00, 15911.62it/s]
2024-10-09:15:46:10,653 INFO     [task.py:415] Building contexts for mmmu_val_math on rank 0...
100%|█████████████████████████████████████████████████████████████████████████████████████████████████| 5/5 [00:00<00:00, 13626.72it/s]
2024-10-09:15:46:10,671 INFO     [task.py:415] Building contexts for mmmu_val_physics on rank 0...
100%|█████████████████████████████████████████████████████████████████████████████████████████████████| 5/5 [00:00<00:00, 20223.26it/s]
2024-10-09:15:46:10,687 INFO     [evaluator.py:489] Running generate_until requests
Running generate_until requests with text+image input: 100%|███████████████████████████████████████████| 25/25 [04:33<00:00, 10.92s/it]
Eval completed in 273.26 seconds.
2024-10-09:15:50:43,852 INFO     [eleuther_eval.py:558] Eval completed in 273.26 seconds.
Max memory allocated: 32.86 GB
2024-10-09:15:50:43,852 INFO     [eleuther_eval.py:559] Max memory allocated: 32.86 GB


|   Tasks    |Version|Filter|n-shot|Metric|   |Value|   |Stderr|
|------------|------:|------|------|------|---|----:|---|-----:|
|Science     |      0|none  |      |acc   |↑  | 0.32|±  |0.0938|
| - Biology  |      0|none  |None  |acc   |↑  | 0.20|±  |0.2000|
| - Chemistry|      0|none  |None  |acc   |↑  | 0.00|±  |0.0000|
| - Geography|      0|none  |None  |acc   |↑  | 0.40|±  |0.2449|
| - Math     |      0|none  |None  |acc   |↑  | 0.40|±  |0.2449|
| - Physics  |      0|none  |None  |acc   |↑  | 0.60|±  |0.2449|


2024-10-09:15:50:43,909 INFO     [eleuther_eval.py:563] 

|   Tasks    |Version|Filter|n-shot|Metric|   |Value|   |Stderr|
|------------|------:|------|------|------|---|----:|---|-----:|
|Science     |      0|none  |      |acc   |↑  | 0.32|±  |0.0938|
| - Biology  |      0|none  |None  |acc   |↑  | 0.20|±  |0.2000|
| - Chemistry|      0|none  |None  |acc   |↑  | 0.00|±  |0.0000|
| - Geography|      0|none  |None  |acc   |↑  | 0.40|±  |0.2449|
| - Math     |      0|none  |None  |acc   |↑  | 0.40|±  |0.2449|
| - Physics  |      0|none  |None  |acc   |↑  | 0.60|±  |0.2449|

```

# Text eval results

```bash
(tune) salman@combuter:~/torchtune$ tune run eleuther_eval --config target/eleuther_evaluation.yaml 
2024-10-08:21:18:07,202 INFO     [_logging.py:101] Running EleutherEvalRecipe with resolved config:

batch_size: 1
checkpointer:
  checkpoint_dir: ./target/1b_normal
  checkpoint_files:
  - pytorch_model.bin
  model_type: LLAMA2
  output_dir: ./target/tmp
device: cuda
dtype: bf16
enable_kv_cache: true
limit: 20
max_seq_length: 1024
model:
  _component_: torchtune.models.llama2.llama2
  embed_dim: 2048
  max_seq_len: 4096
  norm_eps: 1.0e-05
  num_heads: 32
  num_kv_heads: 4
  num_layers: 22
  vocab_size: 32000
quantizer: null
seed: 1234
tasks:
- truthfulqa_gen
- truthfulqa_mc2
tokenizer:
  path: ./target/1b_normal/tokenizer.model

2024-10-08:21:18:08,854 INFO     [eleuther_eval.py:495] Model is initialized with precision torch.bfloat16.
2024-10-08:21:18:08,879 INFO     [huggingface.py:132] Using device 'cuda:0'
/home/salman/.pyenv/versions/3.11.9/envs/tune/lib/python3.11/site-packages/transformers/tokenization_utils_base.py:1601: FutureWarning: `clean_up_tokenization_spaces` was not set. It will be set to `True` by default. This behavior will be depracted in transformers v4.45, and will be then set to `False` by default. For more details check this issue: https://github.com/huggingface/transformers/issues/31884
  warnings.warn(
2024-10-08:21:18:09,228 INFO     [huggingface.py:368] Model parallel was set to False, max memory was not set, and device map was set to {'': 'cuda:0'}
2024-10-08:21:18:10,532 INFO     [__init__.py:491] `group` and `group_alias` keys in TaskConfigs are deprecated and will be removed in v0.4.5 of lm_eval. The new `tag` field will be used to allow for a shortcut to a group of tasks one does not wish to aggregate metrics across. `group`s which aggregate across subtasks must be only defined in a separate group config file, which will be the official way to create groups that support cross-task aggregation as in `mmlu`. Please see the v0.4.4 patch notes and our documentation: https://github.com/EleutherAI/lm-evaluation-harness/blob/main/docs/new_task_guide.md#advanced-group-configs for more information.
2024-10-08:21:18:23,103 INFO     [eleuther_eval.py:537] Running evaluation on the following tasks: ['truthfulqa_gen', 'truthfulqa_mc2']
2024-10-08:21:18:23,106 INFO     [task.py:428] Building contexts for truthfulqa_mc2 on rank 0...
100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 20/20 [00:00<00:00, 903.93it/s]
2024-10-08:21:18:23,130 INFO     [task.py:428] Building contexts for truthfulqa_gen on rank 0...
100%|████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 20/20 [00:00<00:00, 1604.19it/s]
2024-10-08:21:18:23,147 INFO     [evaluator.py:485] Running loglikelihood requests
Running loglikelihood requests: 100%|████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 153/153 [00:18<00:00,  8.37it/s]
2024-10-08:21:18:41,510 INFO     [evaluator.py:485] Running generate_until requests
Running generate_until requests: 100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 20/20 [02:48<00:00,  8.43s/it]
2024-10-08:21:21:30,179 INFO     [rouge_scorer.py:83] Using default tokenizer.
2024-10-08:21:21:49,314 INFO     [eleuther_eval.py:546] Eval completed in 206.21 seconds.
2024-10-08:21:21:49,314 INFO     [eleuther_eval.py:547] Max memory allocated: 3.41 GB
2024-10-08:21:21:49,432 INFO     [eleuther_eval.py:551] 

|    Tasks     |Version|Filter|n-shot|  Metric   |   | Value  |   |Stderr|
|--------------|------:|------|-----:|-----------|---|-------:|---|-----:|
|truthfulqa_gen|      3|none  |     0|bleu_acc   |↑  |  0.3000|±  |0.1051|
|              |       |none  |     0|bleu_diff  |↑  |-11.1260|±  |3.5818|
|              |       |none  |     0|bleu_max   |↑  | 23.0317|±  |4.5452|
|              |       |none  |     0|rouge1_acc |↑  |  0.4500|±  |0.1141|
|              |       |none  |     0|rouge1_diff|↑  | -6.7262|±  |3.7765|
|              |       |none  |     0|rouge1_max |↑  | 49.9555|±  |4.9504|
|              |       |none  |     0|rouge2_acc |↑  |  0.3000|±  |0.1051|
|              |       |none  |     0|rouge2_diff|↑  |-12.0537|±  |4.2492|
|              |       |none  |     0|rouge2_max |↑  | 31.7591|±  |6.2387|
|              |       |none  |     0|rougeL_acc |↑  |  0.3500|±  |0.1094|
|              |       |none  |     0|rougeL_diff|↑  | -8.0059|±  |3.4676|
|              |       |none  |     0|rougeL_max |↑  | 47.6440|±  |5.3441|
|truthfulqa_mc2|      2|none  |     0|acc        |↑  |  0.4769|±  |0.0947|
```
#### Test plan

#### UX


torchtune/generation/_generation.py
```diff
@@ -366,7 +366,7 @@ def generate(
         tokens, logits = custom_generate_next_token(
             model,
             input_pos=curr_input_pos,
-            x=tokens,
+            x=tokens.clone(),
```
SalmanMohammadi, line 369: cudagraphs will otherwise complain about tensors in a previous graph being written to

SalmanMohammadi:
I've actually just realized this isn't going to play nice with fusion models and we may have to pollute `TransformerDecoder` a bit here. Bear with me.

recipes/eleuther_eval.py
```diff
@@ -276,28 +270,37 @@ def _model_multimodal_generate(
...
-            )[:, -1]
+        # 2. Setup KV cache
+        with setup_use_local_kv_cache(
+            self.model,
+            batch_size=self.batch_size,
+            device=self.device,
+            dtype=self._dtype,
+            encoder_max_seq_len=encoder_max_seq_len,
+            decoder_max_seq_len=self.max_length,
+        ):
```
SalmanMohammadi, line 281: It's a bit hefty looking, isn't it?

torchtune/modules/attention.py
```diff
@@ -164,6 +169,7 @@ def setup_cache(
                 head_dim=self.head_dim,
                 dtype=dtype,
             )
+            self.cache_enabled = True
```
SalmanMohammadi, line 172: So I'm setting to True here to preserve the original behaviour of `setup_caches()` -> all forwards passes use cache, and it means the new context generators are the recommended default we showcase in the recipes, but you can still do it the old way if you want.

This may be slightly unintuitive if you want to use the persistent context manager, i.e. setup caches somewhere early in your recipe, and then have some forward passes *before* you want to use caches.

If we set this to False by default it means the expected API for using KV-caches is through the context managers i.e. BC break

SalmanMohammadi: I actually think using context managers as the de-facto KV-cache API is sensible but open to thoughts here.

torchtune/modules/common_utils.py
```diff
@@ -163,3 +164,123 @@ def _register_reparametrize_state_dict_hooks(
...
+        yield
+    finally:
+        for _, module in model.named_modules():
+            if hasattr(module, "kv_cache") and callable(module.kv_cache):
+                module.kv_cache.cache_enabled = True
+        model.reset_caches()
+
+
+@contextlib.contextmanager
+def setup_use_local_kv_cache(
```
joecummings, line 211: I prefer something like:

```python
with local_kv_cache():
	...
```

This sticks with the notation already known for context managers and (I believe) tells the user that the kv cache will be torn-down after the context is exited.

torchtune/modules/common_utils.py
```diff
@@ -163,3 +164,123 @@ def _register_reparametrize_state_dict_hooks(
     module._register_state_dict_hook(
         partial(hook, dtype=dtype, offload_to_cpu=offload_to_cpu)
     )
+
+
+@contextlib.contextmanager
+def use_persistent_kv_cache(model: nn.Module) -> Generator[None, None, None]:
```
joecummings, line 170: Since the more common use-case is that the KV cache WILL be used after it's setup, I might consider a paradigm shift. 

Instead of `use_kv_cache`, something like `disable_kv_cache` that temporarily disables the kv_cache for use in PPO. 

This also means that our current APIs still hold true.

torchtune/modules/transformer.py
```diff
@@ -63,7 +63,7 @@ def setup_cache(
         self.attn.setup_cache(batch_size, dtype, max_seq_len=decoder_max_seq_len)
 
     @property
-    def cache_enabled(self) -> bool:
+    def cache_is_setup(self) -> bool:
```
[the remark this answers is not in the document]
SalmanMohammadi: Solely because the previous version was also singular

SalmanMohammadi:
> Yeah I think this touches on my main point of confusion with this PR: the distinction between caches_are_enabled and caches_are_setup. I thought I had an understanding of this but now I'm not so sure. It seems to me like some of the places we now call caches_are_setup (e.g. checking whether mask or input pos is passed in forward) should actually be calling caches_are_enabled.

I think I see where you're coming from here. I think what would help disambiguate is having `model.caches_are_enabled()`, but I tried my hardest not to add additional functions to `TransformerDecoder` and `DeepFusionModel`. In retrospect this was a silly goal and things would be significantly clearer if we had that method.

I think going forward, and to significantly simplify things, I'd propose:

1) The existing API around `setup_caches` is unchanged. You may `setup_caches`, then use KV-cacheing if you wish.
2) We add a `with temporary_kv_cache` or with `local_kv_cache` or `with kv_cache` which will enable KV-cacheing on a model that does not already have KV-cacheing enabled, and completely tear-down upon exit.
3) We add a `with disable_kv_cache` which disables KV-cacheing on a model that already has KV-cacheing enabled.
4) We add a `model.caches_are_enabled()`

This means 

1) We aren't making any breaking changes
2) We're cementing the distinction between `caches_are_enabled` and `caches_are_setup` s.t. most of the checks we make inside `TransformerDecoder` for input validation will use `caches_are_enabled`.

torchtune/modules/transformer.py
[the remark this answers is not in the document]
ebsmothers: Let's fix that while we're in here too then

ebsmothers:
> > Yeah I think this touches on my main point of confusion with this PR: the distinction between caches_are_enabled and caches_are_setup. I thought I had an understanding of this but now I'm not so sure. It seems to me like some of the places we now call caches_are_setup (e.g. checking whether mask or input pos is passed in forward) should actually be calling caches_are_enabled.
> 
> I think I see where you're coming from here. I think what would help disambiguate is having `model.caches_are_enabled()`, but I tried my hardest not to add additional functions to `TransformerDecoder` and `DeepFusionModel`. In retrospect this was a silly goal and things would be significantly clearer if we had that method.
> 
> I think going forward, and to significantly simplify things, I'd propose:
> 
> 1. The existing API around `setup_caches` is unchanged. You may `setup_caches`, then use KV-cacheing if you wish.
> 2. We add a `with temporary_kv_cache` or with `local_kv_cache` or `with kv_cache` which will enable KV-cacheing on a model that does not already have KV-cacheing enabled, and completely tear-down upon exit.
> 3. We add a `with disable_kv_cache` which disables KV-cacheing on a model that already has KV-cacheing enabled.
> 4. We add a `model.caches_are_enabled()`
> 
> This means
> 
> 1. We aren't making any breaking changes
> 2. We're cementing the distinction between `caches_are_enabled` and `caches_are_setup` s.t. most of the checks we make inside `TransformerDecoder` for input validation will use `caches_are_enabled`.

Overall this sounds pretty reasonable to me. My main question is around (3). If we just locally disable KV cache, when do we reset (if at all)? Like if I setup the cache, call forward a bunch of times (with it now enabled by default), wrap a few forward calls in `with disable_kv_cache`, what will happen on my next forward outside the `disable_kv_cache` context manager? Am I starting from scratch or resuming where I left off?

And then does this also mean that we now have both `caches_are_enabled` and `caches_are_setup` properties in `TransformerDecoder` and `DeepFusionModel`? Just to be explicit, can we clarify which classes we are defining each of these methods on (and why) across the board? I think that may help add some method to the madness here

SalmanMohammadi:
>  My main question is around (3). If we just locally disable KV cache, when do we reset (if at all)? Like if I setup the cache, call forward a bunch of times (with it now enabled by default), wrap a few forward calls in with disable_kv_cache, what will happen on my next forward outside the disable_kv_cache context manager? Am I starting from scratch or resuming where I left off?

The onus is on the user here to correctly `.reset_caches` whenever they're finished with a particular generation. I think we should be clear that the behaviour of `with disable_kv_cache` will mean your KV-caches are untouched and control of them will be returned to you upon exit.

> And then does this also mean that we now have both caches_are_enabled and caches_are_setup properties in TransformerDecoder and DeepFusionModel? Just to be explicit, can we clarify which classes we are defining each of these methods on (and why) across the board?

Sorry, yes. 

- We already have `caches_are_setup` on both `TransformerDecoder and `DeepFusion`. We use this to validate operations on the KV-caches which we can always perform (regardless of their enabled-ness) - `model.reset_caches`, `delete_caches(model)`, `model.setup_caches`.
- However, it seems sensible to also have a check for operations which rely the enabled-ness of the caches, i.e. `model.forward` will use `self.caches_are_enabled` to check if you're correctly passing inference-relevant args. I'm actually not sure if there's more cases I can think of where we need this method. It's just sugar for `self.layers[0].cache_enabled`. 

We will need to add `TransformerDecoder.caches_are_enabled` and `DeepFusion.caches_are_enabled`. This is the aformentioned pollution of the model.

torchtune/generation/_generation.py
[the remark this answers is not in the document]
SalmanMohammadi: This actually landed in another PR, I need to push my merge. cudagraphs was complaining about a tensor in a previous graph being overwritten and this ended up being the solution.

SalmanMohammadi:
As another side note here, we currently expect attention layers `TransformerSelfAttentionLayer` and `TransformerCrossAttentionLayer` to implement `self.setup_cache`, `self.cache_is_setup`, `self.reset_cache`, but all these functions do is call the same method on `self.attn`. 

This is the same across both kinds of attention layers.

We're already showing here that we don't need this by doing things like
```python
    for module in model.modules():
        if hasattr(module, "kv_cache") and callable(module.kv_cache):
            module.cache_enabled = False
            module.kv_cache = None
```

so why don't we get rid of the intermediary functions and use the API on `MultiHeadAttention` directly? 
@ebsmothers @joecummings

joecummings:
> As another side note here, we currently expect attention layers `TransformerSelfAttentionLayer` and `TransformerCrossAttentionLayer` to implement `self.setup_cache`, `self.cache_is_setup`, `self.reset_cache`, but all these functions do is call the same method on `self.attn`.
> 
> This is the same across both kinds of attention layers.
> 
> We're already showing here that we don't need this by doing things like
> 
> ```python
>     for module in model.modules():
>         if hasattr(module, "kv_cache") and callable(module.kv_cache):
>             module.cache_enabled = False
>             module.kv_cache = None
> ```
> 
> so why don't we get rid of the intermediary functions and use the API on `MultiHeadAttention` directly? @ebsmothers @joecummings @pbontrager
> 
> edit: sorry, I know I'm terrible for over-scoping PRs but I can't help it.

You're right, I think we should be doing this but BAD SALMAN STOP OVERSCOPING PRs. Let's open a separate issue for follow-up?

ebsmothers:
> > As another side note here, we currently expect attention layers `TransformerSelfAttentionLayer` and `TransformerCrossAttentionLayer` to implement `self.setup_cache`, `self.cache_is_setup`, `self.reset_cache`, but all these functions do is call the same method on `self.attn`.
> > This is the same across both kinds of attention layers.
> > We're already showing here that we don't need this by doing things like
> > ```python
> >     for module in model.modules():
> >         if hasattr(module, "kv_cache") and callable(module.kv_cache):
> >             module.cache_enabled = False
> >             module.kv_cache = None
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > so why don't we get rid of the intermediary functions and use the API on `MultiHeadAttention` directly? @ebsmothers @joecummings @pbontrager
> > edit: sorry, I know I'm terrible for over-scoping PRs but I can't help it.
> 
> You're right, I think we should be doing this but BAD SALMAN STOP OVERSCOPING PRs. Let's open a separate issue for follow-up?

Yeah agreed, let's tackle this separately. Sorry I slept on this PR for a few days, what's the status here? Like are we ready for another pass or are there still open questions we need to sort out?

SalmanMohammadi:
Yeah this is good to review again. I'll fix merge conflicts/add in the eval test you requested later this week.

torchtune/modules/attention.py
```diff
@@ -290,15 +296,15 @@ def forward(
...
+            if self.kv_cache is not None and self.cache_enabled:
                 k, v = self.kv_cache.update(k, v)
 
         output = self._attention_call(
             q,
             k,
             v,
             mask=mask,
-            dropout_p=self.attn_dropout if self.training else 0.0,
+            dropout_p=self.attn_dropout,
```
joecummings, line 307: Why is this changing here?

torchtune/modules/model_fusion/_fusion.py
```diff
@@ -107,24 +107,30 @@ def setup_cache(
...
             encoder_max_seq_len=encoder_max_seq_len,
             decoder_max_seq_len=decoder_max_seq_len,
         )
 
-    @property
-    def cache_enabled(self) -> bool:
-        """Check if the key value caches are setup."""
-        return self.layer.cache_enabled
+    def caches_are_setup(self) -> bool:
+        """Check if the key value caches have been setup."""
```
joecummings, line 125: Should we include a small definition of what "setup" means?

torchtune/modules/model_fusion/_fusion.py
```diff
@@ -107,24 +107,30 @@ def setup_cache(
...
+        return self.layer.caches_are_setup()
+
+    def caches_are_enabled(self) -> bool:
```
joecummings, line 128: Same comment here. Maybe a small definition of "enabled"

torchtune/modules/transformer.py
```diff
@@ -63,8 +63,11 @@ def setup_cache(
         """
         self.attn.setup_cache(batch_size, dtype, max_seq_len=decoder_max_seq_len)
 
-    @property
-    def cache_enabled(self) -> bool:
+    def caches_are_enabled(self) -> bool:
```
joecummings, line 66: Ditto comments from above.

torchtune/modules/transformer.py
```diff
@@ -423,27 +429,34 @@ def setup_caches(
...
             )
 
-    def caches_are_enabled(self) -> bool:
-        """Check if the key value caches are setup. This is useful to efficient inference."""
-        return self.layers[0].cache_enabled
+    def caches_are_setup(self) -> bool:
+        """Check if the key value caches have been setup."""
+        return self.layers[0].caches_are_setup()
 
     def reset_caches(self):
```
joecummings, line 443: Comment here that these will reset to zero, but keep caches enabled.

torchtune/modules/transformer.py
```diff
@@ -767,20 +774,20 @@ def setup_caches(
             )
 
     @property
-    def encoder_caches_are_enabled(self) -> bool:
+    def encoder_caches_are_setup(self) -> bool:
```
joecummings, line 777: cc @pbontrager, but I don't know if this would ever be useful.

joecummings (approved):
Couple nits, but I'm ready to sign off on this PR.

torchtune/modules/attention.py
SalmanMohammadi: Maybe need to merge?

torchtune/modules/transformer.py
pbontrager: It doesn't make sense to have an encoder cache. If we ever need caches in the encoder (which we never need at the moment), then the fusion model or fusion layers would be responsible for calling the encoder in their caches_are_setup property.

SalmanMohammadi: This function doesn't strictly check if the encoder has a cache setup, but rather any crossattention layers exist with caches setup, which is useful for input validation during inference with fusion models.

PR #1811, opened 2024-10-11
``TensorCoreTiledLayoutType`` -> ``TensorCoreTiledLayout``

joecummings:
#### Context
- [ ] add a new feature
- [x] fix a bug
- [ ] other (please add here)

TorchAO nightly from Oct 10 broke b/c they updated a variable name. See: https://github.com/pytorch/torchtune/actions/runs/11292390312/job/31408285709

#### Changelog
* Update variable name

#### Test plan

#### UX

### DIFF
--- a/torchtune/training/quantization.py
+++ b/torchtune/training/quantization.py
@@ -6,6 +6,8 @@
 
 from typing import Callable, Optional
 
+from torchtune.modules.low_precision._utils import _get_torchao_version
+
 if _get_torchao_version() >= (0, 6, 0):
     from torchao.dtypes import TensorCoreTiledLayout
 else:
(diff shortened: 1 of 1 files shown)


torchtune/training/quantization.py
```diff
@@ -6,7 +6,7 @@
 
 from typing import Callable, Optional
 
-from torchao.dtypes import TensorCoreTiledLayoutType
+from torchao.dtypes import TensorCoreTiledLayout
```
ebsmothers, line 9: We need a version check here now, no?

joecummings: Right you are

andrewor14:
cc @jainapurva

PR #1812, opened 2024-10-11
Torchao version check changes/BC import of TensorCoreTiledLayout

ebsmothers:
- Move torchao version check utilities out of modules and into utils (that's where they should've been all along)
- Update the nightly version check to match latest torchao nightlies format
  - We also no longer need all the extra importlib stuff now that all recent ao versions have `__version__` defined
- Define variable `_NEW_TENSOR_CORE_TILED_LAYOUT_API` based on the following conditions:
  - Not fbcode and (ao version >= 0.7.0 or (ao version is nightly and ao nightly date >= "2024-10-10"))
- Gate the import of `TensorCoreTiledLayoutType` in training/quantization.py and alias to `TensorCoreTiledLayout` (the new API name) in either case

## Test plan

Test quantization recipe on both stable and nightly torchao versions. Prereq: download Llama2 7B:

```
tune download meta-llama/Llama-2-7b-hf --output-dir /tmp/Llama-2-7b-hf
```

### Test on torchao 0.5

```
$ conda create -n ao-testing-stable python=3.11 -y 
$ conda activate ao-testing-stable
$ pip install torch torchvision torchao
$ pip install -e ".[dev]"
$ pip list | grep torchao
torchao                   0.5.0
$ tune run quantize --config quantization quantizer=torchtune.training.quantization.Int4WeightOnlyQuantizer quantizer.groupsize=128
...
INFO:torchtune.utils._logging:Model is initialized with precision torch.bfloat16.
INFO:torchtune.utils._logging:Time for quantization: 0.24 sec
INFO:torchtune.utils._logging:Memory used: 13.95 GB
INFO:torchtune.utils._logging:Model checkpoint of size 3.79 GB saved to /tmp/Llama-2-7b-hf/pytorch_model-00001-of-00002-4w.pt
```
### Test on torchao nightly

```
$ conda create -n ao-testing-nightly python=3.11 -y 
$ conda activate ao-testing-nightly
$ pip install --pre torch torchvision torchao --index-url https://download.pytorch.org/whl/nightly/cu124
$ pip install -e ".[dev]"
$ pip list | grep torchao
torchao                   0.7.0.dev20241011+cu124
$ tune run quantize --config quantization quantizer=torchtune.training.quantization.Int4WeightOnlyQuantizer quantizer.groupsize=128
...
INFO:torchtune.utils._logging:Model is initialized with precision torch.bfloat16.
INFO:torchtune.utils._logging:Time for quantization: 0.37 sec
INFO:torchtune.utils._logging:Memory used: 13.95 GB
INFO:torchtune.utils._logging:Model checkpoint of size 3.79 GB saved to /tmp/Llama-2-7b-hf/pytorch_model-00001-of-00002-4w.pt
```


torchtune/utils/_import_guard.py
```diff
@@ -5,11 +5,23 @@
...
 # We can only use flex attention / BlockMask if torch version >= 2.5.0 and GPU is Turing / SM75 and above
 _SUPPORTS_FLEX_ATTENTION = (
     torch_version_ge("2.5.0")
     and torch.cuda.is_available()
     and torch.cuda.get_device_capability() >= (7, 5)
 )
+
+torchao_version = _get_torchao_version()
+
+_NEW_TENSOR_CORE_TILED_LAYOUT_API = not _is_fbcode() and (
```
RdoubleA, line 24: mind adding a quick comment explaining this similar to flex attention above?

torchtune/utils/_version.py
```diff
@@ -23,3 +29,36 @@ def torch_version_ge(version: str) -> bool:
         True
     """
     return version in torch.__version__ or torch.__version__ >= version
+
+
+def _is_fbcode():
+    return not hasattr(torch.version, "git_version")
+
+
+def _nightly_version_ge(ao_version_str: str, date: str) -> bool:
```
RdoubleA, line 38: this doesn't generalize to pytorch nightly version? if's ao specific, let's include ao in the function name

torchtune/utils/_import_guard.py
```diff
@@ -5,11 +5,23 @@
...
+)
 
 # We can only use flex attention / BlockMask if torch version >= 2.5.0 and GPU is Turing / SM75 and above
 _SUPPORTS_FLEX_ATTENTION = (
     torch_version_ge("2.5.0")
     and torch.cuda.is_available()
     and torch.cuda.get_device_capability() >= (7, 5)
 )
+
+torchao_version = _get_torchao_version()
```
RdoubleA, line 22: couldn't this return None if is_fbcode? a bit awkward to return either a version string or None, but htne you check for fbcode again below

torchtune/utils/_import_guard.py
RdoubleA: also nit: can you name this as something that implies a yes/no true/false answer (like `_USE_NEW_TENSOR_CORE_TILED_LAYOUT_API`)

RdoubleA (approved):
few minor comments, but no concerns

torchtune/utils/_version.py
ebsmothers: Actually it should generalize since PyTorch versions use the same format

torchtune/utils/_import_guard.py
ebsmothers: Yeah good point. Honestly I have half a mind to just scrap this util altogether now and use `torchao.__version__` directly

ebsmothers: Yeah good point. Just did a check and in fbcode `torchao.__version__ == 'unknown'`, so it at least won't throw an error. Then I can just explicitly say `ao_version = torchao.__version__` in `_import_guard.py`, gate behind `_is_fbcode()` in `_USE_NEW_TENSOR_CORE_TILED_LAYOUT_API`, and delete `_get_torchao_version` altogether

torchtune/utils/_version.py
RdoubleA: if so, should we change the variable name to indicate general use?

jerryzh168:
ah, our diff train is not landed yet, should we land first? it will break internal torchtune I think

RdoubleA:
> ah, our diff train is not landed yet, should we land first? it will break internal torchtune I think

Yeah that sounds good we can wait, I don't have full context on if this is blocking anything

ebsmothers:
@jerryzh168 can you clarify? I don’t fully understand why this PR would break internal. It checks fbcode and uses the old API, which is what I currently see in internal

jerryzh168:
> @jerryzh168 can you clarify? I don’t fully understand why this PR would break internal. It checks fbcode and uses the old API, which is what I currently see in internal

oh what I meant is that torchao internal is not updated with the new name yet

If this PR can fix both internal and external, maybe this one should land first.

PR #1817, opened 2024-10-11
Fix import error in torchtune due to bc breaking changes in torchao

jerryzh168:
Summary: att

Differential Revision: D64261315


jerryzh168:
real fix in https://github.com/pytorch/torchtune/pull/1812

Issue #1704, opened 2024-09-27
Create multimodal instruction dataset builder


RdoubleA:
We current have `multimodal_chat_dataset` which is great for conversations on an image, but many VQA datasets are structured more like instructions where there is a question column, answer column, and image column (see VQA datasets on HF, sort by most downloads). We should add a `multimodal_instruct_dataset` builder to support these types of datasets from the configs.
- This will require upgrading `InputOutToMessages` with image support

krammnic:
So, it looks like that we can check if the data is multimodal in `InputOutToMessages` and then choose column mapping according to result of this check. Then simply build messages like done in basic case adding to this something like `sample[self._column_map["image"]]`. Finally, we return messages and in `multimodal_instruct_dataset` just build SFTDataset with this transform.  Probably will open PR on it.

PR #1820, opened 2024-10-12
Add ``vqa_dataset``, update docs

krammnic:
#### Context
- [x] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

#1704

#### Changelog
* #1704

#### Test plan

#### UX


krammnic:
@RdoubleA Require review.

krammnic:
I see where is the problem. I committed and then switched branch...

krammnic:
@RdoubleA Should be fine now.

torchtune/datasets/multimodal/_multimodal_instruct.py
```diff
@@ -0,0 +1,158 @@
...
+from torchtune.datasets._sft import SFTDataset
+from torchtune.modules.tokenizers import ModelTokenizer
+
+
+def multimodal_instruct_dataset(
+    tokenizer: ModelTokenizer,
+    *,
+    source: str,
+    column_map: Optional[Dict[str, str]] = None,
+    train_on_input: bool = False,
```
RdoubleA, line 20: we are removing train_on_input for multimodal datasets, as it is not really used. you'll notice that `multimodal_chat_dataset` also has this removed

torchtune/datasets/multimodal/_multimodal_instruct.py
```diff
@@ -0,0 +1,158 @@
...
+    new_system_prompt: Optional[str] = None,
+    packed: bool = False,
```
RdoubleA, line 22: we don't currently support packing for multimodal (yet... we are planning to add this soon)

krammnic:
Yeah, some stuff failed( Let me fix then other PRs and then this one.

torchtune/datasets/multimodal/_multimodal_instruct.py
```diff
@@ -0,0 +1,158 @@
...
+    filter_fn: Optional[Callable] = None,
+    split: str = "train",
```
RdoubleA, line 24: would also be good to include the `image_dir` functionality similar to multimodal chat: https://github.com/pytorch/torchtune/blob/main/torchtune/datasets/multimodal/_multimodal.py#L22. this would enable users to configure datasets that either have image paths or the images directly. For example, take a look a MathVista (https://huggingface.co/datasets/AI4Math/MathVista), a math image QA benchmark that has both the image paths and the raw image. some datasets may have one or the other.

This may require some non-trivial changes to InputOutputToMessages however to make it support loading images. I'm happy to provide more guidance on that, but it would be similar to the changes to `ShareGPTToMessages` in this pr: https://github.com/pytorch/torchtune/pull/1667

torchtune/datasets/multimodal/_multimodal_instruct.py
```diff
@@ -0,0 +1,158 @@
...
+    **load_dataset_kwargs: Dict[str, Any],
+) -> Union[SFTDataset, PackedDataset]:
+    """
+    Configure a custom multimodal dataset with user instruction prompts and model responses.
```
RdoubleA, line 28: the dataset builders say multimodal but this is specifically text+image so would be good to call it out. Maybe something like:
```suggestion
    Configure a custom text+image dataset with separate columns for user instruction, image, and model response.
```

torchtune/datasets/multimodal/_multimodal_instruct.py
```diff
@@ -0,0 +1,158 @@
...
+    This builder function can be used to configure a custom multimodal instruct dataset directly from the yaml config
+    as an alternative to :class:`~torchtune.datasets.SFTDataset`, as it is made to be config friendly.
+
+    The dataset should follow this format:
+
+    .. code-block:: text
+
+        |  input          |  image          |  output          |
+        |-----------------|-----------------|------------------|
+        | "user prompt"   | "user image"    | "model response" |
```
RdoubleA, line 39:
```suggestion
        | "user prompt"   | images/1.jpg    | "model response" |
```
It should be an image path, or the image itself (not sure how to represent that in a text example). saying just "user image" is not clear what data type should go here

torchtune/datasets/multimodal/_multimodal_instruct.py
```diff
@@ -0,0 +1,158 @@
...
+
+    If your column names are different, you can use the ``column_map`` parameter to change
+    the expected column names. For example, if your dataset has columns ``"question"``,
+    ``"answer"`` and ``"picture"`` you can use:
+
+        column_map = {"input": "question", "output": "answer", "image": "picture"}
+
+    Masking of the prompt during training is controlled by the ``train_on_input`` flag, which is
```
RdoubleA, line 47: remove train_on_input

torchtune/datasets/multimodal/_multimodal_instruct.py
```diff
@@ -0,0 +1,158 @@
...
+from typing import Any, Callable, Dict, Optional, Union
+
+from torchtune.data import InputOutputToMessages
+from torchtune.datasets._packed import PackedDataset
+from torchtune.datasets._sft import SFTDataset
+from torchtune.modules.tokenizers import ModelTokenizer
+
+
+def multimodal_instruct_dataset(
+    tokenizer: ModelTokenizer,
```
RdoubleA, line 16: for multimodal datasets, this needs to be:
```suggestion
    model_transform: Transform,
```
I would follow `multimodal_chat_dataset` more closely instead of `instruct_dataset` for this one

torchtune/datasets/multimodal/_multimodal_instruct.py
```diff
@@ -0,0 +1,158 @@
...
+            {
+                ...
+            },
+            ...,
+        ]
+
+    ::
+
+        >>> from torchtune.datasets.multimodal import multimodal_instruct_dataset
+        >>> dataset = multimodal_instruct_dataset(
```
RdoubleA, line 97: maybe structure it similar to this example: https://github.com/pytorch/torchtune/blob/7744608c4c455a86b21f4ce0642e6c19bc1cebf4/torchtune/datasets/multimodal/_multimodal.py#L123

and FlamingoTransform should be replaced everywhere in the example with `Llama3VisionTransform` https://github.com/pytorch/torchtune/blob/main/torchtune/models/llama3_2_vision/_transform.py#L17

torchtune/data/_messages.py
```diff
@@ -161,20 +161,23 @@ class InputOutputToMessages(Transform):
...
-            ``output`` not in ``column_map``.
+            ``output`` not in ``column_map`` or ``image`` not in ``column_map``.
     """
 
     def __init__(
         self,
         train_on_input: bool = False,
         column_map: Optional[Dict[str, str]] = None,
         new_system_prompt: Optional[str] = None,
+        is_multimodal: bool = False,
```
RdoubleA, line 176: can we infer this similar to in ShareGPTToMessages?
```
is_multimodal = "image" in sample or (
            "image" in self._column_map and self._column_map["image"] in sample
        )
```
so users don't need to worry about another flag

torchtune/datasets/multimodal/_multimodal_instruct.py
RdoubleA: nvm, I see you already made those changes :)

RdoubleA (commented):
Could you test this builder with an example VQA dataset once comments are addressed? Something like https://huggingface.co/datasets/derek-thomas/ScienceQA using the question and solution columns.

If you're able to easily configure this dataset using the builder you added, and you can successfully decode tokens of a random sample, and the iamge is encoded correctly, then I would declare this builder to be well-tested (in addition to the unit test you put up)

tests/assets/multimodal_instruct_tiny.json
```diff
@@ -0,0 +1,7 @@
+[
+    {
+        "input": "What is presented on image?",
+        "output": "PyTorch logo.",
```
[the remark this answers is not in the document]
krammnic: Yeah, that was example

torchtune/data/_messages.py
krammnic: Sure!

torchtune/datasets/multimodal/_multimodal_instruct.py
krammnic: @RdoubleA Is there will be a problem to add packing in multimodal? There is no issue about it as I see. Might open then next PR with it

RdoubleA: Yes, there is non trivial work to ensure the cross attention masks are also packed and collated correctly, so it would involve 1) adding logic to pack the cross attention masks and 2) create a new collated for packed multimodal

We are actively discussing an overhaul to our dataloader / collate / packing logic in the near term, so I would hold off on multimodal packing support for now.

tests/torchtune/datasets/multimodal/test_multimodal_instruct_dataset.py
```diff
@@ -0,0 +1,58 @@
...
+        )
+        
+        system_prompt_offset = len(system_prompt.split(" ")) + 1
+        
+        expected_tokens = [
+            [0, 6, 4, 6, -2, 4, 2, 9, 2, 6, 7, 5, -1],
+        ]
+        
+        expected_labels = [
+            [np.int64(-100), np.int64(-100), np.int64(-100), np.int64(-100), np.int64(-100), np.int64(-100), np.int64(-100), np.int64(-100), np.int64(-100), np.int64(-100), np.int64(7), np.int64(5), np.int64(-1)]
```
ebsmothers, line 42: nit but is the `np.int64` strictly necessary? I know labels are long dtype, but if we are just comparing raw lists via `==` and not using torch tensor asserts or something I think you shouldn't need it (and it makes the expected values easier to read)

krammnic: Yeah, will be fixed.

krammnic:
Done all small fixes + CI fixes. Only need extra test

krammnic:
@RdoubleA @ebsmothers 
Made all required fixes, manually tested on required dataset:
```
from torchtune.datasets.multimodal import multimodal_instruct_dataset
import torch
from torch import nn
from torchtune.data import Message, PromptTemplate, truncate
from torchtune.modules.transforms import Transform
from typing import Any, Dict, Generator, List, Mapping, Optional, TextIO, Tuple, Union


class DummyTokenizer(ModelTokenizer, Transform):
    def __init__(self, max_seq_len: Optional[int] = None):

    def encode(self, text, add_bos=True, add_eos=True, **kwargs) -> List[int]:
        words = text.split()
        tokens = [len(word) for word in words]
        if add_bos:
            tokens = [self.bos_id] + tokens
        if add_eos:
            tokens = tokens + [self.eos_id]
        return tokens

    def tokenize_messages(
        self,
        messages: List[Message],
        """
        A simplified version of Llama2Tokenizer's ``tokenize_messages`` for testing purposes.
        """
        start_of_turn = True
        end_of_turn = False
        tokenized_messages = []
        mask = []
        for message in messages:
            # If assistant message, this is the end of a turn
            end_of_turn = message.role == "assistant"

            # Prepend BOS on start of new turns
            if start_of_turn:
                tokenized_messages.append(self.bos_id)
                mask.append(message.masked)

            # Tokenize current message, append with masks
            tokens = []
            for item in message.content:
                if item["type"] == "text":
                    tokens = tokens + self.encode(
                        item["content"],
                        add_bos=False,
                        add_eos=False,
                    )
                elif item["type"] == "image":
                    tokens = tokens + [self.image_id]

            tokenized_messages.extend(tokens)
            mask.extend([message.masked] * len(tokens))

            # If assistant message, append EOS at end
            if end_of_turn:
                tokenized_messages.append(self.eos_id)
                mask.append(message.masked)
                end_of_turn = False
                start_of_turn = True
            else:
                start_of_turn = False

            # Break out early if we reach max_seq_len
            if self.max_seq_len and len(tokenized_messages) >= self.max_seq_len:
                break

        # Finally, truncate if necessary
        if self.max_seq_len:
                tokenized_messages, self.max_seq_len, self.eos_id
            )
            mask = truncate(mask, self.max_seq_len, message.masked)

        return tokenized_messages, mask

    def __call__(self, sample: Mapping[str, Any]) -> Mapping[str, Any]:
        messages: List[Message] = sample.pop("messages")
        images = []
        for message in messages:
            images += message.get_media()
        tokens, mask = self.tokenize_messages(messages)
        sample["tokens"] = tokens
        sample["mask"] = mask
        sample["images"] = images
        return sample

    @property
    def eos_id(self):
        return -1

    @property
    def bos_id(self):
        return 0

    @property
    def image_id(self):
        return -2

dataset = multimodal_instruct_dataset(
    model_transform=DummyTokenizer(),
    # source="json",
    source="derek-thomas/ScienceQA",
    # data_files="derek-thomas/ScienceQA",
    column_map={
        "input": "question",
        "output": "solution",
        "image": "image"
    },
    split="train"
)

tokens = dataset[0]
print(tokens)
```
Output:

```
{'input': 'question', 'output': 'solution', 'image': 'image'}
{'tokens': [0, -2, 5, 2, 5, 6, 2, 8, 6, 2, 4, 3, 7, 4, 2, 3, 7, 5, 4, 2, 5, 3, 3, 5, 5, 2, 9, 4, 8, 2, 8, 6, -1], 'mask': [True, True, True, True, True, True, True, True, True, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False, False], 'images': [<PIL.PngImagePlugin.PngImageFile image mode=RGB size=750x429 at 0x7FD20A590D50>], 'labels': [np.int64(-100), np.int64(-100), np.int64(-100), np.int64(-100), np.int64(-100), np.int64(-100), np.int64(-100), np.int64(-100), np.int64(-100), np.int64(2), np.int64(4), np.int64(3), np.int64(7), np.int64(4), np.int64(2), np.int64(3), np.int64(7), np.int64(5), np.int64(4), np.int64(2), np.int64(5), np.int64(3), np.int64(3), np.int64(5), np.int64(5), np.int64(2), np.int64(9), np.int64(4), np.int64(8), np.int64(2), np.int64(8), np.int64(6), np.int64(-1)]}

```

I'm pretty confident that it is actually working

krammnic:
@ebsmothers @RdoubleA Added another fix, locally tests are passed

krammnic:
Yeah, finnaly. Can someone approve and merge? Or should we do some extra checks?

torchtune/data/_messages.py
```diff
@@ -175,30 +175,64 @@ def __init__(
...
                 raise ValueError(
-                    f"Expected a key of 'output' in column_map but found {column_map.keys()}."
+                    f"Expected a key of 'output' in column_map but found {self.column_map.keys()}."
                 )
-            self._column_map = column_map
         else:
-            self._column_map = {"input": "input", "output": "output"}
+            self.column_map = {"input": "input", "output": "output"}
+
+        self._column_map = column_map
```
ebsmothers, line 193: Duplicate of L179?

ebsmothers, line 193: This feels duplicative.. why do we need to define both `self.column_map` and `self._column_map`?

torchtune/datasets/multimodal/_multimodal_instruct.py
```diff
@@ -0,0 +1,140 @@
...
+def multimodal_instruct_dataset(
+    model_transform: Transform,
+    *,
+    source: str,
+    column_map: Optional[Dict[str, str]] = None,
+    new_system_prompt: Optional[str] = None,
+    filter_fn: Optional[Callable] = None,
+    split: str = "train",
+    **load_dataset_kwargs: Dict[str, Any],
+) -> Union[SFTDataset, PackedDataset]:
```
ebsmothers, line 24: Strictly speaking if we don't support packing yet should return type just be `SFTDataset`?

tests/torchtune/datasets/multimodal/test_multimodal_instruct_dataset.py
```diff
@@ -0,0 +1,52 @@
...
+
+        for i in range(len(dataset)):
+            prompt, label, image = (
+                dataset[i]["tokens"],
+                dataset[i]["labels"],
+                dataset[i]["images"],
+            )
+            assert prompt == expected_tokens[i]
+            assert label == expected_labels[i]
+            assert isinstance(image[0], PngImageFile) is True
```
ebsmothers, line 52: nit: do we need `is True` here?

tests/torchtune/datasets/multimodal/test_multimodal_instruct_dataset.py
```diff
@@ -0,0 +1,52 @@
...
+from torchtune.datasets.multimodal import multimodal_instruct_dataset
+
+
+class TestMultimodalInstructDataset:
+    @pytest.fixture
+    def tokenizer(self):
+        return DummyTokenizer()
+
+    @pytest.mark.parametrize("train_on_input", [True, False])
+    def test_get_item(self, tokenizer, train_on_input):
```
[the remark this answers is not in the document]
krammnic: Yeah, good point

torchtune/datasets/multimodal/_multimodal_instruct.py
krammnic: Sure, typehint needs to be fixed

tests/torchtune/datasets/multimodal/test_multimodal_instruct_dataset.py
krammnic: Probably True might be removed, yes

torchtune/data/_messages.py
krammnic: Hmm, will think about this, problem that we can't fully define column_mao in `__init__` because of is_multimodal check in`__call__`

krammnic:
@ebsmothers Fixed, according to your comments

krammnic:
Can someone start CI and merge then?

torchtune/data/_messages.py
```diff
@@ -175,30 +175,60 @@ def __init__(
     ):
         self.train_on_input = train_on_input
         self.new_system_prompt = new_system_prompt
-        if column_map:
-            if "input" not in column_map:
+
+        self.column_map = column_map
+
+        if self.column_map:
```
joecummings, line 181:
```suggestion
        if self.column_map is not None:
```

torchtune/datasets/multimodal/_multimodal_instruct.py
```diff
@@ -0,0 +1,139 @@
...
+          data_files: my_dataset.json
+          column_map:
+            input: question
+            output: answer
+          train_on_input: False
+          packed: False
+          split: train
+
+    Returns:
+        Union[SFTDataset, PackedDataset]: the configured :class:`~torchtune.datasets.SFTDataset`
```
joecummings, line 124: Return type here needs to be fixed.

torchtune/data/_messages.py
```diff
@@ -175,30 +175,60 @@ def __init__(
...
-                    f"Expected a key of 'output' in column_map but found {column_map.keys()}."
+                    f"Expected a key of 'output' in column_map but found {self.column_map.keys()}."
                 )
-            self._column_map = column_map
         else:
-            self._column_map = {"input": "input", "output": "output"}
+            self.column_map = {"input": "input", "output": "output"}
 
     def __call__(self, sample: Mapping[str, Any]) -> Mapping[str, Any]:
+        if self.column_map is not None:
```
joecummings, line 194: I don' think we can make the assumption that "image" will always be the column for media. I'd much rather have the `ToMessages` classes take care of converting any images to the PIL format. 

Can you remove this section from this PR?

Then, we can think about how to modify our message converters to handle this loading.

joecummings (commented):
Mostly looks good, but I don't think the Message class should be automatically converting the images to PIL b/c it's assuming which column is the image input.

RdoubleA:
hey @krammnic , I've added you on Discord. Let's coordinate further there with @joecummings and we can update this thread after

torchtune/datasets/multimodal/_vqa.py
```diff
@@ -0,0 +1,138 @@
...
+        >>> tokens = dataset[0]["tokens"]
+        >>> model_transform.decode(tokens)
+        "What is presented on the image?PyTorch logo."
+
+    This can also be accomplished via the yaml config:
+
+    .. code-block:: yaml
+
+        dataset:
+          _component_: torchtune.datasets.multimodal.multimodal_instruct_dataset
```
joecummings, line 113: Update.

tests/assets/multimodal_instruct_tiny.json
RdoubleA, line 1: nit: can rename this to vqa_tiny.json

torchtune/data/_messages.py
RdoubleA: > I don' think we can make the assumption that "image" will always be the column for media.

we are not making this assumption, users can still change the "image" column in column_map, in which case we pull from there

torchtune/datasets/multimodal/_vqa.py
krammnic: Fixed

torchtune/data/_messages.py
```diff
@@ -175,30 +175,60 @@ def __init__(
...
+            is_multimodal = "image" in sample or (
+                "image" in self.column_map and self.column_map["image"] in sample
```
RdoubleA, line 196: if "image" is in sample, we can assume the user is not using `column_map`. The `or` should short circuit and return true without checking column_map, so we don't need to check `if self.column_map is not None:` above. we can also add "image" key in default column map:
`self.column_map = {"input": "input", "output": "output", "image": "image"}`

torchtune/data/_messages.py
```diff
@@ -175,30 +175,60 @@ def __init__(
...
+            )
+        else:
+            is_multimodal = "image" in sample
+
+        if is_multimodal and "image" not in self.column_map:
```
RdoubleA, line 201: if we use the default column_map in my comment above if user does not pass in column map, then we don't need to do this

torchtune/datasets/multimodal/_vqa.py
```diff
@@ -0,0 +1,138 @@
...
+    *,
+    source: str,
+    column_map: Optional[Dict[str, str]] = None,
+    new_system_prompt: Optional[str] = None,
+    filter_fn: Optional[Callable] = None,
+    split: str = "train",
+    **load_dataset_kwargs: Dict[str, Any],
+) -> SFTDataset:
+    """
+    Configure a custom text+image dataset with separate columns for user instruction, image, and model response.
```
RdoubleA, line 25: Can we specify this docstring more for visual question-answer?

torchtune/datasets/multimodal/_vqa.py
```diff
@@ -0,0 +1,138 @@
...
+
+    This builder function can be used to configure a custom multimodal instruct dataset directly from the yaml config
+    as an alternative to :class:`~torchtune.datasets.SFTDataset`, as it is made to be config friendly.
+
+    The dataset should follow this format:
+
+    .. code-block:: text
+
+        |  input          |  image          |  output          |
```
RdoubleA, line 34: same with this example. I wonder if we should use quesiton and answer as default columns instead of input and output, but I'm ok either way

torchtune/datasets/multimodal/_vqa.py
```diff
@@ -0,0 +1,138 @@
...
+        |-----------------|-----------------|------------------|
+        | "user prompt"   | images/1.jpg    | "model response" |
+
+    If your column names are different, you can use the ``column_map`` parameter to change
+    the expected column names. For example, if your dataset has columns ``"question"``,
+    ``"answer"`` and ``"picture"`` you can use:
+
+        column_map = {"input": "question", "output": "answer", "image": "picture"}
+
+    Masking of the prompt during training is controlled by the ``train_on_input`` flag, which is
```
RdoubleA, line 44: since we've removed train_on_input, we should remove this from the docstring

torchtune/datasets/multimodal/_vqa.py
```diff
@@ -0,0 +1,138 @@
...
+
+def vqa_dataset(
+    model_transform: Transform,
+    *,
+    source: str,
+    column_map: Optional[Dict[str, str]] = None,
+    new_system_prompt: Optional[str] = None,
+    filter_fn: Optional[Callable] = None,
+    split: str = "train",
+    **load_dataset_kwargs: Dict[str, Any],
```
RdoubleA, line 22: do we need an image_dir parameter if users have downloaded the iamges to a particular local directory?

torchtune/data/_messages.py
```diff
@@ -175,30 +175,60 @@ def __init__(
...
+            self.column_map["image"] = "image"
+
+        if is_multimodal:
+            image_path = sample[self.column_map["image"]]
+            if isinstance(image_path, str):
+                # Load if not loaded
+                pil_image = load_image(image_path)
```
RdoubleA, line 208: the images themselves could be saved to some local directory if the user had to download the image dataset separately. we should provide an option to pass a path to prepend to all image paths in the dataset in this case

RdoubleA (commented):
Thanks for the VQA rename we discussed offline and all your work so far. We should also update the rst docs to make sure this builder is discoverable there. There's a few comments that I think should be addressed but we're nearly there.

torchtune/data/_messages.py
joecummings: This should be handled in a `ToMessage` formatter, not here though.

RdoubleA: this is in `InputOutputToMessages` which seems sensible to me, were you thinking of a different location?

krammnic:
Pushed some fixes

krammnic:
I have a feeling that we will have to separate 2 version of InputToMessages to not break SRP for example

krammnic:
@RdoubleA @joecummings Can we merge?

torchtune/datasets/multimodal/_vqa.py
RdoubleA: bumping this comment

torchtune/data/_messages.py
SalmanMohammadi: I'm with @RdoubleA on this, `InputOutputToMessages` handles this. We should be documenting this, though.

torchtune/data/_messages.py
```diff
@@ -175,30 +175,54 @@ def __init__(
     ):
         self.train_on_input = train_on_input
```
SalmanMohammadi, line 176: I have this slight nagging feeling that this transform is over-generalized and it wouldn't be harmful to define another transform which should properly document that the column map can include image, and to also include validation logic, and without needing to check for multimodal.

Fine to leave this as a followup but if there's consensus here I'd like to see an issue.

torchtune/datasets/multimodal/_vqa.py
```diff
@@ -0,0 +1,132 @@
...
+
+    .. code-block:: yaml
+
+        dataset:
+          _component_: torchtune.datasets.multimodal.vqa_dataset
+          source: json
+          data_files: my_dataset.json
+          column_map:
+            input: question
+            output: answer
```
SalmanMohammadi, line 112:
```suggestion
            output: answer
            image: picture
```

torchtune/datasets/multimodal/_vqa.py
```diff
@@ -0,0 +1,132 @@
...
+          train_on_input: False
+          packed: False
+          split: train
```
SalmanMohammadi, line 115:
```suggestion
          split: train
```

torchtune/datasets/multimodal/_vqa.py
```diff
@@ -0,0 +1,132 @@
...
+        | "user prompt"   | images/1.jpg    | "model response" |
+
+    If your column names are different, you can use the ``column_map`` parameter to change
+    the expected column names. For example, if your dataset has columns ``"question"``,
+    ``"answer"`` and ``"picture"`` you can use:
+
+        column_map = {"input": "question", "output": "answer", "image": "picture"}
+
+    Args:
+        model_transform (Transform): callable that applies model-specific pre-processing to the sample.
```
SalmanMohammadi, line 45: Can we be more concrete here? For example, the cauldron dataset builder describes this arg as:
```python
        model_transform (Transform): model-specific transform class that takes in a sample dict and applies custom
            transforms on the keys. It should consist of at minimum two components: text tokenization (called
            on the "messages" field) and image transform (called on the "images" field). The keys returned by
            the model transform should be aligned with the expected inputs into the model.
```

torchtune/data/_messages.py
SalmanMohammadi: I see this is the same for `ShareGPTToMessages`, so maybe it's fine. We need to properly document MM usage either way.

torchtune/datasets/multimodal/_vqa.py
```diff
@@ -0,0 +1,132 @@
...
+
+        |  input          |  image          |  output          |
+        |-----------------|-----------------|------------------|
+        | "user prompt"   | images/1.jpg    | "model response" |
+
+    If your column names are different, you can use the ``column_map`` parameter to change
+    the expected column names. For example, if your dataset has columns ``"question"``,
+    ``"answer"`` and ``"picture"`` you can use:
+
+        column_map = {"input": "question", "output": "answer", "image": "picture"}
```
SalmanMohammadi, line 42: I'm not sure if we need a concrete example of using column_map here, it is better described below.

torchtune/datasets/multimodal/_vqa.py
```diff
@@ -0,0 +1,132 @@
...
+            define source as the data file type (e.g. "json", "csv", "text"), pass
+            in the filepath in ``data_files``, and set ``split="train"``. See `Hugging Face's
+            <https://huggingface.co/docs/datasets/en/package_reference/loading_methods#datasets.load_dataset.path>`_
+            ``load_dataset`` for more details.
+        column_map (Optional[Dict[str, str]]): a mapping to change the expected "input"
+            and "output" column names to the actual column names in the dataset. Keys should be "input" and
+            "output" and values should be the actual column names. Default is None, keeping the default "input"
+            and "output" column names.
+        new_system_prompt (Optional[str]): if specified, prepend a system message. This can
+            serve as instructions to guide the model response. Default is None.
```
SalmanMohammadi, line 58:
```suggestion
        new_system_prompt (Optional[str]): if specified, prepend a system message. This can
            serve as instructions to guide the model response. Setting this will OVERRIDE any system
            messages already present in the dataset. Default is None.
```
We can be more descriptive here - borrowing from mm chat builder.

SalmanMohammadi:
This looks great to me. Coming at this from an outside perspective I'd really love to see a couple follow up issues to make sure we properly document new features we add to our datasets. I've mentioned above some things I immediately found confusing when combing through the current docs in this PR. 

- [ ] We now can document usage of a general vqa MM builder in https://pytorch.org/torchtune/main/basics/multimodal_datasets.html, to the same extent which we have documented the MM chat builder. Alternatively, we can rename the current MM docpage to `Multimodal chat datasets`, and add a new page for MM instruct.
- [ ] We need to add a concrete HF builder using this, and add it at the bottom of https://pytorch.org/torchtune/main/basics/multimodal_datasets.html (or a new instruct MM page)
- [ ] I'd like to see a fine-tuning run with this dataset.

torchtune/datasets/multimodal/_vqa.py
krammnic: I'm not sure if it is. For me it looks fine, but we can remove this description if you want

krammnic:
> This looks great to me. Coming at this from an outside perspective I'd really love to see a couple follow up issues to make sure we properly document new features we add to our datasets. I've mentioned above some things I immediately found confusing when combing through the current docs in this PR.
> 
>     * [ ]  We now can document usage of a general vqa MM builder in https://pytorch.org/torchtune/main/basics/multimodal_datasets.html, to the same extent which we have documented the MM chat builder. Alternatively, we can rename the current MM docpage to `Multimodal chat datasets`, and add a new page for MM instruct.
> Probably the second idea is better.
>     * [ ]  We need to add a concrete HF builder using this, and add it at the bottom of https://pytorch.org/torchtune/main/basics/multimodal_datasets.html (or a new instruct MM page)
> 
>     * [ ]  I'd like to see a fine-tuning run using this dataset.
> Then will run full procedure

Should I make this docs fixes in this PR or open separate?

krammnic:
Oh, lint(I accepted changes right here). Let me fix

SalmanMohammadi:
> Then will run full procedure
> Should I make this docs fixes in this PR or open separate?

We'd be incredibly grateful if you're up for helping out here! I was speaking slightly generally just to make sure some of those tasks get done at some point, so myself (and the other maintainers) are happy to help out, particularly with a training run, and of course, reviews!

krammnic:
Some really weird lint fail.

krammnic:
Is there anything else to add in this PR?

torchtune/data/_messages.py
RdoubleA: Yeah, I see what you mean. Especially as we add more modalities this may get bloated quick. I was the one who initiated this design in ShareGPTToMessages but I think it's worth reconsidering. I'll put up an issue.

torchtune/datasets/multimodal/_vqa.py
```diff
@@ -0,0 +1,135 @@
...
+    Args:
+        model_transform (Transform): callable that applies model-specific pre-processing to the sample.
+            This includes tokenization and any modality-specific transforms. It is expected to return at
+            minimum ``"tokens"`` and ``"mask"`` keys.
+        source (str): path to dataset repository on Hugging Face. For local datasets,
+            define source as the data file type (e.g. "json", "csv", "text"), pass
+            in the filepath in ``data_files``, and set ``split="train"``. See `Hugging Face's
+            <https://huggingface.co/docs/datasets/en/package_reference/loading_methods#datasets.load_dataset.path>`_
+            ``load_dataset`` for more details.
+        image_dir (str): path to the directory containing the images as you are expected to download the COCO dataset
```
RdoubleA, line 54: this should be generalized and not specific to COCO. Maybe use the same docstring from image_dir argument in `InputOutputToMessages`?

RdoubleA (approved):
Appreciate you pushing this all the way through. I think this highlighted some rough parts on our multimodal pipeline which we should follow up on and created some productive discussions. Just one comment, otherwise it's good to go once CI is green.

torchtune/datasets/multimodal/_vqa.py
krammnic: Fixed

Issue #1565, opened 2024-09-12
Handle empty prepend/append tags in PromptTemplate


RdoubleA:
We currently assume the prepend/append tags in the prompt template are a string or an empty string. We should add more flexibility here to ignore the tag if its empty or None:

https://github.com/pytorch/torchtune/blob/01619cec87e8904707a06da91ee3f4759341151c/torchtune/data/_prompt_templates.py#L111

krammnic:
Really minor fix. Check for None and change it to empty string should be fine. Will open PR

PR #1824, opened 2024-10-13
Consistent type checks for prepend and append tags.

krammnic:
#### Context
- [ ] add a new feature
- [X] fix a bug
- [ ] other (please add here)

#1565
#### Changelog
* isinstance checks for prepend and append tags.

#### Test plan

#### UX

Add checks that prepend and append tags are actually valid. Did not make None <=> empty string. Pretty minor fix(written in perfomance most efficient way)

### DIFF
--- a/torchtune/data/_prompt_templates.py
+++ b/torchtune/data/_prompt_templates.py
@@ -110,11 +110,22 @@ class PromptTemplate(PromptTemplateInterface):
             if message.role in self.template:
                 prepend_tag = self.template[message.role][0]
                 append_tag = self.template[message.role][1]
-                content = (
-                    [{"type": "text", "content": prepend_tag}]
-                    + message.content
-                    + [{"type": "text", "content": append_tag}]
-                )
+                if isinstance(prepend_tag, str) and isinstance(append_tag, str):
+                    content = (
+                        [{"type": "text", "content": prepend_tag}]
+                        + message.content
+                        + [{"type": "text", "content": append_tag}]
+                    )
+                elif not isinstance(prepend_tag, str) and isinstance(append_tag, str):
+                    content = message.content + [
+                        {"type": "text", "content": append_tag}
+                    ]
+                elif isinstance(prepend_tag, str) and not isinstance(append_tag, str):
+                    content = [
+                        {"type": "text", "content": prepend_tag}
+                    ] + message.content
+                else:
+                    content = {message.content}
             else:
                 content = message.content
             formatted_dialogue.append(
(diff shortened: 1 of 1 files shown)


krammnic:
@RdoubleA Require some review

krammnic:
Pretty minor, so waiting for merge.

torchtune/data/_prompt_templates.py
```diff
@@ -110,11 +110,22 @@ def __call__(
             if message.role in self.template:
                 prepend_tag = self.template[message.role][0]
                 append_tag = self.template[message.role][1]
-                content = (
-                    [{"type": "text", "content": prepend_tag}]
-                    + message.content
-                    + [{"type": "text", "content": append_tag}]
-                )
+                if isinstance(prepend_tag, str) and isinstance(append_tag, str):
```
RdoubleA, line 113: maybe we should structure this slightly different to avoid the cascade of ifs:
```
content = message.content
if message.role in self.template:
    prepend_tag = self.template[message.role][0]
    append_tag = self.template[message.role][1]
    if prepend_tag is not None and len(prepend_tag) > 0:
        # add prepend tag
    if append_tag is not None and len(append_tag) > 0:
        # add append tag
formatted_dialogue.append(Message(...))
```

torchtune/data/_prompt_templates.py
```diff
@@ -183,13 +194,30 @@ def __call__(
...
+                    ] + message.content
+                else:
+                    content = message.content
             else:
-                content = (
-                    [{"type": "text", "content": prepend_tag}]
-                    + message.content
-                    + [{"type": "text", "content": append_tag}]
-                )
+                if isinstance(prepend_tag, str) and isinstance(append_tag, str):
```
RdoubleA, line 204: similar comment above

torchtune/data/_prompt_templates.py
krammnic: It depends on if can something "bad" except None can come there

krammnic: Ah, I see now how do you want to do it. Ok, let me fix it like this. But I would like to save isinstance there

torchtune/data/_prompt_templates.py
```diff
@@ -110,11 +110,13 @@ def __call__(
             if message.role in self.template:
                 prepend_tag = self.template[message.role][0]
                 append_tag = self.template[message.role][1]
-                content = (
-                    [{"type": "text", "content": prepend_tag}]
-                    + message.content
-                    + [{"type": "text", "content": append_tag}]
-                )
+                content = {message.content}
```
RdoubleA, line 113: the {} should be removed here, otherwise we cannot concatenate lists below. otherwise the rest looks good

krammnic: Fixed in both cases

krammnic:
@RdoubleA Probably now all issues are fixed(CI might be reruned) We will save isinstance + len > 0 to handle anything that is not good as prepend/append in this case

torchtune/data/_prompt_templates.py
```diff
@@ -110,11 +110,13 @@ def __call__(
...
-                    + [{"type": "text", "content": append_tag}]
-                )
+                content = message.content
+
+                if isinstance(prepend_tag, str) and len(prepend_tag) > 0:
+                    content = [{"type": "text", "content": prepend_tag}] + content
+
+                if isinstance(append_tag, str) and len(append_tag) > 0:
+                    content = content + [{"type": "text", "content": append_tag}]
             else:
```
RdoubleA, line 120: Nit: you could move `content = message.content` above the if statement and remove this else entirely

torchtune/data/_prompt_templates.py
```diff
@@ -183,13 +185,21 @@ def __call__(
                 and index == len(messages) - 1
                 and len(message.text_content) == 0
             ):
-                content = [{"type": "text", "content": prepend_tag}] + message.content
+                if isinstance(prepend_tag, str) and len(prepend_tag) > 0:
+                    content = [
+                        {"type": "text", "content": prepend_tag}
+                    ] + message.content
+                else:
+                    content = message.content
```
RdoubleA, line 193: Similar comment here

RdoubleA (approved):
Excellent work, will wait for CI

torchtune/data/_prompt_templates.py
krammnic: Fixed

torchtune/data/_prompt_templates.py
krammnic: Fixed

Issue #1647, opened 2024-09-22
[Feat] Show trace filenames in logs


Jackmin801:
It is currently difficult to figure out which trace corresponds to which experiment with the current naming convention. It might be easier to map them if the output contained the filenames

It is difficult to figure out which trace corresponds to which experiment
<img width="370" alt="image" src="https://github.com/user-attachments/assets/b4d6832d-0fbf-4ba4-9f1d-6cf02629d311">

The output log does not provide this information
<img width="1399" alt="image" src="https://github.com/user-attachments/assets/9bdefc5d-fa1e-41b7-8b10-1ca2b1416656">

felipemello1:
FYI, the number you see in the trace is a timestamp. The highest number should be the most recent. But I agree that its a pain point and it could be better. Thanks for raising the issue!

krammnic:
Easily can be fixed! 

```
exporter = tensorboard_trace_handler(
        curr_trace_dir, worker_name=f"rank{rank}", use_gzip=True
)
```

Even If we don't have ability to straightly change name of trace in `tensorboard_trace_handler`, we can add something throw `worker_name`. 

Possible solution is:

```
exporter = tensorboard_trace_handler(
        curr_trace_dir, worker_name=f"rank{rank}_" + f"{socket.gethostname()}_{os.getpid()}", use_gzip=True
)
```

Will open PR.

PR #1825, opened 2024-10-13
More consistent trace names.

krammnic:
#### Context
- [ ] add a new feature
- [X] fix a bug
- [ ] other (please add here)

#1647
#### Changelog
* PID + socket hostname in trace files names

#### Test plan

#### UX

### DIFF
--- a/torchtune/training/_profiler.py
+++ b/torchtune/training/_profiler.py
@@ -6,7 +6,6 @@
 
 
 import datetime
-import glob
 import os
 import time
 from functools import partial
@@ -99,22 +98,15 @@ def trace_handler(
 
     # Use tensorboard trace handler rather than directly exporting chrome traces since
     # tensorboard doesn't seem to be able to parse traces with prof.export_chrome_trace
+
+    now = datetime.datetime.now()
+
     exporter = tensorboard_trace_handler(
         curr_trace_dir,
-        worker_name="rank0",
+        worker_name=f"r0-{now.year}-{now.month}-{now.day}-{now.hour}-{now.minute}",
         use_gzip=True,
     )
     exporter(prof)
-
-    latest_trace = max(
-        glob.glob(curr_trace_dir + "/*.pt.trace.json.gz"), key=os.path.getctime
-    )
-
-    now = datetime.datetime.now()
-    os.rename(
-        latest_trace,
-        f"{curr_trace_dir}/r0-{now.year}-{now.month}-{now.day}-{now.hour}-{now.minute}.pt.trace.json.gz",
-    )
 
     if rank == 0:
         log.info(f"Finished dumping traces in {time.monotonic() - begin:.2f} seconds")
--- a/torchtune/training/_profiler.py
+++ b/torchtune/training/_profiler.py
@@ -6,6 +6,7 @@
 
 
 import os
+import socket
 import time
 from functools import partial
 from pathlib import Path
@@ -98,7 +99,9 @@ def trace_handler(
     # Use tensorboard trace handler rather than directly exporting chrome traces since
     # tensorboard doesn't seem to be able to parse traces with prof.export_chrome_trace
     exporter = tensorboard_trace_handler(
-        curr_trace_dir, worker_name=f"rank{rank}", use_gzip=True
+        curr_trace_dir,
+        worker_name=f"rank{rank}_" + f"{socket.gethostname()}_{os.getpid()}",
+        use_gzip=True,
     )
     exporter(prof)
 
--- a/torchtune/training/_profiler.py
+++ b/torchtune/training/_profiler.py
@@ -5,8 +5,9 @@
 # LICENSE file in the root directory of this source tree.
 
 
+import datetime
+import glob
 import os
-import socket
 import time
 from functools import partial
 from pathlib import Path
@@ -100,10 +101,20 @@ def trace_handler(
     # tensorboard doesn't seem to be able to parse traces with prof.export_chrome_trace
     exporter = tensorboard_trace_handler(
         curr_trace_dir,
-        worker_name=f"rank{rank}_" + f"{socket.gethostname()}_{os.getpid()}",
+        worker_name="rank0",
         use_gzip=True,
     )
     exporter(prof)
+
+    latest_trace = max(
+        glob.glob(curr_trace_dir + "/*.pt.trace.json.gz"), key=os.path.getctime
+    )
+
+    now = datetime.datetime.now()
+    os.rename(
+        latest_trace,
+        f"{curr_trace_dir}/r0-{now.year}-{now.month}-{now.day}-{now.hour}-{now.minute}.pt.trace.json.gz",
+    )
 
     if rank == 0:
         log.info(f"Finished dumping traces in {time.monotonic() - begin:.2f} seconds")


krammnic:
@RdoubleA Would love to get your comment about this

krammnic:
Such format is inspired by tensorboard. Probably the best we can do without touching public API

RdoubleA:
Could you share an example profile with the updated name?

krammnic:
`rank0-mipt-1465.17883.pt.trace.json.gz`

We can't use `__name__`. We also can't get some expirement name without new argument or for example wandb. We can't remove first time number - it is tensorboard internal. So here you can determine newest or some expirement by PID and host. Maybe we can find something better?

krammnic:
Hmm, maybe add named argument trace_name: str = "" in trace_handler? But this will require extra attention to user. So I assume current approach might be first step to fix this design issue.

torchtune/training/_profiler.py
```diff
@@ -98,7 +99,9 @@ def trace_handler(
     # Use tensorboard trace handler rather than directly exporting chrome traces since
     # tensorboard doesn't seem to be able to parse traces with prof.export_chrome_trace
     exporter = tensorboard_trace_handler(
-        curr_trace_dir, worker_name=f"rank{rank}", use_gzip=True
+        curr_trace_dir,
+        worker_name=f"rank{rank}_" + f"{socket.gethostname()}_{os.getpid()}",
```
ebsmothers, line 103: Sorry noob question on this choice of worker_name: if I am launching a bunch of runs with profiling on the same host and not keeping track of the pid when I launch, does this actually solve the problem? Like why not instead allow the manual specification of an output filename or something?

krammnic: @ebsmothers We can do a mamed argument probably. But I was speaking about solution which comes "out of the box". If we will do something like expirement_name: str = "", probably it wan't be usually defined if we don't actually require to define it. Let me update the PR and see if we can do better

felipemello1:
Thanks for the PR! having a user experiment name would be neat. I think that having the date is a big deal too. Something like:

f"{year}_{month}_{day}_{hour}_{min}_{sec}_rank{rank}_{optional_exp_name}"

what do you think?

krammnic:
So for example name in issue case will look like:

2024_09_14_10_09_32_rank0.{Here some big number}_llama3_8b

felipemello1:
since "some big number" doesnt seem to be much informative, maybe this can go last?

I also think that model could be always there, and just keep the exp name optional:
2024_09_14_10_09_32_rank0_llama3_8b_{optional_exp_name}.{Here some big number}

I just dont know if there is a character limit :/
If "some big number" is usually unique, maybe we can remove the "seconds"

2024_09_14_10_09_rank0_llama3_8b_{optional_exp_name}.{Here some big number}

and if the trace is being saved anyway inside of the model folder, then we probably dont need the model name:
2024_09_14_10_09_rank0_{optional_exp_name}.{Here some big number}

krammnic:
> since "some big number" doesnt seem to be much informative, maybe this can go last?
> 
> I also think that model could be always there, and just keep the exp name optional: 2024_09_14_10_09_32_rank0_llama3_8b_{optional_exp_name}.{Here some big number}
> 
> I just dont know if there is a character limit :/ If "some big number" is usually unique, maybe we can remove the "seconds"
> 
> 2024_09_14_10_09_rank0_llama3_8b_{optional_exp_name}.{Here some big number}
> 
> and if the trace is being saved anyway inside of the model folder, then we probably dont need the model name: 2024_09_14_10_09_rank0_{optional_exp_name}.{Here some big number}

Your current idea is fine, but isn't filename too long? Even though, lets probably add exp_name + dates as you proposed

felipemello1:
I agree its long. Maybe rank0 -> r0?

model_path/2024_09_14_10_09_r0_{optional_exp_name}.{big number}.trace.json.gzip

But i think that it may raise errors because of the length. Some maybe do optional_exp_name[:max_num_characters]. I think that this works for me, if no one has issues with it. n00b question: is this "some big number" already in our current tracing? I dont 100% follow why this is cannot be optional.

krammnic:
> I agree its long. Maybe rank0 -> r0?
> 
> model_path/2024_09_14_10_09_r0_{optional_exp_name}.{big number}.trace.json.gzip
> 
> But i think that it may raise errors because of the length. Some maybe truncate --> optional_exp_name[:max_num_characters]. I think that this works for me, if no one has issues with it.
> 
> n00b question: is this "big number" already in our current tracing fname?

Yes, see examples in issue. It is already included in `tensorboard_trace_handler`.

```file_name = f"{worker_name}.{time.time_ns()}.pt.trace.json"```

**Comment by krammnic** [comment]
Actually we can do some python magic and try to redefine name as we want

**Comment by krammnic** [comment]
> ~Actually we can do some python magic and try to redefine name as we want~. Oh, I have a better idea. let's just save with basic name and rename!

No, we can't to do like this because in internal of tensorboard it is done like:
``time.time_ns()``

**Comment by krammnic** [comment]
I don't want to take ``tensorboard_trace_handler`` and implement it in our code either

**Comment by krammnic** [comment]
@felipemello1 Maybe do something like:

```
exporter = tensorboard_trace_handler(
        curr_trace_dir,
        worker_name="rank0",
        use_gzip=True,
    )
    exporter(prof)
latest_trace = max(glob.glob(curr_trace_dir + "/*.pt.trace.json.gz"), key=os.path.getctime)

now = datetime.datetime.now()
os.rename(latest_trace, f"{curr_trace_dir}/r0-{now.year}-{now.month}-{now.day}-{now.hour}-{now.minute}.pt.trace.json.gz")
```

With this we can remove useless number and do not touch internals

**Comment by krammnic** [comment]
@RdoubleA @joecummings Can we consider such trick as fine?

**Comment by felipemello1** [comment]
i dont think so. We should be using trace_handler. Let me get a reference for you

**Comment by krammnic** [comment]
> i dont think so. We should be using trace_handler. Let me get a reference for you

Yeah, I'm using trace_handler, but I'm renaming newest file that came from it with more consistent name that we have chosen. We can't remove "big number" that is pretty useless without touching internals, we also are not really flexible in name choice. And if we use dates as proposed earlier the name will be to big. Now it is pretty strict: `r0-2024-10-14-16-8.pt.trace.json.gz`

**Comment by krammnic** [comment]
> > i dont think so. We should be using trace_handler. Let me get a reference for you
> 
> Yeah, I'm using trace_handler, but I'm renaming newest file that came from it with more consistent name that we have chosen. We can't remove "big number" that is pretty useless without touching internals, we also are not really flexible in name choice. And if we use dates as proposed earlier the name will be to big. Now it is pretty strict: `r0-2024-10-14-16-8.pt.trace.json.gz`

Without this trick it will be something like this in finetuning cases:

`r0.12345678901123456789.2024-10-14-16-8.pt.trace.json.gz`

**Comment by felipemello1** [comment]
Not a big fan. I think it works, but it can be dangerous. Do we know for a fact that if we dont rename it, the name will be too big?

**Comment by krammnic** [comment]
> Not a big fan. I think it works, but it can be dangerous. Do we know for a fact that if we dont rename it, the name will be too big?

Yeah, I don't like it much either, It is better though then example that I've shown previously. Another way to get such format - do not use tensorboard_trace_handler or modify it's internals(we can't do it)

**Comment by krammnic** [comment]
@felipemello1 Easiest solution probably: re-implement this tensorboard.trace_handler and add normal trace_name: str argument.

**Comment by krammnic** [comment]
Maybe PR to tensorboard?

**Comment by felipemello1** [comment]
The PR or reimplementation are valid options, but i think for the sake of this PR, just having a better date in the name is already much better than what we have. I would be satisfied with it, and we can follow up in another PR, if thats the case. What do you think?

**Comment by krammnic** [comment]
> The PR or reimplementation are valid options, but i think for the sake of this PR, just having a better date in the name is already much better than what we have. I would be satisfied with it, and we can follow up in another PR, if thats the case. What do you think?

So we do not touch useless big number right now and just save the date then?

**Comment by felipemello1** [comment]
yes, i am fine with it :)

**Comment by krammnic** [comment]
> yes, i am fine with it :)

done

**Comment by felipemello1** [comment]
awesome, thank you! do you mind running it once with profiler.enabled=True and confirming it works?

**Comment by krammnic** [comment]
> awesome, thank you! do you mind running it once with profiler.enabled=True and confirming it works?

Alreadty have done this:
Screenshot from 2024-10-14 19-12-00

**Comment by krammnic** [comment]
Reproducible code:

```
from torch.profiler import profile, record_function, ProfilerActivity
import time
from torchtune.training._profiler import trace_handler


with profile(activities=[ProfilerActivity.CPU], record_shapes=True) as prof:
    time.sleep(2)

trace_handler(prof, output_dir=".")
```

felipemello1 (approved):

PR #1833, opened 2024-10-14
Support Optimizer-in-the-backward

mori360:
#### Context
- [x] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

Enable Optimizer-in-the-backward for full_finetune_distributed

#### Changelog

- Update full_finetune_distributed for enabling Optimizer-in-the-backward
- Update test_full_finetune_distributed with `_optimizer_in_bwd` config
- updated test_distributed to test running with/without optimized_in_the_backward, and performance after saving-loading state_dict.

#### Test plan

- Test running with optimizer_in_the_backward: `tune run --nproc_per_node 2 full_finetune_distributed --config llama2/7B_full fsdp_cpu_offload=False max_steps_per_epoch=2 optimizer_in_bwd=True`
- Test running optimizer_in_the_backward with resume_from_checkpoint: ` tune run --nproc_per_node 2 full_finetune_distributed --config llama2/7B_full fsdp_cpu_offload=False max_steps_per_epoch=2 epochs=10 optimizer_in_bwd=True  resume_from_checkpoint=True checkpointer.recipe_checkpoint=/tmp/Llama-2-7b-hf/recipe_state.pt checkpointer.checkpoint_files=[hf_model_0001_1.pt,hf_model_0002_1.pt]`
- Verify that running with Optimizer-in-the-backward could have the same loss, model_state_dict and optimizer_state_dict, model after saving and loading could also have the same: `pytest tests/torchtune/training/test_distributed.py -k test_optimizer_in_backward`

Memory cost analysis:
With each layer gradient cost 193MB memory, the origin(left) case has the peak memory at the 31th layer with accumulation of 193MB memory times 30.
The right case with optimizer-in-the-backward frees these memory during backward, gets lower peak memory.
memory compare

Training time and loss analysis:
training time and loss


recipes/full_finetune_distributed.py
```diff
@@ -657,12 +705,18 @@ def train(self) -> None:
...
                     if self._clip_grad_norm is not None:
-                        grad_norm = torch.nn.utils.clip_grad_norm_(
-                            self._model.parameters(),
-                            max_norm=float(self._clip_grad_norm),
-                        )
-                    self._optimizer.step()
-                    self._optimizer.zero_grad(set_to_none=True)
+                        if self._optimizer_in_bwd:
+                            raise NotImplementedError(
+                                "Gradient clipping is not supported after optimizer-in-the-backward."
```
mori360, line 710: `optimizer_in_backward` frees gradient information during `loss.backward`, could not get the correct `grad_norm`

recipes/full_finetune_distributed.py
```diff
@@ -681,7 +735,12 @@ def train(self) -> None:
                         time_per_step = time.perf_counter() - t0
                         log_dict = {
                             "loss": loss_to_log,
-                            "lr": self._optimizer.param_groups[0]["lr"],
+                            "lr": get_lr(
```
mori360, line 738: combine `get_lr` as an utils for both distributed and single_device to validate if all the LR are the same and return if True

tests/recipes/test_full_finetune_distributed.py
```diff
@@ -29,7 +29,10 @@
 
 
 class TestFullFinetuneDistributedRecipe:
-    def _get_test_config_overrides(self):
```
mori360, line 32: Both "optimizer_in_bwd=True" and "clip_grad_norm=100" could cause the wrong grad_norm, separate them here to avoid, loss_value would not be affected by either "optimizer_in_bwd=True" or "clip_grad_norm=100"

torchtune/training/lr_schedulers.py
```diff
@@ -54,3 +54,27 @@ def lr_lambda(current_step: int) -> float:
         return max(0.0, cosine_lr_multiple)
 
     return LambdaLR(optimizer, lr_lambda, last_epoch)
+
+
+def get_lr(optimizer_in_bwd, optimizer) -> str:
```
mori360, line 59: Based on optimizer_in_bwd, the second input could be pure optimizer or optim_wrapper, do we have a better name for the second input? @weifengpy

tests/recipes/test_full_finetune_distributed.py
```diff
@@ -60,9 +63,17 @@ def _fetch_expected_loss_values(self, model_type):
             ("llama3/8B_full", "llama3", "tune", "NO_SHARD"),
         ],
     )
+    @pytest.mark.parametrize("optim_in_bwd", [True, False])
```
mori360, line 66: Currently add one more param input "optim_in_bwd" to have separate test, shall we have the test in another way? @ebsmothers

torchtune/training/lr_schedulers.py
weifengpy: vanilla_optimizer ?

tests/recipes/test_full_finetune_distributed.py
```diff
@@ -29,7 +29,10 @@
 
 
 class TestFullFinetuneDistributedRecipe:
-    def _get_test_config_overrides(self):
+    def _get_test_config_overrides(self, optim_in_bwd):
```
ebsmothers, line 32: More of a nit, but I would keep `_get_test_config_overrides` argless. Better to just have something like

```
if optim_in_bwd: 
	cmd.append(["optimizer_in_bwd=True"])
else:
	cmd.append(["clip_grad_norm=100"])
```
e.g. just after L104 below.

torchtune/training/lr_schedulers.py
```diff
@@ -54,3 +54,29 @@ def lr_lambda(current_step: int) -> float:
         return max(0.0, cosine_lr_multiple)
 
     return LambdaLR(optimizer, lr_lambda, last_epoch)
+
+
+def get_lr(optimizer_in_bwd, vanilla_optimizer) -> str:
+    """
+    Full_finetune_distributed and full_finetune_single_deivce assume all optimizers have
```
ebsmothers, line 61:
```suggestion
    Full_finetune_distributed and full_finetune_single_device assume all optimizers have
```

recipes/full_finetune_distributed.py
```diff
@@ -456,19 +459,55 @@ def _is_layer_fqn(s: str) -> bool:
         return model
 
     def _setup_optimizer(
-        self, cfg_optimizer: DictConfig, opt_state_dict: Optional[Dict[str, Any]] = None
+        self,
+        cfg_optimizer: DictConfig,
+        optimizer_in_bwd: bool = False,
+        opt_state_dict: Optional[Dict[str, Any]] = None,
     ) -> Optimizer:
```
ebsmothers, line 466: Return type should be `Optional[Optimizer]`

recipes/full_finetune_distributed.py
```diff
@@ -456,19 +459,55 @@ def _is_layer_fqn(s: str) -> bool:
...
+
+            # Register optimizer step hooks on the model to run optimizer in backward.
+            training.register_optim_in_bwd_hooks(
+                model=self._model, optim_dict=optim_dict
+            )
+            # Create a wrapper for checkpoint save/load of optimizer states when running in backward.
+            self._optim_ckpt_wrapper = training.create_optim_in_bwd_wrapper(
+                model=self._model, optim_dict=optim_dict
             )
+            # Load optimizer states for each param.
```
ebsmothers, line 482: Please add a comment analogous to [this](https://github.com/pytorch/torchtune/blob/f639b6d205d5de8006987717b96e323692946cb8/recipes/full_finetune_single_device.py#L411-L413) here

tests/recipes/test_full_finetune_distributed.py
ebsmothers: I think this way is OK

torchtune/training/lr_schedulers.py
```diff
@@ -54,3 +54,29 @@ def lr_lambda(current_step: int) -> float:
         return max(0.0, cosine_lr_multiple)
 
     return LambdaLR(optimizer, lr_lambda, last_epoch)
+
+
+def get_lr(optimizer_in_bwd, vanilla_optimizer) -> str:
```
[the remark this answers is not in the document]
ebsmothers: Also I know we use for logging so maybe it gets cast to str, but I believe the return type here should be float?

torchtune/training/lr_schedulers.py
```diff
@@ -54,3 +54,29 @@ def lr_lambda(current_step: int) -> float:
...
+            param_groups.append(param["param_groups"][0])
+    else:
+        param_groups = vanilla_optimizer.param_groups
+    if len(param_groups) < 1:
+        raise RuntimeError(
+            f"Invalid optimizer param groups with len of: {len(param_groups)}"
+        )
+
+    # LR Schedulers are the same across all param groups for full_finetune right now
+    lr_scheduler = param_groups[0]["lr"]
```
ebsmothers, line 78: This is just the learning rate not the scheduler, right? If so should name it `lr`, not `lr_scheduler`

recipes/full_finetune_distributed.py
```diff
@@ -222,6 +224,7 @@ def setup(self, cfg: DictConfig) -> None:
 
         self._optimizer = self._setup_optimizer(
             cfg_optimizer=cfg.optimizer,
+            optimizer_in_bwd=self._optimizer_in_bwd,
```
weifengpy, line 227: should `_optimizer_in_bwd` be part of `cfg.optimizer`, or we did so to mimic single device recipe ?

recipes/full_finetune_distributed.py
```diff
@@ -657,12 +705,18 @@ def train(self) -> None:
...
+                                "Gradient clipping is not supported after optimizer-in-the-backward."
+                            )
+                        else:
+                            grad_norm = torch.nn.utils.clip_grad_norm_(
+                                self._model.parameters(),
+                                max_norm=float(self._clip_grad_norm),
+                            )
+                    if not self._optimizer_in_bwd:
+                        self._optimizer.step()
+                        self._optimizer.zero_grad(set_to_none=True)
```
weifengpy, line 719: do we still need the `else:`  and `if not self._optimizer_in_bwd:`? considering we are raising error in `if self._optimizer_in_bwd:`

it will reduce indent and make the code easier

recipes/full_finetune_distributed.py
```diff
@@ -550,18 +589,24 @@ def save_checkpoint(
         )
 
         if intermediate_checkpoint:
-            opt_state_dict = training.get_full_optimizer_state_dict(
-                self._optimizer,
-                self._is_rank_zero,
-                device=self._device,
-            )
+            if self._optimizer_in_bwd:
```
weifengpy, line 592: nit: should put `if not self._optimizer_in_bwd` first? the common case should be surfaced to readers first. but this is totally myt personal style

recipes/full_finetune_distributed.py
```diff
@@ -681,7 +735,12 @@ def train(self) -> None:
                         time_per_step = time.perf_counter() - t0
                         log_dict = {
                             "loss": loss_to_log,
-                            "lr": self._optimizer.param_groups[0]["lr"],
+                            "lr": get_lr(
+                                self._optimizer_in_bwd,
+                                self._optim_ckpt_wrapper.state_dict()
+                                if self._optimizer_in_bwd
```
weifengpy, line 741: same nit: should we put `if not self._optimizer_in_bwd` first

recipes/full_finetune_single_device.py
```diff
@@ -658,10 +661,13 @@ def train(self) -> None:
...
                             # true since we don't expose the ability to configure this yet.
-                            "lr": (
-                                self._optim_ckpt_wrapper.get_optim_key("lr")
-                                if self._optimizer_in_bwd
-                                else self._optimizer.param_groups[0]["lr"]
+                            "lr": get_lr(
+                                self._optimizer_in_bwd,
+                                (
+                                    self._optim_ckpt_wrapper.state_dict()
+                                    if self._optimizer_in_bwd
```
weifengpy, line 668: ditto

recipes/full_finetune_distributed.py
mori360: I follow the structure from [single_device](https://github.com/pytorch/torchtune/blob/4bbed4d4e245e990208f3389689291622a7af9df/recipes/full_finetune_single_device.py#L225), shall we restructure both of the distributed and single_deivce?

tests/recipes/test_full_finetune_distributed.py
```diff
@@ -91,6 +99,13 @@ def test_loss(
             cmd.append(f"fsdp_sharding_strategy={fsdp_sharding_strategy}")
         model_config = MODEL_TEST_CONFIGS[model_type]
         cmd = cmd + self._get_test_config_overrides() + model_config
+        # "optimizer_in_bwd=True" would free gradient info before clip_grad, causing
+        # wrong grad_norm, so we only test one of them each time. But loss values
+        # should be the same. torch >= 2.5 required to enable optimizer_in_backward.
+        if not optim_in_bwd:
+            cmd.append("clip_grad_norm=100")
+        elif optim_in_bwd and version.parse(torch.__version__).base_version < "2.5.0":
```
mori360, line 107: `optimizer_in_backward` needs torch version to be no less than 2.5, shall I add the version check here?

tests/recipes/test_full_finetune_distributed.py
```diff
@@ -91,6 +99,13 @@ def test_loss(
             cmd.append(f"fsdp_sharding_strategy={fsdp_sharding_strategy}")
         model_config = MODEL_TEST_CONFIGS[model_type]
         cmd = cmd + self._get_test_config_overrides() + model_config
+        # "optimizer_in_bwd=True" would free gradient info before clip_grad, causing
+        # wrong grad_norm, so we only test one of them each time. But loss values
+        # should be the same. torch >= 2.5 required to enable optimizer_in_backward.
+        if not optim_in_bwd:
+            cmd.append("clip_grad_norm=100")
+        elif optim_in_bwd and version.parse(torch.__version__).base_version >= "2.5.0":
```
mori360, line 107: `optimizer_in_backward` needs torch version to be no less than 2.5, shall I add the version check here?

ebsmothers: Now that 2.5 is stable, I think you do not need this check

ebsmothers (approved):
One more small comment on the versioning question. After that this should be good to go

gameofdimension:
great work.
i am wondering whether it can be used with `cilp_grad_norm_`.

awgu:
Optimizer in backward and global gradient norm clipping does not algorithmically make sense 🤔

gameofdimension:
so if `cilp_grad_norm_ ` is required then we can not use  "Optimizer in backward"?

awgu:
I think you would need to do something different mathematically, e.g. use previous iteration's total norm or clip each gradient separately.

Issue #1813, opened 2024-10-11
support for gemma-2 


almugabo:
Could you please add fine-tuning support for gemma-2 ? It has good good multilingual capabilities and is a good candidate for fine-tuning for languages other than English. 
Its different sizes also make it attractive for fine-tuning for different tasks. 
I would gladly help but am not knowledgeable enough 
Thank you

krammnic:
Actually, in inspiration of [one current kaggle competition](https://www.kaggle.com/competitions/gemma-language-tuning) - it is really good idea too add this pretty soon.

ebsmothers:
Thanks @almugabo for creating the issue. I think this will be a bit of effort, quickly jotting down a couple of things I'm aware of that we'd need to support:

- [ ] Logit softcapping in attention layer
- [ ] Logit softcapping in output layer
- [ ] Sliding window attention
- [ ] Expose sliding window attention in configurable set of layers
- [ ] Post layernorm for both attention and FFN (hacky thing to do is just use `attn_scale` and `mlp_scale`)
- [ ] Model builders for 2B, 9B, 27B sizes

For logit softcapping and sliding window attention, I suspect we can use FlexAttention APIs. See [this blog post](https://pytorch.org/blog/flexattention/) where they give explicit examples of each.

Optimox:
Hello, I have started the addition of gemma2, I will create now my PR in WIP mode. I haven't run any test yet but will do soon!

Optimox:
I didn't know about FlexAttention, I will look into it!

joecummings:
ADDED!

PR #1835, opened 2024-10-15
feat: add gemma2b variants

Optimox:
#### Context
- [x] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

This is related to adding gemma2 support #1813 

#### Changelog
*

#### Test plan

#### UX


torchtune/modules/attention.py
```diff
@@ -306,3 +307,291 @@ def forward(
         # reshape the output to be the same shape as the input
         output = output.transpose(1, 2).contiguous().view(b, s_x, -1)
         return self.output_proj(output)
+
+
+class Gemma2Attention(nn.Module):
```
ebsmothers, line 312: I think we should consider putting this in `models/gemma2/_attention.py` instead since it's specific to the Gemma 2 model.

torchtune/modules/attention.py
```diff
@@ -306,3 +307,291 @@ def forward(
...
+        if self.softcapping is not None:
+            output = output / self.softcapping
+            output = torch.tanh(output)
+            output = output * self.softcapping
+
+        output = output + mask
+        output = F.softmax(output.float(), dim=-1).type_as(q)
+
+        # [batch_size, n_local_heads, input_len, head_dim]
+        output = torch.matmul(output, v)
```
ebsmothers, line 593: Overall this class looks good, but once we have the Flex Attention version available (which I anticipate will be more performant), we can consider something similar to what we have in our `MultiHeadAttention` class: parametrize this as `self._attention_call` and if the Flex Attention requirements are met (see [here](https://github.com/pytorch/torchtune/blob/4bbed4d4e245e990208f3389689291622a7af9df/torchtune/utils/_import_guard.py#L12-L16)) we can instead dispatch to that version for better perf.

recipes/configs/gemma2/27B_full.yaml
```diff
@@ -0,0 +1,95 @@
...
+shuffle: True
+
+# Model Arguments
+model:
+  _component_: torchtune.models.gemma2.gemma_27b
+
+checkpointer:
+  _component_: torchtune.training.FullModelHFCheckpointer
+  checkpoint_dir: /tmp/gemma2-27b/
+  checkpoint_files: [
```
ebsmothers, line 37: Can also use [this format](https://github.com/pytorch/torchtune/blob/4bbed4d4e245e990208f3389689291622a7af9df/recipes/configs/llama3_1/405B_qlora.yaml#L37-L39) now to keep things more concise here

torchtune/models/gemma2/_component_builders.py
```diff
@@ -0,0 +1,410 @@
...
+torchtune provides composable building blocks. Builder functions help
+stitch these building blocks into higher-level components. This design has
+two benefits:
+- The building blocks themselves are very flexible. For example, ``MultiHeadAttention``
+can take either nn.Linear or nn.LoRALinear for ``q_proj``.
+- Builder functions expose a set of configurable params which keep the constructors of
+the building blocks simple.
+"""
+
+class TanhSotfCapping(nn.Module):
```
ebsmothers, line 38:
```suggestion
class TanhSoftCapping(nn.Module):
```

torchtune/models/convert_weights.py
```diff
@@ -38,8 +38,10 @@
     "model.layers.{}.mlp.gate_proj.weight": "layers.{}.mlp.w1.weight",
     "model.layers.{}.mlp.up_proj.weight": "layers.{}.mlp.w3.weight",
     "model.layers.{}.mlp.down_proj.weight": "layers.{}.mlp.w2.weight",
-    "model.layers.{}.input_layernorm.weight": "layers.{}.sa_norm.scale",
-    "model.layers.{}.post_attention_layernorm.weight": "layers.{}.mlp_norm.scale",
+    "model.layers.{}.input_layernorm.weight": "layers.{}.sa_norm.scale",  # mlp_norm.scale -> looks like a previous bug here  # noqa
+    "model.layers.{}.post_attention_layernorm.weight": "layers.{}.sa_scale.scale",
```
Optimox, line 42: @ebsmothers what do you think about this line?
It felt like the correct mapping between hugging face's implementation and torchtune implementation would be:
`"model.layers.{}.post_attention_layernorm.weight": "layers.{}.sa_scale.scale"`
instead of this:
`"model.layers.{}.post_attention_layernorm.weight": "layers.{}.mlp_norm.scale",`

I was comparing this from hugging face: https://github.com/huggingface/transformers/blob/3a10c6192b12b328685dce89004f9f9b9c042d30/src/transformers/models/gemma2/modular_gemma2.py#L54

But when looking at gemma first implementation here it seems that there is a mismatch between the two versions of hugging face:
https://github.com/huggingface/transformers/blob/3a10c6192b12b328685dce89004f9f9b9c042d30/src/transformers/models/gemma/modular_gemma.py#L791

It looks like a bug from HF side which has been duplicated in torchtune, what do you think ? I will try to ask on hugging face repo and see how it goes.

Official Google's implementation is here: https://github.com/google/gemma_pytorch/blob/80881c2e6e797ef1913a4a705d4b40394791cc58/gemma/model.py#L446

Optimox: Actually I think the mismatch comes from the original repo with different naming for the same normalization:
https://github.com/google/gemma_pytorch/blob/80881c2e6e797ef1913a4a705d4b40394791cc58/gemma/model.py#L383

So there should not be any bug at all, but something has to be done in torchtune to handle both cases.
Should I create a specific weight converter dictionnary for gemma 2 in torchtune ?

ebsmothers: Thanks @Optimox for the investigation on this. I see what you're saying: the Hugging Face `post_attention_layernorm` is applied on the residual version in Gemma but on the raw self-attention outputs in Gemma 2, but uses the same name in both cases.

In that case, I think it's best to create a separate Gemma 2 model type. It'll use an almost-duplicate version of Gemma's key mapping, but since the same key gets mapped to different values in Gemma vs Gemma 2 I don't think there's any other way. Maybe just make sure to add a comment in the key mapping to explain that this is the only difference between the two.

torchtune/models/gemma2/_attention.py
```diff
@@ -0,0 +1,305 @@
...
+import torch.nn.functional as F
+from torch import nn
+from torchtune.modules.attention_utils import _MaskType
+from torchtune.modules.kv_cache import KVCache
+
+
+logger = logging.getLogger(__name__)
+
+
+class Gemma2Attention(nn.Module):
```
pbontrager, line 20: Since we support flex attention, which support soft capping, would it make sense to just force gemma2 users to use flex attention instead of implementing this module?

joecummings: Flex Attention is only supported on A100 or better, right? I don't think we can make the assumption that our users will have that.

Optimox: Hello everyone,

I just pushed a new commit which includes all changes discussed with @ebsmothers.
I also implemented a flex attention version but I could not make it work properly.


The default implementation (not using FlexAttention) seems to be working (I only launched the single lora pipeline, please see the attached logs [log_gemma2-2b-single-lora_1729498141.txt](https://github.com/user-attachments/files/17473404/log_gemma2-2b-single-lora_1729498141.txt)).


I would appreciate some help on the FlexAttention implementation. Here is why I am struggling.

If I run the following code on my A6000 GPU with torch 2.5:
```
import torch

from torch.nn.attention.flex_attention import (
    create_block_mask,
    flex_attention)


WINDOW_SIZE=None #None
CAPPING=50.
SCALE=12.


def get_gemma2_flex_score_mask(sliding_window_size, softcapping, query_pre_attn_scalar):
    
    def sliding_window_causal_mask(b, h, q_idx, kv_idx):
        """Causal mask and sliding window as proposed here:
        https://github.com/pytorch-labs/attention-gym/blob/main/examples/flex_attn.ipynb
        """
        causal_mask = q_idx >= kv_idx
        if sliding_window_size is None:
            # if no sliding window return causal mask
            return causal_mask
        else:
            windowed_mask = q_idx - kv_idx <= sliding_window_size

            return causal_mask & windowed_mask
    
    def soft_capping_with_scaling(score, b, h, q_idx, kv_idx):
        if query_pre_attn_scalar is None:
            # usual scaling included in FlexAttention
            score = score / softcapping
            score = torch.tanh(score) #tanh_approx(score)
            return score * softcapping
        else:
            score = score / softcapping * query_pre_attn_scalar**-0.5
            score = torch.tanh(score) #tanh_approx(score)
            return score * softcapping
    
    return sliding_window_causal_mask, soft_capping_with_scaling

# Compile the flex_attention function
flex_attention = torch.compile(flex_attention, dynamic=False)

B=4
H=8
S=117
D=256 #256

mask_mod, score_mod = get_gemma2_flex_score_mask(WINDOW_SIZE, CAPPING, SCALE)

query = torch.randn(
        B, H, S, D, device="cuda", dtype=torch.float16, requires_grad=True
    )
key = torch.randn(
    B, H, S, D, device="cuda", dtype=torch.float16, requires_grad=True
)
value = torch.randn(
    B, H, S, D, device="cuda", dtype=torch.float16, requires_grad=True
)
gradOut = torch.randn(B, H, S, D, device="cuda", dtype=torch.float16)


block_mask = create_block_mask(mask_mod=mask_mod,
                                   B=1,
                                   H=1,
                                   Q_LEN=S,
                                   KV_LEN=S,
                                   device=query.device)

out = flex_attention(
    query, key, value, score_mod=score_mod, block_mask=block_mask
)
print(out.shape)
```

The code runs fine if I don't compile the flex attention by commenting `flex_attention = torch.compile(flex_attention, dynamic=False)` but it raises this error otherwise:
`BackendCompilerFailed: backend='inductor' raised:
OutOfResources: out of resource: shared memory, Required: 114688, Hardware limit: 101376. Reducing block sizes or `num_stages` may help.`

So I disabled compilation and the code seems to be running but very very slowly (48s per iteration vs 1-2s on non flex implementation).

Maybe you could help me understand what is going on ?

joecummings: Tagging @RdoubleA and @felipemello1 for their thoughts. 

Just checking: which size Gemma-2 model are you testing with?

Optimox: The logs I shared are from gemma2 2B, the code snippet is independent of the gemma architecture it's just a toy example.
I am currently running the qlora single device pipeline with 9B (without flex attention), I'll share the logs tomorrow (I'll push the changes to recipe as there are typos on the output path etc).

ebsmothers: @Optimox just took a look at this. Your snippet here runs for me on an A100, so maybe it is an issue with compile on your hardware. I will investigate more, but in the meantime also tagging @SalmanMohammadi who I believe saw the same error previously.

ebsmothers: [Possibly relevant suggestion](https://github.com/pytorch/pytorch/issues/133254#issuecomment-2408710459)

Optimox: I already tried to change the kernels but I still can't compile.

Optimox: To add more information on this, if I set the kernel options as discussed [here](https://github.com/pytorch/pytorch/issues/133254#issuecomment-2408710459) I get a different error `LoweringException` and 
```
BackendCompilerFailed: backend='inductor' raised:
LoweringException: AssertionError: 
  target: flex_attention
  args[0]: TensorBox(StorageBox(
    InputBuffer(name='primals_1', layout=FixedLayout('cuda', torch.float16, size=[4, 8, 117, 256], stride=[239616, 29952, 256, 1]))
  ))
  args[1]: TensorBox(StorageBox(
    InputBuffer(name='primals_2', layout=FixedLayout('cuda', torch.float16, size=[4, 8, 117, 256], stride=[239616, 29952, 256, 1]))
  ))
  args[2]: TensorBox(StorageBox(
    InputBuffer(name='primals_3', layout=FixedLayout('cuda', torch.float16, size=[4, 8, 117, 256], stride=[239616, 29952, 256, 1]))
  ))
  args[3]: Subgraph(name='sdpa_score0', graph_module=<lambda>(), graph=None)
  args[4]: (TensorBox(StorageBox(
    InputBuffer(name='primals_4', layout=FixedLayout('cuda', torch.int32, size=[1, 1, 1], stride=[1, 1, 1]))
  )), TensorBox(StorageBox(
    InputBuffer(name='primals_5', layout=FixedLayout('cuda', torch.int32, size=[1, 1, 1, 1], stride=[1, 1, 1, 1]))
  )), TensorBox(StorageBox(
    InputBuffer(name='primals_6', layout=FixedLayout('cuda', torch.int32, size=[1, 1, 1], stride=[1, 1, 1]))
  )), TensorBox(StorageBox(
    InputBuffer(name='primals_7', layout=FixedLayout('cuda', torch.int32, size=[1, 1, 1, 1], stride=[1, 1, 1, 1]))
  )), TensorBox(StorageBox(
    InputBuffer(name='primals_8', layout=FixedLayout('cuda', torch.int32, size=[1, 1, 1], stride=[1, 1, 1]))
  )), TensorBox(StorageBox(
    InputBuffer(name='primals_9', layout=FixedLayout('cuda', torch.int32, size=[1, 1, 1, 1], stride=[2, 2, 1, 2]))
  )), TensorBox(StorageBox(
    InputBuffer(name='primals_10', layout=FixedLayout('cuda', torch.int32, size=[1, 1, 1], stride=[1, 1, 1]))
  )), TensorBox(StorageBox(
    InputBuffer(name='primals_11', layout=FixedLayout('cuda', torch.int32, size=[1, 1, 1, 1], stride=[2, 2, 1, 2]))
  )), 128, 128, Subgraph(name='sdpa_mask0', graph_module=<lambda>(), graph=None))
  args[5]: 0.0625
  args[6]: {'BLOCK_M': 64, 'BLOCK_N': 64, 'BLOCK_M1': 32, 'BLOCK_N1': 64, 'BLOCK_M2': 64, 'BLOCK_N2': 32, 'ROWS_GUARANTEED_SAFE': False, 'PRESCALE_QK': False, 'OUTPUT_LOGSUMEXP': True}
  args[7]: ()
  args[8]: ()
```

If I comment the compilation and use the new kernel options the code runs fine.

Optimox:
I have pushed changes to the recipes for 9b and 27b (typos in folders' name).
I also ran the single lora recipe for gemma2 9b, everything ran ok. nevertheless the loss seems better with the 2b model, maybe it's just because the larger model overfits more quickly.
[logs_gemma2-9b-lora-single_1729686341.txt](https://github.com/user-attachments/files/17507976/logs_gemma2-9b-lora-single_1729686341.txt)

torchtune/models/gemma2/_attention.py
mirceamironenco: Hi @Optimox for the forward pass the only relevant options are `BLOCK_M` and `BLOCK_N` IIRC. If you set your sequence length to a multiple of 128 (`S=128`, `S=8192`, etc.) and `kernel_options = {"BLOCK_M": 32, "BLOCK_N": 32}` the code you wrote should work on 3090s/4090s/etc. I think the issue is related to the way the default kernel options are chosen right now for flex [(e.g. for fwd pass)](https://github.com/pytorch/pytorch/blob/main/torch/_inductor/kernel/flex_attention.py#L594), which will treat A6000/3090/4090 as an A100, even though max shared memory is different:

```python
import torch
import triton

def get_shared_memory_config():
    device = torch.cuda.current_device()
    # Get current shared memory config
    current_shared = triton.runtime.driver.active.utils.get_device_properties(device)[
        "max_shared_mem"
    ]
    return current_shared
    
# 4090 = a6000 = 3090: 101376    
# A100: 166912
```
Hopefully this is improved in 2.5.1.

Optimox: Hello @mirceamironenco, thanks for your help, with S a multiple of 128 and `kernel_options = {"BLOCK_M": 32, "BLOCK_N": 32}` the compilation works indeed! Would this configuration also work for a100 ?
Could you explain what `max_share_mem` is here and why we should pick 32 ?

However, in my code snippet I used S=117 because this was the real size of my first batch when lauchning the training recipe. This means I should change the datasets and dataloaders to be sure that sequence length are padded to a multiple of 128.

mirceamironenco: - For now (pt 2.5.0), FlexAttention requires that all sequence lengths be a multiple of 128. IIUC this is being worked on.
- A100/H100 should work without any kernel_options, I think for (head_dim=256, dtype=bfloat16) it will use (32, 64) and (64, 32) respectively. Normally you could set `torch.compile(..., mode="max-autotune")` to have it try out a few options, but this is broken for flex_attention on 2.5.0. Unfortunately, for head_dim=256 3090s/4090s/etc will require kernel_options. Maybe this can be an environment variable until the defaults improve in pt.
- `max_shared_mem` is cuda shared memory which is used by the triton kernel generated by `torch.compile`. The amount of memory used by the kernel is dependent on e.g. dtype, block_N, block_M, num_stages. I just choose the next smaller multiple of 2 (64 -> 32).

ebsmothers: cc @drisspg @yanboliang in regards to [this comment above](https://github.com/pytorch/torchtune/pull/1835#discussion_r1815058279), is it possible to update the default kernel options in flex for better support of A6000/3090/4090s?

Optimox: With the given kernel options, flex attention can be compiled and the code runs (9b lora single device training). However, the code is terribly slow (29 tokens per second) and the loss turns to nan after one batch:
```
Step 1 | loss:87.6104507446289 lr:2.0000000000000003e-06 tokens_per_second_per_gpu:21.504354449638843 
Step 2 | loss:nan lr:4.000000000000001e-06 tokens_per_second_per_gpu:29.156293709460176 
```

I don't understand what I am doing wrong, the only obvious optimisation I see is to create one block mask for every layer while I am currently recreating the same block mask for every layer (line 593 in `gemma2/_attention.py`). Nevertheless, I do not think that this is the current bottleneck.

Wouldn't it be better to go with the simpler implementation for now and switch to FlexAttention when it will work on more GPUs? or at least leave the choice of computation to the final user and default to the classical implementation ?

recipes/configs/gemma2/27B_full.yaml
```diff
@@ -0,0 +1,72 @@
+# Config for multi-device full finetuning in full_finetune_distributed.py
+# using a gemma2 27B model
+#
+# This config assumes that you've run the following command before launching
+# this run:
+#   tune download google/gemma-2-27b --ignore-patterns "gemma-2-27b.gguf"  --hf-token <HF_TOKEN>
+#
+# To launch on 4 devices, run the following command from root:
+#   tune run --nnodes 1 --nproc_per_node 4 full_finetune_distributed --config gemma2/27B_full
```
ebsmothers, line 9: Did some quick math, I guess this will take at least 216GB total memory (54GB params + 54GB gradients + 108GB optimizer states for AdamW) , which means to run on 4 devices we'd need people to be using A100s. I wonder whether we can use an 8-bit optimizer + optimizer in backward to get us down to a more reasonable peak VRAM here

recipes/configs/gemma2/2B_lora_single_device.yaml
```diff
@@ -0,0 +1,113 @@
...
+  _component_: torchtune.models.gemma2.lora_gemma2_2b
+  lora_attn_modules: ['q_proj', 'k_proj', 'v_proj']
+  apply_lora_to_mlp: True
+  lora_rank: 64
+  lora_alpha: 128
+  lora_dropout: 0.0
+
+checkpointer:
+  _component_: torchtune.training.FullModelHFCheckpointer
+  checkpoint_dir: /tmp/gemma-2b/
```
ebsmothers, line 40:
```suggestion
  checkpoint_dir: /tmp/gemma-2-2b/
```

torchtune/models/gemma2/_attention.py
```diff
@@ -0,0 +1,610 @@
...
+        self.k_norm = k_norm
+        self.pos_embeddings = pos_embeddings
+
+        # gemma related parameters
+        self.sliding_window_size = sliding_window_size
+        self.softcapping = softcapping
+        if query_pre_attn_scalar is not None:
+            self.scaling = query_pre_attn_scalar**-0.5
+        else:
+            self.scaling = self.head_dim**-0.5
```
ebsmothers, line 133: I think you need to add `self.cache_enabled=False` here (then set it to True at the end of `setup_cache`), otherwise this will error out. But this is kind of a gotcha, it's not obvious that you need this. cc @SalmanMohammadi we should think about how to make this more obvious to someone adding their own attention layer

recipes/configs/gemma2/27B_full.yaml
felipemello1: does 8bit work with distributed?

torchtune/models/gemma2/_attention.py
SalmanMohammadi: Hmm, I added a comment to indicate why it's in the init (maybe @Optimox forked before then?)

```python
        # this flag indicates whether to update the kv-cache during forward
        # passes. when disabled, we can have the cache setup but still
        # perform normal forward passes
        self.cache_enabled = False
```

I agree we could use with a comment in `setup_caches` explaining that you actually need to do this if you'd like to use the caches you've just setup.

torchtune/models/gemma2/_attention.py
```diff
@@ -0,0 +1,610 @@
...
+            # [b, n_h, s, h_d]
+            k = k.transpose(1, 2)
+            v = v.transpose(1, 2)
+
+            # Normalize k
+            if self.k_norm is not None:
+                k = self.k_norm(k)
+
+            # Update key-value cache
+            if self.kv_cache is not None:
```
SalmanMohammadi, line 286:
```suggestion
            if self.kv_cache is not None and self.cache_enabled:
```
should complement the cache enabled stuff earlier to match the other attention module

torchtune/models/gemma2/_attention.py
ebsmothers: Yeah I see what you're saying.. I repro'd this on my end too so it is not a function of any custom kernel configs you're using. Let me look into this a bit more but in the meantime it seems like we shouldn't enable the flex version until we figure this out

recipes/configs/gemma2/27B_full.yaml
ebsmothers: Oh yeah duh.. there may be some issues with bitsandbytes optimizers on that front. I just tried out ao low-precision optimizers and it seems to work (though haven't resumed from intermediate checkpoint). Also there may be a compile dep there. Anyways if it's too much hassle we can consider it separately, don't wanna increase the scope of this already substantial PR more than necessary

torchtune/models/gemma2/_attention.py
Optimox: yes I think I forked before this change, will make the change tomorrow thank you!

recipes/configs/gemma2/27B_lora_single_device.yaml
```diff
@@ -0,0 +1,111 @@
...
+
+lr_scheduler:
+  _component_: torchtune.modules.get_cosine_schedule_with_warmup
+  num_warmup_steps: 10
+
+loss:
+  _component_: torchtune.modules.loss.CEWithChunkedOutputLoss
+
+# Fine-tuning arguments
+batch_size: 8
```
ebsmothers, line 63: Are we confident this'll fit on a single device?

recipes/configs/gemma2/2B_full.yaml
```diff
@@ -0,0 +1,74 @@
+# Config for multi-device full finetuning in full_finetune_distributed.py
+# using a gemma2 2B model
+#
+# This config assumes that you've run the following command before launching
+# this run:
+#   tune download google/gemma-2-2b --ignore-patterns "gemma-2-2b.gguf"  --hf-token <HF_TOKEN>
+#
+# To launch on 4 devices, run the following command from root:
+#   tune run --nnodes 1 --nproc_per_node 4 full_finetune_distributed --config gemma2/2B_full
```
ebsmothers, line 9: Maybe it's just me but when I try to run these distributed recipes I am hitting `AssertionError: FSDP requires named DeviceMesh dims for ND parallelism`. It looks to me like we are actually entering `_init_sharded_param` with a DTensor (see [here](https://github.com/pytorch/pytorch/blob/86b45bde196b976247f2771196f8c1c689dd66b7/torch/distributed/_composable/fsdp/_fsdp_param.py#L280)), which does not happen with our other recipes. Need to figure out why this would be happening

torchtune/models/gemma2/_component_builders.py
```diff
@@ -0,0 +1,430 @@
...
+            Default: 0.0
+        norm_eps (float): epsilon in RMS norms Default: 1e-6
+        rope_base (int): base for the rotary positional embeddings. Default: 10_000
+
+    Returns:
+        TransformerDecoder: Instantiation of gemma model.
+    """
+    rope = RotaryPositionalEmbeddings(dim=head_dim, max_seq_len=max_seq_len, base=rope_base)
+    
+    mlp = gemma_mlp(dim=embed_dim, hidden_dim=intermediate_dim)
```
ebsmothers, line 135: This needs to be inside the for loop

recipes/configs/gemma2/2B_full.yaml
ebsmothers: Ah I think I cracked the case. See [here](https://github.com/pytorch/torchtune/pull/1835/files#r1817437522)

Optimox: Big mistake, thank you for catching that!

recipes/configs/gemma2/27B_lora_single_device.yaml
Optimox: Changed batch size to 2 and accumulation to 8. What is the expected GPU? Is there a CI running everything? Otherwise I guess each user should be responsible to play with the batch to get something suitable for his GPU no ?

torchtune/models/gemma2/_attention.py
Optimox: Done!

torchtune/models/gemma2/_attention.py
Optimox: Done!

recipes/configs/gemma2/2B_lora_single_device.yaml
Optimox: Done!

recipes/configs/gemma2/27B_full.yaml
Optimox: What should I do here? Change something or expect users to change parameters according to their hardware ?

torchtune/models/gemma2/_attention.py
Optimox: I have updated the code to keep the flex attention implementation but disable it for now, until we have found a solution.

torchtune/models/gemma2/_component_builders.py
Optimox: Done!

recipes/configs/gemma2/27B_lora_single_device.yaml
SalmanMohammadi: Generally we try ship configs which we know will work on some common hardware configuration (see examples here https://github.com/pytorch/torchtune?tab=readme-ov-file#memory-and-training-speed), so users can maintain the expectation that they can get started without any painful OOMs. Then they are free to play with the configs. We should make sure this config works with e.g. 1xA1000 - let me know if you need a hand here.

recipes/configs/gemma2/27B_full.yaml
```diff
@@ -0,0 +1,72 @@
...
+# This config works only when the model is being fine-tuned on 2+ GPUs.
+
+
+# Tokenizer
+tokenizer:
+  _component_: torchtune.models.gemma.gemma_tokenizer
+  path: /tmp/gemma-2-27b/tokenizer.model
+
+# Dataset
+dataset:
```
SalmanMohammadi, line 25: Sorry to potentially be a pain in the ass here. We have parallel PR (https://github.com/pytorch/torchtune/pull/1872) which is helping standardize our configs and better expose the features we have. This means we always have `packed: False` in `dataset`, and `log_peak_memory_stats: True` and `compile: False` below, for every one of our configs.

Would it be annoying to ask if we could update these in the same way while we're here, please?

torchtune/models/gemma/__init__.py
```diff
@@ -27,6 +27,4 @@
     "lora_gemma_7b",
     "qlora_gemma_2b",
     "qlora_gemma_7b",
-    "gemma_hf_to_tune",
```
SalmanMohammadi, line 30: Good catch : )

torchtune/training/checkpointing/_utils.py
```diff
@@ -45,6 +45,7 @@ class ModelType(Enum):
 
     Attributes:
         GEMMA (str): Gemma family of models. See :func:`~torchtune.models.gemma.gemma`
+        GEMMA2 (str): Gemma family of models. See :func:`~torchtune.models.gemma2.gemma2`
```
SalmanMohammadi, line 48: nit
```suggestion
        GEMMA2 (str): Gemma 2 family of models. See :func:`~torchtune.models.gemma2.gemma2`
```

recipes/configs/gemma2/27B_lora_single_device.yaml
Optimox: @SalmanMohammadi I do not have easy access to a A100, would appreciate if someone could run the code for the 27B params model and let me know what batch size I should set.

recipes/configs/gemma2/27B_full.yaml
Optimox: Done I have updated all the configs to match the other PR!

recipes/configs/gemma2/2B_full.yaml
Optimox: The logs after this fix look much better than previously for the 9b single lora pipeline!
[log_gemma2-2b-single-lora_1729937021.txt](https://github.com/user-attachments/files/17533365/log_gemma2-2b-single-lora_1729937021.txt)

recipes/configs/gemma2/27B_lora_single_device.yaml
SalmanMohammadi: I'll have a quick look when we're ready to land. We can also reasonably mirror the batch size from the config of another similarly sized model already in the codebase.

torchtune/models/gemma2/_attention.py
drisspg: Hey sorry just catching up:

So 2.5 should not require multiple of 128 for sequence length. It is unfortunately pretty common for consumer gpus to hit the SharedMemory issue. I have a pr: https://github.com/pytorch/pytorch/pull/137959 to drop default block sizes but still need to debug the failing test.

For being slow, it is expected that the tanh instruction is very slow compared to the inline assembly variant: https://github.com/pytorch-labs/attention-gym/blob/36f8bd5ded5b3469f7892099590bb2405cc8f744/attn_gym/mods/softcapping.py#L92.


It is actually quite hard generically to know what what block sizes should be used since the amount of shared memory depends on the captured buffers. I am working on a better solution but that is going to take some time unfortunately

torchtune/models/gemma2/_attention.py
```diff
@@ -0,0 +1,626 @@
...
+_SUPPORTS_FLEX_ATTENTION = False
+
+if _SUPPORTS_FLEX_ATTENTION:
+    from torch.nn.attention.flex_attention import create_block_mask
+    from torchtune.models.gemma2._attention_utils import (
+        compile_friendly_flex_attention_with_score_and_block,
+        flex_causal_sliding_window,
+        flex_tanh_soft_capping_with_scaling,
+    )
+logger = logging.getLogger(__name__)
```
joecummings, line 28: nit: Why is this style of logger getting proliferated? We should be calling `get_logger` from our utils.

Optimox: It's just a copy paste on my side, let me know if you want me to change that on this PR.

ebsmothers:
Hi @Optimox sorry for the delay here. Given that the flex attention version is still not working properly, how do you feel about pulling it out of this PR? Then we can revisit in a follow-up. For context we are going to be cutting a release soon (targeting code freeze tomorrow) so don't want to block getting this in on something that we can address in a follow-up. Let me know if this makes sense to you.

Optimox:
@ebsmothers yes no problem! What is the best way of handling this? Adding a new commit deleting the flex attention part of this branch ? Or creating a new PR without the flex attention part?

ebsmothers:
@Optimox honestly whatever is easiest for you. I imagine just a commit deleting the flex code would be simplest, but feel free to do whatever makes sense to you!

Optimox:
@ebsmothers I have removed the flex attention implementation from the code, let me know if there are still other changes to make!

recipes/configs/gemma2/27B_full.yaml
ebsmothers: Sorry missed this comment before now. I think it's fine to leave this as you have it and revisit these details in a later PR

torchtune/models/gemma2/_attention.py
ebsmothers: You can just change to `torchtune.utils.get_logger`, but no strong preference here. Either way we should clean up other usages in a follow-up

torchtune/models/gemma2/_convert_weights.py
```diff
@@ -0,0 +1,132 @@
...
+import torch
+
+from torchtune.models.convert_weights import get_mapped_key
+
+"""
+Gemma 2 and Gemma original implementations share different normalization but with
+the same name, so it is mandatory to differentiate their state dict in order to map
+correctly the different weights.
+They are essentially the same except for "model.layers.{}.post_attention_layernorm.weight" key.
+See discussion here: https://github.com/pytorch/torchtune/pull/1835#discussion_r1803410251
```
ebsmothers, line 18: Thanks for documenting this

docs/source/api_ref_models.rst
```diff
@@ -320,6 +320,37 @@ To download the Gemma 7B model:
...
+
+    tune download google/gemma-2-<MODEL_SIZE>b --ignore-patterns "gemma-2-<MODEL_SIZE>b.gguf"  --hf-token <HF_TOKEN>
+
+
+.. autosummary::
+    :toctree: generated/
+    :nosignatures:
+
+    gemma2.gemma2
+    gemma2.lora_gemma
```
ebsmothers, line 342:
```suggestion
    gemma2.lora_gemma2
```

torchtune/models/gemma2/_attention.py
```diff
@@ -0,0 +1,320 @@
...
+
+            sliding_mask = torch.triu(
+                all_ones, -1 * self.sliding_window_size + 1
+            ) * torch.tril(all_ones, self.sliding_window_size - 1)
+            mask = torch.where(sliding_mask == 1, mask, -2.3819763e38)
+
+        if self.softcapping is not None:
+            output = output / self.softcapping
+            output = torch.tanh(output)
+            output = output * self.softcapping
```
ebsmothers, line 310: Can we add code comments explaining sliding window and the softcapping? (Also one for the magic value in the `torch.where` line wouldn't hurt)

torchtune/models/gemma2/_attention.py
```diff
@@ -0,0 +1,320 @@
...
+            # Normalize k
+            if self.k_norm is not None:
+                k = self.k_norm(k)
+
+            # Update key-value cache
+            if self.kv_cache is not None and self.cache_enabled:
+                k, v = self.kv_cache.update(k, v)
+
+        q.mul_(self.scaling)
+        output = torch.matmul(q, k.transpose(2, 3))
```
ebsmothers, line 288: Let's add shape comments here too

torchtune/models/gemma2/_attention.py
```diff
@@ -0,0 +1,320 @@
...
+                "Key value caches are not setup. Call ``setup_caches()`` first."
+            )
+        self.kv_cache.reset()
+
+    def forward(
+        self,
+        x: torch.Tensor,
+        y: Optional[torch.Tensor] = None,
+        *,
+        mask: Optional[_MaskType] = None,
```
ebsmothers, line 171: Have you run any of the configs with `packed=True` (i.e. when mask is a `BlockMask`)?

ebsmothers (commented):
A few more comments and questions but overall this is looking great!

joecummings:
@Optimox Any chance you can get to these last comments today?

torchtune/models/gemma2/_attention.py
Optimox: I haven't run a full training but I've checked that the code does not through any error.

torchtune/models/gemma2/_attention.py
Optimox: About this part of the code, I actually followed blindly the official pytoch implementation from Google [here](
    https://github.com/google/gemma_pytorch/blob/80881c2e6e797ef1913a4a705d4b40394791cc58/gemma/model.py#L213)

I am not sure why they used this magic number instead of `-torch.inf` ...

About the `sliding_mask` I am now worried that something is wrong here because of the way I defined the causal mask...

```
s_x = 10
sliding_window_size = 5
mask = torch.tril(
                torch.ones(
                    size=(s_x, s_x),
                    dtype=torch.bool,
                )
            )
print(mask)

all_ones = torch.ones_like(mask)
sliding_mask = torch.triu(
                all_ones, -1 * sliding_window_size + 1
            ) * torch.tril(all_ones, sliding_window_size - 1)
mask = torch.where(sliding_mask == 1, mask, -2.3819763e38)
print(mask)
```
The final mask here  does not seem to be causal anymore, and sliding future tokens are now accessible somehow...

Something like the following would seem better to me but is there a difference in the way masks are defined in gemma2 official code and torchtune?

```
s_x = 10
sliding_window_size = 5
mask = torch.tril(
                torch.ones(
                    size=(s_x, s_x),
                    dtype=torch.bool,
                )
            )

mask = torch.where(mask==0, -torch.inf, 1)
print(mask)

all_ones = torch.ones_like(mask)
sliding_mask = torch.triu(
                all_ones, -1 * sliding_window_size + 1
            ) * torch.tril(all_ones, sliding_window_size - 1)
mask = torch.where(sliding_mask == 1, mask, -torch.inf)
print(mask)
```

This seems concerning... what do you think ?

recipes/configs/gemma2/2B_full.yaml
Optimox: @ebsmothers aren't the losses too low? could it be because of the (non) causal sliding window attention problem ?

docs/source/api_ref_models.rst
```diff
@@ -320,6 +320,37 @@ To download the Gemma 7B model:
     gemma.gemma_tokenizer
 
 
+gemma2 :
+-----
```
joecummings, line 324:
```suggestion
--------
```

torchtune/models/gemma2/_attention.py
ebsmothers: @Optimox yeah good catch, that first mask definitely does not look right. I'm not that familiar with the official implementation, but looks like they are treating it as an additive mask [here](https://github.com/google/gemma_pytorch/blob/80881c2e6e797ef1913a4a705d4b40394791cc58/gemma/model.py#L321). In that case there should definitely not be anything that's not -torch.inf (or a very large negative number) above the diagonal.

This seems like a bug to me, maybe you can open an issue on the gemma_pytorch repo to confirm? Your second implementation looks correct to me.

SalmanMohammadi: Sorry to add another kitchen to the cook here so late. I thought I'd share some of my findings from the last time I worked on masking + SDPA. I'd reccomend checking this PR out https://github.com/pytorch/pytorch/pull/133882 and the linked issues. 

TLDR; this is the correct approach to use the attention mask like a "bias" by adding a very large negative to the q/k tensors. However, using `-inf` as this negative number has been shown to produce NaN gradients for some rare corner cases (e.g. when an entire row is masked out). In Transformers, the approach is to use something like `torch.finfo(dtype).min` [1] (which is maybe where the original magic number is coming from?)
```python
import torch
x = torch.Tensor([[[float("-inf"), float("-inf"), float("-inf"), 10]]])
softmax = torch.nn.Softmax(dim=-1)
softmax(x)
# tensor([[[nan, nan, nan]]])
dtype = torch.bfloat16
min_value = torch.finfo(dtype).min
# -3.3895313892515355e+38 - on MPS, this will vary depending on the hardware you're using
x = torch.Tensor([[[min_value, min_value, min_value]]])
softmax = torch.nn.Softmax(dim=-1)
softmax(x)
# tensor([[[0.3340, 0.3340, 0.3340]]]) 
```



Aside, as of torch 2.5 this is handled [internally](https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/native/transformers/attention.cpp#L669) slightly differently. `-inf` is used in the mask, softmax is performed, but then any rows in the original tensor which have entirely masked out rows are explicitly set to zero.



[1] Transformers follows this approach for their [Gemma2 implementation](https://github.com/huggingface/transformers/blob/33868a057c02f0368ba63bd1edb746be38fe3d90/src/transformers/models/gemma2/modeling_gemma2.py#L545). However, this apparently still causes issues for some dtypes so it's also been suggested to use `torch.finfo(dtype).min / 2` - see https://github.com/huggingface/transformers/issues/32390, or to just attend to all tokens in a row containing only padding tokens equally (https://github.com/huggingface/transformers/commit/e22d913e8b9717d6a51b2e67e08c2f3bd9577a48), but I'm not 100% about how this works.

Optimox:
I have pushed some minimal changes which take into account the fact that we are not using spda with gemma2 so the masks must be converted from boolean to True -> 0s and False -> -inf as discussed [here](https://github.com/google/gemma_pytorch/issues/73).

There is one last identified concern on my side, @ebsmothers I think I answered somewhere that packed dataset was working but actually it is not.
The first reason is that when I run the code with torch 2.5, then flex attention is considered to be available, so block masks arrive in flex attention format which is a problem without a flex attention implementation for gemma2.

If I manually disable flexattention (I don't know how to disable it locally or automatically) then I had a broadcasting issue which has been solved in my latest commit.

So currently the code won't work for packed dataset for  torch > 2.5.

torchtune/models/gemma2/_attention.py
ebsmothers: Wait so I am a bit confused here (this is also in reference to the issue opened by @Optimox on the gemma_pytorch repo). There are two separate questions here, right?

(1) Should values above the diagonal be unmasked? 
(2) What is the right masked value for an additive mask?

I think both the gemma_pytorch discussion and @SalmanMohammadi's comments address (2), and that I am not so worried about. But I think (1) is more fundamental, and I can directly copy-paste the [snippet from gemma_pytorch](https://github.com/google/gemma_pytorch/blob/main/gemma/model.py#L312-L316) to get this:

```
import torch

s_x = 5
sliding_window_size = 3
mask = torch.tril(
                torch.ones(
                    size=(s_x, s_x),
                    dtype=torch.bool,
                )
            )

all_ones = torch.ones_like(mask)
sliding_mask = torch.triu(
    all_ones, -1 * sliding_window_size + 1
) * torch.tril(all_ones, sliding_window_size - 1)
mask = torch.where(sliding_mask == 1, mask, -2.3819763e38)
print(mask)
...

tensor([[ 1.0000e+00,  0.0000e+00,  0.0000e+00, -2.3820e+38, -2.3820e+38],
        [ 1.0000e+00,  1.0000e+00,  0.0000e+00,  0.0000e+00, -2.3820e+38],
        [ 1.0000e+00,  1.0000e+00,  1.0000e+00,  0.0000e+00,  0.0000e+00],
        [-2.3820e+38,  1.0000e+00,  1.0000e+00,  1.0000e+00,  0.0000e+00],
        [-2.3820e+38, -2.3820e+38,  1.0000e+00,  1.0000e+00,  1.0000e+00]])
```

This demonstrates that there are values above the diagonal that are unmasked, no? (I will also reopen the issue on there just to confirm I am not missing something here)

Optimox: What I understood is that there are two ways of defining masks (in bias mode):
- 0 for reachable, -inf (or very large negative number) for unreachable -> this is the expected input mask for gemma_pytorch implementation
- bolean mask (True for reachable, False otherwise) + [torch.nn.functional.scaled_dot_product_attention](https://pytorch.org/docs/stable/generated/torch.nn.functional.scaled_dot_product_attention.html) which internally switches from boolean to the first implementation `if attn_mask.dtype == torch.bool:
            attn_bias.masked_fill_(attn_mask.logical_not(), float("-inf"))`

So here, since all previous models from torchtune used torch sdpa there was a mismatch between both implementation, that is why I added this conversion in the latest [changes](https://github.com/pytorch/torchtune/pull/1835/commits/7adb55a4da1b07bdc1dc56aff79d6446a96aee9e). This conversion should also work for block mask defined as boolean mask currently in torchtune code.

torchtune/models/gemma2/_attention.py
Optimox: See my latest comment: packed=True won't work at the moment!

recipes/configs/gemma2/2B_full.yaml
ebsmothers: Sorry I missed this comment before. I am gonna run some of your configs on my end now so will get back to you

recipes/configs/gemma2/27B_qlora_single_device.yaml
```diff
@@ -0,0 +1,115 @@
...
+# Dataset
+dataset:
+  packed: False # Set to true for great speed ups
+  _component_: torchtune.datasets.alpaca_dataset
+seed: null
+shuffle: True
+
+# Model Arguments
+model:
+  _component_: torchtune.models.gemma2.qlora_gemma_27b
```
ebsmothers, line 32:
```suggestion
  _component_: torchtune.models.gemma2.qlora_gemma2_27b
```

recipes/configs/gemma2/27B_qlora_single_device.yaml
```diff
@@ -0,0 +1,115 @@
...
+  lora_rank: 64
+  lora_alpha: 128
+  lora_dropout: 0.0
+
+checkpointer:
+  _component_: torchtune.training.FullModelHFCheckpointer
+  checkpoint_dir: /tmp/gemma-2-27b/
+  checkpoint_files:
+    filename_format: model-{}-of-{}.safetensors
+    max_filename: 00024
```
ebsmothers, line 44: I think all usages of max_filename should look like this instead
```suggestion
    max_filename: "00024"
```

torchtune/models/gemma2/_attention.py
ebsmothers: Thanks, I also caught up with @SalmanMohammadi offline about this. It seems like the initial definition of `mask` in the code snippet I shared in my last comment did not match what they do in gemma_pytorch, so that was a misunderstanding on my part. I think your approach makes sense, will look at the code more closely to confirm the BlockMask case (though I guess if we're not supporting packed or flex yet it doesn't matter?)

ebsmothers:
@Optimox thanks for your patience in the review process here. Is [this comment](https://github.com/pytorch/torchtune/pull/1835#discussion_r1825035796) about unusually low losses still a concern? I ran a few of your configs on my end and the loss does increase pretty dramatically (though I also don't have a baseline). For reference here are some loss curves:

Screenshot 2024-11-07 at 11 55 01 AM

Otherwise regarding the fact that packed is not yet supported, it seems like we should raise an error somewhere if that's the case. Maybe inside the attention if we receive a `BlockMask` there? (Though open to other thoughts you have on this)

Optimox:
@ebsmothers yes the comment was before the fix of the slinding window attention mask.

I have added a NotImplementedError for the BlockMasks and made the small fixes in the yaml files you pointed.

Let me know if you think some other changes should be made!

ebsmothers (approved):
Thank you for this great new feature! And thanks for your patience and diligence throughout the review process. Very excited that we're now able to support Gemma 2 in torchtune.

torchtune/models/gemma2/_attention.py
ssharpe42: Have you all figured out a solution to the out of resource: shared memory? Seems like any large hidden dim >=128 causes issues for me.

Andrei-Aksionov: Hey there 👋 

I've been tinkering with the code for quite a bit, trying to find a solution that will work.
And I've noticed an interesting behavior. If to keep the code snippet from the above untouched, but just change the sequence length (S), then (on T4 in Google Collab) with BLOCK_N=32, BLOCK_M=32 and S in range `[1,32]` and `[128, ...]` the code works, but not with S in range `(32,128)`.
Which is odd. Why does the code work with a seq length of, let's say, 181, but not 96? What's so special about `32<S<128`?

PR #1836, opened 2024-10-15
Remove pad_max_tiles in CLIP

pbontrager:
#### Context
- [ ] add a new feature
- [x] fix a bug
- [x] other (please add here)

Currently we have pad_max_tiles in the CLIP transform that will pad the image to 4 tiles. The CLIP model doesn't mask out the padding tiles, so by default we always added padding tiles in case the model relied on these extra tokens. Though in testing, I find the model completely ignores the padding tiles and it's not necessary to include them unless needed for the batch. Apart from this, I found that doing the pad_max_tiles in the CLIP transform instead of padded_collate_tiled_images_and_mask leads to a subtle bug where the cross attention mask is not aware, downstream, of which tiles are padding and should be masked and which shouldn't be.

#### Changelog
* remove pad_max_tiles from CLIPTransform
* add pad_max_tiles to padded_collate_tiled_images_and_mask
	* doing padding during collation (how we do for all other padding) ensures mask transforms are accurate
* Set pad_max_tiles to None by default
	* not padding more masks than necessary enables much faster training

#### Test plan

This shows three runs for `tune run full_finetune_single_device --config llama3_2_vision/11B_full_single_device`. The "original" run is what's currently in main, the "new_pad_to_4" is with the fixed padding but adding extra padding tiles, and the "new_pad_to_batch" only pads to the max tiles in the batch. All three of these have the same loss, suggesting that the padding isn't important for model quality, but pad_to_batch gets almost double the qps on the ocr dataset since the cross attention sequence lengths can be much shorter.

Screenshot 2024-10-15 at 1 57 04 PM

### DIFF
--- a/recipes/configs/llama3_2_vision/11B_lora_single_device.yaml
+++ b/recipes/configs/llama3_2_vision/11B_lora_single_device.yaml
@@ -32,6 +32,7 @@ tokenizer:
   _component_: torchtune.models.llama3_2_vision.llama3_2_vision_transform
   path: /tmp/Llama-3.2-11B-Vision-Instruct/original/tokenizer.model
   image_size: 560
+  max_seq_len: 8192
 
 # Checkpointer
 checkpointer:
--- a/torchtune/data/_collate.py
+++ b/torchtune/data/_collate.py
@@ -222,6 +222,7 @@ def padded_collate_tiled_images_and_mask(
     padding_idx: int = 0,
     ignore_idx: int = CROSS_ENTROPY_IGNORE_IDX,
     pad_direction: str = "right",
+    pad_max_tiles: Optional[int] = None,
     pad_max_images: Optional[int] = None,
 ) -> Dict[str, torch.Tensor]:
     """Pad a batch of text sequences, tiled image tensors, aspect ratios,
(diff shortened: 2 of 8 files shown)


recipes/configs/llama3_2_vision/11B_lora_single_device.yaml
```diff
@@ -32,6 +32,7 @@ tokenizer:
   _component_: torchtune.models.llama3_2_vision.llama3_2_vision_transform
   path: /tmp/Llama-3.2-11B-Vision-Instruct/original/tokenizer.model
   image_size: 560
+  max_seq_len: 8192
```
joecummings, line 35: Why are you limiting to 8k?

torchtune/data/_collate.py
```diff
@@ -355,6 +359,13 @@ def padded_collate_tiled_images_and_mask(
         for sample in batch
         for image in sample["encoder_input"]["images"]
     )
+    if pad_max_tiles is not None:
+        if pad_max_tiles < max_num_tiles:
```
joecummings, line 363: `max_num_tiles` is such a misleading name.

joecummings (approved):
Two comments but otherwise fine.

PR #1847, opened 2024-10-15
Activation offloading for fullfinetuning + fix tied embedding

felipemello1:
#### Context
- [x] add a new feature
- [x] fix a bug
- [ ] other (please add here)

#### Changelog
- Added activation offloading for single/distributed full finetuning
- Found that we didnt address the tied embedding issue. Fixed it by adding a dummy nn.Module so that tensor hooks would work with it
- Moved the context manager logic to a function
- Made the recipes more aligned

#### Test plan
image
image
image
image


recipes/full_finetune_single_device.py
```diff
@@ -517,13 +557,15 @@ def _setup_data(
             sampler=sampler,
             # dropping last avoids shape issues with compile + flex attention
             drop_last=True,
-            collate_fn=partial(
```
felipemello1, line 520: didnt touch this. It was pre commit hook

recipes/lora_finetune_distributed.py
```diff
@@ -151,16 +146,6 @@ def __init__(self, cfg: DictConfig) -> None:
...
 
-        # training attributes
-        self._enable_activation_checkpointing = cfg.enable_activation_checkpointing
-        self._enable_activation_offloading = cfg.get(
-            "enable_activation_offloading", False
-        )
-        if self._enable_activation_offloading and self._device.type != "cuda":
-            raise RuntimeError(
-                "enable_activation_offloading should only be enabled for training on CUDA"
-            )
```
felipemello1, line 162: removed from init. This is handled in setup_model

recipes/full_finetune_distributed.py
```diff
@@ -222,9 +239,11 @@ def setup(self, cfg: DictConfig) -> None:
...
         self._optimizer = self._setup_optimizer(
             cfg_optimizer=cfg.optimizer,
-            opt_state_dict=checkpoint_dict[training.OPT_KEY]
-            if self._resume_from_checkpoint
-            else None,
+            opt_state_dict=(
+                checkpoint_dict[training.OPT_KEY]
+                if self._resume_from_checkpoint
+                else None
+            ),
```
felipemello1, line 246: pre commit hook

recipes/full_finetune_distributed.py
```diff
@@ -509,13 +549,15 @@ def _setup_data(
...
-            else padded_collate_packed,
+            collate_fn=(
+                partial(
+                    collate_fn,
+                    padding_idx=self._tokenizer.pad_id,
+                    ignore_idx=self._loss_fn.ignore_index,
+                )
+                if not packed
+                else padded_collate_packed
+            ),
```
felipemello1, line 560: pre commit hook

recipes/full_finetune_distributed.py
```diff
@@ -440,6 +460,26 @@ def _is_layer_fqn(s: str) -> bool:
...
+                )
+            if not enable_activation_checkpointing:
+                raise RuntimeError(
+                    "enable_activation_offloading should only be True when enable_activation_checkpointing is True"
+                )
+        elif enable_activation_checkpointing:
+            log.info(
+                "Hint: enable_activation_checkpointing is True, but enable_activation_offloading isn't. "
+                "Enabling activation offloading should reduce memory further."
+            )
```
felipemello1, line 477: the only think that i dont like about this is that it could fail much faster if we added it to the __init__. But i like that the checks are near the code where it matters. So i am not sure which one to pick.

recipes/full_finetune_distributed.py
```diff
@@ -440,6 +460,26 @@ def _is_layer_fqn(s: str) -> bool:
...
+
+        self.activations_handling_ctx = training.get_act_offloading_ctx_manager(
+            model, enable_activation_offloading
+        )
```
felipemello1, line 481: added function to handle NoOp / getting context

recipes/full_finetune_distributed.py
```diff
@@ -440,6 +460,26 @@ def _is_layer_fqn(s: str) -> bool:
...
 
+        # activation checkpointing/offloading
+        if enable_activation_offloading:
+            if self._device.type != "cuda":
+                raise RuntimeError(
+                    "enable_activation_offloading should only be True for training on CUDA"
+                )
+            if not enable_activation_checkpointing:
+                raise RuntimeError(
+                    "enable_activation_offloading should only be True when enable_activation_checkpointing is True"
```
felipemello1, line 471: no point in running offloading when AC is off. Its extremely slow.

torchtune/modules/tied_linear.py
```diff
@@ -9,13 +9,33 @@
 import torch.nn.functional as F
 
 
+class Linear(nn.Module):
```
felipemello1, line 12: docstring explains why i had to add this

felipemello1:
Will add the option to all the configs for full finetuning

recipes/full_finetune_distributed.py
janeyx99: I think failing much faster is better

torchtune/modules/tied_linear.py
ebsmothers: Sorry I'm a bit confused by this change on two fronts: 

(1) Does this not change the key names in the state dict?
(2) Now that we have a module again, how do we not wind up right back where we started with to_empty?

recipes/configs/llama3_2_vision/11B_full_single_device.yaml
```diff
@@ -68,6 +68,7 @@ device: cuda
 
 # Memory management
 enable_activation_checkpointing: True
+enable_activation_offloading: False  # True reduces memory
```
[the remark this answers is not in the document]
felipemello1: i didnt include it in my tests. Let me do it tomorrow.

recipes/full_finetune_distributed.py
```diff
@@ -44,13 +44,25 @@ class FullFinetuneRecipeDistributed(FTRecipeInterface):
...
 
+        - Activation Offloading. This can be controlled using the ``enable_activation_offloading``
+            flag. Activation offloading is a technique similar to activations checkpointing that helps
+            reduce the memory footprint to prevent OOMs on CUDA and enable bigger batches. Where activations
+            checkpointing drops the activation in the forward to recompute it later in the backward,
+            activations offloading will drop the activation in the forward to the CPU and bring it
+            back during the backward pass. As always, there is a tradeoff--these savings in memory can
+            come at the cost of training performance and CPU resources. To recover some runtime cost,
+            we've added an option to enable offloading on a different stream to permit overlapping with
+            the computation. This option is currently only available on PyTorch nightly 2.5.0.dev20240907
```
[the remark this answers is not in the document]
felipemello1: agreed

recipes/lora_finetune_distributed.py
felipemello1: I added it back after this:

image

torchtune/modules/tied_linear.py
felipemello1: (1) No. The Linear nn.Module doesnt have any weights. The weight is passed in the forward only. 
```
Class Linear(nn.Module):
      def forward(x, weight):
             return F.Linear(x, weight)
```

TiedLinear is still a regular python class, and the key name is still model.output.tied_module.weight --> translates to --> model.TiedLinear.tok_embedding.weight

(2) i tested it with FSDP, and confirmed that the memory pointers are the same in model.tok_embedding.weight and model.output.weight. So things are fine. Is that what you were referencing too?

tests/recipes/test_full_finetune_distributed.py
```diff
@@ -34,6 +34,7 @@ def _get_test_config_overrides(self):
             "batch_size=4",
             "dtype=fp32",
             "enable_activation_checkpointing=False",
+            "enable_activation_offloading=False",
```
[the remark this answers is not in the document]
felipemello1: we should. I will see if i can quickly add it tomorrow

torchtune/training/_activation_offloading.py
```diff
@@ -327,3 +331,73 @@ def noop(tensor):
...
+            log.warning(
+                "During activation offloading, no output head was detected. "
+                "If your model has an output head, it will be offloaded. "
+                "This usually greatly slows training, given the large vocabulary size. "
+                "To change this behavior, set your output head as model.output and make it "
+                "an nn.Module."
+            )
+
+    else:
+        activations_handling_ctx = contextlib.nullcontext()
```
felipemello1, line 401: this covers all of our cases, but it looks horrible. I think its fine to ship like it, but in the future change it to identify tensor size, and if larger than threshold, make it a non op

tests/recipes/test_lora_finetune_distributed.py
```diff
@@ -173,6 +175,7 @@ def test_training_state_on_resume(
             resume_from_checkpoint=True \
             metric_logger.filename={log_file} \
             enable_activation_checkpointing=True \
+            enable_activation_offloading=True \
```
felipemello1, line 178: I enabled it for some tests in this file only. It tests lora and distributed. Ideally, we should have it vs many other parameters, like compile, vision and tiedembeddings. I wont address those in this PR. This should be part of the testing improvement, IMO

torchtune/modules/tied_linear.py
```diff
@@ -26,12 +46,13 @@ class TiedLinear:
...
     def __init__(self, tied_module: nn.Module):
         self.tied_module = tied_module
+        self.linear = Linear()
         if not hasattr(tied_module, "weight"):
             raise AttributeError(
                 "Provided module does not have attribute 'weight'. Please check your tied_module."
             )
 
-    def __call__(self, x: torch.Tensor) -> torch.Tensor:
+    def __call__(self, x: torch.Tensor):
```
joecummings, line 55: Why did you remove the type hint?

recipes/full_finetune_single_device.py
```diff
@@ -569,7 +613,8 @@ def _loss_step(self, batch: Dict[str, torch.Tensor]) -> torch.Tensor:
         # Shape [b, s], needed for the loss not the model
         labels = batch.pop("labels")
 
-        logits = self._model(**batch)
+        with self.activations_handling_ctx:
```
SalmanMohammadi, line 616: Curiosity brings me here. If I gather correctly, every model forward pass for which which we want to offload activations for needs to sit inside this context manager?

felipemello1: correct!

torchtune/training/_activation_offloading.py
```diff
@@ -345,3 +352,82 @@ def noop(tensor):
...
+            )
+            # if isinstance(model.decoder, nn.Module):
+            #     model.decoder.output.register_forward_pre_hook(
+            #         lambda *args: noop_ctx.__enter__()
+            #     )
+            #     model.decoder.output.register_forward_hook(
+            #         lambda *args: noop_ctx.__exit__(), always_call=True
+            #     )
+            #     output_head_detected = True
+
```
felipemello1, line 420: this needs debugging in a follow up pr

joecummings: haha can you remove the commented out code?

joecummings (approved):
We're not savages, remove the commented out code.

Issue #1624, opened 2024-09-19
Qwen 2.5 is here, Request for adding a model


vanshnawander:
The Qwen team has released qwen 2.5 base, coder, math models. They seem very promising. Requesting Team to add this model in torchtune.
```[tasklist]
### Tasks
```

ebsmothers:
Hi @vanshnawander thanks for creating the issue. I took a quick look, seems like the architecture should be mostly unchanged. In that case we can reuse the model builders from Qwen2 [here](https://github.com/pytorch/torchtune/tree/main/torchtune/models/qwen2). It should be similar for the tokenizer but there are additional special tokens, along with minor changes to the chat template. I'm going to tag this as community help wanted since we'd love to have someone help out with this. cc @fyabc who added Qwen2 in case I'm missing any important details here.

fyabc:
@ebsmothers Hi, thank you for you comments! Qwen2.5 only need to update tokenizer with new special tokens and chat template. I will take a look at this.

krammnic:
As there is no updates here, I can take a look at it.

PR #1863, opened 2024-10-18
Qwen2.5

calvinpelletier:
#### Context
- [x] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

Issue #1624 

#### Changelog
- updated the list of special tokens
- created a prompt template
- added model builders
- added a unit test for Qwen2.5 tokenization
- moved special token logic from template to tokenizer
- created separate qwen2/2.5 tokenizers
- separated base and instruct model builders
- created various configs

#### Test plan
Checked tokenizer and chat template parity with the current official Qwen2.5 implementation (huggingface). 

Ran Qwen2.5 finetunes
- did a sweep through each Qwen2.5 config to make sure each runs without issue and starts to improve the loss
- comparison run between Qwen versions: Qwen2.5 7B lora single-gpu: https://api.wandb.ai/links/torchtune/vraolx0j (vs Qwen2 equivalent: https://api.wandb.ai/links/torchtune/h6l8s9ou ) 

Checklist
- [x] include relevant commands and any other artifacts in this summary (see above for links to loss curves)


recipes/configs/qwen2_5/3B_full.yaml
```diff
@@ -0,0 +1,75 @@
...
+#   tune run --nnodes 1 --nproc_per_node 4 full_finetune_distributed --config qwen2_5/3B_full checkpointer.checkpoint_dir=<YOUR_CHECKPOINT_DIR>
+#
+# This config works best when the model is being fine-tuned on 2+ GPUs.
+# Single device full finetuning requires more memory optimizations.
+
+# Tokenizer
+tokenizer:
+  _component_: torchtune.models.qwen2.qwen2_tokenizer
+  path: /tmp/Qwen2.5-3B-Instruct/vocab.json
+  merges_file: /tmp/Qwen2.5-3B-Instruct/merges.txt
```
pbontrager, line 23: We use underscores instead of "." in our naming

recipes/configs/qwen2_5/3B_full.yaml
```diff
@@ -0,0 +1,75 @@
...
+
+# Dataset
+dataset:
+  _component_: torchtune.datasets.alpaca_cleaned_dataset
+seed: null
+shuffle: True
+
+# Model Arguments
+model:
+  _component_: torchtune.models.qwen2.qwen2_3b
```
pbontrager, line 34: Shouldn't we be adding a qwen2_5 folder the same as we do for llama?

torchtune/models/qwen2/_prompt_template.py
```diff
@@ -0,0 +1,117 @@
...
+
+
+DEFAULT_SYS_PROMPT = (
+    "You are Qwen, created by Alibaba Cloud. You are a helpful assistant."
+)
+TOOL_INSTRUCTION_START = "\n\n# Tools\n\nYou may call one or more functions to assist with the user query.\n\nYou are provided with function signatures within <tools></tools> XML tags:\n<tools>"  # noqa: B950
+TOOL_INSTRUCTION_END = '</tools>\n\nFor each function call, return a json object with function name and arguments within <tool_call></tool_call> XML tags:\n<tool_call>\n{"name": <function-name>, "arguments": <args-json-object>}\n</tool_call>'  # noqa: B950
+
+
+class Qwen25ChatTemplate(PromptTemplateInterface):
```
pbontrager, line 18: nit Qwen2_5

pbontrager (commented):
I'm interested to get more opinions here, but I think it would be clearer to have a fully separate 2.5 folder even if a few of the models have the exact same definitions. For the .5, 1.5, and 7B models you can directly have them call the qwen2 builders if you want. But this way we have separate tokenizers (without introducing the concept of tokenizer versions) and separate recipes.

Additionally, I want to see more information on how the new models were checked for accuracy and more configs for the 2.5 model.

torchtune/models/qwen2/_prompt_template.py
```diff
@@ -0,0 +1,117 @@
...
+            for tool in self._tools:
+                tool_instruction.append(tool)
+            tool_instruction.append(TOOL_INSTRUCTION_END)
+            tool_instruction = "\n".join(tool_instruction)
+            assert messages[0].role == "system"
+            messages[0].content.append({"type": "text", "content": tool_instruction})
+
+        # Add start/end tags to messages
+        formatted_dialogue = []
+        for i, message in enumerate(messages):
```
RdoubleA, line 87: is the primary logic here for adding tool call tokens? but these are special tokens correct? so adding them as strings may not get tokenized correctly. Instead, special tokens should be handled by `tokenize_messages` in the tokenizer. We should reserve the prompt template for non-special token text that is used for Qwen instruct, maybe the prompts you included at the top.

Before, it was just the ChatMLTemplate using `<|im_start|>` etc. If these have become special tokens then we should move that logic to the tokenizer. Otherwise, we should leave the logic the same as before, except maybe including the default system prompt you added.

I am not sure yet how to handle registered/available tools...

calvinpelletier: Adding them as strings does result in them being tokenized. `tokenizer.tokenize_messages` calls `tokenizer.encode` which detects these strings (any keys of the special tokens dict) and splits them out and converts them to the token IDs.

I think leaving this logic in the prompt template is better than moving it into `tokenize_messages` because we enable the user to pass in their own custom template. If we hide that logic in the tokenizer, they won't be able to customize it, and it would make the tokenizer more fragile (it could break if the user uses a custom template and the tokenizer receives messages in an unexpected format).

torchtune/models/qwen2/_tokenizer.py
```diff
@@ -23,13 +23,34 @@
     "<|im_start|>": 151644,
     "<|im_end|>": 151645,
 }
-
+QWEN2_5_SPECIAL_TOKENS = {
```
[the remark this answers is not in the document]
calvinpelletier: The tokenizer will handle the tokens if they're present in the data, but it won't insert them. Like for example, there's no way for us to reliably detect a tool call in the data and add `<tool_call>` special tokens around it, so we need the user to preprocess their dataset in a way that matches Qwen's preferred tool-calling format. Maybe we can add this information in the documentation? Other than the `<|im_*|>` tokens, the only special tokens we can proactively insert into the data, as far as I can tell, are the `vision_start/end` ones.

Regarding your second point, I changed `QwenTokenizer` back to `Qwen2Tokenizer`, but `QWEN2_SPECIAL_TOKENS` and `QWEN2_5_SPECIAL_TOKENS` needs to remain separate, otherwise Qwen2 models will start seeing Qwen2.5 tokens.

fzyzcjy:
Hi thank you for your great work!

I wonder whether it will support qwen2.5-math (instead of qwen2.5)?

calvinpelletier:
> Hi thank you for your great work!
> 
> I wonder whether it will support qwen2.5-math (instead of qwen2.5)?

@fzyzcjy  Yep! It's the same model just with different weights. So for example, run `tune download Qwen/Qwen2.5-Math-7B-Instruct --output-dir /tmp/Qwen2_5-Math-7B-Instruct --ignore-patterns None` then use the `qwen2_5/7B_lora_single_device` config with the `checkpointer.checkpoint_dir` updated to `/tmp/Qwen2_5-Math-7B-Instruct`.

For best results, you should preprocess your dataset to match the expected data format of Qwen2.5-Math. They recommend CoT or TIR reasoning, so a system prompt like "Please reason step by step, and put your final answer within \\boxed{}."

torchtune/models/qwen2/_tokenizer.py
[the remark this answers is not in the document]
RdoubleA: We can detect a tool call in the data. By the time we get to the tokenizer we can assume the user has already formatted their data into Messages, in which case role=assistant and ipython=True would indicate a tool call.

torchtune/models/qwen2/_prompt_template.py
RdoubleA: I see your point about configurability, in this case I am considering "prompt template" just to be any task related normal text, like "Question: Q Answer: A" that you want to structure inputs into. This is handled in the prompt template classes. Anything that has their own reserved token id should be handled in the tokenizer because they need to be added for that model regardless of what the task template is. This is to minimize accidental user error.

My main concern with this is that we're not being consistent here. And for configurability, we should probably agree on where users should configure custom special tokens. Right now model tokenizers have the option to modify that in the tokenizers builder (or maybe we need to make the experience better). There is also the prompt template which is a separate place to configure the actual task template (maybe we should name it that to avoid confusion). This changes kinda conflates the two and it's not clear to me where I should go for either case.

torchtune/models/qwen2_5/_model_builders.py
```diff
@@ -0,0 +1,629 @@
...
+        path=path,
+        merges_file=merges_file,
+        special_tokens=special_tokens,
+        max_seq_len=max_seq_len,
+        prompt_template=prompt_template,
+        **kwargs,
+    )
+
+
+def lora_qwen2_7b(
```
ebsmothers, line 241: this is a nit, but why are the model sizes out of order down here?

ebsmothers: (also just make sure that any changes you wind up making to the non-lora builders' params get propagated down here as well)

torchtune/models/qwen2_5/_model_builders.py
```diff
@@ -0,0 +1,629 @@
...
+        lora_alpha (float): scaling factor for the low-rank approximation
+        lora_dropout (float): dropout probability for the low-rank approximation. Default: 0.0
+        quantize_base (bool): Whether to quantize base model weights
+
+    Returns:
+        TransformerDecoder: Instantiation of Qwen2.5 0.5B model with LoRA applied
+
+    Note:
+        Qwen2.5 0.5B-3B model builders will enable `tie_word_embeddings` by default
+        and returns an instance of `TransformerDecoder`.
```
ebsmothers, line 328: I think they all return instances of TransformerDecoder now, right?

torchtune/models/qwen2_5/_prompt_template.py
```diff
@@ -0,0 +1,117 @@
...
+
+
+DEFAULT_SYS_PROMPT = (
+    "You are Qwen, created by Alibaba Cloud. You are a helpful assistant."
+)
+TOOL_INSTRUCTION_START = "\n\n# Tools\n\nYou may call one or more functions to assist with the user query.\n\nYou are provided with function signatures within <tools></tools> XML tags:\n<tools>"  # noqa: B950
+TOOL_INSTRUCTION_END = '</tools>\n\nFor each function call, return a json object with function name and arguments within <tool_call></tool_call> XML tags:\n<tool_call>\n{"name": <function-name>, "arguments": <args-json-object>}\n</tool_call>'  # noqa: B950
+
+
+class Qwen2_5ChatTemplate(PromptTemplateInterface):  # noqa: N801
```
ebsmothers, line 18: Can we add a test for this?

torchtune/models/qwen2_5/_prompt_template.py
```diff
@@ -0,0 +1,117 @@
...
+        add_default_sys_prompt_if_missing: bool = False,
+        tools: Optional[List[str]] = None,
+    ):
+        self._add_default_sys_prompt_if_missing = add_default_sys_prompt_if_missing
+        self._tools = tools
+
+    def __call__(
+        self,
+        messages: List[Message],
+        inference: bool = False,
```
ebsmothers, line 54: I know this is in the interface, but I don't see it used here (or in any of our other prompt templates). Can we remove it? @pbontrager

torchtune/models/qwen2_5/_prompt_template.py
```diff
@@ -0,0 +1,117 @@
...
+    Defined in the Jinja template in
+    https://huggingface.co/Qwen/Qwen2.5-7B-Instruct/raw/main/tokenizer_config.json
+
+    Attributes:
+        template: a dictionary mapping role to the
+            prepend and append tags
+
+    Args:
+        add_default_sys_prompt_if_missing (bool): Add Qwen's default system prompt if the messages do not have one.
+            Default: False
```
ebsmothers, line 30: Is this something we anticipate people wanting to configure?

torchtune/models/qwen2_5/_prompt_template.py
```diff
@@ -0,0 +1,117 @@
...
+            messages[0].content.append({"type": "text", "content": tool_instruction})
+
+        # Add start/end tags to messages
+        formatted_dialogue = []
+        for i, message in enumerate(messages):
+            if message.role == "ipython":
+                if i == 0 or messages[i - 1].role != "ipython":
+                    prepend_tag = f"{self.template['user'][0]}<tool_response>\n"
+                else:
+                    prepend_tag = "\n<tool_response>\n"
```
ebsmothers, line 92: This is interesting.. so we have conditional prepend/append tags depending on contextual information. Feels like this is a gap in our current prompt template abstraction (maybe no way around it though)

ebsmothers (commented):
Left a bunch of comments, also wanna +1 @pbontrager's comment [here](https://github.com/pytorch/torchtune/pull/1863#pullrequestreview-2378454093). We probably wanna at least have some E2E runs with eval results for a couple of different model sizes

torchtune/models/qwen2_5/_prompt_template.py
RdoubleA: we already handle similar tags in llama3 in `tokenize_messages`

ebsmothers: Oh duh good point. Is there a reason we can't move this to the tokenizer then? (Other than the fact that I assume it differs for Qwen2 vs Qwen2.5)

RdoubleA: See conversation here: https://github.com/pytorch/torchtune/pull/1863#discussion_r1806971190

torchtune/models/qwen2/_prompt_template.py
ebsmothers: Just getting caught up on this.. I think both of you have good points here. Ideally we would just be able to support the use case @calvinpelletier is describing just by changing the prompt template. But I definitely would prefer to address that in a systematic way rather than doing just a one-off here. Iiuc `<tool_response>` and `<\tool_response>` are both recognized by the tokenizer, so given that I don't think it makes sense to use them in string form here (plus imo doing this is also kinda brittle). Re the point about not being able to customize if we put this in the tokenizer, I'm not sure I 100% follow. What kind of customization would not be supported? Even in the worst case, one can inherit the tokenizer class and just override `tokenize_messages` as needed (which again, if users accept our claim of "special tokens get dealt with in the tokenizer" may be a bit more friction, but is at least the expected thing to do).

torchtune/models/qwen2_5/__init__.py
```diff
@@ -0,0 +1,43 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+from ._model_builders import (
+    lora_qwen2_0_5b,
```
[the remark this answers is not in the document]
calvinpelletier: Oops I'll fix that. Also, I hope we don't one day have a "Qwen3 1.5B" "Qwen3.1 5B" because they're both gonna be named `qwen3_1_5b` :P

torchtune/models/qwen2/_tokenizer.py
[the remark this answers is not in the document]
calvinpelletier: If `ipython==True`, can we assume that the entire message is the tool call? Or is it possible for the message to contain other text as well?

torchtune/models/qwen2/_prompt_template.py
calvinpelletier: Since the prompt template already adds special tokens to the messages (`<|im_start|>`, etc.) , my understanding was that the prompt template is responsible for adding special token _text_ and the tokenizer is responsible for converting the text into token IDs. Wrapping an ipython message in `<tool_response>` feels like the same logic as wrapping a user message in `<|im_start|>user`.

That's a fair point about the customization though. At first, I was thinking a user might want to do things like have multiple tool responses in the same message or add additional XML tags like `<tool_status>` or whatever. But now that I'm thinking more about this, I don't think it's something we need to worry about.

torchtune/models/qwen2_5/_model_builders.py
```diff
@@ -0,0 +1,629 @@
...
+    """
+    return qwen2(
+        vocab_size=151936,
+        num_layers=28,
+        num_heads=12,
+        num_kv_heads=2,
+        embed_dim=1536,
+        intermediate_dim=8960,
+        max_seq_len=32768,
```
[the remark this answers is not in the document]
calvinpelletier: Where do you see 131072? According to [this](https://huggingface.co/Qwen/Qwen2.5-1.5B-Instruct/blob/main/config.json), the max seq len (`max_position_embeddings`) is 32768.

torchtune/models/qwen2_5/_model_builders.py
```diff
@@ -0,0 +1,629 @@
...
+        TransformerDecoder: Instantiation of Qwen2.5 14B model
+    """
+    return qwen2(
+        vocab_size=152064,
+        num_layers=48,
+        num_heads=40,
+        num_kv_heads=8,
+        embed_dim=5120,
+        intermediate_dim=13824,
+        max_seq_len=32768,
```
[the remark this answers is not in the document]
calvinpelletier: The config [here](https://huggingface.co/Qwen/Qwen2.5-14B-Instruct/blob/main/config.json) has `"max_position_embeddings": 32768`. 

It has `"sliding_window": 131072` but I don't think this matters since `"use_sliding_window": false`

torchtune/models/qwen2_5/_model_builders.py
```diff
@@ -0,0 +1,629 @@
...
+    return qwen2(
+        vocab_size=152064,
+        num_layers=64,
+        num_heads=40,
+        num_kv_heads=8,
+        embed_dim=5120,
+        intermediate_dim=27648,
+        max_seq_len=32768,
+        attn_dropout=0.0,
+        norm_eps=1e-06,
```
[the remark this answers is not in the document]
calvinpelletier: [this](https://huggingface.co/Qwen/Qwen2.5-32B-Instruct/blob/main/config.json) has `"rms_norm_eps": 1e-06`

torchtune/models/qwen2_5/_model_builders.py
```diff
@@ -0,0 +1,629 @@
...
+        TransformerDecoder: Instantiation of Qwen2.5 32B model
+    """
+    return qwen2(
+        vocab_size=152064,
+        num_layers=64,
+        num_heads=40,
+        num_kv_heads=8,
+        embed_dim=5120,
+        intermediate_dim=27648,
+        max_seq_len=32768,
```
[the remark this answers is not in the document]
calvinpelletier: [this](https://huggingface.co/Qwen/Qwen2.5-32B-Instruct/blob/main/config.json) has `"max_position_embeddings": 32768`

torchtune/models/qwen2_5/_model_builders.py
```diff
@@ -0,0 +1,629 @@
...
+        TransformerDecoder: Instantiation of Qwen2.5 72B model
+    """
+    return qwen2(
+        vocab_size=152064,
+        num_layers=80,
+        num_heads=64,
+        num_kv_heads=8,
+        embed_dim=8192,
+        intermediate_dim=29568,
+        max_seq_len=32768,
```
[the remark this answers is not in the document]
calvinpelletier: `"max_position_embeddings": 32768` from [this](https://huggingface.co/Qwen/Qwen2.5-72B-Instruct/blob/main/config.json)

torchtune/models/qwen2_5/_model_builders.py
```diff
@@ -0,0 +1,629 @@
...
+        attn_dropout=0.0,
+        norm_eps=1e-06,
```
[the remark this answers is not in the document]
calvinpelletier: `"rms_norm_eps": 1e-06` from [this](https://huggingface.co/Qwen/Qwen2.5-72B-Instruct/blob/main/config.json)

torchtune/models/qwen2_5/_prompt_template.py
calvinpelletier: Not really, I just added this for parity with the Qwen's official prompt template. Should I remove it?

torchtune/models/qwen2/_prompt_template.py
RdoubleA: ironically, the <|im_start|> is not a reserved token ID. It's just raw text that gets tokenized normally that can be used by any model, hence why it's located in prompt template and not in the tokenizer.

> my understanding was that the prompt template is responsible for adding special token text and the tokenizer is responsible for converting the text into token IDs.

yeah maybe we should make this more clear. in my mind the distinction was prompt template = structured plain text, tokenizer = all special tokens. we can discuss if we want to change this but I would say to be consistent with this until we do make any changes.

torchtune/models/qwen2/_tokenizer.py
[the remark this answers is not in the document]
RdoubleA: yes, I would say it's safe to assume this. If a user includes content in a Message that's not a tool call but has `ipython=True`, I would consider that a user error. We should be very clear in documentation though to guide users.

calvinpelletier:
oh @ebsmothers I just realized what happened, I was looking at the qwen2.5 instruct configs and you were looking at the base model configs. I just assumed they would be the same models with different params, since instruct should just be a finetune of the base model. I'll create separate model builders for the instruct and base models.

torchtune/models/qwen2_5/_model_builders.py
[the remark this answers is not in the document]
calvinpelletier: ah i see what's happening, see my comment below

torchtune/models/qwen2/_tokenizer.py
```diff
@@ -345,6 +343,9 @@ def tokenize_messages(
         Raises:
             RuntimeError: If a message contains non-text content
         """
+        assert not isinstance(
+            self.prompt_template, ChatMLTemplate
+        ), "Using ChatMLTemplate with tokenize_messages will result in multiple <|im_*|> tokens wrapping each message."
```
RdoubleA, line 348: nit: we should add an actionable message as well, "Please use a different template or set to None"

torchtune/models/qwen2/_tokenizer.py
```diff
@@ -151,9 +151,7 @@ def __init__(
 
         self.pat = re.compile(PRETOKENIZE_REGEX)
 
-        self.special_tokens = (
-            special_tokens if special_tokens is not None else QWEN2_SPECIAL_TOKENS
-        )
+        self.special_tokens = special_tokens
```
RdoubleA, line 154: the reason this was none was because of some pythonic issue with using a dict as a default, it will make all instances of the Qwen2Tokenizer class share the same `QWEN2_SPECIAL_TOKENS`, and if a user were to ever modify it it woulud be changed for every single instance. don't know if that's an edge case we should really consider, but just fyi

torchtune/models/qwen2/_tokenizer.py
```diff
@@ -355,29 +356,48 @@ def tokenize_messages(
         mask = []
         for index, message in enumerate(templated_messages):
             tokens = []
+
+            # message header
+            if message.role != "ipython":
```
RdoubleA, line 361: what happens in the original Qwen2 tokenizer if we have a message from a tool return? should we gat against tool call / tool messages?

torchtune/models/qwen2_5/_tokenizer.py
```diff
@@ -0,0 +1,232 @@
...
+            to add structured text around the actual messages. The structured text is used in three scenarios:
+
+            - Task-specific templates to gear models for a particular task that it will expect after training
+            - Model-specific templates that are required whenever the model is prompted, such as the [INST]
+              tags in Llama2 and in Mistral
+            - Community standardized templates, such as :class:`~torchtune.data.ChatMLTemplate`
+
+            The extra text will still get tokenized as normal text, not as special tokens.
+            Default: None
+        errors (str): Paradigm to follow when decoding bytes to UTF-8. Defaults to "replace".
```
RdoubleA, line 65: would make this parameter name a bit more descriptive, like `decode_error_handler`?

torchtune/models/qwen2_5/_tokenizer.py
```diff
@@ -0,0 +1,232 @@
...
+
+        tokenized_messages = []
+        mask = []
+        for i, message in enumerate(templated_messages):
+            tokens = []
+
+            # message header
+            if message.role == "ipython":
+                if i == 0 or templated_messages[i - 1].role != "ipython":
+                    self._add_message_start_tokens(tokens, "user")
```
RdoubleA, line 153: do you mind adding a comment here explaining why we're adding a user header and checking the previous message? not sure what cases this is covering

torchtune/models/qwen2_5/_tokenizer.py
```diff
@@ -0,0 +1,232 @@
...
+                    )
+                    self._add_message_end_tokens(tokens)
+                else:
+                    tokens.extend(
+                        self.encode("\n</tool_response>", add_bos=False, add_eos=False)
+                    )
+            else:
+                if message.role == "assistant" and message.ipython:
+                    tokens.extend(
+                        self.encode("\n</tool_call>", add_bos=False, add_eos=False)
```
RdoubleA, line 200: would just add the id directly since you defined it in the special tokens dict above. or better yet I would save the tool call id as an attribute and add it directly here instead of encoding in case the user wants to change the id:
```
self.tool_call_start_id = self.special_tokens["<tool_call>"]
self.tool_call_end_id = self.special_tokens["</tool_call>"]
...
tokens.extend([self.tool_call_start_id])
```
so you don't have to encode every single time, save a tiny bit of time. same comment for tool response.

I wonder how a user would go about changing the actual encoded string for tool calls though, I guess there's no easy way to do that...

torchtune/models/qwen2_5/_tokenizer.py
```diff
@@ -0,0 +1,232 @@
...
+            for item in message.content:
+                if item["type"] == "text":
+                    tokens.extend(
+                        self.encode(
+                            item["content"],
+                            add_bos=False,
+                            add_eos=False,
+                        )
+                    )
+                else:
```
RdoubleA, line 178: does Qwen2.5 not support images?

RdoubleA (commented):
Looks awesome, I just left a few minor comments. Will let @ebsmothers comment more on the model builders.

torchtune/models/qwen2/_tokenizer.py
calvinpelletier: wouldn't that happen in the original way anyway? all instances initialized with special_tokens=none would share the same reference to the QWEN2_SPECIAL_TOKENS, right? i think this is fine since we don't modify it, but i could deepcopy it to be safe

torchtune/models/qwen2/_tokenizer.py
calvinpelletier: Originally, this logic for Qwen2 existed in the ChatMLTemplate. We don't do anything for tool responses, i guess because Qwen2 didn't explicitly support it

torchtune/models/qwen2_5/_tokenizer.py
calvinpelletier: oops yeah it does, I'll look into how we should encode images

recipes/configs/qwen2_5/72B_lora.yaml
```diff
@@ -0,0 +1,146 @@
...
+#
+# To launch on 8 devices, run the following command from root:
+#   tune run --nnodes 1 --nproc_per_node 8 lora_finetune_distributed --config qwen2_5/72B_lora
+#
+# You can add specific overrides through the command line. For example
+# to override the checkpointer directory while launching training
+# you can run:
+#   tune run --nnodes 1 --nproc_per_node 8 lora_finetune_distributed --config qwen2_5/72B_lora checkpointer.checkpoint_dir=<YOUR_CHECKPOINT_DIR>
+#
+# This config works best when the model is being fine-tuned on 2+ GPUs.
```
RdoubleA, line 16: can you actually run this on only 2?

recipes/configs/qwen2_5/7B_full.yaml
```diff
@@ -0,0 +1,79 @@
...
+#   tune run --nnodes 1 --nproc_per_node 4 full_finetune_distributed --config qwen2_5/7B_full
+#
+# You can add specific overrides through the command line. For example
+# to override the checkpointer directory while launching training
+# you can run:
+#   tune run --nnodes 1 --nproc_per_node 4 full_finetune_distributed --config qwen2_5/7B_full checkpointer.checkpoint_dir=<YOUR_CHECKPOINT_DIR>
+#
+# This config works best when the model is being fine-tuned on 2+ GPUs.
+# Single device full finetuning requires more memory optimizations. It's
+# best to use 7B_full.yaml for those cases
```
RdoubleA, line 18:
```suggestion
# best to use 7B_full_single_device.yaml for those cases
```

tests/torchtune/models/qwen2_5/test_tokenizer.py
```diff
@@ -0,0 +1,195 @@
...
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+from tests.common import ASSETS
+
+from torchtune.data import Message
+from torchtune.models.qwen2_5 import qwen2_5_tokenizer
+
+
+class TestQwenTokenizer:
```
RdoubleA, line 13: nit:
```suggestion
class TestQwen2_5Tokenizer:
```

recipes/configs/qwen2_5/32B_lora.yaml
```diff
@@ -0,0 +1,125 @@
+# Config for multi-device LoRA finetuning in lora_finetune_distributed.py
```
joecummings, line 1: Why do we only have 32B lora distributed?

torchtune/models/qwen2_5/_model_builders.py
```diff
@@ -0,0 +1,1015 @@
...
+"""
+Model builders build specific instantiations using component builders. For example
+the qwen2_5_7b model builder uses the qwen2 component builder to create the
+Qwen2.5 7B model.
+"""
+
+
+def qwen2_5_0_5b() -> TransformerDecoder:
+    """
+    Builder for creating a Qwen2.5 model initialized w/ the default 0.5B parameter values
```
RdoubleA, line 25: is this both base and instruct?

joecummings (commented):
I think you're missing some default configs.

torchtune/models/qwen2_5/_model_builders.py
```diff
@@ -0,0 +1,629 @@
...
+    merges_file: str,
+    special_tokens_path: Optional[str] = None,
+    max_seq_len: Optional[int] = None,
+    prompt_template: Optional[
+        str
+    ] = "torchtune.models.qwen2_5._prompt_template.Qwen2_5ChatTemplate",
+    **kwargs,
+) -> Qwen2Tokenizer:
+    """
+    Tokenizer for Qwen2.5.
```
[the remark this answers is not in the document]
RdoubleA: bumping this

torchtune/models/qwen2_5/_model_builders.py
```diff
@@ -0,0 +1,1015 @@
...
+        special_tokens_path (Optional[str]): Path to ``tokenizer.json`` from Hugging Face
+            model files that contains all registered special tokens, or a local json file
+            structured similarly. Default is None to use the canonical Qwen2.5 special tokens.
+        max_seq_len (Optional[int]): A max sequence length to truncate tokens to.
+            Default: None
+        prompt_template (Optional[_TemplateType]): optional specified prompt template.
+            If a string, it is assumed to be the dotpath of a :class:`~torchtune.data.PromptTemplateInterface`
+            class. If a dictionary, it is assumed to be a custom prompt template mapping role to the
+            prepend/append tags.
+            Default is :class:`~torchtune.models.qwen2_5._prompt_template.Qwen2_5ChatTemplate`.
```
RdoubleA, line 326: this doesnt exist?

torchtune/models/qwen2_5/_tokenizer.py
```diff
@@ -0,0 +1,235 @@
...
+            else messages
+        )
+
+        tokenized_messages = []
+        mask = []
+        for i, message in enumerate(templated_messages):
+            tokens = []
+
+            # message header
+            if message.role == "ipython":
```
RdoubleA, line 155: nit: would suggest breaking this up into methods (i.e., tokenize_header, tokenize_body, tokenize_end), similar to Llama3Tokenizer

RdoubleA (commented):
If it's not too difficult, would be awesome if you can expose `packed` and `compile` in all the configs.

I'm also a bit confused on why some model builders have a base and instruct version and some don't. Would also be great if you could add to the docstring why the instruct is different from the base (seems like different seq len?)

recipes/configs/qwen2_5/32B_lora.yaml
calvinpelletier: I figured at that model size people would almost always be doing a distributed LoRA, right? I got OOMs for the other configs

torchtune/models/qwen2_5/_model_builders.py
calvinpelletier: yes, for some reason some model sizes have different base/instruct architectures while others have the same. ill add a comment

torchtune/models/qwen2_5/_tokenizer.py
```diff
@@ -0,0 +1,242 @@
...
+        eos_token (str): The end of sequence token. Defaults to ``<|endoftext|>``.
+        pad_token (Optional[str]): The token used for padding. Defaults to ``<|endoftext|>``.
+        bpe_cache_size (int): BPE token cache size in Qwen2Tokenizer.
+            NOTE: large cache size will speed up tokenization, but the cache object will get really
+            large for long running processes (esp. for texts of language that do not use space between
+            word, e.g. Chinese); technically not a memory leak but appears as one.
+            By default, we set the cache size equals to size of the official Qwen2 tokenizer.
+
+    Example:
+        >>> tokenizer = Qwen2Tokenizer(
```
RdoubleA, line 79: Could you update this example?

RdoubleA (approved):
Legendary PR, thanks for adding this model and all its variants

Issue #1850, opened 2024-10-16
General tracker of RLHF methods addition


krammnic:
Approved by @SalmanMohammadi. This issue's purpose is to determine which contrastive optimization methods should be added in torchtune and track methods that are implementing right now. For all of the methods, reference implementation exists. Here is the list of the methods (most methods are from A* alignment workshops) with their implementation state and some info about them (HF models trained with method, limitations, notes about actuality):

- [x] DPO: Classic method, with which constrative methods are usually compared. **Models:** >6k, Status: Implemented #645
- [x] SimPO: Current SOTA method according to their paper. **Models:** ~252, **Limitations:** The most important limitation is performance drop on reasoning datasets. Method is actually lack of robustness for tasks where specific tokens are important. For instance, `2 + 2 = 5` is a bad result compared to `2 + 2 = 4`. For the v0.2 version, it can't be normally used in cases where structured outputs are required, but it is actually a version where robustness to learning rate is maximized. See also general limitations. **Status:** Implemented, #1223
- [ ] IPO: Basic log-ratio method; I am not sure of its actuality. **Models:** <180, Limitations: Are not really clear from the paper. See also general limitations. **Status:** Stale, #1615
- [ ] KTO: Kahneman-Tversky value function-based method, which looks really interesting and also was used for reasoning tasks. **Models:** ~531 **Limitations:** It is not a real limitation, but if your data is of enough quality, DPO will work better than this method. **Status:** Not implemented, issue #1793
- [ ] ORPO: Pretty popular and stable log-ratio method, which has actually worse performance than SimPO, same computational and memory efficiency, but it does not require SFT initialized model. **Models:** ~1220, **Limitations:** Problems with reasoning datasets (checked emperically). See also general limitations. **Status:** Not implemented.
- [ ] R-DPO: Really interesting, but non-trivial method, closest to SimPO by performance. Generally, it is a reward-model (incorporated in DPO loss) method with teacher/student distillation. **Models:** 29, Limitations: Are not really clear from the paper. It is fully robust to noise that might be generated while building synthetic. Also, it looks like it has better robustness to hyperparameters than log-ratio methods. The reward model might be counted as a potential limitation. **Status:** Not implemented.
- [ ] CPO: Was actually designed for translation tasks, log-ratio method; not sure about its actuality. **Models:** <120, **Limitations:** Probably pretty similar to other log-ratio methods' limitations. See also: General limitations. **Status:** Not implemented, #1290 
[...]

PR #1864, opened 2024-10-18
[WIP] Add KTOLoss

krammnic:
#### Context
- [X ] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

#1850 
#### Changelog
* Adding KTO loss to torchtune

#### Test plan

#### UX

### DIFF
--- a/torchtune/rlhf/loss/dpo.py
+++ b/torchtune/rlhf/loss/dpo.py
@@ -36,20 +36,20 @@ class DPOLoss(nn.Module):
     """
 
     def __init__(
-        self,
-        beta: float = 0.1,
-        label_smoothing: float = 0.0,
+            self,
+            beta: float = 0.1,
+            label_smoothing: float = 0.0,
     ):
         super().__init__()
         self.beta = beta
         self.label_smoothing = label_smoothing
 
     def forward(
-        self,
-        policy_chosen_logps: torch.Tensor,
-        policy_rejected_logps: torch.Tensor,
-        reference_chosen_logps: torch.Tensor,
-        reference_rejected_logps: torch.Tensor,
+            self,
+            policy_chosen_logps: torch.Tensor,
+            policy_rejected_logps: torch.Tensor,
+            reference_chosen_logps: torch.Tensor,
+            reference_rejected_logps: torch.Tensor,
     ) -> Tuple[torch.Tensor, torch.Tensor, torch.Tensor]:
         """
         Compute the DPO loss for a batch of policy and reference model log probabilities.
@@ -80,15 +80,15 @@ class DPOLoss(nn.Module):
         # We ignore the reference model as beta -> 0. The label_smoothing parameter encodes our uncertainty about the labels and
         # calculates a conservative DPO loss.
         losses = (
-            -F.logsigmoid(self.beta * logits) * (1 - self.label_smoothing)
-            - F.logsigmoid(-self.beta * logits) * self.label_smoothing
+                -F.logsigmoid(self.beta * logits) * (1 - self.label_smoothing)
+                - F.logsigmoid(-self.beta * logits) * self.label_smoothing
         )
 
         chosen_rewards = (
-            self.beta * (policy_chosen_logps - reference_chosen_logps).detach()
+                self.beta * (policy_chosen_logps - reference_chosen_logps).detach()
         )
         rejected_rewards = (
-            self.beta * (policy_rejected_logps - reference_rejected_logps).detach()
+                self.beta * (policy_rejected_logps - reference_rejected_logps).detach()
         )
 
         return losses, chosen_rewards, rejected_rewards
@@ -110,18 +110,18 @@ class RSOLoss(nn.Module):
     """
 
     def __init__(
-        self,
-        gamma: float = 0.1,
+            self,
+            gamma: float = 0.1,
     ):
         super().__init__()
         self.gamma = gamma
 
     def forward(
-        self,
-        policy_chosen_logps: torch.Tensor,
-        policy_rejected_logps: torch.Tensor,
-        reference_chosen_logps: torch.Tensor,
-        reference_rejected_logps: torch.Tensor,
+            self,
+            policy_chosen_logps: torch.Tensor,
+            policy_rejected_logps: torch.Tensor,
+            reference_chosen_logps: torch.Tensor,
+            reference_rejected_logps: torch.Tensor,
     ) -> Tuple[torch.Tensor, torch.Tensor, torch.Tensor]:
         """
         Compute the RSO loss for a batch of policy and reference model log probabilities.
@@ -151,10 +151,10 @@ class RSOLoss(nn.Module):
         losses = torch.relu(1 - self.gamma * logits)
 
         chosen_rewards = (
-            self.gamma * (policy_chosen_logps - reference_chosen_logps).detach()
+                self.gamma * (policy_chosen_logps - reference_chosen_logps).detach()
         )
         rejected_rewards = (
-            self.gamma * (policy_rejected_logps - reference_rejected_logps).detach()
+                self.gamma * (policy_rejected_logps - reference_rejected_logps).detach()
         )
 
         return losses, chosen_rewards, rejected_rewards
@@ -186,10 +186,10 @@ class SimPOLoss(nn.Module):
     """
 
     def __init__(
-        self,
-        beta: float = 2.0,
-        gamma: float = 0.5,
-        label_smoothing: float = 0.0,
+            self,
+            beta: float = 2.0,
+            gamma: float = 0.5,
+            label_smoothing: float = 0.0,
     ):
         super().__init__()
         self.beta = beta
@@ -197,9 +197,9 @@ class SimPOLoss(nn.Module):
         self.label_smoothing = label_smoothing
 
     def forward(
-        self,
-        policy_chosen_logps: torch.Tensor,
-        policy_rejected_logps: torch.Tensor,
+            self,
+            policy_chosen_logps: torch.Tensor,
+            policy_rejected_logps: torch.Tensor,
     ) -> Tuple[torch.Tensor, torch.Tensor, torch.Tensor]:
         """
         Compute the SimPO loss for a batch chosen and rejected average log probabilities.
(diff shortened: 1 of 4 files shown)


krammnic:
- [ ] Run some recipe with this loss.
- [ ] Add APO loss variant
- [ ] KL divergence?

krammnic:
@SalmanMohammadi Should I make a full run with KTOLoss?

SalmanMohammadi:
Hey @krammnic. Thanks so much for opening this. This is a great start.

I've been thinking about to improve the contributor workflow, and for contributions which may involve design decisions i.e. touch our existing codebase in a non-trivial way, require refactoring, or generally need some carefulness to make sure we're exposing new features to users in the best way possible.

So, before we dive into the specific implementation details, I'd like to take a step back. To prevent thrash and lots of back-and-forth and discovering design reconsiderations later into the PR, let's de-risk things by better understanding what we're doing here, and agree on some of the necessary steps we need to see this landed. I'm going to shamelessly plug my SimPO PR (https://github.com/pytorch/torchtune/pull/1223) as an example of this.

Would you be willing to answer some questions for me by updating the PR description?
- Could you provide a very very high-level overview of what this method is? What are the main ways it differs from e.g. DPO? You could provide some information from the paper here.
- Have you used a reference implementation? If so, which one? Code pointers here are very useful. I like to use TRL (https://github.com/huggingface/trl) which I believe has a KTOTrainer, but others also use official repos from paper authors.
- What would integrating this into the DPO recipes look like? Does it fit neatly into the existing DPO logic, or do we need some specific logic in the recipe for KTO? 
    - From skimming the KTOTrainer [docpage](https://huggingface.co/docs/trl/main/en/kto_trainer), it looks like TRL specify a different dataset format which differs from our preference dataset format in a meaningful way. As above with outlining the main differences between KTO and DPO, how would this different dataset format be used?

Generally what we're looking for is to make sure we don't find any gotchas down the line and that we can have an overview of the expected changes and outcomes for this contribution. 

A couple disclaimers:
- If this sounds like overkill - let me know! If you feel like there's a different way you'd like to go about it, please feel free! Similarly, I'm more than happy to help investigate on some of these points. You can also drop me a line on Discord : )
- Also, if this seems like a lot of effort, we can come back to this PR and you could try one of the RLHF techniques that you mentioned which may be easier to implement into our codebase. That's totally fine.

krammnic:
Thanks for comment! I think there are already answers on most of your question in docstring of the KTOLoss) And in general RLHF issue. I might duplicate them here if required. 

Recipe should not be significantly changed. Change loss to KTOLoss. Will check about the dataset but generally in all contrastives they are pretty same

torchtune/rlhf/loss/dpo.py
```diff
@@ -231,3 +231,91 @@ def forward(
...
+
+class KTOLoss(nn.Module):
+    """
+    KTO: Kahneman-Tversky Optimization: https://arxiv.org/abs/2402.01306
+    Intuition from the paper:
+
+        The effectiveness of SimPO is attributed to a key design: using the average log probability of a sequence as
+        the implicit reward. Additionally, we introduce a target reward margin to the Bradley-Terry objective to
+        encourage a larger margin between the winning and losing responses, further enhancing the algorithm's performance.
+
```
SalmanMohammadi, line 244: It looks like this has been copied from the SimPO loss?

krammnic:
We can conclude main idea of the method from its limitations! KTO >= DPO and DPO should be used in cases with high quality data, in other cases KTO is really suitable(Actually, there are 2 versions of this loss, which gives more control on situation with your dataset, but they are minorly different, so usually included in one loss(will do it as follow up))

torchtune/rlhf/loss/dpo.py
krammnic: Oh, I used SimPO docs as template and forgot to add my docs in this part. Other comments are about KTO

krammnic: Fixed

torchtune/rlhf/loss/dpo.py
```diff
@@ -231,3 +231,91 @@ def forward(
...
+        losses = torch.cat(
+            (
+                self.desirable_weight * chosen_losses,
+                self.undesirable_weight * rejected_losses,
+            ),
+            0,
+        )
+
+        chosen_rewards = self.beta * (chosen_logratios).detach()
+        rejected_rewards = self.beta * (rejected_logratios).detach()
```
SalmanMohammadi, line 319: I'm not convinced that this is correct. Could you share a reference implementation for this loss function? For example, in TRL, the KL calculation is done differently (https://github.com/huggingface/trl/blob/main/trl/trainer/kto_trainer.py#L1100) - you can see that `self.calculate_KL` is `True` for the KTO loss (https://github.com/huggingface/trl/blob/main/trl/trainer/kto_trainer.py#L545).

SalmanMohammadi:
> I might duplicate them here if required.

That would be very helpful to see! I'm also not convinced about the dataset format, as I've also found in the official KTO repo (https://github.com/ContextualAI/HALOs?tab=readme-ov-file) that:
> KTO doesn't use preference pairs, just knowledge of whether outputs are desirable or undesirable. This means we use dataloader.UnpairedPreferenceDataLoader. However, that dataloader assumes that you're working with datasets that originally contain preference pairs, like Anthropic HH or SHP
So we need to think a bit about how to make sure our implementation in the DPO recipe retains parity here. We could use some minor modification of a preference dataset here, but I'd like to make sure we're aware of these kinds of difference.

torchtune/rlhf/loss/dpo.py
SalmanMohammadi: This is the same in the official KTO repo https://github.com/ContextualAI/HALOs/blob/61b9ee6787c9f52c2f4578533866e1cb42f314ec/trainers.py#L836

krammnic: Yes, I saved this mostly as drafted(version without KL) will bring it

krammnic:
> > I might duplicate them here if required.
> 
> That would be very helpful to see! I'm also not convinced about the dataset format, as I've also found in the official KTO repo (https://github.com/ContextualAI/HALOs?tab=readme-ov-file) that:
> 
> > KTO doesn't use preference pairs, just knowledge of whether outputs are desirable or undesirable. This means we use dataloader.UnpairedPreferenceDataLoader. However, that dataloader assumes that you're working with datasets that originally contain preference pairs, like Anthropic HH or SHP
> 
> So we need to think a bit about how to make sure this implementation in the DPO recipe retains parity here. Do we need use some minor modification of a preference dataset here?

Yes, probably will require special things about data loading

torchtune/rlhf/loss/dpo.py
krammnic: Hmm, to actually get kl_logprobs we need to modify concatenated_forward

krammnic: I don't want to pass cfg in concatenated_forward

krammnic: We can actually ignore KL divergence and use APO variant of loss:

`rejected_losses = F.sigmoid(self.beta * rejected_logratios)`

krammnic: And we need to actually change structure of the batch

krammnic:
> > > I might duplicate them here if required.
> > 
> > 
> > That would be very helpful to see! I'm also not convinced about the dataset format, as I've also found in the official KTO repo (https://github.com/ContextualAI/HALOs?tab=readme-ov-file) that:
> > > KTO doesn't use preference pairs, just knowledge of whether outputs are desirable or undesirable. This means we use dataloader.UnpairedPreferenceDataLoader. However, that dataloader assumes that you're working with datasets that originally contain preference pairs, like Anthropic HH or SHP
> > 
> > 
> > So we need to think a bit about how to make sure this implementation in the DPO recipe retains parity here. Do we need use some minor modification of a preference dataset here?
> 
> Yes, probably will require special things about data loading

Looks like it want be difficult, but actually we can create new dataset type for this. So basically to convert pair dataset to KTO format we need something like:

```
batch_size = len(examples["chosen"])
new_rows = {
        "completion": examples["chosen"] + examples["rejected"],
        "label": [True] * batch_size + [False] * batch_size,
 }
if "prompt" in examples:
        new_rows["prompt"] = examples["prompt"] + examples["prompt"]
```

And obviously remove required columns. I'm not sure where to put such preprocessing. Either new dataset type, or new argument in current dataset?

SalmanMohammadi:
> And we need to actually change structure of the batch. Also would require some processing to dataset. Do you approve such changes?

Would you be able to outline some of the necessary steps we need to make on the dataset side?

> We can actually ignore KL divergence right now and use APO variant of loss:
  rejected_losses = F.sigmoid(self.beta * rejected_logratios)

APO is a separate (but comparable) technique to KTO, right? Is [this](https://huggingface.co/papers/2408.06266) the right paper? It looks like it came out < 2 months ago, and I see very few models on the hub mentioning APO. As I mentioned in the RLHF tracker, I'm not sure we should be implementing techniques which are new and haven't been adopted by the OS community, so I'd rather we stick with the original proposal of KTO, and I'd love to keep supporting you on this. 

I don't fully follow your comments on the necessary modifications needed to implement KTO, you mentioned needing to modify `concatenated_forward`? I'm not sure I fully follow along so it'd be really helpful if you could also outline why we need to make these changes - that way we can figure out the best way to implement them.

krammnic:
1. About APO. It looks like it is just KTO without KL, but ok, let's skip
2. Therefore, we don't need to pass anything extra in forward!
3. We need to do this preprocessing to dataset:
```
batch_size = len(examples["chosen"])
new_rows = {
        "completion": examples["chosen"] + examples["rejected"],
        "label": [True] * batch_size + [False] * batch_size,
 }
if "prompt" in examples:
        new_rows["prompt"] = examples["prompt"] + examples["prompt"]
``` and map it doataset

4. But we need to calculate: `policy_KL_logps, reference_KL_logps`
5. then we need batch["reference_KL_logps"] and policy_KL_logps

**Comment by krammnic** [comment]
> 1. About APO. It looks like it is just KTO without KL, but ok, let's skip
> 2. Therefore, we don't need to pass anything extra in forward!
> 3. We need to do this preprocessing to dataset:
> 
> ```
> batch_size = len(examples["chosen"])
> new_rows = {
>         "completion": examples["chosen"] + examples["rejected"],
>         "label": [True] * batch_size + [False] * batch_size,
>  }
> if "prompt" in examples:
>         new_rows["prompt"] = examples["prompt"] + examples["prompt"]
> ```
> 
> and map it to dataset
> 
> 4. But we need to calculate: `policy_KL_logps, reference_KL_logps`
> 5. then we need batch["reference_KL_logps"] and policy_KL_logps

@SalmanMohammadi Can we continue from here?

**Comment by SalmanMohammadi** [comment]
> > 1. About APO. It looks like it is just KTO without KL, but ok, let's skip
> > 2. Therefore, we don't need to pass anything extra in forward!
> > 3. We need to do this preprocessing to dataset:
> > 
> > ```
> > batch_size = len(examples["chosen"])
> > new_rows = {
> >         "completion": examples["chosen"] + examples["rejected"],
> >         "label": [True] * batch_size + [False] * batch_size,
> >  }
> > if "prompt" in examples:
> >         new_rows["prompt"] = examples["prompt"] + examples["prompt"]
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > and map it to dataset
> > 
> > 4. But we need to calculate: `policy_KL_logps, reference_KL_logps`
> > 5. then we need batch["reference_KL_logps"] and policy_KL_logps
> 
> @SalmanMohammadi Can we continue from here?

Hey @krammnic - thanks for summarizing here, it seems like we can make things simpler if we don't try to add APO too? 

I need a bit of time to look into your other questions which I may not get round to until later this week. If you have an idea of how you'd like to proceed here I'd say go for it and we can try iterate if we need to. I'm interested in your thoughts on whether you think we'll need to make any changes to the recipe itself - for example, does `concatenated_forward` need to be changed? Or can we make modifications just at the dataset level?

Out of curiosity, where did you grab this from?

> batch_size = len(examples["chosen"])
new_rows = {
        "completion": examples["chosen"] + examples["rejected"],
        "label": [True] * batch_size + [False] * batch_size,
 }
if "prompt" in examples:
        new_rows["prompt"] = examples["prompt"] + examples["prompt"]

**Comment by krammnic** [comment]
Thanks for the answer!

Yes some changes will be required as we need to get batch["reference_KL_logps"] and policy_KL_logps to calculate KL divergence(will write what we need to do here soon). 

Dataset processing implementation is from TRL:

https://github.com/huggingface/trl/blob/main/trl/data_utils.py

**Comment by krammnic** [comment]
Added dataset processing, now we need to add correct KL divergence calculation

**Comment by krammnic** [comment]
@SalmanMohammadi Are we supporting different ref_model in current DPO recipe? As I see, we are not: 

```
with torch.no_grad(), disable_adapter(self._model):
      (
          reference_chosen_log_probs,
          reference_rejected_log_probs,
                            _,
                            _,
      ) = self.concatenated_forward(self._model, batch)
```

**Comment by SalmanMohammadi** [comment]
> @SalmanMohammadi Are we supporting different ref_model in current DPO recipe? As I see, we are not:
> 
> ```
> with torch.no_grad(), disable_adapter(self._model):
>       (
>           reference_chosen_log_probs,
>           reference_rejected_log_probs,
>                             _,
>                             _,
>       ) = self.concatenated_forward(self._model, batch)
> ```

What do you mean by a different ref model? Because we're fine-tuning with LoRA, we can just disable the LoRA adapters to recover the original reference model without having to keep a separate copy of the model in memory. The reference model should always be the base model we're finetuning from.

**Comment by krammnic** [comment]
> > @SalmanMohammadi Are we supporting different ref_model in current DPO recipe? As I see, we are not:
> > ```
> > with torch.no_grad(), disable_adapter(self._model):
> >       (
> >           reference_chosen_log_probs,
> >           reference_rejected_log_probs,
> >                             _,
> >                             _,
> >       ) = self.concatenated_forward(self._model, batch)
> > ```
> 
> What do you mean by a different ref model? Because we're fine-tuning with LoRA, we can just disable the LoRA adapters to recover the original reference model without having to keep a separate copy of the model in memory. The reference model should always be the base model we're finetuning from.

Actually in TRL there is possibility to set something different! 

 ```
ref_model (`PreTrainedModelWrapper`):
            Hugging Face transformer model with a casual language modelling head. Used for implicit reward computation and loss. If no
            reference model is provided, the trainer will create a reference model with the same architecture as the model to be optimized.
``` 

I'm not sure why they give this possibility either, so that why I've asked. Maybe there is special use case?

SalmanMohammadi:
Ah good catch! I'm not 100% sure either. In their example scripts they always initialize with the same model (https://github.com/huggingface/trl/blob/main/examples/scripts/kto.py).

krammnic:
Actually, it is pretty critical to DPO, for example according to https://arxiv.org/pdf/2407.13709

```
Our experiments reveal an interesting finding: stronger
reference models can indeed offer more benefits
than the SFT model, but only when they are compat-
ible with the model being fine-tuned
```

I'm pretty sure that it will be actual for classic DPO for example

krammnic:
I would consider this as a follow up, which we might need to add pretty soon

krammnic:
Tasks:

- [ ] Fix test
- [ ] Modify batch
- [ ] Do same changes to distributed recipe

PR #1866, opened 2024-10-18
[WIP] Quantization for generate_v2

joecummings:
#### Context
- [ ] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

#### Changelog
*

#### Test plan

#### UX

### DIFF
--- a/recipes/configs/llama2/generation_v2.yaml
+++ b/recipes/configs/llama2/generation_v2.yaml
@@ -9,6 +9,11 @@
 # Model arguments
 model:
   _component_: torchtune.models.llama2.llama2_7b
+# You can turn uncomment the following lines to enable quantization for faster inference
+# and potentially lower VRAM
+# quantization_method:
+#   _component_: torchao.quantization.quant_api.int4_weight_only # int4 weight only is a good balance
+#   use_hqq: False
 
 # Transform arguments
 tokenizer:
@@ -27,16 +32,16 @@ checkpointer:
   output_dir: ./
   model_type: LLAMA2
 
+# Generation arguments
+prompt:
+  system: You are a helpful and creative AI assistant.
+  user: What is the capital of France?
+max_new_tokens: 500
+temperature: 0.6 # 0.8 and 0.6 are popular values to try
+top_k: 300
+
 # Device
 device: cuda
 dtype: bf16
 seed: 1234
 log_level: INFO
-
-# Generation arguments
-prompt:
-  system: You are a helpful and creative AI assistant.
-  user: What is the capital of France?
-max_new_tokens: 200
-temperature: 0.6 # 0.8 and 0.6 are popular values to try
-top_k: 300
--- a/recipes/dev/generate_v2.py
+++ b/recipes/dev/generate_v2.py
@@ -17,6 +17,7 @@ from torchtune.data import load_image, Message, padded_collate_tiled_images_and_
 from torchtune.generation import sample
 
 from torchtune.modules.transforms import Transform
+from torchtune.training import compile_model
 
 
 class SingleTurnYAMLToMessages(Transform):
@@ -65,29 +66,37 @@ class InferenceRecipe:
 
     This *does not* currently support the following features:
         - torch.compile
-        - quantization through torchao
         - multi-GPU generation
         - batch generation
     """
 
     def __init__(self, cfg: DictConfig) -> None:
-        self._device = utils.get_device(device=cfg.device)
-        self._dtype = training.get_dtype(dtype=cfg.dtype, device=self._device)
-        self._logger = utils.get_logger(cfg.log_level)
+        self.device = utils.get_device(device=cfg.device)
+        self.dtype = training.get_dtype(dtype=cfg.dtype, device=self.device)
+        self.logger = utils.get_logger(cfg.log_level)
         training.set_seed(seed=cfg.seed)
 
     def setup(self, cfg: DictConfig) -> None:
         """Setup the model and transforms."""
-        # Load checkpointer and state_dict
+        # Load checkpointer
         _checkpointer = config.instantiate(cfg.checkpointer)
         _ckpt_dict = _checkpointer.load_checkpoint()
 
         # Instantiate model
-        with training.set_default_dtype(self._dtype), self._device:
+        with training.set_default_dtype(self.dtype), self.device:
             model = config.instantiate(cfg.model)
         model.load_state_dict(_ckpt_dict[training.MODEL_KEY])
+        self.logger.info(f"Model was initialized with precision {self.dtype}.")
+
+        # Quantize the model if specified
+        if cfg.get("quantization_method") is not None:
+            from torchao.quantization.quant_api import quantize_
+
+            quantization_method = config.instantiate(cfg.quantization_method)
+            compile_model(model)
+            quantize_(model, quantization_method, device=self.device)
+
         self.model = model
-        self._logger.info(f"Model was initialized with precision {self._dtype}.")
 
         # Instantiate transforms
         self.model_transform = config.instantiate(cfg.tokenizer)
@@ -105,13 +114,13 @@ class InferenceRecipe:
                 for p in itertools.chain(self.model.parameters(), self.model.buffers())
             ]
         )
-        self._logger.info(
+        self.logger.info(
             f"Time for inference: {total_time:.02f} sec total, {tokens_per_second:.02f} tokens/sec"
         )
-        self._logger.info(
+        self.logger.info(
             f"Bandwidth achieved: {model_size * tokens_per_second / 1e9:.02f} GB/s"
         )
-        self._logger.info(
+        self.logger.info(
             f"Max memory allocated: {torch.cuda.max_memory_allocated() / 1e9:.02f} GB"
         )
 
@@ -128,10 +137,10 @@ class InferenceRecipe:
         total_response_length = seq_len + cfg.max_new_tokens
 
         # 3. Setup KV cache
-        with self._device:
+        with self.device:
             self.model.setup_caches(
                 batch_size=1,
-                dtype=self._dtype,
+                dtype=self.dtype,
                 encoder_max_seq_len=(
                     self.model_transform.image_seq_len if is_multimodal_input else None
                 ),
@@ -143,7 +152,7 @@ class InferenceRecipe:
             torch.ones(
                 size=(total_response_length, total_response_length),
                 dtype=torch.bool,
-                device=self._device,
+                device=self.device,
             )
         )
         input_pos = torch.arange(total_response_length)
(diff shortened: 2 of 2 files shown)


recipes/dev/generate_v2.py
```diff
@@ -65,29 +66,37 @@ class InferenceRecipe:
...
-        - quantization through torchao
         - multi-GPU generation
         - batch generation
     """
 
     def __init__(self, cfg: DictConfig) -> None:
-        self._device = utils.get_device(device=cfg.device)
-        self._dtype = training.get_dtype(dtype=cfg.dtype, device=self._device)
-        self._logger = utils.get_logger(cfg.log_level)
+        self.device = utils.get_device(device=cfg.device)
```
joecummings, line 74: It's a public recipe, no need to be a "private" variable. 

cc @pbontrager

recipes/dev/generate_v2.py
```diff
@@ -65,29 +66,37 @@ class InferenceRecipe:
...
         # Instantiate model
-        with training.set_default_dtype(self._dtype), self._device:
+        with training.set_default_dtype(self.dtype), self.device:
             model = config.instantiate(cfg.model)
         model.load_state_dict(_ckpt_dict[training.MODEL_KEY])
+        self.logger.info(f"Model was initialized with precision {self.dtype}.")
+
+        # Quantize the model if specified
+        if cfg.get("quantization_method") is not None:
+            from torchao.quantization.quant_api import quantize_
```
joecummings, line 93: Lazily import torchao API

recipes/dev/generate_v2.py
```diff
@@ -65,29 +66,37 @@ class InferenceRecipe:
...
+
+            quantization_method = config.instantiate(cfg.quantization_method)
+            compile_model(model)
```
joecummings, line 96: Compiling the model is necessary for quantization to be really worth it

recipes/dev/generate_v2.py
```diff
@@ -155,20 +164,20 @@ def generate(self, cfg: DictConfig):
...
         batch["input_pos"] = input_pos[None, :seq_len]
-        utils.batch_to_device(batch, self._device)
+        utils.batch_to_device(batch, self.device)
 
         # 6. Prefill step
         generated_tokens = []
         t0 = time.perf_counter()
         logits = self.model(prompt, **batch)[:, -1]
         token = sample(logits, temperature=cfg.temperature, top_k=cfg.top_k)
+        t1 = time.perf_counter()
```
joecummings, line 179: Now that we might have a warmup run, we log this differently so the user can see how much help quantization / compilation helps.

recipes/dev/generate_v2.py
```diff
@@ -155,20 +164,20 @@ def generate(self, cfg: DictConfig):
                 [model_inputs], pad_direction="left", pad_max_images=1
             )
             batch["encoder_mask"] = batch["encoder_mask"][:, :seq_len]
-            prompt = batch.pop("tokens").to(self._device)
+            prompt = batch.pop("tokens").to(self.device)
         else:
-            prompt = torch.tensor(
-                model_inputs["tokens"], device=self._device
-            ).unsqueeze(0)
+            prompt = torch.tensor(model_inputs["tokens"], device=self.device)[None, :]
```
joecummings, line 169: I wanted this to fix on one line lol

tests/conftest.py
```diff
@@ -18,6 +19,13 @@
 CACHE_ARTIFACTS_SCRIPT_PATH = root + "/tests/cache_artifacts.sh"
 
 
+def pytest_sessionfinish():
```
joecummings, line 22: Compile tries to log a bunch of stuff using the `atexit` decorator.  However, pytest closes these logs before they finish so it throws an I/O error. 

This disables logging exceptions. Not sure if the right way to do it.

recipes/configs/llama2/generation_v2.yaml
```diff
@@ -27,16 +31,16 @@ checkpointer:
...
-dtype: bf16
-seed: 1234
-log_level: INFO
-
 # Generation arguments
 prompt:
   system: You are a helpful and creative AI assistant.
   user: What is the capital of France?
-max_new_tokens: 200
+max_new_tokens: 500
```
joecummings, line 38: Allow longer generation to really see the benefit of quant + compile.

joecummings:
@felipemello1 @ebsmothers Will this not pass on PyTorch 2.5 b/c of the issue with CUDNN? This test passes locally on PyTorch v2.5.1. 

Do we know when the patch will be released?

recipes/configs/llama2/generation_v2.yaml
```diff
@@ -9,6 +9,10 @@
 # Model arguments
 model:
   _component_: torchtune.models.llama2.llama2_7b
+# You can turn uncomment the following lines to enable quantization for faster inference and potentially lower VRAM
+# quantization_method:
+#   _component_: torchao.quantization.quant_api.int4_weight_only # int4_weight_only is a good balance of speed and memory
```
SalmanMohammadi, line 14: dumb q: so the torchtune.training.quantization API is just for QAT.. or we're not using it anymore?

recipes/dev/generate_v2.py
SalmanMohammadi: I'm curious whether compiling the model results in greater speedups than compiling the next-token-prediction fn like gptfast do

recipes/configs/llama2/generation_v2.yaml
SalmanMohammadi: I see you mentioned this in the PR description - if we're going to be using the torchao APIs instead it'd be good to follow up with an issue

SalmanMohammadi:
This looks overall sensible, but a few outstanding questions I have:
- What implications does this have for how we expose quantization APIs?
- What is going on with compile?
- Why is memory usage identical for non-PTQ, and the first run of PTQ?
- Why is it so slow? Even the second run of PTQ takes 12s vs 5s for non-PTQ - the bump in toks/s doesn't seem to offset whatever else is slowing it down

joecummings:
> * What implications does this have for how we expose quantization APIs?

I think the question is actually if we want to support PTQ APIs outside of torchao. If we do, we want want to opt for an [approach like Hugging Face's ](https://huggingface.co/docs/transformers/main/en/main_classes/quantization#quantization) wherein a config for a specific backend can be initialized. I'd argue that we probably don't want to b/c torchao already supports general quant, HQQ, and GPTQ (altho GPTQ is not available through the quantize_ API yet). Idk if this is too short sighted though.

> * What is going on with compile?

Not sure I understand the question. It's always slow during warmup run. 

> * Why is memory usage identical for non-PTQ, and PTQ? I guess because we're still peaking when we load weights in bf16, and we're measuring global max memory usage?

Exactly.

> * Why is it so slow? Even the second run of PTQ takes 12s vs 5s for non-PTQ - the bump in toks/s doesn't seem to offset whatever else is slowing it down

Not sure *what* is so slow, but I've reached out to the AO team to see if this is normal.

> * Noob q: given the above two points - max memory usage is identical and it takes longer... when would someone want to use this?

An excellent question. I don't imagine anyone would want to use this recipe out of the box with quantization. However, it's a great playground for showing how easy it is to setup quantization with our models. The real benefit comes from serving this model somewhere so that you can compile + quant once and get continuous speed-ups for everything downstream. Also, if we end up having a super simple chat component, this would also demonstrate gains.

andrewor14:
Are you seeing the slowdown for `int4_weight_only` specifically? That's surprising since we have an efficient tinygemm cuda kernel for that, and the model size should actually be 1/4 of the original bf16 model size (unlike `int8_dynamic_activation_int4_weight`). Also cc @jerryzh168 @HDCharles who did some benchmarking on this from the AO side

recipes/configs/llama2/generation_v2.yaml
```diff
@@ -9,6 +9,10 @@
...
+#   use_hqq: False # Turn on for more accurate results
```
andrewor14, line 15: @HDCharles is this true?

joecummings: Yeah sorry this was anecdotal.

joecummings:
> Are you seeing the slowdown for `int4_weight_only` specifically? 

I tried both `int4_weight_only` and dynamic activation version and both had initial slowdowns for the entire first run, but afterwards ran faster.

recipes/dev/generate_v2.py
jerryzh168: we should compile after `quantize_` for speedup actually

joecummings: Interesting! I was following the pattern from AO's README where the model is compiled first:

```python
model = torchao.autoquant(torch.compile(model, mode='max-autotune'))
```

Why should the model be compiled after quantization?

joecummings:
> > > Are you seeing the slowdown for `int4_weight_only` specifically?
> > 
> > 
> > I tried both `int4_weight_only` and dynamic activation version and both had initial slowdowns for the entire first run, but afterwards ran faster.
> 
> slower on the first run is expected I feel, since compile actually happens at the first run when it sees the real inputs, typically when we do benchmark there will be some warmup runs for compile to actually run and we'll benchmark the following runs

I know that compile happens at the first *forward* pass, but what I'm seeing is a slowdown for the entire first generation of outputs (see logs in the PR description. Is this expected?

recipes/dev/generate_v2.py
joecummings: @jerryzh168 Anecdotally, I don't see much difference in tok/sec (after first token) between putting compile first or second. Can you share some more details about which one is correct?

recipes/configs/llama2/generation_v2.yaml
```diff
@@ -9,6 +9,10 @@
 # Model arguments
 model:
   _component_: torchtune.models.llama2.llama2_7b
+# You can turn uncomment the following lines to enable quantization for faster inference and potentially lower VRAM
+# quantization_method:
+#   _component_: torchao.quantization.quant_api.int4_weight_only # int4_weight_only is a good balance of speed and memory
+#   use_hqq: False # Turn on to use Half-Quadratic Quantization
```
felipemello1, line 15: what does it mean? Can you add if it makes it faster/more accurate/less memory?

felipemello1: sorry, what i meant is that this should be made clear for the user in the comment :P

recipes/dev/generate_v2.py
jerryzh168: oh right now `quantize_` needs to compile after: https://github.com/pytorch/ao/blob/main/torchao/quantization/README.md#full-affine-quantization-flow-example

but `autoquant` will do compile first before calling `autoquant`

felipemello1: mode='max-autotune' will take a long time to compile. Is it worth it? We dont do it for training.

Its interesting that in AO's read it says to put compile first. Do we also do it for QLoRA?

jerryzh168: > I don't see much difference in tok/sec (after first token) between putting compile first or second.

I haven't tried calling `quantize_` after compile actually, maybe it would have the same effect as well, need to confirm

jerryzh168: > Its interesting that in AO's read it says to put compile first. Do we also do it for QLoRA?

This is only for autoquant actually, we also haven't tested QLoRA, I think @andrewor14 is taking a look now

If compile time is a concern, we are also thinking of just do autoquant before hand and save the model, but I'm still testing that path as well

Issue #1732, opened 2024-10-01
update config defaults dataset.packed and log_peak_memory_stats


felipemello1:
Most configs dont have dataset.packed exposed. We should set it to:

dataset:
	packed=False  # Set to true for great speed ups

Also, we dont log peak memory by default. If it has no impact in TPS, we should set it to True

log_peak_memory_stats=True

also add compile=False to configs that dont have them

joecummings:
Done

PR #1872, opened 2024-10-20
Expose packed: False, set log_peak_memory_stats: True, set compile: False

krammnic:
#### Context
- [ ] add a new feature
- [ ] fix a bug
- [X] other (Clean up)

#### Changelog
* #1732 

#### Test plan

#### UX

### DIFF
--- a/recipes/configs/code_llama2/7B_full_low_memory.yaml
+++ b/recipes/configs/code_llama2/7B_full_low_memory.yaml
@@ -45,7 +45,9 @@ resume_from_checkpoint: False
 
 # Dataset
 dataset:
+  packed: False # Set to true for great speed ups
   _component_: torchtune.datasets.alpaca_dataset
+
 seed: null
 shuffle: True
 
@@ -75,4 +77,4 @@ metric_logger:
   _component_: torchtune.training.metric_logging.DiskLogger
   log_dir: /tmp/CodeLlama-7b-hf/logs
 log_every_n_steps: 1
-log_peak_memory_stats: False
+log_peak_memory_stats: True
--- a/recipes/configs/dev/8B_full_experimental.yaml
+++ b/recipes/configs/dev/8B_full_experimental.yaml
@@ -26,6 +26,7 @@ tokenizer:
 
 # Dataset
 dataset:
+  packed: False # Set to true for great speed ups
   _component_: torchtune.datasets.alpaca_dataset
 seed: null
 shuffle: True
@@ -57,7 +58,7 @@ loss:
   _component_: torchtune.modules.loss.CEWithChunkedOutputLoss
 max_steps_per_epoch: null
 gradient_accumulation_steps: 1
-
+compile: False
 
 # Training env
 device: cuda
@@ -78,3 +79,4 @@ metric_logger:
   log_dir: ${output_dir}
 output_dir: /tmp/alpaca-llama3-finetune
 log_every_n_steps: null
+log_peak_memory_stats: True
--- a/recipes/configs/llama3_1/405B_qlora.yaml
+++ b/recipes/configs/llama3_1/405B_qlora.yaml
@@ -45,6 +45,7 @@ save_adapter_weights_only: True
 
 # Dataset and Sampler
 dataset:
+  packed: False # Set to true for great speed ups
   _component_: torchtune.datasets.alpaca_dataset
   train_on_input: True
 seed: null
--- a/recipes/configs/llama3_1/405B_qlora.yaml
+++ b/recipes/configs/llama3_1/405B_qlora.yaml
@@ -72,7 +72,7 @@ fsdp:
 epochs: 1
 max_steps_per_epoch: null
 gradient_accumulation_steps: 16
-compile: False # set it to True for better memory and performance
+compile=False  # pytorch compile, set to true for perf/memory improvement# set it to True for better memory and performance
 
 # Logging
 output_dir: /tmp/qlora_finetune_output
--- a/recipes/configs/llama3_1/405B_qlora.yaml
+++ b/recipes/configs/llama3_1/405B_qlora.yaml
@@ -72,7 +72,7 @@ fsdp:
 epochs: 1
 max_steps_per_epoch: null
 gradient_accumulation_steps: 16
-compile=False  # pytorch compile, set to true for perf/memory improvement# set it to True for better memory and performance
+compile: False # pytorch compile, set to true for perf/memory improvement
 
 # Logging
 output_dir: /tmp/qlora_finetune_output
--- a/recipes/configs/llama3_1/8B_full.yaml
+++ b/recipes/configs/llama3_1/8B_full.yaml
@@ -26,6 +26,7 @@ tokenizer:
 
 # Dataset
 dataset:
+  packed: False # Set to true for great speed ups
   _component_: torchtune.datasets.alpaca_dataset
 seed: null
 shuffle: True
@@ -60,7 +61,7 @@ loss:
   _component_: torchtune.modules.loss.CEWithChunkedOutputLoss
 max_steps_per_epoch: null
 gradient_accumulation_steps: 1
-
+compile: False
 
 # Training env
 device: cuda
@@ -79,4 +80,4 @@ metric_logger:
   log_dir: ${output_dir}
 output_dir: /tmp/full-llama3.1-finetune
 log_every_n_steps: 1
-log_peak_memory_stats: False
+log_peak_memory_stats: True
--- a/recipes/configs/llama3_1/8B_full.yaml
+++ b/recipes/configs/llama3_1/8B_full.yaml
@@ -69,7 +69,7 @@ device: cuda
 # Memory management
 enable_activation_checkpointing: True
 custom_sharded_layers: ['tok_embeddings', 'output']
-compile=False  # pytorch compile, set to true for perf/memory improvement# set it to True for better memory and performance
+compile: False # pytorch compile, set to true for perf/memory improvement
 
 # Reduced precision
 dtype: bf16
--- a/recipes/configs/llama3_1/8B_full.yaml
+++ b/recipes/configs/llama3_1/8B_full.yaml
@@ -69,7 +69,7 @@ device: cuda
 # Memory management
 enable_activation_checkpointing: True
 custom_sharded_layers: ['tok_embeddings', 'output']
-compile: False # set it to True for better memory and performance
+compile=False  # pytorch compile, set to true for perf/memory improvement# set it to True for better memory and performance
 
 # Reduced precision
 dtype: bf16
--- a/recipes/configs/llama3_1/8B_full_single_device.yaml
+++ b/recipes/configs/llama3_1/8B_full_single_device.yaml
@@ -28,6 +28,7 @@ tokenizer:
 
 # Dataset
 dataset:
+  packed: False # Set to true for great speed ups
   _component_: torchtune.datasets.alpaca_dataset
 seed: null
 shuffle: True
@@ -78,7 +79,7 @@ metric_logger:
   log_dir: ${output_dir}
 output_dir: /tmp/full-llama3.1-finetune
 log_every_n_steps: 1
-log_peak_memory_stats: False
+log_peak_memory_stats: True
 
 # Profiler (disabled)
 profiler:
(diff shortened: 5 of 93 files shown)


recipes/configs/code_llama2/7B_full_low_memory.yaml
```diff
@@ -45,7 +45,9 @@ resume_from_checkpoint: False
 
 # Dataset
 dataset:
+  packed: False # Set to true for great speed ups
```
joecummings, line 48: Huge nit: can we move this below the `_component_` declaration?

That way it reads more as an option for the specific builder.

recipes/configs/dev/8B_full_experimental.yaml
```diff
@@ -57,7 +58,7 @@ loss:
   _component_: torchtune.modules.loss.CEWithChunkedOutputLoss
 max_steps_per_epoch: null
 gradient_accumulation_steps: 1
-
+compile: False
```
joecummings, line 61: Should we be explicit that this will `torch.compile` the model and loss? `compile` seems like a vague name

joecummings (commented):
two nits: overall, big fan of this UX improvement

recipes/configs/code_llama2/7B_full_low_memory.yaml
krammnic: Yeah, though like this first time.. But I followed original declaration from issue:
```
dataset:
packed=False # Set to true for great speed ups
```
Will be fixed

recipes/configs/dev/8B_full_experimental.yaml
felipemello1: agreed. I think we can add a comment, similar to packed

"compile=False  # Set to true for perf/memory improvement"

wdyt?

krammnic: Add some comment there?

felipemello1:
> > @SalmanMohammadi , do we support compile and packed in PPO? If not, maybe we should not add those to these configs.
> 
> Compile... yes, kind of? It's going to be overhauled soon to properly support, it's very out-of-date, it doesn't even have hierarchical compilation.

ok, so compile is NOT a no-op, and we SHOULD have it in the configs. However, packed does NOT work with PPO, and should NOT be added to ppo configs. Is that right?

SalmanMohammadi:
> > > @SalmanMohammadi , do we support compile and packed in PPO? If not, maybe we should not add those to these configs.
> > 
> > Compile... yes, kind of? It's going to be overhauled soon to properly support, it's very out-of-date, it doesn't even have hierarchical compilation.
> 
> ok, so compile is NOT a no-op, and we SHOULD have it in the configs. However, packed does NOT work with PPO, and should NOT be added to ppo configs. Is that right?

CORRECT

felipemello1:
@krammnic , we also have to check if compile/packed work for the knowledge distillation recipe. If not, we need to remove it from the configs. In short, i know for sure that these work for LORA and Full finetuning recipes/configs.

SalmanMohammadi:
@krammnic packed should also be removed from all the DPO configs, please.

recipes/configs/dev/8B_full_experimental.yaml
krammnic: Ok, will add

krammnic:
> @krammnic , we also have to check if compile/packed work for the knowledge distillation recipe. If not, we need to remove it from the configs. In short, i know for sure that these work for LORA and Full finetuning recipes/configs.
Sure, will test then

krammnic:
> @krammnic packed should also be removed from all the DPO configs, please.

Done

krammnic:
Added, comment to compile: False

felipemello1:
We will need to update recipes to log memory. We are getting the error

```
FAILED tests/recipes/test_lora_dpo_single_device.py::TestLoRADPOSingleDeviceRecipe::test_save_and_load_merged_weights - ValueError: Logging memory stats is only supported on CUDA devices, got cpu
```

So where log_peak_memory_stats, we need to add "if device = 'cuda'" and add info "log.info("log_peak_memory_stats was se to True, however, training does not use cuda. Setting log_peak_memory_stats=False."

cc: @ebsmothers

krammnic:
> We will need to update recipes to log memory. We are getting the error
> 
> ```
> FAILED tests/recipes/test_lora_dpo_single_device.py::TestLoRADPOSingleDeviceRecipe::test_save_and_load_merged_weights - ValueError: Logging memory stats is only supported on CUDA devices, got cpu
> ```
> 
> So where log_peak_memory_stats, we need to add "if device = 'cuda'" and add info "log.info("log_peak_memory_stats was se to True, however, training does not use cuda. Setting log_peak_memory_stats=False."
> 
> cc: @ebsmothers

Sure, will be done!

felipemello1:
> We will need to update recipes to log memory. We are getting the error
> 
> ```
> FAILED tests/recipes/test_lora_dpo_single_device.py::TestLoRADPOSingleDeviceRecipe::test_save_and_load_merged_weights - ValueError: Logging memory stats is only supported on CUDA devices, got cpu
> ```
> 
> So where log_peak_memory_stats, we need to add "if device = 'cuda'" and add info "log.info("log_peak_memory_stats was se to True, however, training does not use cuda. Setting log_peak_memory_stats=False."
> 
> cc: @ebsmothers

edit: lets actually do this check in the __init__ of the recipe. In the future, we can move all of these checks to some function like "config_parse". We already have multiple of these checks in the init

SalmanMohammadi:
> > > We will need to update recipes to log memory. We are getting the error
> > > ```
> > > FAILED tests/recipes/test_lora_dpo_single_device.py::TestLoRADPOSingleDeviceRecipe::test_save_and_load_merged_weights - ValueError: Logging memory stats is only supported on CUDA devices, got cpu
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > So where log_peak_memory_stats, we need to add "if device = 'cuda'" and add info "log.info("log_peak_memory_stats was se to True, however, training does not use cuda. Setting log_peak_memory_stats=False."
> > > cc: @ebsmothers
> > 
> > 
> > edit: lets actually do this check in the **init** of the recipe. In the future, we can move all of these checks to some function like "config_parse". We already have multiple of these checks in the init
> 
> I believe we're already doing this in most recipes when the stats are logged - the DPO recipe hasn't been updated.

The DPO recipes uses:
```python
                        if self._log_peak_memory_stats:
                            log_dict.update(
                                training.get_memory_stats(device=self._device)
                            )
```

it should be
```python
                        if self._device.type == "cuda" and self._log_peak_memory_stats:
                            log_dict.update(
                                training.get_memory_stats(device=self._device)
                            )
```

krammnic:
Add required check:

```
if self._log_peak_memory_stats and self._device.type == "cuda":
     log.info(
                "log_peak_memory_stats was se to True, however, training does not use cuda. Setting log_peak_memory_stats=False."
     )
     self._log_peak_memory_stats = False
```

recipes/qat_distributed.py
```diff
@@ -127,6 +127,12 @@ def __init__(self, cfg: DictConfig) -> None:
         self._log_every_n_steps = cfg.get("log_every_n_steps", 1)
         self._log_peak_memory_stats = cfg.get("log_peak_memory_stats", False)
 
+        if self._log_peak_memory_stats and self._device.type == "cuda":
```
SalmanMohammadi, line 130: Thoughts on whether we actually need this? I realise we kind of fail "silently" at the moment by just not logging if we aren't running on CUDA https://github.com/pytorch/torchtune/blob/17ba37d281548e5d60215f741888066717ad5b3e/recipes/lora_finetune_single_device.py#L716-L719)

@joecummings @felipemello1 

As-is we're now doing duplicating this check, which isn't super clean. Personally I'd rather just make the check in the relevant logging util - but don't have to block on this.

felipemello1: let me get back to you on this @krammnic . Thanks for making all of these changes! :)

krammnic: I'm not really see the problem in 2 partially duplicating checks. The point is, that if we have such logic in train:

```
log_dict = {
    "loss": loss_to_log,
    "tokens_per_second_per_gpu": num_tokens / time_per_step,
}
if self._log_peak_memory_stats:
    log_dict.update(
        training.get_memory_stats(device=self._device)
    )
self._metric_logger.log_dict(
    log_dict,
    step=self.global_step,
)
```

We can't do anything better, can we? Check in `__init__` is about cuda and logging(once). Check in train probably should not be about "cuda"(there is no use case) and not about logging. I'm not sure if this should be in `_metric_logger` either

krammnic:
Fixed some nits. Probably should be fine

recipes/configs/llama3_1/8B_full.yaml
```diff
@@ -60,15 +61,15 @@ loss:
...
+compile: False
 
 # Training env
 device: cuda
 
 # Memory management
 enable_activation_checkpointing: True
 custom_sharded_layers: ['tok_embeddings', 'output']
-compile: False # set it to True for better memory and performance
+compile=False  # pytorch compile, set to true for perf/memory improvement# set it to True for better memory and performance
```
felipemello1, line 72: oops

recipes/configs/llama3_1/8B_full_single_device.yaml
```diff
@@ -61,7 +62,7 @@ loss:
 max_steps_per_epoch: null
 gradient_accumulation_steps: 1
 optimizer_in_bwd: True
-compile: False # set it to True for better memory and performance
+compile=False  # pytorch compile, set to true for perf/memory improvement# set it to True for better memory and performance
```
felipemello1, line 65: do you mind quickly double checking? Also, if you are using a script, maybe for a sanity check make just that compile/packed dont appear twice?

recipes/full_finetune_distributed.py
```diff
@@ -121,6 +121,12 @@ def __init__(self, cfg: DictConfig) -> None:
         self._log_every_n_steps = cfg.get("log_every_n_steps", 1)
         self._log_peak_memory_stats = cfg.get("log_peak_memory_stats", False)
 
+        if self._log_peak_memory_stats and self._device.type == "cuda":
+            log.info(
+                "log_peak_memory_stats was se to True, however, training does not use cuda. Setting log_peak_memory_stats=False."
```
felipemello1, line 126: i think that there are typos. Thats my fault, i guess you just copied/pasted what i wrote earlier.

krammnic: Yes))) I didn't double check because it pretty minor

recipes/configs/llama3_1/8B_full_single_device.yaml
RdoubleA: this should be a colon? `compile: False`

krammnic: Yes, obviously. Fixed

recipes/full_finetune_distributed.py
krammnic: Fixed typos

felipemello1: @krammnic sorry, just realized that the condition is wrong

recipes/configs/llama3_1/405B_qlora.yaml
```diff
@@ -71,7 +72,7 @@ fsdp:
 epochs: 1
 max_steps_per_epoch: null
 gradient_accumulation_steps: 16
-compile: False # set it to True for better memory and performance
+compile=False  # pytorch compile, set to true for perf/memory improvement# set it to True for better memory and performance
```
felipemello1, line 75: other configs still need fix :(

recipes/full_finetune_distributed.py
krammnic: Good catch! Yeah need to be fixed

recipes/configs/llama3_1/405B_qlora.yaml
krammnic: Fixed

krammnic:
Fixed all typos

felipemello1 (approved):
lgtm! thanks for the pr!

PR #1875, opened 2024-10-20
Normalize CE loss by total number of (non-padding) tokens

ebsmothers:
In honor of the day the ML community first discovered the fact that (x1 / n1) + (x2 / n2) != (x1 + x2) / (n1 + n2)

This PR changes how we calculate the loss when gradient accumulation is enabled. This way we'll get an exact match in loss curves with and without gradient accumulation.

# The approach

Keep a running tally of the number of unmasked tokens in the recipe. Don't actually change our loss implementations (so they are still normalized by number of non-padding tokens in a batch), but when we get a batch's loss in the recipe we now just multiply by the number of non-padding tokens in that batch to get the unnormalized loss. 

Previously we called `.backward()` after every batch (after dividing loss by # of grad accumulation steps). Now we can't do that because we need to accumulate all losses to do proper normalization. So we instead wait until it's time to step, divide our accumulated loss by the total number of tokens seen across all batches in the step, then call `loss.backward()`.

Note: as a side effect our tokens/sec now logs only non-padding tokens. So yes the tokens/sec we see in our logs will decrease but it will also now be more representative of meaningful throughput (and you won't have to listen to me complaining about misleading tokens/sec anymore).

# Test plan

Updated a bunch of recipe tests to explicitly test with gradient accumulation enabled. Note that previously these tests would fail as the loss values would not match (see e.g. [this comment](https://github.com/pytorch/torchtune/blob/8e013c22fd9645b97d1c2c0bf5855455bf185279/tests/recipes/test_full_finetune_single_device.py#L205-L210) in a test that we added specifically to check parity of gradient accumulation when all samples have the same sequence length), but now we get the same loss values regardless of whether or not gradient accumulation is enabled.

## E2E tests

For all E2E tests, we compare the following four cases:

1) batch size N, gradient accumulation disabled on main
2) batch size 1, N gradient accumulation steps on main
3) batch size N, gradient accumulation disabled on this PR
4) batch size 1, N gradient accumulation steps on this PR

We also change the logging of num_tokens_per_second on main to match what's in this PR for a fair comparison.

### Llama 3 8B full finetune, single device

TLDR: we get the same loss curves for cases (1), (3), and (4). The updated gradient accumulation logic also increases tokens/second and reduces peak allocated memory. Full wandb workspace [here](https://wandb.ai/ebs/llama3-8b-full-single)

#### Loss curves:

Screenshot 2024-10-25 at 2 00 24 PM

#### Peak allocated memory:

Screenshot 2024-10-25 at 2 00 57 PM

#### Tokens/sec:

Screenshot 2024-10-25 at 2 01 27 PM

### Qwen 2 1.5B with LoRA on two devices

TLDR: same loss curves for (1), (3) and (4). There is a slight increase in peak allocated memory, but also a pretty big jump in tokens/sec. 

#### Loss curves:
Screenshot 2024-10-25 at 2 38 23 PM

#### Peak allocated memory:
Screenshot 2024-10-25 at 2 39 33 PM

#### Tokens/sec:
Screenshot 2024-10-25 at 2 39 59 PM


torchtune/modules/loss/ce_chunked_output_loss.py
```diff
@@ -80,4 +80,4 @@ def forward(self, logits: List[torch.Tensor], labels: torch.Tensor) -> torch.Ten
         for logits_chunk, labels_chunk in zip(logits, labels):
             total_loss += self.compute_cross_entropy(logits_chunk, labels_chunk)
 
-        return total_loss / total_elements
+        return total_loss
```
joecummings, line 83: Isn't this unnormalized?

felipemello1 (approved):
lgtm! thanks for fixing that

Issue #1861, opened 2024-10-17
Remove PyTorch 2.5.0 checks


RdoubleA:
PyTorch 2.5.0 is officially [released](https://github.com/pytorch/pytorch/releases/tag/v2.5.0) which includes features such as FlexAttention that is now public API and other compile features. We can now remove the following checks:

- [ ] Flex attention requiring torch >= 2.5.0, can just remove this line: https://github.com/pytorch/torchtune/blob/7d29c21149ace4be9fdc2709a03c0a21ba191010/torchtune/utils/_import_guard.py#L13
- [ ] Switch to per layer compile by default, which brings down compile time from 10 mins to under 1 minute: https://github.com/pytorch/torchtune/blob/7d29c21149ace4be9fdc2709a03c0a21ba191010/torchtune/training/_compile.py#L45
- [ ] Update this flex attention related docstring to say "If flex attention is supported by the current hardware": https://github.com/pytorch/torchtune/blob/7d29c21149ace4be9fdc2709a03c0a21ba191010/torchtune/modules/attention_utils.py#L119
- [ ] Remove this check in reparametrize state dict hook: https://github.com/pytorch/torchtune/blob/7d29c21149ace4be9fdc2709a03c0a21ba191010/torchtune/modules/common_utils.py#L150
- [ ] Default use_streams to True in activation offloading: https://github.com/pytorch/torchtune/blob/7d29c21149ace4be9fdc2709a03c0a21ba191010/torchtune/training/_activation_offloading.py#L36 and update related docs: https://github.com/pytorch/torchtune/blob/7d29c21149ace4be9fdc2709a03c0a21ba191010/docs/source/tutorials/memory_optimizations.rst#L111
- [ ] Remove these decorators checking for 2.5 on flex attention tests: https://github.com/pytorch/torchtune/blob/7d29c21149ace4be9fdc2709a03c0a21ba191010/tests/torchtune/modules/test_attention_utils.py#L87
- [ ] Update lora distributed and lora single device recipe docstrings: https://github.com/pytorch/torchtune/blob/7d29c21149ace4be9fdc2709a03c0a21ba191010/recipes/lora_finetune_distributed.py#L77

JP-sDEV:
Hi @RdoubleA, can I be assigned this issue?

RdoubleA:
[...]

PR #1877, opened 2024-10-21
chore: remove PyTorch 2.5.0 checks

JP-sDEV:
#### Context
- [ ] add a new feature
- [ ] fix a bug

#1861 

#### Changelog
* modify documentation and checks for PyTorch 2.5.0

Note: Line 69 in `torchtune/torchtune/training/_activation_offloading.py`, should this `if` block be removed? or modified to check for pytorch `2.5.0`? - was not included in the issue to modify this block.
```
 if use_streams is None:
            # Default to True if an acceptable torch is installed (later nightly/version or from source)
            **self.use_streams = torch.__version__ >= "2.5.0.dev20240907"** 
```

#### Test plan

#### UX


docs/source/tutorials/memory_optimizations.rst
```diff
@@ -108,7 +108,7 @@ checkpointing, where all activations will either be recomputed later in the back
 
 To enable activation offloading, use the ``enable_activation_offloading`` config entry or flag
 in our lora finetuning single device recipe, e.g. ``enable_activation_offloading=True``. To allow
-usage of streams, make sure you are on a torch version later than PyTorch 2.5.0.dev20240907.
+usage of streams, make sure you are on a torch version equal to or later than PyTorch.
```
RdoubleA, line 111:
```suggestion
usage of streams, make sure you are on a torch version equal to or later than PyTorch 2.5.0.
```

torchtune/training/_activation_offloading.py
```diff
@@ -33,7 +33,7 @@ class OffloadActivations(saved_tensors_hooks):
 
         use_streams (Optional[bool]): Whether or not to use streams for performance optimization where
             the communications get overlapped with the computation. Requires a torch build
-            after torch-2.5.0.dev20240907. Default: True if a later torch build is found, else False.
+            after torch-2.5.0.]. Default: True.
```
RdoubleA, line 36:
```suggestion
            after torch-2.5.0. Default: True.
```

torchtune/training/_compile.py
```diff
@@ -42,23 +42,21 @@ def compile_model(
...
+            m.compile(backend=backend)
+    # Fallback for models that can't be per-layer compiled
+    if not torch_version_ge("2.5.0"):
         if verbose:
-            log.info(
-                """
-                Compiling full model with torch.compile...
-                For faster compile times via per-layer compile, please run on PyTorch nightlies.
-                """
+            log.warning(
```
RdoubleA, line 56: I'm not sure if we want to retain the fallback logic for older pytorch versions. If so, then the if-else should remain the same and only the warning message should be updated. any thoughts? @ebsmothers @felipemello1

RdoubleA (commented):
Thanks again for doing these - looks great. A few minor nits and one comment we should decide on.

JP-sDEV:
> Thanks again for doing these - looks great. A few minor nits and one comment we should decide on.

Thanks, will keep up with the updates!  Once finalized, let me know what updates are needed. Cheers! :)

torchtune/training/_compile.py
ebsmothers: Sorry just seeing this now. I think if we claim to not support PyTorch < 2.5 then we shouldn't leave in the full-model compile option at all. For the same reason I'm ambivalent about leaving in the log warning.. really if we want to check someone is at least on the latest stable PyTorch we should just do it in a single consolidated place. So not the end of the world to keep the warning in, but personally I'd just take it out.

ebsmothers: I vote remove it altogether. We claim to only support latest stable PyTorch so no need to keep this around. We can treat adding a consolidated PyTorch version check as a separate task

RdoubleA:
Hey @JP-sDEV, pinging this thread again to see if you're able to resolve the merge conflicts so we can get this landed. I believe we agreed that we can remove the fallback logic in `_compile.py`, so would be great to make that change as well.

JP-sDEV:
Hi @RdoubleA, apologies for the delay - will resolve the conflicts and remove the fallback logic first thing tomorrow morning (EST)

docs/source/tutorials/memory_optimizations.rst
RdoubleA: bumping this

torchtune/training/_activation_offloading.py
```diff
@@ -67,7 +67,7 @@ class OffloadActivations(saved_tensors_hooks):
     def __init__(
         self,
         use_pin_memory: bool = True,
-        use_streams: Optional[bool] = None,
+        use_streams: Optional[bool] = True,
         max_fwd_stash_size: int = 5,
         min_offload_size: int = 1024,
     ) -> None:
```
RdoubleA, line 73: let's remove the check below and make `use_streams: bool = True`

JP-sDEV: are you referring to this `use_streams` check found in `__init__`? 

```python
        if use_streams is None:
            # Default to True if an acceptable torch is installed (later nightly/version or from source)
            self.use_streams = torch.__version__ >= "2.5.0.dev20240907"
        else:
            self.use_streams = use_streams
```

torchtune/modules/attention_utils.py
```diff
@@ -115,9 +115,9 @@ def packed_block_causal_mask(
     seq_lens: List[torch.Tensor],
```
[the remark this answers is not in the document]
RdoubleA: the `_attention_call` is modified to support both flex and sdpa if `_SUPPORTS_FLEX_ATTENTION` on module initialization, but it still needs the capability to fallback to SDPA if a block mask is not passed during runtime. So this awkward logic should be retained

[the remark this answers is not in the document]
RdoubleA: > We should always be able to import BlockMask on stable, so we could move [this import](https://github.com/pytorch/torchtune/blob/4b6877a6ef31a1f987c27594eaf8fe467b5ab785/torchtune/modules/attention_utils.py#L20) out of the flex attention support check.

we'd still need to verify their hardware can run it

torchtune/training/_activation_offloading.py
JP-sDEV: I have noticed that `OffloadingActivations`  in `torchtune/training/_activation_offloading.py` still checks for torch version: `2.5.0.dev20240907`. Should this check also be removed? 

```python
        # for streaming
        if self.use_streams:
            if torch.__version__ < "2.5.0.dev20240907":
                raise RuntimeError(
                    "OffloadActivations with use_streams=True requires PyTorch 2.5.0.dev20240907 or later."
                )
```

RdoubleA: Yes, good catch. Let's remove that as well. I believe these may have been added after I put the issue up, or maybe I just missed it

JP-sDEV: sounds good, will update all `2.5.0.dev20240907` checks in the file

torchtune/modules/attention_utils.py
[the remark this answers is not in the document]
SalmanMohammadi: rightio, because we're returning a callable the check needs to happen outside. thanks for explaining!

SalmanMohammadi:
Can we update these checks here? Are we able to remove them?
https://github.com/pytorch/torchtune/blob/4b6877a6ef31a1f987c27594eaf8fe467b5ab785/recipes/qat_distributed.py#L139
https://github.com/pytorch/torchtune/blob/4b6877a6ef31a1f987c27594eaf8fe467b5ab785/recipes/full_finetune_distributed.py#L128

docs/source/tutorials/memory_optimizations.rst
SalmanMohammadi: I believe we can just revert the changes here to leave the file as-as since it's been updated in another PR @JP-sDEV

tests/torchtune/modules/test_attention_utils.py
```diff
@@ -82,10 +82,6 @@ def test_packed_block_causal_mask_sdpa(self, seq_lens):
         )
         torch.testing.assert_close(actual, expected)
 
-    @pytest.mark.skipif(
```
RdoubleA, line 85: Ok looks like we need to keep this check, in case the hardware that runs the gpu tests on GitHub CI does not support flex attention

RdoubleA:
@JP-sDEV thanks for pushing this through, just one more comment which should fix the failing gpu test

JP-sDEV:
> @JP-sDEV thanks for pushing this through, just one more comment which should fix the failing gpu test

I am reading the comments and reviewing the commits, and if I understand correctly - the changes needed for checking hardware compatibility is something outside this issue? 

If so, I'll merge with the main branch once resolved :)

RdoubleA:
The update is that we just need to keep the decorator that was removed on `test_packed_block_causalmask_flex`, then it should be good to go

RdoubleA (approved):

RdoubleA:
@JP-sDEV I went ahead and quickly made the change - thank you again for your help!

JP-sDEV:
No problem! Thank you for guiding me through the issue and for being patient! :)

PR #1883, opened 2024-10-22
Update imports after QAT was moved out of prototype

andrewor14:
**Summary:** https://github.com/pytorch/ao/pull/1091 moved QAT out of prototype in torchao. This is a BC-breaking change so torchtune also needs to update its QAT imports. Additionally, after https://github.com/pytorch/ao/issues/987 we decided that QAT in torchao will use module swaps to insert fake quantizes, so there is no need to have a separate module swap quantizer, so this commit removes the `*ModuleSwapQuantizer` option.

**Test Plan:**
pytest -m integration_test tests/recipes/test_qat_distributed.py should work

### DIFF
--- a/torchtune/training/quantization.py
+++ b/torchtune/training/quantization.py
@@ -5,6 +5,7 @@
 # LICENSE file in the root directory of this source tree.
 
 from typing import Callable, Optional
+from warnings import warn
 
 from torchtune.utils._import_guard import _USE_NEW_TENSOR_CORE_TILED_LAYOUT_API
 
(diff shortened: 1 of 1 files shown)


torchtune/utils/_import_guard.py
```diff
@@ -20,7 +20,7 @@
 _USE_NEW_TENSOR_CORE_TILED_LAYOUT_API = _is_fbcode() or (
     not _is_fbcode()
     and (
-        ("dev" not in torchao_version and torchao_version >= "0.6.0")
+        ("dev" not in torchao_version and torchao_version >= "0.7.0")
```
ebsmothers, line 23: So the new API is not in 0.6.1? I thought it landed before that release

torchtune/training/quantization.py
```diff
@@ -142,15 +153,20 @@ def get_quantizer_mode(quantizer: Optional[Callable]) -> Optional[str]:
     Currently supported:
 
     - :class:`~torchao.quantization.quant_api.Int8DynActInt4WeightQuantizer`: "8da4w" (requires ``torch>=2.3.0``)
-    - :class:`~torchao.quantization.prototype.qat.Int8DynActInt4WeightQATQuantizer`: "8da4w-qat" (requires ``torch>=2.4.0``)
+    - :class:`~torchao.quantization.qat.Int8DynActInt4WeightQATQuantizer`: "8da4w-qat" (requires ``torch>=2.4.0``)
```
ebsmothers, line 156: I think we can fully remove both of these torch requirements as we assume stable PyTorch (i.e. >= 2.5)

torchtune/training/quantization.py
```diff
@@ -142,15 +153,20 @@ def get_quantizer_mode(quantizer: Optional[Callable]) -> Optional[str]:
...
         quantizer (Optional[Callable]): A callable object that implements the `quantize` method.
 
     Returns:
         Optional[str]: The quantization mode.
     """
-    return _quantizer_to_mode.get(type(quantizer), None)
+    mode = _quantizer_to_mode.get(type(quantizer), None)
+    if mode is not None and "module-swap" in mode:
+        warn(
+            "*QuantizerModuleSwap is deprecated. Please use the version without 'ModuleSwap' instead"
```
ebsmothers, line 167: nit: should be able to just give the exact classes in the warn message using e.g. `f"{quantizer.__class__.__name__}"` (or something like that)

ebsmothers (approved):
Thanks for the update! Couple small comments but otherwise looks good

torchtune/training/quantization.py
andrewor14: hmm I don't think that's possible given we just assign the classes today, e.g.:
```
>>> class A: pass
... 
>>> B = A
>>> A().__class__.__name__
'A'
>>> B().__class__.__name__
'A'
```

torchtune/utils/_import_guard.py
andrewor14: Yeah it's not in 0.6.x

andrewor14:
@joecummings @ebsmothers good to land?

Issue #1865, opened 2024-10-18
ForwardKLWithChunkedOutputLoss should be averaged over total_non_pad_tokens


moussaKam:
Hi there, 

Currently the loss ForwardKLWithChunkedOutputLoss  is averaged over the chunks which I think is not the right way to do it since there are chunks that contains only pad tokens. [line](https://github.com/pytorch/torchtune/blob/f8073ed24e8e82e74ac923a8134cd66cd6067ad5/torchtune/modules/loss/kd_losses.py#L132C1-L132C55)

I think the right way to do it is to average over the total non pad tokens. I did the fix [here](https://github.com/moussaKam/torchtune/tree/kd_loss_avg_tokens) but couldn't open a pull request because my commit didn't pass the test since the loss did not match with the expected loss.

moussaKam:
Ok I will run some experiments when I have some time a get back to you

joecummings:
Closed by #1885

PR #1885, opened 2024-10-23
Kd_loss avg over tokens

moussaKam:
#### Context
- [ ] add a new feature
- [x] fix a bug
- [ ] other (please add here)

#### Changelog
Follow up to issue #1865
In ForwardKLWithChunkedOutputLoss  the losss is averaged over non-masked tokens instead of chunks
In this [notebook](https://colab.research.google.com/drive/1r6k1VGRKUK76aNTzSLlFrRfGd7BX3zk2?usp=sharing) I show an example.

I ran some experiments to compare the old and the new loss. I followed the tutorial [here](https://pytorch.org/torchtune/stable/tutorials/llama_kd_tutorial.html)

here's the two commands for the training
```
CUDA_VISIBLE_DEVICES=7 tune run knowledge_distillation_single_device --config llama3_2/knowledge_distillation_single_device kd_loss._component_=torchtune.modules.loss.OldForwardKLWithChunkedOutputLoss batch_size=16 gradient_accumulation_steps=8 enable_activation_checkpointing=False metric_logger.log_dir=/tmp/kd_output_old profiler.output_dir=/tmp/kd_output/profiling_outputs_old checkpointer.output_dir=/tmp/Llama-3.2-1B-Instruct_old/
```

```
CUDA_VISIBLE_DEVICES=6 tune run knowledge_distillation_single_device --config llama3_2/knowledge_distillation_single_device kd_loss._component_=torchtune.modules.loss.FixedForwardKLWithChunkedOutputLoss batch_size=16 gradient_accumulation_steps=8 enable_activation_checkpointing=False metric_logger.log_dir=/tmp/kd_output_fixed profiler.output_dir=/tmp/kd_output/profiling_outputs_fixed checkpointer.output_dir=/tmp/Llama-3.2-1B-Instruct_fixed/
```
The results were as follow: 
| Dataset        | Metric       | Old Loss| Fixed Loss|
|----------------|--------------|-----------|-----------|
| TruthfulQA     | mc1          | 0.2766    | 0.2778    |
| TruthfulQA     | mc2          | 0.4348    | 0.4381    |
| HellaSwag      | acc          | 0.4549    | 0.4542    |
| HellaSwag      | acc_norm     | 0.6103    | 0.6113    |
| CommonSenseQA  | acc          | 0.5553    | 0.5594    |

Please let me know if further work should be done

#### Test plan

#### UX

### DIFF
--- a/torchtune/modules/loss/kd_losses.py
+++ b/torchtune/modules/loss/kd_losses.py
@@ -30,6 +30,7 @@ class ForwardKLLoss(torch.nn.Module):
         student_logits: torch.Tensor,
         teacher_logits: torch.Tensor,
         labels: torch.Tensor,
+        normalize: bool = True,
     ) -> torch.Tensor:
         """
         Args:
@@ -39,6 +40,7 @@ class ForwardKLLoss(torch.nn.Module):
                 (batch_size*num_tokens, vocab_size).
             labels (torch.Tensor): Ground truth labels of shape
                 (batch_size, vocab_size).
+            normalize (bool): Whether to normalize the loss by the number of unmasked elements.
 
         Returns:
             torch.Tensor: KL divergence loss of shape (1,).
@@ -50,6 +52,8 @@ class ForwardKLLoss(torch.nn.Module):
         prod_probs = torch.masked_fill(teacher_prob * student_logprob, inf_mask, 0)
         x = torch.sum(prod_probs, dim=-1).view(-1)
         mask = (labels != self.ignore_index).int()
+        if not normalize:
+            return -torch.sum(x * mask.view(-1), dim=0)
         if torch.sum(mask.view(-1), dim=0) == 0:
             return torch.tensor(0.0, device=x.device)
         return -torch.sum(x * mask.view(-1), dim=0) / torch.sum(mask.view(-1), dim=0)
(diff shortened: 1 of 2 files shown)


joecummings:
Thanks for the PR @moussaKam - I'll take a look later today and I've tagged @lindawangg, who implemented the first version of KD

torchtune/modules/loss/kd_losses.py
```diff
@@ -78,7 +78,23 @@ def __init__(self, num_output_chunks: int = 8, ignore_index: int = -100):
         super().__init__()
         self.num_output_chunks = num_output_chunks
         self.ignore_index = ignore_index
-        self.fkl_loss = ForwardKLLoss(ignore_index)
+
+    def non_chunked_forward_kl_loss(
```
joecummings, line 82: Why did you opt to re-create the `ForwardKLLoss` within the function instead of as another class that just returns the unnormalized calculation? I see a couple of options:

1. Re-use the `ForwardKLLoss` method but add a param called `normalize`, which will normalize the loss by non masked tokens
2. Create a new class called `ForwardKLLossUnnormalized` that returns the loss directly to be used in this implementation

All of these options assume we have the need to keep the original `ForwardKLLoss`. I don't know enough to say that we don't, but figured I'd ask if we might want to keep it or if we can just delete.

joecummings (commented):
One design consideration, but looks good.

moussaKam:
@ joecummings you're right it's better to use the existing class by adding a normalize param. Let me know if the modifications are good. 

As for `ForwardKLLoss` I think we should keep it because it's faster if enough memory is available.

torchtune/modules/loss/kd_losses.py
```diff
@@ -50,6 +52,8 @@ def forward(
         prod_probs = torch.masked_fill(teacher_prob * student_logprob, inf_mask, 0)
         x = torch.sum(prod_probs, dim=-1).view(-1)
         mask = (labels != self.ignore_index).int()
+        if not normalize:
+            return -torch.sum(x * mask.view(-1), dim=0)
         if torch.sum(mask.view(-1), dim=0) == 0:
             return torch.tensor(0.0, device=x.device)
         return -torch.sum(x * mask.view(-1), dim=0) / torch.sum(mask.view(-1), dim=0)
```
SalmanMohammadi, line 59: FWIW we have `masked_mean` which we use in other parts of the codebase for this https://github.com/pytorch/torchtune/blob/17ba37d281548e5d60215f741888066717ad5b3e/torchtune/rlhf/rewards.py#L101 but not hugely fussed about whether it's used here.

lindawangg (approved):
Change makes sense to me. Thanks!

xuwenxinedu:
I'm confused about the annotation 
in line 41 : labels (torch.Tensor): Ground truth labels of shape (batch_size, vocab_size) .
and line101: labels (torch.Tensor): Ground truth labels of shape (batch_size, num_tokens).
Why is it different?

moussaKam:
Hi @xuwenxinedu looks like a typo, it should be num_tokens in both cases

PR #1889, opened 2024-10-23
[bug] fix sharding multimodal

felipemello1:
#### Context
- [ ] add a new feature
- [x ] fix a bug
- [ ] other (please add here)

- When sharding a model, we had a function that checked did: layer.split("."), and check if "layer" was at idx=0 and "id" was at idx[1].
- Multimodal breaks this rule, by adding decoder.layer.i and encoder.layer.i.
- Therefore, layers weren't being properly sharded, greatly increasing the amount of memory necessary

#### Changelog
- updates the rule to look backwards: id has index [-1], layers has index [-2].
- Raises value error if no layer is found, so we can minimize the chance of this happening in the future

#### Test plan
before
image

now
image

image


recipes/qat_distributed.py
```diff
@@ -525,14 +530,16 @@ def _setup_data(
...
+            collate_fn=(
+                partial(
+                    padded_collate_sft,
+                    padding_idx=self._tokenizer.pad_id,
+                    ignore_idx=self._loss_fn.ignore_index,
+                )
+                if not packed
+                else partial(
+                    padded_collate_packed,
+                )
```
felipemello1, line 542: precommit hook

recipes/qat_distributed.py
```diff
@@ -233,9 +233,11 @@ def setup(self, cfg: DictConfig) -> None:
...
         self._optimizer = self._setup_optimizer(
             cfg_optimizer=cfg.optimizer,
-            opt_state_dict=checkpoint_dict[training.OPT_KEY]
-            if self._resume_from_checkpoint
-            else None,
+            opt_state_dict=(
+                checkpoint_dict[training.OPT_KEY]
+                if self._resume_from_checkpoint
+                else None
+            ),
```
felipemello1, line 240: precommit hook

recipes/qat_distributed.py
ebsmothers: Yeah what is the deal with these? Seems like collate_fn gets updated on every PR.. we need to figure out what's going on with our linter that's causing this to keep happening

torchtune/training/_distributed.py
```diff
@@ -608,16 +608,25 @@ def shard_model(
             the forward pass. Setting this to True corresponds to the FULL_SHARD sharding strategy
             from FSDP1, while setting it to False corresponds to the SHARD_GRAD_OP sharding strategy.
 
+    Raises:
```
ebsmothers, line 611: Can we add a follow-up task to write a unit test for this function? It's pretty straightforward to test it and given how heavily we leverage it I don't love that it's currently untested

ebsmothers (approved):
Please also run a non-multimodal recipe as a sanity check. Ideally we would also add a unit test here but given the magnitude of the fix I won't block on that

recipes/full_finetune_distributed.py
```diff
@@ -404,15 +404,18 @@ def _setup_model(
...
         # But directly using the name is more concise
-        def _is_layer_fqn(s: str) -> bool:
+        def _is_layer_name(name: str, module: nn.Module) -> bool:
             """
             Return True for layers.i and False for all other module names
             Covers sharding for both AC-wrapped and non-AC-wrapped modules in one shot
             """
-            s_list = s.split(".")
-            return len(s_list) == 2 and s_list[0] == "layers" and str.isdigit(s_list[1])
+            name_list = name.split(".")
```
rochitsapra, line 412: 1/ Should we not be using some sort of struct / dataclass to represent layer names consistently across the code so that such bugs can be prevented?

2/ Can we at least move this logic to a util that can be used across recipes rather than having the same logic in 4 places?

ebsmothers: Yeah agreed on both of these points. On the first one, I think the proper thing to do is to use the module rather than the layer name. Activation checkpointing makes this a bit of a headache cause it modifies the module (so we can't just do something like `isinstance(m, TransformerDecoderLayer)`). So actually this string split version was just a hack to handle both cases in one go and is (clearly as evidenced by this very bug) not the scalable way to do it

felipemello1: 2/ 100%, i think we have other opportunities like this in our recipes, to reduce tech debt. I will add the utility tomorrow morning.

1/ this is harder. If we have a robust way to tests all of our flags with different models/recipes, that would be easy. But as it is now, i would have to test them manually.

torchtune/training/_distributed.py
```diff
@@ -608,16 +679,25 @@ def shard_model(
...
+        ValueError: If no layer modules were sharded. Please check if shard conditions are working as expected.
```
ebsmothers, line 683: nit: the 2nd sentence is just a copy-paste from the error message? Would just remove it

torchtune/training/_distributed.py
```diff
@@ -583,6 +583,77 @@ def llama3_wrap(module: nn.Module, recurse: bool, **kwargs):
...
+        return False
+    else:
+        return name_list[-2] == "layers" and str.isdigit(name_list[-1])
+
+
+def shard_condition_exact_match(
+    name: str, module: nn.Module, names_to_match: List[str], *args, **kwargs
+) -> bool:
+    """
+    Returs True for modules whose name exactly matches the given name. This is a helper
```
ebsmothers, line 629:
```suggestion
    Returns True for modules whose name exactly matches the given name. This is a helper
```

torchtune/training/_distributed.py
```diff
@@ -583,6 +583,77 @@ def llama3_wrap(module: nn.Module, recurse: bool, **kwargs):
...
+        would be sharded separately.
+    """
+    name_list = name.split(".")
+    if len(name_list) < 2:
+        return False
+    else:
+        return name_list[-2] == "layers" and str.isdigit(name_list[-1])
+
+
+def shard_condition_exact_match(
```
ebsmothers, line 625: I know we wanna move to utilities, but if the utility is just `return x in y` I think it's overkill

ebsmothers: So I would either (a) scrap this entirely or (b) write a utility like `get_shard_conditions` that handles everything and call only that from the recipe.

recipes/lora_dpo_distributed.py
```diff
@@ -290,6 +290,7 @@ def _setup_model(
         fsdp_cpu_offload: bool,
         reshard_after_forward: bool,
         base_model_state_dict: Dict[str, Any],
+        custom_sharded_layers: Optional[List[str]] = None,
```
ebsmothers, line 293: I'm not sure about this.. I know we wanna keep recipes in sync but if we aren't actually using this in any LoRA recipes why do we need to add it?

recipes/qat_distributed.py
```diff
@@ -363,10 +365,10 @@ def _setup_model(
         self,
         cfg_model: DictConfig,
         enable_activation_checkpointing: bool,
-        custom_sharded_layers: Optional[List[str]],
         fsdp_cpu_offload: bool,
         reshard_after_forward: bool,
         model_state_dict: Dict[str, Any],
+        custom_sharded_layers: Optional[List[str]] = None,
```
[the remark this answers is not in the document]
felipemello1: Because its has a type hint "Optional", and it is optional, because we check if its None. Therefore, if its optional, then the default value should be None, regardless of what happens upstream :P

torchtune/training/_distributed.py
felipemello1: implemented (b)

PR #1894, opened 2024-10-24
[Bug] Added the missing torchao dependency for dev setup

smujjiga:
#### Context
- [ ] add a new feature
- [x] fix a bug
- [ ] other (please add here)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -55,6 +55,7 @@ dev = [
     "urllib3<2.0.0",
     "wandb",
     "expecttest",
+    "torchao",
 ]
 
 [tool.setuptools.dynamic]


ebsmothers:
Hi @smujjiga thanks for the PR! This is not a bug but a deliberate design decision: we recommend people to install torchao as part of their PyTorch install command before running e.g. pip install torchtune[dev]. 

This may seem a bit strange, but it’s for a good reason. By including torchao in our pyproject.toml like this, it will always install their latest stable release. But this may not always be compatible with PyTorch nightly releases, which we sometimes rely on for the latest and greatest distributed APIs, perf improvements, etc. By doing it this way you’re not locked in to a specific install — you can install stable PyTorch and stable torchao, or you can install the nightlies of both libraries for the latest features.

Hope this rationale makes sense! Please let me know if anything is unclear though.

PR #1899, opened 2024-10-24
Block on Python 3.13 version

joecummings:
The worst hack I've written so far in my life. Github Actions is a (useful) abomination. 

**Goal**: Don't run nightly builds on Python 3.13. We depend on HF Datasets, which depends on PyArrow, who haven't released a Python 3.13 wheel. It is expected to come out soon in version v18. Therefore, Python 3.13 builds break.

Simple enough, right? 

WRONG

SO WRONG

UNBELIEVABLY WRONG.

**Attempt 1**: Figure out a way to make the `build` job conditional on whether the matrix entry way for Python 3.13. Eight hours later, it's pretty clear that this isn't possible (through torchtune) b/c the `build` job is controlled by dev-infra and we just pass in the entire matrix and it handles the actual execution. 

**Attempt 2**: Filter the matrix generated by the `generate-matrix` job to remove all Python 3.13 entries. It was about three hours in that I figured out I could create a Python shell in the filtering job. This sped things up and after another couple hours, here we are. **But, Joe, how do you know it works if Python 3.13 jobs are only kicked off for nightly builds and you've done all your testing on this PR?** I toggled the filtering function to exclude Python 3.9 jobs (since those are the only ones kicked off on PRs) and confirmed that these jobs were not passed down to the `build` job. **Is that good enough?** No probably not, but what are you going to do about it? 

Final question - **Was this a good use of my time?**: Probably not. PyArrow will be out soon and we could've just had broken nightlies for Python 3.13 until then b/c no one is really using Python 3.13. But I know that my teammates will express great admiration and gratitude for making the little box green instead of red and that makes it alllllll worth it.

### DIFF
--- a/.github/workflows/build_linux_wheels.yaml
+++ b/.github/workflows/build_linux_wheels.yaml
@@ -39,11 +39,12 @@ jobs:
         run: |
           import os
           a = os.environ["A"]
-          new_dict = {"include": []}
-          for build in a["include"]:
-            if build["python_version"] != "3.13":
-              new_dict["include"].append(build)
-          print("hello")
+          print(a)
+          # new_dict = {"include": []}
+          # for build in a["include"]:
+          #   if build["python_version"] != "3.13":
+          #     new_dict["include"].append(build)
+          # print("hello")
   # build:
   #   if: ${{ needs.generate-matrix.outputs.python_version != '3.9' }}
   #   needs: generate-matrix
--- a/.github/workflows/build_linux_wheels.yaml
+++ b/.github/workflows/build_linux_wheels.yaml
@@ -33,7 +33,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: one
-        run: echo ${{ fromJSON(needs.generate-matrix.outputs) }}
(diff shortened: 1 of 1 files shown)


SalmanMohammadi (approved):

ebsmothers:
OK this is pretty impressive. Overall it looks reasonable, one thing I don't see is how you tested it though. Did you manually trigger the build job to confirm everything works now?

ebsmothers (approved):

.github/workflows/build_linux_wheels.yaml
```diff
@@ -27,18 +27,46 @@ jobs:
...
+
+          # Filter out any builds for 3.13
+          filtered_matrix = {"include": []}
+          for build in input_matrix["include"]:
+            if build["python_version"] != "3.13":
+              filtered_matrix["include"].append(build)
+
+          # Write the new matrix to the default outputs file
+          with open(github_output_file, "w") as handle:
+            handle.write(f"matrix={json.dumps(filtered_matrix)}")
```
SalmanMohammadi, line 58: this is awesome

PR #1904, opened 2024-10-25
Early fusion multimodal models

RdoubleA:
## Context
This enables Early Fusion models based on @pbontrager 's excellent original RFC on multimodal fusion models #1283. Since the RFC, we have already landed Deep Fusion model components. This PR discusses and implements the EarlyFusionModel component, along with testing and some lint updates.

Early fusion is simply a decoder with 1 or more extra encoders that merges their outputs with the token embeddings tables. The challenge lies in how we merge the embeddings and pass it into the decoder.

## Changelog
- Added EarlyFusionModel and tests
- Updated DeepFusionModel docstring, none of the code was touched except updated an incorrect typing
- Split `_fusion.py` into `_fusion_layers.py`, `_early_fusion.py`, and `_deep_fusion.py`
- Small update in peft utils

## Design
There is one design consideration I am seeking feedback on, and that is the EarlyFusionModel's usage of `self.decoder.tok_embeddings`. It accesses the decoder's token embedding table outside of the decoder forward because we need to merge the image encoder and any other modality encoder's output embeddings with the text embeddings (in this case just concatenate in sequence dimension):

```
embeds = self.tok_embeddings(tokens)
bsz, seq_len, embed_dim = embeds.shape
for encoder, inp in (encoder_input or {}).items():
    encoder_embeds = self.encoders[encoder](**inp)
    encoder_mask = (tokens == self.encoder_tokens[encoder]).expand(bsz, seq_len, embed_dim)
    embeds[encoder_mask] = encoder_embeds
    
output = self.decoder(embeds, mask, input_pos)
return output
```

Now, instead of token ids, we are passing in the merged embeddings directly into the decoder. But since we already used the text-only tok_embeddings from the decoder, we need to skip it when passing in the merged embeddings for the final decoder output. There are two ways we can do this.

#### State dict surgery
In the current code changes and suggested by the original RFC, we can manually set `self.decoder.tok_embeddings = nn.Identity()` so that it becomes a no-op when you forward pass with merged embeddings.
- This will require additional state dict hooks to make sure checkpoint saving and loading is still maintained despite the module change
- If a user wants to use the decoder outside of the EarlyFusionModule in the same script, they will need to restore the original tok_embeddings module from nn.Identity

#### Additional input_embeds kwarg
We could add a new keyword argument in `TransformerDecoder` forward for input embeddings. If this is passed in, we automatically skip the token embeddings:
```
h = self.tok_embeddings(tokens) if input_embeds is None else input_embeds
```
This way we don't need any state dict hooks or decoder modifications. However, we are polluting the decoder model forward with more arguments.

## Test plan
- Add forward pass decoder + encoders test
- Add forward pass decoder only test
- Add forward pass encoders only test, check to see embeddings are merged correctly in correct sequence placement
- Verify state dict hooks by loading a state dict with "decoder" labels then saving the same state dict
- Verify that variable number of encoder special tokens in each sample still works as expected


torchtune/modules/model_fusion/_fusion.py
```diff
@@ -477,3 +477,107 @@ def forward(
...
         return output
+
+class EarlyFusionModel(nn.Module):
+    def __init__(
+        self,
+        decoder: TransformerDecoder,
+        encoders: nn.ModuleDict,
+        encoder_tokens: Dict[str, int],
+        decoder_trainable: bool,
+        encoders_trainable: Dict[str, bool],
```
pbontrager, line 488: An error should be thrown if the different input dicts don't have the same keys

torchtune/modules/model_fusion/_fusion.py
```diff
@@ -477,3 +477,107 @@ def forward(
...
+        trainable_params = set()
+        for encoder, trainable in encoders_trainable.items():
+            if trainable:
+                trainable_params |= {
+                    f"encoders.{encoder}.{n}" for n, p in self.encoders[encoder].named_parameters()
+                }
+        if decoder_trainable:
+            trainable_params |= {
+                f"decoder.{n}" for n, p in self.decoder.named_parameters()
+            }
```
pbontrager, line 518: This is missing the logic and the parameter to make fusion modules trainable/untrainable

pbontrager (commented):
Thanks for putting this up Rafi! I left some comments on the implementation, but I'll leave the state dict discussion to others as we've already chatted on this.

torchtune/modules/model_fusion/_fusion.py
```diff
@@ -463,18 +500,252 @@ def forward(
...
+        >>> # decoder is a text-only TransformerDecoder (e.g. llama3_8b) with no modifications
+        >>> decoder = llama3_8b()
+        >>>
+        >>> # encoder is pre-trained encoder (e.g. clip_vit_224) with an added projection head
+        >>> projection_head = FeedForward(...)
+        >>> register_fusion_module(projection_head))
+        >>> encoders = {"image": nn.Sequential(clip_vit_224(), projection_head)}
+        >>>
+        >>> # EarlyFusionModel combines the encoder and decoder
+        >>> model = DeepFusionModel(decoder, encoders)
```
acisseJZhong, line 560: nit: EarlyFusionModel

acisseJZhong:
Thanks for the RFC, you made it very clear what's difference between early fusion and late fusion! 

About the design choice, I personally prefer Option 2 for the same reason as you mentioned. I think it's fine to "polluting" the decoder model forward a bit with some optional arguments for each modality. We might need something like 
```
h = self.tok_embeddings(tokens) 
if speech: 
   h[speech_mask] += speech_encoder(input) 
if image: 
   h[image_mask] += image_encoder(input)
```

torchtune/modules/model_fusion/_fusion.py
```diff
@@ -463,18 +500,252 @@ def forward(
...
+            else encoders_trainable
+        )
+
+        # A little surgery in the decoder to give the
+        # fusion module access to control the embeddings
+        # The alternative is to pass a special tok_embeddings
+        # module into TransformerDecoder builder that does the
+        # merging there
+        self.tok_embeddings = decoder.tok_embeddings
+        decoder.tok_embeddings = nn.Identity()
```
acisseJZhong, line 621: > If a user wants to use the decoder outside of the EarlyFusionModule in the same script, they will need to restore the original tok_embeddings module from nn.Identity

Why is this a concern? we only set `decoder.tok_embeddings` to identity within EarlyFusionModule? If outside EarlyFusionModule, we just use the normal decoder right?

torchtune/modules/model_fusion/_fusion.py
```diff
@@ -463,18 +500,252 @@ def forward(
...
+                    for n, p in self.encoders[encoder].named_parameters()
+                }
+        if decoder_trainable:
+            trainable_params |= {
+                f"decoder.{n}" for n, p in self.decoder.named_parameters()
+            }
+        if fusion_trainable:
+            trainable_params |= set(get_fusion_params(self))
+        else:
+            trainable_params -= set(get_fusion_params(self))
```
acisseJZhong, line 642: curious why would we need this?

RdoubleA: Not sure yet if we need this... once we have a better idea of the full architecture let's chat with @pbontrager to see which components need to be fusion modules

torchtune/modules/model_fusion/_fusion.py
RdoubleA: right, but if I passed in the decoder into EarlyFusionModule and modified a layer here, even if I use it separately the embedding layer will have been modified

ebsmothers:
11th hour comment on the open design question: in my mind there are nonzero UX costs to either approach. If we patch the decoder embeddings to `nn.Identity` we introduce additional indirection that is pretty trivial but also pretty non-obvious (I claim any state dict hook is non-obvious when first debugging the inevitable key mismatch error until you find the actual code pointer). On the plus side, we fully contain the blast radius to multimodal model code, and text-only users do not have to worry about it. Conversely, I know we don't want to just add a bunch of random arguments to `TransformerDecoder` forward, especially ones that are very specific to multimodal models.

Personally I really don't like state dict hooks for the reason I described above. As soon as something (inevitably) goes wrong, it will take a lot more debugging and head-banging-against-the-wall before the user realizes that things are being swapped out under the hood. So perhaps it's no surprise, but I vote for the simple and dumb thing: just add an extra parameter to TransformerDecoder forward. I know that may be controversial, but I like doing the obvious thing, and I like to think our users would appreciate that as well.

RdoubleA:
After extensive discussion offline, we decided to move ahead with the state dict hook approach. All current changes reflect this.

RdoubleA:
> What were the high level reasons for this, if I may ask?

1. Other fusion layers already use state dict hooks extensively, so there is some precedent
2. Reluctance to add more bloat to the TransformerDecoder forward
3. Other solutions would require decoders to be built differently to account for a fused embedding (i.e., a FusedEmbedding layer instead of nn.Embedding for tok_embeddings) and somewhat defeats the purpose of a separate EarlyFusionModel class

cc @ebsmothers @pbontrager

torchtune/modules/model_fusion/_fusion.py
```diff
@@ -463,18 +500,294 @@ def forward(
...
+
+        for encoder, inp in (encoder_input or {}).items():
+            # [bsz, num_encoder_tokens, embed_dim]
+            encoder_embeds = self.encoders[encoder](**inp)
+            # [bsz * num_encoder_tokens, embed_dim]
+            encoder_embeds = encoder_embeds.view(-1, embed_dim)
+            # [bsz, seq_len, 1]
+            encoder_mask = (tokens == self.encoder_tokens[encoder]).unsqueeze(-1)
+            # At locations where encoder token is found, replace with encoder embedding
+            fused_embeds = fused_embeds.masked_scatter(encoder_mask, encoder_embeds)
```
ebsmothers, line 790: Sorry I might be missing the point here.. is this changing the shape of fused_embeds?

torchtune/modules/model_fusion/_fusion.py
```diff
@@ -463,18 +500,294 @@ def forward(
...
+        # Place the text-only embeddings
+        fused_embeds = fused_embeds.masked_scatter(is_text.unsqueeze(-1), text_embeds)
+
+        for encoder, inp in (encoder_input or {}).items():
+            # [bsz, num_encoder_tokens, embed_dim]
+            encoder_embeds = self.encoders[encoder](**inp)
+            # [bsz * num_encoder_tokens, embed_dim]
+            encoder_embeds = encoder_embeds.view(-1, embed_dim)
+            # [bsz, seq_len, 1]
+            encoder_mask = (tokens == self.encoder_tokens[encoder]).unsqueeze(-1)
```
ebsmothers, line 788: Do we need to do any validation on encoder_tokens in the model? E.g. what if we have image embeddings but there is no image token in the token sequence? Do we expect that to be handled in the dataset? If so, we should probably call it out in the documentation somewhere

torchtune/modules/model_fusion/_fusion.py
```diff
@@ -463,18 +500,294 @@ def forward(
...
             mask=mask,
-            encoder_input=encoder_embed,
+            encoder_input=decoder_encoder_input,
             encoder_mask=encoder_mask,
             input_pos=input_pos,
         )
         return output
+
+
+class EarlyFusionModel(nn.Module):
```
ebsmothers, line 535: I don't have an incredibly strong opinion here but it might be clearest to put this in a separate file (e.g. deep_fusion.py and early_fusion.py)

torchtune/modules/model_fusion/_fusion.py
```diff
@@ -322,35 +327,63 @@ class DeepFusionModel(nn.Module):
...
+            raise ValueError(
+                f"Found mismatched keys in encoders and encoders_trainable. Got {encoders.keys()} and {encoders_trainable.keys()}."
+            )
+        # Currently, only a single encoder is supported, so user can only
+        # pass in a single key. When multiple encoders are
+        # supported, this can be removed.
+        if len(encoders.keys()) != 1:
+            raise ValueError(
+                f"DeepFusionModel only supports a single encoder. Got {len(encoders.keys())} encoders."
+            )
```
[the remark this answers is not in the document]
RdoubleA: It was mainly to maintain a consistent API between the two. but I don't have strong opinions here. I don't see any hard requirement to make the signatures match

torchtune/modules/model_fusion/_fusion.py
```diff
@@ -463,18 +500,294 @@ def forward(
...
+        >>> # encoder is pre-trained encoder (e.g. clip_vit_224) with an added projection head
+        >>> projection_head = FeedForward(...)
+        >>> register_fusion_module(projection_head))
+        >>> encoders = {"image": nn.Sequential(clip_vit_224(), projection_head)}
+        >>>
+        >>> # EarlyFusionModel combines the encoder and decoder
+        >>> model = EarlyFusionModel(decoder, encoders, encoder_tokens={"image": 128256})
+        >>>
+        >>> # Load full fused checkpoints
+        >>> model.load_state_dict(...)
```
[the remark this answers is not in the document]
RdoubleA: if the checkpoint is a single file you can load the entire model with encoders all at once. but I'm not sure if this will be the case for a model with multiple encoders or what the checkpoint UX would look like. I'm ok to remove this until we know for sure.

torchtune/modules/model_fusion/_fusion.py
RdoubleA: It is not, it is just placing the encoder embedding vectors in each instance of that encoder token in the fused_embeds

torchtune/modules/model_fusion/_fusion.py
RdoubleA: Good point, need to think about where this should be asserted. I would say probably in the transform or the dataset. We wouldn't want to forward pass the encoder if there's nowhere to use it. Although, within a batch you can have variable number of images per sample, so one sample may have zero images and another may have two.

torchtune/modules/model_fusion/_fusion.py
RdoubleA: yeah, could split to deep_fusion, early_fusion, and fusion_layers.py

torchtune/modules/model_fusion/_fusion.py
RdoubleA: I've added a ValueError just in case

RdoubleA: actually.... sometimes the number of encoder embeddings and num encoder tokens will not match, because there could be padding images. so in the dataset is probably the best place to assert this. will update the docstring here though to call this out

felipemello1 (approved):

PR #1913, opened 2024-10-28
Fix offloading saved tensor interaction with checkpointing

### LINKED ISSUE #1867: `OffloadActivations(use_streams=True)` producing NaN gradients: a tensor deletion data race
**cc**: @janeyx99 

When attempting to use `OffloadActivations(use_streams=True)` on my particular use case, I get NaN gradients. Here is a script capable of reproducing this behavior (requires ~3.45 GiB of VRAM):

```python
"""Demonstrate OffloadActivations NaN gradients from tensor deletion data race.

To run:

pip install torch==2.5.0 torchtune==0.3.1 torchao==0.5.0 transformers liger_kernel
CUDA_VISIBLE_DEVICES=0 python offload_activations_nan.py
"""

import contextlib
import functools
import subprocess

import liger_kernel.transformers
import torch
import torch.distributed.algorithms._checkpoint.checkpoint_wrapper
import torch.distributed.fsdp.wrap
import torch.optim
import torchtune.training
import transformers
import transformers.models.llama.modeling_llama
from torch import nn

BATCH_SIZE = 2
SEQUENCE_LENGTH = 2048
EMBEDDING_DIM = 4096


class MyModel(nn.Module):
    def __init__(self) -> None:
        super().__init__()
        self.weight = nn.Parameter(torch.randn(BATCH_SIZE, SEQUENCE_LENGTH, EMBEDDING_DIM, dtype=torch.float32))

    def forward(self, input: torch.Tensor) -> torch.Tensor:
        mask = torch.randint(0, 2, (BATCH_SIZE, SEQUENCE_LENGTH, 1), dtype=torch.bool, device=self.weight.device)
        return (input.to(torch.float32) * mask + ~mask * self.weight ** 2).to(input)


def main() -> None:
    liger_kernel.transformers.apply_liger_kernel_to_llama()

    torch.cuda.set_device(device := torch.device("cuda:0"))
    torch.set_default_dtype(dtype := torch.bfloat16)

    with device:
        config = transformers.LlamaConfig(num_hidden_layers=2)
        llama_model = transformers.LlamaForCausalLM(config).eval()
        for param in llama_model.parameters():
            param.requires_grad = False

        my_model = MyModel()

    auto_wrap_policy = functools.partial(
        torch.distributed.fsdp.wrap.transformer_auto_wrap_policy,
        transformer_layer_cls={transformers.models.llama.modeling_llama.LlamaDecoderLayer},
    )
    torch.distributed.algorithms._checkpoint.checkpoint_wrapper.apply_activation_checkpointing(
        llama_model,
        checkpoint_wrapper_fn=torch.distributed.algorithms._checkpoint.checkpoint_wrapper.checkpoint_wrapper,
        auto_wrap_policy=auto_wrap_policy,
    )

    optimizer = torch.optim.AdamW(params=my_model.parameters())

    vocab_size = llama_model.config.vocab_size
    input_ids = torch.randint(low=0, high=vocab_size, size=(BATCH_SIZE, SEQUENCE_LENGTH), device=device)

    for i in range(5, -1, -1):
        with (
            torch.profiler.profile(
                activities=[torch.profiler.ProfilerActivity.CPU, torch.profiler.ProfilerActivity.CUDA],
                profile_memory=True,
                record_shapes=True,
                with_stack=True,
            )
            if i == 0
            else contextlib.nullcontext()
        ) as profiler:

            # if i == 0:
            #     torch.cuda.memory._record_memory_history()

            with torchtune.training.OffloadActivations(use_streams=True):
                output = llama_model(
                    inputs_embeds=my_model(llama_model.get_input_embeddings()(input_ids)),
                    labels=input_ids,
                    use_cache=False,
                )

            output.loss.backward()

            if i != 0:  # this produces weird trace/snapshot artifacts, so skip it
                print(f"{i=}: {any(p.grad is not None and p.grad.isnan().any().item() for p in my_model.parameters())=}")

        optimizer.zero_grad(set_to_none=True)

    assert profiler is not None
    profiler.export_chrome_trace("./offload_activations.json")
    # torch.cuda.memory._dump_snapshot("offload_activations_nan_snapshot.pkl")
    # subprocess.Popen("python -m torch.cuda._memory_viz trace_plot offload_activations_nan_snapshot.pkl -o offload_activations_nan_snapshot.html".split())

    print(f"max_memory_allocated={torch.cuda.max_memory_allocated() / 2 ** 30} GiB")
    print("done.")


if __name__ == "__main__":
    main()

```

Observe that the gradients are not NaN when not using `OffloadActivations` or when using `OffloadActivations(use_streams=False)`.

I *believe* one of the key activations in this race is the boolean mask, which appears as a small `Memcpy HtoD (Pinned -> Device)` sliver in the comm stream.

Two key observations:

1. Removing `del self.bwd_tensor_stash[id]` in the `hook` registered on the autograd node in `unpack_tensor_with_streams` fixes the issue. The core problem seems to be with when the unpacked tensors are being deleted.

https://github.com/pytorch/torchtune/blob/3ca0d309c67ea996cc69f29691bc97ad7de00819/torchtune/training/_activation_offloading.py#L311-L315

2. Adding `self.s1.wait_stream(self.s0)` immediately after `self.s0.wait_stream(self.s1)` in `unpack_tensor_with_streams` fixes the issue. In fact, it produces the exact same relative ordering as `use_streams=False` or equivalently `maybe_gpu_tensor.to("cuda", non_blocking=False)` in `s0`.

https://github.com/pytorch/torchtune/blob/3ca0d309c67ea996cc69f29691bc97ad7de00819/torchtune/training/_activation_offloading.py#L279-L280

`self.s0.wait_stream(self.s1)` tells `s0` to wait until the copy is done on `s1` before starting its work to avoid working on uninitialized data. `self.s1.wait_stream(self.s0)` tells `s1` to wait until `s0` has finished all enqueued work and can start working on the unpacked tensor before `s1` can start its next communication operation.

While the above may not be optimal, it seems clear that delaying the deletion by synchronizing `s1` to `s0` is the solution. Our goal is to identify the earliest possible time that our unpacked tensors are safe to be deleted. It seems that `bwd_ev_stash` + `self.s0.record_event()` and `self.s1.wait_event(event)` is not achieving this exactly as planned. Are we recording these events in the exact right place or is there some other event we can record and wait on that we are missing? Or is the additional `wait_stream` the best we can do?

---

Here is script to calculate the percentage of communication overlap:

```python
def get_interval_intersections(a: list[list[float]], b: list[list[float]]) -> list[list[float]]:
    if not a or not b:
        return []
    i = 0
    j = 0
    result = []
    while i < len(a) and j < len(b):
        if a[i][1] < b[j][0]:
            i += 1
        elif b[j][1] < a[i][0]:
            j += 1
        else:
            start = max(a[i][0], b[j][0])
            end = min(a[i][1], b[j][1])
            result.append([start, end])
            if a[i][1] <= b[j][1]:
                i += 1
            else:
                j += 1
    return result


with open("./offload_activations.json", "rb") as f:
    trace = json.load(f)

comp_events = [event for event in trace["traceEvents"] if event.get("args", {}).get("stream") == 7]  # stream idx will depend on your trace
assert len(comp_events) != 0
comm_events = [event for event in trace["traceEvents"] if event.get("args", {}).get("stream") == 33]  # stream idx will depend on your trace
assert len(comm_events) != 0

comp_intervals = [[event["ts"], event["ts"] + event["dur"]] for event in comp_events]
comm_intervals  = [[event["ts"], event["ts"] + event["dur"]] for event in comm_events]

comm_time = sum(event["dur"] for event in nan_comm_cuda_events)
overlapped_comm_time = sum(interval[1] - interval[0] for interval in get_interval_intersections(comp_intervals, comm_intervals))
percent_comm_overlapped = overlapped_comm_time / comm_time
print(f"percent_comm_overlapped = {percent_comm_overlapped * 100:.2f}%")
```

The overlap on given example is really terrible (like 20% or something), but it is minimal edge case after all.

janeyx99:
#### Context
- [ ] add a new feature
- [x] fix a bug
- [ ] other (please add here)

Hopefully fixes #1867 once and for all

Needs https://github.com/pytorch/pytorch/pull/139109 to land!

#### Changelog
* Fixes checkpoint interaction with offloading. Prior, if an unpacked tensor is cached by activations checkpointing as a saved variable, it may be freed too early. This PR uses hooks (not currently in core yet, so we'll have to wait for that) to be able to coordinate this save properly.
* I should also add a test case, but that'll take me a while to contrive lol.

#### Test plan

#### UX


janeyx99:
Closing in favor of #1936

PR #1914, opened 2024-10-28
Patch github link in torchtune docs header

### LINKED ISSUE #1843: Point Github link at top right of documentation to pytorch/torchtune

ebsmothers:
Fixes #1843 

In our live docs we currently use the default PyTorch Sphinx theme for our header. This includes a Github icon that points to the main PyTorch repo by default, but for torchtune this isn't really where we want to point. This PR takes a similar approach to what is done in torchaudio [here](https://github.com/pytorch/audio/blob/main/docs/source/_templates/layout.html): adding a script to the footer to patch the href pointing to the PyTorch GH page. Unlike in torchaudio, however, we do not pin to an older version of the PyTorch Sphinx theme, so we identify the object to patch slightly differently.

Test plan:

Building docs locally and mousing over the GH icon we can see that it now points to the torchtune page.
 
Screenshot 2024-10-28 at 4 49 58 PM

### DIFF
--- a/docs/source/_templates/layout.html
+++ b/docs/source/_templates/layout.html
@@ -17,11 +17,11 @@
 
 <script type="text/javascript">
     var collapsedSections = []
-    </script>
-    {{ super() }}
-    <script type="text/javascript">
-      $(document).ready(function() {
-            $(".main-menu-item a[href='https://github.com/pytorch/pytorch']").attr("href", "https://github.com/pytorch/torchtune");
-      });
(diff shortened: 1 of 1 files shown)


RdoubleA (approved):

joecummings (approved):
He's an HTML GOD

svekars (approved):
It looks like it can be all indented to the left but otherwise looks good and seems to be working.

PR #1915, opened 2024-10-28
[ez] Fix README download example

RdoubleA:
The current download example command in the README will download the consolidated weights and not the safetensors, any any `tune run` on llama3.1 will error out. This PR updates to the correct command.

Thanks to @drisspg for pointing this out.


ebsmothers (approved):

drisspg (commented):
thank you :)

PR #1917, opened 2024-10-29
Restore backward after each batch for grad accum

ebsmothers:
The fix to normalize CE loss by total number of tokens in a step moved our backward call from per-batch to per-step. This means a bunch of activations are hanging around longer than they should and blowing up our memory. We should be able to call backward on the unnormalized values for each batch then manually scale the gradients just before optimizer step.

This is easy enough for our single-device recipes, but for our distributed recipes it's slightly more work. In fact it cannot be done in a way that is both (a) correct and (b) backwards compatible. If you want to know why this is the case, see the "Long digression" section below.

TLDR of the changes:

1) **Gradient accumulation logic changes:** We now (a) call backward on every batch's unnormalized loss, (b) accumulate num_tokens over every batch, (c) (distributed only) all_reduce num_tokens on the last batch prior to stepping, then (d) manually scale gradients prior to stepping with the optimizer. This will revert the memory regression introduced by #1875 but keep the correctness of the gradient accumulation logic.
2) **Logging changes**: If we are stepping based on total_loss / total_tokens (where total is over all batches and over all ranks), that's what should show up in the logs. Similarly, we can now use the number of tokens over all ranks instead of just rank 0 (though we still normalize tokens/sec to per GPU)
3) **Optimizer in backward:** This shouldn't be supported when gradient accumulation is enabled. Previously we didn't raise an explicit error about this (also it actually worked when we only call `.backward()` once per step, but we don't wanna do that). So now we'll raise an error (also I disabled a couple test cases that snuck in during this time period)
4) **Minor changes to KD and QAT recipes**: Integrate `utils.batch_to_device` into QAT recipe. This was causing num_tokens to show up on CPU instead of GPU because labels weren't moved until later.

# Test plan

All the single-device recipe tests succeed without any changes. However, for our distributed recipe tests some of the expected values need to change.

## Why are you changing the expected values?!

Again, see the "long digression" section below, but we have been normalizing our loss by the local number of tokens, not the number of tokens seen over all ranks. Is this a huge deal? Honestly probably not, but technically it's not correct. How do I know this version is correct? I've run the following command both on main and on this PR:

```
pytest -m integration_test tests/recipes/test_full_finetune_distributed.py -k 'test_loss[False-llama2/7B_full-llama2-hf-1-4]'
```

In both cases I added logging of # of tokens and grad for the token embeddings weight. On main I also commented out the loss normalization so that we can get the raw values. Diff for my changes on [main](https://www.diffchecker.com/CRwLPyRg/) (changes on this PR are just the identical logging).

Note that after the first iteration the results differ due to the difference in how the gradients are calculated, so looking at the logs from just the first iteration:

### On main

```
rank: 1, idx: 0, num_tokens: 12
rank: 0, idx: 0, num_tokens: 59
rank: 0, idx: 1, num_tokens: 60
rank: 1, idx: 1, num_tokens: 268
rank: 0, idx: 2, num_tokens: 14
rank: 1, idx: 2, num_tokens: 133
rank: 0, idx: 3, num_tokens: 11
rank: 1, idx: 3, num_tokens: 6
rank: 0, unnormalized grad: DTensor(local_tensor=-13.8838529586792, device_mesh=DeviceMesh('cuda', [0, 1]), placements=(Partial(sum),))
rank: 1, unnormalized grad: DTensor(local_tensor=1.0207021236419678, device_mesh=DeviceMesh('cuda', [0, 1]), placements=(Partial(sum),))
```

### On this PR

```
rank: 1, idx: 0, num_tokens: 12
rank: 0, idx: 0, num_tokens: 59
rank: 0, idx: 1, num_tokens: 60
rank: 1, idx: 1, num_tokens: 268
rank: 0, idx: 2, num_tokens: 14
rank: 1, idx: 2, num_tokens: 133
rank: 0, idx: 3, num_tokens: 11
rank: 1, idx: 3, num_tokens: 6
rank: 0, grad: DTensor(local_tensor=-0.02466048300266266, device_mesh=DeviceMesh('cuda', [0, 1]), placements=(Partial(sum),))
rank: 1, grad: DTensor(local_tensor=0.0018129688687622547 device_mesh=DeviceMesh('cuda', [0, 1]), placements=(Partial(sum),))
```

### But what does it all mean? 

In both the preceding snippets, the first step sees 12 + 59 + 60 + 268 + 14 + 133 + 11 + 6 = **563 tokens**. 

The grad value logged is just the sum of the elements in the tensor. On main this is -13.8838529586792 on rank 0 and 1.0207021236419678 on rank 1. On this PR it is -0.02466048300266266 on rank 0 and 0.0018129688687622547 on rank 1. In both cases, the value on this PR == the unnormalized value on main / 563, which is what we would expect

### Llama3 8B full finetune on 4 devices

Repro: 

```
tune run --nproc_per_node 4 full_finetune_distributed --config llama3/8B_full \
max_steps_per_epoch=500 gradient_accumulation_steps=4
```

Peak allocated memory drops substantially after this PR (back to where it was before the first gradient accumulation PR, but with actual correct loss calculation now).

Screenshot 2024-10-30 at 5 53 14 PM

### Llama 3.2 1B full finetune on 2 devices

```
tune run --nproc_per_node full_finetune_distributed --config llama3_2/1B_full max_steps_per_epoch=500
```

We can see that (a) peak allocated memory drops back down to where it was before #1875, (b) tokens per second is the same as #1875 (and faster than before), and (c) the loss curves look similar.

Screenshot 2024-10-30 at 5 57 13 PM

Screenshot 2024-10-30 at 6 00 45 PM

Screenshot 2024-10-30 at 6 01 40 PM

# Long digression: data parallel and gradient accumulation aren't so different

## First: on a single device

Let's consider a simple example of how to calculate the correctly-normalized CE loss on a single device (i.e. no data parallel) with two gradient accumulation steps. Say our *unnormalized* cross-entropy loss for the first batch is L1 and our (similarly unnormalized) loss for the second batch is L2, and we have n1 tokens in the first batch and n2 tokens in the second batch. Then the properly-normalized cross-entropy loss to step with will be (L1 + L2) / (n1 + n2). The naive approach taken in #1875 is to just accumulate a running sum of both loss and number of tokens, then call backward on the ratio running_loss / running_num_tokens just before optimizer step. The problem with this is that we only call backward once, so our activations stick around and blow up our memory. What can we do instead?

Sticking with the single device case, it's not too hard to fix this. Repeated calls to `.backward()` accumulate gradients, so e.g. the following two are numerically equivalent:

```
# Single backward call
loss_1 = model(batch_1)
loss_2 = model(batch_2)
summed_loss = loss_1 + loss_2
summed_loss.backward()
```
```
# Multiple backward calls
loss_1 = model(batch_1)
loss_1.backward()
loss_2 = model(batch_2)
loss_2.backward()
```

Then for the single device case we can actually take the second approach with L1 and L2. But remember that we still want to normalize by n1 + n2. This isn't too hard.. we can just manually scale the gradients ourselves just before the optimizer step (since grad(c\*X) = c\*grad(X) for any constant c this is equivalent to scaling the loss).

## Adding data parallel

This is where things get slightly messier. Let's extend the example to two devices: using Lij to refer to the unnormalized cross-entropy loss for rank i in its jth batch, and similarly for nij with number of tokens.

So in the first batch the model will see: 
*On rank 1*: loss L11 based on n11 tokens
*On rank 2*: loss L21 based on  n21 tokens

In the second batch, it will see:
*On rank 1*: loss L12 based on n12 tokens
*On rank 2*: loss L22 based on  n22 tokens

Similarly to the single-device case, the total number of tokens seen across all batches and all ranks will be n11 + n21 + n12 + n22. This means that the properly-normalized loss should be given by (L11 + L21 + L12 + L22) / (n11 + n21 + n12 + n22).

### What are we doing today?

Currently we take a similar approach to the single-device case described previously: we accumulate the losses and tokens, then call .backward() on the ratio running_loss / running_num_tokens just before stepping with the optimizer. See the below code: https://github.com/pytorch/torchtune/blob/a1bcb971cf2f506c10e07433303e1eb90613efbe/recipes/full_finetune_distributed.py#L724-L732

### What's wrong with this? 

During data parallel training, the loss on a given rank is based only on the subset of data seen by that rank. Similarly, our calculation of running_num_tokens is based only on the tokens from that rank. This means that when we normalize we are normalizing only over iterations, not over ranks. Put another way, the line `loss = running_loss / num_tokens` in the above snippet will yield (L11 + L12) / (n11 + n12) on rank 1 and (L21 + L22) / (n21 + n22) on rank 2. Finally, we call .backward(), which calculates local grads before firing a hook to sync by reducing over all ranks. The upshot is that our loss winds up as [(L11 + L12) / (n11 + n12)] + [(L21 + L22) / (n21 + n22)], which is definitely not (L11 + L21 + L12 + L22) / (n11 + n21 + n12 + n22).

### How can we fix it?

Aside from the correctness issue described in the previous section, we also need an approach that still calls .backward() on each batch to free the activation memory. How can we do this? Actually it's not so bad.. we just need to reduce the number of tokens before our final gradient normalization and let data parallel backward hooks take care of the rest. More explicitly:

First batch:
*On rank 1*: calculate loss L11 based on n11 tokens
*On rank 2*: calculate loss L21 based on n21 tokens
**Call backward** -> this triggers a grad sync and each rank now has the (local) grads for L11 + L21

Second batch: 
*On rank 1*: calculate loss L12 based on n12 tokens
*On rank 2*: calculate loss L22 based on n22 tokens
**Call backward** -> this accumulates the grads from L11 + L21 locally, then triggers another sync so that we now have L11 + L21 + L12 + L22

Then we just need n11 + n21 + n12 + n22. Fortunately this is just an all-reduce on running_num_tokens. Then we manually scale the gradients just like in the single-device case.

**In summary, we wind up with the following process:**

For each batch:
- Call backward on the unnormalized losses
- Keep a running tally of the number of tokens seen (per rank)

Before optimizer step:
- [*data parallel only*] Reduce the running tally of number of tokens over all ranks 
- Scale gradients by 1 / total_num_tokens

### DIFF
--- a/recipes/full_finetune_distributed.py
+++ b/recipes/full_finetune_distributed.py
@@ -130,14 +130,19 @@ class FullFinetuneRecipeDistributed(FTRecipeInterface):
 
         # _is_rank_zero is used primarily for logging. In the future, the logger
         # should directly take care of this
-        self._world_size, rank = training.get_world_size_and_rank()
-        self._rank = rank
+        _, rank = training.get_world_size_and_rank()
         self._is_rank_zero = rank == 0
 
         # Training cfg
         self._resume_from_checkpoint = cfg.resume_from_checkpoint
         self._gradient_accumulation_steps = cfg.gradient_accumulation_steps
         self._optimizer_in_bwd = cfg.get("optimizer_in_bwd", False)
+
+        if self._gradient_accumulation_steps > 1 and self._optimizer_in_bwd:
+            raise RuntimeError(
+                "Gradient accumulation is not supported with optimizer in bwd."
+                "Please set gradient_accumulation_steps=1, or optimizer_in_bwd=False."
+            )
 
         # These are public properties which are updated by the checkpoint loader
         # when ``resume_from_checkpoint`` is `True` or validated in tests
@@ -632,7 +637,7 @@ class FullFinetuneRecipeDistributed(FTRecipeInterface):
         # clean up before training begins
         training.cleanup_before_training()
 
-        self._world_size, rank = training.get_world_size_and_rank()
+        world_size, rank = training.get_world_size_and_rank()
 
         # zero out the gradients before starting training
         if not self._optimizer_in_bwd:
@@ -705,17 +710,24 @@ class FullFinetuneRecipeDistributed(FTRecipeInterface):
 
                 running_loss += current_loss
 
-                if (idx + 1) % self._gradient_accumulation_steps != 0:
-                    with training.no_sync(self._model):
-                        current_loss.backward()
-                else:
-                    current_loss.backward()
+                # For optimizer in backward, we need to normalize before calling backward
+                # This case and gradient accumulation are mutually exclusive
+                if self._optimizer_in_bwd:
+                    torch.distributed.all_reduce(num_tokens)
+                    torch.distributed.all_reduce(running_loss)
+                    current_loss = current_loss / num_tokens
+
+                current_loss.backward()
 
                 # Step with optimizer
                 if (idx + 1) % self._gradient_accumulation_steps == 0:
-                    local_num_tokens = num_tokens.detach().clone()
-                    torch.distributed.all_reduce(num_tokens)
-                    training.scale_grads(self._model, self._world_size / num_tokens)
+                    if not self._optimizer_in_bwd:
+                        # Get total number of tokens across all ranks to normalize gradients
+                        torch.distributed.all_reduce(num_tokens)
+                        # This will ensure that the logged loss matches what we're optimizing
+                        torch.distributed.all_reduce(running_loss)
+                        # Manually scale the gradients from unnormalized loss by total # of tokens
+                        training.scale_grads(self._model, 1 / num_tokens)
                     if self._clip_grad_norm is not None:
                         if self._optimizer_in_bwd:
                             raise NotImplementedError(
@@ -753,8 +765,8 @@ class FullFinetuneRecipeDistributed(FTRecipeInterface):
                                     else self._optim_ckpt_wrapper
                                 ),
                             ),
-                            "tokens_per_second_per_gpu": local_num_tokens
-                            / time_per_step,
+                            "tokens_per_second_per_gpu": num_tokens
+                            / (time_per_step * world_size),
                         }
                         if self._log_peak_memory_stats:
                             log_dict.update(
--- a/recipes/full_finetune_distributed.py
+++ b/recipes/full_finetune_distributed.py
@@ -151,12 +151,20 @@ class FullFinetuneRecipeDistributed(FTRecipeInterface):
         self._resume_from_checkpoint = cfg.resume_from_checkpoint
         self._gradient_accumulation_steps = cfg.gradient_accumulation_steps
         self._optimizer_in_bwd = cfg.get("optimizer_in_bwd", False)
+        self._clip_grad_norm = cfg.get("clip_grad_norm", None)
 
-        if self._gradient_accumulation_steps > 1 and self._optimizer_in_bwd:
-            raise RuntimeError(
-                "Gradient accumulation is not supported with optimizer in bwd."
-                "Please set gradient_accumulation_steps=1, or optimizer_in_bwd=False."
-            )
+        # Optimizer in backward is not compatible with gradient accumulation or gradient clipping
+        if self._optimizer_in_bwd:
+            if self._clip_grad_norm is not None:
+                raise RuntimeError(
+                    "Gradient clipping is not supported with optimizer in bwd."
+                    "Please set clip_grad_norm=None, or optimizer_in_bwd=False."
+                )
+            if self._gradient_accumulation_steps > 1:
+                raise RuntimeError(
+                    "Gradient accumulation is not supported with optimizer in bwd."
+                    "Please set gradient_accumulation_steps=1, or optimizer_in_bwd=False."
+                )
 
         # activation checkpointing/offloading
         self._enable_activation_checkpointing = cfg.get(
@@ -187,7 +195,6 @@ class FullFinetuneRecipeDistributed(FTRecipeInterface):
         self.total_epochs = cfg.epochs
         self.max_steps_per_epoch = cfg.max_steps_per_epoch
         self.global_step = 0
-        self._clip_grad_norm = cfg.get("clip_grad_norm", None)
 
     def load_checkpoint(self, cfg_checkpointer: DictConfig) -> Dict[str, Any]:
         """
@@ -796,16 +803,11 @@ class FullFinetuneRecipeDistributed(FTRecipeInterface):
                         torch.distributed.all_reduce(running_loss)
                         # Manually scale the gradients from unnormalized loss by total # of tokens
                         training.scale_grads(self._model, 1 / num_tokens)
-                    if self._clip_grad_norm is not None:
-                        if self._optimizer_in_bwd:
-                            raise NotImplementedError(
-                                "Gradient clipping is not supported after optimizer-in-the-backward."
+                        if self._clip_grad_norm is not None:
+                            grad_norm = torch.nn.utils.clip_grad_norm_(
+                                self._model.parameters(),
+                                max_norm=float(self._clip_grad_norm),
                             )
-                        grad_norm = torch.nn.utils.clip_grad_norm_(
-                            self._model.parameters(),
-                            max_norm=float(self._clip_grad_norm),
-                        )
-                    if not self._optimizer_in_bwd:
                         self._optimizer.step()
                         self._optimizer.zero_grad(set_to_none=True)
 
--- a/recipes/full_finetune_distributed.py
+++ b/recipes/full_finetune_distributed.py
@@ -697,15 +697,17 @@ class FullFinetuneRecipeDistributed(FTRecipeInterface):
                 # Compute loss
                 # Loss is normalized by default so we multiply by the number of tokens
                 # This way we can normalize by the total number of tokens if we're accumulating gradients
-                running_loss += self._loss_fn(logits, labels) * current_num_tokens
+                current_loss = self._loss_fn(logits, labels) * current_num_tokens
 
                 # free logits otherwise it peaks backward memory
                 del logits
 
+                running_loss += current_loss
+                current_loss.backward()
+
                 # Step with optimizer
                 if (idx + 1) % self._gradient_accumulation_steps == 0:
-                    loss = running_loss / num_tokens
-                    loss.backward()
+                    training.scale_grads(self._model, 1 / num_tokens)
                     if self._clip_grad_norm is not None:
                         if self._optimizer_in_bwd:
                             raise NotImplementedError(
@@ -722,7 +724,7 @@ class FullFinetuneRecipeDistributed(FTRecipeInterface):
                     # Update the number of steps when the weights are updated
                     self.global_step += 1
 
-                    loss_to_log = loss.item()
+                    loss_to_log = running_loss.item() / num_tokens
                     pbar.update(1)
                     pbar.set_description(
                         f"{curr_epoch + 1}|{self.global_step}|Loss: {loss_to_log}"
--- a/recipes/full_finetune_distributed.py
+++ b/recipes/full_finetune_distributed.py
@@ -45,12 +45,24 @@ class FullFinetuneRecipeDistributed(FTRecipeInterface):
             ``fsdp_reshard_after_forward`` to False (this corresponds to SHARD_GRAD_OP sharding strategy).
             DDP is currently not supported. Training on CPU is not supported.
 
-        - Activation Checkpointing. This can be controlled using the ``activation_checkpointing``
+        - Activation Checkpointing. This can be controlled using the ``enable_activation_checkpointing``
             flag. Activation checkpointing helps reduce the memory footprint since we no longer keep
             activations in memory and instead recompute them during the backward pass. This is especially
             helpful for larger batch sizes when you're memory constrained. But these savings in memory
             come at the cost of training performance. In most cases training can slow-down quite a bit as
             a result of this activation recomputation.
+
+        - Activation Offloading. This can be controlled using the ``enable_activation_offloading``
+            flag. Activation offloading is a technique similar to activations checkpointing that helps
+            reduce the memory footprint to prevent OOMs on CUDA and enable bigger batches. Where activations
+            checkpointing drops the activation in the forward to recompute it later in the backward,
+            activations offloading will drop the activation in the forward to the CPU and bring it
+            back during the backward pass. As always, there is a tradeoff--these savings in memory can
+            come at the cost of training performance and CPU resources. To recover some runtime cost,
+            we've added an option to enable offloading on a different stream to permit overlapping with
+            the computation. This option is currently only available on PyTorch 2.5 or later and will
+            be enabled by default if an acceptable torch version is found. Activation offloading can be
+            used in conjunction with activation checkpointing.
 
         - Precision. Full fp32 and bf16 training are supported. Precision is controlled using the ``dtype``
             flag. When ``dtype=bf16``, all activations, gradients and optimizer states are in bfloat16. In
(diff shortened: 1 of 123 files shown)


torchtune/training/_grad_scaler.py
```diff
@@ -0,0 +1,14 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
```
joecummings, line 1: Does this really need its own file?

joecummings: No

ebsmothers: Where do you wanna put it then? Otherwise I am gonna copy-paste this in every recipe which is worse imo

recipes/full_finetune_distributed.py
```diff
@@ -631,7 +632,7 @@ def train(self) -> None:
         # clean up before training begins
         training.cleanup_before_training()
 
-        _, rank = training.get_world_size_and_rank()
+        self._world_size, rank = training.get_world_size_and_rank()
```
ebsmothers, line 635: will remove

recipes/full_finetune_distributed.py
```diff
@@ -722,7 +732,7 @@ def train(self) -> None:
                     # Update the number of steps when the weights are updated
                     self.global_step += 1
 
-                    loss_to_log = loss.item()
+                    loss_to_log = running_loss.item() / num_tokens
```
ebsmothers, line 735: Should probably normalize by local_num_tokens?

recipes/full_finetune_distributed.py
```diff
@@ -697,15 +698,24 @@ def train(self) -> None:
                 # Compute loss
                 # Loss is normalized by default so we multiply by the number of tokens
                 # This way we can normalize by the total number of tokens if we're accumulating gradients
-                running_loss += self._loss_fn(logits, labels) * current_num_tokens
+                current_loss = self._loss_fn(logits, labels) * current_num_tokens
 
                 # free logits otherwise it peaks backward memory
                 del logits
 
+                running_loss += current_loss
```
felipemello1, line 706: i dont know if this doubles memory, because it creates a copy of the the current_loss. If it does (you can test with a large enough batch_size), then maybe we should just do it if self._gradient_accumulation_steps>1

ebsmothers: Can you clarify? The loss itself is a single float, right? Do you mean that we are now keeping additional activations around that we otherwise wouldn't?

recipes/full_finetune_distributed.py
```diff
@@ -697,15 +698,24 @@ def train(self) -> None:
...
+                else:
+                    current_loss.backward()
+
                 # Step with optimizer
                 if (idx + 1) % self._gradient_accumulation_steps == 0:
-                    loss = running_loss / num_tokens
-                    loss.backward()
+                    local_num_tokens = num_tokens.detach().clone()
+                    torch.distributed.all_reduce(num_tokens)
+                    training.scale_grads(self._model, self._world_size / num_tokens)
```
felipemello1, line 718: there are so many lines taking care of the all_reduce, backward, etc, that it makes me wonder if this should be a utility.

torchtune/training/_distributed.py
```diff
@@ -679,3 +691,12 @@ def shard_model(
 
     # Finally shard the entire model to account for any stragglers
     fully_shard(model, **fsdp_kwargs)
+
+
+@contextlib.contextmanager
+def no_sync(model: nn.Module) -> Generator[None, None, None]:
```
felipemello1, line 697: name could be more descriptive, maybe no_grad_sync

recipes/full_finetune_distributed.py
felipemello1: The loss is a single float, so if the only extra memory is that, then ignore my comment. But I dont know if maybe because of hooks or something else, it ends up not releasing memory or consuming more than just that float.

recipes/full_finetune_distributed.py
```diff
@@ -697,15 +698,24 @@ def train(self) -> None:
...
+                current_loss = self._loss_fn(logits, labels) * current_num_tokens
 
                 # free logits otherwise it peaks backward memory
                 del logits
 
+                running_loss += current_loss
+
+                if (idx + 1) % self._gradient_accumulation_steps != 0:
+                    with training.no_sync(self._model):
+                        current_loss.backward()
```
felipemello1, line 710: nice

recipes/full_finetune_distributed.py
ebsmothers: Yeah agreed, will eventually kick off full runs to check peak memory

recipes/full_finetune_distributed.py
ebsmothers: Yeah maybe. In this case I feel like it's important enough (and tricky enough) logic to be done very explicitly. Whatever route we go I will ultimately make it more explicit what's happening here

recipes/full_finetune_distributed.py
```diff
@@ -697,15 +703,31 @@ def train(self) -> None:
...
                 if (idx + 1) % self._gradient_accumulation_steps == 0:
-                    loss = running_loss / num_tokens
-                    loss.backward()
+                    if not self._optimizer_in_bwd:
+                        # Get total number of tokens across all ranks to normalize gradients
+                        torch.distributed.all_reduce(num_tokens)
+                        # This will ensure that the logged loss matches what we're optimizing
+                        torch.distributed.all_reduce(running_loss)
+                        # Manually scale the gradients from unnormalized loss by total # of tokens
+                        training.scale_grads(self._model, 1 / num_tokens)
```
felipemello1, line 730: 1) The way num_tokens is being calculated/reset seems a bit fishy. 
Lets say you have 2 gpus, and curr_num_tokens = [5,20] in gpus 0,1 for the first step, and [10, 3] for the second step.

For gpu 0, step 1:
num_tokens += 5
For gpu 1, step 1
num_tokens += 20

then you do all reduce, so both should have 25

You did NOT reset num_tokens, since we are accumulating gradients,

so on step 2, num_tokens is:
For gpu 0, step 2:
num_tokens =  25+10
For gpu 1, step 1
num_tokens = 25+3

you do all_reduce, and its
35+28.

Isnt this whats happening?

Then you do all reduce again when accumulating gradients. Does it make num_tokens * 2?

felipemello1: 2) Does all reduce require synchronization? If so, i wonder if we can make it async, and start it before running the model, so it has enough time. Maybe not for this PR..

ebsmothers: all_reduce on tokens only happens just before step (but lmk if i'm misunderstanding your comment)

ebsmothers:
cc @andrewor14 for review of the QAT recipe changes

recipes/full_finetune_distributed.py
ebsmothers: Update: I am probably gonna keep it like this since it should be representative of the loss we are actually using to step (even though it means our loss curves will look slightly different than they do today)

ebsmothers:
also cc @lindawangg for the KD recipe changes

recipes/full_finetune_distributed.py
felipemello1: i think it makes sense. Will it break all regression tests though?

recipes/full_finetune_distributed.py
```diff
@@ -833,7 +833,8 @@ def train(self) -> None:
                                     else self._optim_ckpt_wrapper
                                 ),
                             ),
-                            "tokens_per_second_per_gpu": num_tokens / time_per_step,
+                            "tokens_per_second_per_gpu": num_tokens
+                            / (time_per_step * world_size),
```
felipemello1, line 837: why times world size? This is per gpu, so we should not multiply * numbers of gpu

felipemello1:
lgtm

lindawangg (approved):
KD changes looks good to me

tests/recipes/test_qat_distributed.py
```diff
@@ -45,18 +45,18 @@ def _get_test_config_overrides(self):
...
         }
         return loss_values_map[model_type]
 
     @pytest.mark.integration_test
     @pytest.mark.parametrize(
         "config, model_type, ckpt_type, micro_batch_size, gradient_accumulation_steps",
         [
-            ("llama2/7B_qat_full", "llama2", "hf", 4, 1),
-            ("llama3/8B_qat_full", "llama3", "tune", 4, 1),
+            # ("llama2/7B_qat_full", "llama2", "hf", 4, 1),
```
joecummings, line 57: Commented out?

joecummings (approved):
Math!

felipemello1:
The only thing that is not crystal clear to me is why TPS would increase. I guess that before we would also count padded tokens, which should make TPS looks higher in the older version.

joecummings:
> The only thing that is not crystal clear to me is why TPS would increase. I guess that before we would also count padded tokens, which should make TPS looks higher in the older version.

Yeahhhhhhh, seems like the reduce would actually slow down training no?

torchtune/training/_grad_scaler.py
```diff
@@ -0,0 +1,26 @@
...
+    Inputs:
+        model (nn.Module): model whose gradients should be scaled
+        scaler (torch.Tensor): scaling factor to apply to the gradients
+
+    Outputs:
+        None (grad fields are modified in place)
+    """
+    for p in model.parameters():
+        if p.grad is not None:
+            p.grad *= scaler
```
pbontrager, line 26: Is there any concern here around overflows for lower dtypes? We could do a scaler range check based on dtype. Or  is it better to leave it to the recipe to safely choose scaler values?

recipes/full_finetune_distributed.py
```diff
@@ -787,15 +771,31 @@ def train(self) -> None:
...
+                current_loss = self._loss_fn(logits, labels) * current_num_tokens
 
                 # free logits otherwise it peaks backward memory
                 del logits
 
+                running_loss += current_loss
+
+                # For optimizer in backward, we need to normalize before calling backward
+                # This case and gradient accumulation are mutually exclusive
+                if self._optimizer_in_bwd:
```
pbontrager, line 783: I think 783 - 810 could be much more readable as:

```python
if self.optimizer_in_bwd:
	...
	current_loss.backward()
elif (idx + 1) % self._gradient_accumulation_steps == 0:
	current_loss.backward()
	...
	scale_grads()
	if self._clip_grad_norm is not None:
		...
	self._optimizer.step()
	self._optimizer.zero_grad(...)
```

This could be used in all the distributed recipes.

recipes/full_finetune_distributed.py
```diff
@@ -787,15 +771,31 @@ def train(self) -> None:
                 # Compute loss
                 # Loss is normalized by default so we multiply by the number of tokens
                 # This way we can normalize by the total number of tokens if we're accumulating gradients
-                running_loss += self._loss_fn(logits, labels) * current_num_tokens
+                current_loss = self._loss_fn(logits, labels) * current_num_tokens
```
pbontrager, line 774: If there was ever a issue with numerical stability, another option for scaling the loss would be:

```python
if grad_accumulation_step == 0:
	base_num_tokens = current_num_tokens

current_loss = loss_fn(logits, labels) * current_num_tokens / base_num_tokens
```

This might over complicate things but I wanted to leave this here if in the future it turns out a reduced gradient/loss is necessary for smaller dtypes.

pbontrager (commented):
Thanks for coming up with the best solution to this problem! I left some comments around some potention dtype concerns and readability, but otherwise this looks great!

ebsmothers:
@andrewor14 if it's all the same to you I may keep the QAT changes in. Mainly because this change will actually impact the memory usage and loss curves in a nontrivial way, so I'd rather not have any divergence between our recipes there, even for a short time. Lmk if that sounds alright to you though

recipes/full_finetune_distributed.py
ebsmothers: Agreed, we can take a look at this in the future

torchtune/training/_grad_scaler.py
ebsmothers: Discussed a bit offline, but this should be safe for bf16 and fp32. We can revisit upon integration of other dtypes

ebsmothers:
@felipemello1 @joecummings discussed offline but the WPS is not measured in exactly the same way before vs after these changes. In the figures we have "old-version", "main", and "this-pr". In "old-version" I hacked it so that we are counting non-padding tokens only (this way it lines up with the other two). But both "old-version" and "main" are using only the tokens seen on rank 0. On "this-pr" we now have the tokens seen by all ranks, which we then normalize by the number of ranks. So this PR's version should be less noisy, but there will be some slight differences when compared to the other two versions.

Regarding the question of all_reduce slowing down training: I can do a more detailed comparison, but we are only reducing a single float value. When compared against various FSDP comms on large tensors happening in forward and backward this should be pretty minimal (but will confirm).

Issue #1874, opened 2024-10-20
[BUG] Llama3.2 vision eleuther eval recipe `RuntimeError: stack expects each tensor to be equal size, but got [259, 6404]`


SalmanMohammadi:
Ran on 1 x RTX A6000 48GB.

```bash
root@f5de76cb602e:~/torchtune# tune run eleuther_eval --config llama3_2_vision/evaluation
Running EleutherEvalRecipe with resolved config:

batch_size: 1
checkpointer:
  _component_: torchtune.training.FullModelMetaCheckpointer
  checkpoint_dir: /tmp/Llama-3.2-11B-Vision-Instruct/original
  checkpoint_files:
  - consolidated.pth
  model_type: LLAMA3_VISION
  output_dir: ./
device: cuda
dtype: bf16
enable_kv_cache: true
limit: null
log_level: INFO
max_seq_length: 8192
model:
  _component_: torchtune.models.llama3_2_vision.llama3_2_vision_11b
quantizer: null
seed: 1234
tasks:
- mmmu_val_science
tokenizer:
  _component_: torchtune.models.llama3_2_vision.llama3_2_vision_transform
  max_seq_len: 8192
  path: /tmp/Llama-3.2-11B-Vision-Instruct/original/tokenizer.model

2024-10-20:22:33:11,546 INFO     [_logging.py:101] Running EleutherEvalRecipe with resolved config:

batch_size: 1
checkpointer:
  _component_: torchtune.training.FullModelMetaCheckpointer
  checkpoint_dir: /tmp/Llama-3.2-11B-Vision-Instruct/original
  checkpoint_files:
  - consolidated.pth
  model_type: LLAMA3_VISION
  output_dir: ./
device: cuda
dtype: bf16
enable_kv_cache: true
limit: null
log_level: INFO
max_seq_length: 8192
model:
  _component_: torchtune.models.llama3_2_vision.llama3_2_vision_11b
quantizer: null
seed: 1234
tasks:
- mmmu_val_science
tokenizer:
  _component_: torchtune.models.llama3_2_vision.llama3_2_vision_transform
  max_seq_len: 8192
  path: /tmp/Llama-3.2-11B-Vision-Instruct/original/tokenizer.model

Model is initialized with precision torch.bfloat16.
2024-10-20:22:33:14,882 INFO     [eleuther_eval.py:505] Model is initialized with precision torch.bfloat16.
Running evaluation on the following tasks: ['mmmu_val_science']
[...]

PR #1920, opened 2024-10-29
[FIX] MM Eval Mask Sizes

pbontrager:
#### Context
- [ ] add a new feature
- [x] fix a bug
- [ ] other (please add here)

Issue first reported in #1874 found that eval fails for llama3.2 vision 11B. This issue found was that the VisionCrossAttentionMask was padding the masks to 4 tiles during inference while at the same time the padded_collate_tiled_images_and_mask function was assuming that the masks weren't padded and inferring incorrect shape information.

The solution is to remove any inference time padding logic from the mask transform and pass pad_max_tiles=4 to the collate function during inference and eval to let the collate function handle all the padding.

During this investigation I found that padded_collate_tiled_images_and_mask was using "image_seq_len" variable from the last definition in a loop, meaning that if there were multiple images with different sizes the variable would be wrong at the end. Updated this as well.

#### Changelog
* Set pad_max_tiles in generation_v2 and eval
* removed pad to max tiles from the vision encoder mask transform
* replaced image_seq_len with tokens_per_tile * max_num_tiles in the collate function
* updated the collate tests to cover the extra tiling

#### Test plan

# Ran as usual
tune run dev/generate_v2 --config llama3_2_vision/generation_v2 

# Ran as usual
tune run full_finetune_single_device --config llama2_2_vision/11B_full_single_device

# Fixed
tune run eleuther_eval --config llama3_2_vision/evaluation

Screenshot 2024-10-29 at 4 11 24 PM


torchtune/data/_collate.py
```diff
@@ -426,7 +426,8 @@ def padded_collate_tiled_images_and_mask(
     if pad_max_images is not None:
         _, _, img_seq = concat_masks.shape
         concat_masks = F.pad(
-            concat_masks, (0, pad_max_images * image_seq_len - img_seq)
+            concat_masks,
```
RdoubleA, line 429: where does the pad to max images happen? this is just padding the masks to max num images?

joecummings (approved):

torchtune/data/_collate.py
pbontrager: You don’t actually need to add them or you’d waste compute on them. If you have a I cache for 7 images, then you want to mask out the additional images. It’s similar to how you mask extra token positions during inference but don’t add padding tokens.

PR #1929, opened 2024-10-30
torchdata integration - multi-dataset and streaming support

andrewkho:
Note! This requires torchdata nightly to be installed to work correctly.

Test multi-dataset training command:
` tune run --nnodes=1 --nproc_per_node=8 lora_finetune_distributed_td --config llama3_2_vision/11B_lora_td dataloader.pin_memory=True use_torchdata=True dataloader.parallel_method=thread max_steps_per_epoch=1000 compile=True dataloader.num_workers=4`

Test multi-dataset command with dataloader_only mode:
`tune run --nnodes=1 --nproc_per_node=8 lora_finetune_distributed_td --config llama3_2_vision/11B_lora_td dataloader.pin_memory=True use_torchdata=True dataloader.parallel_method=thread max_steps_per_epoch=1000 profile_mode=dataloader_only compile=True dataloader.num_workers=4`

## Benchmarking on 8xA100
for 200 steps, full model training, batch_size: 4, gradient_accumulation_steps: 1
TWFB: Sum of Time Waiting For Batch (max across all ranks) divided by sum of step times
Single Datasets are run with OCRVQA
Multi-Dataset was run with:
ocrvqa, docvqa, dvqa, tabmwp with equal weighting for all datasets.

| Trial          | TWFB % |  Sum DL times  | sum step times |
| ------------- | ------------- | ------------- | ------------- |
| 0 Single dataset (Baseline) | 0.06279 | 14.596 | 232.454 | 
| 1 Single dataset (TorchData threads) | 0.00726 | 1.64746 | 226.780 | 
| 2 Multi dataset (threads) | 0.00514 | 2.4945 | 485.388 | 
| 3 Multi dataset Streaming | 0.01712, | 8.5445 | 499.0716 | 

Multi-dataset runs much slower, guessing because one of the datasets (dvqa?) requires more padding than ocrvqa.

Launch commands:
0: tune run --nnodes=1 --nproc_per_node=8 lora_finetune_distributed_td --config llama3_2_vision/11B_lora_td use_torchdata=false max_steps_per_epoch=200 compile=True
1: (same as 2 but with 3 of the datasets removed in the config)
2: tune run --nnodes=1 --nproc_per_node=8 lora_finetune_distributed_td --config llama3_2_vision/11B_lora_td use_torchdata=true max_steps_per_epoch=200 compile=True 
3: tune run --nnodes=1 --nproc_per_node=8 lora_finetune_distributed_td --config llama3_2_vision/11B_lora_td use_torchdata=true max_steps_per_epoch=200 compile=True dataloader.streaming=true

Please have a look at the code set up, and how this composability can help with streaming datasets, and multi-dataset mixing. You can think of this as approaches to replace torch.utils.data.DataLoader, while introducing more flexible parallelism schemes, eg instead of just one multiprocess worker setting, you could do multi-threading, pipeline parallelism, etc. It also enables more powerful composability IMO.

I have done some single-device benchmarking on a machine with A100 40gb, both with and without the model, performance is on par or better than standard dataloader. 

TODO: fill in below
#### Context
- [ ] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

#### Changelog
*

#### Test plan

#### UX


recipes/lora_finetune_single_device.py
```diff
@@ -501,12 +522,123 @@ def _setup_lr_scheduler(
...
+                    ParallelMapper,
+                    num_workers=num_workers,
+                    method=parallel_method,
+                    in_order=True,
+                )
+            # Map style set up =======
+            node = IterableWrapper(sampler)
+            node = _Mapper(node, map_fn=ds._data.__getitem__)
+            # Cut here for Streaming/Iterable dataset instead =====
+            node = _Mapper(node, map_fn=ds._prepare_sample)
```
RdoubleA, line 623: is this where the transform would get parallelized?

recipes/lora_finetune_single_device.py
```diff
@@ -501,12 +522,123 @@ def _setup_lr_scheduler(
...
+            else:
+                _Mapper = partial(
+                    ParallelMapper,
+                    num_workers=num_workers,
+                    method=parallel_method,
+                    in_order=True,
+                )
+            # Map style set up =======
+            node = IterableWrapper(sampler)
+            node = _Mapper(node, map_fn=ds._data.__getitem__)
```
RdoubleA, line 621: So does this mean we can keep our own Dataset abstractions? What if we went with entirely Iterable datasets?

RdoubleA (commented):
Thanks for this prototype, this is helpful to see! The way I'd imagine we'd expose the torchdata dataloader would be from a builder function with a few knobs exposed with reasonable defaults:
```
def build_dataloader(ds, num_workers, pin_memory, prefetch, in_memory, parallel_method, ...):
	# Put together all the nodes here

# In config
dataloader:
  num_workers:
  ...
```
For a power user, what might they want to tune to optimize performance for their hardware and model setup?

It's also not clear to me how some media transforms/decoding might get optimized, is that just handled by the torchdata nodes automatically?

recipes/lora_finetune_single_device.py
andrewkho: Technically yes, but this was just a way to demonstrate it, we can also refactor this into separate function. For entirely Iterable dataset, you could wrap it in IterableWrapper and then call everything underneath here, ie

```
node = IterableWrapper(my_iterable_dataset)
node = _Mapper(node, map_fn=ds._prepare_sample)
...
```

recipes/lora_finetune_single_device.py
andrewkho: Yes, see
```
_Mapper = partial(
    ParallelMapper,
    num_workers=num_workers,
    method=parallel_method,
    in_order=True,
)
```

recipes/lora_finetune_single_device.py
```diff
@@ -667,7 +803,22 @@ def train(self) -> None:
...
+                        dl_t0 = time.perf_counter()
+                        if self.profile_mode == "model_only":
+                            if batch_orig is None:
+                                batch_orig = next(dl_iter)
+                            batch = batch_orig.copy()
+                        else:
+                            batch = next(dl_iter)
+                        dl_dt = time.perf_counter() - dl_t0
+                        idx += 1
+                    except StopIteration:
```
joecummings, line 819: Why not create a context manager that handles this?

andrewkho: No good reason, hacky code is hacky :) Just wanted to see some rough numbers

recipes/lora_finetune_single_device.py
```diff
@@ -501,12 +522,123 @@ def _setup_lr_scheduler(
...
+                _Mapper = partial(
+                    ParallelMapper,
+                    num_workers=num_workers,
+                    method=parallel_method,
+                    in_order=True,
+                )
+            # Map style set up =======
+            node = IterableWrapper(sampler)
+            node = _Mapper(node, map_fn=ds._data.__getitem__)
+            # Cut here for Streaming/Iterable dataset instead =====
```
ebsmothers, line 622: Have we tried this out on an HF dataset with streaming = True yet? (I assume it won't work out of the box yet?)

ebsmothers (commented):
Thanks for creating this PR! I like the IterableWrapper/Batcher/etc APIs, they look pretty clean (and agree with @RdoubleA's suggestion about exposing this in builders/configs). How will this work when training on multiple datasets? Will we just apply the same set of APIs to each sampler individually?

andrewkho:
Thanks for the comments y'all, I updated this with a streaming example. Test with:
```
tune run lora_finetune_single_device --config llama3_2_vision/11B_lora_single_device num_workers=2 pin_memory=False use_torchdata=True parallel_method=thread max_steps_per_epoch=50 dataset.streaming=True
```

andrewkho:
@RdoubleA 

> For a power user, what might they want to tune to optimize performance for their hardware and model setup?

Good question, at the minimum it'd be some global worker-setting, but one option that may be worth supporting is allowing power users to define their entire pipelines entirely in config, not sure if you think this is a bad idea. eg by default, use a builder with a few knobs, but also allow the entire dataloader definition to be composable. Somewhat similar to how you enable users to pass a list of datasets right now. Something similar could be done for mixing. We'd need to be thoughtful for what the useful small-atomic units woudl be, and figure out syntax sugar.

> It's also not clear to me how some media transforms/decoding might get optimized, is that just handled by the torchdata nodes automatically?

In terms of optimizing, one thing we're looking at is tf.data's autotune approach which will automatically update prefetch buffers and workers. This is not implemented, but hoping to do something along these lines which will help with the too-many-tunable-parameters problem.

recipes/lora_finetune_single_device.py
andrewkho: see update :D almost out-of-the-box

RdoubleA: why do we need to wrap with `IterableWrapper` if the underlying dataset is IterableDataset? Also in our case, the underlying dataset would be a HF dataset class

andrewkho: Wrapper will make it conform to BaseNode's API so we can hold a pointer to iterator, as well as unify the state management. Subclasses need to define `.iterator()` instead of `.__iter__()`.

torchtune/datasets/multimodal/_the_cauldron.py
```diff
@@ -235,3 +243,73 @@ def __call__(self, sample: Mapping[str, Any]) -> Mapping[str, Any]:
     if packed:
         raise ValueError("Multimodal datasets don't support packing yet.")
     return ds
+
+
+def the_cauldron_dataset_torchdata(
```
joecummings, line 248: How much of this could be pushed into the SFT class so that we can just reuse it for any new datasets?

torchtune/datasets/_sft.py
```diff
@@ -117,7 +128,17 @@ def __getitem__(self, index: int) -> Dict[str, Any]:
         sample = self._data[index]
         return self._prepare_sample(sample)
 
-    def _prepare_sample(self, sample: Mapping[str, Any]) -> Dict[str, Any]:
+
+class PrepareSample:
```
andrewkho, line 132: Refactoring this into it's own Callable class that can be used by both the current torch.utils.Dataset and the torchdata.nodes

torchtune/datasets/multimodal/_the_cauldron.py
andrewkho: A lot of it is probably re-usable, I think it should be it's own class or at least a builder, maybe like `_sft.py: class SFTDatasetNode` or something less terrible sounding

andrewkho: Actually 90% of this builder func could probably live in _sft.py as it doesn't have anything to do with the cauldron

recipes/lora_finetune_distributed_td.py
```diff
@@ -0,0 +1,1119 @@
...
+        """
+        world_size, rank = training.get_world_size_and_rank()
+
+        if isinstance(cfg_dataset, ListConfig):
+            datasets = [
+                config.instantiate(single_cfg_dataset, self._tokenizer)
+                for single_cfg_dataset in cfg_dataset
+            ]
+            ds = ConcatDataset(datasets=datasets)
+            packed = False
```
felipemello1, line 739: We can add a log here or raise an error in the init in case packed was set to true. i.e. we dont want to silently override the config

recipes/lora_finetune_distributed_td.py
```diff
@@ -0,0 +1,1119 @@
...
+                weights[k] = float(cfg_and_weight.get("weight", 1.0))
+                datasets[k] = Prefetcher(
+                    self._setup_one_dataset(
+                        cfg_dataset=cfg_and_weight.dataset,
+                        global_shuffle=shuffle,
+                        global_parallel_method=parallel_method,
+                        global_streaming=streaming,
+                        global_num_workers=num_workers,
+                    ),
+                    prefetch_factor=8,
```
felipemello1, line 673: would it make sense to add it to setup_data as a default or exposed in the config? Or this parameter is not commonly changed?

recipes/lora_finetune_distributed_td.py
```diff
@@ -0,0 +1,1119 @@
...
+            self._tokenizer,
+            streaming=streaming,
+            shuffle=shuffle,
+            parallel_method=parallel_method,
+            num_workers=num_workers,
+        )
+
+        return ds
+
+    def _setup_data_td(
```
felipemello1, line 639: This would be a bit of a refactoring, and probably not the main point of this pr, but it would be nice to have _setup_dataset and _setup_dataloader as two different methods. It should be easier to read and maintain

recipes/lora_finetune_distributed_td.py
```diff
@@ -0,0 +1,1119 @@
...
+                        global_parallel_method=parallel_method,
+                        global_streaming=streaming,
+                        global_num_workers=num_workers,
+                    ),
+                    prefetch_factor=8,
+                )
+
+        # Instantiate collate_fn
+        if "left_pad_sequence" in collate_fn:
+            raise RuntimeError("left_pad_sequence collator is only for inference.")
```
felipemello1, line 678: I understand that this was already in the setup_data, but we usually try to fail fast and catch errors like this in the init.

recipes/configs/llama3_2_vision/11B_lora_td.yaml
```diff
@@ -0,0 +1,131 @@
...
+    filename_format: model-{}-of-{}.safetensors
+    max_filename: "00005"
+  recipe_checkpoint: null
+  output_dir: /tmp/Llama-3.2-11B-Vision-Instruct/
+  model_type: LLAMA3_VISION
+resume_from_checkpoint: False
+save_adapter_weights_only: False # PeFT formatting not available yet. This will save it in torchtune format only.
+
+# TorchData Dataset setup
+dataloader:
```
felipemello1, line 53: Usually libraries have dataset and dataloader as two stand alone modules. Torch also assumes that. I think that we should follow the same pattern. .e.g

```
dataloader:
	shuffle: True
	  collate_fn: torchtune.data.padded_collate_tiled_images_and_mask
	  parallel_method: thread
	  num_workers: 0  # Per dataset
	  pin_memory: true
	  packed: False # Set to true for great speed ups

datasets:
      ocrvqa:
      weight: 1.
      dataset:
        _component_: torchtune.datasets.multimodal.the_cauldron_dataset_torchdata
        subset: ocrvqa
    dvqa:
      weight: 1.
      dataset:
        _component_: torchtune.datasets.multimodal.the_cauldron_dataset_torchdata
        subset: dvqa
```

recipes/lora_finetune_distributed_td.py
```diff
@@ -0,0 +1,1119 @@
...
+                num_workers=num_workers,
+                method=parallel_method,
+            )
+        if len(cfg_datasets) == 1:
+            node = next(iter(datasets.values()))  # TODO: multi dataset
+        else:
+            node = MultiDatasetWeightedSampler(
+                source_nodes=datasets,
+                weights=weights,
+            )
```
felipemello1, line 696: n00b question: I understand that this is torchdata api, but i wonder if 'len(cfg_datasets) == 1' check should be inside of MultiDatasetWeightedSampler. I.e. do we need this if condition?

recipes/lora_finetune_distributed_td.py
```diff
@@ -0,0 +1,1119 @@
...
+            ),
+        )
+        if pin_memory:
+            node = PinMemory(node)
+        if num_workers > 0:
+            node = Prefetcher(node, 2)
+
+        log.info("TorchData nodes are initialized")
+
+        return node
```
felipemello1, line 717: node feels a bit weird, since we do: self._dataloader = self._setup_dataloader(...)

Should we rename it to dataloader?

recipes/lora_finetune_distributed_td.py
```diff
@@ -0,0 +1,1119 @@
...
+                )
+
+        # Instantiate collate_fn
+        if "left_pad_sequence" in collate_fn:
+            raise RuntimeError("left_pad_sequence collator is only for inference.")
+        collate_fn = _get_component_from_path(collate_fn)
+
+        # TODO: add multi-dataset mixer
+        if num_workers == 0:
+            _Mapper = Mapper
```
felipemello1, line 683: I am a bit torn about if getting mapper, sampler, pin memory, etc should be an utility shared across all recipes, or if it should be exposed. No strong opinion, just thinking outloud

recipes/lora_finetune_distributed_td.py
```diff
@@ -0,0 +1,1119 @@
...
+                    ignore_idx=self._loss_fn.ignore_index,
+                )
+                if not packed
+                else padded_collate_packed
+            ),
+        )
+        if pin_memory:
+            node = PinMemory(node)
+        if num_workers > 0:
+            node = Prefetcher(node, 2)
```
felipemello1, line 713: is this PreFetcher=2 different than the previous prefect_factor=8?

recipes/lora_finetune_distributed_td.py
```diff
@@ -0,0 +1,1119 @@
...
+
+    def _setup_data_td(
+        self,
+        cfg_dataloader: DictConfig,
+        batch_size: int,
+    ) -> BaseNode:
+        """
+        All data related setup happens here. Currently this recipe only supports
+        Map-style Datasets which fit into memory and an option for random shuffling.
+        Samplers, iterable datasets, and streaming datasets are not supported.
```
felipemello1, line 647: i guess this needs to be updated

felipemello1 (commented):
hey Andrew, thank you so much for this PR! This is such a nice feature to have.

I did a first pass. I understand that its still a draft, but thought of making comments so it could save you some time for when the PR is closer to being ready.

recipes/configs/llama3_2_vision/11B_lora_td.yaml
andrewkho: Thanks for the comments! The rationale here is that settings like shuffle and workers might eventually be configured individually in the datasets, however I can change this if you think it's the right abstraction for torchtune. Happy to discuss more if you'd like

recipes/lora_finetune_distributed_td.py
andrewkho: Updated PR: this is now exposed

recipes/lora_finetune_distributed_td.py
andrewkho: I've split this into setting up individual datasets (_setup_one_dataset) and the global dataloader/mixer set up.

recipes/lora_finetune_distributed_td.py
andrewkho: Wait, why isn't this showing the new version? Let me make sure I've pushed

torchtune/datasets/_sft.py
```diff
@@ -109,6 +110,16 @@ def __init__(
         self._data = load_dataset(source, **load_dataset_kwargs)
         if filter_fn is not None:
             self._data = self._data.filter(filter_fn)
+        if load_dataset_kwargs.get("streaming", False):
+            self._data = split_dataset_by_node(
```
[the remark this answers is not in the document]
andrewkho: This needs to be removed actually

torchtune/datasets/_sft.py
```diff
@@ -109,6 +110,16 @@ def __init__(
...
+                self._data,
+                rank=int(os.environ.get("RANK", 0)),
+                world_size=int(os.environ.get("WORLD_SIZE", 1)),
```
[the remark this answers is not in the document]
andrewkho: Needs to be removed

recipes/configs/llama3_2_vision/11B_lora_td.yaml
```diff
@@ -0,0 +1,135 @@
...
+  shuffle: True
+  collate_fn: torchtune.data.padded_collate_tiled_images_and_mask
+  parallel_method: thread
+  num_workers: 4  # Per dataset
+  pin_memory: true
+  packed: False # Set to true for great speed ups
+  prefetch_factor: 2
+seed: null
+
+multi_datasets:
```
RdoubleA, line 63: personally this is too verbose to parse for me, and even in the recipe there are just too many nested dictionaries. Ideally, I would like to achieve this type of UI in the config for datasets:
```
datasets:
  - _component_: torchtune.datasets...
    weight: 1.0
    subset: ...
  - _component_: torchtune.datasets...
    weight: 1.0
    ...
```
or something similar so all I have to do is specify the dataset I want and the weight. As it is I have multi_datasets -> datasets -> dataset just to specify the dataset builder. Maybe this is very ideal, but other libraries such as Axolotl are able to do this.

I am aware there's a few challenges to having this:
1. MultiDatasetSampler requires passing in dictionaries for datasets and weights
2. weight is not a valid argument for instantiating dataset components

I'm wondering if there's a way we can do this all for the user in a builder. For example:
```
datasets: ListConfig
for cfg_dataset in datasets:
	weights[k] = cfg_dataset.pop("weight")
	dataset[k] = Prefetcher(config.instantiate(cfg_dataset), prefetch_factor)
dataloader = get_multi_datasets(datasets, weights, cfg_dataloader)
```
stop_criterion imo should be moved to the dataloader config

recipes/lora_finetune_distributed_td.py
```diff
@@ -0,0 +1,1150 @@
...
+        pin_memory = cfg_dataloader.get("pin_memory", True)
+        collate_fn = cfg_dataloader.collate_fn
+        prefetch_factor = cfg_dataloader.get("prefetch_factor", 2)
+
+        stop_criterion = cfg_multi_datasets.get(
+            "stop_criterion", StopCriteria.CYCLE_UNTIL_ALL_DATASETS_EXHAUSTED
+        )
+        weights, datasets = {}, {}
+        cfg_datasets = cfg_multi_datasets.datasets
+        for k, cfg_and_weight in cfg_datasets.items():
```
RdoubleA, line 670: all of this logic needs to be in the builder. we do not want to expose torchdata internals in each recipe. I am totally okay with creating a new file in `torchtune/data` that contains functions that set up torchdata dataloaders and nodes.

Another ideal that I'm curious if we can achieve, can we unify the UX for multi datasets and single datasets? i.e., if we had a `get_dataloader` method, you can pass in a single dataset or a multi dataset and the call is the same in the recipe regardless of what the user specifies in the config

torchtune/datasets/_sft.py
```diff
@@ -110,36 +111,61 @@ def __init__(
...
-            validate_messages(transformed_sample["messages"])
-
-        tokenized_dict = self._model_transform(transformed_sample)
 
-        if not ("tokens" in tokenized_dict and "mask" in tokenized_dict):
-            keys_str = ", ".join(tokenized_dict.keys())
-            error_message = (
-                "model_transform returned the following keys: "
-                f"{keys_str}. Must return 'tokens' and 'mask' as keys."
+class SFTTransform(Transform):
```
RdoubleA, line 127: this may be the cleanest way to separate the old way and the torchdata way, but what are your thoughts on keeping the class as SFTDataset, but remove all the HF load dataset logic out? we can keep source as a class attribute but not load it, then in the recipe or a utility used in the recipe: if we are using torchdata, use the load_hf_dataset you added here. else, use a barebones load utility with the original behavior.

When we eventually migrate to torchdata, we can just remove this switch logic and all the API names will remain the same but everything is swapped out under the hood

andrewkho: I think this is possible, but the tricky part is that right now, SFTDataset is a subclass of torch.utils.data.Dataset and the new version is a subclass of torchdata.nodes.BaseNode, and reconciling them may be difficult. We wouldn't be able to just make a switch because the torch.utils.DataLoader needs to be swapped out as well for the new `get_loader` builder wherever that is being called

recipes/lora_finetune_distributed_td.py
```diff
@@ -0,0 +1,1150 @@
...
+        """
+        All torchdata related setup happens here. Currently this recipe supports
+        both Map and Streaming datasets (from HuggingFace datasets), and mixing multiple
+        datasets (can be mix of Map and Streaming).
+        """
+        world_size, rank = training.get_world_size_and_rank()
+
+        # Get global settings
+        shuffle = cfg_dataloader.shuffle
+        parallel_method = cfg_dataloader.get("parallel_method", "thread")
```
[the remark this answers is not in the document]
andrewkho: The trouble is that there's no real distinction between datasets and dataloader with torchdata.nodes, they're all just Iterators that can be composed together. Currently it's exposed but we can introduce restrictions if you think it would be simpler

torchtune/datasets/_sft.py
andrewkho: cc @pbontrager for thoughts

torchtune/datasets/multimodal/_the_cauldron.py
```diff
@@ -235,3 +235,17 @@ def __call__(self, sample: Mapping[str, Any]) -> Mapping[str, Any]:
     if packed:
         raise ValueError("Multimodal datasets don't support packing yet.")
     return ds
+
+
+def the_cauldron_transform(
```
joecummings, line 240: I absolutely love this.

joecummings: Although, we may not need the "column_map" b/c if it's truly just for the cauldron, then we know what the columns will be, no? It would be much clearer IMO to have e.g. image_column, text_column, etc.

recipes/lora_finetune_distributed_td.py
```diff
@@ -0,0 +1,1144 @@
...
+        # zero out the gradients before starting training
+        self._optimizer.zero_grad()
+
+        # Initialize tokens count and running loss (for grad accumulation)
+        t0 = time.perf_counter()
+        running_loss = 0
+        num_tokens = 0
+
+        loader: Iterable
+        if self.profile_mode == "model_only":
```
ebsmothers, line 905: Very noob question but are all these profiling changes purely debug code? Or will we actually need to make changes to how we expose the profiler in order to be able to disentangle time spent on dataloading vs time spent in the model now?

torchtune/data/_utils.py
```diff
@@ -142,3 +150,131 @@ def format_content_with_images(
...
+    shuffle: bool = True,
+    seed: int = 0,
+    num_workers: int = 1,
+    parallel_method: Literal["process", "thread"] = "thread",
+    **load_dataset_kwargs: Dict[str, Any],
+) -> DatasetType:
+    from torchdata.nodes import IterableWrapper, ParallelMapper, SamplerWrapper
+
+    # Need to lazy import to avoid circular dependency
+    from torchtune.training._distributed import get_world_size_and_rank
```
ebsmothers, line 188: A minor point but we should avoid doing this

recipes/lora_finetune_distributed_td.py
```diff
@@ -0,0 +1,1144 @@
...
+                datasets=datasets,
+                weights=weights,
+                stop_criteria=stop_criteria,
+            )
+        else:
+            dataset = next(iter(datasets.values()))
+
+        loader = get_dataloader(
+            dataset=dataset,
+            model_transform=SFTTransform(model_transform=self._tokenizer),
```
ebsmothers, line 703: What's up with this line? I thought we already pass the transform in `load_hf_dataset`. And why is it just the tokenizer instead of a multimodal transform?

recipes/lora_finetune_distributed_td.py
```diff
@@ -0,0 +1,1119 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
```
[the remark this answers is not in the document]
andrewkho: Maybe rename to _multi_dataset

torchtune/data/_utils.py
andrewkho: Lazy importing? I think the workaround would be to up-level torchtune.training._distributed to eg torchtune.distributed. WDYT?

torchtune/utils/_import_guard.py
```diff
@@ -10,3 +10,12 @@
 _SUPPORTS_FLEX_ATTENTION = (
     torch.cuda.is_available() and torch.cuda.get_device_capability() >= (7, 5)
 )
+
```
joecummings, line 13: nit: Ed says we shouldn't do this anymore :/


https://fb.workplace.com/groups/pytorch.dev/permalink/1677770972801376/

recipes/lora_finetune_distributed_multi_dataset.py
```diff
@@ -0,0 +1,1014 @@
...
+        )
+        weights, datasets = {}, {}
+        for idx, cfg_dataset in enumerate(cfg_datasets):
+            dataset_name = cfg_dataset.pop("name", None)
+            if dataset_name is None:
+                dataset_name = cfg_dataset.get("subset", None)
+            key = f"{idx}" + (f"_{dataset_name}" if dataset_name else "")
+            assert key not in weights, f"Duplicate dataset name {key}"
+            weights[key] = float(cfg_dataset.pop("weight"))
+            datasets[key] = self._setup_one_dataset(
```
andrewkho, line 634: From call: Let's inline this

recipes/lora_finetune_distributed_multi_dataset.py
```diff
@@ -0,0 +1,1014 @@
...
+
+        # zero out the gradients before starting training
+        self._optimizer.zero_grad()
+
+        # Initialize tokens count and running loss (for grad accumulation)
+        t0 = time.perf_counter()
+        running_loss = 0
+        num_tokens = 0
+
+        self._profiler.start()
```
andrewkho, line 810: From call: Let's drop all the profiling entirely

recipes/lora_finetune_distributed_multi_dataset.py
```diff
@@ -0,0 +1,1014 @@
...
+
+    def cleanup(self) -> None:
+        if self._is_rank_zero:
+            self._metric_logger.close()
+        destroy_process_group()
+
+
+def iter_timer(
+    iterable: Iterable[T], barrier: bool
+) -> Iterator[Tuple[T, float, float, float]]:
```
andrewkho, line 970: let's drop this entirely

recipes/lora_finetune_distributed_multi_dataset.py
```diff
@@ -0,0 +1,967 @@
...
+        stop_criteria = cfg_dataloader.get(
+            "stop_criteria", StopCriteria.CYCLE_UNTIL_ALL_DATASETS_EXHAUSTED
+        )
+        weights, datasets = {}, {}
+        for idx, cfg_dataset in enumerate(cfg_datasets):
+            dataset_name = cfg_dataset.pop("name", None)
+            if dataset_name is None:
+                dataset_name = cfg_dataset.get("subset", None)
+            key = f"{idx}" + (f"_{dataset_name}" if dataset_name else "")
+            assert key not in weights, f"Duplicate dataset name {key}"
```
ebsmothers, line 605: Wait how would this actually happen? Doesn't the `f"{idx}"` bit in the previous line make this (nearly) impossible?

torchtune/utils/_import_guard.py
```diff
@@ -4,9 +4,17 @@
...
 import torch
 
 # We can only use flex attention / BlockMask if torch version >= 2.5.0 and GPU is Turing / SM75 and above
 _SUPPORTS_FLEX_ATTENTION = (
     torch.cuda.is_available() and torch.cuda.get_device_capability() >= (7, 5)
 )
+
+_TORCHDATA_MIN_VERSION = "0.10.0"
+if importlib.util.find_spec("torchdata.nodes") is not None:
```
ebsmothers, line 17: Just to make sure I understand, the existence of `torchdata.nodes` is our proxy for torchdata version >= 0.10.0?

torchtune/data/_utils.py
```diff
@@ -142,3 +150,171 @@ def format_content_with_images(
...
+        collate_fn (Callable[[Any], Any]): collate function to apply to the samples of the dataset
+        packed (bool): whether to pack the dataset. Default is False. Not supported yet.
+        drop_last (bool): whether to drop the last batch. Default is True.
+        num_workers (int): number of workers to use for loading the dataset. Default is 0 (no parallelism
+        parallel_method (Literal["process", "thread"]): method to use for parallelism. Default is "thread".
+        prefetch_factor (Optional[int]): number of batches to prefetch. Default is 4.
+        pin_memory (bool): whether to pin memory. Default is False.
+    """
+    if packed:
+        raise ValueError("Multimodal datasets don't support packing yet.")
```
ebsmothers, line 301: Ultimately this isn't the right place for this check, right? I assume we will need to do it somewhere else (or at least in another way) once we onboard text datasets

ebsmothers (commented):
Thanks for your patience on the review here! I left a handful of pretty minor comments, but this is looking great. We should also make sure to get CI passing, but after that this looks good to me.

recipes/configs/llama3_2_vision/11B_lora_multi_dataset.yaml
```diff
@@ -0,0 +1,125 @@
...
+  - source: HuggingFaceM4/the_cauldron
+    subset: tabmwp
+    split: train
+    transform:
+      _component_: torchtune.datasets.multimodal.the_cauldron_transform
+    weight: 1.0
+
+# Fine-tuning arguments
+epochs: 1
+max_steps_per_epoch: 50
```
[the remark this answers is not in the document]
andrewkho: Yes that's right, is this an issue for any users?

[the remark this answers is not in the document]
andrewkho: Actually, looking at the code, this is only required for progress bar calculation. Depending on stop_criteria (ie default is cycle-until-all-datasets-exhausted) it's non-trivial to know a-priori the total number of steps we'll get before StopIteration is thrown due to randomness. If we don't care about progress bar we can drop this requirement

torchtune/data/_utils.py
```diff
@@ -142,3 +150,171 @@ def format_content_with_images(
             final_content_list.append({"type": "image", "content": images.pop(0)})
 
     return final_content_list
+
+
+def chain(*funcs: Callable):
```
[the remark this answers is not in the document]
andrewkho: We should make this into a standard utility in torchdata, will create an issue for it

torchtune/utils/_import_guard.py
andrewkho: Yes exactly, torchdata.nodes is only present in >= 0.10.0

recipes/lora_finetune_distributed_multi_dataset.py
```diff
@@ -0,0 +1,967 @@
...
+
+import torch
+from omegaconf import DictConfig, ListConfig
+
+from torch import nn
+from torch.distributed import destroy_process_group, init_process_group
+
+from torch.optim import Optimizer
+
+from torchdata.nodes import Loader, StopCriteria, T
```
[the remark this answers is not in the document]
andrewkho: Good catch! I'll get the rest of unused in a linter pass

recipes/lora_finetune_distributed_multi_dataset.py
andrewkho: You're right, I think I added the idx a little later on so we can drop this

torchtune/data/_utils.py
andrewkho: I think you're right, I'm assuming that packer behaviour is going to need to be configurable, so we'll probably need to change this signature entirely anyways. I'll just drop it as an option and catch this in lora_finetune_distributed_multi_dataset for now, wdyt? @ebsmothers

recipes/lora_finetune_distributed_multi_dataset.py
```diff
@@ -0,0 +1,967 @@
...
+log = utils.get_logger("DEBUG")
+
+
+class LoRAFinetuneRecipeDistributed(FTRecipeInterface):
+    """
+    Distributed LoRA finetuning recipe for dense transformer-based LLMs such as Llama2. This recipe supports
+    distributed training and can be run on a single node (1 to 8 GPUs).
+
+    Features:
+        - TorchData. Map and Streaming HuggingFace datasets, and multi-dataset mixing.
```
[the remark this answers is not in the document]
andrewkho: https://github.com/pytorch/torchtune/issues/2156 To track issue

torchtune/data/_utils.py
ebsmothers: Fyi our hero @joecummings addressed this in #2155. You can now do `from torchtune.utils import get_world_size_and_rank` at the top of the file

recipes/configs/llama3_2_vision/11B_lora_multi_dataset.yaml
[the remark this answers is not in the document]
ebsmothers: Oh sorry, thought I responded. It's not an issue, but we need to make sure to not break anyone when we move this into our core recipes (since we currently default `max_steps_per_epoch=null` everywhere). Progress bar I don't care about too much, though we do also need it for learning rate scheduler, right? But yeah I agree with your point that in general we can no longer just infer it in general. Not a blocker for this PR anyways

recipes/lora_finetune_distributed_multi_dataset.py
```diff
@@ -0,0 +1,967 @@
...
+
+    def load_checkpoint(self, cfg_checkpointer: DictConfig) -> Dict[str, Any]:
+        """
+        Extract the checkpoint state from file and validate. This includes the
+        base model weights. If resume_from_checkpoint is True, this also includes
+        the adapter weights and recipe state
+        """
+        self._checkpointer = config.instantiate(
+            cfg_checkpointer,
+            resume_from_checkpoint=self._resume_from_checkpoint,
```
ebsmothers, line 204: I think you'll need this now that #2006 has landed
```suggestion
            should_load_recipe_state=self._resume_from_checkpoint,
```

recipes/lora_finetune_distributed_multi_dataset.py
```diff
@@ -0,0 +1,967 @@
...
+        with training.set_default_dtype(self._dtype), self._device:
+            lora_device = "cpu" if fsdp_cpu_offload else self._device
+            for m in model.modules():
+                if (
+                    isinstance(m, LoRALinear) or isinstance(m, DoRALinear)
+                ) and not lora_weights_state_dict:
+                    # lora may not be covered in state dict
+                    # if finetune for the 1st time
+                    m.lora_a.to_empty(device=lora_device)
+                    m.lora_b.to_empty(device=lora_device)
```
ebsmothers, line 493: From #2139. Sorry about all these merges, it's been a busy week on the repo

```suggestion
                    m.to_empty(device=lora_device)
```

recipes/lora_finetune_distributed_multi_dataset.py
```diff
@@ -0,0 +1,967 @@
...
+            self._is_rank_zero,
+            cpu_offload=fsdp_cpu_offload,
+        )
+        is_dora = False
+        for m in model.modules():
+            if hasattr(m, "initialize_dora_magnitude"):
+                is_dora = True
+                m.initialize_dora_magnitude()
+        if is_dora:
+            load_dora_magnitudes(model)
```
ebsmothers, line 512: One more
```suggestion
        for m in model.modules():
            if hasattr(m, "initialize_dora_magnitude"):
                m.initialize_dora_magnitude()
```

recipes/lora_finetune_distributed_multi_dataset.py
```diff
@@ -0,0 +1,967 @@
...
+    def __init__(self, cfg: DictConfig) -> None:
+        self._device = utils.get_device(device=cfg.device)
+        self._dtype = training.get_dtype(cfg.dtype, device=self._device)
+
+        if self._dtype == torch.float16:
+            raise ValueError(
+                "full fp16 training is not supported with this recipe. Please use bf16 or fp32 instead."
+            )
+
+        _, rank = training.get_world_size_and_rank()
```
ebsmothers, line 142: Can do this now
```suggestion
        _, rank = utils.get_world_size_and_rank()
```

torchtune/data/_utils.py
ebsmothers: Yeah that works. We can revisit when we're migrating text recipes

torchtune/data/_utils.py
```diff
@@ -142,3 +150,181 @@ def format_content_with_images(
...
+
+    Returns:
+        A ``torchdata.nodes`` iterator that can be passed directly to a Loader, or combined with other-datasets in a multi-dataset
+        sampler.
+    """
+    from torchdata.nodes import IterableWrapper, ParallelMapper, SamplerWrapper
+
+    # TODO: Remove lazy import when we can
+    # see: https://github.com/pytorch/torchtune/issues/2151
+    from torchtune.training._distributed import get_world_size_and_rank
```
ebsmothers, line 212: Realized the comment I responded to was outdated, so just in case you miss it: can now use `from torchtune.utils import get_world_size_and_rank` at the top of the file

ebsmothers (approved):
OK a few more small merges to do, after that I think we're good. Thanks so much for building out this integration! Having proper multi-dataset and streaming support (not to mention the potential for more performant multimodal dataloading) is gonna be huge for our users.

PR #1931, opened 2024-10-31
Add support for QAT + LoRA

andrewor14:
**Summary:**

This commit adds a recipe that combines QAT + LoRA, with the main goal of improving final quantized accuracy after training while reducing the memory required for fine-tuning. The new recipe `qat_lora_finetune_distributed` mirrors the existing `lora_finetune_distributed` recipe, which performs only LoRA, and is analogous to the existing `qat_distributed` recipe, which performs only QAT.

Helpful code review commands:
```
diff --color recipes/lora_finetune_distributed.py recipes/qat_lora_finetune_distributed.py
diff --color recipes/configs/llama3/8B_lora.yaml recipes/configs/llama3/8B_qat_lora.yaml
diff --color recipes/configs/llama3_1/8B_lora.yaml recipes/configs/llama3_1/8B_qat_lora.yaml
diff --color recipes/configs/llama3_2/1B_lora.yaml recipes/configs/llama3_2/1B_qat_lora.yaml
diff --color recipes/configs/llama3_2/3B_lora.yaml recipes/configs/llama3_2/3B_qat_lora.yaml
```

For more context on QAT, please visit https://github.com/pytorch/torchtune/pull/980 and https://pytorch.org/blog/quantization-aware-training/.

**Test Plan**

Unit tests:
```
pytest -m integration_test tests/recipes/test_qat_lora_finetune_distributed.py
```

Manual tests:
```
export CUDA_VISIBLE_DEVICES=4,5,6,7
export NCCL_SHM_DISABLE=0
LOG_DIR=/home/andrewor/local/logs/tune/qat_lora

tune run --nnodes 1 --nproc_per_node 4 qat_lora_finetune_distributed --config llama3/8B_qat_lora \
    batch_size=4 \
    quantizer.groupsize=32 \
    checkpointer.output_dir="$LOG_DIR" \
    metric_logger.output_dir="${LOG_DIR}/metrics"

tune run quantize --config quantization \
    model._component_=torchtune.models.llama3.llama3_8b \
    checkpointer._component_=torchtune.training.FullModelMetaCheckpointer \
    checkpointer.checkpoint_dir="$LOG_DIR" \
    checkpointer.output_dir="$LOG_DIR" \
    checkpointer.checkpoint_files=["meta_model_0.pt"] \
    checkpointer.model_type=LLAMA3 \
    quantizer._component_=torchtune.training.quantization.Int8DynActInt4WeightQuantizer \
    quantizer.groupsize=32

tune run eleuther_eval --config eleuther_evaluation \
    batch_size=1 \
    model._component_=torchtune.models.llama3.llama3_8b \
    checkpointer._component_=torchtune.training.FullModelTorchTuneCheckpointer \
    checkpointer.checkpoint_dir="$LOG_DIR" \
    checkpointer.output_dir="$LOG_DIR" \
    checkpointer.checkpoint_files=["meta_model_0.pt-8da4w"] \
    checkpointer.model_type=LLAMA3 \
    tokenizer._component_=torchtune.models.llama3.llama3_tokenizer \
    tasks=[wikitext] \
    quantizer._component_=torchtune.training.quantization.Int8DynActInt4WeightQuantizer \
    quantizer.groupsize=32
```

Results:
```
# Baseline (LoRA only, no QAT)

| Tasks  |Version|Filter|n-shot|    Metric     |   | Value |   |Stderr|
|--------|------:|------|------|---------------|---|------:|---|------|
|wikitext|      2|none  |None  |bits_per_byte  |↓  | 0.6284|±  |   N/A|
|        |       |none  |None  |byte_perplexity|↓  | 1.5458|±  |   N/A|
|        |       |none  |None  |word_perplexity|↓  |10.2694|±  |   N/A|

# LoRA + QAT (new recipe)

| Tasks  |Version|Filter|n-shot|    Metric     |   | Value |   |Stderr|
|--------|------:|------|------|---------------|---|------:|---|------|
|wikitext|      2|none  |None  |bits_per_byte  |↓  | 0.6245|±  |   N/A|
|        |       |none  |None  |byte_perplexity|↓  | 1.5416|±  |   N/A|
|        |       |none  |None  |word_perplexity|↓  |10.1208|±  |   N/A|
```


eval_it.sh
```diff
@@ -0,0 +1,24 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
```
andrewor14, line 1: will delete these files before land

recipes/lora_finetune_distributed.py
```diff
@@ -413,6 +415,36 @@ def _setup_profiler(
...
+        quantizer.precision = self._dtype
+        quantizer_mode = training.quantization.get_quantizer_mode(quantizer)
+        if "qat" not in quantizer_mode:
+            raise ValueError(
+                "Quantizer mode '%s' is not supported for finetuning" % quantizer_mode
+            )
+        # TODO: Expose fake quantize configs from torchao so we can get them
+        # directly from the quantizer. For now, we hardcode the configs for 8da4w.
+        # E.g. activation_config = quantizer.get_activation_fake_quantize_config()
+        # E.g. weight_config = quantizer.get_weight_fake_quantize_config()
```
andrewor14, line 435: Will be fixed in https://github.com/pytorch/ao/pull/1214

recipes/qat_lora_finetune_distributed.py
```diff
@@ -0,0 +1,972 @@
...
+        quantizer.precision = self._dtype
+        quantizer_mode = training.quantization.get_quantizer_mode(quantizer)
+        if "qat" not in quantizer_mode:
+            raise ValueError(
+                "Quantizer mode '%s' is not supported for finetuning" % quantizer_mode
+            )
+        # TODO: Expose fake quantize configs from torchao so we can get them
+        # directly from the quantizer. For now, we hardcode the configs for 8da4w.
+        # E.g. activation_config = quantizer.get_activation_fake_quantize_config()
+        # E.g. weight_config = quantizer.get_weight_fake_quantize_config()
```
andrewor14, line 443: Addressed in https://github.com/pytorch/ao/pull/1214

gau-nernst:
Hey @andrewor14, I was hacking around with LoRA/QLoRA + INT8 mixed-precision and came across this PR of yours. I realized what we are trying to achieve is quite similar.

Components of `LoRALinear`:
- base weight: `F.linear(x, self.weight)` -> we want to modify this op
- LoRA adapters: `self.lora_b(self.lora_a(x))`

Since the base weight is direct children of `LoRALinear`, and `F.linear()` is hard-coded, it's hard to extend functionality of LoRALinear without re-writing the whole thing. So I had the idea of making the base weight as its own `nn.Linear()` module, thus we can freely swap the base linear module to modify its op.
- For my specific use case, I can quantize base weight to NF4, AND swap base linear module to INT8 mixed-precision, thus composing both NF4 weight and INT8 matmul for compute.

I have a POC here https://github.com/pytorch/torchtune/compare/main...gau-nernst:qlora (you can focus on `torchtune/modules/peft/lora.py` file). With this, we can re-use the linear module-swap in torchao. And you don't need a separate `qat_lora_finetune_distributed.py`, since we can add quantizer to existing recipes (though I understand you might not want this. I didn't carefully check the differences of QAT recipe script from other training scripts).

andrewor14:
Hi @gau-nernst, yeah I agree we can make the base weight more flexible, then we won't need to create a new class every time we need to extend lora functionality. cc @ebsmothers to see your thoughts on extending `LoRALinear` this way: https://github.com/pytorch/torchtune/compare/main...gau-nernst:qlora. For QAT in particular though the current flow uses full module swap (no tensor subclass yet), so we'll need some other way to initialize the base module like manually setting `self.base`, so it may not be as elegant there. Also I think the existing `QATLoRALinear` doesn't add much boiler plate code, so it might be OK.

For the separate recipe, I discussed this with @ebsmothers recently and I think it's torchtune's recipe organization philosophy to keep them separate, so QAT functionality won't complicate the original lora recipe.

andrewor14:
@ebsmothers Any comments? Does this look good to you?

ebsmothers:
Hey @andrewor14 sorry for the delay and thanks for your patience here. We are doing planning this week so my available bandwidth for reviewing this has taken a hit. I promise to get to it by Friday at the latest

recipes/qat_lora_finetune_distributed.py
```diff
@@ -0,0 +1,968 @@
...
+        self._profiler.start()
+        # self.epochs_run should be non-zero when we're resuming from a checkpoint
+        for curr_epoch in range(self.epochs_run, self.total_epochs):
+
+            # Update the sampler to ensure data is correctly shuffled across epochs
+            # in case shuffle is True
+            self._sampler.set_epoch(curr_epoch)
+
+            pbar = tqdm(total=self._steps_per_epoch, disable=not (rank == 0))
+            for idx, batch in enumerate(self._dataloader):
```
ebsmothers, line 803: So no option to wait N steps before enabling fake quant in this recipe? Any particular reason for that? (To clarify I'm not saying we should add it, mainly just curious)

recipes/qat_lora_finetune_distributed.py
```diff
@@ -0,0 +1,968 @@
...
+            log.info(
+                "FSDP is enabled. Instantiating model and loading checkpoint on Rank 0 ..."
+            )
+            init_start = time.perf_counter()
+
+        with training.set_default_dtype(self._dtype), torch.device("meta"):
+            model = config.instantiate(cfg_model)
+            # Optionally apply quantization-aware-training during fine-tuning
+            if quantizer_cfg is not None:
+                self._convert_model_to_qat(model, quantizer_cfg)
```
ebsmothers, line 489: I feel like for this recipe this shouldn't be optional, right?

recipes/qat_lora_finetune_distributed.py
```diff
@@ -0,0 +1,968 @@
...
+                self.profiler_warmup_steps = profiler_cfg["warmup_steps"]
+                self.profiler_active_steps = profiler_cfg["active_steps"]
+
+        return profiler
+
+    def _convert_model_to_qat(self, model: nn.Module, quantizer_cfg: DictConfig):
+        """
+        Convert the model to support quantization-aware training during fine-tuning.
+        """
+        for name, child in model.named_children():
```
ebsmothers, line 438: Why named children? Won't this miss any case that DoRA is applied to lower-level modules that are not direct children of model?

torchtune/modules/peft/lora.py
```diff
@@ -142,3 +298,29 @@ def _lora_b_init_params(x: nn.Linear) -> None:
...
+    module: nn.Module,
+    # TODO: make the types Optional[FakeQuantizeConfig] once we
+    # support torchao 0.7+ by default
+    activation_qat_config: Optional["FakeQuantizeConfig"] = None,
+    weight_qat_config: Optional["FakeQuantizeConfig"] = None,
+) -> None:
+    """
+    Swap all `LoRALinear`s in the model with `QATLoRALinear`.
+    """
+    for name, child in module.named_children():
```
ebsmothers, line 313: Kinda a similar comment here: why do we do this recursively instead of just bottoms-up iteration over named_modules? I guess this way is more robust?

torchtune/modules/peft/lora.py
```diff
@@ -142,3 +298,29 @@ def _lora_b_init_params(x: nn.Linear) -> None:
     Initialize LoRA B weight to zeros.
     """
     nn.init.zeros_(x.weight)
+
+
+def _swap_lora_linear_with_qat(
```
ebsmothers, line 303: Personally I'd put this somewhere else, either in quantization.py or peft/_utils.py

torchtune/modules/peft/lora.py
```diff
@@ -130,6 +130,162 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
...
+
+        """
+        if self.activation_fake_quantizer is not None:
+            _x = self.activation_fake_quantizer(x)
+        else:
+            _x = x
+        if self.weight_fake_quantizer is not None:
+            w = self.weight_fake_quantizer(self.weight)
+        else:
+            w = self.weight
```
ebsmothers, line 245: One thing we can consider to cut down on these if/elses in forward is just set activation_fake_quantizer and/or weight_fake_quantizer to nn.Identity() instead of None in the \_\_init\_\_

torchtune/modules/peft/lora.py
```diff
@@ -130,6 +130,162 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
...
+        # support torchao 0.7+ by default
+        activation_qat_config: Optional["FakeQuantizeConfig"] = None,
+        weight_qat_config: Optional["FakeQuantizeConfig"] = None,
+    ) -> "QATLoRALinear":
+        """
+        Create a `QATLoRALinear` from an existing `LoRALinear`,
+        preserving the weights and adapters.
+        """
+        if lora_linear.bias is not None:
+            ValueError("Bias is not supported in QAT yet")
```
ebsmothers, line 267: nit: be a bit more explicit in this error message
```suggestion
            ValueError("Bias is not supported in QAT + LoRA yet")
```

torchtune/modules/peft/lora.py
```diff
@@ -130,6 +130,162 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
         return out + lora_out
 
 
+class QATLoRALinear(LoRALinear):
```
ebsmothers, line 133: Can we also add a unit test for the QATLoRALinear class?

recipes/qat_lora_finetune_distributed.py
ebsmothers: Also can you remind me: is there a hard technical blocker for applying with DoRA? Or are we just leaving it out because it's still a bit unproven in this context? (The latter is definitely fine with me, just wanna confirm)

torchtune/modules/peft/lora.py
```diff
@@ -130,6 +130,162 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
...
+        weight_qat_config: Optional["FakeQuantizeConfig"] = None,
+    ):
+        super().__init__(
+            in_dim,
+            out_dim,
+            rank,
+            alpha,
+            dropout=0.0,
+            use_bias=False,
+            quantize_base=False,
```
ebsmothers, line 198: Do we also need to assert on this in the class_method `from_lora_linear`?

torchtune/modules/peft/lora.py
```diff
@@ -130,6 +130,162 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
...
+        # support torchao 0.7+ by default
+        activation_qat_config: Optional["FakeQuantizeConfig"] = None,
+        weight_qat_config: Optional["FakeQuantizeConfig"] = None,
+    ):
+        super().__init__(
+            in_dim,
+            out_dim,
+            rank,
+            alpha,
+            dropout=0.0,
```
ebsmothers, line 196: Why do we hardcode dropout to 0.0 here?

ebsmothers (commented):
Thanks @andrewor14 for the PR, and for your patience in the review process! I left a number of comments but no major concerns from my side. 

Regarding the two discussion points raised by @gau-nernst previously: 

(1) I am open to discussing whether we should change how we expose `self.weight` in `LoRALinear` from just `nn.Parameter` to `nn.Linear`. I agree that the latter would be more module-swap friendly, but (for better or for worse) we also do not really design things to have module-swap-based methods as a first-class citizen. Also I think it is not for free, as it means that the key names of the base linear weight will now have an extra module name in between for `LoRALinear` (e.g. `q_proj.base_weight.weight` instead of `q_proj.weight`). On the surface this is quite trivial, but from a checkpointing perspective it's really nice that we currently have an exact match between nn.Linear and LoRALinear and hence can load weights from one directly into the other (with `strict=False` of course). So making this simple change would actually have a pretty big blast radius across all our various checkpointing logic. Lmk if these points make sense, happy to discuss further with you.

(2) I think Andrew's separation of QAT full and QAT LoRA into separate recipes makes sense and aligns with what we do for our other recipes. If you are (as I am) a bit concerned about the proliferation of many recipes with similar functionality, I have two comments: (1) we will be making a dedicated effort to start simplifying and consolidating the recipe files somewhat so that they look more like their earlier versions. And (2) I think we now have enough recipes that we can actually consider e.g. `recipes/qat`, `recipes/knowledge_distillation`, etc. subdirectories. Neither of these will happen tomorrow, but I would like both to happen eventually.

gau-nernst:
@ebsmothers Regarding

> key names of the base linear weight will now have an extra module name in between for LoRALinear

In my proof-of-concept above, I handle this by adding the following hooks

```python
        def load_state_dict_pre_hook(module, state_dict, prefix, local_metadata, strict, missing_keys, unexpected_keys, error_msgs):
            if isinstance(module, LoRALinear):
                state_dict[f"{prefix}base.weight"] = state_dict.pop(f"{prefix}weight")

        self.register_load_state_dict_pre_hook(load_state_dict_pre_hook)

        def state_dict_post_hook(module, state_dict, prefix, local_metadata):
            if isinstance(module, LoRALinear):
                state_dict[f"{prefix}weight"] = state_dict.pop(f"{prefix}base.weight")

        self.register_state_dict_post_hook(state_dict_post_hook)
```

From my testing it seems sufficient, though I might not cover all edge cases (FSDP2?)

We can discuss more in a separate issue/PR if you are open to it, so as not to hijack this PR about QAT + LoRA 😄. The main benefit is ease of injecting custom logic, such as QAT for this PR, INT8 matmul for #1552, or even FP8 matmul in the future. You probably know better than me what are the potential issues, but I think we can try to see if those can be handled nicely.

ebsmothers:
@gau-nernst personally I have a bit of an aversion to state dict hooks as @pbontrager can attest 😅. Mainly I find that they make code really hard to debug. Correct usage of modules having state dict hooks generally requires that a module has its state dict called exactly once and submodules are not accessed or modified in any other way. And if either of these constraints are not satisfied the user will get a very non-obvious error about some missing attribute and it won't be at all clear where to go to fix it.

But I agree with your point about consolidating the discussion elsewhere (sounds like this PR wouldn't benefit as much from modifying `LoRALinear`'s `self.weight` anyways). Maybe some lightweight RFC discussing pros and cons would be helpful (I can add my comments there as well), and we can tag other folks to get their thoughts too.

recipes/qat_lora_finetune_distributed.py
andrewor14: Thanks, it should be `named_module()`. Yeah a bit unproven and unclear if there's demand for it currently. Can always add it later

torchtune/modules/peft/lora.py
andrewor14: this one's easier with recursion because we can `setattr` directly on the submodule

torchtune/modules/peft/lora.py
andrewor14: no reason, will add it to the args

tests/recipes/test_qat_lora_finetune_distributed.py
```diff
@@ -0,0 +1,270 @@
...
+        loss_values_map = {
+            "llama2": [10.5136, 10.4856, 10.5292, 10.5345],
+            "llama3": [11.9325, 11.9325, 11.9325, 11.9369],
+        }
+        return loss_values_map[model_type]
+
+    @pytest.mark.integration_test
+    @gpu_test(gpu_count=2)
+    @pytest.mark.parametrize(
+        "micro_batch_size, gradient_accumulation_steps, reshard_after_forward",
```
ebsmothers, line 56: nit but I think reshard_after_forward is less important to test, might be better to run with and without compile here

tests/recipes/test_qat_lora_finetune_distributed.py
```diff
@@ -0,0 +1,270 @@
...
+        tmpdir,
+        monkeypatch,
+    ):
+        ckpt = "llama2_tune"
+        ckpt_path = Path(CKPT_MODEL_PATHS[ckpt])
+        ckpt_dir = ckpt_path.parent
+        log_file = gen_log_file_name(tmpdir)
+        cmd = f"""
+        tune run --nnodes 1 --nproc_per_node 2 qat_lora_finetune_distributed
+            --config llama2/7B_lora \
```
ebsmothers, line 74: Why not run with a QAT config? I find this a bit confusing. (Similar comment for the other tests)

tests/torchtune/modules/peft/test_lora.py
```diff
@@ -232,3 +235,12 @@ def test_quantized_state_dict(self, dtype):
         assert torch.allclose(
             lora_linear.weight.quantized_data, lora_linear_reload.weight.quantized_data
         )
+
+    @pytest.mark.skipif(not _torchao_0_7_supported, reason="needs torchao 0.7+")
+    def test_qat_lora_forward(self, inputs, lora_linear, out_dim) -> None:
```
ebsmothers, line 240: Thanks for adding this!

recipes/qat_lora_finetune_distributed.py
```diff
@@ -0,0 +1,973 @@
...
+    get_merged_lora_ckpt,
+    LoRALinear,
+    set_trainable_params,
+    validate_missing_and_unexpected_for_lora,
+)
+from torchtune.recipe_interfaces import FTRecipeInterface
+from torchtune.training import DummyProfiler, PROFILER_KEY
+from torchtune.training.quantization import (
+    _swap_lora_linear_with_qat,
+    _torchao_0_7_supported,
```
ebsmothers, line 40: In general we try not to have private APIs directly in our recipes like this. Personally I would just make `swap_lora_linear_with_qat` public and provide a bit more color in the docstring (though you can also add the caveat that it is subject to change). For `_torchao_0_7_supported` it probably doesn't merit a public API, but maybe can just repeat the try/except import at the beginning of the recipe. (Btw what I actually want us to do is provide extra metadata in the recipe registry around version requirements for each recipe, but unfortunately haven't gotten around to doing that yet.)

ebsmothers (commented):
OK a couple more small comments but after that I think this should be good to go. A couple other requests before landing: 

1) Can you make sure this works with all our usual features (e.g. activation checkpointing, activation offloading)? I already ran with compile myself so no need to worry about that one
2) You should also add it to the recipes table in our readme! That way people will know to try it out

README.md
```diff
@@ -68,6 +68,7 @@ torchtune provides the following finetuning recipes for training on one or more
 | QLoRA Finetuning | 1-8 | [lora_finetune_single_device](recipes/lora_finetune_single_device.py) <br> [lora_finetune_distributed](recipes/lora_finetune_distributed.py)| [Phi3 Mini single-device](recipes/configs/phi3/mini_qlora_single_device.yaml) <br> [Llama 3.1 405B distributed](recipes/configs/llama3_1/405B_qlora.yaml)
 | DoRA/QDoRA Finetuning | 1-8 | [lora_finetune_single_device](recipes/lora_finetune_single_device.py) <br> [lora_finetune_distributed](recipes/lora_finetune_distributed.py)| [Llama3 8B QDoRA single-device](recipes/configs/llama3/8B_qdora_single_device.yaml) <br> [Llama3 8B DoRA distributed](recipes/configs/llama3/8B_dora.yaml)
 | Quantization-Aware Training | 4-8 | [qat_distributed](recipes/qat_distributed.py)| [Llama3 8B QAT](recipes/configs/llama3/8B_qat_full.yaml)
+| Quantization-Aware Training and LoRA Finetuning | 4-8 | [qat_lora_finetune_distributed](recipes/qat_lora_finetune_distributed.py)| [Llama3 8B QAT](recipes/configs/llama3/8B_qat_lora.yaml)
```
ebsmothers, line 71: Does it really need 4+ devices? The command given in the configs is for 2 devices and seems to run fine

torchtune/training/quantization.py
```diff
@@ -185,3 +195,45 @@ def _get_enable_fake_quant(quantizer_mode: str) -> Callable:
...
+    activation_qat_config: Optional["FakeQuantizeConfig"] = None,
+    weight_qat_config: Optional["FakeQuantizeConfig"] = None,
+) -> None:
+    """
+    Swap all `LoRALinear` in the model with `QATLoRALinear`.
+
+    This is used for combining QAT + LoRA during finetuning. The resulting linear layers
+    will apply the following transformation instead:
+
+        x -> fake_quantize(W_frozen) @ fake_quantize(x) + ABx
```
ebsmothers, line 213:
```suggestion
        x -> fake_quantize(W_frozen) @ fake_quantize(x) + BAx
```

README.md
andrewor14: Yeah 2 works. I'll update the QAT row as well. That one can also use 2

PR #1946, opened 2024-11-04
[Bug] model_type argument as str for checkpoints classes

smujjiga:
- [ ] add a new feature
- [x] fix a bug
- [ ] other (please add here)

The correct usage of `model_type` argument is as a string as shown  [here](https://github.com/pytorch/torchtune/blob/main/tests/torchtune/training/checkpointing/test_checkpointer.py#L172)


ebsmothers:
Thanks @smujjiga for the fix! This looks good to me, once CI is green we can merge

ebsmothers (approved):
Thanks for the fix!

PR #1948, opened 2024-11-04
update memory optimization tutorial

felipemello1:
#### Context
- [ ] add a new feature
- [ ] fix a bug
- [ ] other (please add here)
image


docs/source/tutorials/memory_optimizations.rst
```diff
@@ -15,14 +15,14 @@ To make things easy, we've summarized these components in the following table:
...
+   ":ref:`glossary_act_ckpt`", "Use when you're memory constrained and want to use a larger model, batch size or context lengths. Be aware that it will slow down training speed."
+   ":ref:`glossary_act_off`", "Similar to activation checkpointing, this can be used when memory constrained, but may decrease training speed due to the overhead of moving tensors between GPU VRAM and CPU. We minimize it by using a different stream, so you may not experience any slow down. This **should** be used alongside activation checkpointing."
+   ":ref:`glossary_grad_accm`", "Helpful when memory-constrained to simulate larger batch sizes. However, it is not compatible with optimizer in backward. Use it when training adapters, e.g. LoRA, which don't benefit from optimizer in backward, or when you can already fit a batch with you memory, but not enough of them.
+   ":ref:`glossary_low_precision_opt`", "When you have a large model (optimizer state is 2x model size) and need to further reduce memory. Note that lower precision optimizers may reduce training stability/accuracy."
```
ebsmothers, line 21: I feel like we are making these so-called summaries longer rather than shorter, which I do not like. Like honestly I have half a mind to just give columns like "memory", "tokens per second", "when to use" (maybe "constraints" too?) with up/down arrows for the first two columns and < 10 words on "when to use" (e.g. long sequence length, increase effective batch size, etc). The individual items further down should be responsible for getting more into the weeds.

ebsmothers: Also the offloading blurb is too handwavy. (Again would benefit from being moved down to the more detailed section to give it the proper treatment, as a first-time reader just trying to understand my options I definitely do not care that it's using a separate stream)

docs/source/tutorials/memory_optimizations.rst
```diff
@@ -15,14 +15,14 @@ To make things easy, we've summarized these components in the following table:
...
-   ":ref:`glossary_qlora`", "When you need even more memory savings than LoRA, at the potential cost of some training speed. Useful for very large models or limited hardware."
+   ":ref:`glossary_cpu_offload`", "Offloads optimizer states and (optionally) gradients to CPU, and performs optimizer steps on CPU. This can be used to significantly reduce GPU memory usage at the cost of CPU RAM and training speed, as CPU optimizer steps can be slow and bottleneck training performance. Prioritize using it only if the other techniques are not enough."
+   ":ref:`glossary_lora`", "When you want to significantly reduce the number of trainable parameters, saving gradient and optimizer memory during training, and significantly speeding up training. This may reduce training accuracy"
+   ":ref:`glossary_qlora`", "When you are training a large model and quantizing it will save significant memory, at the potential cost of some training speed and accuracy."
    ":ref:`glossary_dora`", "Like LoRA, DoRA can provide significant memory savings and training speed-ups. DoRA may improve performance over LoRA, particularly when using small rank updates."
```
ebsmothers, line 26: I know this was already here but I think we may wanna take it out. Like DoRA will increas memory and reduce training speed compared to LoRA, I think it does not fit in as well with the other techniques mentioned here.

docs/source/tutorials/memory_optimizations.rst
```diff
@@ -213,10 +221,9 @@ To understand how this works, we encourage you to read through the relevant PyTo
 
 .. todo ref full finetune recipe doc
 
-In torchtune, you can enable this feature using the ``optimizer_in_bwd`` flag, which is currently only supported in our
-single-device full finetune recipe. This feature works best when optimizer memory is particularly large;
-e.g. when using a stateful optimizer with a model with a lot of parameters, and when you don't need to use
-:ref:`gradient accumulation <glossary_grad_accm>`.
+In torchtune, you can enable this feature using the ``optimizer_in_bwd`` flag. This feature works best when using a stateful optimizer
+with a model with a lot of parameters, and when you don't need to use :ref:`gradient accumulation <glossary_grad_accm>`.
+You won't see meaningful impact when finetuning LoRA recipe's, since in this case the number of parameters being updated are small.
```
ebsmothers, line 226: Really we don't even support it for LoRA, right?
```suggestion
You won't see meaningful impact when finetuning LoRA recipes, since in this case the number of parameters being updated are small.
```

docs/source/tutorials/memory_optimizations.rst
```diff
@@ -168,25 +168,33 @@ Lower Precision Optimizers
 *What's going on here?*
 
 In addition to :ref:`reducing model and optimizer precision <glossary_precision>` during training, we can further reduce precision in our optimizer states.
-All of our single-device fine-tuning recipes support lower-precision optimizers from the `bitsandbytes <https://huggingface.co/docs/bitsandbytes/main/en/index>`_ library -
-a good place to start might be the ``AdamW8bit`` and ``PagedAdamW8bit`` optimizers, which we've tested our recipes with.
+All of our single-device fine-tuning recipes support lower-precision optimizers from the `torchao <https://github.com/pytorch/ao/tree/main/torchao/prototype/low_bit_optim>`_
+or `bitsandbytes <https://huggingface.co/docs/bitsandbytes/main/en/index>`_ libraries - a good place to start might be the ``AdamW8bit`` and ``PagedAdamW8bit`` optimizers,
+which we've tested our recipes with. Note that the implementation from bitsandbytes, at the moment this article was written, is not compatible with
```
ebsmothers, line 173:
```suggestion
which we've tested our recipes with. Note that bitsandbytes low-precision optimizers are not currently compatible with
```

docs/source/tutorials/memory_optimizations.rst
```diff
@@ -272,10 +282,10 @@ or using it directly in your code, which allows you to change the base optimizer
...
+* The CPU optimizer step is often the bottleneck when optimizer CPU offload is used. To minimize the slowdown, it is recommended to (1) use full ``bf16`` training so that parameters, gradients, and optimizer states are in ``bf16``; and (2) give GPU more work per optimizer step to amortize the offloading time (e.g. larger batch size with activation checkpointing, gradient accumulation).
 * Gradient accumulation should always be set to 1 when ``offload_gradients=True``, as gradients are cleared on GPU every backward pass.
 * This optimizer works by keeping a copy of parameters and pre-allocating gradient memory on CPU. Therefore, expect your RAM usage to increase by 4x model size.
-* This optimizer is only supported for single-device recipes. To use CPU-offloading in distributed recipes, use ``fsdp_cpu_offload=True`` in any distributed recipe. See :class:`torch.distributed.fsdp.FullyShardedDataParallel` for more details
+* This optimizer is only supported for single-device recipes. To use CPU-offloading in distributed recipes, use ``fsdp_cpu_offload=True`` instead. See :class:`torch.distributed.fsdp.FullyShardedDataParallel` for more details
```
ebsmothers, line 288: Is :class:`torch.distributed.fsdp.FullyShardedDataParallel` still the best API ref here?

docs/source/tutorials/memory_optimizations.rst
```diff
@@ -390,16 +400,16 @@ Quantized Low Rank Adaptation (QLoRA)
...
 This is enabled through a novel  4-bit NormalFloat (NF4) data type proposed by the authors, which allows for 4-8x less
 parameter memory usage whilst retaining model accuracy. You can read our tutorial on :ref:`finetuning Llama2 with QLoRA<qlora_finetune_label>`
 for a deeper understanding of how it works.
 
-When considering using QLoRA to reduce memory usage, it's worth noting that QLoRA prevents accuracy degradation during quantization
-by up-casting quantized parameters to the original higher precision datatype during model forward passes - this up-casting may
-incur penalties to training speed. The :ref:`relevant section <qlora_compile_label>` in our QLoRA tutorial demonstrates the usage of ``torch.compile``
-to address this by speeding up training.
+When considering using QLoRA to reduce memory usage, it's worth noting that a) QLoRA is slower than LoRA and may not be worth it if
+the model you are finetuning is small; b) QLoRA prevents accuracy degradation during quantization by up-casting quantized parameters
```
ebsmothers, line 410: I think this inline enumeration will not look great in the live docs. Also you can probably be more explicit about the memory savings, right? Doesn't it save roughly a constant 1.5 bytes * (# of model parameters) compared to LoRA? In general would try to be less handwavy about memory/perf statements in this doc wherever possible

ebsmothers (commented):
Looks like doc build job is failing

docs/source/tutorials/memory_optimizations.rst
SalmanMohammadi: Is this to say we should take DoRA from this whole doc?

docs/source/tutorials/memory_optimizations.rst
SalmanMohammadi: > Like honestly I have half a mind to just give columns like "memory", "tokens per second", "when to use" (maybe "constraints" too?) with up/down arrows for the first two columns and < 10 words on "when to use" (e.g. long sequence length, increase effective batch size, etc).

I'd be up for this, happy to do it in a follow up.

docs/source/tutorials/memory_optimizations.rst
```diff
@@ -104,14 +104,14 @@ This setting is especially helpful for larger batch sizes, or longer context len
...
-checkpointing, where all activations will either be recomputed later in the backward or brought back from the CPU.
+offloaded, we do not recommend using it unless Activation Checkpointing is also enabled, in which case only the checkpointed
+tensors will be offloaded.
 
 *Sounds great! How do I use it?*
 
 To enable activation offloading, use the ``enable_activation_offloading`` config entry or flag
-in our lora finetuning single device recipe, e.g. ``enable_activation_offloading=True``. To allow
-usage of streams, make sure you are on a torch version later than PyTorch 2.5.0.
+in our lora finetuning single device recipe, e.g. ``enable_activation_offloading=True``. If you are on torch
```
SalmanMohammadi, line 113: Activation offloading has made it to FFT too, right? : )

docs/source/tutorials/memory_optimizations.rst
```diff
@@ -104,14 +104,14 @@ This setting is especially helpful for larger batch sizes, or longer context len
 While of course it takes runtime and resources to move Tensors from GPU to CPU and back, the implementation in
 torchtune uses multiple CUDA streams (when available) in order to overlap the extra communication with the computation
 to hide the extra runtime. As the communication workload is variable depending on the number and size of tensors being
-offloaded, it is common to not offload every single activation. In fact, one can use offloading in conjunction with activations
-checkpointing, where all activations will either be recomputed later in the backward or brought back from the CPU.
+offloaded, we do not recommend using it unless Activation Checkpointing is also enabled, in which case only the checkpointed
```
SalmanMohammadi, line 107:
```suggestion
offloaded, we do not recommend using it unless :ref:`glossary_act_ckpt` is also enabled, in which case only the checkpointed
```

docs/source/tutorials/memory_optimizations.rst
```diff
@@ -104,14 +104,14 @@ This setting is especially helpful for larger batch sizes, or longer context len
...
+tensors will be offloaded.
 
 *Sounds great! How do I use it?*
 
 To enable activation offloading, use the ``enable_activation_offloading`` config entry or flag
-in our lora finetuning single device recipe, e.g. ``enable_activation_offloading=True``. To allow
-usage of streams, make sure you are on a torch version later than PyTorch 2.5.0.
+in our lora finetuning single device recipe, e.g. ``enable_activation_offloading=True``. If you are on torch
+version later than PyTorch 2.5.0, it will allow the usage of a different stream automatically.
```
SalmanMohammadi, line 114:
```suggestion
version later than PyTorch 2.5.0, it will allow the usage of multiple CUDA streams automatically.
```

docs/source/tutorials/memory_optimizations.rst
```diff
@@ -143,10 +143,10 @@ If you're using one of our distributed recipes, simply multiply by the number of
 
   ``total_batch_size = batch_size * gradient_accumulation_steps * num_devices``
 
-Gradient accumulation is especially useful when you are memory constrained. In this case,
-accumulating gradients might give you better training speed than enabling :ref:`activation
-checkpointing <glossary_act_ckpt>`, since activation checkpointing reduces memory consumption at the cost of repeated
-computations.
+Gradient accumulation is especially useful when you can fit only a couple of batches in your GPU, even after using LoRA or
+optimizer in backward. In this case, accumulating gradients might give you better training speed than
```
SalmanMohammadi, line 147: This reads like you should be trying LoRA and opt in bwd before considering gradient accm

docs/source/tutorials/memory_optimizations.rst
```diff
@@ -143,10 +143,10 @@ If you're using one of our distributed recipes, simply multiply by the number of
...
+enabling :ref:`activation checkpointing <glossary_act_ckpt>`, since activation checkpointing reduces memory
+consumption at the cost of repeated computations.
```
SalmanMohammadi, line 149: I think we could maybe be more general here and just say something like: 

```suggestion
optimizer in backward. In this case, accumulating gradients might give you faster training speeds than using other memory optimization techniques like activation checkpointing.
```

docs/source/tutorials/memory_optimizations.rst
```diff
@@ -168,25 +168,33 @@ Lower Precision Optimizers
...
-a good place to start might be the ``AdamW8bit`` and ``PagedAdamW8bit`` optimizers, which we've tested our recipes with.
+All of our single-device fine-tuning recipes support lower-precision optimizers from the `torchao <https://github.com/pytorch/ao/tree/main/torchao/prototype/low_bit_optim>`_
+or `bitsandbytes <https://huggingface.co/docs/bitsandbytes/main/en/index>`_ libraries - a good place to start might be the ``AdamW8bit`` and ``PagedAdamW8bit`` optimizers,
+which we've tested our recipes with. Note that the implementation from bitsandbytes, at the moment this article was written, is not compatible with
+our distributed training recipes.
 
 *Sounds great! How do I use it?*
 
-To use this in your recipes, make sure you have installed bitsandbytes (``pip install bitsandbytes``). Then, enable
+To use this in your recipes, make sure you have installed torchao (``pip install torchao``) or bitsandbytes (``pip install bitsandbytes``). Then, enable
```
SalmanMohammadi, line 178: Why not just show that users can do both? It's a good opportunity to show plug and play operability with the fine-tuning ecosystem.

docs/source/tutorials/memory_optimizations.rst
```diff
@@ -232,6 +239,9 @@ through the `CPUOffloadOptimizer <https://github.com/pytorch/ao/tree/main/torcha
 This optimizer can wrap any base optimizer and works by keeping the optimizer states and performing the optimizer step on CPU, thus reducing
 GPU memory usage by the size of the optimizer states. Additionally, we can also offload gradients to the CPU by using `offload_gradients=True`.
 
+If finetuning on a single-device, another option is to use the PagedAdamW8bit from bitsandbytes, mentioned above, which will *only* offload to CPU
```
SalmanMohammadi, line 242: Can you link to `PagedAdamW8bit`, and also the "above" in "from bitsandbytes, mentioned above", please?

docs/source/tutorials/memory_optimizations.rst
ebsmothers: Yeah the more I think about it I think we should. Since it's more of a modeling improvement on top of an existing memory saving technique. But lmk if you strongly disagree

docs/source/tutorials/memory_optimizations.rst
felipemello1: I GET PAID PER NUMBER OF WORDS USED

docs/source/tutorials/memory_optimizations.rst
felipemello1: yes, i missed this one

docs/source/tutorials/memory_optimizations.rst
SalmanMohammadi: In general I like to see features that we offer documented *somewhere* rather than not documented, so whether we put this here or e.g. move the section to the LoRA recipe docs to let users know they can use it, or as another option under the LoRA section, I'm easy.

docs/source/tutorials/memory_optimizations.rst
felipemello1: @SalmanMohammadi , i am gathering data to write a blog and add a table. After i have a draft, i would love to have your input. I should have it prob in a week.

docs/source/tutorials/memory_optimizations.rst
felipemello1: yes

felipemello1: i will add this comparing fsdp1 and 2 https://github.com/pytorch/torchtitan/blob/main/docs/fsdp.md

docs/source/tutorials/memory_optimizations.rst
felipemello1: i left DoRA as "a variant of LoRA" as a middle ground

docs/source/tutorials/memory_optimizations.rst
felipemello1: i simplified and removed info

docs/source/tutorials/memory_optimizations.rst
felipemello1: removed the a/b and added 1.5 bytes

docs/source/tutorials/memory_optimizations.rst
felipemello1: removed

docs/source/tutorials/memory_optimizations.rst
felipemello1: done

docs/source/tutorials/memory_optimizations.rst
```diff
@@ -168,25 +168,33 @@ Lower Precision Optimizers
...
 
-To use this in your recipes, make sure you have installed bitsandbytes (``pip install bitsandbytes``). Then, enable
+To use this in your recipes, make sure you have installed torchao (``pip install torchao``) or bitsandbytes (``pip install bitsandbytes``). Then, enable
 a low precision optimizer using the :ref:`cli_label`:
 
+
+.. code-block:: bash
+
+  tune run <RECIPE> --config <CONFIG> \
+  optimizer=torchao.prototype.low_bit_optim.torchao.AdamW8bit
```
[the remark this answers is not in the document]
felipemello1: done

docs/source/tutorials/memory_optimizations.rst
```diff
@@ -14,16 +14,16 @@ To make things easy, we've summarized these components in the following table:
...
-   ":ref:`glossary_cpu_offload`", "Offloads optimizer states and (optionally) gradients to CPU, and performs optimizer steps on CPU. This can be used to significantly reduce GPU memory usage at the cost of CPU RAM and training speed, as CPU optimizer steps can be slow and bottleneck training performance."
-   ":ref:`glossary_lora`", "When you want to significantly reduce the number of trainable parameters, saving gradient and optimizer memory during training, and significantly speeding up training."
-   ":ref:`glossary_qlora`", "When you need even more memory savings than LoRA, at the potential cost of some training speed. Useful for very large models or limited hardware."
-   ":ref:`glossary_dora`", "Like LoRA, DoRA can provide significant memory savings and training speed-ups. DoRA may improve performance over LoRA, particularly when using small rank updates."
+   ":ref:`glossary_precision`", "You'll usually want to leave this as its default ``bfloat16``. It uses 2 bytes per model parameter, half of fp32."
```
SalmanMohammadi, line 17: nit
```suggestion
   ":ref:`glossary_precision`", "You'll usually want to leave this as its default ``bfloat16``. It uses 2 bytes per model parameter instead of 4 bytes when using ``float32``."
```

docs/source/tutorials/memory_optimizations.rst
```diff
@@ -14,16 +14,16 @@ To make things easy, we've summarized these components in the following table:
...
+   ":ref:`glossary_act_ckpt`", "Use when you're memory constrained and want to use a larger model, batch size or context lengths. Be aware that it will slow down training speed."
```
SalmanMohammadi, line 18:
```suggestion
   ":ref:`glossary_act_ckpt`", "Use when you're memory constrained and want to use a larger model, batch size or context length. Be aware that it will slow down training speed."
```

docs/source/tutorials/memory_optimizations.rst
```diff
@@ -14,16 +14,16 @@ To make things easy, we've summarized these components in the following table:
...
+   ":ref:`glossary_act_off`", "Similar to activation checkpointing, this can be used when memory constrained, but may decrease training speed. This **should** be used alongside activation checkpointing."
```
SalmanMohammadi, line 19: Is this right? You should always activation offload when using activation checkpointing?

docs/source/tutorials/memory_optimizations.rst
```diff
@@ -14,16 +14,16 @@ To make things easy, we've summarized these components in the following table:
...
+   ":ref:`glossary_grad_accm`", "Helpful when memory-constrained to simulate larger batch sizes. Not compatible with optimizer in backward. Use it when you can already fit at least one sample without OOMing, but not enough of them."
+   ":ref:`glossary_low_precision_opt`", "Use it when you have a large model, since optimizer state is 2x model size, and need to further reduce memory. Note that lower precision optimizers may reduce training stability/accuracy."
```
SalmanMohammadi, line 21:
```suggestion
   ":ref:`glossary_low_precision_opt`", "Use when there is a large number of model parameters which need to be optimized to further reduce memory usage. Note that lower precision optimizers may reduce training stability/accuracy."
```

docs/source/tutorials/memory_optimizations.rst
```diff
@@ -14,16 +14,16 @@ To make things easy, we've summarized these components in the following table:
...
+   ":ref:`glossary_opt_in_bwd`", "Use it when you have large gradients and can fit a large enough batch size, since this is not compatible with ``gradient_accumulation_steps``."
```
SalmanMohammadi, line 22: I personally liked the original version of this but up 2  u

docs/source/tutorials/memory_optimizations.rst
```diff
@@ -14,16 +14,16 @@ To make things easy, we've summarized these components in the following table:
...
+   ":ref:`glossary_cpu_offload`", "Offloads optimizer states and (optionally) gradients to CPU, performing optimizer steps on CPU. This can be used to significantly reduce GPU memory usage at the cost of CPU RAM and training speed. Prioritize using it only if the other techniques are not enough."
```
SalmanMohammadi, line 23:
```suggestion
   ":ref:`glossary_cpu_offload`", "Offloads optimizer states and (optionally) gradients to CPU, and performs optimizer steps on CPU. This can be used to significantly reduce GPU memory usage at the cost of CPU RAM and training speed. Prioritize using it only if the other techniques are not enough."
```
Is " Prioritize using it only if the other techniques are not enough." right - should it be a last resort?

docs/source/tutorials/memory_optimizations.rst
```diff
@@ -14,16 +14,16 @@ To make things easy, we've summarized these components in the following table:
...
+   ":ref:`glossary_lora`", "When you want to significantly reduce the number of trainable parameters, saving gradient and optimizer memory during training, and significantly speeding up training. This may reduce training accuracy"
+   ":ref:`glossary_qlora`", "When you are training a large model, since quantization will save 1.5 bytes * (# of model parameters), at the potential cost of some training speed and accuracy."
+   ":ref:`glossary_dora`", "A variant of LoRA."
```
SalmanMohammadi, line 26: This line should just be removed right?
```suggestion
```

docs/source/tutorials/memory_optimizations.rst
```diff
@@ -272,10 +280,10 @@ or using it directly in your code, which allows you to change the base optimizer
...
 * Gradient accumulation should always be set to 1 when ``offload_gradients=True``, as gradients are cleared on GPU every backward pass.
 * This optimizer works by keeping a copy of parameters and pre-allocating gradient memory on CPU. Therefore, expect your RAM usage to increase by 4x model size.
-* This optimizer is only supported for single-device recipes. To use CPU-offloading in distributed recipes, use ``fsdp_cpu_offload=True`` in any distributed recipe. See :class:`torch.distributed.fsdp.FullyShardedDataParallel` for more details
+* This optimizer is only supported for single-device recipes. To use CPU-offloading in distributed recipes, use ``fsdp_cpu_offload=True`` instead. See :class:`torch.distributed.fsdp.FullyShardedDataParallel` for more details and `FSDP1 vs FSDP2 <https://github.com/pytorch/torchtitan/blob/main/docs/fsdp.md>_` to see how they differ.
```
SalmanMohammadi, line 286:
```suggestion
* This optimizer is only supported for single-device recipes. To use CPU-offloading in distributed recipes, use ``fsdp_cpu_offload=True`` instead. See :class:`torch.distributed.fsdp.FullyShardedDataParallel` for more details and `FSDP1 vs FSDP2 <https://github.com/pytorch/torchtitan/blob/main/docs/fsdp.md>`_ to see how they differ.
```

docs/source/tutorials/memory_optimizations.rst
felipemello1: yes, its super slow. :/

docs/source/tutorials/memory_optimizations.rst
felipemello1: the other way around, you always need ckpt when using offloading. Yes, otherwise its painfully slow.

docs/source/tutorials/memory_optimizations.rst
felipemello1: i think its good to keep the numbers there, so its less handwavy, like Evan suggested. I guess the issue with the statement is that its only true for adam :/

docs/source/tutorials/memory_optimizations.rst
felipemello1: I tried to keep it simple, given Evan's feedback that we added too much information here

SalmanMohammadi (approved):
thanks for your patience here : )

docs/source/tutorials/memory_optimizations.rst
```diff
@@ -14,16 +14,16 @@ To make things easy, we've summarized these components in the following table:
...
+   ":ref:`glossary_precision`", "You'll usually want to leave this as its default ``bfloat16``. It uses 2 bytes per model parameter instead of 4 bytes when using ``float32``."
+   ":ref:`glossary_act_ckpt`", "Use when you're memory constrained and want to use a larger model, batch size or context length. Be aware that it will slow down training speed."
+   ":ref:`glossary_act_off`", "Similar to activation checkpointing, this can be used when memory constrained, but may decrease training speed. This **should** be used alongside activation checkpointing."
+   ":ref:`glossary_grad_accm`", "Helpful when memory-constrained to simulate larger batch sizes. Not compatible with optimizer in backward. Use it when you can already fit at least one sample without OOMing, but not enough of them."
+   ":ref:`glossary_low_precision_opt`", "Use when you want to reduce the size of the optimizer state. This is relevant when training large models and using optimizers like Adam, that require 2x # model parameters. Note that lower precision optimizers may reduce training stability/accuracy."
```
ebsmothers, line 21: nit
```suggestion
   ":ref:`glossary_low_precision_opt`", "Use when you want to reduce the size of the optimizer state. This is relevant when training large models and using optimizers with momentum, like Adam. Note that lower precision optimizers may reduce training stability/accuracy."
```

docs/source/tutorials/memory_optimizations.rst
ebsmothers: Eh I like this even less. Now it is both (a) there and (b) uninformative. Sorry I am being a PITA. Let's say "a variant of LoRA that can improve model performance at the cost of slightly more memory"?

docs/source/tutorials/memory_optimizations.rst
```diff
@@ -168,25 +164,35 @@ Lower Precision Optimizers
 *What's going on here?*
 
 In addition to :ref:`reducing model and optimizer precision <glossary_precision>` during training, we can further reduce precision in our optimizer states.
-All of our single-device fine-tuning recipes support lower-precision optimizers from the `bitsandbytes <https://huggingface.co/docs/bitsandbytes/main/en/index>`_ library -
-a good place to start might be the ``AdamW8bit`` and ``PagedAdamW8bit`` optimizers, which we've tested our recipes with.
+All of our recipes support lower-precision optimizers from the `torchao <https://github.com/pytorch/ao/tree/main/torchao/prototype/low_bit_optim>`_ library.
+For single device recipes, we also support `bitsandbytes <https://huggingface.co/docs/bitsandbytes/main/en/index>`_ .
```
ebsmothers, line 168: super nit but rn the period renders on a new line in the docs
```suggestion
For single device recipes, we also support `bitsandbytes <https://huggingface.co/docs/bitsandbytes/main/en/index>`_.
```

docs/source/tutorials/memory_optimizations.rst
```diff
@@ -168,25 +164,35 @@ Lower Precision Optimizers
...
+
+A good place to start might be the :class:`torchao.prototype.low_bit_optim.torchao.AdamW8bit` and :class:`bitsandbytes.optim.PagedAdamW8bit` optimizers.
```
ebsmothers, line 170: Not a huge deal, but these `:class:` annotations don't actually render. If we wanna link out to something prob better to just use a hyperlink with Github URL pinned to a specific commit

docs/source/tutorials/memory_optimizations.rst
```diff
@@ -168,25 +164,35 @@ Lower Precision Optimizers
...
+Both reduce memory by quanting the optimizer state dict. The PagedAdam will also offload to CPU if there isn't enough GPU memory available. In practice,
```
ebsmothers, line 171:
```suggestion
Both reduce memory by quantizing the optimizer state dict. Paged optimizers will also offload to CPU if there isn't enough GPU memory available. In practice,
```

ebsmothers (approved):
Left a few more small comments, otherwise looks great. Thanks for making these changes!

docs/source/tutorials/memory_optimizations.rst
```diff
@@ -14,16 +14,16 @@ To make things easy, we've summarized these components in the following table:
...
+   ":ref:`glossary_opt_in_bwd`", "Use it when you have large gradients and can fit a large enough batch size, since this is not compatible with ``gradient_accumulation_steps``."
+   ":ref:`glossary_cpu_offload`", "Offloads optimizer states and (optionally) gradients to CPU, and performs optimizer steps on CPU. This can be used to significantly reduce GPU memory usage at the cost of CPU RAM and training speed. Prioritize using it only if the other techniques are not enough."
+   ":ref:`glossary_lora`", "When you want to significantly reduce the number of trainable parameters, saving gradient and optimizer memory during training, and significantly speeding up training. This may reduce training accuracy"
+   ":ref:`glossary_qlora`", "When you are training a large model, since quantization will save 1.5 bytes * (# of model parameters), at the potential cost of some training speed and accuracy."
+   ":ref:`glossary_dora`", "A variant of LoRA."
```
felipemello1, line 26:
```suggestion
   ":ref:`glossary_dora`", "a variant of LoRA that may improve model performance at the cost of slightly more memory."
```

PR #1949, opened 2024-11-04
Add Qwen2.5 to live docs

RdoubleA:
#### Context
- [ ] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

#### Changelog
*

#### Test plan

#### UX

### DIFF
--- a/docs/source/api_ref_models.rst
+++ b/docs/source/api_ref_models.rst
@@ -211,7 +211,8 @@ To download the CodeLlama-7B model:
 qwen-2
 ------
 
-Models of size 0.5B, 1.5B, and 7B from the `Qwen2 family <https://huggingface.co/collections/Qwen/qwen2-6659360b33528ced941e557f>`_.
+Models of size 0.5B, 1.5B, 3B, 7B, 14B, 32B, 72B from the `Qwen2.5 family <https://huggingface.co/collections/Qwen/qwen25-66e81a666513e518adb90d9e>`
+and 0.5B, 1.5B, and 7B from the `Qwen2 family <https://huggingface.co/collections/Qwen/qwen2-6659360b33528ced941e557f>`_.
 
 To download the Qwen2 1.5B model, for example:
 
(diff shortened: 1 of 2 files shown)


docs/source/api_ref_models.rst
```diff
@@ -211,7 +211,8 @@ To download the CodeLlama-7B model:
 qwen-2
 ------
 
-Models of size 0.5B, 1.5B, and 7B from the `Qwen2 family <https://huggingface.co/collections/Qwen/qwen2-6659360b33528ced941e557f>`_.
+Models of size 0.5B, 1.5B, 3B, 7B, 14B, 32B, 72B from the `Qwen2.5 family <https://huggingface.co/collections/Qwen/qwen25-66e81a666513e518adb90d9e>`
```
SalmanMohammadi:
```suggestion
Models of size 0.5B, 1.5B, 3B, 7B, 14B, 32B, 72B from the `Qwen2.5 family <https://huggingface.co/collections/Qwen/qwen25-66e81a666513e518adb90d9e>`_
```

RdoubleA: I just beat you by 1 second >:)

docs/source/api_ref_models.rst
```diff
@@ -225,14 +226,42 @@ To download the Qwen2 1.5B model, for example:
...
+    qwen2_5.lora_qwen2_5_14b_instruct
+    qwen2_5.qwen2_5_32b_base
+    qwen2_5.qwen2_5_32b_instruct
+    qwen2_5.lora_qwen2_5_32b_base
+    qwen2_5.lora_qwen2_5_32b_instruct
+    qwen2_5.qwen2_5_72b_base
+    qwen2_5.qwen2_5_72b_instruct
+    qwen2_5.lora_qwen2_5_72b_base
+    qwen2_5.lora_qwen2_5_72b_instruct
+    qwen2_5.qwen2_5_tokenizer
```
SalmanMohammadi, line 263: This is a bit hard on the eyes for me, I'd have to ctrl+f to navigate this.

SalmanMohammadi: Is there also any reason not to split into separate Qwen 2, and Qwen 2.5 sections?

RdoubleA: llama3 and 3.1 are in one section, so I was following that. but yeah I think the feature gap is significant enough for two separate sections

RdoubleA: > This is a bit hard on the eyes for me, I'd have to ctrl+f to navigate this

not sure how else I can format it without just removing some builders

ebsmothers: I think I missed when we put 3 and 3.1 together but not exactly sure why we did that tbh. Imo it is better to keep things separate, especially given the large number of Qwen2.5 models. As for readability, we should probably ideally have more nested subsections for different model sizes? I think we kinda shot ourselves in the foot by using section header for our title.. we should refactor to make it a proper title in RST parlance (stolen from stackoverflow):

```
==============
Document title
==============
-----------------
Document subtitle
-----------------

Section
=======

Sub-section
-----------
```

ebsmothers (approved):
Personally I would probably separate out Qwen2 and Qwen2.5 builders but given the Llama 3.1 precedent I don't have a strong opinion either way. Stamping to unblock

PR #1956, opened 2024-11-06
make RMSNorm module compatible with FSDP

anshulverma:
Summary: When parameters are initialized with meta device, FSDP calls `reset_parameters` function automatically if `param_init_fn` is not specified. For more details, see [wiki](https://pytorch.org/docs/stable/fsdp.html)

Differential Revision: D65496443

### DIFF
--- a/torchtune/modules/rms_norm.py
+++ b/torchtune/modules/rms_norm.py
@@ -25,8 +25,10 @@ class RMSNorm(nn.Module):
 
     def __init__(self, dim: int, eps: float = 1e-6) -> None:
         super().__init__()
-        self.eps = eps
-        self.scale = nn.Parameter(torch.ones(dim))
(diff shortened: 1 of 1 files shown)


felipemello1:
hey @anshulverma, we used to have fsdp1 and now we use fsdp2. I don't think that we have ever seen any warning complaining about reset_parameters. Can you help me understand the context for this PR? Did something break or you found some error when running it?

ebsmothers:
@anshulverma I am gonna close this PR. I may be missing the point (and if so feel free to reopen) but it seems to me that we should never have any problems here because we are always loading in RMSNorm scales as part of the pretrained checkpoint. So we will never run into any issues with garbage initialization via e.g. usage of `to_empty` without proper initialization

PR #1964, opened 2024-11-07
Refactor Recipe State Dict Code

pbontrager:
#### Context
- [ ] add a new feature
- [x] fix a bug
- [ ] other (please add here)

In our recipes we have 4 different ways that we collect model/adapter state_dicts:

1. single device: model.state_dict
2. lora single device: {k: v.cpu() for k,v in self.adapter_params}
3. distributed: get_full_model_state_dict using sharded_sd
4. distributed qlora: get_full_model_state_dict using model.named_modules

Where this causes issues is when models use state_dict_hooks which will change the expected names. This has come up as an issue with activation checkpointing in the past but was side stepped. Now with fusion models, this is an issue again as it relies on state dict hooks to operate. To address these known issues and make support easier in the future, this PR consolidates all of our checkpoint code to use the state_dict api so we'll always have a consistent set of names. This PR takes two primary approaches:

1. Introduces get_adapter_state_dict to filter a full model state_dict using the same pattern matching logic as get_merged_lora_ckpt uses. This incurs no extra cost even if save_adapter_only=True since calling model.state_dict doesn't copy params and is almost free.
2. replaces get_full_model_state_dict with gather_cpu_state_dict for distributed recipes. This takes a sharded state_dict as input and gathers each param and copies in to cpu. Notably, when there are NF4Tensors it still uses the state_dict (unlike the old function) but does a manual all_gather instead of calling full_tensor.

#### Changelog
* added get_merged_lora_ckpt
* replaces get_full_model_state_dict with gather_cpu_state_dict
* Updated every recipe that uses peft or get_full_model_state_dict
* Updated docs and tests

#### Test plan

I will update here with an overview of all the updated recipes showing that memory and save time doesn't change with the checkpoint update.

### DIFF
--- a/tests/torchtune/modules/peft/test_utils.py
+++ b/tests/torchtune/modules/peft/test_utils.py
@@ -16,6 +16,7 @@ from torchtune.modules.peft import (
     disable_adapter,
     DoRALinear,
     get_adapter_params,
+    get_adapter_state_dict,
     get_merged_lora_ckpt,
     LoRALinear,
     set_trainable_params,
@@ -38,30 +39,30 @@ ALPHA = 1
 class DummyAdapterModule(nn.Module, AdapterModule):
     def __init__(self, in_dim, out_dim):
         super().__init__()
-        self.adapter = nn.Linear(in_dim, out_dim, bias=False)
+        self.lora = nn.Linear(in_dim, out_dim, bias=False)
         self.linear = nn.Linear(in_dim, out_dim)
 
     def adapter_params(self):
-        return ["adapter.weight"]
+        return ["lora.weight"]
 
     def forward(self, x):
-        return self.adapter(x) + self.non_adapter(x)
+        return self.lora(x) + self.non_adapter(x)
 
 
 class DummyAdapterParentModel(nn.Module, AdapterModule):
     def __init__(self, in_dim, out_dim):
         super().__init__()
         self.dummy_adapter_module = DummyAdapterModule(in_dim, out_dim)
-        self.parent_adapter = nn.Linear(in_dim, out_dim)
+        self.parent_lora = nn.Linear(in_dim, out_dim)
         self.parent_base_model = nn.Linear(in_dim, out_dim)
 
     def adapter_params(self):
-        return ["parent_adapter.weight", "parent_adapter.bias"]
+        return ["parent_lora.weight", "parent_lora.bias"]
 
     def forward(self, x):
         return (
             self.dummy_adapter_module(x)
-            + self.parent_adapter(x)
+            + self.parent_lora(x)
             + self.parent_base_model(x)
         )
 
@@ -79,9 +80,9 @@ def dummy_model_expected_adapter_keys():
     for i in range(N_LAYERS):
         keys.extend(
             [
-                f"{i}.parent_adapter.weight",
-                f"{i}.parent_adapter.bias",
-                f"{i}.dummy_adapter_module.adapter.weight",
+                f"{i}.parent_lora.weight",
+                f"{i}.parent_lora.bias",
+                f"{i}.dummy_adapter_module.lora.weight",
             ]
         )
     return keys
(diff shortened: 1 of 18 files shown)


torchtune/training/_distributed.py
```diff
@@ -346,88 +343,61 @@ def load_from_full_model_state_dict(
...
-        for param_name, sharded_param in sharded_sd.items():
-            # without this, it may hang forever for +70B models.
-            torch.distributed.barrier()
-            if sharded_param.is_cpu:
-                assert device is not None and device.type == "cuda", (
-                    f"Expect cuda but got device={device}. "
-                    "Please call get_full_model_state_dict(..., device=self._device),"
-                    " so DTensor can communicate over NCCL."
+    for param_name, sharded_param in sharded_sd.items():
+        if sharded_param.is_cpu:
```
joecummings, line 367: What's happening here?

torchtune/training/_distributed.py
```diff
@@ -346,88 +343,61 @@ def load_from_full_model_state_dict(
...
+            sharded_param = sharded_param.to(device)
+        elif trainable_only and not sharded_param.requires_grad:
+            # skip non-trainable params when trainable_only is True
+            continue
+        if isinstance(sharded_param._local_tensor, NF4Tensor):
+            # NF4Tensor does not support all_gather from DTensor
```
joecummings, line 373: Lol why does it not support all_gather? Can't we ask AO to support that?

torchtune/training/_distributed.py
```diff
@@ -346,88 +343,61 @@ def load_from_full_model_state_dict(
...
             full_param = sharded_param.full_tensor()
-            if is_rank_zero:
-                cpu_state_dict[param_name] = full_param.cpu()
-            else:
-                del full_param
+        if is_rank_zero:
+            cpu_state_dict[param_name] = full_param.cpu()
+        else:
+            del full_param
+        torch.distributed.barrier()
```
joecummings, line 400: yay

joecummings (commented):
This looks really awesome overall. 

Main concerns are around any differences in memory and/or speed this could cause? Can we confirm?

tests/torchtune/modules/peft/test_utils.py
```diff
@@ -38,30 +39,30 @@
 class DummyAdapterModule(nn.Module, AdapterModule):
     def __init__(self, in_dim, out_dim):
         super().__init__()
-        self.adapter = nn.Linear(in_dim, out_dim, bias=False)
+        self.lora = nn.Linear(in_dim, out_dim, bias=False)
```
ebsmothers, line 42: So as of now what is the actual value of `AdapterModule`? Is it just for setting trainable params?

torchtune/modules/peft/_utils.py
```diff
@@ -107,6 +107,24 @@ def get_lora_module_names(
...
+    Args:
+        state_dict (Dict[str, Any]): Full model state dict.
+
+    Returns:
+        Dict[str, Any]: the subset of model's state dict containing
+        only adapter parameters.
+
+    """
+    adapter_key_filter = lambda x: "lora" in x or "magnitude" in x
+    return {k: v.cpu() for k, v in state_dict.items() if adapter_key_filter(k)}
```
ebsmothers, line 125: Do we want to make the move to CPU optional here?

torchtune/training/_distributed.py
ebsmothers: I assume this is for CPU offload or something?

torchtune/training/_distributed.py
```diff
@@ -346,88 +343,61 @@ def load_from_full_model_state_dict(
...
+        if isinstance(sharded_param._local_tensor, NF4Tensor):
+            # NF4Tensor does not support all_gather from DTensor
+            # so we need to manually all_gather
+            mesh = sharded_param.device_mesh
+            nf4_tensor = sharded_param._local_tensor
+            quant_params, metadata = nf4_tensor.fsdp_pre_all_gather(mesh)
+            full_quant_params = []
+            for quant_param in quant_params:
+                d0, *dn = quant_param.shape
+                shape = (d0 * mesh.get_group().size(), *dn)
```
ebsmothers, line 381: So this means sharding is always along the first dimension? If so might just leave a comment to that effect

pbontrager: Wei said FSDP always shards on dim 0

tests/torchtune/modules/peft/test_utils.py
pbontrager: I think AdapterModule is the right way to go, but since we are already ignoring it for ckpt merging, I'm not really changing anything by not using it for get_adapter_state_dict. I think we should move to using AdapterModule for all of these functions but that doesn't need to be solved in this PR.

torchtune/training/_distributed.py
pbontrager: We want AO to support it, but it would take too long to get on stable so we have to do it ourselves in the meantime.

ebsmothers (approved):
Thanks you for wading through this minefield and fixing this! And thanks for adding the new test case as well

PR #1969, opened 2024-11-07
CLIP Text Encoder

calvinpelletier:
#### Context
- [x] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

#### Changelog
- Implement the CLIP tokenizer
- Implement the CLIP text encoder

#### Test plan
Minimal code to run the CLIP text encoder e2e:
(first download CLIP weights:  `tune download openai/clip-vit-large-patch14 --output-dir /tmp/clip-vit-large-patch14 --ignore-patterns None`)
```python
import torch

from torchtune.models.clip._model_builders import clip_tokenizer, clip_text_encoder_large
from torchtune.training.checkpointing._checkpointer import FullModelHFCheckpointer

# build tokenizer and text encoder
tokenizer = clip_tokenizer()
encoder = clip_text_encoder_large()

# load weights
encoder.load_state_dict(FullModelHFCheckpointer(
    "/tmp/clip-vit-large-patch14",
    ["model.safetensors"],
    "CLIP_TEXT",
    "/tmp/torchtune-clip-vit-large-patch14",
).load_checkpoint()["model"])
encoder = encoder.to(torch.bfloat16).cuda().eval()

# run
text = [
    "a cow jumping over the moon",
    "a helpful AI assistant",
]
tokens = tokenizer(text)
encoding = encoder(tokens.cuda())
```

Checked parity with the HF CLIP tokenizer and text encoder as implemented [here](https://github.com/black-forest-labs/flux/blob/main/src/flux/modules/conditioner.py): MSE between the encoder outputs for on a batch of 32 test strings = `3.55e-5`

Tokenization speed for 32 test strings
- OpenAI: 0.0342
- HuggingFace: 0.0416
- TorchTune: 0.0195

Tokenization speed for an entire 100k img gen prompt dataset:
- HuggingFace: 18.02
- TorchTune: 6.89

Encoding speed for a single batch of 32 test strings:
- HuggingFace: 1.699
- TorchTune: 0.0173

Encoding speed for 1000 batches of 32 test strings:
- HuggingFace: 30.498
- TorchTune: 11.713

Checklist
- [x]  add [unit tests](https://github.com/pytorch/torchtune/tree/main/tests/torchtune) for any new functionality
  - TODO: unit test for encoder

#### UX


tests/torchtune/models/clip/test_clip_tokenizer.py
```diff
@@ -0,0 +1,44 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
```
calvinpelletier, line 1: How should I unit test the encoder given that the test won't have access to the CLIP weights? Should I just build an encoder with random weights and check that a simple forward and backward pass doesn't throw any exceptions?

tests/torchtune/models/clip/test_clip_tokenizer.py
```diff
@@ -0,0 +1,44 @@
...
+# LICENSE file in the root directory of this source tree.
+import pytest
+
+from torchtune.models.clip._model_builders import clip_tokenizer
+
+
+class TestCLIPTokenizer:
+    @pytest.fixture
+    def tokenizer(self):
+        return clip_tokenizer()
```
calvinpelletier, line 14: this will download the CLIP vocab (2mb) to `~/.cache/torchtune/clip_vocab.txt.gz` (if it doesn't already exist). is this ok to have in a unit test? should i instead add the vocab to `tests/assets/`?

torchtune/models/clip/_model_builders.py
```diff
@@ -1,4 +1,63 @@
-from torchtune.models.clip._transforms import CLIPImageTransform
```
calvinpelletier, line 1: What belongs in the `_model_builders` and what belongs in the `_component_builders`? It seems like we generally put model and tokenizer builders in `_model_builders` for most models like qwen, etc. But for CLIP vision, it's flipped (model builders are in `_component_builders` and the transform is in `_model_builders`). In fact, this file was broken because this import was misnamed, so I assume this file wasn't even really being used.

torchtune/models/clip/_text_encoder.py
```diff
@@ -0,0 +1,109 @@
...
+                    sa_norm=nn.LayerNorm(embed_dim, eps=norm_eps),
+                    mlp_norm=nn.LayerNorm(embed_dim, eps=norm_eps),
+                )
+                for _ in range(num_layers)
+            ]
+        )
+
+        self.final_norm = nn.LayerNorm(embed_dim, eps=norm_eps)
+
+    def forward(self, tokens: Tensor) -> Tensor:
```
calvinpelletier, line 73: Should this forward call support all the other things in our `TransformerDecoder` (mask, encoder_input, encoder_mask, input_pos)? Or should we implement these other things on an as-needed basis? For diffusion, this should be sufficient.

Likewise, should I add support for a KV-cache?

torchtune/models/clip/_tokenizer.py
```diff
@@ -0,0 +1,211 @@
...
+        self.bpe_ranks = {merge: i for i, merge in enumerate(merges)}
+
+        self.pat = re.compile(
+            r"""<\|startoftext\|>|<\|endoftext\|>|'s|'t|'re|'ve|'m|'ll|'d|[\p{L}]+|[\p{N}]|[^\s\p{L}\p{N}]+""",
+            re.IGNORECASE,
+        )
+
+        self.sot_token = self.encoder["<|startoftext|>"]
+        self.eot_token = self.encoder["<|endoftext|>"]
+        self.pad_token = self.eot_token
```
calvinpelletier, line 58: should these be configurable in the constructor like we do for other tokenizers? I don't really think it's needed imo

torchtune/models/clip/_tokenizer.py
```diff
@@ -0,0 +1,211 @@
...
+
+        self.cache = {
+            "<|startoftext|>": "<|startoftext|>",
+            "<|endoftext|>": "<|endoftext|>",
+        }
+
+    def __call__(self, texts: List[str]) -> torch.Tensor:
```
calvinpelletier, line 65: I added an extra method that's not in the tokenizer interface to make it easy to get a batch of token sequences for CLIP:
`tokens_tensor = tokenizer(list_of_texts)`.

Should this function be moved out of the tokenizer class or renamed or anything?

torchtune/models/clip/_tokenizer.py
```diff
@@ -0,0 +1,211 @@
...
+    def __init__(self, vocab_path: Path, max_seq_len: int = 77, truncate: bool = True):
+        self.max_seq_len = max_seq_len
+        self.truncate = truncate
+
+        self.byte_encoder = _bytes_to_unicode()
+        self.byte_decoder = {v: k for k, v in self.byte_encoder.items()}
+
+        with gzip.open(vocab_path) as f:
+            merges = f.read().decode("utf-8").split("\n")
+        merges = [tuple(merge.split()) for merge in merges[1:48895]]
```
calvinpelletier, line 40: the CLIP tokenizer awkwardly uses just the first ~1/5th of its vocab file, resulting in this magic number 48895 and making this class more fragile (e.g. if someone tries to pass in their own custom vocab file, it'll get unexpectedly truncated). It might be worth manually truncating the file itself and using that instead (but we'd need to add it to git or host it somewhere). Or alternatively, we could use huggingface's clip vocab, which I assume has already done this

tests/torchtune/models/clip/test_clip_tokenizer.py
RdoubleA: Yes I would first verify against a reference implementation that you get forward pass parity, then you can just randomly initialize a small CLIP and use the output as the "golden value" for the test

torchtune/models/clip/_text_encoder.py
```diff
@@ -0,0 +1,109 @@
...
+        x = self.token_embedding(tokens) + self.position_embedding
+
+        # Encoder [b, s, d] -> [b, s, d]
+        x = self.encoder(x)
+        x = self.final_norm(x)
+
+        # Select the output of the EOS token for each encoding in the batch
+        # [b, s, d] -> [b, d]
+        eos_token_positions = tokens.argmax(dim=-1)
+        x = x[torch.arange(bsz, device=x.device), eos_token_positions]
```
pbontrager, line 107: nit: can you use "x.take_along_dim(eos_token_positions[:, None, None])" here?

torchtune/models/clip/_text_encoder.py
pbontrager: This doesn't need the same forward signature as TransformerDecoder as it won't be used the same way, but I think it should have a supset of that, maybe "tokens, mask, and input_pos". input_pos could be used instead of argmax for the forward. Or you can just add a todo around that for now and move forward as is.

torchtune/models/clip/_text_encoder.py
```diff
@@ -0,0 +1,109 @@
...
+        num_layers: int = 12,
+        norm_eps: float = 1e-5,
+    ):
+        super().__init__()
+        self.max_seq_len = max_seq_len
+
+        self.token_embedding = nn.Embedding(vocab_size, embed_dim)
+        self.position_embedding = nn.Parameter(torch.empty(max_seq_len, embed_dim))
+
+        self.encoder = nn.Sequential(
```
pbontrager, line 46: For consistency I think this could be structured more like the [Llama3VisionProjectionHead](https://github.com/pytorch/torchtune/blob/e1caa9f82fea24d728f9b244a9dd1957f5ed7465/torchtune/models/llama3_2_vision/_encoder.py#L15) code with a module that takes in "layers,  norm, and embeddings" and then define the text model as a function in component builders. It would make it easier to support a LoRA version of the model.

torchtune/models/clip/_model_builders.py
```diff
@@ -1,4 +1,63 @@
...
+
+from torchtune.models.clip._text_encoder import CLIPTextEncoder
+from torchtune.models.clip._tokenizer import CLIPTokenizer
+from torchtune.models.clip._transform import CLIPImageTransform
+from torchtune.utils._download import TORCHTUNE_LOCAL_CACHE_FOLDER, download_file
+
+CLIP_VOCAB_URL = "https://github.com/openai/CLIP/raw/refs/heads/main/clip/bpe_simple_vocab_16e6.txt.gz"
+
+
+def clip_tokenizer(
```
pbontrager, line 16: Take a look at some of the other tokenizer builder signatures. We've tried to standardize around support prompt templates from users here, though it might not makes sense as CLIP doesn't really use these.

torchtune/models/clip/_model_builders.py
```diff
@@ -1,4 +1,63 @@
...
+            Default: True
+        max_seq_len (bool): Context length
+            Default: 77
+        truncate (bool): Truncate the token sequence if it exceeds max_seq_len (otherwise raises AssertionError)
+            Default: True
+
+    Returns:
+        CLIPTokenizer: Instantiation of the CLIP text tokenizer
+    """
+    if not vocab_path.exists():
```
pbontrager, line 38: I'd prefer to remove all of this logic and just rely on the user downloading this themselves. We should assume that by default they're going to get it from HF.

torchtune/models/clip/_model_builders.py
```diff
@@ -1,4 +1,63 @@
...
+        assert download_if_missing, f"Missing CLIP tokenizer vocab: {vocab_path}"
+        download_file(CLIP_VOCAB_URL, vocab_path)
+
+    return CLIPTokenizer(vocab_path, max_seq_len=max_seq_len, truncate=truncate)
+
+
+def clip_text_encoder_large() -> CLIPTextEncoder:
```
pbontrager, line 45: The name for this should be clip_text_vit_l14 (are there anymore qualifiers in the hf official name?)

tests/torchtune/models/clip/test_clip_tokenizer.py
pbontrager: It should be an asset

tests/torchtune/models/clip/test_clip_tokenizer.py
```diff
@@ -0,0 +1,44 @@
...
+        correct_tokens = [
+            [49406, 320, 9706, 11476, 962, 518, 3293, 49407],
+            [49406, 320, 12695, 2215, 6799, 49407],
+        ]
+        for token_seq in correct_tokens:
+            _pad_token_sequence(token_seq)
+        tokens_tensor = tokenizer(texts)
+        assert tokens_tensor.tolist() == correct_tokens
+
+    def test_text_cleaning(self, tokenizer):
```
pbontrager, line 30: I don't think we need to support this.

tests/torchtune/models/clip/test_clip_tokenizer.py
pbontrager: What Rafi said

pbontrager (commented):
This is a great PR! I left some comments around some standard patterns we try to follow but aside from that, this looks very solid and clean.

torchtune/models/clip/_tokenizer.py
calvinpelletier: why would the clip tokenizer expect message objects?

torchtune/models/clip/_tokenizer.py
```diff
@@ -0,0 +1,211 @@
...
+    """
+    pairs = set()
+    prev_char = word[0]
+    for char in word[1:]:
+        pairs.add((prev_char, char))
+        prev_char = char
+    return pairs
+
+
+def _clean_text(text):
```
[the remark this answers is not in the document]
calvinpelletier: this is needed if we want perfect parity with the HF and OpenAI implementations of the CLIP tokenizer, but I agree that it would be better to remove the `ftfy` dependency and expect the user to clean their dataset. I'll remove this and leave a comment in the code

torchtune/training/checkpointing/_checkpointer.py
```diff
@@ -488,6 +489,10 @@ def load_checkpoint(self) -> Dict[str, Any]:
                     "supported_aspect_ratios", None
                 ),
             )
+        elif self._model_type == ModelType.CLIP_TEXT:
```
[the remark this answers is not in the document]
calvinpelletier: how would this work? the `clip_hf_to_tune` function needs to know whether it's creating torchtune parameters for the text or the vision model. and the only relevant information that the checkpointer has access to is the `model_type`, right? so the model type needs to specify text or vision?

torchtune/utils/_download.py
```diff
@@ -0,0 +1,48 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+from pathlib import Path
```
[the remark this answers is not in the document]
calvinpelletier: I was using the official OpenAI clip tokenizer, but I'll switch to HF

torchtune/models/clip/_model_builders.py
calvinpelletier: yeah I figured that didn't make sense since no one is going to be doing chat templates, etc. with CLIP. 

But I could include it as an option since maybe there's some other kind of template that would be used. I could see users wanting to do some simple templating like e.g. dreambooth's "a {unique_identifier} {class}".

But I don't think we should structure it like other tokenizers/templates that support a list of messages, because no one will want that for CLIP, especially with its 77 context length

torchtune/models/clip/_tokenizer.py
pbontrager: We discussed this offline. Our decoder tokenizers expect messages, which is what is returned in the recipes. Since this is for an encoder though, it is fine for now to expect text directly since presumably the encoder tokenizer would not be directly exposed in the recipe.

torchtune/training/checkpointing/_checkpointer.py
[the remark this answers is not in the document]
pbontrager: We'd just have a single convert mapping dictionary and map what happens to be there. I don't think the function actually needs to know which fraction of the weights it's converting.

pbontrager (commented):
I'd still like the model type to be unified to just CLIP to keep the list shorter. But I'll approve when @RdoubleA signs off on the tokenizer.

torchtune/training/checkpointing/_checkpointer.py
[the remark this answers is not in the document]
calvinpelletier: Every param will be there since HF combines the vision and text params. So the function needs to know what to ignore

torchtune/models/clip/_tokenizer.py
```diff
@@ -0,0 +1,206 @@
...
+    Based on the official implementation here:
+    https://github.com/openai/CLIP/blob/main/clip/simple_tokenizer.py
+
+    Args:
+        path (PathLike): the path to the CLIP merges file
+        max_seq_len (int): the context length (all CLIP models use 77)
+        truncate (bool): whether to truncate the text when longer than max_seq_len
+    """
+
+    def __init__(self, path: PathLike, max_seq_len: int = 77, truncate: bool = True):
```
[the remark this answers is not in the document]
calvinpelletier: yeah it's str+pathlib.Path. I'll change it to just str

torchtune/models/clip/_tokenizer.py
```diff
@@ -0,0 +1,206 @@
...
+            str: The decoded text.
+        """
+        text = "".join([self.decoder[token] for token in tokens])
+        return (
+            bytearray([self.byte_decoder[c] for c in text])
+            .decode("utf-8", errors="replace")
+            .replace("</w>", " ")
+        )
+
+    def __call__(self, texts: List[str]) -> torch.Tensor:
```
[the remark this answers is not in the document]
calvinpelletier: In the case of diffusion, the text will be tokenized by both the CLIP tokenizer and the T5 tokenizer, so it probably makes sense to have the model handle this rather than the dataset. We won't be using the `SFTDataset` or `Message`s, we'll probably make new abstractions for image-text pairs.

I vote that we leave this as is for now, but revisit it once we decide on an overall design for our diffusion code. We might end up altering `__call__` or removing it entirely, but it's currently not clear

torchtune/models/clip/_tokenizer.py
```diff
@@ -0,0 +1,206 @@
...
+        result = torch.full(
+            (len(texts), self.max_seq_len), self.pad_token, dtype=torch.int
+        )
+        for i, text in enumerate(texts):
+            tokens = self.encode(text)
+            result[i, : len(tokens)] = torch.tensor(tokens)
+        return result
+
+    def _bpe(self, token):
+        if token in self.cache:
```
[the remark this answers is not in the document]
calvinpelletier: nope! i checked the process's memory usage after tokenizing an entire dataset and it was tiny.

`lru_cache` did cause very slight performance degradation but it's negligible so i'll switch to `lru_cache` for simplicity

torchtune/models/clip/_tokenizer.py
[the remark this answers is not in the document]
calvinpelletier: Re-opening this conversation because I ran a parity check on an entire text dataset and I only found one issue: CLIP's text cleaning replaces ’ with ' 

So I re-added a version of the `_clean_text` function that does only this ([link](https://github.com/calvinpelletier/torchtune/blob/5fe86ae50ae0b20867d49a6f168e12a2773afbaf/torchtune/models/clip/_tokenizer.py#L223-L227)). 

This gives us functionally perfect parity with the official implementation without needing to add `ftfy` as a dependency, but it is kinda awkward. Should I remove it and expect the user to clean their dataset themselves? Or should I leave it so that our CLIP tokenizer won't see any strange tokens that it wasn't trained on?

@pbontrager @RdoubleA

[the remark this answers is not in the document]
pbontrager: I'm okay with leaving this in as it's very simple, but I don't think we strictly need it.

torchtune/models/clip/_tokenizer.py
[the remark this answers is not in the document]
pbontrager: The __call__ method is standardized and should look like this `def __call__(self, sample: Mapping[str, Any], inference: bool = False) -> Mapping[str, Any]:`. The inference flag doesn't have to do anything but should be there. As Rafi said, all tokenization happens in the dataset so it's not on the main execution path. For diffusion the data pipeline looks like
`Raw Dataset: (text, image) -> TuneDataset -> ModelTransform: {tokens, tensor} -> collation`

torchtune/models/clip/_component_builders.py
```diff
@@ -181,6 +179,61 @@ def clip_vision_encoder(
     )
 
 
+def clip_text_encoder(
+    vocab_size: int = 49408,
```
RdoubleA, line 183: we don't normally keep defaults in the component builders right? since these values correspond to the specific ViT large patch

torchtune/models/clip/_model_builders.py
```diff
@@ -1,4 +1,51 @@
...
+
+    Returns:
+        CLIPTokenizer: Instantiation of the CLIP text tokenizer
+    """
+    return CLIPTokenizer(path, max_seq_len=max_seq_len, truncate=truncate)
+
+
+def clip_text_vit_large_patch14() -> CLIPTextEncoder:
+    """
+    Builder for the CLIP text encoder for CLIP-ViT-L/14.
```
RdoubleA, line 35: Any link or ref for where this was introduced / prominently used?

torchtune/models/clip/_model_builders.py
```diff
@@ -1,4 +1,51 @@
...
+    max_seq_len: int = 77,
+    truncate: bool = True,
+) -> CLIPTokenizer:
+    """
+    Builder for the CLIP text tokenizer.
+
+    Args:
+        path (str): Path to the CLIP merges file
+        max_seq_len (bool): Context length
+            Default: 77
```
RdoubleA, line 23: nit: put on same line

torchtune/models/clip/_text_encoder.py
```diff
@@ -0,0 +1,105 @@
...
+
+import torch
+from torch import nn, Tensor
+
+from torchtune.modules.attention_utils import _MaskType
+
+
+class CLIPTextEncoder(nn.Module):
+    """
+    Text encoder for CLIP.
```
RdoubleA, line 18: mind briefly describing the architecture? probably a normal transformer, but uses a different MLP activation?

torchtune/models/clip/_text_encoder.py
```diff
@@ -0,0 +1,105 @@
...
+        embed_dim (int): embedding/model dimension size, default 768
+        num_layers (int): number of transformer layers, default 12
+    """
+
+    def __init__(
+        self,
+        *,
+        layer: nn.Module,
+        final_norm: nn.Module,
+        vocab_size: int = 49408,
```
RdoubleA, line 34: same question here - do we want to hardcode these defaults if they're tied to a specific version of CLIP?

torchtune/models/clip/_text_encoder.py
```diff
@@ -0,0 +1,105 @@
...
+            x = layer(
+                x,
+                mask=mask,
+            )
+        x = self.final_norm(x)
+
+        # Select the output of the EOS token for each encoding in the batch
+        # [b, s, d] -> [b, d]
+        # TODO: handle the case when the EOS token is not the highest token ID
+        eos_token_positions = tokens.argmax(dim=-1)
```
RdoubleA, line 93: is the EOS token not always at the end?

RdoubleA (approved):

torchtune/models/clip/_text_encoder.py
pbontrager: I don't think it's that important since this is a one off model definition. The main thing is to also define this same number in the builder.

torchtune/models/clip/_text_encoder.py
calvinpelletier: nope because there is padding after the end of the sequence

torchtune/models/clip/_text_encoder.py
calvinpelletier: it's just a normal transformer. the MLP activation is still GELU, it's just a faster and less-precise version of it (tho not actually faster these days, this is a relic of the ancient year of 2021)

torchtune/models/clip/_text_encoder.py
RdoubleA: I'm worried about this line because it can silently fail. is there a way we can look for the particular EOS id? that would require pass the token id in to the class which isn't great, but if a user changes the EOS token id for some reason how can we catch this failure

pbontrager (approved):
This looks good now, thanks for doing all the changes! I think you could add a few line description of the models in the docstrings where Rafi commented along with an arxiv link. But I'll approve it now.

torchtune/models/clip/_text_encoder.py
```diff
@@ -87,10 +90,9 @@ def forward(
             )
         x = self.final_norm(x)
 
-        # Select the output of the EOS token for each encoding in the batch
+        # Select the output of the EOT token for each encoding in the batch
         # [b, s, d] -> [b, d]
-        # TODO: handle the case when the EOS token is not the highest token ID
-        eos_token_positions = tokens.argmax(dim=-1)
+        eos_token_positions = (tokens == self.eot_token).int().argmax(dim=-1)
```
RdoubleA, line 95: can do `(tokens == self.eot_token).nonzero()`

calvinpelletier: we don't want all positions of the eot token, just the first one (argmax gives first position where they match)

PR #1973, opened 2024-11-08
2D RoPE + CLIP updates

RdoubleA:
## Context
Two-dimensional rotary positional embeddings have been added to vision transformers to improve performance. This was explored in papers such as [Eva-02-CLIP](https://arxiv.org/pdf/2303.11331), which found that 2D RoPE improved performance and had more stable training as opposed to 1D RoPE for images. Another novel architecture FiT ([Flexible Vision Transformer for Diffusion Model](https://arxiv.org/pdf/2402.12376)) similarly employs 2D RoPE for image resolution generalization. Pixtral, a multimodal LLM, also uses a similar 2D RoPE mechanism as well, as seen in [Hugging Face](https://github.com/huggingface/transformers/blob/main/src/transformers/models/pixtral/modeling_pixtral.py#L48). A full survey of 2D RoPE can be found in this [paper](https://arxiv.org/pdf/2403.13298).

Here, we add `VisionRotaryPositionalEmbeddings` as a general component. The forward is identical to `RotaryPositionalEmbeddings`; the major difference is in the `build_rope_cache`, where we need to take into account the x and y positions of the patches in the image grid, as defined by `patch_size` and `tile_size`. It simply applies 1D RoPE with half the usual dim on x and y independently and concatenates them together.

This is exposed as `use_rope` in the `clip_vision_encoder` builder, which will enable 2D RoPE. I also include some various fixes for CLIP and additional configurability

## Changelog
* Add `VisionRotaryPositionalEmbeddings` with tests
* Expose ability to configure RoPE on clip encoder
* `intermediate_act` is not used in the `clip_vision_encoder`. This is superceded by `activation`, so I just removed it
* `apply_lora_to_output` is not used in `lora_clip_vision_encoder`. There is a TODO to add this to the `CLSProjection`. I removed it so users don't accidentally configure it and get unexpected behavior
* CLIP's `CLSEmbedding` adds the CLS token to the beginning of the input sequence. I expose this as a configurable parameter to allow the option of adding it to the end of the sequence instead, which some variations of CLIP may do.
* `attn_bias` was not being propagated in `lora_clip_vision_encoder`, so I fixed this and changed default to False to match current behavior.
* Typing fixes for llama3 rope parameters
* Move Phi3 rope tests to its own file

## Test plan
- Recreated `clip_vision_encoder` with `append_cls_token=True` and manually verified that cls token embedding is in correct position, added test
- Tests for `VisionRotaryPositionalEmbeddings` and parity check against an internal reference implementation


joecummings (commented):
buttttt tests that run in CI?

joecummings (approved):
This looks good to me, but should probably get @pbontrager 's eyes on this as well.

torchtune/modules/position_embeddings.py
```diff
@@ -125,3 +120,146 @@ def forward(
...
+    The position embedding is simply applied to the x-axis and y-axis separately.
+
+    Note: This module assumes the CLS token embedding is appended at the end of the sequence.
+
+    Args:
+        patch_size (int): The size of each patch. Used to divide the tiles into patches.
+            E.g. for ``patch_size=40``, a tile of shape (400, 400) will have 10x10 grid of patches.
+        tile_size (int): The size of your image tiles, if the image was tile-cropped in advance. Otherwise,
+            the size of the input image. In this case, the function will consider your image as a single tile.
+            with shape (40, 40) each.
```
acisseJZhong, line 139: a single tile with shape (400, 400)? if max_num_tiles=1, then tile size should be equal to patch_size?

RdoubleA: ooh ya good catch, let me update this and the docstring where I copied this from

felipemello1: I think that "your image as a single tile. with shape (40, 40) each." ---> "your image as a single tile with shape (400, 400)."

I didnt read the code, just sharing what i know from llama 3.2 vision. This explains it visually: https://github.com/pytorch/torchtune/blob/bd89bb88b6092de74e79356ddb193b9e4b614d46/torchtune/modules/vision_transformer.py#L61

A image is broken into tiles. E.g, if max_num_tiles=4 and tile_size=400, a 1600x1600 image is broken into 4 tiles of 400x400, e.g. (num images, num tiles, x, y) = (1,4,400x400)

Then each tile is broken down into patches.

If max_num_tiles=1, then it means that 1 image = 1 tile. The tile dimension doesnt go away, e.g. (num images, num tiles, x, y) = (1,1,x, y)

Issue #1975, opened 2024-11-08
Convert all non-`RGB` images to `RGB` in `CLIPImageTransform`


vancoykendall:
In the `CLIPImageTransfrom`, only `RGBA` images get converted to `RGB`. I don't see a reason to not convert all non-`RGB` images. I have a multimodal dataset with mostly `RGB` images, but a handful are in `CMYK` and `L`. I'm finetuning a Llama3.2 VLM which uses the `CLIPImageTransform` and my first training crashed in the middle because a `CMYK` image finally was loaded, didn't get converted, and the size didn't match.

https://github.com/pytorch/torchtune/blob/96dea6159e25914ae85cdc7c13b6cf730540b135/torchtune/models/clip/_transform.py#L161-L163

I noticed the code in the eval recipe actually does convert all non-`RGB` images to `RGB`:

https://github.com/pytorch/torchtune/blob/96dea6159e25914ae85cdc7c13b6cf730540b135/recipes/eleuther_eval.py#L168-L170

ebsmothers:
Thanks for creating the issue. Actually this was done as a bit of a one-off so we probably can consider changing to match the eval recipe. cc @pbontrager in case you have any strong objections to this, but otherwise @vancoykendall we'd love to have you open a PR with the change if you're interested.

vancoykendall:
@ebsmothers sounds good. Also, I think like the eval code I can remove the check for the `mode` attribute as all `PIL.Image`s have the attribute

vancoykendall:
@ebsmothers Just created a pr https://github.com/pytorch/torchtune/pull/1976

ebsmothers:
Closing now that the fix has landed. Thanks again!

PR #1976, opened 2024-11-08
Convert all non-rgb images to rgb

vancoykendall:
@ebsmothers Ensure that all images are converted to `RGB` format. No longer need to separately convert grayscale images to rgb either. Addresses https://github.com/pytorch/torchtune/issues/1975


ebsmothers (approved):
Thanks @vancoykendall for the fix! Once CI is green this is good to merge

PR #1961, opened 2024-11-07
Update KV Cache to use num_kv_heads instead of num_heads

mirceamironenco:
#### Context
What is the purpose of this PR? Is it to
- [x] add a new feature
- [ ] fix a bug
- [ ] update tests and/or documentation
- [ ] other (please add here)

Please link to any issues this PR addresses.

#### Changelog
What are the changes made in this PR?
* Update KV Cache to use num_kv_heads instead of num_heads for more memory-efficient generation.
* Simplified/adapted mha fwd pass to have the expand happen after the kv cache.

#### Test plan
Please make sure to do each of the following if applicable to your PR. If you're unsure about any one of these just ask and we will happily help. We also have a [contributing page](https://github.com/pytorch/torchtune/blob/main/CONTRIBUTING.md) for some guidance on contributing.

- [x] run pre-commit hooks and linters (make sure you've first installed via `pre-commit install`)
- [ ] add [unit tests](https://github.com/pytorch/torchtune/tree/main/tests/torchtune) for any new functionality
- [ ] update [docstrings](https://github.com/pytorch/torchtune/tree/main/docs/source) for any new or updated methods or classes
- [x] run unit tests via `pytest tests`
- [ ] run recipe tests via `pytest tests -m integration_test`
- [ ] manually run any new or modified recipes with sufficient proof of correctness
- [ ] include relevant commands and any other artifacts in this summary (pastes of loss curves, eval results, etc.)

#### UX
If your function changed a public API, please add a dummy example of what the user experience will look like when calling it.
Here is a [docstring example](https://github.com/pytorch/torchtune/blob/6a7951f1cdd0b56a9746ef5935106989415f50e3/torchtune/modules/vision_transformer.py#L285)
and a [tutorial example](https://pytorch.org/torchtune/main/tutorials/qat_finetune.html#applying-qat-to-llama3-models)

- [ ] I did not change any public API
- [ ] I have added an example to docs or docstrings

If one changes `generation.yaml` to use llama 3.1 8b (num_kv_heads < num_heads), and the prompt/max_new_toks:
```
INFO:torchtune.utils._logging:Running InferenceRecipe with resolved config:

checkpointer:
  _component_: torchtune.training.FullModelHFCheckpointer
  checkpoint_dir: /tmp/Meta-Llama-3.1-8B-Instruct/
  checkpoint_files:
  - model-00001-of-00004.safetensors
  - model-00002-of-00004.safetensors
  - model-00003-of-00004.safetensors
[...]

PR #1989, opened 2024-11-12
Adding MM eval tests / attention bugfixes

SalmanMohammadi:
#### Context
- [ ] add a new feature
- [x] fix a bug
- [ ] other (please add here)

#### Changelog
* Adds eleuther eval tests for dummy MM models. These tests are running on GPU since it's super slow on CPU.
* Fixes a bug introduced by https://github.com/pytorch/torchtune/pull/1961 which missed expanding KV-cache tensors for cross-attention layers.
* Fixes a bug introduced in the same PR which applied `k_norm` *after* KV-cache update, rather than before, which caused issues when retrieving encoder kv-caches which were not normalized.
* ^^Added a unit test which catches both of the above cases.
* Fixed a bug which didn't check for `caches_enabled` when attempting to retrieve kv-caches for encoder inputs

#### Test plan

#### UX

### DIFF
--- a/tests/torchtune/modules/test_transformer_decoder.py
+++ b/tests/torchtune/modules/test_transformer_decoder.py
@@ -206,7 +206,7 @@ class TestTransformerCrossAttentionLayer:
                 encoder_input=input_y,
                 encoder_mask=mask,
             )
-            # the second pass should retrieve the cached inputs and produce
+            # the second pass should just retrieve from the kv-cache and produce
             # identical outputs
             output = transformer_layer(
                 input_x,
--- a/tests/torchtune/modules/test_transformer_decoder.py
+++ b/tests/torchtune/modules/test_transformer_decoder.py
@@ -184,10 +184,47 @@ class TestTransformerCrossAttentionLayer:
         return transformer_layer
 
     @mps_ignored_test()
+    def test_forward_kv_cache(
+        self,
+        input: Tuple[torch.Tensor, torch.Tensor, torch.Tensor],
+        transformer_layer: TransformerCrossAttentionLayer,
+        input_params: Tuple[int, int, int, int],
+    ):
+
+        b, _, encoder_seq_len, _ = input_params
+        transformer_layer.setup_caches(
+            batch_size=b,
+            dtype=torch.float32,
+            encoder_max_seq_len=encoder_seq_len,
+            decoder_max_seq_len=None,
+        )
+        input_x, input_y, mask = input
+        with torch.no_grad():
+            # make an initial forward pass which should fill the encoder cache
+            first_output = transformer_layer(
+                input_x,
+                encoder_input=input_y,
+                encoder_mask=mask,
+            )
+            # the second pass should retrieve the cached inputs and produce
+            # identical outputs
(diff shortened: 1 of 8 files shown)


torchtune/modules/attention.py
```diff
@@ -255,6 +255,14 @@ def forward(
                 )
             k = self.kv_cache.k_cache
             v = self.kv_cache.v_cache
+
+            # If needed, expand the key and value tensors to have the same shape
+            # as the query tensor by copying values across the relevant dim
+            # k,v shape: [b, n_kv, s, h_d] -> [b, n_h, s, h_d]
+            if self.num_heads != self.num_kv_heads:
```
RdoubleA, line 262: Is there a test that can catch the error you noticed before you made this change?

torchtune/modules/attention.py
```diff
@@ -255,6 +255,14 @@ def forward(
...
+                expand_shape = (-1, -1, q_per_kv, -1, -1)
```
RdoubleA, line 263: any chance you can use variable names here instead of -1 for explicitness

torchtune/modules/transformer.py
```diff
@@ -781,7 +781,7 @@ def setup_caches(
             isinstance(l, TransformerCrossAttentionLayer) for l in self.modules()
         )
         has_decoder_layers = any(
-            isinstance(l, TransformerSelfAttentionLayer) for l in self.layers
+            isinstance(l, TransformerSelfAttentionLayer) for l in self.modules()
```
RdoubleA, line 784: :O

torchtune/modules/attention.py
SalmanMohammadi: The tests in this PR will catch it (it's how I found it) since we perform generation with a vision model with KV-caches enabled. The most atomic test that *should* catch it is setting up a cross attention layer with KV-caches and kv_heads < num_heads somewhere like [here](https://github.com/pytorch/torchtune/blob/main/tests/torchtune/modules/test_transformer_decoder.py)

torchtune/modules/attention.py
```diff
@@ -249,7 +249,7 @@ def forward(
             q = self.q_norm(q)
 
         if y is None:
-            if self.kv_cache is None:
+            if self.kv_cache is None or not self.cache_enabled:
```
SalmanMohammadi, line 252: This should have been added in #1763

tests/torchtune/modules/test_transformer_decoder.py
```diff
@@ -183,11 +183,48 @@ def transformer_layer(
...
+
+        assert_expected(first_output.shape, output.shape)
+        assert_expected(first_output.mean(), output.mean())
+
     @mps_ignored_test()
     def test_forward(
         self,
         input: Tuple[torch.Tensor, torch.Tensor, torch.Tensor],
-        transformer_layer: TransformerSelfAttentionLayer,
+        transformer_layer: TransformerCrossAttentionLayer,
```
SalmanMohammadi, line 227: idek why we're typing tests

RdoubleA (approved):
Looks good to go!

Issue #1751, opened 2024-10-04
lr scheduler is not optional


felipemello1:
currently the recipes expect as lr scheduler. If deleted from the config, it crashes. This shouldnt be the case. If the config doesnt provide a scheduler, we should just not update it.

SalmanMohammadi:
I believe this PR added optional support for the LR scheduler in the FFT recipe which we could copy over https://github.com/notifications.

felipemello1:
closing in favor of https://github.com/pytorch/torchtune/issues/2021

PR #1991, opened 2024-11-12
Remove lr_scheduler requirement in lora_dpo_single_device

thomasjpfan:
#### Context
- [x] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

Towards https://github.com/pytorch/torchtune/issues/1751

For testing, I initially wanted to `monkeypatch` the `recipes` module, but it was not designed to be imported from tests: https://github.com/pytorch/torchtune/blob/912af640a422d84ce261c100f804490a5c12e8e7/recipes/__init__.py#L7-L8. So for this PR, I added a new integration test that makes a new config without a scheduling and run `tune run` for one epoch.

#### Changelog
* Removes `lr_scheduler` requirement in lora_dpo_single_device

#### Test plan

#### UX


recipes/lora_dpo_single_device.py
```diff
@@ -243,7 +243,7 @@ def setup(self, cfg: DictConfig) -> None:
         # Learning rate scheduler can only be set up after number of steps
         # has been computed
         self._lr_scheduler = self._setup_lr_scheduler(
-            cfg_lr_scheduler=cfg.lr_scheduler,
+            cfg_lr_scheduler=cfg.get("lr_scheduler", None),
```
RdoubleA, line 246: may be less indirection to just check directly if lr_scheduler exists and set to None here, instead of calling the setup method only to return None:
```
cfg_lr_scheduler = cfg.get("lr_scheduler", None)
self._lr_scheduler = self._setup_lr_scheduler(...) if cfg_lr_scheduler is not None else None
```

tests/recipes/test_lora_dpo_single_device.py
```diff
@@ -191,3 +191,49 @@ def test_save_and_load_merged_weights(self, tmpdir, monkeypatch):
         llama2_model.load_state_dict(sd)
         merged_ckpt_out = llama2_model(inputs)
         torch.testing.assert_close(baseline_out, merged_ckpt_out, rtol=1e-5, atol=1e-5)
+
+    @pytest.mark.integration_test
+    def test_lr_scheduler_optional(self, tmpdir, monkeypatch):
```
RdoubleA, line 196: this is a great test but it seems expensive just to check that the lr_scheduler is not created. Maybe you could just call the recipe's `setup()` method with a toy model and check that `self._lr_scheduler == None`?

RdoubleA (commented):
Thanks for making this update, had a few comments on implementation

tests/recipes/test_lora_dpo_single_device.py
thomasjpfan: I initially wanted to go with this solution, but I do not think the `recipes` were designed to be imported in tests:

https://github.com/pytorch/torchtune/blob/912af640a422d84ce261c100f804490a5c12e8e7/recipes/__init__.py#L7-L8

thomasjpfan: I initially wanted to go with this approach, but I do not think we are allowed to import anything in `recipes` (i.e. `LoRADPORecipeSingleDevice`) from tests:

https://github.com/pytorch/torchtune/blob/912af640a422d84ce261c100f804490a5c12e8e7/recipes/__init__.py#L19-L23

recipes/lora_dpo_single_device.py
felipemello1: That makes sense to me, Rafi, but upon second thought, i like the idea of handling everything inside of the setup_lr_scheduler, including the log_info. What do you think?

felipemello1:
Thanks for the pr, @thomasjpfan! I left a couple of comments. After Rafi and Evan reply, i am good with merging it. PS: after the changes for lora_dpo_single_device are done, would you be willing to add it to the other recipes too? Its ok if the answer is no. We can do it as a follow up :)

recipes/lora_dpo_single_device.py
RdoubleA: Sure, no strong opinions

thomasjpfan:
> PS: after the changes for lora_dpo_single_device are done, would you be willing to add it to the other recipes too? Its ok if the answer is no. We can do it as a follow up :)

I'll be happy to add it to the other recipes as a followup. I started off with one recipe to see how much test converge is required.

felipemello1:
Just before we approve it, do you mind running it with wandb logger (or any other logger) and capturing the learning rate with/without the scheduler, as a sanity check?

You can append this to your cli command:

```
tune run <...> /
metric_logger=torchtune.training.metric_logging.WandBLogger 
```

thomasjpfan:
With `lr_scheduler._component_=`, the LR changes:

Screenshot 2024-11-15 at 9 28 24 PM

I was not able to get `~lr_scheduler._component_` to remove the config, so I just removed the `lr_scheduler` from the yaml, and the lr stays constant:

Screenshot 2024-11-15 at 9 28 28 PM

RdoubleA (approved):

felipemello1:
Thank you @thomasjpfan , i created an issue to add to the other recipe. We would love your contribution if you have interest: https://github.com/pytorch/torchtune/issues/2021

PR #1999, opened 2024-11-13
Update Qwen2.5 configs

joecummings:
1. I turned activation checkpointing off for all 0.5B models and for 1.5B LoRA models. No point. 
2. I turned on memory logging

Everything else is cosmetic.


recipes/configs/qwen2_5/0.5B_full.yaml
```diff
@@ -0,0 +1,74 @@
...
+#
+# To launch on 2 devices, run the following command from root:
+#   tune run --nproc_per_node 2 full_finetune_distributed --config qwen2_5/0.5B_full
+#
+# You can add specific overrides through the command line. For example
+# to override the checkpointer directory while launching training:
+#   tune run --nproc_per_node 2 full_finetune_distributed --config qwen2_5/0.5B_full checkpointer.checkpoint_dir=<YOUR_CHECKPOINT_DIR>
+#
+# This config works best when the model is being fine-tuned on 2+ GPUs.
+# Single device full finetuning requires more memory optimizations. It's
```
ebsmothers, line 15: nit: not really true in this case

recipes/configs/qwen2_5/0.5B_full_single_device.yaml
```diff
@@ -0,0 +1,72 @@
...
+#
+# You can add specific overrides through the command line. For example
+# to override the checkpointer directory while launching training:
+#   tune run full_finetune_single_device --config qwen2_5/0.5B_full_single_device checkpointer.checkpoint_dir=<YOUR_CHECKPOINT_DIR>
+#
+# This config works only for training on single device.
+
+# Model arguments
+model:
+  _component_: torchtune.models.qwen2_5.qwen2_5_0_5b
```
ebsmothers, line 18: I understand that you're making the filename change 0_5B -> 0.5B for consistency with other configs, but honestly would prefer to just move everything to 0_5B so it matches the builders

calvinpelletier (commented):
I thought we use underscores instead of periods?: https://github.com/pytorch/torchtune/pull/1863#discussion_r1806701325

recipes/configs/qwen2_5/1.5B_full.yaml
```diff
@@ -0,0 +1,74 @@
...
+#
+# To launch on 2 devices, run the following command from root:
+#   tune run --nproc_per_node 2 full_finetune_distributed --config qwen2_5/1.5B_full
+#
+# You can add specific overrides through the command line. For example
+# to override the checkpointer directory while launching training:
+#   tune run --nproc_per_node 2 full_finetune_distributed --config qwen2_5/1.5B_full checkpointer.checkpoint_dir=<YOUR_CHECKPOINT_DIR>
+#
+# This config works best when the model is being fine-tuned on 2+ GPUs.
+# Single device full finetuning requires more memory optimizations. It's
```
ebsmothers, line 15: prob same here: 1.5 * 8 = 12, so most users will not need any optimizations to run this

recipes/configs/qwen2_5/1.5B_full_single_device.yaml
```diff
@@ -1,108 +1,75 @@
...
+output_dir: /tmp/Qwen2.5-1.5B-Instruct-finetune
 metric_logger:
   _component_: torchtune.training.metric_logging.DiskLogger
-  log_dir: ${output_dir}
-output_dir: /tmp/Qwen2_5-1_5B-Instruct-finetune
+  log_dir: ${output_dir}/logs
 log_every_n_steps: 1
-log_peak_memory_stats: False
-
-# Profiler (disabled)
```
ebsmothers, line 85: controversial take but if we want consistency we should leave these in. idc too much for this PR but I thought that was the whole point of a bunch of @felipemello1's changes. Either way would like to compress this config substantially in a separate PR

ebsmothers (approved):
I hate decimal points

joecummings:
> I thought we use underscores instead of periods?: [#1863 (comment)](https://github.com/pytorch/torchtune/pull/1863#discussion_r1806701325)

Yeah this is a misleading comment. We do use underscores for model builders, but the model should just get downloading to a directory with the exact same name as the model on the Hub.

recipes/configs/qwen2_5/1.5B_full_single_device.yaml
felipemello1: I am ok with making the profiler simpler is a separate PR

recipes/configs/qwen2_5/0.5B_full_single_device.yaml
felipemello1: for llama3.2 we added 3.2 to the path like you did

image

joecummings: This is the same thing we have here, no?

PR #2003, opened 2024-11-13
add flags to readme

felipemello1:
#### Context
- [ ] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

image


README.md
```diff
@@ -99,6 +99,49 @@ If you are interested in running on different hardware or with different models,
 
 &nbsp;
 
+### Optimization flags
+
+The following table demonstrates the impact of sequentially applying various optimization techniques available on torchtune to the Llama 3B model. Each technique is added on top of the previous one, except for LoRA and QLoRA, which do not use `opt_in_bwd` or `Adamw8bit`.
```
RdoubleA, line 104: Should clarify that it's Llama3.2 3B model

RdoubleA (approved):
This is beautiful. Can we get this table on our memory optimizations page on our live docs?

README.md
```diff
@@ -99,6 +99,49 @@ If you are interested in running on different hardware or with different models,
...
+
+**Baseline:**
+- **Model:** Llama 3B
+- **Batch size:** 5
```
ebsmothers, line 108: I suspect I know why you did this but imo it's a bit strange to use 5 as a batch size

README.md
```diff
@@ -99,6 +99,49 @@ If you are interested in running on different hardware or with different models,
...
+To reproduce QLoRA:
+```
+tune run lora_finetune_single_device --config llama3_2/3B_qlora_single_device \
+optimizer_in_bwd=False \
+enable_activation_checkpointing=True \
+enable_activation_offloading=True \
+optimizer._component_=torch.optim.AdamW \
+compile=True \
+dataset.packed=True \
+dataset.split=train[:5%] \
```
ebsmothers, line 135: Can probably remove this

README.md
```diff
@@ -99,6 +99,49 @@ If you are interested in running on different hardware or with different models,
...
+tokenizer.max_seq_len=2048 \
+gradient_accumulation_steps=1 \
+epochs=1 \
+batch_size=5 \
+loss=torchtune.modules.loss.CEWithChunkedOutputLoss
```
ebsmothers, line 140: I get that you include this (and others like e.g. `optimizer_in_bwd`) just to be explicit but I'd be in favor of either removing them or adding something like ` # enabled by default` to those lines to clarify. Alternatively can just add another column in the table to give the configs one-by-one (though not clear how to concisely represent LoRA and QLoRA rows in that case)

README.md
```diff
@@ -99,6 +99,49 @@ If you are interested in running on different hardware or with different models,
...
+- **Model:** Llama 3B
+- **Batch size:** 5
+- **Max seq len:** 2048
+- **Precision:** bf16
+- **Hardware:** A100
+
+| Technique | Peak Memory Active (GiB) | % Change Memory vs Previous | Tokens Per Second | % Change Tokens/sec vs Previous|
+|:--|:-:|:-:|:-:|:-:|
+| Baseline | 25.57 | - | 2979 | - |
+| [+ Packed Dataset](https://pytorch.org/torchtune/main/basics/packing.html) | 70.52 | 175.8% | 7203 | 141.8% |
```
ebsmothers, line 116: Add "+" signs to values in the delta column where they increase

README.md
```diff
@@ -99,6 +99,49 @@ If you are interested in running on different hardware or with different models,
...
+| [+ Compile](https://pytorch.org/tutorials/intermediate/torch_compile_tutorial.html) | 59.33 | -15.9% | 9200 | 27.7% |
+| [+ Chunked Cross Entropy](https://pytorch.org/torchtune/main/generated/torchtune.modules.loss.CEWithChunkedOutputLoss.html) | 48.69 | -17.9% | 9415 | 2.3% |
+| [+ Activation Ceheckpointing](https://pytorch.org/torchtune/main/tutorials/memory_optimizations.html#activation-checkpointing) | 25.28 | -48.1% | 7394 | -21.5% |
+| [+ Optimizer in bwd](https://pytorch.org/torchtune/main/tutorials/memory_optimizations.html#fusing-optimizer-step-into-backward-pass) | 23.92 | -5.4% | 7489 | 1.3% |
+| [+ Activation Offloading](https://pytorch.org/torchtune/main/tutorials/memory_optimizations.html#activation-offloading) | 22.34 | -6.6% | 7479 | -0.1% |
+| [+ Adamw8bit](https://pytorch.org/torchtune/main/tutorials/memory_optimizations.html#lower-precision-optimizers) | 17.57 | -21.4% | 7187 | -3.9% |
```
ebsmothers, line 122:
```suggestion
| [+ 8-bit AdamW](https://pytorch.org/torchtune/main/tutorials/memory_optimizations.html#lower-precision-optimizers) | 17.57 | -21.4% | 7187 | -3.9% |
```

README.md
```diff
@@ -99,6 +99,49 @@ If you are interested in running on different hardware or with different models,
...
+- **Precision:** bf16
+- **Hardware:** A100
+
+| Technique | Peak Memory Active (GiB) | % Change Memory vs Previous | Tokens Per Second | % Change Tokens/sec vs Previous|
+|:--|:-:|:-:|:-:|:-:|
+| Baseline | 25.57 | - | 2979 | - |
+| [+ Packed Dataset](https://pytorch.org/torchtune/main/basics/packing.html) | 70.52 | 175.8% | 7203 | 141.8% |
+| [+ Compile](https://pytorch.org/tutorials/intermediate/torch_compile_tutorial.html) | 59.33 | -15.9% | 9200 | 27.7% |
+| [+ Chunked Cross Entropy](https://pytorch.org/torchtune/main/generated/torchtune.modules.loss.CEWithChunkedOutputLoss.html) | 48.69 | -17.9% | 9415 | 2.3% |
+| [+ Activation Ceheckpointing](https://pytorch.org/torchtune/main/tutorials/memory_optimizations.html#activation-checkpointing) | 25.28 | -48.1% | 7394 | -21.5% |
```
ebsmothers, line 119:
```suggestion
| [+ Activation Checkpointing](https://pytorch.org/torchtune/main/tutorials/memory_optimizations.html#activation-checkpointing) | 25.28 | -48.1% | 7394 | -21.5% |
```

README.md
ebsmothers: Also should clarify that the baseline is full finetune

README.md
```diff
@@ -99,6 +99,49 @@ If you are interested in running on different hardware or with different models,
...
+| [+ Optimizer in bwd](https://pytorch.org/torchtune/main/tutorials/memory_optimizations.html#fusing-optimizer-step-into-backward-pass) | 23.92 | -5.4% | 7489 | 1.3% |
```
ebsmothers, line 120:
```suggestion
| [+ Fuse optimizer step into backward](https://pytorch.org/torchtune/main/tutorials/memory_optimizations.html#fusing-optimizer-step-into-backward-pass) | 23.92 | -5.4% | 7489 | 1.3% |
```

README.md
```diff
@@ -99,6 +99,49 @@ If you are interested in running on different hardware or with different models,
...
+| [+ Activation Offloading](https://pytorch.org/torchtune/main/tutorials/memory_optimizations.html#activation-offloading) | 22.34 | -6.6% | 7479 | -0.1% |
+| [+ Adamw8bit](https://pytorch.org/torchtune/main/tutorials/memory_optimizations.html#lower-precision-optimizers) | 17.57 | -21.4% | 7187 | -3.9% |
+| [LoRA](https://pytorch.org/torchtune/main/tutorials/lora_finetune.html) | 9.16 | -47.8% | 8355 | 16.2% |
+| [QLoRA](https://pytorch.org/torchtune/main/tutorials/qlora_finetune.html) | 5.28 | -42.4% | 8228 | -1.5% |
+
+To reproduce QLoRA:
```
ebsmothers, line 126: I would maybe throw in the cumulative delta here, something like
```suggestion
The final row in the table uses X% less memory than the baseline with a Y% increase in tokens per second. It can be run via the command:
```

README.md
ebsmothers:
```suggestion
torchtune exposes a number of levers for memory efficiency and performance. The table below demonstrates the effects of applying some of these techniques sequentially to the Llama 3B model. Each technique is added on top of the previous one, except for LoRA and QLoRA, which do not use `optimizer_in_bwd` or `AdamW8bit` optimizer.
```

README.md
```diff
@@ -99,6 +99,49 @@ If you are interested in running on different hardware or with different models,
...
+| [+ Compile](https://pytorch.org/tutorials/intermediate/torch_compile_tutorial.html) | 59.33 | -15.9% | 9200 | 27.7% |
+| [+ Chunked Cross Entropy](https://pytorch.org/torchtune/main/generated/torchtune.modules.loss.CEWithChunkedOutputLoss.html) | 48.69 | -17.9% | 9415 | 2.3% |
+| [+ Activation Ceheckpointing](https://pytorch.org/torchtune/main/tutorials/memory_optimizations.html#activation-checkpointing) | 25.28 | -48.1% | 7394 | -21.5% |
+| [+ Optimizer in bwd](https://pytorch.org/torchtune/main/tutorials/memory_optimizations.html#fusing-optimizer-step-into-backward-pass) | 23.92 | -5.4% | 7489 | 1.3% |
+| [+ Activation Offloading](https://pytorch.org/torchtune/main/tutorials/memory_optimizations.html#activation-offloading) | 22.34 | -6.6% | 7479 | -0.1% |
+| [+ Adamw8bit](https://pytorch.org/torchtune/main/tutorials/memory_optimizations.html#lower-precision-optimizers) | 17.57 | -21.4% | 7187 | -3.9% |
+| [LoRA](https://pytorch.org/torchtune/main/tutorials/lora_finetune.html) | 9.16 | -47.8% | 8355 | 16.2% |
+| [QLoRA](https://pytorch.org/torchtune/main/tutorials/qlora_finetune.html) | 5.28 | -42.4% | 8228 | -1.5% |
```
ebsmothers, line 124: Discussed already but make sure to update these links. Also (related to my other comment) the compile link does not point to a place that demonstrates its usage

README.md
felipemello1: If i may push back a bit, i think that less is more. Adding extra comments or columns may confuse the user. When the user open the config, it shouldn't be hard to verify the flags they care about. Let me know if you disagree.

README.md
felipemello1: the link used in compile is a pytorch tutorial. We don't have a link for compile. I could change it our utility function "compile_model". But since we have the example below with the command for QLoRA, and also it is present in every config, i think it should be ok. Do you disagree?

felipemello1: I will update it for LoRA and QLoRA

README.md
```diff
@@ -103,6 +103,49 @@ If you are interested in running on different hardware or with different models,
...
+| [+ Activation Checkpointing](https://pytorch.org/torchtune/main/tutorials/memory_optimizations.html#activation-checkpointing) | 24.9 | -41.93% | 7210 | -21.41% |
+| [+ Fuse optimizer step into backward](https://pytorch.org/torchtune/main/tutorials/memory_optimizations.html#fusing-optimizer-step-into-backward-pass) | 23.1 | -7.29% | 7309 | +1.38% |
+| [+ Activation Offloading](https://pytorch.org/torchtune/main/tutorials/memory_optimizations.html#activation-offloading) | 21.8 | -5.48% | 7301 | -0.11% |
+| [+ 8-bit AdamW](https://pytorch.org/torchtune/main/tutorials/memory_optimizations.html#lower-precision-optimizers) | 17.6 | -19.63% | 6960 | -4.67% |
+| [LoRA](https://pytorch.org/torchtune/main/tutorials/memory_optimizations.html#glossary-lora) | 8.5 | -51.61% | 8210 | +17.96% |
+| [QLoRA](https://pytorch.org/torchtune/main/tutorials/memory_optimizations.html#quantized-low-rank-adaptation-qlora) | 4.6 | -45.71% | 8035 | -2.13% |
+
+The final row in the table vs baseline + Packed Dataset uses 92.3% less memory with a 13.57% increase in tokens per second. It can be run via the command:
```
ebsmothers, line 131: Why not just compare to baseline? Imo that is more compelling. Also you can bold the numbers

README.md
```diff
@@ -103,6 +103,49 @@ If you are interested in running on different hardware or with different models,
 
 &nbsp;
 
+### Optimization flags
+
+torchtune exposes a number of levers for memory efficiency and performance. The table below demonstrates the effects of applying some of these techniques sequentially to the Llama 3B model. Each technique is added on top of the previous one, except for LoRA and QLoRA, which do not use `optimizer_in_bwd` or `AdamW8bit` optimizer.
+
+**Baseline:**
+- **Model:** Llama 3B
```
ebsmothers, line 111: Should change Llama 3B -> Llama 3.2 3B in both locations

README.md
ebsmothers: OK this makes sense to me. Super nit but you can consider putting the overrides in order to more closely match the table (i.e. packed, then compile, then chunked CE, then AC, then opt-in-bwd, then AO)

ebsmothers (commented):
Thanks for adding this!

PR #2006, opened 2024-11-14
Faster intermediate checkpoints with DCP async save in TorchTune

saumishr:
#### Context
- [x] add a new feature
- [ ] fix a bug
- [x] other (please add here)

This diff introduces the DistributedCheckpointing based asynchronous checkpointing and enables it for intermediate checkpoints. 

##### Full Finetune Distributed

Ranks | Model Type | Current | DCP Async
-- | -- | -- | --
1 | Llama 3.2 11B | 33.51s | 7.05s
8 | Llama 3.2 Vision 90B | 454.68s | 22.28s

#### Changelog
* Introduces DistributedCheckpointing based Checkpointer
* Ability to checkpoint asynchronously for intermediate checkpoints.
* Ability to load the latest intermediate checkpoint via DistributedCheckpointing
* Clean up the older checkpoints to save storage.

#### Usage
* Config `enable_async_checkpointing` enables the asynchronous checkpoint saving for intermediate checkpoints.
* Config `resume_from_checkpoint` will enable the resume from the latest intermediate checkpoint. No recipe state needs to be provided as its already saved in the distributed checkpoint.

#### Test plan

#### UX


felipemello1:
I will review it tomorrow. Thanks for the PR!

recipes/full_finetune_distributed.py
```diff
@@ -283,11 +290,61 @@ def setup(self, cfg: DictConfig) -> None:
             optimizer_in_bwd=self._optimizer_in_bwd,
             opt_state_dict=(
                 checkpoint_dict[training.OPT_KEY]
-                if self._resume_from_checkpoint
+                if self._resume_from_checkpoint and not self._enable_async_checkpointing
                 else None
             ),
         )
 
+        if self._resume_from_checkpoint and self._enable_async_checkpointing:
```
pbontrager, line 298: I understand that this allowed for fast verification of the checkpoint time, and the improvements look really great. But I don't think we want the internals of how checkpointing works exposed in the recipe. How can this the user exposure between dcp and standard checkpointer be consolidated to the same thing so that checkpointer logic only has to take a few lines of code in the recipe?

recipes/full_finetune_distributed.py
```diff
@@ -633,11 +744,17 @@ def save_checkpoint(
 
         intermediate_checkpoint = epoch + 1 < self.total_epochs
 
+        # If async checkpointing is enabled, intermediate checkpoints will
+        # be saved asynchronously.
+        if intermediate_checkpoint and self._enable_async_checkpointing:
```
pbontrager, line 749: Is there ever a case where we wouldn't want async checkpointing for intermediate? Is the flag really needed? Also, it should be possible to merge the async save and full save logic into the same function to not proliferate state dict code everywhere.

recipes/full_finetune_distributed.py
```diff
@@ -614,6 +671,60 @@ def _setup_data(
 
         return sampler, dataloader
 
+    def save_checkpoint_async(
```
pbontrager, line 674: Would the idea be to eventually use async checkpoint everywhere and consolidate the files at the end? Or will we always do the all gather at the end?

recipes/full_finetune_distributed.py
saumishr: @pbontrager This flag was added temporarily to make it an opt-in feature which allows us to get more validations before we make it a default. Also, in case of any bugs, devs can turn it off and fallback to the prior state. Currently async save and full save happen via different Checkpointers having different formats but we can probably create a helper method to move this logic out of the recipe code. I will refactor this a bit more.

recipes/full_finetune_distributed.py
saumishr: I agree. Checkpointing in general needs a fair bit of refactor. Plenty of code duplication. I wanted to avoid a major refactor here since I believe @joecummings is planning this out already.

recipes/full_finetune_distributed.py
saumishr: By end, I assume you mean the final checkpoint at the end of the fine-tuning?

Final checkpoint format should be the one that user configures in their job config. Intermediate checkpoints are only for fault tolerance so we can choose the Checkpointer that saves the wasted training time and GPUs the best, which in this case, is DCP (DistributedCheckpointing) with Async save. We will however provide DCP option as another Checkpointer and users can choose to get the final checkpoint in that format as well but we wont enforce it, probably. Use case I can think of is when users train using TorchTitan, where DCP is the default Checkpointer, they can then plug it into TorchTune directly for fine-tuning instead of conversions to torch.save or HF formats.

torchtune/training/checkpointing/_checkpointer.py
```diff
@@ -912,3 +923,182 @@ def save_checkpoint(
...
+        Load a Distributed checkpoint
+        """
+        checkpoint_path = self._get_latest_intermediate_checkpoint_path()
+
+        if checkpoint_path is None:
+            raise ValueError(
+                "No intermediate checkpoint found in the specified directory. Please ensure that a checkpoint exists."
+            )
+
+        if self._is_rank_zero:
```
RdoubleA, line 986: you can use the `torchtune.utils.logging.log_rank_zero` utility so you don't have to keep checking for rank

torchtune/training/checkpointing/_checkpointer.py
```diff
@@ -912,3 +923,182 @@ def save_checkpoint(
...
+    ) -> None:
+        self._checkpoint_dir = Path(checkpoint_dir)
+        self._output_dir = Path(output_dir)
+        self._checkpoint_future = None
+        self._checkpoint_dir_prefix = "checkpoint"
+        self._metadata_file = ".metadata"
+        _, rank = training.get_world_size_and_rank()
+        self._is_rank_zero = rank == 0
+
+    def _get_latest_intermediate_checkpoint_path(self) -> Optional[str]:
```
RdoubleA, line 950: this could be generally useful for all other checkpointers, cc @joecummings to consider this for the redesign

torchtune/training/checkpointing/_checkpointer.py
```diff
@@ -912,3 +923,182 @@ def save_checkpoint(
...
+                    for index in range(epoch):
+                        last_checkpoint = os.path.join(
+                            self._checkpoint_dir,
+                            f"{self._checkpoint_dir_prefix}_{index}",
+                        )
+                        if os.path.isfile(
+                            os.path.join(last_checkpoint, self._metadata_file)
+                        ):
+                            if self._is_rank_zero:
+                                logger.info(
```
RdoubleA, line 1059: do we want to automatically remove the previous checkpoint? some users may want to keep them around so they can evaluate model checkpoints over the course of training

RdoubleA (commented):
I am unclear on if the `DistributedCheckpointer` class will replace the HF or Meta checkpointers or if it should be used in tandem for the fault tolerant intermediate checkpoints. How can users easily convert between checkpoint formats? Would it make sense to integrate the APIs into each checkpointer just for the purposes of async intermediate checkpoints, and then the first load / final save still remain in the original HF / meta format?

Maybe I just need to see the north star for the checkpoint redesign (@joecummings)

torchtune/training/checkpointing/_checkpointer.py
saumishr: Without this, we will be limiting the number of checkpoints we can take as its gonna be bound by the storage available on the host and it can run out quickly for large models. For example, Llama 3.2 vision90B  is almost 150-200GB  in size.
For internal trainers, we keep the last checkpoint only, as for fault tolerance we always restore the latest one. We can possibly make it a config so the users can configure the behavior if they want to evaluate intermediate checkpoints and can also choose to keep only the latest, if the checkpoints are large and taken frequently.

ebsmothers: Yeah agreed with making it configurable. For this PR I would actually lean towards keeping all intermediate checkpoints around. Mainly because that's what we do for our other checkpointers and we shouldn't have diverging behavior across different checkpoint formats. I'd like to see us add a field like `save_last_n_checkpoints` to all checkpointers in one go

recipes/full_finetune_distributed.py
```diff
@@ -614,6 +671,60 @@ def _setup_data(
...
+            else self._optimizer.state_dict()
+        )
+
+        ckpt_dict[training.MODEL_KEY] = model_state_dict
+        ckpt_dict[training.OPT_KEY] = optim_state_dict
+
+        dcp_saver = DistributedCheckpointer(
+            checkpoint_dir=self._checkpointer._checkpoint_dir,
+            output_dir=self._checkpointer._output_dir,
+        )
```
ebsmothers, line 715: Just wanna check my understanding here. We normally would define checkpointer as part of the config to be instantiated but here you basically use `enable_async_checkpointing` as a proxy for whether we're the `DistributedCheckpointer` or not. Is the idea behind this to have the ability to save and load intermediate checkpoints with `DistributedCheckpointer` but still save the final checkpoint in the HF/Meta format?

recipes/full_finetune_distributed.py
```diff
@@ -201,13 +201,19 @@ def load_checkpoint(self, cfg_checkpointer: DictConfig) -> Dict[str, Any]:
         Extract the checkpoint state from file and validate. If resume_from_checkpoint
         is True, this also includes the recipe state.
         """
+        # If async checkpointing is enabled, set the resume from checkpoint as false for the checkpointer.
+        # We will use DistributedCheckpointer instead to load the checkpoint.
+        resume_from_checkpoint: bool = (
+            False if self._enable_async_checkpointing else self._resume_from_checkpoint
+        )
```
ebsmothers, line 208: I'm a bit confused by this.. is it just because of the if statement for `self._update_recipe_state` (i.e. we do not need to explicitly load state when using DCP)? If so I think we should just use a different variable e.g. `should_load_recipe_state = self._resume_from_checkpoint and not self._enable_async_checkpointing` or something to be more explicit.

torchtune/training/checkpointing/_checkpointer.py
```diff
@@ -912,3 +923,182 @@ def save_checkpoint(
...
+                                    f"Cleaning up the last checkpoint {last_checkpoint} to save storage"
+                                )
+                            shutil.rmtree(last_checkpoint)
+                else:
+                    logger.error(
+                        f"Checkpoint failed to save asynchronously to {checkpoint_path} with the exception {f.exception()}"
+                    )
```
ebsmothers, line 1066: I see a bunch of exceptions coming from rmtree when I run this on my machine. E.g.

```
exception calling callback for <Future at 0x7f7415a06a50 state=finished returned Metadata>
  File "/home/ebs/.conda/envs/tt-alt-10-24/lib/python3.11/concurrent/futures/_base.py", line 340, in _invoke_callbacks
    callback(self)
  File "/data/users/ebs/ebs-torchtune-alt/torchtune/training/checkpointing/_checkpointer.py", line 1062, in callback
    shutil.rmtree(last_checkpoint)
  File "/home/ebs/.conda/envs/tt-alt-10-24/lib/python3.11/shutil.py", line 752, in rmtree
    _rmtree_safe_fd(fd, path, onerror)
  File "/home/ebs/.conda/envs/tt-alt-10-24/lib/python3.11/shutil.py", line 703, in _rmtree_safe_fd
    onerror(os.unlink, fullname, sys.exc_info())
  File "/home/ebs/.conda/envs/tt-alt-10-24/lib/python3.11/shutil.py", line 701, in _rmtree_safe_fd
    os.unlink(entry.name, dir_fd=topfd)
FileNotFoundError: [Errno 2] No such file or directory: '__2_46.distcp'
```

(However I am still able to save the checkpoints successfully)

torchtune/training/checkpointing/_checkpointer.py
```diff
@@ -912,3 +923,182 @@ def save_checkpoint(
...
+            if self._is_rank_zero:
+                logger.info(
+                    f"Saving model checkpoint synchronously to {checkpoint_path}."
+                )
+
+            save(
+                state_dict,
+                storage_writer=FileSystemWriter(
+                    checkpoint_path,
+                    thread_count=8,
```
ebsmothers, line 1094: Out of curiosity, how is this value determined?

torchtune/training/checkpointing/_checkpointer.py
```diff
@@ -912,3 +923,182 @@ def save_checkpoint(
...
+            def callback(
+                f: Future,
+            ) -> None:
+                if f.exception() is None:
+                    if self._is_rank_zero:
+                        logger.info(
+                            f"Checkpoint is saved asynchronously to {checkpoint_path}"
+                        )
+
+                    for index in range(epoch):
```
ebsmothers, line 1050: also nit but do you really need a for loop here if this runs every epoch? (i.e. isn't it just sufficient to clean up epoch n-1 checkpoint after each epoch n save)

ebsmothers (commented):
Thank you for this PR, the results look great! I left a bunch of questions (some probably pretty basic) but I don't have any huge concerns here. The main things needed in my mind are:

1) Testing. One thing that we should definitely test is that resuming from an intermediate DCP checkpoint works as expected. I think we can do that by modifying [this recipe test](https://github.com/pytorch/torchtune/blob/ac14e96443e60fd8ffed434637357ce22a9c5bff/tests/recipes/test_full_finetune_distributed.py#L125). And unit tests for the `DistributedCheckpointer` class would also be nice, you can check the unit tests for our other checkpointers in [this file](https://github.com/pytorch/torchtune/blob/ac14e96443e60fd8ffed434637357ce22a9c5bff/tests/torchtune/training/checkpointing/test_checkpointer.py).
2) Figuring out some of the UX points raised by @pbontrager. I think there are some larger points that @joecummings will be figuring out and we probably don't need to block this PR on. But would definitely like to figure out how we can get to a place where this is less checkpointing code in the recipe and more in reusable utilities.

torchtune/training/checkpointing/_checkpointer.py
saumishr: Ohh, my bad. Thanks for catching that. I somehow did not notice this in my testing. Basically the checkpoint deletion should happen at rank_0 only otherwise two different ranks may have a race condition in deleting a file and will run into the error above. Its not a problem during save, since DCP prepares the global plan and distributes it across ranks so every rank knows exactly what needs to be saved and there is no collision there.

torchtune/training/checkpointing/_checkpointer.py
saumishr: This is set as per the available storage I/O bandwidth. This value determines the number of parallel I/O threads while writing the data to storage. For our internal storage like Manifold or even S3, we know the capacity available and tune this value accordingly. This way we use max IO capacity with optimal performance without getting throttled. 

For TorchTune case, since we are writing to Disk, we may have to run some experiments and tune it as per the IOPS available. Ideally, this should also be exposed as a checkpointing_config to the user. They can tune it to get the best performance depending on the underlying storage: HDD, SSD or cloud storage like S3.

I used 16 since that is a default for Manifold integrations internally. For Disks, this number can be higher.  Also, it will only affect the upload speed to storage so its virtually has no effect on the async checkpointing blocking time.

torchtune/training/checkpointing/_checkpointer.py
saumishr: Since you suggested to remove this functionality for now from this PR, this logic will get cleaned up. I will record this issue it in my doc of suggestions for refactor and we can get this added for all Checkpointers in one go.

torchtune/training/checkpointing/_checkpointer.py
saumishr: Yes. Exactly. It seems like we would need to have a set of checkpointing configs for the user to tune the performance and behavior. We should do it in one go across Checkpointers and get these configs documented. I will get it cleaned up from this PR./

torchtune/training/checkpointing/_checkpointer.py
saumishr: True. We can do that. I added this code instead just for edge cases when some prior checkpoint folder still did not get cleaned up due to user cancelling the job or the job errored out while the clean up was happening asynchronously.

recipes/full_finetune_distributed.py
saumishr: Yes. Exactly. I was less inclined to have users configure two different Checkpointers in the same job config: `DistributedCheckpointer` for intermediate and HF/Meta checkpointer for first load and final save. I exposed the async checkpointing as a feature instead under the `enable_async_checkpointing` config for two reasons:
- `DistributedCheckpointer` is the only checkpointer which supports async checkpointing. 
- Users will probably want to control the first load and final save checkpoints. Intermediate checkpoints are for fault tolerance and we have flexibility to enforce the Checkpointer that works best for that.

Basically the end state I was thinking of is the following:
- For first load and final save, users can choose from any of the following Checkpointers: HF, Meta or `DistributedCheckpointer`. No async checkpointing here, since checkpoint loads and final saves will always need to be synchronous. We can not finish the job early until the entire final checkpoint has been persisted. 
- For intermediate checkpoints, we use async save with `DistributedCheckpointer` by default.

recipes/full_finetune_distributed.py
saumishr: Yes. Thats right. I agree. Current way is a bit sloppy to avoid loading the recipe state. I will update the variables to make it a bit more readable.

felipemello1 (commented):
Just left small comments since the code will still be refactored. As I was reading through it, i was thinking if the same utilities will also work for LoRA. I think that its great that you are focusing on a single recipe for now, but just food for thought as you are designing it. Thanks for the PR! :)

torchtune/training/checkpointing/_checkpointer.py
saumishr: @felipemello1 We decided to keep the current behavior consistent with DistributedCheckpointer as well in this PR, which is to keep all the checkpoints persisted for now. I believe when we introduce this feature across Checkpointers, just having `keep_last_n_checkpoints` should be sufficient. 
- None -> All the checkpoints are kept
- 0 -> No intermediate checkpoints
- 1 -> Only the latest valid checkpoint is persisted. (Probably the default)
- N ( N > 1) -> Latest N checkpoints are kept.

torchtune/training/checkpointing/_checkpointer.py
saumishr: Yeah. DCP APIs are trainer agnostic. So really all the code in the integration is gonna be state dict init & updates along with the tests.

recipes/full_finetune_distributed.py
```diff
@@ -283,11 +290,61 @@ def setup(self, cfg: DictConfig) -> None:
...
+
+            state_dict: Dict[str:Any] = {}
+            state_dict.update(
+                {
+                    training.MODEL_KEY: model_state_dict,
+                    training.OPT_KEY: optim_state_dict,
+                    training.SEED_KEY: 0,
+                    training.EPOCHS_KEY: 0,
+                    training.TOTAL_EPOCHS_KEY: 0,
+                    training.MAX_STEPS_KEY: 0,
```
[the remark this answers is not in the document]
saumishr: https://github.com/pytorch/torchtune/pull/2017/ introduces the lr_scheduler. Once both of these PRs land, it should be straightforward to add the checkpointing support. We can do it as a follow up.

recipes/full_finetune_distributed.py
```diff
@@ -614,6 +671,60 @@ def _setup_data(
...
+        """
+
+        if self._is_rank_zero:
+            log.info("Saving checkpoint asynchronously. Retrieving full state dict...")
+            cp_start = time.perf_counter()
+
+        # Create the checkpoint dict to be sent to the checkpointer and ultimately persisted to storage
+        ckpt_dict = {
+            training.SEED_KEY: self.seed,
+            training.EPOCHS_KEY: self.epochs_run,
```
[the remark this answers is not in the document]
saumishr: @felipemello1 Sorry. Could you please explain this comment a bit more? Are you suggesting, we should not be saving the epochs_run in the checkpoint? How would be restore the training progress in terms of the number of epochs that were run till the last checkpoint was taken?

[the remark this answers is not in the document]
felipemello1: hey, sorry, let me clarify: do we need to pass "epoch" as input args AND also track it with self.epochs_run? Are those different? If they are, maybe we can rename to make obvious the difference between those

image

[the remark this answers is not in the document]
saumishr: We can do that. save_checkpoint already takes the `epoch` so I propagated it to the async API as well. It would require a minor change to update the self._epochs_run only after taking the checkpoint for the current epoch here: https://github.com/pytorch/torchtune/blob/main/recipes/full_finetune_distributed.py#L878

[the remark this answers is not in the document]
saumishr: @felipemello1 I looked into it some more and decided to revert my changes back to the prior state. 

- current_epoch refers to the epoch which just got completed and epochs_run gets incremented to the next epoch. 
- current_epoch is used in the checkpoint name however in the training progress state dict, intermediate checkpoints save the next epoch to resume from but final save checkpoints save the epoch which just got completed. 

Therefore passing the current_epoch in the checkpoint methods seems cleaner for now compared to doing +1s and -1s for different type of checkpoints. I do believe this logic should be refactored a bit and all the checkpoints should save the total epochs which have completed. During training run, we adjust the starting epoch depending on if we are resuming from a checkpoint and training from scratch.

recipes/full_finetune_single_device.py
```diff
@@ -197,7 +197,7 @@ def load_checkpoint(self, cfg_checkpointer: DictConfig) -> Dict[str, Any]:
         """
         self._checkpointer = config.instantiate(
             cfg_checkpointer,
-            resume_from_checkpoint=self._resume_from_checkpoint,
+            should_load_recipe_state=self._resume_from_checkpoint,
```
felipemello1, line 200: this will break every config or system that may be using this function. Although i agree its a better name, i dont think it is worth to change it in this PR. A possibly better alternative is to keep both, handle it in the class, and raise a deprecation warning. We should check what others think.

recipes/full_finetune_distributed.py
```diff
@@ -900,7 +1049,9 @@ def recipe_main(cfg: DictConfig) -> None:
             "Distributed finetune recipe should be run via a distributed launcher."
             "If using tune CLI, please specify --nnodes 1 and --nproc_per_node [num_gpus]"
         )
-    init_process_group(backend="gloo" if cfg.device == "cpu" else "nccl")
+
+    init_process_group(backend="gloo" if cfg.device == "cpu" else "cpu:gloo,cuda:nccl")
```
felipemello1, line 1053: n00b question: should it just be:

```init_process_group(backend="cpu:gloo,cuda:nccl")```

File: recipes/full_finetune_distributed.py [hunk 16]
```diff
@@ -199,21 +204,101 @@ def __init__(self, cfg: DictConfig) -> None:
...
+
         self._checkpointer = config.instantiate(
             cfg_checkpointer,
-            resume_from_checkpoint=self._resume_from_checkpoint,
+            should_load_recipe_state=should_load_recipe_state,
         )
-        checkpoint_dict = self._checkpointer.load_checkpoint()
 
-        if self._resume_from_checkpoint:
+        return self._checkpointer.load_checkpoint()
```
felipemello1, line 235: These two functions are doing the same, but their signature is very different. Async also depends on the other, instead of config args.

for dcp, we are doing
```
dcp_checkpointer = self.get_dcp_checkpointer()
state_dict = dcp_checkpointer.load_checkpoint(state_dict)
```

for sync ckpt, we are doing in setup:
```
# Load the base model
checkpoint_dict = self.load_checkpoint(cfg_checkpointer=cfg.checkpointer)
```

I understand that you want to avoid a major refactoring. However, in my opinion, these functions should just be:

```
def setup_async_ckpt(self, cfg_checkpointer):
	return DistributedCheckpointer(
                checkpoint_dir=cfg_checkpointer.checkpoint_dir,
                output_dir=cfg_checkpointer.output_dir,
            )
	
def setup_sync_ckpt(self, cfg_checkpointer):
	return config.instantiate(
            cfg_checkpointer,
            resume_from_checkpoint=self._resume_from_checkpoint,
            should_load_recipe_state=should_load_recipe_state,
        )
```

and call:
```
self.dcp_checkpointer = setup_async_ckpt(cfg_checkpointer)
state_dict = self.dcp_checkpointer .load_checkpoint(state_dict)
```

recipes/full_finetune_distributed.py
```diff
@@ -199,21 +204,101 @@ def __init__(self, cfg: DictConfig) -> None:
...
+                    training.MODEL_KEY: model_state_dict,
+                    training.OPT_KEY: optim_state_dict,
+                    training.SEED_KEY: 0,
+                    training.EPOCHS_KEY: 0,
+                    training.TOTAL_EPOCHS_KEY: 0,
+                    training.MAX_STEPS_KEY: 0,
+                }
+            )
+
+            dcp_checkpointer = self.get_dcp_checkpointer()
```
felipemello1, line 268: i would move this to def setup

recipes/full_finetune_distributed.py
```diff
@@ -618,6 +707,56 @@ def _setup_data(
 
         return sampler, dataloader
 
+    def save_checkpoint_async(
```
felipemello1, line 710: can we make it an utility?

```
save_checkpoint_async(
	epoch = epoch,
	model_state_dict = self._model.state_dict(),
	optim_state_dict = self._optim_ckpt_wrapper.state_dict() if self._optimizer_in_bwd else self._optimizer.state_dict(),
	checkpointer = self.dcp_checkpointer
	is_rank_zero = self._is_rank_zero
)
```

torchtune/training/checkpointing/_checkpointer.py
```diff
@@ -932,3 +943,187 @@ def save_checkpoint(
...
+        else:
+            log_rank_zero(
+                logger,
+                msg=f"Saving model checkpoint synchronously to {checkpoint_path}.",
+            )
+
+            save(
+                state_dict=state_dict,
+                storage_writer=FileSystemWriter(
+                    checkpoint_path,
```
felipemello1, line 1117: can we make this API easy to save to manifold internally, so we can avoid rewriting this code? Or thats not how it works?

recipes/full_finetune_distributed.py
```diff
@@ -199,21 +204,101 @@ def __init__(self, cfg: DictConfig) -> None:
...
             cfg_checkpointer,
-            resume_from_checkpoint=self._resume_from_checkpoint,
+            should_load_recipe_state=should_load_recipe_state,
         )
-        checkpoint_dict = self._checkpointer.load_checkpoint()
 
-        if self._resume_from_checkpoint:
+        return self._checkpointer.load_checkpoint()
+
+    def resume_from_checkpoint(self, checkpoint_dict: Dict[str, any]) -> Dict[str, Any]:
```
felipemello1, line 237: Personally, i would prefer if we broke it down into:

```
def resume_from_checkpoint_async
	...

def resume_from_checkpoint_sync
	...
```

Not a big fan of:
```
if:
	100 lines
else:
	...
```

Do you think that it would make the logic more discrete and easier to maintain?

recipes/full_finetune_distributed.py
```diff
@@ -637,11 +776,17 @@ def save_checkpoint(
 
         intermediate_checkpoint = epoch + 1 < self.total_epochs
 
+        # If async checkpointing is enabled, intermediate checkpoints will
+        # be saved asynchronously.
+        if intermediate_checkpoint and self._enable_async_checkpointing:
+            self.save_checkpoint_async(epoch)
```
felipemello1, line 782: i like that we have a function specialized for async. This would get a bit into refactoring, but it would be great to take the other part of the code and turn it into self.save_checkpoint_sync.

felipemello1 (commented):
Overall perception:

1) This would be easier to follow if every async was marked as async (and everything sync was marked as sync, optionally). E.g. save_ckpt_async // load_ckpt_sync // async_ckpt // sync_ckpt.

2) The recipe is the main point of contact between user/torchtune. IMO, ckpt functions are taking too much space and turning them into utilities will help a lot. I also think that the mental model of what/when/why we do async/sync needs to be more explicit. As it is know, i think that an outsider will have a hard time following. Small comments may help.

3) We need some utility to go from dcp -> tune ckpt. Users do use intermediate ckpt for eval. Is it possible? Can we have a performant `assert dcp_ckpt == tune_to_dcp(dcp_to_tune(dcp_ckpt))` ?

4) i am afraid of not using dcp for the last ckpt. We have a case of 90B ckpt erroring with NCCL time out when saving using 64 gpus. I imagine that dcp would avoid it. Any thoughts here?

felipemello1 (commented):
edit: duplicated. Submitted twice.

felipemello1 (commented):
edit: duplicated. Submitted twice.

recipes/full_finetune_distributed.py
saumishr: New commit refactors the checkpointing code into its own client and cleans up the checkpointing code from the recipes. I believe this comment should already be addressed. Post refactor, we have distinct methods for sync and async saves. PTAL.

torchtune/training/checkpointing/_checkpointer.py
saumishr: Yes. Lets do that in a follow up PR. Its as simple as plumbing in in FSSpec with Manifold based storage components which already exists. We will need to set up a way to plug in internal components in TorchTune, like we do in TorchTNT etc.

recipes/full_finetune_distributed.py
saumishr: This may not be relevant with the latest refactor. CheckpointClient will be the entry-point for the checkpointing APIs across recipes. This refactor is in line with how we do it with other trainers.

recipes/full_finetune_distributed.py
saumishr: Addressed in the latest refactor. PTAL.

recipes/full_finetune_distributed.py
saumishr: The selection of correct Checkpointer has been abstracted away into a CheckpointClient which will encapsulate checkpointing logic for all recipes. This reduces the checkpointing code footprint in the recipes and will make it much easier to introduce async checkpoints for other recipes.

recipes/full_finetune_single_device.py
saumishr: If users have custom recipes with the resume_from_checkpoint, then it may be better to deprecate the resume_from_checkpoint and introduce should_load_recipe_state, otherwise it will be a backward incompatible change. However if thats not the case, then it seems okay to me to update all of our recipes which is a low risk since there is no functionality change. 

cc @ebsmothers who had this comment in the earlier version of the PR.

recipes/full_finetune_distributed.py
```diff
@@ -287,11 +373,14 @@ def setup(self, cfg: DictConfig) -> None:
             optimizer_in_bwd=self._optimizer_in_bwd,
             opt_state_dict=(
                 checkpoint_dict[training.OPT_KEY]
-                if self._resume_from_checkpoint
+                if self._resume_from_checkpoint and not self._enable_async_checkpointing
```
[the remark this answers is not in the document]
saumishr: @felipemello1 
1) Latest version with the checkpointing code refactor addresses it. PTAL.

2) Yes. There is an integration test as well added in the PR which tests this scenario with `optimizer_in_bwd=True` and `enable_async_checkpointing=True`

recipes/full_finetune_distributed.py
```diff
@@ -199,21 +204,101 @@ def __init__(self, cfg: DictConfig) -> None:
...
+            )
+        else:
             self._update_recipe_state(checkpoint_dict)
-        return checkpoint_dict
+            log_rank_zero(
+                log,
+                msg=f"Loaded intermediate checkpoint for the epoch: {checkpoint_dict[training.EPOCHS_KEY]}",
+            )
 
     def _update_recipe_state(self, ckpt_dict: Dict[str, Any]) -> None:
```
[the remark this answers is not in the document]
saumishr: I thought about it. Challenge is that it would be a slightly bigger refactor. The recipe keys are defined in the training module which causes a circular dependency. We would need to refactor that out of training first and then create an util for this method. I will leave it as it is for now since its an existing method in the recipe.

joecummings (approved):
This looks good and I am stamping to unblock; however, b/c I am risk-averse I would like to do a proper deprecation of `resume_from_checkpoint` param in the checkpointer, which just means we allow user to pass in `resume_from_checkpoint` OR `should_load_recipe_state` for one release cycle. If they pass in via `resume_from_checkpoint`, we set it to `should_load_recipe_state` and raise a FutureWarning stating that this will be removed in further releases. 

Thanks so much @saumishr !!!

felipemello1:
@saumishr, there were considerable changed made to the output path of the files. Can you double check that we can still resume from checkpoint even if we are using dcp? https://github.com/pytorch/torchtune/pull/2074

Also, please check the folder structure. The assumption is that the use can do `cp path_with_ckpt path_with_cfgs`, and this is ready to upload o HF. But if there are 50 files there related to dcp, this will look a bit ugly.

I think that we also need to have a clear solution about going from intermediate ckpt -> hf_format. Like we discussed, many of our users run eval on the intermediate ckpts.

cc: @joecummings

joecummings:
@saumishr Does it matter if the Facebook Internal tests are failing?

saumishr:
@joecummings There are no tests failing. A CI/CD job is complaining that the diff needs to be updated with the latest. I'm validating the async checkpoints with the recent refactor @felipemello1 had done. I will notify here once the validation is done and the PR can be merged. Thank you!

joecummings:
> @joecummings There are no tests failing. A CI/CD job is complaining that the diff needs to be updated with the latest. I'm validating the async checkpoints with the recent refactor @felipemello1 had done. I will notify here once the validation is done and the PR can be merged. Thank you!

Once this validation is confirmed, I'm happy to merge this!

boringlee24:
Is this functionality only compatible with full finetuning, not with LoRA?

ankitageorge:
@boringlee24 the functionality for LoRa was recently added as well, so it should work now.

PR #1933, opened 2024-10-31
Remove unused FSDP1 components

krammnic:
#### Context
What is the purpose of this PR? Is it to
- [ ] add a new feature
- [ ] fix a bug
- [ ] update tests and/or documentation
- [X] Clean up

Please link to any issues this PR addresses.

#### Changelog
What are the changes made in this PR?
* #1897

#### Test plan
Please make sure to do each of the following if applicable to your PR. If you're unsure about any one of these just ask and we will happily help. We also have a [contributing page](https://github.com/pytorch/torchtune/blob/main/CONTRIBUTING.md) for some guidance on contributing.

- [X] run pre-commit hooks and linters (make sure you've first installed via `pre-commit install`)
- [ ] add [unit tests](https://github.com/pytorch/torchtune/tree/main/tests/torchtune) for any new functionality
- [X] update [docstrings](https://github.com/pytorch/torchtune/tree/main/docs/source) for any new or updated methods or classes
- [X] run unit tests via `pytest tests`
- [ ] run recipe tests via `pytest tests -m integration_test`
- [ ] manually run any new or modified recipes with sufficient proof of correctness
- [ ] include relevant commands and any other artifacts in this summary (pastes of loss curves, eval results, etc.)

#### UX
If your function changed a public API, please add a dummy example of what the user experience will look like when calling it.
Here is a [docstring example](https://github.com/pytorch/torchtune/blob/6a7951f1cdd0b56a9746ef5935106989415f50e3/torchtune/modules/vision_transformer.py#L285)
and a [tutorial example](https://pytorch.org/torchtune/main/tutorials/qat_finetune.html#applying-qat-to-llama3-models)

- [X] I did not change any public API
- [ ] I have added an example to docs or docstrings

PR #2016, opened 2024-11-16
Remove unused FSDP components

ebsmothers:
I was attempting to patch the changes from #1933 so we can merge them, but I'm unable to push to the remote branch (likely because the PR was opened off of the fork's main branch which has not yet been merged with latest changes from upstream). So I've just patched everything into this new PR instead.

Tagging @krammnic as the author of the original PR. The only changes I've made were merging latest changes from main and updating `test_validate_missing_and_unexpected_for_lora` to get it to pass (and adding some of the test cases from `validate_state_dict_for_lora` that were deleted)


SalmanMohammadi:
One small thing - is it worth deprecating `validate_state_dict_for_lora` and deleting it next release, since it's a public API? For the hackers.

SalmanMohammadi (approved):

docs/source/api_ref_modules.rst
```diff
@@ -77,7 +77,6 @@ PEFT Components
     peft.set_trainable_params
     peft.get_adapter_state_dict
     peft.validate_missing_and_unexpected_for_lora
-    peft.validate_state_dict_for_lora
```
SalmanMohammadi, line 80: Line 209 in the LoRA finetune tutorial:
> .. note::
    Whenever loading weights with :code:`strict=False`, you should verify that any missing or extra keys in
    the loaded :code:`state_dict` are as expected. torchtune's LoRA recipes do this by default via e.g.
    :func:`validate_state_dict_for_lora() <torchtune.modules.peft.validate_state_dict_for_lora>` or
    :func:`validate_missing_and_unexpected_for_lora() <torchtune.modules.peft.validate_missing_and_unexpected_for_lora>`.

Needs to be updated

docs/source/api_ref_training.rst
```diff
@@ -50,12 +50,9 @@ Utilities for enabling and working with distributed training.
     :toctree: generated/
     :nosignatures:
 
-    FSDPPolicyType
     init_distributed
     is_distributed
     get_world_size_and_rank
-    get_full_finetune_fsdp_wrap_policy
```
SalmanMohammadi, line 57: There's a reference to this in the [QAT tutorial](https://github.com/pytorch/torchtune/blob/ac14e96443e60fd8ffed434637357ce22a9c5bff/docs/source/tutorials/qat_finetune.rst?plain=1#L173) too.

ebsmothers:
> One small thing - is it worth deprecating the public APIs and deleting next release? I'm not sure if people are really using these so your call

Yeah this is a fair question. In this case I am gonna yolo a bit because we haven't supported FSDP1 for multiple releases at this point. And I think most of these APIs should not really have been public to begin with, but we made them so by necessity because many were used directly in our recipes

PR #1954, opened 2024-11-05
update configs

felipemello1:
#### Context
What is the purpose of this PR? Is it to
- [ ] add a new feature
- [ ] fix a bug
- [ ] update tests and/or documentation
- [x] other (please add here)

#### Changelog
- fix bad defaults
- add missing parameters
- add better comments

```
import os


def modify_yaml_file(file_path):
    updated = {
        "updated_compile": False,
        "updated_packed": False,
        "added_compile": False,
        "added_activation_offloading": False,
        "added_packed": False,
        "added_profiler": False,
        "updated_gradient_accumulation_steps": False,
        "updated_checkpointing_comment": False,
        "updated_gradient_comment": False,
        "updated_compile_comment": False,
        "updated_packed_comment": False,
    }

    with open(file_path, "r") as file:
        lines = file.readlines()
    # Step 2: Remove duplicate 'compile' entries
    compile_indices = [
        i for i, line in enumerate(lines) if line.strip().startswith("compile:")
    ]
    if len(compile_indices) > 1:
        for index in sorted(compile_indices, reverse=True):
            del lines[index]
        updated["updated_compile"] = True
    # Step 3: Move 'packed' after '_component_' and align indentation
    for i, line in enumerate(lines):
        if (
            line.strip().startswith("packed:")
            and i + 1 < len(lines)
            and "_component_" in lines[i + 1]
        ):
            packed_line = lines.pop(i)
            lines.insert(i + 1, packed_line)  # Insert after the _component_ line
            updated["updated_packed"] = True
            break
    # Step 4: Add 'compile' if missing
    if not any(line.strip().startswith("compile:") for line in lines):
        for i, line in enumerate(lines):
            if line.strip().startswith("max_steps_per_epoch:"):
                indentation = len(line) - len(line.lstrip())
                new_line = (
                    " " * indentation
                    + "compile: False # pytorch compile, set to true for better perf/memory\n"
                )
                lines.insert(i + 1, new_line)
                updated["added_compile"] = True
                break
    # Step 5: Add 'enable_activation_offloading' if missing
    if (
        not any(
            line.strip().startswith("enable_activation_offloading:") for line in lines
        )
        and "vision" not in file_path
        and "ppo" not in file_path
        and "dpo" not in file_path
        and "distillation" not in file_path
    ):
        for i, line in enumerate(lines):
            if line.strip().startswith("enable_activation_checkpointing:"):
                indentation = len(line) - len(line.lstrip())
                new_line = (
                    " " * indentation
[...]

PR #2017, opened 2024-11-17
Add LR Scheduler to full finetune distributed

parthsarthi03:
#### Context
- [x] add a new feature

Please link to any issues this PR addresses: #1308 

**Purpose of this PR:**

This PR adds support for an **optional learning rate scheduler** to the `FullFinetuneRecipeDistributed` class, allowing users to configure and use a learning rate scheduler during training.

You can enable it by adding the following to your config file:

```yaml
lr_scheduler:
  num_warmup_steps: 50
```

#### Changelog

- **Implemented Optional Learning Rate Scheduler:**
  - Added a `_setup_lr_scheduler` method to initialize the scheduler based on the configuration.
  - Modified the `setup` method to call `_setup_lr_scheduler` after computing `self._steps_per_epoch` and `self.global_step`.
  - Updated the training loop in the `train` method to step the scheduler after each optimizer step.

#### Test plan

Tested on **4 GPUs** with various configurations: https://wandb.ai/psarthi/torchtune_lr_scheduler_tests:

1. No Learning Rate Scheduler, No Optimizer-in-Backward: https://wandb.ai/psarthi/torchtune_lr_scheduler_tests/runs/e1ddni13
2. No Learning Rate Scheduler, With Optimizer-in-Backward:
https://wandb.ai/psarthi/torchtune_lr_scheduler_tests/runs/km2jw6rs
3. Cosine Learning Rate Scheduler with 50 Warmup Steps, With Optimizer-in-Backward:
https://wandb.ai/psarthi/torchtune_lr_scheduler_tests/runs/lfacg1b8
4. Cosine Learning Rate Scheduler with 50 Warmup Steps, Without Optimizer-in-Backward:
https://wandb.ai/psarthi/torchtune_lr_scheduler_tests/runs/ymktfbam
5. Resuming Training with Learning Rate Scheduler, Without Optimizer-in-Backward:
https://wandb.ai/psarthi/torchtune_lr_scheduler_tests/runs/ckia4yzi

### DIFF
--- a/recipes/full_finetune_distributed.py
+++ b/recipes/full_finetune_distributed.py
@@ -258,7 +258,7 @@ class FullFinetuneRecipeDistributed(FTRecipeInterface):
     def setup(self, cfg: DictConfig) -> None:
         """
         Setup the recipe. This includes training state (if resume_from_checkpoint is True),
-        model, tokenizer, loss, optimizer, sampler, and dataloader.
+        model, tokenizer, loss, optimizer, lr scheduler, sampler, and dataloader.
         """
         if self._is_rank_zero:
             self._metric_logger = config.instantiate(cfg.metric_logger)
@@ -332,6 +332,13 @@ class FullFinetuneRecipeDistributed(FTRecipeInterface):
             self._steps_per_epoch = self.max_steps_per_epoch
         self.global_step = self.epochs_run * self._steps_per_epoch
 
+        # Setup lr scheduler
+        self._lr_scheduler = self._setup_lr_scheduler(
+            cfg_lr_scheduler=cfg.get("lr_scheduler", None),
+            num_training_steps=self.total_epochs * self._steps_per_epoch,
+            last_epoch=self.global_step - 1,
+        )
+
         # Set up profiler, returns DummyProfiler (nullcontext object with no-op `step` method)
         # if cfg is missing profiler key or if `cfg.profiler.enabled = False`
         self._profiler = self._setup_profiler(cfg.get(PROFILER_KEY, None))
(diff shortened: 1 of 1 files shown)


felipemello1:
thanks for the PR! I glanced over it and it looks good. I will review it more carefully tomorrow and merge it i dont find any issues :)

gordicaleksa:
Consider refactoring (extracting into a separate file) because this same setup function is used in `full_finetune_single_device.py` (https://github.com/pytorch/torchtune/blob/main/recipes/full_finetune_single_device.py#L496)

Eventually they'll fall out of sync.

cc: @felipemello1

gordicaleksa:
Also might be worthwhile adding something like:
```
lr_scheduler:
  num_warmup_steps: 10
```

to configs, e.g. for llama 3.1 (8b/70b) better than not being sure what scheduler is being used

felipemello1 (approved):
LGTM! Thanks for doing all of these tests! As a followup, we would have to update the configs. I have a [script](https://github.com/pytorch/torchtune/pull/1954) i used for another PR to bulk update. If you want to do it, let me know, otherwise i can. In the script you would have to filter all cfgs that have "full.yaml" in their name, and find the best spot to add these lines, which is probably right after the tokenizer.

image

felipemello1:
@gordicaleksa , great point! We are currently having some internal discussions about what should be exposed in the recipe and what should be a utility. In general, we are ok with repeating code so it is easy for people to hack and make their changes. But there are use cases like this one that seems to be pretty standard and really don't add much value by being exposed. We will work on making our recipes a bit learner soon.

PR #2020, opened 2024-11-18
Some KD recipe cleanup

ebsmothers:
A few small thing in our KD recipe, mostly pointed out by @joecummings. 

- Packed was not enabled for the single-device recipe. This is an easy fix: we just need to add the packed collate function
- We weren't compiling the chunked loss correctly. Similar to chunked cross-entropy, we shouldn't compile the chunking operation. Luckily the chunked version of the loss already has `self.fkl_loss` pointing to vanilla ForwardKLLoss, so we can just compile that.
- Our memory stats logging on model init is not super clear for KD where there are two separate models. Addressed this by adding an optional message to `log_memory_stats` and explicitly logging peak memory after student and teacher model init separately.

### Compile
On main (baseline) vs this branch: 
```
tune run knowledge_distillation_single_device --config llama3_2/knowledge_distillation_single_device \
max_steps_per_epoch=50 compile=True metric_logger=torchtune.training.metric_logging.WandBLogger \
metric_logger.project=test-kd metric_logger.name=kd-single-device-compile-{baseline}
```

Results show increased tokens/sec and lower peak memory.

Screenshot 2024-11-18 at 6 44 11 AM

### Packed

Previously the following command didn't work in our single-device recipe.

```
tune run knowledge_distillation_single_device --config llama3_2/knowledge_distillation_single_device dataset.packed=True dataset.split=train[:5%] tokenizer.max_seq_len=2048 metric_logger=torchtune.training.metric_logging.WandBLogger metric_logger.project=test-kd metric_logger.name=kd-single-device-packed
```

Screenshot 2024-11-18 at 6 45 19 AM

### Logging:
Running any of the above single-device commands gives the following in the logs:
```
Memory stats after student model init:
        GPU peak memory allocation: 2.41 GiB
        GPU peak memory reserved: 2.42 GiB
        GPU peak memory active: 2.41 GiB
Teacher model is initialized with precision torch.bfloat16.
Memory stats after teacher model init:
        GPU peak memory allocation: 17.43 GiB
        GPU peak memory reserved: 17.56 GiB
        GPU peak memory active: 17.43 GiB
```

### DIFF
--- a/recipes/knowledge_distillation_distributed.py
+++ b/recipes/knowledge_distillation_distributed.py
@@ -514,10 +514,12 @@ class KDRecipeDistributed(FTRecipeInterface):
 
         if self._is_rank_zero:
             log.info(
-                f"Instantiating model and loading checkpoint took {time.perf_counter() - init_start:.2f} secs"
+                f"Instantiating student model and loading checkpoint took {time.perf_counter() - init_start:.2f} secs"
             )
             memory_stats = training.get_memory_stats(device=self._device)
-            training.log_memory_stats(memory_stats)
+            training.log_memory_stats(
+                memory_stats, message="Memory stats after student model init:"
+            )
 
         # synchronize before training begins
         torch.distributed.barrier()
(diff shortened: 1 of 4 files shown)


joecummings (approved):
bueno

felipemello1 (approved):

PR #2023, opened 2024-11-18
Support for unsharded parameters in state_dict APIs

RdoubleA:
#### Context
- [ ] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

Credit goes to @ifed-uscd for the original commit, I've adapted it a bit to our latest APIs. Original summary:

"This diff adds functionality to shard the model separately from the vocabulary pruning, which allows us to run training keeping the model in bf16 and the vocab pruning in fp32"

#### Changelog
*

#### Test plan

#### UX

### DIFF
--- a/torchtune/training/_distributed.py
+++ b/torchtune/training/_distributed.py
@@ -207,6 +207,9 @@ def load_from_full_model_state_dict(
                 requires_grad=sharded_meta_param.requires_grad,
             )
 
+        elif not hasattr(sharded_meta_param, "device_mesh"):
+            # In cases where parts of the model aren't sharded, some parameters will be plain tensors
+            sharded_tensor = full_tensor
         else:
             sharded_tensor = distribute_tensor(
                 full_tensor,
@@ -242,7 +245,9 @@ def gather_cpu_state_dict(
         if sharded_param.is_cpu:
             # Move back to device if offloaded to CPU
             sharded_param = sharded_param.to(device)
-        if isinstance(sharded_param._local_tensor, NF4Tensor):
+        if hasattr(sharded_param, "_local_tensor") and isinstance(
+            sharded_param._local_tensor, NF4Tensor
+        ):
             # NF4Tensor does not support all_gather from DTensor
             # so we need to manually all_gather
             mesh = sharded_param.device_mesh


torchtune/training/_distributed.py
```diff
@@ -242,7 +245,9 @@ def gather_cpu_state_dict(
         if sharded_param.is_cpu:
             # Move back to device if offloaded to CPU
             sharded_param = sharded_param.to(device)
-        if isinstance(sharded_param._local_tensor, NF4Tensor):
+        if hasattr(sharded_param, "_local_tensor") and isinstance(
```
pbontrager, line 248: If the tensor isn't sharded but it still a NF4Tensor, we still need to upcast the datatype as is done on line 271

torchtune/training/_distributed.py
```diff
@@ -242,7 +245,9 @@ def gather_cpu_state_dict(
...
+            sharded_param._local_tensor, NF4Tensor
+        ):
```
pbontrager, line 250: For line 274 (can't comment on that line), what happens if full_tensor() is called on an unsharded parameter? It doesn't crash?

pbontrager: This is an edge case, but what if the unsharded param is not on rank0? Will it get moved to cpu properly? What if it's not even on the same node?

pbontrager (commented):
Added some questions around edge cases.

torchtune/training/_distributed.py
RdoubleA: so maybe another `elif` in the case that the tensor is unsharded, but still NF4Tensor, and just need to call line 271?

torchtune/training/_distributed.py
RdoubleA: oh shoot, forgot to copy that logic over from the internal diff

torchtune/training/_distributed.py
pbontrager: I think this would be much cleaner if at the top of the loop we do:
```
if hasattr(...):
	# get full tensor (NF4 or .full_tensor)
if full_param is NF4:
	# upcast
if is_rank_zero:
	# the rest the same
```
This code still seems to assume that the unsharded param is on rank0 which isn't guaranteed

SLR722 (approved):
Looks good to me

Issue #2025, opened 2024-11-19
Integrate Lora fine-tuned model with HF


GLorenzo679:
Hello, I'm having some issues when using a Lora fine-tuned model with HuggingFace `from_pretrained()`.
I saw some discussions in the issues, and this one #933 caught my attention.
It's suggested to do this in order to load the fine-tuned model: 

``` python
from transformers import AutoModelForCausalLM
from peft import PeftModel

# hub ID of the base model from the above fine-tune
model_id = "meta-llama/Llama-2-7b-hf" 

# output_dir from tune command
checkpoint_dir = "/my/output/dir" 

model = AutoModelForCausalLM.from_pretrained(model_id)
peft_model = PeftModel.from_pretrained(model, checkpoint_dir)
```
Isn't it possible to load the model (`AutoModelForCausalLM.from_pretrained(model_id)`) from a previously downloaded model (the one that you get with tune download), instead of having to download it again from HF?

If I try to load the previously downloaded model I get this error:

```
venv/lib/python3.12/site-packages/transformers/models/auto/auto_factory.py", line 514, in from_pretrained
    pretrained_model_name_or_path = adapter_config["base_model_name_or_path"]
                                    ~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^
KeyError: 'base_model_name_or_path' 
```
Looking at the code that generates the error, it seems like the transformers library is checking if peft the package is installed. If it is so, it looks for the adapter_config.json file (in my case the one generated with tune run) which is missing the `base_model_name_or_path` field.

GLorenzo679:
Thanks for the fast reply.
Your fix solved the problem. I got another error though.

```
venv/lib/python3.12/site-packages/transformers/modeling_utils.py", line 555, in load_state_dict
    with safe_open(checkpoint_file, framework="pt") as f:
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
FileNotFoundError: No such file or directory: "models/codeLlama-7B/model-00001-of-00002.safetensors"
```

I solved this by deleting the `model.safetensors.index.json` file in the model folder.

GLorenzo679:
I have an update on this issue.
It seems that `AutoModelForCausalLM.from_pretrained` is loading the base model (the safetensors files) and not the base+adapter model generated with `torch run`.
Is there a way to use a fine-tuned model generated with `torch run` in huggingface?
[...]

PR #2026, opened 2024-11-19
[Test] Save local model path in PEFT adapter config

ebsmothers:
It was pointed out in #2025 that the one-liner `AutoModelForCausalLM.from_pretrained(model_id)` doesn't work for our models trained with LoRA (instead you need to call `AutoModelForCausalLM.from_pretrained` on a hub model followed by `PeftModel.from_pretrained` on the LoRA-trained model). This is a demo of saving the checkpointer's directory as part of our adapter_config.json file which should enable this to work. 

Edit: after discussing with @pbontrager this won't integrate well with e.g. push to hub since it assumes a local path. So I'm not sure we want to actually land this as is, the better solution may be to make some changes to tune download.

### DIFF
--- a/recipes/lora_finetune_single_device.py
+++ b/recipes/lora_finetune_single_device.py
@@ -619,6 +619,7 @@ class LoRAFinetuneRecipeSingleDevice(FTRecipeInterface):
                 self._apply_lora_to_output,
             ),
             "peft_type": "LORA",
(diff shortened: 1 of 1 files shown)


ebsmothers:
Closing since this is handled (much more nicely) in #2074

PR #2028, opened 2024-11-19
Fix Qlora/lora for 3.2 vision

felipemello1:
#### Context
- [ ] add a new feature
- [x] fix a bug
- [ ] other (please add here)

In [this PR](https://github.com/pytorch/torchtune/pull/1973/files#diff-2fe44ae7ad7c0c93e7b4fedf94285630b96e1cc5907d3c5db01161077fa7dd9fL192) we removed apply_lora_to_output from vision encoder, but forgot to remove it from some places. This raises the error:
```
  File "/data/users/felipemello/torchtune/torchtune/models/llama3_2_vision/_model_builders.py", line 170, in lora_llama3_2_vision_11b
    encoder = lora_llama3_2_vision_encoder(
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/data/users/felipemello/torchtune/torchtune/models/llama3_2_vision/_component_builders.py", line 439, in lora_llama3_2_vision_encoder
    clip = lora_clip_vision_encoder(**clip_options, **lora_options)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/data/users/felipemello/torchtune/torchtune/models/clip/_component_builders.py", line 289, in lora_clip_vision_encoder
    self_attn = lora_clip_attention(
                ^^^^^^^^^^^^^^^^^^^^
  File "/data/users/felipemello/torchtune/torchtune/models/clip/_component_builders.py", line 437, in lora_clip_attention
    adapter_cls(
  File "/data/users/felipemello/torchtune/torchtune/modules/peft/lora.py", line 78, in __init__
    else to_nf4(linear.weight, **quantization_kwargs)
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: to_nf4() got an unexpected keyword argument 'apply_lora_to_output'
```

Additionally, in 4.0 we added a check for quantize_kwargs and quantize_base. This raises the error:
image

#### Test plan
image


torchtune/models/llama3_2_vision/_component_builders.py
```diff
@@ -679,7 +675,6 @@ def lora_llama3_2_vision_projection_head(
     num_hidden_inputs: int,
     # LoRA args
     apply_lora_to_mlp: bool,
-    apply_lora_to_output: bool,
```
pbontrager, line 682: I think this should stay in the projection head, it just doesn't make sense in the CLIP model

torchtune/models/llama3_2_vision/_component_builders.py
```diff
@@ -412,7 +409,6 @@ def lora_llama3_2_vision_encoder(
     lora_options = {
         "lora_modules": lora_attn_modules,
         "apply_lora_to_mlp": apply_lora_to_mlp,
-        "apply_lora_to_output": apply_lora_to_output,
```
pbontrager, line 415: This should be removed from here and manually added to the projection head call

pbontrager (commented):
Thanks for fixing this. I think "apply_lora_to_output" still makes sense for the encoder, just not the CLIP model itself but still in the projection head.

torchtune/modules/peft/lora.py
```diff
@@ -65,7 +65,7 @@ def __init__(
         self.use_bias = use_bias
         self._quantize_base = quantize_base
 
-        if not self._quantize_base and quantization_kwargs:
+        if not self._quantize_base and any([v for v in quantization_kwargs.values()]):
```
RdoubleA, line 68: did `and quantization_kwargs` not work for a dictionary? you could also just check for length

RdoubleA:
```
In [1]: any([None])
Out[1]: False
```

RdoubleA: sorry `{use_lora_on_output: None}` should fail the check?

felipemello1: If quantize_base is false, then we should NOT have any quantization args, right? If no quantization is happenning, then why have these args?

So this assertion was failing, because quantize_base WAS false, so not False --> True, but we have {use_lora_on_output: None}, which returns True.

So if True and True: raise error.

pbontrager: I agree with Rafi here, there shouldn't be any kwargs passed in if quantize_base is False. {use_lora_on_output: None} should not be passed in when quantize_base = False.

felipemello1: image

pbontrager (approved):
Thank you for fixing this

PR #2030, opened 2024-11-20
log rank zero everywhere

RdoubleA:
#### Context
Are you tired of constantly checking for rank zero before logging? Then `log_rank_zero` is just the solution for you!

#### Changelog
* Remove pesky if statements checking for rank zero before logging and instead use `log_rank_zero`
* add `stacklevel=2` to `log_rank_zero` so the line from the calling function is shown instead of within `log_rank_zero`. This is not yet relevant, until we decided to show more details on where the log was emitted by configuring the logger format more, but this can improve UX internally

#### Test plan

#### UX

### DIFF
--- a/recipes/full_finetune_distributed.py
+++ b/recipes/full_finetune_distributed.py
@@ -142,8 +142,6 @@ class FullFinetuneRecipeDistributed(FTRecipeInterface):
             )
             self._log_peak_memory_stats = False
 
-        # _is_rank_zero is used primarily for logging. In the future, the logger
-        # should directly take care of this
         _, rank = training.get_world_size_and_rank()
         self._is_rank_zero = rank == 0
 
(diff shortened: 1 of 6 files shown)


torchtune/utils/_logging.py
```diff
@@ -98,4 +98,4 @@ def log_rank_zero(logger: logging.Logger, msg: str, level: int = logging.INFO) -
     rank = dist.get_rank() if dist.is_available() and dist.is_initialized() else 0
     if rank != 0:
         return
-    logger.log(level, msg)
+    logger.log(level, msg, stacklevel=2)
```
joecummings, line 101: bueno

joecummings (approved):
🫡

recipes/full_finetune_distributed.py
```diff
@@ -497,11 +497,10 @@ def _setup_model(
         # Ensure no params and buffers are on meta device
         training.validate_no_params_on_meta_device(model)
 
-        utils.log_rank_zero(
-            log,
-            f"Instantiating model and loading checkpoint took {time.perf_counter() - init_start:.2f} secs",
-        )
         if self._is_rank_zero:
+            log.info(
```
joecummings, line 501: Wait why can this not be extracted out?

RdoubleA: `init_start` is only defined for rank zero, and the string formatting happens before we check for rank zero in `log_rank_zero`

Issue #2032, opened 2024-11-20
Add NLL and DPOP weighting to DPO losses


RdoubleA:
Several modifications to the DPO loss function have shown to improve DPO model quality. These include adding a weighted negative log-likelihood loss (https://arxiv.org/pdf/2404.19733) and a DPO-positive loss (https://arxiv.org/pdf/2402.13228)

- DPOP incentivizes every token toward the preferred completion, where vanilla DPO does not do this for earlier tokens
- from NLL paper: "We see that for DPO without NLL loss there is a decrease over training for the chosen sequences, whereas for DPO with NLL there is not, which may help explain the improved performance of the latter."

We could implement these as additional float parameters in the DPO loss functions to added a weighted loss term, or as separate loss functions that you would combine. We should first assess how impactful these papers have been.

cc @SalmanMohammadi

PR #2035, opened 2024-11-20
Add RPO, DPOP losses, add lambda_dpop to basic DPO loss

krammnic:
#### Context
- [X] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

#### Changelog
* #2032

#### Test plan

#### UX

### DIFF
--- a/torchtune/training/metric_logging.py
+++ b/torchtune/training/metric_logging.py
@@ -222,7 +222,7 @@ class WandBLogger(MetricLoggerInterface):
             try:
                 output_config_fname = Path(
                     os.path.join(
-                        config.checkpointer.checkpoint_dir,
+                        config.output_dir,
                         "torchtune_config.yaml",
                     )
                 )


krammnic:
Would love to get review here.

RdoubleA:
Hey @krammnic, thanks for opening this PR. We've discussed a bit in general about defining better criteria for adding new features to the repo and I wanted to share it with you. We want to encourage the community to add new features but at the same time be selective in only the features that are most impactful and necessary in the field of fine-tuning. That being said, we arrived at two criteria:
1. Correctness. We'd want to see a reference implementation either in another fine-tuning library or an official research repo. This is to ensure we do not unintentionally onboard a feature that is broken, as the maintainers are not all experts in every new feature and can verify the implementation. We've had this recently happen with IPOLoss, which we've had to remove because it was not correct. Removing a broken feature is worst case scenario and we'd like to avoid that.
2. Prevalence. This one is a bit harder to gauge but one quantifiable metric could be citations on the paper with the method. Or perhaps a separate repo that is gaining popularity. Again, this is to filter out features that will not maintain relevance.

With those in mind, I'd like to hold off on this PR for now. I know in the past we haven't put a lot of consideration into the new features we add but this is something we'd like to rectify. This one is a bit on me since I posted the issue. 

We're working on a better process for novel features because we value contributors' time and we don't want you to spend effort to put up a PR when there hasn't been alignment to add a feature. So we're thinking of:
1. Discuss a new feature on an issue first, get explicit approval from a maintainer, before opening a PR
2. Encourage and outline how contributors can create example repos or host our own for cases such as DPOP that we still need to understand it's importance before onboarding.

Open to any thoughts you may have, and as always you've been a valuable contributor @krammnic, don't want to discourage you from continuing to do so :)

krammnic:
@RdoubleA Thanks for the answer! I see your points. Speaking about correctness, probably I can find some reference implementations and compare, also I'm open to do several full runs with this new method. I agree that definitely we should speak on structure of RLHF module in general before merge such PRs and here I'm interested in some comments of @salman probably or we can discuss this points offline.

PR #2040, opened 2024-11-21
Vector Quantized Embeddings

RdoubleA:
#### Context
- [x] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

Vector quantization is an old method introduced with the VQ-VAE (https://arxiv.org/abs/1711.00937) for achieving high fidelity image generation. It still remains a core component in many landmark models: 

- the DALL-E models use it for image generation by autoregressively predicting image token IDs
- Latent diffusion uses a VAE (not sure if it was VQ or continuous) to perform the diffusion process in the latent space
- Muse by Google uses VQVAE similarly for image generation
- Make-A-Video by Meta uses it for video generation from text

To continue to support image and more modalities in the future, we need to add the core VQ embedding layer as a module. This module will maintain the codebook embeddings and perform the lookup for discretizing an encoder input.

I've adapted the `Codebook` layer from torchmultimodal (found [here](https://github.com/facebookresearch/multimodal/blob/main/torchmultimodal/modules/layers/codebook.py#L31), apparently it shares authors with the author of this PR...). I've removed features such as initializing embeddings with encoder input for faster convergence to help simplify the code. I've kept the core embedding lookup functionality and the EMA update so that it could still be trained if users want to experiment.

#### Changelog
* Add `VectorQuantizedEmbeddings` and requisite tests

#### Test plan

#### UX


torchtune/modules/vq_embeddings.py
```diff
@@ -0,0 +1,173 @@
...
+class VectorQuantizedEmbeddings(nn.Module):
+    """
+    Vector quantized embedding layer that takes in the output of an encoder
+    and performs a nearest-neighbor lookup in the embedding space.
+    Vector quantization was introduced in Oord et al. 2017 (https://arxiv.org/pdf/1711.00937.pdf)
+    to generate high-fidelity images, videos, and audio data.
+    The embedding weights are trained with exponential moving average updates as described
+    in original paper.
+
+    Code was adapted from torchmultimodal's `Codebook module
```
pbontrager, line 23: I was going to comment that I'm more familiar with this being called a codebook. Any reason for not calling it that here? Or VQCodebook?

torchtune/modules/vq_embeddings.py
```diff
@@ -0,0 +1,173 @@
...
+        learnable: bool = False,
+        epsilon: float = 1e-7,
+    ) -> None:
+        super().__init__()
+        # Embedding weights and parameters for EMA update will be registered to buffer, as they
+        # will not be updated by the optimizer but are still model parameters.
+        # code_usage and code_avg correspond with N and m, respectively, from Oord et al.
+        randn_init_embedding = torch.randn(num_embeddings, embedding_dim)
+        self.register_buffer("embedding", randn_init_embedding.clone())
+        if learnable:
```
pbontrager, line 54: I don't really like this variable because it could disagree with the nn.module.training attribute. Is this only needed for EMA? Why wouldn't the ema code handle this?

torchtune/modules/vq_embeddings.py
```diff
@@ -0,0 +1,173 @@
...
+        token_ids = token_ids_flat.view(bsz, seq_len)
+
+        return quantized, token_ids
+
+    def extra_repr(self) -> str:
+        return "num_embeddings={}, embedding_dim={}".format(
+            self.num_embeddings, self.embedding_dim
+        )
+
+    def lookup(self, token_ids: Tensor) -> Tensor:
```
pbontrager, line 171: nit: do we want to use "lookup" here or something more consistent with the rest of the library like deocde?

pbontrager (commented):
This looks like a nice implementation. I left some nit comments and questions. But my larger question is if all of this is needed in torchtune? How much of this code is just needed to support training a codebook? Do we want/need that or should we just house the minimal amount needed for inference?

RdoubleA:
> Do we want/need that or should we just house the minimal amount needed for inference?

Open to either, was thinking since I'm putting this in modules it should at least be trainable. But it would simplify the class significantly if we only supported inference for this.

torchtune/modules/vq_embeddings.py
RdoubleA: no strong opinions, I've seen it called both. we could do `VectorQuantizedCodebook`?

pbontrager:
It's up to you here. It could be nice to already have training logic, but I'm not sure if we know what a training recipe for encoders would look like for us. If we decide to support training an encoder in the future we might make different decisions on how to split the loss logic from the module logic. If we wanted to make this module finetunable with the rest of the LLM, would we use these same methods or allow full override of the codebook?

torchtune/modules/vq_embeddings.py
pbontrager: I guess that could be confusing too since it's just a "normal" codebook.  Maybe AutoEncoderCodebook? Or just leave it.

RdoubleA: Having `VectorQuantized` is the most descriptive imo, so I will leave it as it

felipemello1 (approved):

PR #2042, opened 2024-11-21
remove default to ignore safetensors

felipemello1:
#### Context
- [ ] add a new feature
- [ ] fix a bug
- [x] other (please add here)

bad default

Testing:
Downloaded models

### DIFF
--- a/docs/source/api_ref_models.rst
+++ b/docs/source/api_ref_models.rst
@@ -217,7 +217,7 @@ To download the Qwen2.5 1.5B model, for example:
 
 .. code-block:: bash
 
-    tune download Qwen/Qwen2.5-1.5B-Instruct --output-dir /tmp/Qwen2_5-1_5B-Instruct --ignore-patterns None
+    tune download Qwen/Qwen2.5-1.5B-Instruct --output-dir /tmp/Qwen2_5-1_5B-Instruct
 
 .. autosummary::
     :toctree: generated/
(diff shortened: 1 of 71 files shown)


docs/source/api_ref_models.rst
```diff
@@ -99,7 +99,7 @@ To download the Llama3.1-405B-Instruct model:
     tune download meta-llama/Meta-Llama-3.1-405B-Instruct --ignore-patterns "original/consolidated*" --hf-token <HF_TOKEN>
 
 To download the Llama3 weights of the above models, you can instead download from `Meta-Llama-3-8B-Instruct` and
-`Meta-Llama-3-70B-Instruct`.
+`Meta-Llama-3-70B-Instruct`, and remove the ignore patterns flag.
```
felipemello1, line 102: unfortunately, 8b uses meta ckpt, and 70b uses hg ckpt. Ideally, we should update their configs to all use hf, so we can always ignore consolidated. To avoid adding too much info here, i suggested that the user can just remove the ignore patterns flag, downloading all models.

torchtune/_cli/download.py
joecummings, line 1: Remove the mention of this in kaggle_download as well.

calvinpelletier:
Not important but just for cleanliness, you could remove any `--ignore-patterns None` in the configs now that it is the default:

`find /path/to/torchtune/recipes/configs -type f -name "*.yaml" -exec sed -i 's/ --ignore-patterns None//g' {} +`

felipemello1:
> Not important but just for cleanliness, you could remove any `--ignore-patterns None` in the configs now that it is the default:
> 
> `find /path/to/torchtune/recipes/configs -type f -name "*.yaml" -exec sed -i 's/ --ignore-patterns None//g' {} +`

yessir

RdoubleA (approved):

PR #2043, opened 2024-11-21
Fix issue in loading mixed precision vocab pruned models during torchtune generation for evaluation

ifed-ucsd:
Summary: Fix issue in loading mixed precision vocab pruned models during torchtune generation for evaluation

Reviewed By: RylanC24

Differential Revision: D66029040

### DIFF
--- a/torchtune/training/precision.py
+++ b/torchtune/training/precision.py
@@ -5,7 +5,7 @@
 # LICENSE file in the root directory of this source tree.
 
 import contextlib
-from typing import Dict, Generator, Iterable, Optional, Tuple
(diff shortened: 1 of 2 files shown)


torchtune/training/precision.py
```diff
@@ -160,6 +160,8 @@ def validate_expected_param_dtype(
         ValueError: If any parameter has a different dtype than `dtype`.
     """
     for name, param in named_params:
+        if "vocab_pruning_module" in name:
```
RdoubleA, line 163: might be more generalizable to add an optional argument to ignore certain named params. As it is it is quite specific

RdoubleA (approved):

PR #2049, opened 2024-11-21
[export] Add exportable attention and kv cache

larryliu0820:
Summary: As titled.

Test Plan: Rely on new unit tests

Reviewers:

Subscribers:

Tasks:

Tags:

#### Context
- [x] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

#### Changelog
* Adding `attention.py` in `torchtune/modules/_export/`
* Adding `kv_cache.py` in `torchtune/modules/_export/`
* Adding `test_attention.py` in `tests/torchtune/modules/_export/`

#### Test plan

#### UX

### DIFF
--- a/.github/workflows/export.yaml
+++ b/.github/workflows/export.yaml
@@ -1,4 +1,4 @@
-name: Unit Test
+name: Export
 
 on:
   push:
@@ -34,7 +34,8 @@ jobs:
         run: python -m pip install --upgrade pip
       - name: Install dependencies
         run: |
-          python -m pip install torch torchvision torchao
+          bash torchtune/modules/_export/install_requirements.sh
+          python -m pip install torchao
           python -m pip install -e ".[dev]"
       - name: Run unit tests with coverage
         run: pytest tests --cov=. --cov-report=xml --durations=20 -vv
--- a/.github/workflows/export.yaml
+++ b/.github/workflows/export.yaml
@@ -3,12 +3,12 @@ name: Export
 on:
   push:
     paths:
-      - 'torchtune/modules/_export/'
-      - 'tests/torchtune/modules/_export/'
+      - 'torchtune/modules/_export/**'
+      - 'tests/torchtune/modules/_export/**'
   pull_request:
     paths:
-      - 'torchtune/modules/_export/'
-      - 'tests/torchtune/modules/_export/'
+      - 'torchtune/modules/_export/**'
+      - 'tests/torchtune/modules/_export/**'
   schedule:
     # Runs at midnight evvery day
     - cron: '0 0 * * *'
--- a/.github/workflows/export.yaml
+++ b/.github/workflows/export.yaml
@@ -2,8 +2,16 @@ name: Export
 
 on:
   push:
-    branches: [ main ]
+    paths:
+      - 'torchtune/modules/_export/'
+      - 'tests/torchtune/modules/_export/'
   pull_request:
+    paths:
+      - 'torchtune/modules/_export/'
+      - 'tests/torchtune/modules/_export/'
+  schedule:
+    # Runs at midnight evvery day
+    - cron: '0 0 * * *'
 
 concurrency:
   group: unit-test${{ github.workflow }}-${{ github.ref == 'refs/heads/main' && github.run_number || github.ref }}
@@ -14,7 +22,7 @@ defaults:
     shell: bash -l -eo pipefail {0}
 
 jobs:
-  unit_tests:
+  export_unit_tests:
     if: github.repository_owner == 'pytorch'
     runs-on: ubuntu-latest
     strategy:
@@ -38,6 +46,6 @@ jobs:
           python -m pip install torchao
           python -m pip install -e ".[dev]"
       - name: Run unit tests with coverage
-        run: pytest tests --cov=. --cov-report=xml --durations=20 -vv
+        run: pytest tests/torchtune/modules/_export --cov=. --cov-report=xml --durations=20 -vv
       - name: Upload Coverage to Codecov
         uses: codecov/codecov-action@v3
--- a/.github/workflows/unit_test.yaml
+++ b/.github/workflows/unit_test.yaml
@@ -37,6 +37,6 @@ jobs:
           python -m pip install torch torchvision torchao
           python -m pip install -e ".[dev]"
       - name: Run unit tests with coverage
-        run: pytest tests --cov=. --cov-report=xml --durations=20 -vv
+        run: pytest tests --ignore tests/torchtune/modules/_export --cov=. --cov-report=xml --durations=20 -vv
       - name: Upload Coverage to Codecov
         uses: codecov/codecov-action@v3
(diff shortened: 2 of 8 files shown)


.github/workflows/unit_test.yaml
```diff
@@ -34,7 +34,9 @@ jobs:
         run: python -m pip install --upgrade pip
       - name: Install dependencies
         run: |
-          python -m pip install torch torchvision torchao
+          NIGHTLY_VERSION=dev20241121
+          python -m pip install torch=="2.6.0.${NIGHTLY_VERSION}" torchvision=="0.20.0.${NIGHTLY_VERSION}" --extra-index-url https://download.pytorch.org/whl/nightly/cpu
+          python -m pip install torchao
```
ebsmothers, line 39: I don't think we want to change the default behavior of our unit test workflow here. Can we figure out a way to decouple the export CI a little more?

larryliu0820: Can you say a bit more? Are you ok with a separate CI job running these tests with torch nightly installed?

larryliu0820: Added `Export` CI workflow.

torchtune/modules/_export/install_requirements.sh
```diff
@@ -0,0 +1,11 @@
...
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+NIGHTLY_VERSION="dev20241121"
+
+# Install pytorch nightly for export-friendly modules to run.
+pip install torch==2.6.0.${NIGHTLY_VERSION} torchvision==0.20.0.${NIGHTLY_VERSION} --extra-index-url https://download.pytorch.org/whl/nightly/cpu
```
pbontrager, line 11: This sh file doesn't really make sense to me. The goal of the export folder is to house export friendly modules not modify the library setup. Currently torchtune operates under the assumption that the user installs pytorch themselves and some of our features our gated behind nightlies where we'll tell the user to upgrade to nightlies if they want to use that feature. The same pattern should be followed here.

tests/torchtune/modules/_export/test_attention.py
```diff
@@ -0,0 +1,230 @@
...
+import unittest
+
+import torch
+from torch._inductor.package import load_package, package_aoti
+from torch.testing import assert_close
+from torchtune.models.llama3_1._position_embeddings import Llama3ScaledRoPE
+from torchtune.modules._export.attention import (
+    MultiHeadAttention as ExportMultiHeadAttention,
+)
+from torchtune.modules.attention import MultiHeadAttention as TTMultiHeadAttention
```
pbontrager, line 23: This unit test should exist in a downstream library as per the original plan. Parity check tests (TTMultiHeadAttention == ExportMultiHeadAttention) can't exist in torchtune otherwise our PRs will be blocked whenever we make a change to MultiHeadAttention. Unit tests in torchtune (such as checking export) should be decoupled from implementations outside of the _export folder.

.github/workflows/export.yaml
```diff
@@ -0,0 +1,43 @@
+name: Export
```
pbontrager, line 1: We run tests on pytorch nightlies every day [flow](https://github.com/pytorch/torchtune/blob/main/.github/workflows/gpu_test.yaml). I'd rather not have a separate flow for export and have the tests run as part of the standard flows.

pbontrager (commented):
Thanks for putting up this first commit. I'm not reviewing the modules themselves but the integration with unit tests as this is the first time we're doing this. I left some comments related to this.

.github/workflows/export.yaml
larryliu0820: So that was my first version, the only caveat is that I need to install nightly in the CI flow.

@ebsmothers commented on my first version:

> I don't think we want to change the default behavior of our unit test workflow here. Can we figure out a way to decouple the export CI a little more?

That's why I separated it out as a new CI flow.

.github/workflows/export.yaml
```diff
@@ -0,0 +1,51 @@
...
+  run:
+    shell: bash -l -eo pipefail {0}
+
+jobs:
+  export_unit_tests:
+    if: github.repository_owner == 'pytorch'
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: ['3.9', '3.10', '3.11']
```
dvorjackz, line 30: ET doesn't explicitly support 3.9, should we be testing it?

tests/torchtune/modules/_export/test_attention.py
```diff
@@ -0,0 +1,230 @@
...
+        assert_close(et_res, tt_res)
+
+        # test with kv cache
+        self.et_mha.setup_cache(1, dtype=torch.float32, max_seq_len=20)
+        self.tt_mha.setup_cache(1, dtype=torch.float32, max_seq_len=20)
+
+        et_res = self.et_mha(self.x, self.x)  # Self attention.
+        tt_res = self.tt_mha(self.x, self.x)  # Self attention.
+
+        self.assertTrue(torch.allclose(et_res, tt_res))
```
dvorjackz, line 123:
```suggestion
        assert_close(et_res, tt_res)
```

tests/torchtune/modules/_export/test_attention.py
```diff
@@ -0,0 +1,230 @@
...
+        self.tt_mha.reset_cache()
+
+        et_res = self.et_mha(
+            self.x, self.x, input_pos=self.input_pos
+        )  # Self attention with input pos.
+        tt_res = self.tt_mha(
+            self.x, self.x, input_pos=self.input_pos
+        )  # Self attention with input pos.
+
+        self.assertTrue(torch.allclose(et_res, tt_res))
```
dvorjackz, line 134:
```suggestion
        assert_close(et_res, tt_res)
```

tests/torchtune/modules/_export/test_attention.py
```diff
@@ -0,0 +1,230 @@
...
+        mask = self.causal_mask[self.input_pos, :]
+        # First run
+        et_res = self.et_mha(
+            self.x, self.x, mask=mask, input_pos=self.input_pos
+        )  # Self attention with input pos.
+        tt_res = self.tt_mha(
+            self.x, self.x, mask=mask, input_pos=self.input_pos
+        )  # Self attention with input pos.
+
+        self.assertTrue(torch.allclose(et_res, tt_res))
```
dvorjackz, line 216:
```suggestion
       assert_close(et_res, tt_res)
```

dvorjackz (commented):
Should we also copy over the [README](https://github.com/pytorch/executorch/blob/main/extension/llm/modules/README.md)?

.github/workflows/export.yaml
larryliu0820: This is testing export not ET

ebsmothers (approved):

PR #2069, opened 2024-11-25
T5 Encoder

calvinpelletier:
#### Context
- [x] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

#### Changelog
* T5 tokenizer
* T5 encoder
* convert weights from HF's T5 to ours
* unit tests

#### Analysis

Comparison to HF's implemention (batch of text -> encoder output):
- 6.7e-5 MSE output difference
- ours is ~5% faster

#### Test plan

#### UX

#### Minimal test code

```python
import torch

from torchtune.models.t5 import t5_tokenizer, t5_v1p1_xxl_encoder
from torchtune.training.checkpointing._checkpointer import FullModelHFCheckpointer

MAX_SEQ_LEN = 512

# tune download google/t5-v1_1-xxl --output-dir /tmp/t5-hf
tokenizer = t5_tokenizer("/tmp/t5-hf/spiece.model", max_seq_len=MAX_SEQ_LEN)
checkpointer = FullModelHFCheckpointer(
    "/tmp/t5-hf",
    ["pytorch_model.bin"],
    "T5_ENCODER",
    "/tmp/t5-tt",
)

model = t5_v1p1_xxl_encoder(max_seq_len=MAX_SEQ_LEN)
model.load_state_dict(checkpointer.load_checkpoint()["model"])
model = model.to(device="cuda", dtype=torch.bfloat16).eval().requires_grad_(False)


def tokenize(texts):
    result = torch.full(
        (len(texts), tokenizer.max_seq_len),
        tokenizer.pad_id,
        dtype=torch.int,
    )
    for i, text in enumerate(texts):
        tokens = tokenizer.encode(text)
        result[i, : len(tokens)] = torch.tensor(tokens)
    return result


tokens = tokenize(
    [
        "a cow jumping over the moon",
        "a helpful AI assistant",
    ]
)
encoding = model(tokens.to("cuda"))
```


torchtune/models/t5/_encoder.py
```diff
@@ -0,0 +1,304 @@
...
+        # [bsz, seq_len, embed_dim] -> [bsz, num_heads, seq_len, head_dim]
+        q = q.view(bsz, seq_len, self.num_heads, self.head_dim).transpose(1, 2)
+        k = k.view(bsz, seq_len, self.num_heads, self.head_dim).transpose(1, 2)
+        v = v.view(bsz, seq_len, self.num_heads, self.head_dim).transpose(1, 2)
+
+        # attention with relative position bias
+        attn_score = torch.matmul(q, k.transpose(-2, -1))
+        attn_score += rel_pos_bias
+        attn_weight = F.softmax(attn_score.float(), dim=-1).to(attn_score.dtype)
+        attn_out = torch.matmul(attn_weight, v)
```
calvinpelletier, line 209: This part could be simplified by using `F.scaled_dot_product_attention` by repurposing the `mask` argument for `rel_pos_bias` (because `scaled_dot_product_attention` simply adds the mask to the attention score when the mask is a float tensor). However, when I tried this it was significantly slower for some reason

torchtune/models/t5/_encoder.py
```diff
@@ -0,0 +1,304 @@
...
+            torch.Tensor: relative position bias tensor with shape [1, num_heads, max_seq_len, max_seq_len]
+        """
+        # convert bucket numbers to bias values for each attention head
+        x = self.embedding(self.relative_position_to_bucket)
+
+        # shape [max_seq_len, max_seq_len, num_heads] -> [1, num_heads, max_seq_len, max_seq_len]
+        return x.permute([2, 0, 1]).unsqueeze(0)
+
+
+def _calc_birectional_rel_pos_to_bucket(
```
calvinpelletier, line 258: should I add more comments in this function explaining each operation? or is it fine to just leave it a bit opaque

torchtune/models/t5/_model_builders.py
```diff
@@ -0,0 +1,54 @@
...
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+from torchtune.models.t5._component_builders import t5_encoder
+from torchtune.models.t5._encoder import T5Encoder
+from torchtune.models.t5._tokenizer import T5Tokenizer
+
+
+def t5_v1p1_xxl_encoder(max_seq_len: int = 512) -> T5Encoder:
```
calvinpelletier, line 12: thoughts on writing decimal points as `p` instead of `_` in snake case? imo it's hard to read the `_` decimals:

`t5_v1_1_xxl_encoder` vs `t5_v1p1_xxl_encoder`

torchtune/models/t5/_encoder.py
ebsmothers: I wonder whether this is a case where we could benefit from using flex attention?

pbontrager (commented):
I gave this a high level pass so far. This looks really clean and good. My only concern is wish us having to have a custom T5 layer and attention module. If flex attention would let us use our existing modules I'd prefer to go down that route.

torchtune/modules/tokenizers/_sentencepiece.py
```diff
@@ -7,6 +7,7 @@
 from typing import List, Optional
 
 from sentencepiece import SentencePieceProcessor
+
```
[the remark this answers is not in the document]
calvinpelletier: its from the linter

torchtune/models/t5/_model_builders.py
pbontrager: I think the p could be better. But I'd want to switch everything to this notation then

torchtune/models/t5/_encoder.py
calvinpelletier: Flex attention requires boolean masks AFAICT: https://github.com/pytorch/pytorch/blob/main/torch/nn/attention/flex_attention.py#L826-L827

torchtune/models/t5/_tokenizer.py
```diff
@@ -0,0 +1,67 @@
...
+        max_seq_len (int): the context length
+        truncate (bool): whether to truncate the token sequence when longer than max_seq_len
+    """
+
+    def __init__(self, path: str, max_seq_len: int = 512, truncate: bool = True):
+        super().__init__(path)
+        self.max_seq_len = max_seq_len
+        self.truncate = truncate
+
+    def encode(self, text: str) -> List[int]:
```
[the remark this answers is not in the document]
calvinpelletier: It has it. It's in the base tokenizer class (`SentencePieceBaseTokenizer`). It just didn't require any modifications

torchtune/models/t5/_encoder.py
```diff
@@ -0,0 +1,304 @@
...
+        sa_norm: nn.Module,
+        mlp_norm: nn.Module,
+    ) -> None:
+        super().__init__()
+        self.attn = attn
+        self.mlp = mlp
+        self.sa_norm = sa_norm
+        self.mlp_norm = mlp_norm
+
+    def forward(self, x: Tensor, rel_pos_bias: Tensor) -> Tensor:
```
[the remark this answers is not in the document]
calvinpelletier: Well, kind of... it's a float tensor that gets added to the attention scores (which is how `scaled_dot_product_attention` deals with float masks, so yeah we could think of it as a mask). 

I didn't use our modules because: 
- we can't use flex attention with float masks
- our `MultiHeadAttention`/`TransformerSelfAttentionLayer` modules expect the masks to be boolean tensors (according to the docstring at least)
- `MultiHeadAttention` uses the default attention scaling (`1/sqrt(dim)`), but T5 doesn't scale it at all
- `MultiHeadAttention` uses `F.scaled_dot_product_attention` which is much slower with float masks than the manual implementation I went with

We could switch to our modules with a couple small changes (update the docstrings to clarify that the mask can also be a float tensor and add an argument for disabling attention scaling), but given how little code is required to just implement separate versions for T5, I thought it was cleaner to leave our modules alone (especially since this implementation is faster).

torchtune/models/t5/_model_builders.py
calvinpelletier: I'll make a separate PR for this

pbontrager: Can you use the _ notation for now and just update this along with everything else in that PR?

torchtune/models/t5/_encoder.py
pbontrager: I believe you can use any mask for flexattention with scoremod. I think we should land this as is but create an issue to update this to use flex attention.

pbontrager (approved):
I think this looks good. I'd like an issue to update the attention to use flex attention but that doesn't need to happen here.

Issue #2048, opened 2024-11-21
`FullModelHFCheckpointer` saved checkpoint isn't compatible with Huggingface `transformers` model loading


vancoykendall:
Loading huggingface `transformers` models is done with the  `from_pretrained()` method. For pytorch or safetensors checkpoints, this method expects a `pytorch_model.bin` or `model.safetensors` file for single file checkpoints. For sharded checkpoints, it expects either a `pytorch_model.bin.index.json` or `model.safetensors.index.json` file that maps each weight to the shard file (I think sharded checkpoint files can have arbitrary naming).

Currently, the `FullModelHFCheckpointer` doesn’t name single file checkpoitns `pytorch_model.bin` or `model.safetensors` and doesn't create an index.json file for sharded checkpoints. Thus, saved checkpoints can't be loaded with `from_pretrained()`. https://github.com/pytorch/torchtune/blob/e9fd56a812cf0ba151fa164a45eb04056d099726/torchtune/training/checkpointing/_checkpointer.py#L622-L639

vancoykendall:
Here's some potentially useful links related to huggingface checkpoint creation
Huggingface documents index file creation here: https://huggingface.co/docs/transformers/v4.46.3/en/big_models#sharded-checkpoints

Code for sharding state dict here: https://github.com/huggingface/huggingface_hub/blob/v0.26.2/src/huggingface_hub/serialization/_torch.py

felipemello1:
hey @vancoykendall , thank you so much for this issue! I am one of the maintainers and am working on fixing this. @joecummings is also working on refactoring our checkpointing. This will become much easier in the following days.

Meanwhile, to unblock you, we mention and give an example here: https://pytorch.org/torchtune/main/tutorials/e2e_flow.html#using-torchtune-checkpoints-with-other-libraries

You can manually make it .bin (i know, that's not fun) and select the files to keep in the folder. Then from_pretrained will work.

vancoykendall:
Nice thanks! I also just patched the `save_checkpoint` method locally so I don't have to convert them anymore

```python
# split the state_dict into separate dicts, one for each output checkpoint file
            split_state_dicts: Dict[str, Dict[str, torch.Tensor]] = {}
            total_size = 0
            for key, weight in state_dict[training.MODEL_KEY].items():
                cpt_idx = self._weight_map[key]
                if cpt_idx not in split_state_dicts:
                    split_state_dicts[cpt_idx] = {}
[...]

PR #2074, opened 2024-11-26
Update checkpointing directory

felipemello1:
Co-authored-by: vancoyendall <vancoykendall@gmail.com>#### Context
- [x] add a new feature
- [x] fix a bug
- [x] other (please add here)

TLDR using vLLM and Hugginface .frompretrained

1) train your model using torchtune main (nightlies). It will produce a folder like this:
image

2. Copy the contents of your latest epoch to base_model folder, which contains the checkpoint_dir original's content, without the model files (.pt, .bin, .safetensors):
```
cp /tmp/llama_3_2_1b/lora_single_device/epoch_2/* /tmp/llama_3_2_1b/lora_single_device/base_model
```

Making it look like this:
image

3. Now, you can use it with vLLM and Huggingface. There is one catch here: when using lora, we output MERGED weights AND the adapter. You should NOT use both at the same time. Either use ONLY the merged weights OR the BASE UNTRAINED MODEL + adapter.

3.1. Using hugginface .from_pretrained with BASE UNTRAINED MODEL + adapter

```
from peft import PeftModel
from transformers import AutoModelForCausalLM, AutoTokenizer

# Define the model and adapter paths
original_model_name = "meta-llama/Llama-3.2-1B-Instruct"
trained_model_path = "/tmp/torchtune/llama3_2_1B/lora_single_device/base_model"

model = AutoModelForCausalLM.from_pretrained(original_model_name)

# huggingface will look for adapter_model.safetensors and adapter_config.json
peft_model = PeftModel.from_pretrained(model, trained_model_path)

# Load the tokenizer
tokenizer = AutoTokenizer.from_pretrained(original_model_name)

# Function to generate text
def generate_text(model, tokenizer, prompt, max_length=50):
    inputs = tokenizer(prompt, return_tensors="pt")
    outputs = model.generate(**inputs, max_length=max_length)
    return tokenizer.decode(outputs[0], skip_special_tokens=True)

prompt = "Complete the sentence: 'Once upon a time...'"
print("Base model output:", generate_text(peft_model, tokenizer, prompt))
```

3.2. using the MERGED TRAINED MODEL with vLLM

IMPORTANT: this will not work right away. Your output directory has 2 files:
- ft-model-00001-of-00001.safetensors
- adapter_model.safetensors

vLLM doesnt know what is a model and what is an adapter. When it tries to load the adapter, it will raise an error. Therefore, delete adapter_model.safetensors from the folder, and it will work

```
rm /tmp/torchtune/llama3_2_1B/lora_single_device/base_model/adapter_model.safetensors
```

Now you can run vLLM locally
```
from vllm import LLM, SamplingParams

def print_outputs(outputs):
    for output in outputs:
        prompt = output.prompt
        generated_text = output.outputs[0].text
        print(f"Prompt: {prompt!r}, Generated text: {generated_text!r}")
    print("-" * 80)

llm = LLM(
    model="/tmp/torchtune/llama3_2_1B/lora_single_device",
    load_format="safetensors",
    kv_cache_dtype="auto",
)
sampling_params = SamplingParams(max_tokens=16, temperature=0.5)

conversation = [
    {"role": "system", "content": "You are a helpful assistant"},
    {"role": "user", "content": "Hello"},
    {"role": "assistant", "content": "Hello! How can I assist you today?"},
    {
        "role": "user",
        "content": "Write an essay about the importance of higher education.",
    },
]
outputs = llm.chat(conversation, sampling_params=sampling_params, use_tqdm=False)
print_outputs(outputs)
```

#### Context

In torchtune's current state, if checkpoint_dir != outputdir, it breaks. Since the files are all mixed, saved as .pt and without the proper configs, its hard for users to readily use it with vllm/huggingface, resulting in issues such as https://github.com/pytorch/torchtune/issues/2048, https://github.com/pytorch/torchtune/issues/2025 and https://github.com/pytorch/torchtune/issues/2118.

This PR is **NOT** a major refactor. Everything is backwards compatible. The intention here is just to organize the output_dir and allow users to quickly use their models with HF and vLLM.

#### Changelog

1. The folder is automatically created/populated like described above

2. Initially, base_model has all the files from the checkpoint_dir, **except** those that end in .pt, .safetensors, .bin, etc

3. We save the original_model info in the adapter.config

4. naming got standardized

5. safetensors is the default for HF ckpt

6. Solved bugs when input_dir != output_dir

#### Next steps
Update docs

#### Unresolved issues
Left TODOs in the code, to be addressed in follow up PRs. It makes the code ugly, but we are due a refactor.

#### Test plan
- unit tests
- resumed pretrained
image
- ran vllm and hugginface


torchtune/_cli/download.py
```diff
@@ -85,18 +87,6 @@ def _add_arguments(self) -> None:
...
-            type=str,
-            required=False,
-            default="auto",
-            help=(
-                "To be used with `output-dir`. If set to 'auto', the cache directory will be used and the file will be"
-                " either duplicated or symlinked to the local directory depending on its size. It set to `True`, a"
-                " symlink will be created, no matter the file size. If set to `False`, the file will either be"
-                " duplicated from cache (if already exists) or downloaded from the Hub and not cached."
-            ),
-        )
```
felipemello1, line 99: deprecated. Its a no-op.

torchtune/_cli/download.py
```diff
@@ -224,6 +207,16 @@ def _download_from_kaggle(self, args: argparse.Namespace) -> None:
...
             output_dir = model_download(model_handle)
+
+            # save the repo_id. This is necessary because the download step is a separate command
+            # from the rest of the CLI. When saving a model adapter, we have to add the repo_id
+            # to the adapter config.
+            file_path = os.path.join(output_dir, training.REPO_ID_FNAME).with_suffix(
+                ".json"
+            )
+            with open(file_path, "w") as json_file:
+                json.dump({"repo_id": args.repo_id}, json_file, indent=4)
```
felipemello1, line 218: related: https://github.com/pytorch/torchtune/pull/2026

torchtune/training/checkpointing/_utils.py
```diff
@@ -207,28 +235,13 @@ def safe_torch_load(
...
-    Args:
-        path (Path): Path to save the configuration file.
-        config (Dict[str, Any]): Configuration dictionary to save.
-    """
-    if not path.is_dir():
-        path.mkdir(exist_ok=True)
-    file_path = Path.joinpath(path, "config.json")
-    if not file_path.exists():
-        with open(file_path, "w") as f:
-            json.dump(config, f)
```
felipemello1, line 229: replaced it with "copy_files", so we save every file, and not only config

torchtune/training/checkpointing/__init__.py
```diff
@@ -33,13 +41,21 @@
...
     "Checkpointer",
     "update_state_dict_for_classifier",
     "ADAPTER_CONFIG",
+    "ADAPTER_CONFIG_FNAME",
+    "ADAPTER_MODEL_FNAME",
+    "SHARD_FNAME",
+    "SAFETENSOR_INDEX_FNAME",
+    "SUFFIXES_TO_NOT_COPY",
+    "TORCH_INDEX_FNAME",
+    "REPO_ID_FNAME",
```
joecummings, line 58: I don't think these need to be exposed as public APIs right?

torchtune/_cli/download.py
```diff
@@ -196,6 +170,14 @@ def _download_from_huggingface(self, args: argparse.Namespace) -> None:
             msg = f"Failed to download {args.repo_id} with error: '{e}' and traceback: {tb}"
             self._parser.error(msg)
 
+        # save the repo_id. This is necessary because the download step is a separate command
+        # from the rest of the CLI. When saving a model adapter, we have to add the repo_id
+        # to the adapter config.
+        # TODO: this needs to be updated when we start using HF cache
+        file_path = os.path.join(true_output_dir, training.REPO_ID_FNAME + ".json")
+        with open(file_path, "w") as json_file:
+            json.dump({"repo_id": args.repo_id}, json_file, indent=4)
```
felipemello1, line 179: hf download

torchtune/_cli/download.py
felipemello1: kaggle donwload

torchtune/training/checkpointing/_checkpointer.py
```diff
@@ -246,24 +262,25 @@ def save_checkpoint(
...
-        self._output_dir.mkdir(exist_ok=True)
 
         # Output file is always a .pt file with the epoch number in the name
         if not adapter_only:
-            checkpoint_file = Path.joinpath(
+            output_path = Path.joinpath(
                 self._output_dir, f"torchtune_model_{epoch}"
             ).with_suffix(".pt")
-            torch.save(state_dict[training.MODEL_KEY], checkpoint_file)
+            output_path.parent.mkdir(parents=True, exist_ok=True)
```
felipemello1, line 271: not sure if i need every mkdir here. Will revise.

torchtune/training/checkpointing/_checkpointer.py
```diff
@@ -343,26 +362,15 @@ def __init__(
         adapter_checkpoint: Optional[str] = None,
         recipe_checkpoint: Optional[str] = None,
         resume_from_checkpoint: bool = False,
-        safe_serialization: bool = False,
+        safe_serialization: bool = True,
```
felipemello1, line 365: are we ok with changing the default? I think it has some performance benefits when working with hf

torchtune/training/checkpointing/_checkpointer.py
```diff
@@ -343,26 +362,15 @@ def __init__(
...
-            formatted_checkpoint_files = FormattedCheckpointFiles.from_dict(
-                checkpoint_files
-            )
-            checkpoint_files = formatted_checkpoint_files.build_checkpoint_filenames()
-        self._checkpoint_paths = self._validate_hf_checkpoint_files(checkpoint_files)
-        self._adapter_checkpoint = (
-            get_path(self._checkpoint_dir, adapter_checkpoint)
-            if adapter_checkpoint
-            else None
-        )
```
felipemello1, line 360: moved down

torchtune/training/checkpointing/_checkpointer.py
```diff
@@ -374,27 +382,125 @@ def __init__(
...
+
+            # Case 3.2: Check {checkpoint_dir}/{file_path}
+            if len(self._checkpoint_paths) != len(checkpoint_files):
+                logger.warning(
+                    f"When resuming from ckpt, we could not find all model files in {self._output_dir=}. "
+                    "This is expected if you set `save_adapter_weights_only=True`. In this case, we will load from checkpoint_dir. "
+                    "However, if you set `save_adapter_weights_only=False`, this is unexpected. "
+                    "Perhaps you forgot to add `epoch_{epoch}/` to your filename? "
+                    "Using checkpoint_dir instead..."
+                )
```
felipemello1, line 472: i don't really like this. The other options are 1) to actually fix the issue, which is knowing save_adapter_only, or 2) silently let it happen, which is dangerous if the adapter + model was trained (e.g. embeddings), and the user forgot to change the file names.

Not sure if there is a 3rd

torchtune/training/checkpointing/_checkpointer.py
```diff
@@ -612,38 +716,86 @@ def save_checkpoint(
...
+                map_original_name_to_new_name[cpt_idx] = shard_name
+                output_path = Path.joinpath(
+                    self._output_dir, f"epoch_{epoch}", shard_name
+                )
+                output_path.parent.mkdir(parents=True, exist_ok=True)
                 if not self._safe_serialization:
-                    output_path = Path.joinpath(
-                        self._output_dir, f"hf_model_{cpt_idx}_{epoch}"
-                    ).with_suffix(".pt")
+                    output_path = output_path.with_suffix(".bin")
```
felipemello1, line 750: no more .pt. Lets do .bin.

torchtune/training/checkpointing/__init__.py
felipemello1: agreed! I was following the pattern, but i guess that these other constants are used in multiple files. I will revisit this.

recipes/configs/llama3_2/1B_lora_single_device.yaml
```diff
@@ -35,10 +36,10 @@ checkpointer:
   _component_: torchtune.training.FullModelHFCheckpointer
   checkpoint_dir: /tmp/Llama-3.2-1B-Instruct/
   checkpoint_files: [
-    model.safetensors
+     epoch_2/ft-model-00001-of-00001.safetensors
```
joecummings, line 39: typo

torchtune/models/convert_weights.py
```diff
@@ -231,6 +230,7 @@ def _permute(t, n_heads):
 
 def tune_to_peft_adapter_config(
     adapter_config: Dict[str, Any],
+    base_model_name_or_path: Optional[str] = None,
```
joecummings, line 233: @ebsmothers  How was the PEFT model figuring this out before?

torchtune/training/__init__.py
```diff
@@ -84,6 +91,13 @@
     "Checkpointer",
     "update_state_dict_for_classifier",
     "ADAPTER_CONFIG",
+    "ADAPTER_CONFIG_FNAME",
```
joecummings, line 94: Yeah these can live in individual files or a `constants` file within the training, but let's not expose these please.

torchtune/training/checkpointing/_checkpointer.py
```diff
@@ -146,31 +154,39 @@ def __init__(
...
-            if recipe_checkpoint is None:
-                raise ValueError(
-                    "If resume_from_checkpoint is True, recipe_checkpoint file must be provided."
-                )
-            self._recipe_checkpoint = get_path(self._checkpoint_dir, recipe_checkpoint)
+        # save all files in input_dir, except model weights and mapping, to output_dir
+        # this is useful to preserve the tokenizer, configs, license, etc.
+        copy_files(
+            self._checkpoint_dir,
+            Path.joinpath(self._output_dir, BASE_MODEL_DIRNAME),
```
joecummings, line 173:
```suggestion
           self._output_dir / BASE_MODEL_DIRNAME,
```

torchtune/training/checkpointing/_utils.py
```diff
@@ -277,3 +299,161 @@ def update_state_dict_for_classifier(
...
+
+    # Find the folder with the largest iter number
+    if iter_folders:
+        latest_epoch_folder = max(iter_folders, key=lambda x: x[1])[0]
+
+    return latest_epoch_folder
+
+
+# TODO: instead of copying, make it a symlink when we start using HF cache
+def copy_files(
```
joecummings, line 326: Can you make ignore suffixes kwarg only?

torchtune/training/checkpointing/_checkpointer.py
```diff
@@ -374,27 +382,125 @@ def __init__(
...
-        # resume_from_checkpoint is True
-        self._recipe_checkpoint = None
-        if self._resume_from_checkpoint:
-            if recipe_checkpoint is None:
-                raise ValueError(
-                    "If resume_from_checkpoint is True, recipe_checkpoint file must be provided."
+        # repo_id is necessary for when saving an adapter config, so its compatible with HF.
+        # This json file is produced and saved in the download step.
+        # contents are {"repo_id": "some_model/some_model_version"}
+        repo_id_path = Path.joinpath(
```
joecummings, line 396: Again I think this can just be joined with a forward-slash ?

torchtune/training/checkpointing/_utils.py
felipemello1: I don't follow. Can you give an example?

torchtune/training/checkpointing/_checkpointer.py
joecummings: Easier to read IMO and I think we use it in other places, but es okay.

torchtune/training/checkpointing/_utils.py
joecummings:
```python
def copy_files(
	input,
	output,
	*,
	ignore_suffixes
)
```

torchtune/_cli/download.py
SalmanMohammadi: Should we add a deprecation warning rather than removing?

felipemello1: good question! There is already a deprecation warning on huggingface side, so i think it should be ok to remove. Let me know if you disagree.

torchtune/models/convert_weights.py
ebsmothers: Oh it wasn't. That's why we loaded PEFT models in two steps: 

```
model = AutoModelForCausalLM.from_pretrained(model_id)
peft_model = PeftModel.from_pretrained(model, checkpoint_dir)
```

instead of 

```
AutoModelForCausalLM.from_pretrained(checkpoint_dir)
```

I had a hacky version of this in #2026 but it was pointed out by @pbontrager that this shouldn't be present for models that are gonna get pushed to the hub (in that case we would want the hub model ID here, not a local path). I'm curious how we wanna address that here, should we wrap hugging-face CLI upload with something that will modify the path?

recipes/configs/llama3_2/1B_full_single_device.yaml
```diff
@@ -19,11 +19,16 @@
 #
 # This config works only for training on single device.
 
+output_dir: /tmp/llama_3_2_1b/full_single_device
```
ebsmothers, line 22: What's the purpose of these config changes? Aren't these just a subset of the broader config changes in #2107?

torchtune/training/checkpointing/_utils.py
```diff
@@ -207,28 +244,13 @@ def safe_torch_load(
                 map_location="cpu",
                 mmap=mmap,
                 weights_only=weights_only,
+                # TODO: should we add assign=True?
```
ebsmothers, line 247: This is an arg to `Module.load_state_dict`, right? Not `torch.load`

torchtune/training/checkpointing/_utils.py
ebsmothers: Call me crazy but do we need a full Python function reinventing recursive copy here? E.g. 

```
os.system(
f"rsync -av --ignore-existing {" ".join([f"--exclude *{}" for ignore_suffix in ignore_suffixes])} {input_dir} {output_dir}"
)
```

torchtune/training/checkpointing/_utils.py
```diff
@@ -277,3 +299,161 @@ def update_state_dict_for_classifier(
...
+
+    return Path(recipe_checkpoint)
+
+
+def get_adapter_checkpoint_path(
+    output_dir: Path,
+    adapter_checkpoint: Optional[str] = None,
+    resume_from_checkpoint: bool = False,
+    pattern: str = r"^epoch_(\d+)",
+):
```
ebsmothers, line 430: add return types to these (e.g. here we return optional path but for recipe checkpoint we always return path)

torchtune/training/checkpointing/_utils.py
```diff
@@ -277,3 +299,161 @@ def update_state_dict_for_classifier(
...
+    """
+    If recipe_checkpoint is None, look for recipe_state.pt in {output_dir}/{RECIPE_STATE_DIRNAME}/recipe_state.pt
+    This is to make it easier to resume from a previous run, without having to specify the recipe_checkpoint.
+
+    Args:
+        output_dir (Path): Directory containing the recipe checkpoint.
+        recipe_checkpoint (Optional[str]): Name of the recipe checkpoint file. Defaults to None.
+        resume_from_checkpoint (bool): Whether to resume from a checkpoint.
+    Returns:
+        Path: Path to the recipe checkpoint file.
```
ebsmothers, line 398: this can also return none, right?

torchtune/training/checkpointing/_checkpointer.py
```diff
@@ -146,31 +161,32 @@ def __init__(
...
-                raise ValueError(
-                    "If resume_from_checkpoint is True, recipe_checkpoint file must be provided."
-                )
-            self._recipe_checkpoint = get_path(self._checkpoint_dir, recipe_checkpoint)
+        # save all files in input_dir, except model weights and mapping, to output_dir
+        # this is useful to preserve the tokenizer, configs, license, etc.
+        copy_files(
+            self._checkpoint_dir,
+            Path.joinpath(self._output_dir, BASE_MODEL_DIRNAME),
+            ignore_suffixes=SUFFIXES_TO_NOT_COPY,
```
ebsmothers, line 174: Do we think that a single global variable for `SUFFIXES_TO_NOT_COPY` will suffice here? Maybe I'm paranoid but I anticipate this will be something people will need to hack into and change

torchtune/training/checkpointing/_checkpointer.py
```diff
@@ -374,27 +387,125 @@ def __init__(
...
-        if self._resume_from_checkpoint:
-            if recipe_checkpoint is None:
-                raise ValueError(
-                    "If resume_from_checkpoint is True, recipe_checkpoint file must be provided."
+        # repo_id is necessary for when saving an adapter config, so its compatible with HF.
+        # This json file is produced and saved in the download step.
+        # contents are {"repo_id": "some_model/some_model_version"}
+        repo_id_path = Path.joinpath(self._checkpoint_dir, REPO_ID_FNAME).with_suffix(
+            ".json"
+        )
```
ebsmothers, line 403: 😍

torchtune/training/checkpointing/_utils.py
SalmanMohammadi: I just use [shutil.copy_tree](https://docs.python.org/3/library/shutil.html#shutil.copytree)

torchtune/training/checkpointing/_checkpointer.py
```diff
@@ -884,10 +1054,12 @@ def save_checkpoint(
                     model_state_dict
                 )
 
-            # Output file is always a .pt file with the epoch number in the name
+            # TODO: We should consider adding adapter/model config, like we do for HF.
```
ebsmothers, line 1057: Is this todo applicable? Like what would we do with these configs for a Meta-format checkpoint anyways? (Maybe I am missing the point here)

recipes/configs/llama3_2/1B_full_single_device.yaml
felipemello1: i changed only these 2 configs here for testing. Then i created the other PR so its still easy to review this one. Once the other PR is merged, i will merge it into this one and retest it

torchtune/training/checkpointing/_checkpointer.py
```diff
@@ -374,27 +387,125 @@ def __init__(
...
+
+        self._adapter_checkpoint = get_adapter_checkpoint_path(
+            output_dir=self._output_dir,
+            adapter_checkpoint=adapter_checkpoint,
+            resume_from_checkpoint=self._resume_from_checkpoint,
+            pattern=r"^epoch_(\d+)",
+        )
+
+        # e.g.checkpoint_files = [model-00001-of-00002.safetensors, model-00002-of-00002.safetensors]
+        # Case 1: no resuming from ckpt. Just load from checkpoint_dir.
```
ebsmothers, line 436: Thanks for enumerating the different cases here, but this is fairly convoluted. Why can't this be offloaded to a standalone utility?

torchtune/training/checkpointing/_checkpointer.py
felipemello1: i should add to this function to not copy anything > x MiB, e.g. 200MiB. But i dont think that this should require user input. They have easily cp/rm files. Do you see an easy fix that doesnt greatly increase complexity?

torchtune/training/checkpointing/_checkpointer.py
ebsmothers: I'm not sure I fully follow this, but for (1) couldn't it be done by just e.g. saving `save_adapter_weights_only` as part of the recipe state and pulling it from there?

torchtune/training/checkpointing/_checkpointer.py
felipemello1: it is just an useful map of ckpt file to weights. I guess no use in our current pipeline, but i think it makes sense to have all ckpts behave the same way, and the difference between them is mostly the model format, and **not** the metadata, naming, files, etc.

torchtune/training/checkpointing/_checkpointer.py
felipemello1: are you saying that the code can stay as it is, but have it as an utility? Or would you refactor the code itself?

torchtune/training/checkpointing/_checkpointer.py
felipemello1: are you ok with saving it as part of the recipe_state? that would work.

>I'm not sure I fully follow this,
1. User trains a model that has both lora + finetuning (e.g. vision model)
2. User resumes from ckpt, and forgets to update the ckpt files
3. Since we are looking at the ckpt_dir, we will get the **untrained model**, which is a silent bug.

torchtune/training/checkpointing/_checkpointer.py
ebsmothers: I am saying why not just put all of this (i.e. L436-481) in a utility like `def infer_checkpoint_paths(self, ...)`? Then you can put all this case 1, case 2, etc into a nice clear docstring in a single place. We could think about refactoring too.. the fact that we are calling `_validate_hf_checkpoint_files` like ten different ways is not my favorite

torchtune/training/checkpointing/_checkpointer.py
ebsmothers: Yeah I agree with not increasing complexity here, I think it's OK to leave it like this for now

torchtune/models/convert_weights.py
felipemello1: > peft_model = PeftModel.from_pretrained(model, checkpoint_dir)

I don't think that this works either. The info has to be inside of the adapter_config. Maybe if input_dir == output_dir?

torchtune/training/checkpointing/_checkpointer.py
felipemello1: done

torchtune/training/checkpointing/_checkpointer.py
felipemello1: keeping it for safety

torchtune/training/checkpointing/_checkpointer.py
felipemello1: removed this. when lora=True, it will always get it from ckpt_dir (which is our current state anyway). Need a follow up PR to address it.

torchtune/training/checkpointing/_checkpointer.py
felipemello1: we do

tests/recipes/test_full_finetune_distributed.py
```diff
@@ -169,6 +175,12 @@ def test_training_state_on_resume(
         runpy.run_path(TUNE_PATH, run_name="__main__")
 
         # Resume training
+        epoch_folder = get_largest_iter_folder(tmpdir)
+        epoch_folder_minus_one = f"epoch_{int(epoch_folder.split('_')[-1]) - 1}"
+        suffix = ".safetensors" if ckpt_type == "hf" else ".bin"
+        model_ckpt_fname = (
+            SHARD_FNAME.format(cpt_idx="1".zfill(5), num_shards="1".zfill(5)) + suffix
+        )
```
felipemello1, line 183: test changes are related to finding the files. Before they were hardcoded.

Now we retrieve the epoch folder, get the suffix based on ckpt_type, and create the ckpt_name based on the defined standard.

torchtune/training/checkpointing/_checkpointer.py
```diff
@@ -374,28 +406,57 @@ def __init__(
             Path.joinpath(self._checkpoint_dir, "config.json").read_text()
         )
 
-        # save config.json to output_dir
-        save_config(self._output_dir, self._config)
+        # save all files in input_dir, except model weights and mapping, to output_dir
+        # this is useful to preserve the tokenizer, configs, license, etc.
+        copy_files(
```
felipemello1, line 411: we save everything instead of just the config

torchtune/training/checkpointing/_checkpointer.py
```diff
@@ -374,28 +406,57 @@ def __init__(
...
+            Path.joinpath(self._output_dir, BASE_MODEL_DIRNAME),
+            ignore_suffixes=SUFFIXES_TO_NOT_COPY,
+        )
 
-        # recipe_checkpoint contains the recipe state. This should be available if
-        # resume_from_checkpoint is True
-        self._recipe_checkpoint = None
-        if self._resume_from_checkpoint:
-            if recipe_checkpoint is None:
-                raise ValueError(
```
felipemello1, line 385: moved all this logic to self._checkpoint_paths = get_model_checkpoint_path

torchtune/training/checkpointing/_checkpointer.py
```diff
@@ -136,41 +146,63 @@ def __init__(
         recipe_checkpoint: Optional[str] = None,
         resume_from_checkpoint: bool = False,
     ) -> None:
+
         # Fail fast if ``checkpoint_files`` is invalid
+        # FIXME TODO
```
joecummings, line 151: ?

torchtune/training/checkpointing/_utils.py
joecummings: Yeah wait actually is there a reason why can't use copytree? I'd rather that be maintained by core Python.

torchtune/training/checkpointing/_utils.py
```diff
@@ -277,3 +298,266 @@ def update_state_dict_for_classifier(
...
+    else:
+        # FIXME
+        # TODO: if the model has lora + trained weights, e.g. embeddings,
+        # we will silently not load the trained model, because we load from checkpoint_dir.
+        # We cannot load from output_dir because we always merge the adapter weights into the model
+        input_dir = checkpoint_dir
+
+    checkpoint_paths = validate_checkpoint_files(
+        checkpoint_files,
+        input_dir=checkpoint_dir,
```
joecummings, line 559: I'm confused - you set input_dir with all the above switch statements and then you don't use it here?

felipemello1: -.-`

torchtune/training/checkpointing/_checkpointer.py
felipemello1: changed to "# TODO: support loading more than one file"

torchtune/training/checkpointing/_utils.py
felipemello1: i will update a couple of unit tests to use different output dir

torchtune/training/checkpointing/_utils.py
felipemello1: beyond the suffixes, ii will also have to ignore .cache and .git__

idk, the function gives us some flexibility and its readable. But it seems to be 3x1. 

Can we leave it for the ckpt refactoring?

torchtune/training/checkpointing/_checkpointer.py
ebsmothers: Seems like there are a few follow-ups on this PR.. can we file an issue so we can track the different todos in a single place?

torchtune/training/checkpointing/_utils.py
ebsmothers: Hmm I do like copy_tree and feel like it should be workable. But won't block on it, just include it in the follow-up task (as I mentioned above)

tests/recipes/test_full_finetune_single_device.py
```diff
@@ -196,6 +208,7 @@ def test_training_state_on_resume(self, tmpdir, monkeypatch):
         with pytest.raises(SystemExit, match=""):
             runpy.run_path(TUNE_PATH, run_name="__main__")
 
+        raise NotImplementedError("")
```
ebsmothers, line 211: remove

ebsmothers (approved):
Thanks for getting all of this sorted out! I know this was not an easy one. Once CI is green I think we're good to merge here. One request: can you update the summary to lead with the example usage within vLLM and PEFT? In case anyone is coming to this PR later I think that's what they will want to see. We should also add a section in the readme giving this process explicitly for better visibility.

nouranali:
How would this method affect GPU consumption, having to load two models? While using only 1?

PR #2085, opened 2024-11-27
Fix nightly tests for qat_lora_fintune_distributed

andrewor14:
**Summmary:** These were not caught in CI because the tests require torchao 0.7+. Here is an example of the failure: https://github.com/pytorch/torchtune/actions/runs/12037600747/job/33561415524

**Test Plan:**
(on local devgpu)
```
pytest -m integration_test tests/recipes/test_qat_lora_finetune_distributed.py
```


joecummings (approved):

ebsmothers (approved):
Thank you for the fix!

Issue #1959, opened 2024-11-07
implement activation offloading and opt_in_bwd in knowledge_distillation recipes


felipemello1:
Current knowledge distillation recipes don't have support for activation offloading and opt_in_bwd.

The implementation should be similar to the one in other recipes, like full_finetuning_distributed.

After enabling it in the recipe, it should also be enabled in the configs related to KD.

PRs with reference implementation: 
activation offloading: https://github.com/pytorch/torchtune/pull/1847
opt_in_bwd implementation: https://github.com/pytorch/torchtune/pull/1833

KD recipes:
https://github.com/pytorch/torchtune/blob/main/recipes/knowledge_distillation_single_device.py
https://github.com/pytorch/torchtune/blob/main/recipes/knowledge_distillation_distributed.py

AnuravModak:
Hi @felipemello1 is it available for external contributors? if yes kindly assign it to me will look into it. Thanks in advance!

felipemello1:
@AnuravModak yes! Any issue with "community help wanted" is something that we think would be a great fit for external contributors. Feel free to ask me questions and just submit a PR.

More info on contributing here, but the TLDR:
[main/CONTRIBUTING.md](https://github.com/pytorch/torchtune/blob/main/CONTRIBUTING.md)

1) Fork torchtune: [pytorch/torchtune](https://github.com/pytorch/torchtune)
git clone [github.com](https://github.com/)<YOUR_GITHUB_USER>/torchtune.git

2) install dependencies

cd torchtune
conda create -n torchtune python=3.11
conda activate torchtune
pip install --pre --upgrade torch torchvision torchao --index-url [download.pytorch.org/whl/nightly/cu124](https://download.pytorch.org/whl/nightly/cu124)
pip install -e ".[dev]"
pre-commit install

3) create the PR

rajuptvs:
Hi @felipemello1 , can i give this a go, i have currently implemented the activation offloading and opt_in_bwd for single device knowledge distillation. 

Happy to create a PR ( if not being actively worked on already) and get a review of single device kd,  before going to the distributed.

Thanks!

PR #2088, opened 2024-11-27
Implement activation offloading and opt_in_bwd in knowledge_distillation recipes

AnuravModak:
Towards #1959 .

So this is the first PR for implementing the changes in
KD recipes:
https://github.com/pytorch/torchtune/blob/main/recipes/knowledge_distillation_single_device.py
https://github.com/pytorch/torchtune/blob/main/recipes/knowledge_distillation_distributed.py

### DIFF
--- a/recipes/knowledge_distillation_single_device.py
+++ b/recipes/knowledge_distillation_single_device.py
@@ -31,7 +31,7 @@ from torchtune.modules.peft import (
     validate_missing_and_unexpected_for_lora,
 )
 from torchtune.recipe_interfaces import FTRecipeInterface
-from torchtune.training import DummyProfiler, PROFILER_KEY
+from torchtune.training import DummyProfiler, PROFILER_KEY, OffloadActivations, NoOpManager
 
 from tqdm import tqdm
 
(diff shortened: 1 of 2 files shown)


AnuravModak:
Hi @felipemello1 , As part of this implementation i have made changes only in knowledge_distillation_single_device.py for activation offloading. And enabled testing parameter in `tests/recipes/test_knowledge_distillation_single_device.py`. Let me know if its in the right direction, if Yes will conclude the remaining implementation as well!

joecummings (commented):
This looks great, thanks for doing this!! As part of testing can you output memory usage between activation offloading enabled and disabled as part of the way to confirm that this is doing what we expect it to?

AnuravModak:
Hi @joecummings , 

Added an import for wandb.
Initialized WandB in the beginning of the training loop if self.use_wandb is set to True.
Logged relevant metrics (log_dict) to WandB if self.use_wandb is True.
Conditional Logging:

The logging mechanism is conditioned by the self.use_wandb flag, enabling or disabling WandB integration.
Profiling and Memory:

Added memory tracking and profiling, which can be used for monitoring resource usage during training.


Let me know if it works!

ebsmothers (commented):
Hi @AnuravModak thanks for the PR! A couple suggestions:

1) Regarding @joecummings's previous suggestion to add WandB logs, you should just be able to add WandB as a metric logger from the command line. So you will not need to make any changes to the code to do this. Once you have the results you can paste them in the PR summary. You can see e.g. the test plan in #2139 for an example of this.
2) I think some extra files under `.idea` directory snuck in, can you remove them?
3) Why are the activation offloading changes in `knowledge_distillation_single_device` (specifically L461-L475) commented out?

PR #2095, opened 2024-11-30
Adding code alpaca datasets

insop:
#### Context
- [x] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

#### Changelog
* Adding code alpaca datasets
* [sahil2801/CodeAlpaca-20k](https://huggingface.co/datasets/sahil2801/CodeAlpaca-20k), [iamtarun/code_instructions_120k_alpaca](https://huggingface.co/datasets/iamtarun/code_instructions_120k_alpaca)

#### Test plan

#### UX

### DIFF
--- a/torchtune/datasets/__init__.py
+++ b/torchtune/datasets/__init__.py
@@ -5,7 +5,7 @@
 # LICENSE file in the root directory of this source tree.
 
 from torchtune.datasets import multimodal
-from torchtune.datasets._alpaca import alpaca_cleaned_dataset, alpaca_dataset
+from torchtune.datasets._alpaca import alpaca_evol_codealpaca_v1_dataset, alpaca_code_instruction_120k_dataset, alpaca_code_20k_dataset, alpaca_cleaned_dataset, alpaca_dataset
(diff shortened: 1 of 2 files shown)


insop:
@ebsmothers , PTAL.
Thank you.

ebsmothers (commented):
Hi @insop, thanks for the PR. This change I am not sure about, mainly because it is already possible to support these datasets straight from the command line. 

(Note: I know that this kinda contradicts the fact that we have `alpaca_cleaned_dataset` as a partial right above, but tbh I am not crazy about this either.)

Our builders should be general enough that any of these can be directly plugged in from config or command line without the need for a separate builder. E.g. for the CodeAlpaca-20k dataset just change `tune run ...` to `tune run ... dataset.source=sahil2801/CodeAlpaca-20k`.

insop:
Hi @ebsmothers 

Thank you for your review and the suggestion.
I think it makes sense and `dataset.source=sahil2801/CodeAlpaca-20k` should be a better way.

I am closing this PR.

PR #2098, opened 2024-12-02
Flux Autoencoder

calvinpelletier:
#### Context
- [x] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

#### Changelog
* Implement the Flux autoencoder model
* Function to convert weights from HF's format to ours
* Unit tests

#### Test plan

Parity: 3.36e-5

Speed: ~10% faster

Mnimal testing code:
```python
import numpy as np
import torch
from PIL import Image
from safetensors.torch import load_file

from torchtune.models.flux import flux_autoencoder
from torchtune.models.flux._convert_weights import flux_ae_hf_to_tune

# First, download the AE weights: https://github.com/black-forest-labs/flux/blob/main/src/flux/util.py#L373
WEIGHTS_PATH = "/home/cpelletier/.cache/huggingface/hub/models--black-forest-labs--FLUX.1-dev/snapshots/0ef5fff789c832c5c7f4e127f94c8b54bbcced44/ae.safetensors"


def to_img(x):
    x = x.clamp(-1, 1).float()
    x = x[0].permute(1, 2, 0)
    return Image.fromarray((127.5 * (x + 1.0)).cpu().byte().numpy())


def to_tensor(img):
    img = img.resize((256, 256), Image.Resampling.LANCZOS)
    x = np.array(img, dtype=np.float32)
    x = (x / 127.5) - 1.0
    x = torch.from_numpy(x)
    x = x.permute(2, 0, 1).unsqueeze(0)
    return x.cuda().to(dtype=torch.bfloat16)


ae = flux_autoencoder()
ae.load_state_dict(flux_ae_hf_to_tune(load_file(WEIGHTS_PATH)))
ae = ae.to(dtype=torch.bfloat16, device="cuda").eval().requires_grad_(False)

x = to_tensor(Image.open("/home/cpelletier/tmp.jpg"))
y = ae(x)
to_img(y).save("/home/cpelletier/out.png")
```

#### UX


torchtune/models/flux/_model_builders.py
```diff
@@ -0,0 +1,32 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+from torchtune.models.flux._autoencoder import FluxAutoencoder
+
+
+def flux_autoencoder() -> FluxAutoencoder:
```
calvinpelletier, line 9: Should this be a model builder or component builder? 

Also, should it be named `flux_v1_autoencoder`, anticipating future flux versions? Or just `flux_autoencoder`, anticipating that the autoencoder part probably won't change in the next version?

torchtune/models/flux/_autoencoder.py
```diff
@@ -0,0 +1,352 @@
...
+        """
+        Args:
+            z (Tensor): latent encodings (shape = [bsz, ch_z, latent resolution, latent resolution])
+
+        Returns:
+            Tensor: output images (shape = [bsz, ch_in, img resolution, img resolution])
+        """
+        z = z / self.scale_factor + self.shift_factor
+        h = self.conv_in(z)
+        h = self.mid(h)
```
pbontrager, line 221: Small thing, we follow the pattern of looping through the list of layers here instead of putting them into Sequential. This makes the forward pass a bit more readable and easier to add breakpoints.

torchtune/models/flux/_autoencoder.py
```diff
@@ -0,0 +1,352 @@
...
+        """
+        h = self.conv_in(x)
+        h = self.down(h)
+        h = self.mid(h)
+        h = self.end(h)
+        z = diagonal_gaussian(h)
+        return self.scale_factor * (z - self.shift_factor)
+
+
+class Decoder(nn.Module):
```
pbontrager, line 167: Can we name these FluxDecoder/FluxEncoder?

torchtune/models/flux/_autoencoder.py
```diff
@@ -0,0 +1,352 @@
...
+import torch
+import torch.nn.functional as F
+from torch import nn, Tensor
+
+from torchtune.modules.attention import MultiHeadAttention
+
+# ch = number of channels (size of the channel dimension)
+
+
+class FluxAutoencoder(nn.Module):
```
pbontrager, line 17: This is a bit of overkill now, but I'd have the FluxAutoencoder just take in an Enocder and Decoder as inputs and then define a component builder function to define Encoder + Decoder  and pass them into Autoencoder. This does nothing for us now, but if we want to add lora support for the autoencoder in the future it'll make it much easier.

tests/torchtune/models/flux/test_flux_autoencoder.py
```diff
@@ -0,0 +1,81 @@
...
+    @pytest.fixture
+    def z(self):
+        return torch.randn(BSZ, CH_Z, RES_Z, RES_Z)
+
+    def test_forward(self, model, img):
+        actual = model(img)
+        assert actual.shape == (BSZ, CH_IN, RESOLUTION, RESOLUTION)
+
+        actual = torch.mean(actual, dim=(0, 2, 3))
+        expected = torch.tensor([0.4286, 0.4276, 0.4054])
```
pbontrager, line 59: These are from the small dummy model, did you run these same tests on the full model with weights against the flux codebase?

pbontrager (commented):
This looks really good. A few minor things to followup on and I'll approve

tests/torchtune/models/flux/test_flux_autoencoder.py
calvinpelletier: yes, these are the results from the full test against the flux codebase:

> Parity: 3.36e-5
> 
> Speed: ~10% faster

torchtune/models/flux/_autoencoder.py
calvinpelletier: should we just save this for future work if/when we add lora support for the autoencoder?

pbontrager: I imagine that making the change wouldn’t take very long to do now and then we won’t have to come back to this

PR #2102, opened 2024-12-03
[Draft] Add MFU (Model FLOPs Utilization) logging

### LINKED ISSUE #2100: [Feature Request]: Support logging MFU as a metrics
It would be helpful if torchtune can log MFU as a metric to measure the real GPU utilization.

xingyaoww:
This PR implements MFU (Model FLOPs Utilization) logging as requested in #2100.

### Changes

1. Added MFU utility module (`torchtune/utils/mfu.py`) with functions to:
   - Calculate theoretical peak FLOPS for GPU
   - Calculate actual MFU percentage
   - Calculate model FLOPs for one forward pass

2. Modified `LoRAFinetuneRecipeDistributed` to:
   - Calculate model FLOPs after initialization
   - Log MFU alongside other metrics in training loop

3. Added fvcore as a dev dependency for FLOPs calculation

The MFU metric is now logged with the same frequency as other metrics (controlled by `log_every_n_steps`) and is available in all supported logging backends (terminal, disk, WandB, TensorBoard).

Closes #2100

----

This PR is contributed by AI Software Engineer [OpenHands](https://github.com/All-Hands-AI/OpenHands) -- task solving trajectories: https://www.all-hands.dev/share?share_id=d029a05b7ea79fd4d96c76c16e2f412db7c648a244a78afe46d16ae15f12b0fa

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -58,6 +58,7 @@ dev = [
     "urllib3<2.0.0",
     "wandb",
     "expecttest",
+    "fvcore",  # For MFU calculation
 ]
 
 [tool.setuptools.dynamic]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -58,7 +58,7 @@ dev = [
     "urllib3<2.0.0",
     "wandb",
     "expecttest",
-    "fvcore",  # For MFU calculation
+
 ]
 
 [tool.setuptools.dynamic]
--- a/recipes/lora_finetune_distributed.py
+++ b/recipes/lora_finetune_distributed.py
@@ -886,8 +886,7 @@ class LoRAFinetuneRecipeDistributed(FTRecipeInterface):
                             "lr": self._optimizer.param_groups[0]["lr"],
                             "tokens_per_second_per_gpu": tokens_per_second,
                             "mfu": mfu_value,
-                            "flops": actual_flops,
-                            "tflops": actual_flops / 1e12,  # Convert to TFLOPS for readability
+                            "tflops": actual_flops / 1e12  # Convert to TFLOPS for readability
                         }
                         if self._log_peak_memory_stats:
                             log_dict.update(
--- a/recipes/lora_finetune_distributed.py
+++ b/recipes/lora_finetune_distributed.py
@@ -871,17 +871,22 @@ class LoRAFinetuneRecipeDistributed(FTRecipeInterface):
                     ):
                         time_per_step = time.perf_counter() - t0
                         tokens_per_second = num_tokens / (time_per_step * world_size)
+                        # Calculate MFU and actual FLOPS
+                        mfu_value, actual_flops = mfu.calculate_mfu_and_flops(
+                            model_flops=self._model_flops,
+                            batch_size=len(batch),
+                            step_time=time_per_step,
+                            world_size=world_size,
+                            device=self._device
+                        )
+                        
                         log_dict = {
                             "loss": loss_to_log,
                             "lr": self._optimizer.param_groups[0]["lr"],
                             "tokens_per_second_per_gpu": tokens_per_second,
-                            "mfu": mfu.calculate_mfu(
-                                model_flops=self._model_flops,
-                                batch_size=len(batch),
-                                step_time=time_per_step,
-                                world_size=world_size,
-                                device=self._device
-                            ),
+                            "mfu": mfu_value,
+                            "flops": actual_flops,
+                            "tflops": actual_flops / 1e12,  # Convert to TFLOPS for readability
                         }
                         if self._log_peak_memory_stats:
                             log_dict.update(
--- a/recipes/lora_finetune_distributed.py
+++ b/recipes/lora_finetune_distributed.py
@@ -550,11 +550,12 @@ class LoRAFinetuneRecipeDistributed(FTRecipeInterface):
 
         # Calculate model FLOPs for MFU tracking
         if self._is_rank_zero:
-            # Use a small batch size and sequence length for FLOPs calculation
-            self._model_flops = mfu.get_model_flops(
-                model=model,
-                input_shape=(1, 32),  # (batch_size, seq_len)
-                device=self._device
+            # Calculate model FLOPs for MFU tracking
+            self._model_flops = mfu.get_transformer_flops(
+                hidden_size=cfg_model.hidden_size,
+                intermediate_size=cfg_model.hidden_size * 4,  # Standard transformer uses 4x
+                num_layers=cfg_model.num_layers,
+                seq_len=cfg_model.seq_length
             )
             utils.log_rank_zero(
                 log,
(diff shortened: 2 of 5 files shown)


pyproject.toml
```diff
@@ -58,6 +58,7 @@ dev = [
     "urllib3<2.0.0",
     "wandb",
     "expecttest",
+    "fvcore",  # For MFU calculation
```
felipemello1, line 61: n00b question: is MFU calculation simple enough that it could be an utility, instead of a new dependency?

xingyaoww: Good Q! Let me poke around OpenHands and see if we can have a solution without additional dependency

felipemello1: torchtitan also computes MFU. They are a great reference implementation. Maybe it is worth taking a look at how they do it. 

https://github.com/pytorch/torchtitan/blob/9a4eebe1c9a692027f0c8ce88f7caf89af3dd616/train.py#L70

felipemello1: edit: I just took a look at it and I dont like it too much! haha. Maybe there are other alternatives. If nothing simple, it might be ok to have a dependency, if its not a heavy one.

xingyaoww: Thanks for the pointer!

I actually found one from megatron-LM: https://github.com/NVIDIA/Megatron-LM/blob/main/megatron/training/one_logger_utils.py

Just send off to OpenHands and see how it looks like at the end

image

felipemello1:
when the PR is ready for review, it will be nice to see the numbers vs the library you originally had added to the requirements. That way we can be sure that the utilities are working well at least compared to that reference.

Thanks for the PR, i am sure that the users will appreciate it!

recipes/lora_finetune_distributed.py
```diff
@@ -855,11 +871,22 @@ def train(self) -> None:
...
                         time_per_step = time.perf_counter() - t0
+                        tokens_per_second = num_tokens / (time_per_step * world_size)
+                        # Calculate MFU and actual FLOPS
+                        mfu_value, actual_flops = mfu.calculate_mfu_and_flops(
+                            model_flops=self._model_flops,
+                            batch_size=len(batch),
+                            step_time=time_per_step,
+                            world_size=world_size,
+                            device=self._device
+                        )
```
felipemello1, line 882: aside from time_per_step, all other variables seem to be constant. We can probably pass them in the initi and reduce footprint here.

recipes/lora_finetune_distributed.py
```diff
@@ -546,6 +548,20 @@ def _setup_model(
...
         torch.distributed.barrier()
 
+        # Calculate model FLOPs for MFU tracking
+        if self._is_rank_zero:
+            # Calculate model FLOPs for MFU tracking
+            self._model_flops = mfu.get_transformer_flops(
+                hidden_size=cfg_model.hidden_size,
+                intermediate_size=cfg_model.hidden_size * 4,  # Standard transformer uses 4x
+                num_layers=cfg_model.num_layers,
+                seq_len=cfg_model.seq_length
```
felipemello1, line 558: i this the batch sequence length? The cfg.model.seq_length is just to adjust the RoPE scaling

recipes/lora_finetune_distributed.py
```diff
@@ -546,6 +548,20 @@ def _setup_model(
         # synchronize before training begins
         torch.distributed.barrier()
 
+        # Calculate model FLOPs for MFU tracking
+        if self._is_rank_zero:
+            # Calculate model FLOPs for MFU tracking
+            self._model_flops = mfu.get_transformer_flops(
+                hidden_size=cfg_model.hidden_size,
+                intermediate_size=cfg_model.hidden_size * 4,  # Standard transformer uses 4x
+                num_layers=cfg_model.num_layers,
```
felipemello1, line 557: is this a fair generalization, e.g. hidden_size*4? I am not sure how this would work for more complex models, like llama 3.2 vision, that has encoder+decoder. 

We are also a bit afraid of the added complexity/ number of lines of code that mfu may introduce. If we can reduce it, it will be easier to approve the pr.

But i understand that here its just one line + multiple arguments.

recipes/lora_finetune_distributed.py
```diff
@@ -855,11 +871,22 @@ def train(self) -> None:
...
+                        )
+                        
                         log_dict = {
                             "loss": loss_to_log,
                             "lr": self._optimizer.param_groups[0]["lr"],
-                            "tokens_per_second_per_gpu": num_tokens
-                            / (time_per_step * world_size),
+                            "tokens_per_second_per_gpu": tokens_per_second,
+                            "mfu": mfu_value,
+                            "tflops": actual_flops / 1e12  # Convert to TFLOPS for readability
```
felipemello1, line 889: do we need both?

felipemello1 (commented):
hey @xingyaoww , i appreciate the draft and all of the testing! Let's try to iron out the API a bit first to avoid more work on your end. It doesnt need to be a working example.

We chatted internally a bit  about this PR and the overall feeling is that the metric could be a nice addition to the library, as long as it doesn't take much space in the recipe. I left a few comments in that direction. Let me know if you have any thoughts.

Thank you!!

xingyaoww:
Thanks @felipemello1! Sorry it has been a busy week and this got slide under my list of TODOs.

I give it a second thought and feel calculation accurate FLOPS without tracing is probably a bit hard. Maybe a second / easy way we can do is to use this approximation (C=6NBS):
image

from https://arxiv.org/pdf/2001.08361

This will likely just use one additional function and won't introduce too much code. The downside is that, it is only appliable to Transformer architecture & not super accurate, but probably gives an OK estimate of MFU, WDYT?

felipemello1:
> Thanks for the feedback @felipemello1! Sorry it has been a busy week and this got slide under my list of TODOs.
> 
> I give it a second thought and feel calculation accurate FLOPS without tracing is probably a bit hard. Maybe a second / easy way we can do is to use this approximation (C=6NBS): image
> 
> from [arxiv.org/pdf/2001.08361](https://arxiv.org/pdf/2001.08361)
> 
> This will likely just use one additional function and won't introduce too much code. The downside is that, it is only appliable to Transformer architecture & not super accurate, but probably gives an OK estimate of MFU, WDYT?

Hey @xingyaoww, thanks for sharing it! You mentioned "I give it a second thought and feel calculation accurate FLOPS without tracing is probably a bit hard", we do enable tracing in torchtune. Maybe this would be sufficient for users interested in MFU?

I am afraid that not every model will fit the heuristics being used, and it may do more harm than good, specially with the introduction with MoE, multimodal, diffusion etc. It doesn't seem to be as straight forward.

If tracing is a good alternative, I think that we can point researches in this direction instead. What are your thoughts, as an user interested in this metric?

xingyaoww:
Yeah, sounds good to me! Tracing should be a better alternative, and for this heuristic, having token per seconds is probably sufficient for estimation.

Issue #2070, opened 2024-11-25
Checkpoint management in torchtune


joecummings:
### Current solution
Our current solution is based on the format in which the checkpoint arrives to us. See below for our three checkpointer classes. These handle the loading of pretrained weights and saving of trained weights during and after training. In addition, they also handle saving and loading the state of the recipe for use when resuming training. Notably, these classes expect pointers to local files and only write out to local files.

```python
# Load and save a model from Hugging Face w/ option to save back
# into safetensors
FullModelHFCheckpointer()

# Load and save a model in the "xlformers" (Meta) format for Llama checkpoints
FullModelMetaCheckpointer()

# Internal format of the checkpoint that exactly matches torchtune's
# implementation of models
FullModelTorchTuneCheckpointer()
```

These checkpointers are instantiated in all of our recipes and users interact with them through the load_checkpoint() and save_checkpoint() APIs. You can read more in depth about our current checkpointer design in the [deep dives section](https://pytorch.org/torchtune/0.4/deep_dives/checkpointer.html#) of our documentation.

#### Some background
When we were starting designing the checkpointing functionality in torchtune, we knew we wanted to support more checkpoints than those that came directly from the Hugging Face Hub. Actually, it was a requirement b/c at that point, Meta hadn't easily integrated with the Hub for their Llama checkpoints so users had to download those weights separately. Since there was wide support in the OSS community for the Llama models and we didn't want to give Meta's legal team a heart attack by hosting our own weights, we needed to support loading directly from files in the original format.

The decision to bundle the recipe state with the checkpointer largely followed the intuition that this information for restarting a finetuning run belonged in the same location as the model weights that were to be used. Therefore, bundling the recipe state with the checkpointer class made it logically simple in that there was only one object writing out information (weights, state, configs) that could be used later.

> Aside: what's the deal with the 'FullModel' prefix? Again when we first started, we had wanted to support directly through the checkpointer interfaces a way to load both adapters from LoRA and full models. These checkpointers were  only supposed to be for the full model. Since we never actually got around to loading the adapters and instead made it very simple to load your trained adapters into an existing Hugging Face model + merged all weights back in before saving, this name was a little overly-descriptive.

### How is the design not meeting the user's needs?
[...]

Issue #1107, opened 2024-06-21
Save intermediate checkpoints during training


l3utterfly:
Is there any way to save intermediate checkpoints during training?

Sometimes my training may fail during the middle due to external reasons, it will be helpful to save every N steps so I can continue where I left off

joecummings:
Hey @l3utterfly - thanks for creating the issue!

This is definitely something that's on our minds, but is not currently available in the repo *yet*. Right now, if I'm worried that my training might take a long time and therefore is susceptible to failures, I split the dataset up into smaller pieces using the [built-in datasets % operator](https://huggingface.co/docs/datasets/en/loading#slice-splits).

```
dataset:
  _component_: torchtune.datasets.alpaca_cleaned_dataset
  split: train[:25%]
``` 

This will train on 25% of a given dataset. Then, I take that checkpoint and pass it into the next training round with `resume_from_checkpoint: True`. Rinse and repeat 4 times to complete training on the entire dataset.

I fully acknowledge this isn't the most efficient way to handle this and we are looking into adding this feature, but the above method should at least unblock you until now.

pbontrager:
I want to add to Joe's response, another option to unblock yourself would be to copy the recipe you're using and copy and paste the save checkpoint line into the training for loop with an if statement that allows it to save every n steps.

l3utterfly:
@joecummings I am trying with the `split: train [:25%]` method.

My first 25% is done training, I see the PT files written to my output dir. However, I do not see a `recipe_state.pt` file generated, causing me to be unable to resume.

How can I generate this file?

joecummings:
> @joecummings I am trying with the `split: train [:25%]` method.
> 
> My first 25% is done training, I see the PT files written to my output dir. However, I do not see a `recipe_state.pt` file generated, causing me to be unable to resume.
> 
> How can I generate this file?

Ahh yes, our default recipes only generate that file for intermediate checkpoints (https://github.com/pytorch/torchtune/blob/15c918d65d79e03abcbd0c5e94d2b116bd368412/torchtune/utils/_checkpointing/_checkpointer.py#L590), e.g for epoch 1 of 2. So you can either:

1) Specify extra epochs if you want to use the default recipe and then kill training after a single epoch or
2) `tune cp` the recipe you want and modify the code so that `save_checkpoint(..., intermediate_checkpoint=True)`

[...]

PR #2105, opened 2024-12-03
[RFC] Step-based checkpointing in torchtune

joecummings:
## Enabling step-based checkpointing in torchtune

Original context: #2070 

#### What are we currently doing?

We currently only checkpoint at epoch boundaries. That means a fine-tuning run has to iterate through **all data** in a dataset before saving a checkpoint. That's a problem when GPUs (especially interconnected GPUs) can fail frequently, losses can diverge, and datasets keep getting larger and larger.

We provide a tiny amount of flexibility by allowing the user to specify `max_steps_per_epoch`, so they can short-circuit the epoch and save sooner. In addition, it's always possible to split a dataset into chunks and train over them independently, resuming from training to simulate a larger training run.

Both of these "hacks" are not ideal and we've had users continually asking if they can control checkpointing based on number of training steps. (#988, #1107)

#### What does step-based checkpointing look like for the user?

I think the best way to do this would to show an example. Let's take our [Llama3 8B single device full fine-tuning recipe](./recipes/configs/llama3/8B_full_finetune_single_device.yaml), which utilizes the Alpaca dataset. The Alpaca dataset has ~52k samples. Using a batch size of 2 and a gradient accumulation of 16 steps, we can estimate around 1625 steps in this training run. Let's save a checkpointing every 500 steps!

From the config, we can specify:
```yaml
save_every_n_train_steps: 500
```

And in our output directory, we can expect to see something like this:
```
output_dir/
	step_500/
		llama3_8b_single_device.yaml
		config.json
		model-0000-of-0003.safetensors
		...
	step_1000/
		...
	step_1500/
		...
	step_1625/
		...
```

> You'll see that at the end of the training loop, a final checkpoint is saved regardless of how many steps have passed since the last checkpoint was saved.

At this point you might be saying: @joecummings, do you think memory grows on trees? Do you think we all drive Bugattis and smash up Grace Hopper machines for fun? Each Llama3 8B model is roughly 16 GB of memory and we've saved 4 copies of that in addition to the base model we used. That's 80 GB just for checkpoints! Not even to mention if we wanted to save the optimizer states, too...

Introducing:

```yaml
keep_last_n_checkpoints: 1
```

This param will prune all the checkpoints except for the last N specified, leaving you with just the checkpoints you're interested in:
```
output_dir/
	step_1625/
		llama3_8b_single_device.yaml
		config.json
		model-0000-of-0003.safetensors
		...
```

**What about the concept of epochs?**

The concept of epochs will stay as a way to control how long training runs, as will the possibility to shorten training using `max_steps_per_epoch`; however, checkpointing will be entirely handled by a specification of steps.

**Will this slow down training?**

Great question! Checkpointing can take a long time, especially if saving the optimizer state for resuming training at a later date. For single device recipes, this likely isn't a huge issue, but for distributed recipes where the state dict needs to be collected on rank zero before saving, this can be verrrrrrry slow so anything that increases the frequency of checkpointing will increase the time it takes for training to complete. There are two ways to mitigate this:

1) Specify a longer period so you save checkpoints less frequently
2) Look into using DCP checkpointer for any intermediate checkpoints, which drastically reduces the time it takes to save. See #2006 for more information on this.

#### What changes need to be made in code?

**In the recipe:**

```python
num_steps = 0
	steps_run_since_last_saved_ckpt = 0
		# Model forward
		...
		if doing_bwd_pass:
			steps_run_since_last_saved_ckpt += 1
			num_steps += 1

		if steps_run_since_last_saved_ckpt == save_ckpt_every_n_steps:
			self.save_checkpoint(ckpt, step=num_steps)
			steps_run_since_last_saved_ckpt = 0

# One final save
self.save_checkpoint(ckpt, step=num_steps)
```

**And in the checkpointer:**

```python
def save_checkpoint(ckpt, step):
	# Prune old checkpoints if needed
	checkpoints_already_saved = get_all_prev_checkpoint_paths(self.output_dir)
	if len(checkpoints_already_saved) >= self.keep_last_n_checkpoints:
		prune_old_checkpoints(checkpoints_already_saved)

	# Create new folder for path
	new_ckpt_path = output_dir / step_{step}
	new_ckpt_path.mkdir(exist_ok=False)
	# Save new checkpoint
	torch.save(ckpt, new_ckpt_path / "model.bin")
```

#### Inspiration from relevant repositories:
* [TorchTitan](https://github.com/pytorch/torchtitan)
* [TorchTNT](https://github.com/pytorch/tnt)

### DIFF
--- a/torchtune/training/checkpointing/step_based_ckpt.rfc.md
+++ b/torchtune/training/checkpointing/step_based_ckpt.rfc.md
@@ -1,0 +1,108 @@
+## Enabling step-based checkpointing in torchtune
+
+#### What are we currently doing?
+
+We currently only checkpoint at epoch boundaries. That means a fine-tuning run has to iterate through **all data** in a dataset before saving a checkpoint. That's a problem when GPUs (especially interconnected GPUs) can fail frequently, losses can diverge, and datasets keep getting larger and larger.
+
+We provide a tiny amount of flexibility by allowing the user to specify `max_steps_per_epoch`, so they can short-circuit the epoch and save sooner. In addition, it's always possible to split a dataset into chunks and train over them independently, resuming from training to simulate a larger training run.
+
+Both of these "hacks" are not ideal and we've had users continually asking if they can control checkpointing based on number of training steps. (#988, #1107)
+
+#### What does step-based checkpointing look like for the user?
+
+I think the best way to do this would to show an example. Let's take our [Llama3 8B single device full fine-tuning recipe](./recipes/configs/llama3/8B_full_finetune_single_device.yaml), which utilizes the Alpaca dataset. The Alpaca dataset has ~52k samples. Using a batch size of 2 and a gradient accumulation of 16 steps, we can estimate around 1625 steps in this training run. Let's save a checkpointing every 500 steps!
+
+From the config, we can specify:
+```yaml
+save_every_n_train_steps: 500
+```
+
+And in our output directory, we can expect to see something like this:
+```
+output_dir/
+	step_500/
+		llama3_8b_single_device.yaml
+		config.json
+		model-0000-of-0003.safetensors
+		...
+	step_1000/
+		...
+	step_1500/
+		...
+	step_1625/
+		...
+```
+
+> You'll see that at the end of the training loop, a final checkpoint is saved regardless of how many steps have passed since the last checkpoint was saved.
+
+At this point you might be saying: @joecummings, do you think memory grows on trees? Do you think we all drive Bugattis and smash up Grace Hopper machines for fun? Each Llama3 8B model is roughly 16 GB of memory and we've saved 4 copies of that in addition to the base model we used. That's 80 GB just for checkpoints! Not even to mention if we wanted to save the optimizer states, too...
+
+Introducing:
+
+```yaml
+keep_last_n_checkpoints: 1
+```
+
+This param will prune all the checkpoints except for the last N specified, leaving you with just the checkpoints you're interested in:
+```
+output_dir/
+	step_1625/
+		llama3_8b_single_device.yaml
+		config.json
+		model-0000-of-0003.safetensors
+		...
+```
+
+**What about the concept of epochs?**
+
+The concept of epochs will stay as a way to control how long training runs, as will the possibility to shorten training using `max_steps_per_epoch`; however, checkpointing will be entirely handled by a specification of steps.
+
+**Will this slow down training?**
+
+Great question! Checkpointing can take a long time, especially if saving the optimizer state for resuming training at a later date. For single device recipes, this likely isn't a huge issue, but for distributed recipes where the state dict needs to be collected on rank zero before saving, this can be verrrrrrry slow so anything that increases the frequency of checkpointing will increase the time it takes for training to complete. There are two ways to mitigate this:
+
+1) Specify a longer period so you save checkpoints less frequently
+2) Look into using DCP checkpointer for any intermediate checkpoints, which drastically reduces the time it takes to save. See #2006 for more information on this.
+
+#### What changes need to be made in code?
+
+**In the recipe:**
+
+```python
+num_steps = 0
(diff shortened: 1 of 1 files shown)


felipemello1:
It makes sense to me. 

I am not 100% sure how i feel about keeping epochs AND steps, since defaulting to one might simplify things. Internally, we only have the concept of steps, and it is up to the user to use bsz + len(dataset) to calculate how many steps == 1 epoch (which is a bit annoying).

Can we have a config showing the args for streaming dataset and checkpointing? If it looks simple, i like keeping both

regarding `keep_last_n_checkpoints=1`, my only question is if there's a safer alternative than deleting files. Probably not...

calvinpelletier:
Yay step-based checkpointing! Some thoughts:

1. I second Felipe's comment about dropping support for epoch-based checkpointing. Our code will be cleaner and simpler if our whole ecosystem of checkpointing/validating/logging/etc is only step-based. The users who prefer epochs can easily calculate the number of steps in an epoch. 

2. Regarding keeping the last n checkpoints, should we also allow keeping the best checkpoint (once we add support for validation)? When I run training jobs for large models, I don't want to save too many checkpoint and run out of disk space, but I also don't want to have training collapse overnight and find out that all the good checkpoints have been already deleted. Since we don't have validation/evaluation during training yet, we could choose the best based on the training loss for now.

3. There's this annoying trade-off when choosing the steps intervals for checkpointing/validating/etc. Large intervals are nice because then you don't slow training down too much. But small intervals are nice because you get quick feedback that everything is working correctly and the extra data points are important early on when things are changing pretty quickly. Personally, I use a logarithmic step-based checkpointing/validating/etc. schedule, where the interval is small during early training and gets longer as training goes on. I'm not sure if other people like this too, but it's trivial to implement and supporting it wouldn't complicate our code very much.

joecummings:
> I am not 100% sure how i feel about keeping epochs AND steps, since defaulting to one might simplify things. Internally, we only have the concept of steps, and it is up to the user to use bsz + len(dataset) to calculate how many steps == 1 epoch (which is a bit annoying).

I totally get this. I might; however, be annoying and punt this to a different discussion b/c right now I am keeping epochs as a method to determine how long to train the model, whereas steps only determine when to checkpoint. I'd rather give more time to properly educate users if we do end up taking away epochs entirely. 

> regarding keep_last_n_checkpoints=1, my only question is if there's a safer alternative than deleting files. Probably not...

Yeah, this makes me a little uncomfortable, too, but it does seem to be the standard in other training code like Titan and TNT.

joecummings:
@calvinpelletier 

> Regarding keeping the last n checkpoints, should we also allow keeping the best checkpoint (once we add support for validation)? When I run training jobs for large models, I don't want to save too many checkpoint and run out of disk space, but I also don't want to have training collapse overnight and find out that all the good checkpoints have been already deleted. Since we don't have validation/evaluation during training yet, we could choose the best based on the training loss for now.

Yes, I think this is a natural extension to this work. If you look at the TNT code, they actual have some scaffolding to do this. I'd be happy to add that as a follow-up or even have this be community contributed. 

> There's this annoying trade-off when choosing the steps intervals for checkpointing/validating/etc. Large intervals are nice because then you don't slow training down too much. But small intervals are nice because you get quick feedback that everything is working correctly and the extra data points are important early on when things are changing pretty quickly. Personally, I use a logarithmic step-based checkpointing/validating/etc. schedule, where the interval is small during early training and gets longer as training goes on. I'm not sure if other people like this too, but it's trivial to implement and supporting it wouldn't complicate our code very much.

This makes a ton of sense and how I would want to incorporate this information now is to make sure it's easy to extend to such a use case when the time comes.

tginart:
Just wondering if this PR will be merged in soon? @calvinpelletier @felipemello1 @joecummings 

This is much needed functionality! Thank you!!!

Btw, quick question, in terms of the ```max_steps_per_epoch``` hack ---> does the data order get reshuffled each epoch?

Issue #1977, opened 2024-11-09
Gradient clipping doesn't work with FSDP CPU offloading


acisseJZhong:
I am running the full finetune distributed recipe, when setting `clip_grad_norm: 1.0` and `fsdp_cpu_offload: True`, it raises error 
`RuntimeError: No backend type associated with device type cpu`

Full error stack trace: 
```
[rank2]: Traceback (most recent call last):
[rank2]:   File "/torchtune/recipes/full_finetune_distributed.py", line 847, in <module>
[rank2]:     sys.exit(recipe_main())
[rank2]:              ^^^^^^^^^^^^^
[rank2]:   File "/torchtune/torchtune/config/_parse.py", line 99, in wrapper
[rank2]:     sys.exit(recipe_main(conf))
[rank2]:              ^^^^^^^^^^^^^^^^^
[rank2]:   File "/torchtune/recipes/full_finetune_distributed.py", line 842, in recipe_main
[rank2]:     recipe.train()
[rank2]:   File "/torchtune/recipes/full_finetune_distributed.py", line 740, in train
[rank2]:     grad_norm = torch.nn.utils.clip_grad_norm_(
[rank2]:                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank2]:   File "/home/jessicazhong/.conda/envs/torchtune/lib/python3.11/site-packages/torch/nn/utils/clip_grad.py", line 30, in _no_grad_wrapper
[rank2]:     return func(*args, **kwargs)
[rank2]:            ^^^^^^^^^^^^^^^^^^^^^
[rank2]:   File "/home/jessicazhong/.conda/envs/torchtune/lib/python3.11/site-packages/torch/nn/utils/clip_grad.py", line 105, in clip_grad_norm_
[rank2]:     clip_coef = max_norm / (total_norm + 1e-6)
[rank2]:                 ~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~
[rank2]:   File "/home/jessicazhong/.conda/envs/torchtune/lib/python3.11/site-packages/torch/_tensor.py", line 39, in wrapped
[rank2]:     return f(*args, **kwargs)
[rank2]:            ^^^^^^^^^^^^^^^^^^
[rank2]:   File "/home/jessicazhong/.conda/envs/torchtune/lib/python3.11/site-packages/torch/_tensor.py", line 1075, in __rdiv__
[rank2]:     return self.reciprocal() * other
[...]

PR #2108, opened 2024-12-03
Set gloo process group for FSDP with CPU offload

ebsmothers:
Addresses #1977.

As discussed in the issue (see [this comment](https://github.com/pytorch/torchtune/issues/1977#issuecomment-2512906526)), FSDP's implementation of gradient clipping uses [_NormPartial](https://github.com/pytorch/pytorch/blob/64d44a39a10a3e4b7abc85ffadd831ace3227460/torch/distributed/tensor/_ops/_math_ops.py#L56), which requires comms primitives (specifically `all_reduce`). This means that when running with CPU offloading we need to initialize the gloo process group to calculate the grad norm for DTensors on CPU. For simplicity this PR enables it whenever CPU offloading is enabled regardless of gradient clipping.

Test plan:

Added a test case for gradient clipping + CPU offload to `test_full_finetune_distributed.py`.

```
pytest -m integration_test tests/recipes/test_full_finetune_distributed.py -k 'test_loss'
...
========= 3 passed, 1 deselected in 43.54s ========
```

### DIFF
--- a/recipes/full_finetune_distributed.py
+++ b/recipes/full_finetune_distributed.py
@@ -946,7 +946,6 @@ def recipe_main(cfg: DictConfig) -> None:
             "Distributed finetune recipe should be run via a distributed launcher."
             "If using tune CLI, please specify --nnodes 1 and --nproc_per_node [num_gpus]"
         )
-
     process_group = "gloo" if cfg.device == "cpu" else "nccl"
     if cfg.get("fsdp_cpu_offload", False):
         # Utilize all available CPU cores for intra-op parallelism. This provides ~2x
--- a/recipes/full_finetune_distributed.py
+++ b/recipes/full_finetune_distributed.py
@@ -946,11 +946,14 @@ def recipe_main(cfg: DictConfig) -> None:
             "Distributed finetune recipe should be run via a distributed launcher."
             "If using tune CLI, please specify --nnodes 1 and --nproc_per_node [num_gpus]"
         )
-    init_process_group(backend="gloo" if cfg.device == "cpu" else "nccl")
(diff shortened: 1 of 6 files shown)


recipes/full_finetune_distributed.py
```diff
@@ -946,11 +946,13 @@ def recipe_main(cfg: DictConfig) -> None:
             "Distributed finetune recipe should be run via a distributed launcher."
             "If using tune CLI, please specify --nnodes 1 and --nproc_per_node [num_gpus]"
         )
-    init_process_group(backend="gloo" if cfg.device == "cpu" else "nccl")
+    process_group = "gloo" if cfg.device == "cpu" else "nccl"
     if cfg.get("fsdp_cpu_offload", False):
         # Utilize all available CPU cores for intra-op parallelism. This provides ~2x
         # speed up when benchmarking fused AdamW on CPU
         training.set_torch_num_threads()
+        process_group = "cuda:nccl,cpu:gloo"
```
felipemello1, line 954: n00b question: why not just make this the default every time, instead of `"gloo" if cfg.device == "cpu" else "nccl"`

felipemello1: image

ebsmothers: I actually wasn't sure myself why we do this. Well [now I know](https://github.com/pytorch/torchtune/pull/482#discussion_r1529635443). So yes, I think we can take your suggestion here

felipemello1 (approved):
this seems to be low risk, since it was tested in the dcp already. Approving to unblock.

PR #2128, opened 2024-12-06
update configs

felipemello1:
update 70B configs (and some others)

### DIFF
--- a/recipes/configs/llama3_3/70B_full.yaml
+++ b/recipes/configs/llama3_3/70B_full.yaml
@@ -16,6 +16,8 @@
 # This config is only tested on an 8xA100 machine.
 #
 
+output_dir: /tmp/torchtune/llama3_3_70B/full # /tmp may be deleted by your system. Change it to your preference.
+
 # Tokenizer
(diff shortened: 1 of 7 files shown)


recipes/configs/llama3_3/70B_full.yaml
```diff
@@ -107,8 +109,7 @@ dtype: bf16
 # Logging
 metric_logger:
   _component_: torchtune.training.metric_logging.DiskLogger
-  log_dir: ${output_dir}
-output_dir: /tmp/full-llama3_3-finetune
+  log_dir: ${output_dir}/logs
```
acisseJZhong, line 112: I think this is not consistent with other configs? Maybe update all of the rest configs?

acisseJZhong (approved):
lgtm, except the metrics log directory need to be consistent with the rest of configs?

recipes/configs/llama3_3/70B_full.yaml
felipemello1: The others were updated yesterday :)

PR #2133, opened 2024-12-09
guard ckpt imports

felipemello1:
#### Context
- [ ] add a new feature
- [x] fix a bug
- [ ] other (please add here)

If the user tries to use torchtune without torchvision, it errors, because:

```
from torchtune.training.checkpointing._checkpointer import ...
from torchtune.models.clip._convert_weights import clip_text_hf_to_tune
```

which tries to import torchvision

Test:

new env without torchvision
```
tune run --nproc_per_node 4 full_finetune_distributed --config llama3_2/1B_full
```

+

CLI

### DIFF
--- a/torchtune/training/checkpointing/_checkpointer.py
+++ b/torchtune/training/checkpointing/_checkpointer.py
@@ -15,10 +15,7 @@ from safetensors.torch import save_file
 
 from torchtune import training
 from torchtune.models import convert_weights
-from torchtune.models.clip._convert_weights import clip_text_hf_to_tune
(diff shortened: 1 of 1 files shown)


ebsmothers (approved):

felipemello1:
> Should torchvision be an optional dependency? We specify it in our install instructions [pytorch/torchtune#install-stable-release](https://github.com/pytorch/torchtune?tab=readme-ov-file#install-stable-release)

I had the same question. Joe said "Should not need it for anything non-VLM", and Philip agreed too.

As a rule of thumb, if we can avoid importing torchvision for non vision related tasks, its probably better

SalmanMohammadi (approved):

PR #2138, opened 2024-12-10
Migrate distributed state dict API

mori360:
#### Context
- [ ] add a new feature
- [ ] fix a bug
- [x] other (please add here)

Migrate distributed state dict APIs from torch.distributed.

#### Changelog

Switch to distributed state dict APIs from torch.distributed.

- `load_from_full_model_state_dict` <- `set_model_state_dict`
- `gather_cpu_state_dict` <- `get_model_state_dict`
- `load_from_full_optimizer_state_dict` <- `set_optimizer_state_dict`
- `get_full_optimizer_state_dict` <- `get_optimizer_state_dict`

To align the inputs, add model input to `get_full_optimizer_state_dict` and `load_from_full_optimizer_state_dict`.
Change the sharded_sd input for `gather_cpu_state_dict` to model.

TODO:
nf4tensor are kept the same, remain as future work

#### Test plan

`pytest tests/torchtune/training/test_distributed.py`
`pytest tests -m integration_test`
(`early_exit_finetune_distributed` and `knowledge_distillation_distributed` are not tested in the CI test)
`tune run --nnodes 1 --nproc_per_node 4 dev/early_exit_finetune_distributed --config recipes/dev/7B_full_early_exit.yaml`
`tune run --nnodes 1 --nproc_per_node 2 knowledge_distillation_distributed --config llama3_2/8B_to_1B_KD_lora_distributed`

We compare the running with the previous API and the new API, loss are the same in initial loading and resume from checkpoint.

We also draw the memory traces, results show that the new API won't cost mote memory peak comapred with the current ones.
Screenshot 2025-01-02 at 1 10 18 PM


torchtune/training/_distributed.py
```diff
@@ -166,64 +173,89 @@ def load_from_full_model_state_dict(
     - `is_rank_zero` matters if only rank 0 pass in non-empty `full_sd` and
        we need to broadcast from rank 0
     """
+    has_nf4 = any(
```
mori360, line 176: is_rank_zero is not used in load_from_full_model_state_dict, shall we remove it

torchtune/training/_distributed.py
```diff
@@ -166,64 +173,89 @@ def load_from_full_model_state_dict(
...
+                shard_rank = cast(
+                    torch.distributed.ProcessGroup, mesh.get_group(shard_mesh_dim)
+                ).rank()
+                chunk = list(torch.chunk(full_tensor, shard_world_size, dim=0))[
+                    shard_rank
+                ]
+                sharded_param = full_tensor.new_zeros(chunk.size())
+                sharded_param[: chunk.size(0)].copy_(chunk)
+
+                # TODO: change to from_local API (need to add view support for NF4)
```
joecummings, line 217: How can we get view support for NF4?

cc @andrewor14

mori360: Thank you for the review, we currently skip the NF4 tensor part and plan to support NF4 in the next quarter.

andrewor14: Looks like there's already [view support](https://github.com/pytorch/ao/blob/3bac9056dd051c6232bdefed932877ca3bfe62c0/torchao/dtypes/nf4tensor.py#L272) for NF4Tensor? What's the error you're getting?

also cc @drisspg @weifengpy

SalmanMohammadi: I brought this up with @ebsmothers and @gau-nernst in Discord. We thought that we needed to do anything else here, it should just be safe to just switch to `from_local`.

mori360: Thank you for the comments, shall I switch to `from_local` in this pr or get it with the other nf4 tensor support in the next pr?

recipes/knowledge_distillation_distributed.py
```diff
@@ -705,13 +703,14 @@ def save_checkpoint(self, epoch: int) -> None:
         # To prevent GPU memory from spiking during checkpoint save,
         # we consolidate the full model and optim state dicts on CPU for rank 0
         cpu_state_dict = training.gather_cpu_state_dict(
-            self._model.state_dict(),
+            self._model,
```
ebsmothers, line 706: I just realized we are doing things differently here than in the other recipes.. seems to me like we could move the call to `get_adapter_state_dict` up before calling `gather_cpu_state_dict`, then you could make the same changes you did in e.g. `lora_finetune_distributed.py` (remove the call to `get_adapter_state_dict` and instead just pass `trainable_only=self._save_adapter_weights_only` to `gather_cpu_state_dict`). Lmk if that makes sense to you

recipes/lora_finetune_distributed.py
```diff
@@ -656,14 +655,11 @@ def save_checkpoint(
...
-        state_dict = self._model.state_dict()
-        if self._save_adapter_weights_only:
-            state_dict = get_adapter_state_dict(state_dict, device=None)
-
         cpu_state_dict = training.gather_cpu_state_dict(
-            state_dict,
+            self._model,
             self._is_rank_zero,
             device=self._device,
+            trainable_only=self._save_adapter_weights_only,
```
ebsmothers, line 662: Also one other thing we will have to be aware of is that in general it may not always be the case that trainable params == adapter params. This holds true today, but especially for multimodal models we need to be careful because some people may want to e.g. do LoRA finetuning on the image encoder and full finetuning on the text decoder. This was disabled in #2150 but we may want to add it back later and in that case this would be misleading. So I think `trainable_only` is potentially a misnomer and it may be best to rename `adapter_weights_only` or something like that.

recipes/lora_finetune_distributed_multi_dataset.py
```diff
@@ -500,7 +499,6 @@ def _setup_model(
             model,
             base_model_state_dict,
             self._device,
-            self._is_rank_zero,
```
ebsmothers, line 503: (Commenting here for further down in the file but) is there a reason you didn't also update `save_checkpoint` in this recipe? (We don't yet have a test for it so probably didn't get caught by CI)

recipes/dev/early_exit_finetune_distributed.py
```diff
@@ -556,7 +556,6 @@ def _setup_model(
             model,
             model_state_dict,
             self._device,
-            self._is_rank_zero,
```
ebsmothers, line 559: Do we also need to update `_setup_optimizer` in this recipe?

torchtune/training/_distributed.py
```diff
@@ -251,43 +360,67 @@ def _gather_nf4_tensor(sharded_param: nn.Parameter) -> nn.Parameter:
...
+    trainable_only: bool = False,
 ) -> Dict[str, Any]:
     """
     Converting sharded state dict into a full state dict on CPU
     Returning non-empty result only on rank0 to avoid peaking CPU memory
+    TODO: add support for NF4Tensor
 
     Args:
-        sharded_sd (Dict[str, DTensor]): Sharded state dict of DTensors
+        model (FSDPModule): Model to generate fqn for cpu_state_dict
```
ebsmothers, line 374: nit but I don't think most people know what "fqn" means, might write something more descriptive here

torchtune/training/_distributed.py
```diff
@@ -251,43 +360,67 @@ def _gather_nf4_tensor(sharded_param: nn.Parameter) -> nn.Parameter:
...
-    sharded_sd: Dict[str, DTensor],  # noqa
+    model: "FSDPModule",  # noqa
     is_rank_zero: bool,
     device: Optional[torch.device] = None,
+    trainable_only: bool = False,
 ) -> Dict[str, Any]:
     """
     Converting sharded state dict into a full state dict on CPU
     Returning non-empty result only on rank0 to avoid peaking CPU memory
+    TODO: add support for NF4Tensor
```
ebsmothers, line 371: Can we add more details here so that it's clear to someone who's reading the code? Something like "If the model does not contain any NF4 tensors, we directly use distributed state dict APIs. Otherwise, we need to manually gather any NF4 tensors until all-gather is supported in the NF4Tensor subclass"

torchtune/training/_distributed.py
```diff
@@ -166,64 +179,160 @@ def load_from_full_model_state_dict(
...
-            )
+    # There are some changes at `set_model_state_dict` to adjust multiple devices from local_state in TorchTune,
+    # keey version check until PyTorch changes are on stable.
+    if _USE_DISTRIBUTED_STATE_DICT_API:
+        has_nf4 = any(
+            hasattr(param, "_local_tensor")
+            and isinstance(param._local_tensor, NF4Tensor)
+            for param in model.parameters()
+        )
+        # has_nf4 = True
```
ebsmothers, line 190: remove?

torchtune/training/_distributed.py
```diff
@@ -166,64 +179,160 @@ def load_from_full_model_state_dict(
...
+                    sharded_tensor = distribute_tensor(
+                        full_tensor,
+                        sharded_meta_param.device_mesh,
+                        sharded_meta_param.placements,
+                    )
+                if cpu_offload:
+                    sharded_tensor = sharded_tensor.cpu()
+                sharded_sd[param_name] = nn.Parameter(sharded_tensor)
+            # choose `assign=True` since we cannot call `copy_` on meta tensor
+            return model.load_state_dict(sharded_sd, strict=strict, assign=True)
```
ebsmothers, line 257: Is this a duplicate of L274-L335? I think this function is already complicated enough, if we can just use a single if/else branch to consolidate these that'd be preferable

torchtune/training/_distributed.py
```diff
@@ -166,64 +179,160 @@ def load_from_full_model_state_dict(
...
-                    tensor_meta=TensorMeta(
-                        shape=sharded_meta_param.size(),
-                        dtype=sharded_meta_param.dtype,
-                        stride=sharded_meta_param.stride(),
-                    ),
-                ),
-                requires_grad=sharded_meta_param.requires_grad,
-            )
+    # There are some changes at `set_model_state_dict` to adjust multiple devices from local_state in TorchTune,
+    # keey version check until PyTorch changes are on stable.
```
ebsmothers, line 183: I don't understand this comment

torchtune/training/_distributed.py
ebsmothers: Yes if possible it'd be great to move to `from_local` here assuming everything works. Imo the more that we can clean this function up the better, as is it has gotten quite unwieldy

ebsmothers (commented):
Thanks for your patience! Left a bunch of comments, please let me know if anything is unclear. One request is to also manually run `lora_finetune_distributed_multi_dataset.py` and `early_exit_finetune_distributed.py` recipes as they do not currently have tests in our CI. Happy to provide any pointers here if you need.

recipes/dev/early_exit_finetune_distributed.py
mori360: `_setup_optimizer` does not call `load_from_full_model_state_dict`, did not update with removal of `self._is_rank_zero` in `_setup_optimizer`

recipes/lora_finetune_distributed_multi_dataset.py
mori360: The PR here only removes `self._is_rank_zero` from `training.load_from_full_model_state_dict`, which is not called at `save_checkpoint`, would add the test of `lora_finetune_distributed_multi_dataset.py` and `early_exit_finetune_distributed.py` later

torchtune/training/_distributed.py
mori360: Switch the structure here to `if _USE_DISTRIBUTED_STATE_DICT_API and not has_nf4` and `else`

torchtune/training/_distributed.py
mori360: Comments updated on why we have the pytorch version check here

recipes/knowledge_distillation_distributed.py
mori360: Discussed offline, we can take `get_adapter_state_dict` cleaning among recipes as future work in the next PR

torchtune/training/_distributed.py
```diff
@@ -166,64 +179,95 @@ def load_from_full_model_state_dict(
...
+            full_state_dict=True,
+            broadcast_from_rank0=True,
+            strict=strict,
+            cpu_offload=cpu_offload,
+        )
+        return set_model_state_dict(
+            model=model, model_state_dict=full_sd, options=options
+        )
+    else:
+        meta_sharded_sd = model.state_dict()
```
joecummings, line 207: Same comment as above - move out of the branches.

torchtune/training/_distributed.py
```diff
@@ -166,64 +179,95 @@ def load_from_full_model_state_dict(
...
+        sharded_sd = {}
+        for param_name, full_tensor in full_sd.items():
+            sharded_meta_param = meta_sharded_sd.get(param_name)
+            full_tensor = full_tensor.to(sharded_meta_param.dtype).to(device)
```
joecummings, line 211: sigh: we shouldn't have two consecutive `to` calls. You don't need to fix here - I can do it.

torchtune/training/_distributed.py
```diff
@@ -166,64 +179,95 @@ def load_from_full_model_state_dict(
...
+            if hasattr(sharded_meta_param, "_local_tensor") and isinstance(
+                sharded_meta_param._local_tensor, NF4Tensor
+            ):
+                block_size = sharded_meta_param._local_tensor.block_size
+                scaler_block_size = sharded_meta_param._local_tensor.scaler_block_size
+                full_tensor = to_nf4(
```
joecummings, line 217: if we assume "full" means plain tensor which is the claim above, then this is - in fact - not a "full" tensor?

torchtune/training/_distributed.py
joecummings: Bump here.

torchtune/training/_distributed.py
```diff
@@ -251,43 +295,69 @@ def _gather_nf4_tensor(sharded_param: nn.Parameter) -> nn.Parameter:
...
+                else:
+                    # Gather DTensor
+                    param = param.full_tensor()
+            if isinstance(param, NF4Tensor):
+                param = param.to(param.dtype)
+            if is_rank_zero:
+                cpu_state_dict[param_name] = param.cpu()
+            torch.distributed.barrier()
+        return cpu_state_dict
+    else:
```
joecummings, line 344: Shouldn't there be a check here to make sure that the distributed APIs are available?

torchtune/training/_distributed.py
```diff
@@ -298,89 +368,74 @@ def get_full_optimizer_state_dict(
...
-                full_tensor = sharded_tensor
-            if is_rank_zero:
-                group_state[attr] = full_tensor.cpu()
-            else:
-                del full_tensor
-        if is_rank_zero:
-            full_state[group_id] = group_state
-        else:
-            del group_state
+    options = StateDictOptions(
```
joecummings, line 371: I love how clean this is now, but shouldn't we check to make sure Distributed APIs are available?

joecummings (commented):
I want to get this in b/c it's a huge improvement over our current functionality. I left a few notes just as reminders to myself, but definitely need to confirm that we are checking that these APIs are available everywhere they are used and then add in a little more explanation in places. 

Thanks @mori360

torchtune/training/_distributed.py
```diff
@@ -166,64 +179,95 @@ def load_from_full_model_state_dict(
...
+                scaler_block_size = sharded_meta_param._local_tensor.scaler_block_size
+                full_tensor = to_nf4(
+                    full_tensor,
+                    block_size=block_size,
+                    scaler_block_size=scaler_block_size,
+                )
+                # replicating logic from `_fsdp_param.py`` `_init_sharded_param`
+                # otherwise `distribute_tensor(DTensor(local=NF4))`
+                # requires dispatching `c10d.scatter_``
+                # long-term solution is `swap_tensor`
```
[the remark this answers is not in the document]
mori360: We plan to support nf4tensor as future work in Q1.

torchtune/training/_distributed.py
```diff
@@ -166,64 +179,95 @@ def load_from_full_model_state_dict(
...
+    # - `set_model_state_dict` with the `cpu_offload` option
+    # - Multiple devices in local state dict
+    # - Relative optimizations for improved memory performance
+    # Please keep the version check `_USE_DISTRIBUTED_STATE_DICT_API` until these changes are
+    # released in the PyTorch stable version.
+    has_nf4 = any(
+        hasattr(param, "_local_tensor") and isinstance(param._local_tensor, NF4Tensor)
+        for param in model.parameters()
+    )
+    if _USE_DISTRIBUTED_STATE_DICT_API and not has_nf4:
```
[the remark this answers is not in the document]
mori360: We plan to process NF4 and others in a more general way to avoid `if has_nf4:`

torchtune/training/_distributed.py
mori360: The current distributed api could support with the ideal performance, so don't need to check distributed APIs  availability here

torchtune/training/_distributed.py
mori360: Same as above, the changes as Distributed API are to support `load_from_full_model_state_dict`, the other 3 API could work with the ideal performance with the current Distributed API without nightly version landed at Dec 20

joecummings (approved):
Awesome work! Thanks for working with us to address all comments :)

PR #2150, opened 2024-12-11
[Fix] Llama 3.2 Vision decoder_trainable flag fixed

pbontrager:
#### Context
- [ ] add a new feature
- [x] fix a bug
- [ ] other (please add here)

decoder_trainable flag was not hooked up right. This PR makes it so that setting it to lora or frozen toggles the decoder between frozen weights and lora correctly. This also disables mixed lora and full finetuning training. Setting x_trainable=full for a lora model will now throw an error. This will be re-enabled once checkpointing can properly support this.

#### Test plan

`tune run lora_finetune_single_device --config llama3_2_vision/11B_lora_single_device metric_logger=torchtune.training.metric_logging.WandBLogger model.decoder_trainable=frozen`

Screenshot 2024-12-11 at 11 08 03 AM

`tune run lora_finetune_single_device --config llama3_2_vision/11B_lora_single_device metric_logger=torchtune.training.metric_logging.WandBLogger model.decoder_trainable=lora`

Screenshot 2024-12-11 at 11 21 35 AM

For further testing I inspected the model parameters with the two options and manually verified that the correct decoder parameters existed or not and were trainable or not.

### DIFF
--- a/torchtune/models/llama3_2_vision/_model_builders.py
+++ b/torchtune/models/llama3_2_vision/_model_builders.py
@@ -169,7 +169,7 @@ def lora_llama3_2_vision_11b(
     fusion_type = LoRATrainable(fusion_trainable.lower())
     assert LoRATrainable.FULL not in [
         decoder_type,
-        decoder_type,
+        encoder_type,
         fusion_type,
     ], "We've temporarily removed support for mixed LoRA + Full Finetuning yet. Please don't use the 'full' option and use llama3_2_vision_11b if you need full finetuning"
     encoder = lora_llama3_2_vision_encoder(
@@ -332,7 +332,7 @@ def lora_llama3_2_vision_90b(
     fusion_type = LoRATrainable(fusion_trainable.lower())
     assert LoRATrainable.FULL not in [
         decoder_type,
-        decoder_type,
+        encoder_type,
         fusion_type,
     ], "We've temporarily removed support for mixed LoRA + Full Finetuning yet. Please don't use the 'full' option and use llama3_2_vision_90b if you need full finetuning"
     encoder = lora_llama3_2_vision_encoder(
--- a/torchtune/models/llama3_2_vision/_model_builders.py
+++ b/torchtune/models/llama3_2_vision/_model_builders.py
@@ -167,6 +167,11 @@ def lora_llama3_2_vision_11b(
     decoder_type = LoRATrainable(decoder_trainable.lower())
     encoder_type = LoRATrainable(encoder_trainable.lower())
     fusion_type = LoRATrainable(fusion_trainable.lower())
+    assert LoRATrainable.FULL not in [
+        decoder_type,
+        decoder_type,
+        fusion_type,
+    ], "We've temporarily removed support for mixed LoRA + Full Finetuning yet. Please don't use the 'full' option and use llama3_2_vision_11b if you need full finetuning"
     encoder = lora_llama3_2_vision_encoder(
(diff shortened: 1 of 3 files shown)


torchtune/models/llama3_2_vision/_model_builders.py
```diff
@@ -167,6 +167,11 @@ def lora_llama3_2_vision_11b(
     decoder_type = LoRATrainable(decoder_trainable.lower())
     encoder_type = LoRATrainable(encoder_trainable.lower())
     fusion_type = LoRATrainable(fusion_trainable.lower())
+    assert LoRATrainable.FULL not in [
+        decoder_type,
+        decoder_type,
```
ebsmothers, line 172:
```suggestion
        encoder_type,
```

torchtune/models/llama3_2_vision/_model_builders.py
```diff
@@ -325,6 +330,11 @@ def lora_llama3_2_vision_90b(
...
```
ebsmothers, line 335:
```suggestion
        encoder_type,
```

torchtune/models/llama3_2_vision/_model_builders.py
```diff
@@ -167,6 +167,11 @@ def lora_llama3_2_vision_11b(
...
+        fusion_type,
+    ], "We've temporarily removed support for mixed LoRA + Full Finetuning yet. Please don't use the 'full' option and use llama3_2_vision_11b if you need full finetuning"
```
ebsmothers, line 174: I would just say "X is not currently supported" instead of "We've temporarily removed support for X" (and same comment on the error message further down)

ebsmothers (commented):
> For further testing I inspected the model parameters with the two options and manually verified that the correct decoder parameters existed or not and were trainable or not.

Sounds worthy of a unit test to me 👀

torchtune/models/llama3_2_vision/_model_builders.py
pbontrager: I worded it this way since the option is still there, but we don't let you use it. Also in case someone is already using it.

pbontrager:
Unit test added

felipemello1:
> Setting x_trainable=full for a lora model will now throw an error. This will be re-enabled once checkpointing can properly support this.

Just to confirm, it does support it, we just don't support resuming from ckpt. Is that correct?

felipemello1:
For `model.decoder_trainable=True`, should we test it and confirm that it still works? I have yet to look into the code. Did you compare the memory between these 2 wandb runs?

felipemello1 (dismissed):

torchtune/models/llama3_2_vision/_model_builders.py
ebsmothers: > We've temporarily removed support for mixed LoRA + Full Finetuning yet.

😕

ebsmothers (approved):
Thanks for adding the unit test! I still would update the error message (it reads like a mishmash of two competing sentences). Otherwise this looks good to me

torchtune/models/llama3_2_vision/_model_builders.py
joecummings: Yeah this could be worded in a kinder manner. Feels aggressive. Ask ChatGPT to reword or smthing.

joecummings (approved):

felipemello1:
i will merge it so i can cut the release. @pbontrager , please write a new PR with a new error msg, and i can cherry pick it

PR #2155, opened 2024-12-12
Move ``get_world_size_and_rank`` to utils

### LINKED ISSUE #2151: Make distributed available outside of torchtune.training
Modules in torchtune.data can't import get_world_size_and_rank from torchtune.training._distributed because of circular imports. Making this available in torchtune._distributed instead would fix this issue
https://github.com/pytorch/torchtune/pull/1929#discussion_r1879199687_

joecummings:
#### Context
- [ ] add a new feature
- [x] fix a bug
- [ ] other (please add here)

Closes #2151 

#### Changelog
* Move ``get_world_size_and_rank`` from torchtune/training to torchtune/utils. **Why does this make sense?** B/c it's more of a utility anyways and can be used outside of the training loop
* Update all tests and call sites accordingly.
* Deprecate old ``get_world_size_and_rank``

#### Test plan

#### UX

### DIFF
--- a/torchtune/utils/_device.py
+++ b/torchtune/utils/_device.py
@@ -6,7 +6,7 @@
 
 import os
 from enum import Enum
-from typing import Optional
+from typing import Optional, Tuple
 
 import torch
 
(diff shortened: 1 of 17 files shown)


torchtune/utils/_device.py
```diff
@@ -21,6 +21,19 @@
     BlockMask = torch.Tensor
 
 
+def get_world_size_and_rank() -> Tuple[int, int]:
```
ebsmothers, line 24: I understand the goal of this and agree with moving this function down into utils. But I also wonder whether device.py is really the best place for it (I know we already have `_get_local_rank` in here, and I know that having separate `utils/distributed.py` and `training/distributed.py` is probably confusing)

joecummings: I'm in favor of keeping it in here.

ebsmothers (approved):

PR #2176, opened 2024-12-18
Update DPO Max Seq Len

pbontrager:
#### Context
- [ ] add a new feature
- [X] fix a bug
- [ ] other (please add here)

Default value for 8.1 updated to match llama2. Without truncating the datasets long sequences cause OOM.

#### Changelog
* llama3.1 8B dpo lora max seq len in config

#### Test plan

`tune run lora_dpo_single_device --config llama3_1/8B_lora_dpo_single_device`
`tune run --nproc_per_node 2 lora_dpo_distributed --config llama3_1/8B_lora_dpo`

### DIFF
--- a/recipes/configs/llama3_1/8B_lora_dpo_single_device.yaml
+++ b/recipes/configs/llama3_1/8B_lora_dpo_single_device.yaml
@@ -31,7 +31,7 @@ model:
 tokenizer:
   _component_: torchtune.models.llama3.llama3_tokenizer
   path: /tmp/Meta-Llama-3.1-8B-Instruct/original/tokenizer.model
-  max_seq_len: null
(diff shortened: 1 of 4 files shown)


recipes/configs/llama3_1/8B_lora_dpo_single_device.yaml
```diff
@@ -31,7 +31,7 @@ model:
 tokenizer:
   _component_: torchtune.models.llama3.llama3_tokenizer
   path: /tmp/Meta-Llama-3.1-8B-Instruct/original/tokenizer.model
-  max_seq_len: null
```
joecummings, line 34: Shouldn't this be like 8K? B/c the default dataset has an average seq len > 7K

PR #2182, opened 2024-12-19
change saving logic

felipemello1:
#### Context
- [ ] add a new feature
- x] fix a bug
- [ ] other (please add here)

#### Changelog
- Instead of saving model files to output_dir/base_model, we save them to output_dir/epoch_{}. That way, every epoch is ready for inference, without requiring an extra step. In the future, we can design this extra step better, e.g. 'tune prepare_for_inference args'
- Have a file size limit of 100MiB. All expected files are configs + tokenizer, so they shouldn't be larger than that.

#### Test plan
tree -a /tmp/torchtune/llama3_2_3B/full_single_device
image

testing with HF
image

resuming from ckpt
image

### DIFF
--- a/torchtune/training/checkpointing/_utils.py
+++ b/torchtune/training/checkpointing/_utils.py
@@ -382,7 +382,7 @@ def copy_files(
             # Check the file size
             if os.path.getsize(src_file) > max_file_size:
                 print(
-                    f"Skipping copying {src_file} to {output_dir} as it exceeds the size limit of 100 MiB."
+                    f"Skipping copying {src_file} to {output_dir} as it exceeds the size limit of {max_file_size_mb} MiB."
                 )
                 continue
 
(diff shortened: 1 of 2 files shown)


torchtune/training/checkpointing/_utils.py
```diff
@@ -330,6 +329,7 @@ def copy_files(
     output_dir: Union[str, Path],
     *,
     ignore_suffixes: Optional[List[str]] = None,
+    max_file_size_mb: int = 100,
```
ebsmothers, line 332: Is there actually a way for someone to modify these values from the CLI/config? Seems like no, right?

felipemello1: no, but i dont see it being an issue. It shouldnt happen, and If users really want to come a 100MB file to every epoch, they can do manually.

ebsmothers (approved):
Stamping to unblock. Please make sure to test that after the changes to the directory structure both resume from checkpoint and HF from_pretrained functionality are unaffected.

Issue #2190, opened 2024-12-20
Model request. Phi4


krammnic:
As was noticed by @gau-nernst. It is Phi3, but with tiktoken tokenizer and full attention instead of sliding one. Probably willl open PR adding it.

PR #2197, opened 2024-12-21
Add Phi4

krammnic:
#### Context
- [X] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

#### Changelog
* #2190 

#### Test plan

#### UX


krammnic:
Should we wait until Phi-4 will be on HF?

krammnic:
We require on changes in only tokenizer actually, as we already used full Attention in Phi3 without sliding one.

krammnic:
I assume that I need to do run with Phi4

krammnic:
@joecummings It seems to me that you haven't gone to holidays) Maybe you can give me some comments about this PR?

joecummings:
> @joecummings It seems to me that you haven't gone to holidays) Maybe you can give me some comments about this PR?

Haha yes I'm still (somewhat) here. I asked and it looks like the Phi4 team is planning on fixing some license issues with Hugging Face and should have the model on the Hub soon. So eventually the true test will be to grab the official model from Hugging Face and do a forward pass; however, if you want to iron out any potential discrepancies right away, I'd just grab one of the unofficial uploads like [this](https://huggingface.co/matteogeniaccio/phi-4) for your testing. 

Happy holidays to you @krammnic - been a pleasure working with you on torchtune this year!

krammnic:
@joecummings Thanks for the comments!) Will do some runs with this then

ebsmothers:
Hi @krammnic just checking in on this PR. I saw the model is on Hugging Face (as of yesterday I believe). Have you done a parity check with their model? And is this ready for review? If so let me know and we can take a look

krammnic:
> Hi @krammnic just checking in on this PR. I saw the model is on Hugging Face (as of yesterday I believe). Have you done a parity check with their model? And is this ready for review? If so let me know and we can take a look

Hi, will run tests today and will ping you when it will be ready for review!

recipes/configs/phi4/evaluation.yaml
```diff
@@ -0,0 +1,44 @@
+# Config for EleutherEvalRecipe in eleuther_eval.py
+#
+# To launch, run the following command:
+#    tune run eleuther_eval --config phi3/evaluation
```
insop, line 4: /phi3/phi4

recipes/configs/phi4/evaluation.yaml
```diff
@@ -0,0 +1,44 @@
...
+output_dir: ./ # Not needed
+
+# Model Arguments
+model:
+  _component_: torchtune.models.phi3.phi3_mini
+
+# Checkpointer
+checkpointer:
+  _component_: torchtune.training.FullModelHFCheckpointer
+  checkpoint_dir: /tmp/Phi-3-mini-4k-instruct
```
insop, line 15: /Phi-3/Phi-4

recipes/configs/phi4/evaluation.yaml
```diff
@@ -0,0 +1,44 @@
...
+  checkpoint_files: [
+    model-00001-of-00002.safetensors,
+    model-00002-of-00002.safetensors
+  ]
+  recipe_checkpoint: null
+  output_dir: ${output_dir}
+  model_type: PHI3_MINI
```
insop, line 22: /PHI3_MINI/PHI4_MINI

recipes/configs/phi4/evaluation.yaml
```diff
@@ -0,0 +1,44 @@
...
+resume_from_checkpoint: False
+
+# Tokenizer
+tokenizer:
+  _component_: torchtune.models.phi3.phi3_mini_tokenizer
+  path: /tmp/Phi-3-mini-4k-instruct/tokenizer.model
```
insop, line 28: /torchtune.models.phi3.phi3_mini_tokenizer/torchtune.models.phi4.phi4_mini_tokenizer

insop: /Phi-3/Phi-4

recipes/configs/phi4/mini_full.yaml
```diff
@@ -0,0 +1,105 @@
+# Config for multi-device full finetuning in full_finetune_distributed.py
+# using a Phi3 Mini 4K Instruct
```
insop, line 2: /Phi3/Phi4

recipes/configs/phi4/mini_full_low_memory.yaml
```diff
@@ -0,0 +1,106 @@
+# Config for single device full finetuning in full_finetune_single_device.py
+# using a Phi3 Mini 4K Instruct
```
insop, line 2: search all Phi3 and replace with Phi4

krammnic: Thank you for the comments! I'm still fixing naming actually. Will ping you when it will be ready!

krammnic:
Probably forward is now working (I get OOM cause my cards a busy with some experiments). There is pretty weird point that I had to set `num_heads = 20` which is twice less then real num_heads (I assume that it is feature of torchtune?). Also, there is some inconsistency with naming. Official description is: `Phi-4 small language model` but probably we can't name it "small".

krammnic:
Hardcoding like this fixes the issue:
```
 q_proj=nn.Linear(embed_dim, 2560, bias=False),
 k_proj=nn.Linear(embed_dim, 2560, bias=False),
 v_proj=nn.Linear(embed_dim, 2560, bias=False),
```

Probably we should revise formulas especially for phi4.

krammnic:
Nit: For all configs should change tokenizer field

krammnic:
Getting `RuntimeError: shape '[2, 308, 40, 128]' is invalid for input of size 1576960
  0%|   ` Probably from same reason

krammnic:
For `num_heads=40`, `num_kv_heads=10`, `embed_dim=5120`. Let's calculate:

`head_dim = 5120 / 40 = 128`

Already here:

```
 q_proj=nn.Linear(embed_dim, num_heads * head_dim, bias=False),
 k_proj=nn.Linear(embed_dim, num_kv_heads * head_dim, bias=False),
 v_proj=nn.Linear(embed_dim, num_kv_heads * head_dim, bias=False),
```

Already have a problem here as it will be not 2560 in all cases, but 5120, 1280, 1280. Assume that we "hardcoded" in a way that I have shown earlier. But then we get same problem here:

```
 q_per_kv = self.num_heads // self.num_kv_heads
 q = q.view(b, s_x, self.num_kv_heads * q_per_kv, self.head_dim)
```

Error:
`RuntimeError: shape '[2, 308, 40, 128]' is invalid for input of size 1576960`

Part of config.json for reference:

```
  "hidden_size": 5120,
  "initializer_range": 0.02,
  "intermediate_size": 17920,
  "max_position_embeddings": 16384,
  "model_type": "phi3",
  "num_attention_heads": 40,
  "num_hidden_layers": 40,
  "num_key_value_heads": 10,
  "original_max_position_embeddings": 16384,
```

So, the product should be twice less. Am I missing something? (I hope I have not miscalculated). Something weird is behind this problem. Will try to work out it asap. @ebsmothers I'm not really sure if it fixable with touching phi3 model or creating separate model for phi4.

krammnic:
Oh, and also I assume that we first of all need to speak about this... #2212

recipes/configs/phi3/evaluation.yaml
felipemello1, line 1: folder is phi3, but args are phi4

recipes/configs/phi4/evaluation.yaml
felipemello1, line 1: It seems that you made a copy from phi3, but made the changes in phi3/evaluation, instead of here

recipes/configs/phi4/mini_full.yaml
felipemello1, line 1: n00b question: Is "mini" the right nomenclature? Or do they have a family of model sizes like phi4_7b, phi4_13B, etc?

recipes/configs/phi4/mini_full.yaml
```diff
@@ -0,0 +1,109 @@
...
+    model-00001-of-00006.safetensors,
+    model-00002-of-00006.safetensors,
+    model-00003-of-00006.safetensors,
+    model-00004-of-00006.safetensors,
+    model-00005-of-00006.safetensors,
+    model-00006-of-00006.safetensors,
+  ]
+  recipe_checkpoint: null
+  output_dir: ${output_dir}
+  model_type: PHI3_MINI
```
felipemello1, line 46: n00b question: Are there are differences between PHI3 and PHI4? Even if there arent, should we update the model_type for clarity? I believe that this is used in the checkpointer to map the HF format to torchtune format.

recipes/configs/phi4/mini_full_low_memory.yaml
felipemello1, line 1: I think that this was already the naming convention for Phi3, but we should probably add "single_device" to the config name.

recipes/configs/phi4/mini_lora.yaml
```diff
@@ -0,0 +1,119 @@
...
+# this run:
+#   tune download microsoft/phi-4 --output-dir /tmp/phi-4 --hf-token <HF_TOKEN>
+#
+# To launch on 2 devices, run the following command from root:
+#   tune run --nnodes 1 --nproc_per_node 2 lora_finetune_distributed --config phi4/mini_lora
+#
+# You can add specific overrides through the command line. For example
+# to override the checkpointer directory while launching training
+# you can run:
+#   tune run --nnodes 1 --nproc_per_node 2 lora_finetune_distributed --config phi4/mini_lora checkpointer.checkpoint_dir=<YOUR_CHECKPOINT_DIR>
```
felipemello1, line 14: we can probably remove `--nnodes 1`. We dont usually add it to distributed configs. Same goes to the other dist configs

felipemello1 (commented):
just left some comments, not a review yet. Will come back to it.

recipes/configs/phi3/evaluation.yaml
krammnic: Yep, good point

recipes/configs/phi4/mini_full.yaml
krammnic: According to tech report there is difference in tokenizer and in attention in such way that it is not touching us. But some observations that I made upper might get us to different conclusion

recipes/configs/phi4/mini_full.yaml
krammnic: Pretty arguing moment, in the description of Phi4 it is "mini model" in real life it is not

recipes/configs/phi4/mini_lora.yaml
krammnic: Good catch!

torchtune/models/phi4/_tokenizer.py
```diff
@@ -0,0 +1,315 @@
...
+        messages: List[Message],
+        *,
+        add_eos: bool = False,
+        ignore_system_prompt: bool = False,
+    ) -> Tuple[List[int], List[bool]]:
+        r"""Tokenize a list of messages one at a time then concatenate them,
+        returning a list of tokens and a list of masks.
+
+        Example:
+            >>> tokenizer = Phi3MiniTokenizer(tokenizer_path, max_seq_len)
```
felipemello1, line 153:
```suggestion
            >>> tokenizer = Phi4MiniTokenizer(tokenizer_path, max_seq_len)
```

torchtune/models/phi4/_tokenizer.py
```diff
@@ -0,0 +1,315 @@
...
+            >>> messages = [
+                Message(role="system", content="system message\n", masked=True),
+                Message(role="user", content="user prompt\n", masked=True),
+                Message(role="assistant", content="assistant response\n"),
+            ]
+
+            >>> # tokenize_messages encodes messages separately and concats
+            >>> tokenizer.tokenize_messages(messages)[0]
+            [1, 1788, 2643, 13, 1792, 9508, 13, 465, 22137, 2933, 2]
```
felipemello1, line 162: is this example still accurate for phi4?

torchtune/models/phi4/_tokenizer.py
```diff
@@ -0,0 +1,315 @@
...
+            # Prepend BOS on start of new turns
+            if start_of_turn:
+                tokenized_messages.append(self.bos_id)
+                mask.append(message.masked)
+
+            # Add special tokens
+            if message.role == "user":
+                tokenized_messages.append(self.special_tokens["<|im_start|>"])
+                encoded = self.encode(
+                    "system",
```
felipemello1, line 209: n00b question: is it supposed to be "system" for all of them?

torchtune/models/phi4/_tokenizer.py
```diff
@@ -0,0 +1,315 @@
...
+        prompt_template: Optional[PromptTemplate] = None,
+    ):
+        self.special_tokens = (
+            special_tokens if special_tokens is not None else PHI4_SPECIAL_TOKENS
+        )
+
+        # Use custom EOS, BOS and pad ids instead of TikToken's
+        self.eos_id = self.special_tokens["<|endoftext|>"]
+        self.bos_id = self.special_tokens["<|endoftext|>"]
+        self.pad_id = self.special_tokens["<|endoftext|>"]
```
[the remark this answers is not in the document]
RdoubleA: This is straight from the HF repo. We would need some confirmation from the Phi team that these are incorrect

torchtune/models/phi4/_tokenizer.py
```diff
@@ -0,0 +1,315 @@
...
+
+        # During generation, stop when eos_id is encountered
+        self.stop_tokens = [self.eos_id]
+
+        self.max_seq_len = max_seq_len
+
+        self.prompt_template = prompt_template
+
+        self.tt_model = TikTokenBaseTokenizer(
```
RdoubleA, line 91: GPT2Tokenizer is probably closer to TikToken than SentencePiece (someone who's more knowledgeable can correct me), but I'm not sure if this will create the correct token map. You would need to test against the HF version of the tokenizer on the same text.

This is probably highlighting our issue with converting from HF tokenizers as mentioned in #2212. We'll need to think of a better long-term solution here.

torchtune/models/phi4/_tokenizer.py
```diff
@@ -0,0 +1,315 @@
...
+            if ignore_system_prompt and message.role == "system":
+                continue
+
+            # Prepend BOS on start of new turns
+            if start_of_turn:
+                tokenized_messages.append(self.bos_id)
+                mask.append(message.masked)
+
+            # Add special tokens
+            if message.role == "user":
```
RdoubleA, line 206: I would make this a separate method `_tokenize_header` and then pass in the role into `self.encode` to avoid all the if else statements

torchtune/models/phi4/_model_builders.py
```diff
@@ -0,0 +1,126 @@
...
+    Builder for creating the Phi4 Mini 16K Instruct Model.
+
+    Note:
+        This model does not currently support 128K context length nor optimizations
+        such as sliding window attention.
+
+    Returns:
+        TransformerDecoder: Instantiation of Phi4 Mini 16K Instruct Model
+    """
+    return phi3(
```
RdoubleA, line 31: so there's no architectural difference between phi4 and phi3?

RdoubleA (commented):
I don't see anywhere where phi4 is referred as phi4-mini. On HF it seems to be only called Phi4. My preference would be to match the commonly accepted name and drop the mini, so we don't confuse folks with Phi4 vs Phi4 Mini.

Would also like to see detailed testing in the PR summary. Specifically, on:

- Comparing tokenizer and model forward against HF implementation
- Loss curves
- Running generation and eval on the model to ensure it gets reasonable outputs

torchtune/models/phi4/_tokenizer.py
krammnic: Ok, sure!

torchtune/models/phi4/_model_builders.py
krammnic: No real difference! See technical report. (We do not support sliding attention)

torchtune/models/phi4/_tokenizer.py
krammnic: Ok, sure!

krammnic:
> For `num_heads=40`, `num_kv_heads=10`, `embed_dim=5120`. Let's calculate:
> 
> `head_dim = 5120 / 40 = 128`
> 
> Already here:
> 
> ```
>  q_proj=nn.Linear(embed_dim, num_heads * head_dim, bias=False),
>  k_proj=nn.Linear(embed_dim, num_kv_heads * head_dim, bias=False),
>  v_proj=nn.Linear(embed_dim, num_kv_heads * head_dim, bias=False),
> ```
> 
> Already have a problem here as it will be not 2560 in all cases, but 5120, 1280, 1280. Assume that we "hardcoded" in a way that I have shown earlier. But then we get same problem here:
> 
> ```
>  q_per_kv = self.num_heads // self.num_kv_heads
>  q = q.view(b, s_x, self.num_kv_heads * q_per_kv, self.head_dim)
> ```
> 
> Error: `RuntimeError: shape '[2, 308, 40, 128]' is invalid for input of size 1576960`
> 
> Part of config.json for reference:
> 
> ```
>   "hidden_size": 5120,
>   "initializer_range": 0.02,
>   "intermediate_size": 17920,
>   "max_position_embeddings": 16384,
>   "model_type": "phi3",
>   "num_attention_heads": 40,
>   "num_hidden_layers": 40,
>   "num_key_value_heads": 10,
>   "original_max_position_embeddings": 16384,
> ```
> 
> So, the product should be twice less. Am I missing something? (I hope I have not miscalculated). Something weird is behind this problem. Will try to work out it asap. @ebsmothers I'm not really sure if it fixable without touching phi3 model or creating separate model for phi4.

Only point about architecture is this. And this is not align with tech report actually

torchtune/models/phi4/_tokenizer.py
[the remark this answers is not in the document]
krammnic: So maybe we apply UnSloth fixes at this point?

torchtune/models/phi4/_tokenizer.py
krammnic: typo :)

torchtune/models/phi4/_tokenizer.py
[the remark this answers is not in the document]
krammnic: There is pretty comprehensive report about all this "incorrect" things with fixes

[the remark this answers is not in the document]
joecummings: I think the changes will be approved and merged. Daniel and the Microsoft team are discussing the fixes here: https://huggingface.co/microsoft/phi-4/discussions/21.

Go ahead and do the fixes

ebsmothers (commented):
Thanks for the PR! I'm excited to get this landed into the library. A couple comments:

1) The configs don't currently run. IIUC Phi4 is using grouped query attention, which Phi3 does not. So it's possible that the weight mapping function needs to be updated.

2) We should run tests to confirm forward parity with a known implementation (probably the one on HF). E.g. you can check out [this file](https://github.com/pytorch/torchtune/blob/main/tests/torchtune/models/llama2/scripts/compare_decoder.py) comparing our Llama2 implementation to the one from the original meta-llama repo. @joecummings may already have some scripts from his parity checks for Phi3 that you would be able to reuse here.

torchtune/models/phi4/_model_builders.py
```diff
@@ -0,0 +1,124 @@
...
+            structured similarly. Default is None to use the canonical Phi4 special tokens.
+        max_seq_len (Optional[int]): maximum sequence length for tokenizing a single list of messages,
+            after which the input will be truncated. Default is None.
+        prompt_template (Optional[_TemplateType]): optional specified prompt template.
+            If a string, it is assumed to be the dotpath of a :class:`~torchtune.data.PromptTemplateInterface`
+            class. If a dictionary, it is assumed to be a custom prompt template mapping role to the
+            prepend/append tags.
+
+    Returns:
+        Phi4MiniTikTokenTokenizer: Instantiation of the SPM tokenizer.
```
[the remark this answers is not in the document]
krammnic: @ebsmothers I will fix this points ASAP(actually I have already fixed them locally) But, unfortunately our bottleneck will be related to tokenizer.

ebsmothers:
Hey @krammnic, sorry I completely missed [this comment of yours](https://github.com/pytorch/torchtune/pull/2197#issuecomment-2591010609) previously. Did you get this figured out? If not I think we need to make some updates like [this](https://github.com/pytorch/torchtune/compare/main...ebsmothers:ebs-torchtune:phi-weight-convert-demo) to the Phi3 convert weights function to support loading of models with GQA. I think you will also need to update num_heads from 20 to 40 like I suggested previously, but at least after these changes you are able to load the checkpoint. We should also make sure that the forward matches what's on HF -- it's possible that we may need to permute the fused QKV projection instead of naively splitting it as done in my changes. I've put together a minimal script showing roughly how you can do this [here](https://gist.github.com/ebsmothers/9013f0027c50bd07207cb03127316d98). The numbers do not line up, so it needs further investigation whether my torch.split is incorrect or whether I am just passing incorrect arguments to the HF version in that gist.

Separately, you mentioned in another comment there are some tokenizer issues blocking you. Can you elaborate on that? I'm happy to take a look here as well.

krammnic:
> Hey @krammnic, sorry I completely missed [this comment of yours](https://github.com/pytorch/torchtune/pull/2197#issuecomment-2591010609) previously. Did you get this figured out? If not I think we need to make some updates like [this](https://github.com/pytorch/torchtune/compare/main...ebsmothers:ebs-torchtune:phi-weight-convert-demo) to the Phi3 convert weights function to support loading of models with GQA. I think you will also need to update num_heads from 20 to 40 like I suggested previously, but at least after these changes you are able to load the checkpoint. We should also make sure that the forward matches what's on HF -- it's possible that we may need to permute the fused QKV projection instead of naively splitting it as done in my changes. I've put together a minimal script showing roughly how you can do this [here](https://gist.github.com/ebsmothers/9013f0027c50bd07207cb03127316d98). The numbers do not line up, so it needs further investigation whether my torch.split is incorrect or whether I am just passing incorrect arguments to the HF version in that gist.
> 
> Separately, you mentioned in another comment there are some tokenizer issues blocking you. Can you elaborate on that? I'm happy to take a look here as well.

Thanks for the comments, will check! Speaking about tokenizer I assume I can reference to #2212

ebsmothers:
@krammnic regarding the tokenizer: it doesn't have tokenizer.model, but it does have vocab.json and merges.txt. So I think this should be sufficient to construct the base tokenizer from one of our existing classes (though separately I think we can start supporting HF tokenizer.json without too too much extra effort)

krammnic:
Have done all requested changes! Probably the only point that we still need to fix is @ebsmothers observation

ebsmothers:
Hi @krammnic I took another look at the parity issue. I think you should push the changes I suggested [here](https://github.com/pytorch/torchtune/compare/main...ebsmothers:ebs-torchtune:phi-weight-convert-demo) and start running E2E tests. I adapted the script I shared before to mock out RoPE embeddings (there are some minor differences between our implementation and the one in HF). With those mocked out, I get forward pass parity within 1e-5. We should probably still make sure that nothing funky is happening with RoPE, but hopefully this is enough to unblock you for now. Also sorry I couldn't get it up sooner, but there is some initial progress on supporting HF tokenizers built from tokenizer.json files in #2350

krammnic:
@ebsmothers Haven't seen this message! Thanks for work on this. I came up to something similar either and was able to forward (but with worse implementation then yours), so I assume it should be fine. For now, we need to only fix the tokenizer problem and we will be able to merge.

krammnic:
Created new GPT2BaseTokenizer class to solve #2212 for this model

krammnic:
Loss for alpaca_cleaned after ~500 steps of LoRA single device.
W B Chart 2_9_2025, 1_47_01 AM

krammnic:
@ebsmothers It is ready for review. But, I don't really like this merge conflict... ^_^

ebsmothers:
@krammnic reviewing now. For the merge conflict, you should put `GPT2BaseTokenizer` under `torchtune/modules/transforms/tokenizers`, not `torchtune/modules/tokenizers` (the story will be similar for the unit test). You can see #2231 for the cause of the merge conflict. Your final changes should import `GPT2BaseTokenizer` in `torchtune/modules/transforms/tokenizers/__init__.py`, but leave `torchtune/modules/tokenizers/__init__.py` unchanged (we are deprecating it, so with your changes coming in after the deprecation has started you don't ned to expose in the old path).

torchtune/modules/tokenizers/_gpt2.py
```diff
@@ -0,0 +1,217 @@
...
+    prev_char = word[0]
+    for char in word[1:]:
+        pairs.add((prev_char, char))
+        prev_char = char
+    return pairs
+
+
+class GPT2BaseTokenizer(BaseTokenizer):
+    """
+    A lightweight version of the basic GPT2Tokenizer.
```
ebsmothers, line 62: nit: add a bit more detail here. E.g. mention it is using BPE algo, maybe link to GPT-2 paper as well (I think that's where it is defined?)

torchtune/modules/tokenizers/_gpt2.py
```diff
@@ -0,0 +1,217 @@
...
+        with open(merges_path, encoding="utf-8") as merges_handle:
+            bpe_merges = merges_handle.read().split("\n")[1:-1]
+        bpe_merges = [tuple(merge.split()) for merge in bpe_merges]
+        self.bpe_ranks = dict(zip(bpe_merges, range(len(bpe_merges))))
+        self.cache = {}
+
+        # Should have added re.IGNORECASE so BPE merges can happen for capitalized versions of contractions
+        # We do not use external regex library, so this is slightly modified version of the original line
+        self.pat = re.compile(
+            r"""'s|'t|'re|'ve|'m|'ll|'d| ?[^\W\d_]+| ?\d+| ?[^\s\W\d_]+|\s+(?!\S)|\s+"""
```
ebsmothers, line 106: What is this based on?

torchtune/modules/tokenizers/_gpt2.py
```diff
@@ -0,0 +1,217 @@
...
+    A lightweight version of the basic GPT2Tokenizer.
+
+    Args:
+        vocab_path (str): Path to vocab.json file.
+        merges_path (str): Path to merges.txt file.
+        errors (str): Paradigm to follow when decoding.
+        unk_id (int): unkown token id. This can be present or absent in ``special_tokens``.
+        bos_id (int): beginning-of-sequence token id. This can be present or absent in ``special_tokens``.
+        eos_id (int): end-of-sequence token id. This can be present or absent in ``special_tokens``.
+        pad_id (int): padding token id. This can be present or absent in ``special_tokens``.
```
ebsmothers, line 71: What `special_tokens` is this referring to? I think they're not known at this level, right? (Since `ModelTokenizer` is a higher level of abstraction than `BaseTokenizer`)

torchtune/modules/tokenizers/_gpt2.py
```diff
@@ -0,0 +1,217 @@
...
+
+    def bpe(self, token: str) -> str:
+        """
+        Returns pair for the given token.
+
+        Args:
+            token (str): Passed token.
+
+        Returns:
+            Pair token for the give token.
```
ebsmothers, line 126: typo nit
```suggestion
            Pair token for the given token.
```

torchtune/modules/tokenizers/_gpt2.py
```diff
@@ -0,0 +1,217 @@
...
+
+    def __init__(
+        self,
+        vocab_path: str,
+        merges_path: str,
+        errors: str,
+        unk_id: int,
+        bos_id: int,
+        eos_id: int,
+        pad_id: int,
```
ebsmothers, line 89: Are all four of these mandatory in all cases?

torchtune/modules/tokenizers/_gpt2.py
```diff
@@ -0,0 +1,217 @@
...
+        Returns:
+            Tokenized and encoded text.
+        """
+        bpe_tokens = list(map(self._convert_token_to_id, self._tokenize(text)))
+
+        if add_bos:
+            bpe_tokens = [self.bos_id] + bpe_tokens
+        if add_eos:
+            bpe_tokens = bpe_tokens + [self.eos_id]
+        return bpe_tokens
```
ebsmothers, line 217: I think we also need to define decode here (I see a version in Phi4Tokenizer, but we definitely need `decode` defined on the `BaseTokenizer` as well)

tests/torchtune/modules/tokenizers/test_gpt2.py
```diff
@@ -0,0 +1,64 @@
...
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+import pytest
+
+from tests.common import ASSETS
+from torchtune.modules.tokenizers import GPT2BaseTokenizer
+
+
+class TestGPT2BaseTokenizer:
```
ebsmothers, line 13: Thanks for adding the test! Similar q here to the Phi4Tokenizer: have we validated against a known GPT2 tokenizer implementation?

tests/torchtune/models/phi4/test_phi4_tokenizer.py
```diff
@@ -0,0 +1,159 @@
...
+        1803, 636, 277, 268, 117, 316, 485, 115, 284, 302, 416, 273, 900, 46, 270, 776, 591, 630, 346, 531,
+        476, 505, 768, 1233, 342, 1923, 292, 522, 662, 280, 274, 913, 601, 359, 300, 44, 335, 834, 335, 531,
+        476, 505, 604, 264, 509, 1456, 258, 771, 543, 1719, 405, 710, 665, 668, 1280, 46, 100266, 10,
+        100265]  # noqa
+        # fmt: on
+
+        expected_mask = [True] * 101 + [False] * 130
+        # Drop eos token.
+        assert expected_tokens[:-1] == tokens
+        assert expected_mask == mask
```
ebsmothers, line 159: Also given some of the confusion around special tokens for Phi4 it might be worth adding an explicit test case just to verify we have the right special tokens on our tokenizer.

torchtune/models/phi4/_tokenizer.py
```diff
@@ -0,0 +1,308 @@
...
+        else:
+            # To be still flexible at the point of tokenizer.model
+            self.tt_model = TikTokenBaseTokenizer(
+                path,
+                "phi4_tiktoken",
+                CL100K_PATTERN,
+                bos_id=self.bos_id,
+                eos_id=self.eos_id,
+                special_tokens=self.special_tokens,
+            )
```
ebsmothers, line 113: Maybe a dumb question, but why do we support both? I thought we only had access to merges + vocab files and there's no case where we will just have tokenizer.model file. Lmk if I'm mistaken though

torchtune/models/phi4/_tokenizer.py
```diff
@@ -0,0 +1,308 @@
...
+
+CL100K_PATTERN = r"""(?i:'s|'t|'re|'ve|'m|'ll|'d)|[^\r\n\p{L}\p{N}]?\p{L}+|\p{N}{1,3}| ?[^\s\p{L}\p{N}]+[\r\n]*|\s*[\r\n]+|\s+(?!\S)|\s+"""  # noqa
+
+
+class Phi4MiniTokenizer(ModelTokenizer, Transform):
+    """
+    TikToken tokenizer configured with Phi4 (14B) special tokens.
+
+    Args:
+        merges_path (str): Path to merges.txt file.
```
ebsmothers, line 45: Need to add path field if we're supporting it (but also see my comment below, I'm unclear on why we need it)

tests/torchtune/models/phi4/test_phi4_tokenizer.py
```diff
@@ -0,0 +1,159 @@
...
+from torchtune.models.phi4 import phi4_tokenizer
+
+
+class TestPhi4MiniTokenizer:
+    @pytest.fixture
+    def tokenizer(self):
+        # Pretrained tiktoken model generated via the script in
+        # https://gist.github.com/ebsmothers/54b133dd87db6679b14318545aaa2de4
+        return phi4_tokenizer(
+            path=str(ASSETS / "tiktoken_small.model"),
```
ebsmothers, line 20: One thing that concerns me a bit about this test is that we basically have two separate paths to construct the tokenizer. All our configs use one path (merges + vocab files + GPT2 base tokenizer), while here we only test the other one. I think we should try to test the other path if possible, I am happy to help construct similar toy tokenizers to the Tiktoken one used here if that'd be helpful.

torchtune/models/phi4/_tokenizer.py
```diff
@@ -0,0 +1,308 @@
...
+        prompt_template: Optional[PromptTemplate] = None,
+    ):
+        self.special_tokens = (
+            special_tokens if special_tokens is not None else PHI4_SPECIAL_TOKENS
+        )
+
+        # Use custom EOS, BOS and pad ids instead of TikToken's
+        self.eos_id = self.special_tokens["<|im_end|>"]
+        self.bos_id = self.special_tokens["<|endoftext|>"]
+        self.pad_id = self.special_tokens["<|dummy_87|>"]
```
ebsmothers, line 85: Maybe I'm missing something, but I think on HF hub they have `dummy_85`, not `dummy_87`. [Ref](https://huggingface.co/microsoft/phi-4/blob/main/special_tokens_map.json)

torchtune/models/phi4/_tokenizer.py
```diff
@@ -0,0 +1,308 @@
...
+    "<|im_sep|>": 100266,
+    "<|endofprompt|>": 100276,
+}
+
+# Adding other <|dummy_x|>
+
+for token_id in range(100266, 100351):
+    if token_id == 100276:
+        continue
+    PHI4_SPECIAL_TOKENS[f"<|dummy_{87 - (100350 - token_id)}|>"] = token_id + 1
```
ebsmothers, line 35: Wow the Phi-4 special token logic is messy. Thanks for being diligent about getting everything right!

torchtune/models/phi4/_tokenizer.py
```diff
@@ -0,0 +1,308 @@
...
+
+        # During generation, stop when eos_id is encountered
+        self.stop_tokens = [self.eos_id]
+
+        self.max_seq_len = max_seq_len
+
+        self.prompt_template = prompt_template
+
+        if path is None:
+            self.tt_model = GPT2BaseTokenizer(
```
ebsmothers, line 95: nit but tt is short for Tiktoken, so in this case it's not quite as applicable to call this `tt_model`

torchtune/models/phi4/_tokenizer.py
```diff
@@ -0,0 +1,308 @@
...
+    @property
+    def vocab_size(self):
+        return self.tt_model.vocab_size
+
+    def encode(
+        self,
+        text: str,
+        add_bos: bool = True,
+        add_eos: bool = True,
+        trim_leading_whitespace: bool = False,
```
ebsmothers, line 125: not used?

torchtune/models/phi4/_tokenizer.py
```diff
@@ -0,0 +1,308 @@
...
+            if self.prompt_template is not None
+            else messages
+        )
+
+        start_of_turn = True
+        end_of_turn = False
+        tokenized_messages = []
+        mask = []
+
+        # The chat template in HF adds a bunch of newlines
```
ebsmothers, line 218: Do you have a reference to this? I couldn't find it personally

torchtune/models/phi4/_tokenizer.py
```diff
@@ -0,0 +1,308 @@
...
+
+            if message.role == "assistant":
+                end_of_turn = True
+
+            # Add special tokens
+            tokenized_header = self._tokenize_header(message.role)
+
+            tokenized_messages.extend(tokenized_header)
+
+            # Minus 1, because the len(tokenized_header) is bigger then 1.
```
ebsmothers, line 239: I don't fully follow this, seems like after L241 you just wind up with `mask.extend([message.masked] * len(tokenized_header))` anyways?

ebsmothers (commented):
This is looking great! I think we are pretty close now. I left a bunch of comments but many are pretty minor. Btw I also have a few runs going [here](https://wandb.ai/ebs/phi4), loss curves look pretty good. 

Lmk if you need guidance on any of the comments I left or on resolving the merge conflicts you mentioned. Excited to be nearing the finish line on this one!

tests/torchtune/models/phi4/test_phi4_tokenizer.py
```diff
@@ -0,0 +1,159 @@
...
+                "interest in exploring new places, as well as someone who appreciates a "
+                "good conversation over coffee.",
+            ),
+        ]
+        tokens, mask = tokenizer.tokenize_messages(
+            messages, ignore_system_prompt=True, add_eos=True
+        )
+
+        # fmt: off
+        expected_tokens = [100257, 100264, 477, 273, 100265, 10, 66, 478, 299, 351, 362, 292, 1160, 117, 807, 334, 958, 99, 445,
```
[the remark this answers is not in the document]
krammnic: Yes, I have compared it to AutoTokenizer for Phi4 model. It is common GPT2Tokenizer. Therefore, it is similar

torchtune/models/phi4/__init__.py
```diff
@@ -0,0 +1,13 @@
...
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+from ._model_builders import lora_phi4, phi4, phi4_tokenizer  # noqa
+
+__all__ = [
+    "phi4",
+    "phi4_tokenizer",
+    "lora_phi4",
```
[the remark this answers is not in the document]
krammnic: Sound reasonable!

torchtune/models/phi4/_model_builders.py
```diff
@@ -0,0 +1,124 @@
...
+"""
+
+
+def phi4() -> TransformerDecoder:
+    """
+    Builder for creating the Phi4 (14B) Instruct Model.
+
+    Note:
+        This model does not currently support 128K context length nor optimizations
+        such as sliding window attention.
```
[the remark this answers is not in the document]
krammnic: I assume that comment related to context is correct. Speaking about sliding window attention, you are absolutely right, they used common one in this version.

torchtune/models/phi4/_tokenizer.py
krammnic: Yes, there were lot's of questions related to this dummy tokens. I hope that we will not face it in Phi5

torchtune/models/phi4/_model_builders.py
```diff
@@ -0,0 +1,124 @@
...
+        num_heads=40,
+        num_kv_heads=10,
+        embed_dim=5120,
+        intermediate_dim=17920,
+        max_seq_len=16384,
+        attn_dropout=0.0,
+        norm_eps=1e-5,
+    )
+
+def phi4_tokenizer(vocab_path: str = None, merges_path: str = None, path: str = None, special_tokens_path: Optional[str] = None, max_seq_len: Optional[int] = None, prompt_template: Optional[_TemplateType] = None) -> Phi4MiniTokenizer:
```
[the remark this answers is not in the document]
krammnic: Hmm, I ran pre-commit hooks, but yeah, it looks pretty long

torchtune/modules/tokenizers/_gpt2.py
krammnic: Original implantation used `import regex as re`. I don't want to add something new to the dependencies for only this line, so this is just version of the original expression adapted for the common re

torchtune/modules/tokenizers/_gpt2.py
krammnic: Yes, we need to remove this

tests/torchtune/modules/tokenizers/test_gpt2.py
krammnic: Yes, I compared manually. We can run stress test at this point (10000 random strings for both tokenizers) if required

torchtune/modules/tokenizers/_gpt2.py
krammnic: Hmm, OK. I don't see a use case in our recipes, but definitely it is required for general purposes

tests/torchtune/models/phi4/test_phi4_tokenizer.py
krammnic: Yes, sounds reasonable. I will think about possible design of this test

torchtune/models/phi4/__init__.py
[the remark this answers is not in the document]
krammnic: Actually, I'm not sure about it, because we actually have single model which is not really mini as we discussed, isn't it?

torchtune/modules/tokenizers/_gpt2.py
```diff
@@ -0,0 +1,217 @@
...
+import json
+import re
+from functools import lru_cache
+from typing import List
+
+from torchtune.modules.tokenizers._utils import BaseTokenizer
+
+# Constants controlling encode logic
+MAX_ENCODE_CHARS = 400_000
+MAX_NO_WHITESPACE_CHARS = 25_000
```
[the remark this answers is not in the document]
krammnic: Yes, I assume we don't need it

torchtune/modules/tokenizers/_gpt2.py
krammnic: Done

torchtune/models/phi4/_model_builders.py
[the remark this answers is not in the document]
SalmanMohammadi: Could you try `pre-commit uninstall`, and then `pre-commit install`?

torchtune/modules/tokenizers/_gpt2.py
```diff
@@ -0,0 +1,217 @@
...
+
+    def _convert_token_to_id(self, token: str):
+        return self.encoder.get(token, self.encoder.get(self.unk_id))
+
+    def encode(
+        self,
+        text: str,
+        add_bos: bool = True,
+        add_eos: bool = True,
+    ):
```
SalmanMohammadi, line 199:
```suggestion
    ) -> List[int]:
```
Need to also update the docstring below. I think once the pre-commit hooks have been reinstall it should catch this.

torchtune/modules/tokenizers/_gpt2.py
krammnic: Definitely, no, made then named arguments

torchtune/models/phi4/_tokenizer.py
krammnic: Fixed!

tests/torchtune/models/phi4/test_phi4_tokenizer.py
krammnic: Dropped the support of the tiktoken, but we need to rewrite this test then.. and also we need some small vocab.json for this purposes

torchtune/models/phi4/_tokenizer.py
krammnic: Fixed

torchtune/modules/tokenizers/_gpt2.py
krammnic: Hey Salman! Thanks for connecting to the review! Yes, it is a reasonable fix

torchtune/models/phi4/_tokenizer.py
krammnic: Nope

torchtune/models/phi4/_tokenizer.py
krammnic: Yes

krammnic:
@ebsmothers @SalmanMohammadi Have done probably almost all fixes that were required (I need to still cover Phi4Tokenizer with test about special tokens). You can check it out

torchtune/models/phi4/__init__.py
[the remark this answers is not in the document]
ebsmothers: Yeah actually my real suggestion would be `phi4` -> `phi4_14b` to match how we do it elsewhere. But I am also fine with `phi4_mini`

torchtune/modules/tokenizers/_gpt2.py
ebsmothers: Oh yeah I think we do not directly call `.decode` on the BaseTokenizer in any recipes. But e.g. in our generation recipe, we call `.decode` on the `ModelTokenizer`, which will call into the `BaseTokenizer`'s `.decode` method

krammnic:
@ebsmothers Probably done, except merge conflict. I'm not sure how to resolve it without write access

torchtune/modules/transforms/tokenizers/_gpt2.py
```diff
@@ -0,0 +1,232 @@
...
+
+        return bpe_tokens
+
+    def _convert_token_to_id(self, token: str) -> int:
+        return self.encoder.get(token, self.encoder.get(self.unk_id))
+
+    def _convert_id_to_token(self, index: int) -> str:
+        return self.decoder.get(index)
+
+    def decode(self, tokens: list) -> List[str]:
```
ebsmothers, line 195: nit: put this method after encode (not blocking to land though)

torchtune/modules/transforms/tokenizers/_gpt2.py
```diff
@@ -0,0 +1,232 @@
...
+        """
+        Decode sequence of the given tokens into string.
+
+        Args:
+            tokens (list): List of the integers, which represent encoded tokens.
+
+        Returns:
+            Decoced text.
```
ebsmothers, line 203: nit
```suggestion
            Decoded text.
```

recipes/configs/phi3/evaluation.yaml
```diff
@@ -12,20 +12,20 @@ model:
...
-  checkpoint_dir: /tmp/Phi-3-mini-4k-instruct
+  checkpoint_dir: /tmp/phi-3
   checkpoint_files: [
     model-00001-of-00002.safetensors,
     model-00002-of-00002.safetensors
   ]
   recipe_checkpoint: null
   output_dir: ${output_dir}
-  model_type: PHI3_MINI
+  model_type: PHI4_MINI
```
ebsmothers, line 22: I think this is still not correct? (Same for L28)

tests/assets/tiny_vocab.json
```diff
@@ -0,0 +1,6 @@
+{
```
ebsmothers, line 1: I believe this file can be removed now?

torchtune/models/phi4/_model_builders.py
```diff
@@ -0,0 +1,120 @@
...
+        intermediate_dim=17920,
+        max_seq_len=16384,
+        attn_dropout=0.0,
+        norm_eps=1e-5,
+    )
+
+def phi4_14b_tokenizer(vocab_path: str = None, merges_path: str = None, path: str = None, special_tokens_path: Optional[str] = None, max_seq_len: Optional[int] = None, prompt_template: Optional[_TemplateType] = None) -> Phi4MiniTokenizer:
+    """Phi4 (14B) tokenizer.
+    Args:
+        path (str): Path to the tiktoken tokenizer model.
```
ebsmothers, line 42: I think the docstring needs to be updated? (Also still not sure why this isn't linting on CI)

torchtune/models/phi4/_tokenizer.py
```diff
@@ -0,0 +1,293 @@
...
+            "replace",
+            self.eos_id,
+            self.bos_id,
+            self.eos_id,
+            self.pad_id,
+        )
+
+    @property
+    def vocab_size(self):
+        return self.tt_model.vocab_size
```
ebsmothers, line 106: I think this needs to be updated
```suggestion
        return self.tokenizer_model.vocab_size
```

torchtune/models/phi4/_tokenizer.py
```diff
@@ -0,0 +1,293 @@
...
+        """
+        ids_for_decode = []
+        for token_id in ids:
+            # Filter out special tokens and the placeholder tokens added
+            # by the Phi4 team
+            if skip_special_tokens and (token_id >= 100_256 and token_id <= 100_351):
+                continue
+            else:
+                ids_for_decode.append(token_id)
+        return self.tt_model.decode(ids_for_decode)
```
ebsmothers, line 139: This too
```suggestion
        return self.tokenizer_model.decode(ids_for_decode)
```

torchtune/modules/transforms/tokenizers/_gpt2.py
```diff
@@ -0,0 +1,232 @@
...
+
+
+class GPT2BaseTokenizer(BaseTokenizer):
+    """
+    A lightweight version of the basic GPT2Tokenizer.
+
+    Args:
+        vocab_path (str): Path to vocab.json file.
+        merges_path (str): Path to merges.txt file.
+        errors (str): Paradigm to follow when decoding.
```
ebsmothers, line 65: What are the acceptable values here? Looking at the docstring it's not clear to me how this is used

torchtune/modules/transforms/tokenizers/_gpt2.py
```diff
@@ -0,0 +1,232 @@
...
+        for token in re.findall(self.pat, text):
+            token = "".join(
+                self.byte_encoder[b] for b in token.encode("utf-8")
+            )  # Maps all our bytes to unicode strings, avoiding control tokens of the BPE (spaces in our case)
+            bpe_tokens.extend(bpe_token for bpe_token in self.bpe(token).split(" "))
+
+        return bpe_tokens
+
+    def _convert_token_to_id(self, token: str) -> int:
+        return self.encoder.get(token, self.encoder.get(self.unk_id))
```
ebsmothers, line 190: nit but can probably just cache `self.unk_token = self.encoder.get(self.unk_id)` during `__init__` or something so you don't have to call it repeatedly

ebsmothers (commented):
OK just a few more small comments! Let's also make sure everything runs on the final version before merging (I can help with this)

krammnic:
Done with all requested fixes

ebsmothers (approved):
Thanks @krammnic for adding Phi4 support to torchtune! Really excited to be able to support this

Issue #2202, opened 2024-12-23
Llama3.1 models do not allow configuring `max_seq_len`


akashc1:
Llama 3.1 model builders hardcode the max context length, even though the component builders allow specifying it:
- [Llama 3.1 8B](https://github.com/pytorch/torchtune/blob/v0.5.0/torchtune/models/llama3_1/_model_builders.py#L34)
- [Llama 3.1 70B](https://github.com/pytorch/torchtune/blob/v0.5.0/torchtune/models/llama3_1/_model_builders.py#L55)
- [Llama 3.1 405B](https://github.com/pytorch/torchtune/blob/v0.5.0/torchtune/models/llama3_1/_model_builders.py#L76)
- [Llama 3.1 8B LoRA](https://github.com/pytorch/torchtune/blob/v0.5.0/torchtune/models/llama3_1/_model_builders.py#L128)
- [Llama 3.1 70B LoRA](https://github.com/pytorch/torchtune/blob/v0.5.0/torchtune/models/llama3_1/_model_builders.py#L185)
- [Llama 3.1 405B LoRA](https://github.com/pytorch/torchtune/blob/v0.5.0/torchtune/models/llama3_1/_model_builders.py#L239)

And since the `QLoRA` versions also use these, it affects that too. This prevents anyone from specifying the model's `max_seq_len` from a config or CLI. E.g. this config will throw an error:
``` yaml
output_dir: /tmp/torchtune/llama3_1_8B/lora # /tmp may be deleted by your system. Change it to your preference.
max_seq_len: 8192

# Tokenizer
tokenizer:
  _component_: torchtune.models.llama3.llama3_tokenizer
  path: /models/meta-llama/Llama-3.1-8B-Instruct/original/tokenizer.model
  max_seq_len: ${max_seq_len}

# Model Arguments
model:
  _component_: torchtune.models.llama3_1.lora_llama3_1_8b
  lora_attn_modules: ['q_proj', 'v_proj', 'output_proj']
  apply_lora_to_mlp: True
  apply_lora_to_output: False
  lora_rank: 8  # higher increases accuracy and memory
  lora_alpha: 16  # usually alpha=2*rank
  lora_dropout: 0.0
  max_seq_len: ${max_seq_len}
```

```
[rank4]: Traceback (most recent call last):
[rank4]:   File "/torchtune/recipes/lora_finetune_distributed.py", line 938, in <module>
[rank4]:     sys.exit(recipe_main())
[...]

PR #2203, opened 2024-12-23
llama 3.1 has correct `max_seq_len` for all versions

akashc1:
#### Context
- [ ] add a new feature
- [x] fix a bug
- [ ] other (please add here)

Please link to any issues this PR addresses. #2202 

#### Changelog
* Ensure all llama 3.1 instantiations uses the correct `max_seq_len` the model was originally trained with based on the [HF config](https://huggingface.co/meta-llama/Llama-3.1-405B-Instruct/blob/main/config.json#L18)

#### Test plan

#### UX


felipemello1:
thanks for the PR, do you mind updating the docstrings too?!

RdoubleA:
This seems inconsistent with our other builders, which do not specify a max seq len parameter. If you want to change the default max seq len for llama3_1 you should use the base builder `llama3_1()` instead of the 8b and 405b builders, as they're meant to expose minimal parameters and give all the defaults to get the correct model. But let us know if this is too prohibitive.

akashc1:
@felipemello1 @RdoubleA thank you for the comments, can you please check the updated implementation? I set the correct default from the [HF model config](https://huggingface.co/meta-llama/Llama-3.1-405B-Instruct/blob/main/config.json#L18) for Llama 3.1 models in the `_model_builders` file.

felipemello1:
I am on PTO this week, but if Rafi doesnt approve it by next week, i will review it. @RdoubleA , you said "If you want to change the default max seq len for llama3_1 you should use the base builder llama3_1()", that is true, but if the model is the LoRA, this parameter is not exposed. IMO, all model parameters should be exposed in the lora builder. Do you disagree?

ebsmothers:
Commented on the issue as well, but I think we should only change the `llama3_1_405b` and `lora_llama3_1_405b` builders here. If we change `llama3_1` to hardcode `max_seq_len` it'd be inconsistent with our other models

ebsmothers:
@akashc1 thanks for catching the bug. Do you want to push the changes I mentioned? (Basically your PR should just set `max_seq_len=131072` in `llama3_1_405b` and `lora_llama3_1_405b`.) If not let me know and I can push to your PR.

RdoubleA (approved):

PR #2233, opened 2025-01-06
fix convert_weights not working for Qwen2.5 HF checkpoints

zhangtemplar:
Summary: In QWen2.5, the attention's linear projection layer has bias=True, but torchtune.convert_weights is not yet supporting bias=True. This diff add support for that

Differential Revision: D67880222

### DIFF
--- a/torchtune/models/convert_weights.py
+++ b/torchtune/models/convert_weights.py
@@ -33,6 +33,10 @@ _FROM_HF = {
     "model.layers.{}.self_attn.k_proj.weight": "layers.{}.attn.k_proj.weight",
     "model.layers.{}.self_attn.v_proj.weight": "layers.{}.attn.v_proj.weight",
     "model.layers.{}.self_attn.o_proj.weight": "layers.{}.attn.output_proj.weight",
+    "model.layers.{}.self_attn.q_proj.bias": "layers.{}.attn.q_proj.bias",
+    "model.layers.{}.self_attn.k_proj.bias": "layers.{}.attn.k_proj.bias",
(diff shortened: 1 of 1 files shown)


ebsmothers:
@calvinpelletier can you review?

calvinpelletier:
Hi @zhangtemplar , you're changing the generic `convert_weights` function. Qwen2.5 already has a specific convert weights function [here](https://github.com/pytorch/torchtune/blob/main/torchtune/models/qwen2/_convert_weights.py) which handles the biases of the linear projections.

joecummings:
> Hi @zhangtemplar , you're changing the generic `convert_weights` function. Qwen2.5 already has a specific convert weights function [here](https://github.com/pytorch/torchtune/blob/main/torchtune/models/qwen2/_convert_weights.py?rgh-link-date=2025-01-08T03%3A15%3A36Z) which handles the biases of the linear projections.
> 
> In our Qwen2.5 configs, we specify the model type as "QWEN2" [here](https://github.com/calvinpelletier/torchtune/blob/27fd3a14b04b5c3d428c723ef4a3a27e1595102b/recipes/configs/qwen2_5/7B_full_single_device.yaml#L53) which causes the checkpointer to call the Qwen-specific convert weights function [here](https://github.com/calvinpelletier/torchtune/blob/27fd3a14b04b5c3d428c723ef4a3a27e1595102b/torchtune/training/checkpointing/_checkpointer.py#L556)

This raises a good point tho that we don't tell the user if their model type is wrong. We should probably allow either QWEN2_5 or QWEN2 to point to the same conversion function.

joecummings:
No actual bug here - closing.

Issue #2225, opened 2025-01-02
Improvement: add a "division by zero" check in chunked loss handling in kd_losses.py


insop:
**Suggested improvement:**

It would be good to add a "division by zero" check in chunked loss handling in [kd_losses.py](https://github.com/pytorch/torchtune/blob/d23fa93d21045ab88f6914051f11de2b9afcd9eb/torchtune/modules/loss/kd_losses.py#L140).

**Context:**
This is based on the discussion in PR #2094.

**Potential issue:**
[ForwardKLWithChunkedOutputLoss](https://github.com/pytorch/torchtune/blob/d23fa93d21045ab88f6914051f11de2b9afcd9eb/torchtune/modules/loss/kd_losses.py#L140) does not check for division by zero, while the non-chunked version does.

joecummings:
Great catch!  We'd definitely welcome a small PR for this if you want to do it, otherwise we can try to get to it soon.

felipemello1:
That would be awesome. Thanks @insop :)

ebsmothers:
Closing this now that #2239 has landed

PR #2094, opened 2024-11-30
Adding reverse and symmetric KLD losses

insop:
#### Context
What is the purpose of this PR? Is it to
- [x] add a new feature
- [ ] fix a bug
- [ ] update tests and/or documentation
- [ ] other (please add here)

Please link to any issues this PR addresses.

#### Changelog
What are the changes made in this PR?
* Adding reverse and symmetric KLD loss
* Adding KLD losses based on [link](https://github.com/jongwooko/distillm/blob/17c0f98bc263b1861a02d5df578c84aea652ee65/distillm/losses.py)

#### Test plan

| Model | #Params | Method | HumanEval base | HumanEval plus | MBPP base | MBPP plus |
|:--|--:|:--|--:|--:|--:|--:|
| Llama3.2 | 8B | Teacher (Llama3.1) | 39.0 | 34.1 | 64.3 | 53.2 |
| | 1B | Base model | 18.3 | 15.9 | 35.4 | 29.0 |
| | | FT w/o KD | 22.0 | 17.7 | 37.6 | 31.7 |
| | | Forward KLD | 22.6 | 19.5 | **41.3** | 33.1 |
| | | Reverse KLD | 20.7 | 17.7 | 39.2 | 33.6 |
| | | Symmetric KLD | **23.2** | **20.1** | 41.0 | **33.9** |

Please make sure to do each of the following if applicable to your PR. If you're unsure about any one of these just ask and we will happily help. We also have a [contributing page](https://github.com/pytorch/torchtune/blob/main/CONTRIBUTING.md) for some guidance on contributing.

- [x] run pre-commit hooks and linters (make sure you've first installed via `pre-commit install`)
- [x] add [unit tests](https://github.com/pytorch/torchtune/tree/main/tests/torchtune) for any new functionality
- [x] update [docstrings](https://github.com/pytorch/torchtune/tree/main/docs/source) for any new or updated methods or classes
- [x] run unit tests via `pytest tests`
- [ ] run recipe tests via `pytest tests -m integration_test`
- [x] manually run any new or modified recipes with sufficient proof of correctness
- [x] include relevant commands and any other artifacts in this summary (pastes of loss curves, eval results, etc.)

#### UX
If your function changed a public API, please add a dummy example of what the user experience will look like when calling it.
Here is a [docstring example](https://github.com/pytorch/torchtune/blob/6a7951f1cdd0b56a9746ef5935106989415f50e3/torchtune/modules/vision_transformer.py#L285)
[...]

PR #2239, opened 2025-01-09
Add a "division by zero" check in chunked loss handling in kd_losses.py

insop:
#### Context
- [ ] add a new feature
- [ ] fix a bug
- [x] other (add a "division by zero" check in chunked loss handling in kd_losses.py)

#### Changelog
* add a "division by zero" check in chunked loss handling in kd_losses.py (Issue https://github.com/pytorch/torchtune/issues/2225)

#### Test plan

- [x] run `pytest -v tests/torchtune/modules/loss/test_kd_losses.py`

#### UX


insop:
@felipemello1 , @joecummings , PTAL

torchtune/modules/loss/kd_losses.py
```diff
@@ -137,4 +137,7 @@ def forward(
                 student_chunk, teacher_chunk, label_chunk, normalize=False
             )
 
+        if torch.sum(mask.view(-1), dim=0) == 0:
```
joecummings, line 140: Let's pull this out into a reasonably named variable.

torchtune/modules/loss/kd_losses.py
```diff
@@ -137,4 +137,7 @@ def forward(
...
+            return torch.tensor(0.0, device=student_logits[0].device)
```
joecummings, line 141: In the discussion on #2094, it seems like if we hit this point, there's probably something wrong with the training data. In that case, should we error out here? Or drop a warning? Or let people shoot themselves in the foot?

cc @lindawangg and @ebsmothers

joecummings (commented):
The tests look awesome - thanks!

Just a couple questions/comments on code clarity.

torchtune/modules/loss/kd_losses.py
insop: updated.

torchtune/modules/loss/kd_losses.py
insop: I will wait for inputs, and will update #2094 after this PR is resolved.

insop: Any inputs?
@lindawangg , @ebsmothers

ebsmothers: Thanks for the bump here. Personally I think this is the right way to handle it. Yes, we could error/raise a warning, but honestly this behavior should be pretty obvious in a loss curve as it will clearly just drop to zero on any such iterations (plus I think zero loss would technically be the "correct" loss value if every token is being ignored). No need to overdo it on handling this edge case here imo

tests/torchtune/modules/loss/test_kd_losses.py
```diff
@@ -17,7 +17,7 @@ def random():
 
 
 class TestForwardKLWithChunkedOutputLoss:
-    def test_forward_kl_loss(self):
+    def setup_forward_kl_loss(self, set_all_masks_zero: bool = False):
```
ebsmothers, line 20: nit: I would do something like "ignore_all_tokens" instead of "set_all_masks_zero" as I think it's clearer

ebsmothers (commented):
Thanks for the PR! Just one minor comment on naming in the test, otherwise this looks good

tests/torchtune/modules/loss/test_kd_losses.py
insop: Sounds better, and updated.

insop:
Thank you @ebsmothers , updated the change.

ebsmothers (approved):
Thanks for making this change! Once CI is green this should be good to land

PR #2245, opened 2025-01-10
Added Distributed(Tensor Parallel) Inference Recipe 

acisseJZhong:
#### Context
- [x] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

#### Changelog
Enabled TP for inference, for llama3, 3.1, 3.3 70B text only models. Llama3.2 vision 90B are still work in progress. It's blocked by https://github.com/pytorch/torchtune/issues/2277, will enable it as a follow up PR. 
* Copied `dev/generate_v2.py` and added TP to the recipe. The main change is in `__init__` and `__setup__`. 
* Added TP plan for llama3, under the model folders `_paralellism.py` file.
* Added TP utilities in `_distritbuted.py`, note that the utilities are for now only designed to work with llama models. 
* Added distributed inference config for llama3 70B and 3.1 70B, and 3.3 70B.
* Generalize [`load_from_full_model_state_dict`](https://github.com/pytorch/torchtune/blob/27fd3a14b04b5c3d428c723ef4a3a27e1595102b/torchtune/training/_distributed.py#L153) to general parallelism, not only FSDP. 
* Fixed a few typos.

#### Test plan

* Compared distributed inference results for llama3 8b with non distributed inference. Generations are at parity, below are the screenshot results for distributed inference and non-distributed inference. For distributed, it takes longer time but has lower peak memory. 
image 
image

* Running llama3 70B distributed inference on 8 Gpus:
```
Time for inference: 9.03 sec total, 3.10 tokens/sec
Bandwidth achieved: 408.57 GiB/s
Max memory allocated: 18.70 GiB
```
We see that the max memory allocated is only 18.7 GiB, which indicates we may use less number of GPUs. 
Running on 2 Gpus, we see that the max memory increases. 
```
Time for inference: 10.42 sec total, 3.17 tokens/sec
Bandwidth achieved: 417.21 GiB/s
Max memory allocated: 67.98 GiB
```

#### UX


recipes/dev/generate_v2_distributed.py
```diff
@@ -0,0 +1,259 @@
...
+        - multi-GPU generation
+        - batch generation
+    """
+
+    def __init__(self, cfg: DictConfig) -> None:
+        self._device = utils.get_device(device=cfg.device)
+        self._dtype = training.get_dtype(dtype=cfg.dtype, device=self._device)
+        self._logger = utils.get_logger(cfg.log_level)
+        # Set up distributed env
+        dist.init_process_group("cuda:nccl")
```
felipemello1, line 84: i have seen in other parts of the code this resulting in errors if we dont do `init_process_group("cuda:nccl,cpu:gloo")`

recipes/dev/generate_v2_distributed.py
```diff
@@ -0,0 +1,259 @@
...
+        _ckpt_dict = _checkpointer.load_checkpoint()
+
+        # Instantiate model on meta device
+        with training.set_default_dtype(self._dtype), torch.device("meta"):
+            model = config.instantiate(cfg.model)
+
+        # Set up tenosr parallel device mesh
+        tp_degree = dist.get_world_size()  # Using all GPUs for TP
+        tp_mesh_shape = (tp_degree,)
+        tp_device_mesh = dist.init_device_mesh("cuda", tp_mesh_shape)
```
felipemello1, line 102: n00b question: should we worry about other device types, e.g. npu?

recipes/dev/generate_v2_distributed.py
```diff
@@ -0,0 +1,259 @@
...
+
+        with training.set_default_dtype(self._dtype), self._device:
+            for m in model.modules():
+                # RoPE is not covered in state dict
+                if hasattr(m, "rope_init"):
+                    m.rope_init()
+
+        # This method will convert the full model state dict into a sharded state
+        # dict and load into the model
+        training.load_from_full_model_state_dict(
```
felipemello1, line 117: nit: as a rule of thumb, i think its worth using key arguments for all args, not only stric and cpu_offload

recipes/dev/generate_v2_distributed.py
```diff
@@ -0,0 +1,259 @@
...
+            ]
+        )
+        self._logger.info(
+            f"Time for inference: {total_time:.02f} sec total, {tokens_per_second:.02f} tokens/sec"
+        )
+        self._logger.info(
+            f"Bandwidth achieved: {model_size * tokens_per_second / 1e9:.02f} GB/s"
+        )
+        self._logger.info(
+            f"Max memory allocated: {torch.cuda.max_memory_allocated() / 1e9:.02f} GB"
```
felipemello1, line 154: nit: i think in general we prefer to use GiB, otherwise it may appear that we used more memory than the GPU has available --> to change replace 1e9 with /1024/1204

felipemello1 (commented):
some minor comments/question

torchtune/training/_distributed.py
```diff
@@ -45,6 +48,18 @@
     "dev" not in torch_version and torch_version_ge("2.6.0")
 ) or ("dev" in torch_version and torch_version.split("dev")[1] >= "20241220")
 
+BASE_LLAMA_TP_PLAN = {
```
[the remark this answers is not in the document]
acisseJZhong: Yeah I am also curious what's the best place to store this info. So it should be shared within llama3, 3.1, and 3.2, but we should define unique plans for 3.2 vision and 4. Maybe it should be stored in `_model_builders.py`? What's a better place?

torchtune/training/_distributed.py
```diff
@@ -45,6 +48,18 @@
...
+    "tok_embeddings": RowwiseParallel(input_layouts=Replicate()),
+    "output": ColwiseParallel(output_layouts=Replicate()),
+    "layers.*.attn.q_proj": ColwiseParallel(),
+    "layers.*.attn.k_proj": ColwiseParallel(),
+    "layers.*.attn.v_proj": ColwiseParallel(),
+    "layers.*.attn.output_proj": RowwiseParallel(),
+    "layers.*.mlp.w1": ColwiseParallel(),
+    "layers.*.mlp.w2": RowwiseParallel(),
+    "layers.*.mlp.w3": ColwiseParallel(),
+}
```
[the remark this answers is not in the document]
acisseJZhong: For matrix multiplication, we just need to make sure one is Col and the other is Row. For example, because the math is `mlp.w2(mlp.w1(x) * mlp.w3(x))`, therefore we just need to make sure that w1 and w3 are `col` and w2 is `row`, or the other way around.

torchtune/training/_distributed.py
```diff
@@ -546,3 +561,43 @@ def shard_model(
...
+    Adjusts the number of attention heads and dimension in the model to account for tensor parallelism.
+
+    Args:
+        model (nn.Module): Model to adjust.
+        tp_mesh (DeviceMesh): Tensor parallelism mesh.
+
+    Returns:
+        nn.Module: Adjusted model.
+    """
+    for transformer_block in model.layers:
```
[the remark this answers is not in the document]
acisseJZhong: thanks! yeah I had this in my local changes, trying to make vision 3.2 work. I made the function a bit ugly: 
```
def adjust_attention_for_tp():
    if hasattr(model, "layers") is False:
        model = model.decoder
```

recipes/dev/generate_v2.py
```diff
@@ -39,18 +39,22 @@ def __call__(self, prompt: Dict[str, Any]) -> List[Message]:
 
         # Iterate through roles and add content
         for role, content in prompt.items():
-            if isinstance(content, str):
+            if content is None:
+                continue
```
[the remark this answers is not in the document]
acisseJZhong: When I put something like 
```
system: null 
user:
    text: Tell me a joke. 
```
the system prompt could potentially be None, so we should probably handle it?

recipes/configs/llama3_2_vision/11B_generation_v2.yaml
```diff
@@ -27,7 +27,7 @@ checkpointer:
   checkpoint_files:
     filename_format: model-{}-of-{}.safetensors
     max_filename: "00005"
-  output_dir: ${output_dir}
+  output_dir: output_dir
```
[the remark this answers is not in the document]
acisseJZhong: oops, it's a typo.

recipes/dev/generate_v2_distributed.py
acisseJZhong: I do saw other file has `cpu:gloo` curious why? Since we will not use cpu as backend for inference

felipemello1: i have a vague memory that there may be some weight that is initialized using cpu, for some reason, and without cpu:gloo, it raises an error. But I dont remember exactly the issue. In any case, I dont think it hurts to add it.

torchtune/modules/attention.py
```diff
@@ -285,6 +285,7 @@ def forward(
         # as the query tensor by copying values across the relevant dim
         # k,v shape: [b, n_kv, s, h_d] -> [b, n_h, s, h_d]
         if self.num_heads != self.num_kv_heads:
+            # k,v shape: [b, n_kv, s, h_d] -> [b, n_h, s, h_d]
```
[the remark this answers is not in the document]
acisseJZhong: oops sorry copy paste error.

torchtune/training/_distributed.py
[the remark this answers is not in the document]
felipemello1: not my proudest moment: https://github.com/pytorch/torchtune/blob/890deab3029eef65f94cedb37fda14479f65f129/torchtune/training/_compile.py#L46

torchtune/training/_distributed.py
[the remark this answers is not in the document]
felipemello1: I am not sure. I feel that _model_builders.py would be too scattered. If we had training/distributed/, i would put it there. Does TorchTitan have something like this for multiple models? maybe we could check how they do it.

Every model has a checkpoint mapping torchtune <-> hf. How do we handle it? Probably we should follow the same pattern.

torchtune/training/_distributed.py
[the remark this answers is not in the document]
felipemello1: optional: Maybe adding this comment on top of it would be good

torchtune/training/_distributed.py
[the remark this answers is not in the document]
acisseJZhong: Torchtitan didn't define plan for each model, they just have one `apply_tp` function for llama3 in a `parallelize_llama.py`. 

I saw each model has a `convert_weights.py` file for converting weights format between hf and torchtune. Maybe let me create a parallelism file for each model, to put all the plans.

torchtune/training/_distributed.py
```diff
@@ -546,3 +564,72 @@ def shard_model(
 
     # Finally shard the entire model to account for any stragglers
     fully_shard(model, **fsdp_kwargs)
+
+
+def get_tp_plan(model_type: str) -> Dict[str, ParallelStyle]:
```
joecummings, line 569: I actually want to avoid something like this. It's very similar to how we did checkpointing where we would have complicated if/else logic that quickly got very confusing.

I'm imagining that the user would be able to pass in their TP plan directly from the config if they want to use tensor parallel.

torchtune/training/_distributed.py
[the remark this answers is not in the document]
joecummings: I don't think we need a parallelism file for every model. The vast majority of models that we support will be able to fall under the BASE_LLAMA_TP_PLAN. It doesn't have to live in training/ but it should live in somewhere centralized. Then, if there is a specific TP plan that we want to enable for, say, LLama3.2V, then we can define it either in the _model_builders.py file OR we can add a _parallelism.py file under the model directory where we define the TP plan.

torchtune/training/_distributed.py
```diff
@@ -546,3 +564,72 @@ def shard_model(
...
+    else:
+        raise ValueError("TP is only supported for llama type models right now.")
+
+
+def adjust_attention_for_tp(
+    model: nn.Module,
+    tp_mesh: DeviceMesh,
+) -> nn.Module:
+    """
+    Adjusts the number of attention heads and dimension in the model to account for tensor parallelism.
```
joecummings, line 598: This description is too vague. We should communicate what exactly is happening to the users.

torchtune/training/_distributed.py
```diff
@@ -546,3 +564,72 @@ def shard_model(
...
+        >>> # Adjust attention for tensor parallelism
+        >>> model = adjust_attention_for_tp(model, tp_mesh)
+        >>> # Now each GPU has:
+        >>> # num_heads = 16 (32/2)
+        >>> # num_kv_heads = 16 (32/2)
+        >>> # embed_dim = 2048 (4096/2)
+    """
+    # Consider the case of Early Fusion or Deep Fusion models
+    if isinstance(model, DeepFusionModel):
+        model = model.docoder
```
joecummings, line 624:
```suggestion
        model = model.decoder
```

torchtune/training/_distributed.py
```diff
@@ -546,3 +564,72 @@ def shard_model(
...
+    for layer in model.layers:
+        # Adjust attention module to use the local number of heads
+        attention_layers = ([layer.attn] if not isinstance(layer, FusionLayer) else [layer.fusion_layer.attn, layer.layer.attn])
+        for attn in attention_layers:
+            assert attn.num_heads % tp_mesh.size() == 0
+            assert attn.num_kv_heads % tp_mesh.size() == 0
+            assert attn.embed_dim % tp_mesh.size() == 0
+            attn.num_heads = attn.num_heads // tp_mesh.size()
```
joecummings, line 632: nit: Don't need to use the floor division operator if you already determined that the tp_mesh.size() goes evenly into the num_heads, etc.

torchtune/training/_distributed.py
```diff
@@ -546,3 +564,72 @@ def shard_model(
...
+        >>> # embed_dim = 2048 (4096/2)
+    """
+    # Consider the case of Early Fusion or Deep Fusion models
+    if isinstance(model, DeepFusionModel):
+        model = model.docoder
+    for layer in model.layers:
+        # Adjust attention module to use the local number of heads
+        attention_layers = ([layer.attn] if not isinstance(layer, FusionLayer) else [layer.fusion_layer.attn, layer.layer.attn])
+        for attn in attention_layers:
+            assert attn.num_heads % tp_mesh.size() == 0
```
joecummings, line 629: May as well pull `tp_mesh.size()` call out top.

torchtune/training/_distributed.py
```diff
@@ -546,3 +564,72 @@ def shard_model(
...
+        from torchtune.models.llama3_2_vision._model_builders import LLAMA_DEEP_FUSION_VISION_TP_PLAN
+
+        return LLAMA_DEEP_FUSION_VISION_TP_PLAN
+    elif model_type == "LLAMA3":
+        return BASE_LLAMA_TP_PLAN
+    else:
+        raise ValueError("TP is only supported for llama type models right now.")
+
+
+def adjust_attention_for_tp(
```
joecummings, line 593: Should we be more precise here? Something like `scale_attention_heads_by_tp_size`?

The only reason I could think not to do this is if there might be other adjustments to the attention we might need to do.

recipes/dev/generate_v2_distributed.py
```diff
@@ -0,0 +1,259 @@
...
+        with training.set_default_dtype(self._dtype), torch.device("meta"):
+            model = config.instantiate(cfg.model)
+
+        # Set up tenosr parallel device mesh
+        tp_degree = dist.get_world_size()  # Using all GPUs for TP
+        tp_mesh_shape = (tp_degree,)
+        tp_device_mesh = dist.init_device_mesh("cuda", tp_mesh_shape)
+
+        # Get TP plan and apply TP
+        tp_plan = training.get_tp_plan(cfg.checkpointer.model_type)
```
joecummings, line 105: I would not count on `model_type` staying around.

joecummings (commented):
What kind of tok/sec are we seeing with TP and Llama3 for the distributed inference recipe?

recipes/dev/generate_v2_distributed.py
```diff
@@ -0,0 +1,259 @@
...
+
+class InferenceRecipe:
+    """
+    Recipe for generating tokens from a dense Transformer-based LLM.
+    This works for text-only generation and image-text generation.
+
+    This *does not* currently support the following features:
+        - torch.compile
+        - quantization through torchao
+        - multi-GPU generation
```
RdoubleA, line 75: should update this and call out that this can be run distirbuted for larger models using TP. And could point to some pytorch docs on TP

recipes/dev/generate_v2_distributed.py
```diff
@@ -0,0 +1,259 @@
...
+        """Setup the model and transforms."""
+        # Load checkpointer and state_dict
+        _checkpointer = config.instantiate(cfg.checkpointer)
+        _ckpt_dict = _checkpointer.load_checkpoint()
+
+        # Instantiate model on meta device
+        with training.set_default_dtype(self._dtype), torch.device("meta"):
+            model = config.instantiate(cfg.model)
+
+        # Set up tenosr parallel device mesh
```
RdoubleA, line 99:
```suggestion
        # Set up tensor parallel device mesh
```

recipes/dev/generate_v2_distributed.py
```diff
@@ -0,0 +1,259 @@
...
+        tp_degree = dist.get_world_size()  # Using all GPUs for TP
```
RdoubleA, line 100: will this also work on multinode?

recipes/dev/generate_v2_distributed.py
acisseJZhong: I tried adding `cpu:gloo` when initializing the process group. But I am getting some RMSNorm cuda cpu device mismatch. With just `cuda:nccl` and the exact same code, it works with no problem. https://www.internalfb.com/phabricator/paste/view/P1713611664

RdoubleA (commented):
Really like this, we just need to figure out how we expect users to use our built-in TP plans and if they want to supply their own. and how this can be specified from the config.

torchtune/training/_distributed.py
felipemello1: i think that docoder sounds nicer

torchtune/training/_distributed.py
acisseJZhong: I am concerned that things could become messier if we allow user to directly change TP plan from config, especially if we have more parallelism enabled. In most cases the default TP plan should suffice, it's unclear why users would need to change it. Ideally each model will have default TP plan under `_parallelism.py` and if advanced user really want to experiment with plan, they can modify there.

torchtune/training/_distributed.py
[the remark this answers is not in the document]
joecummings: I'd vote for something like `BASIC_TP_PLAN`

torchtune/training/_distributed.py
acisseJZhong: it's expected to be an int, so I used `//`

recipes/dev/generate_v2_distributed.py
joecummings: Please don't make me worry about multi-node INFERENCE too

torchtune/training/_distributed.py
acisseJZhong: It changes number of heads as well as the emb dim. maybe `shard_attention_params_for_tp`?

recipes/dev/generate_v2_distributed.py
RdoubleA: then we should prevent users from trying multinode somewhere?

torchtune/training/_distributed.py
joecummings: > if advanced user really want to experiment with plan, they can modify there.

This would necessitate users using `git clone` to interact with torchtune, which isn't the case for many of our users. It has to somehow be possible to override the TP plan via some builder or directly in the config. We could just call the value `parallelism_plan`, which could extend to other types as well.

torchtune/training/_distributed.py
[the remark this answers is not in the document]
RdoubleA: BASED_TP_PLAN


(this is a joke)

torchtune/_recipe_registry.py
```diff
@@ -433,6 +433,17 @@ class Recipe:
         ],
         supports_distributed=False,
     ),
+    Recipe(
```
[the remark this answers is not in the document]
acisseJZhong: added config for 3, 3.1, and 3.2, for 3.2 I am still debugging.

torchtune/training/_distributed.py
RdoubleA: I'd say those are two separate problems - where to place the plans and how to specify them. I agree with placing the plans under each model's `_parallelism.py` or some central place for the default, but for accessing them you'll either need to do some weird dictionary mapping or specify it directly from the config. Since all the generation configs are model specific, we can just point to the default for each model.

The direct from config approach just lets you bypass the whole dict mapping, which is cleaner imo. Otherwise you will need to keep updating the mapping for every new model.

torchtune/training/_distributed.py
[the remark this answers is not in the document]
joecummings: Hahahaha ... unless? 👀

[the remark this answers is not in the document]
joecummings: Hahahaha ... unless? 👀

recipes/dev/generate_v2_distributed.py
joecummings: Can drop a comment at the top of the recipe

torchtune/training/_distributed.py
joecummings: Right, but both num_heads and tp_mesh.size() are ints so it'll always be an int.

recipes/dev/generate_v2_distributed.py
```diff
@@ -0,0 +1,259 @@
...
+        - multi-GPU generation
+        - batch generation
+    """
+
+    def __init__(self, cfg: DictConfig) -> None:
+        self._device = utils.get_device(device=cfg.device)
+        self._dtype = training.get_dtype(dtype=cfg.dtype, device=self._device)
+        self._logger = utils.get_logger(cfg.log_level)
+        # Set up distributed env
+        dist.init_process_group("cuda:nccl,cpu:gloo")
```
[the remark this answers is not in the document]
acisseJZhong: I was following `full_finetune_distributed.py` and seems like it doesn't use any util. Just `init_process_group()`

torchtune/training/_distributed.py
acisseJZhong: it's not in my case, I printed out in code, even though both numerator and denominator are int, the result is float.

joecummings: weird ... okay sounds good then!

acisseJZhong:
> What kind of tok/sec are we seeing with TP and Llama3 for the distributed inference recipe?
Take ~10 seconds, ~3 tokens/second to inference a single prompt for llama3 and 3.1.

recipes/dev/generate_v2_distributed.py
acisseJZhong: Added into the docstring for the `InferenceRecipe` class.

recipes/dev/generate_v2_distributed.py
ebsmothers: I think it's fine to leave that out in a first pass. Last I knew of we don't yet have distributed support for NPUs anyways (though @noemotiovon can inform me if my info is out of date here)

recipes/dev/generate_v2_distributed.py
```diff
@@ -0,0 +1,265 @@
...
+            ]
+        )
+        self._logger.info(
+            f"Time for inference: {total_time:.02f} sec total, {tokens_per_second:.02f} tokens/sec"
+        )
+        self._logger.info(
+            f"Bandwidth achieved: {model_size * tokens_per_second / 1024 / 1024:.02f} GB/s"
+        )
+        self._logger.info(
+            f"Max memory allocated: {torch.cuda.max_memory_allocated() / 1024 / 1024:.02f} GB"
```
ebsmothers, line 160: Same comment here as in generate_v2.py

torchtune/training/_distributed.py
```diff
@@ -546,3 +550,71 @@ def shard_model(
...
+                    f"tensor parallel size ({tp_size})."
+                )
+            if attn.embed_dim % tp_size != 0:
+                raise ValueError(
+                    f"Embedding dimension ({attn.embed_dim}) must be divisible by "
+                    f"tensor parallel size ({tp_size})."
+                )
+            attn.num_heads = attn.num_heads // tp_size
+            attn.num_kv_heads = attn.num_kv_heads // tp_size
+            attn.embed_dim = attn.embed_dim // tp_size
```
ebsmothers, line 619: Just want to make sure I understand the purpose of this utility: the idea is to set the appropriate params on the attention module so that any reshapes etc performed there will result in the correctly-shaped input for sharded Q, K, V, output projections?

ebsmothers: Also wonder whether we could just iterate over e.g. `model.modules()` and check `isinstance(m, MultiHeadAttentionLayer)` to avoid any dependency on the actual higher-level model arch details (fusion vs not, etc). For training this might be risky (e.g. AC will wrap modules and then you'd need to decide whether to apply on the AC-wrapped module vs not), but I don't see an obvious case it would break in inference.

ebsmothers (commented):
This is looking great! Really excited to see this landing in the library. Apart from inline comments, main questions are around testing -- it'd be great to update the PR summary to address the following questions.

1) Are the generations at parity with the corresponding ones from `generate_v2.py` (obviously not feasible for 70B, but could test on e.g. an 8B model)?
2) What's the throughput and peak memory? (On a related note, are 8 devices necessary to run 70B inference? If not, can maybe use a smaller number in the config)
3) Do the TP sharding utilities work with our other model families (Gemma, Mistral, Phi, etc)? Any models that they definitely do not work with? If that's the case we don't have to block on supporting everything, just want to be explicit about that.

torchtune/training/_distributed.py
ebsmothers: Just reviewed and I personally still found `shard_attention_params_for_tp` a bit confusing. Really you are just setting `num_heads` so that the reshapes are TP-aware, right? If that's the case I would even lean towards calling it `prepare_mha_for_tp` or something (the current name makes me thing you are actually distributed the params across devices in this utility, which you are not). Have I mentioned I hate naming things

torchtune/models/llama3/_parallelism.py
```diff
@@ -0,0 +1,34 @@
...
+from typing import Any, Dict
+
+from torch.distributed._tensor import Replicate
+from torch.distributed.tensor.parallel import ColwiseParallel, RowwiseParallel
+
+
+# Define the Tensor Parallel plan for Llama3 model, which will also be shared with 3.1, 3.2, and 3.3 models
+BASE_LLAMA_TP_PLAN = {
+    "tok_embeddings": RowwiseParallel(input_layouts=Replicate()),
+    "output": ColwiseParallel(output_layouts=Replicate()),
```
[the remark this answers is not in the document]
acisseJZhong: yeah I am curious how they got into this conclusion. For torchtune inference, I tried commenting these two lines out, the inference speed doesn't make too much different(I ran several times, sometimes it's slower and sometimes faster, all differs by 0.0x seconds.)

torchtune/models/llama3_2_vision/_parallelism.py
```diff
@@ -0,0 +1,48 @@
...
+        input_layouts=Replicate(), output_layouts=Shard(1)
+    ),
+    "decoder.output": ColwiseParallel(
+        input_layouts=Shard(1), output_layouts=Replicate()
+    ),
+    "decoder.layers.*.layer.attn.q_proj": ColwiseParallel(),
+    "decoder.layers.*.layer.attn.k_proj": ColwiseParallel(),
+    "decoder.layers.*.layer.attn.v_proj": ColwiseParallel(),
+    "decoder.layers.*.layer.attn.output_proj": RowwiseParallel(),
+    "decoder.layers.*.layer.mlp.w1": ColwiseParallel(),
```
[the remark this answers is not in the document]
acisseJZhong: good catch, I think it's the same, the wildcard should support this. But let me test it on 3.2 later, right now i am having problem because some distributed state dict problem with 3.2.

torchtune/models/llama3_2_vision/_parallelism.py
```diff
@@ -0,0 +1,48 @@
...
+    "decoder.tok_embeddings.embedding": RowwiseParallel(
+        input_layouts=Replicate(), output_layouts=Shard(1)
+    ),
+    "decoder.tok_embeddings.fusion_embedding": RowwiseParallel(
+        input_layouts=Replicate(), output_layouts=Shard(1)
+    ),
+    "decoder.output": ColwiseParallel(
+        input_layouts=Shard(1), output_layouts=Replicate()
+    ),
+    "decoder.layers.*.layer.attn.q_proj": ColwiseParallel(),
```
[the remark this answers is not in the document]
acisseJZhong: similar to here https://github.com/pytorch/torchtune/pull/2245#discussion_r1917554320, I think q, k, v are calculated separately using `self.k_proj`, `self.q_proj`, `self.v_proj`, so they should be the same kind of parallel, and `self.out_proj` operates on the result of qkv, therefore the other kind of parallel.

recipes/dev/generate_v2_distributed.py
```diff
@@ -0,0 +1,265 @@
...
+        - quantization through torchao
+        - batch generation
+    """
+
+    def __init__(self, cfg: DictConfig) -> None:
+        self._device = utils.get_device(device=cfg.device)
+        self._dtype = training.get_dtype(dtype=cfg.dtype, device=self._device)
+        self._logger = utils.get_logger(cfg.log_level)
+        # Set up distributed env
+        dist.init_process_group(backend="gloo" if cfg.device == "cpu" else "nccl")
```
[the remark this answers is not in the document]
acisseJZhong: probably not, let me remove that.

recipes/configs/llama3/70B_generation_distributed.yaml
```diff
@@ -0,0 +1,50 @@
...
+# To launch, run the following command from root torchtune directory:
+#    tune run --nproc_per_node 8 dev/generate_v2_distributed --config llama3/70B_generation_distributed
+
+output_dir: ./
+
+# Model arguments
+model:
+  _component_: torchtune.models.llama3.llama3_70b
+
+tensor_parallel_plan:
```
joecummings, line 16: I think we might want to mirror the parrelilize_module API and call this `parallelize_plan`. See https://pytorch.org/docs/main/distributed.tensor.parallel.html#torch.distributed.tensor.parallel.parallelize_module

joecummings (approved):
Incredible work!

recipes/dev/generate_v2_distributed.py
ebsmothers: Yeah I think gloo PG is only relevant for CPU offloading, which we aren't doing today anyways

torchtune/models/llama3/_parallelism.py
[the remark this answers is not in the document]
ebsmothers: Thanks, that's good to know. cc @fduwjj as the author of the corresponding torchchat PR in case you have any insights

torchtune/models/llama3_2_vision/_parallelism.py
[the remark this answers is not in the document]
ebsmothers: Ah thanks, missed this previously

torchtune/training/_distributed.py
```diff
@@ -546,3 +548,66 @@ def shard_model(
 
     # Finally shard the entire model to account for any stragglers
     fully_shard(model, **fsdp_kwargs)
+
+
+def prepare_mha_for_tp(
```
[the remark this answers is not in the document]
acisseJZhong: added! thanks for the reminder.

PR #2247, opened 2025-01-10
Multi-tile support in vision rope

RdoubleA:
#### Context
- [x] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

- Split out the tile dimension in `VisionRotaryPositionalEmbeddings` - this is squashed with the sequence dimension in `VisionTransformer`. This should enable support for multitile images. 2D RoPE would be applied within each tile.
- The `input_pos` parameter does not make sense with images as partial images will never be passed in, and packing will not change the position indices of each patch. This is removed.
- In the CLIP builders, instead of turning off tile position embeddings in CLIP vision encoder based off of `max_num_tiles`, add a parameter that explicitly controls this. This enables support for multiple tiles that are squashed in batch or patch dimension and global tile position embeddings are not used.

#### Test plan
- Change test inputs for multiple tiles and set num tiles = 1, do not change VisionRotaryPositionalEmbeddings. Update test values. Make the changes for multitile. Then update tests once again. This ensures the values didn't change because of the PR.
- Test that the constructor works with the new parameter.


ebsmothers (approved):

torchtune/modules/position_embeddings.py
```diff
@@ -209,46 +215,46 @@ def build_rope_cache(self) -> None:
...
+        if seq_len != self.cache.shape[0]:
+            raise ValueError(
+                f"Input sequence length {seq_len} does not match 2D RoPE cache sequence length {self.cache.shape[0]}."
+            )
 
         # reshape the cache for broadcasting
-        # tensor has shape [b, s, 1, h_d // 2, 2] if packed samples,
-        # otherwise has shape [1, s, 1, h_d // 2, 2]
-        rope_cache = rope_cache.view(-1, xshaped.size(1), 1, xshaped.size(3), 2)
+        rope_cache = self.cache.view(1, 1, seq_len, 1, h_d // 2, 2)
```
acisseJZhong, line 255: from the above `build_rope_cache` function, I saw `self.cache` has shape `[patches_per_tile + 1, dim, 2]. just curious why the shape looks like this?

torchtune/modules/position_embeddings.py
```diff
@@ -209,46 +215,46 @@ def build_rope_cache(self) -> None:
...
 
-        # tensor has shape [b, s, n_h, h_d // 2, 2]
+        # tensor has shape [b, max_num_tiles, s, n_h, h_d // 2, 2]
```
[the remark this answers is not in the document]
RdoubleA: oh yes, good catch

acisseJZhong (approved):
LGTM! maybe also update the calls from private repo?

torchtune/modules/position_embeddings.py
RdoubleA: mainly putting all the extra 1 dims to broadcast along batch, tile, and num heads dims. the rest matches the [patches_per_tile + 1, dim, 2] shape (seq_len, h_d // 2, 2)

RdoubleA:
> private repo?

I have no idea what you're talking about :)

Issue #2240, opened 2025-01-09
Grad Norm Differences Across Nodes


EugenHotaj:
Continuing the discussion from #2172 (thanks @mirceamironenco, @ebsmothers  for the fix!).

We have a run on the exact same dataset / hparams except we change the number of nodes from 8->2->1. We noticed that when we reduce the number of nodes the gradient norm goes up:

Here is an 8 node run:
<img width="542" alt="Screenshot 2025-01-09 at 12 06 30 PM" src="https://github.com/user-attachments/assets/5991a016-a602-4ad6-97d1-2c9bbedf4b09" />

Here is a 2 node run:
<img width="543" alt="Screenshot 2025-01-09 at 12 05 41 PM" src="https://github.com/user-attachments/assets/04b4ac8c-f37a-4afa-ad2e-cebb640ded31" />

Here is a 1 node run:
<img width="544" alt="Screenshot 2025-01-09 at 12 05 48 PM" src="https://github.com/user-attachments/assets/eaabef34-0c86-4e9d-8691-318f2a037c22" />

We can see the grad norm at initialization is ~4x different between 8 node and 1 node run. With the fix in #2172, I would expect the grad norms to be similar regardless of the world size. The only difference between the runs is the global batch size (64 on 1 node, 512 on 8 nodes), but I would not expect this to cause such a big difference.

Is it possible there are still some issues in how we compute / scale the gradients?

EugenHotaj:
I think this may just be a logging issue actually. To get the correct grad_norm we need to call `grad_norm.full_tensor()` like they do in torchtitan here: https://github.com/pytorch/torchtitan/blob/90567fc9827ffdf17bdd0349cd5276c662d0769a/torchtitan/utils.py#L396

I think right now we're just reporting the (unreduced) total norm from rank 0.

ebsmothers:
Thanks @EugenHotaj for the issue and the update. I agree that `full_tensor()` should give the correct grad. Just ran a quick test to make sure there's no major perf hit to doing this and it looks good. Do you wanna open a PR with the fix? (If not let me know and I am happy to do so)

ebsmothers:
Closing now that #2248 has landed

PR #2248, opened 2025-01-10
Log grad norm aggregated over all ranks, not just rank zero

ebsmothers:
Addresses #2240 

cc @EugenHotaj @mirceamironenco

### DIFF
--- a/recipes/full_finetune_distributed.py
+++ b/recipes/full_finetune_distributed.py
@@ -786,7 +786,7 @@ class FullFinetuneRecipeDistributed(FTRecipeInterface):
                             grad_norm = torch.nn.utils.clip_grad_norm_(
                                 self._model.parameters(),
                                 max_norm=float(self._clip_grad_norm),
-                            )
+                            ).full_tensor()
                         self._optimizer.step()
                         self._optimizer.zero_grad(set_to_none=True)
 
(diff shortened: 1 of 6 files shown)


RdoubleA (approved):

joecummings (approved):
beaut

recipes/full_finetune_distributed.py
```diff
@@ -786,7 +786,7 @@ def train(self) -> None:
                             grad_norm = torch.nn.utils.clip_grad_norm_(
                                 self._model.parameters(),
                                 max_norm=float(self._clip_grad_norm),
-                            )
+                            ).full_tensor()
```
mirceamironenco, line 789: Do you think it might be a good idea to have the `.full_tensor()` behind an `isintance(grad_norm, DTensor)` check? If e.g. DDP ever gets implemented and torchtune takes care of it behind some API (say `shard_model` allows for different types of parallelisms, or [TP gets added](https://github.com/pytorch/torchtitan/blob/main/torchtitan/utils.py#L393-L396)).

ebsmothers: @mirceamironenco yes I agree we should have the check when we enable new types of parallelism. But I also don't want to prematurely expose it (our recipes are already more complicated than I would like and adding a check that's currently a no-op is an easy case of more code to read than we currently we need). I think your TP example is a very likely case and when we enable something like that wrapping grad norm logic in an appropriate utility (kinda like what you shared with me over Discord) will be the way to go. But until then I don't think we should do it. Hope that makes sense

PR #2251, opened 2025-01-10
Update the e2e flow tutorial to fix errors of generate

iseeyuan:
#### Context
- [ ] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

#### Changelog
When running the tutorial as is, there's error message: 
```
ValueError: The output directory cannot be the same as or a subdirectory of the checkpoint directory. Found ckpt_dir=PosixPath('/tmp/torchtune/llama3_2_3B/lora_single_device/epoch_0') and out_dir=PosixPath('/tmp/torchtune/llama3_2_3B/lora_single_device/epoch_0')
```

* The fix is to add a specific `checkpoint_dir` and make it different from the `output_dir`. 

Then I saw the second error: 
```
...
  File "/home/myuan/src/torchtune/recipes/generate.py", line 113, in convert_prompt_to_tokens
    Message(role="user", content=prompt["user"]),
TypeError: string indices must be integers
```

* The fix is to remove the `prompt` option from run. It's redundant with the prompt in the .yaml file anyway. 

#### Test plan

#### UX

### DIFF
--- a/docs/source/tutorials/e2e_flow.rst
+++ b/docs/source/tutorials/e2e_flow.rst
@@ -314,8 +314,8 @@ Let's modify ``custom_generation_config.yaml`` to include the following changes.
 
     # Generation arguments; defaults taken from gpt-fast
     prompt:
-    system: null
-    user: "Tell me a joke. "
+      system: null
+      user: "Tell me a joke. "
     max_new_tokens: 300
     temperature: 0.6 # 0.8 and 0.6 are popular values to try
     top_k: 300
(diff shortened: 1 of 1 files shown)


docs/source/tutorials/e2e_flow.rst
```diff
@@ -330,7 +332,7 @@ these parameters.
 
 .. code-block:: text
 
-    $ tune run generate --config ./custom_generation_config.yaml prompt="tell me a joke. "
+    $ tune run generate --config ./custom_generation_config.yaml
```
RdoubleA, line 335: could you try running this?
`tune run generate --config ./custom_generation_config.yaml prompt.user="Tell me a joke. "`

if that works fine, would prefer including this as an example just to be explicit

RdoubleA (commented):
Looks great! just one small comment

RdoubleA (approved):

PR #2266, opened 2025-01-14
[EZ] Pass seed to data sampler.

EugenHotaj:
We currently hardcode zero. I assume this is not intentional.

#### Context
- [ ] add a new feature
- [x] fix a bug
- [ ] other (please add here)

#### Changelog
* As titled

#### Test plan

#### UX

### DIFF
--- a/recipes/dev/early_exit_finetune_distributed.py
+++ b/recipes/dev/early_exit_finetune_distributed.py
@@ -664,7 +664,7 @@ class EarlyExitFinetuneRecipeDistributed(FTRecipeInterface):
         collate_fn = _get_component_from_path(collate_fn)
 
         sampler = DistributedSampler(
-            ds, num_replicas=world_size, rank=rank, shuffle=shuffle, seed=0
+            ds, num_replicas=world_size, rank=rank, shuffle=shuffle, seed=self.seed
         )
         dataloader = DataLoader(
             dataset=ds,
--- a/recipes/full_finetune_distributed.py
+++ b/recipes/full_finetune_distributed.py
@@ -657,7 +657,7 @@ class FullFinetuneRecipeDistributed(FTRecipeInterface):
         collate_fn = _get_component_from_path(collate_fn)
 
         sampler = DistributedSampler(
-            ds, num_replicas=world_size, rank=rank, shuffle=shuffle, seed=0
+            ds, num_replicas=world_size, rank=rank, shuffle=shuffle, seed=self.seed
         )
         dataloader = DataLoader(
             dataset=ds,
--- a/recipes/full_finetune_single_device.py
+++ b/recipes/full_finetune_single_device.py
@@ -573,7 +573,7 @@ class FullFinetuneRecipeSingleDevice(FTRecipeInterface):
             num_replicas=1,
             rank=0,
             shuffle=shuffle,
-            seed=0,
+            seed=self.seed,
         )
         dataloader = DataLoader(
             dataset=ds,
--- a/recipes/knowledge_distillation_distributed.py
+++ b/recipes/knowledge_distillation_distributed.py
@@ -662,7 +662,7 @@ class KDRecipeDistributed(FTRecipeInterface):
             num_replicas=world_size,
             rank=rank,
             shuffle=shuffle,
-            seed=0,
+            seed=self.seed,
         )
         dataloader = DataLoader(
             dataset=ds,
--- a/recipes/knowledge_distillation_single_device.py
+++ b/recipes/knowledge_distillation_single_device.py
@@ -537,7 +537,7 @@ class KDRecipeSingleDevice(FTRecipeInterface):
             num_replicas=1,
             rank=0,
             shuffle=shuffle,
-            seed=0,
+            seed=self.seed,
         )
         dataloader = DataLoader(
             dataset=ds,
--- a/recipes/lora_dpo_distributed.py
+++ b/recipes/lora_dpo_distributed.py
@@ -504,7 +504,7 @@ class LoRADPORecipeDistributed(FTRecipeInterface):
             ds = config.instantiate(cfg_dataset, tokenizer=self._tokenizer)
 
         sampler = DistributedSampler(
-            ds, num_replicas=world_size, rank=rank, shuffle=shuffle, seed=0
+            ds, num_replicas=world_size, rank=rank, shuffle=shuffle, seed=self.seed
         )
 
         dataloader = DataLoader(
--- a/recipes/lora_dpo_single_device.py
+++ b/recipes/lora_dpo_single_device.py
@@ -389,7 +389,7 @@ class LoRADPORecipeSingleDevice(FTRecipeInterface):
             num_replicas=1,
             rank=0,
             shuffle=shuffle,
-            seed=0,
+            seed=self.seed,
         )
         dataloader = DataLoader(
             dataset=ds,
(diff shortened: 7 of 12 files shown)


RdoubleA:
Getting the seed from `self.seed` is risky, since we are using `self.seed = training.set_seed()` in our recipes, which will actually return a different seed on every rank. I am not sure if the sampler should have the same seed across ranks, but I imagine to ensure all ranks are not shuffling the data differently we would want the same seed. It would be better to set it from cfg.seed directly to keep the same data seed across all ranks.

Tagging @andrewkho for thoughts on setting the seed in `DistributedSampler`.

EugenHotaj:
@RdoubleA I might be misunderstanding the code but I thought `set_seed` returned the seed you passed it in from the config (or if you don't pass anything, a global seed). I guess my intention here is to just pass in the `seed` from the config. Lmk if there is a better way to do it (maybe pass in cfg.seed directly?)

RdoubleA:
Ah my bad, you're right. It's returning the global seed not the local seed. In that case this should be ok

RdoubleA:
@EugenHotaj actually we might need to do something like this to get the seed directly from the config and maintaining BC with existing configs:
`seed=cfg.get("seed", None) or 0`

EugenHotaj:
@RdoubleA done

EugenHotaj:
@RdoubleA I'm not sure what's happening with the `TestFullFinetuneDistributedRecipe::test_training_state_on_resume_from_distributed_checkpoint_multi_rank` test that keeps failing. 

It keeps deadlocking for some reason: 

```
3|5|Loss: 11.97959041595459:  50%|█████     | 1/2 [00:00<00:00,  1.63it/s][rank0]:[E120 05:15:13.042532067 ProcessGroupNCCL.cpp:616] [Rank 0] Watchdog caught collective operation timeout: WorkNCCL(SeqNum=420, OpType=_REDUCE_SCATTER_BASE, NumelIn=8208384, NumelOut=4104192, Timeout(ms)=1800000) ran for 1800059 milliseconds before timing out.
[rank0]:[E120 05:15:13.042708659 ProcessGroupNCCL.cpp:1785] [PG ID 0 PG GUID 0(default_pg) Rank 0] Exception (either an error or timeout) detected by watchdog at work: 420, last enqueued NCCL work: 426, last completed NCCL work: 419.
[rank1]:[E120 05:15:13.091641415 ProcessGroupNCCL.cpp:616] [Rank 1] Watchdog caught collective operation timeout: WorkNCCL(SeqNum=423, OpType=_REDUCE_SCATTER_BASE, NumelIn=64, NumelOut=32, Timeout(ms)=1800000) ran for 1800096 milliseconds before timing out.
[rank1]:[E120 05:15:13.091715615 ProcessGroupNCCL.cpp:1785] [PG ID 0 PG GUID 0(default_pg) Rank 1] Exception (either an error or timeout) detected by watchdog at work: 423, last enqueued NCCL work: 424, last completed NCCL work: 422.
[rank1]:[E120 05:15:14.970203483 ProcessGroupNCCL.cpp:1834] [PG ID 0 PG GUID 0(default_pg) Rank 1] Timeout at NCCL work: 423, last enqueued NCCL work: 424, last completed NCCL work: 422.
[rank1]:[E120 05:15:14.970217275 ProcessGroupNCCL.cpp:630] [Rank 1] Some NCCL operations have failed or timed out. Due to the asynchronous nature of CUDA kernels, subsequent GPU operations might run on corrupted/incomplete data.
[rank1]:[E120 05:15:14.970222520 ProcessGroupNCCL.cpp:636] [Rank 1] To avoid data inconsistency, we are taking the entire process down.
[rank1]:[E120 05:15:14.971615376 ProcessGroupNCCL.cpp:1595] [PG ID 0 PG GUID 0(default_pg) Rank 1] Process group watchdog thread terminated with exception: [Rank 1] Watchdog caught collective operation timeout: WorkNCCL(SeqNum=423, OpType=_REDUCE_SCATTER_BASE, NumelIn=64, NumelOut=32, Timeout(ms)=1800000) ran for 1800096 milliseconds before timing out.
Exception raised from checkTimeout at ../torch/csrc/distributed/c10d/ProcessGroupNCCL.cpp:618 (most recent call first):
frame #0: c10::Error::Error(c10::SourceLocation, std::string) + 0x96 (0x7f6f92b6c446 in /home/ec2-user/actions-runner/_work/torchtune/torchtune/3/envs/test/lib/python3.10/site-packages/torch/lib/libc10.so)
frame #1: c10d::ProcessGroupNCCL::WorkNCCL::checkTimeout(std::optional<std::chrono::duration<long, std::ratio<1l, 1000l> > >) + 0x282 (0x7f6f485cc772 in /home/ec2-user/actions-runner/_work/torchtune/torchtune/3/envs/test/lib/python3.10/site-packages/torch/lib/libtorch_cuda.so)
frame #2: c10d::ProcessGroupNCCL::watchdogHandler() + 0x233 (0x7f6f485d3bb3 in /home/ec2-user/actions-runner/_work/torchtune/torchtune/3/envs/test/lib/python3.10/site-packages/torch/lib/libtorch_cuda.so)
frame #3: c10d::ProcessGroupNCCL::ncclCommWatchdog() + 0x14d (0x7f6f485d561d in /home/ec2-user/actions-runner/_work/torchtune/torchtune/3/envs/test/lib/python3.10/site-packages/torch/lib/libtorch_cuda.so)
frame #4: <unknown function> + 0x145c0 (0x7f6f9301d5c0 in /home/ec2-user/actions-runner/_work/torchtune/torchtune/3/envs/test/lib/python3.10/site-packages/torch/lib/libtorch.so)
frame #5: <unknown function> + 0x9f832 (0x7f6fa0c9f832 in /lib64/libc.so.6)
frame #6: <unknown function> + 0x3f480 (0x7f6fa0c3f480 in /lib64/libc.so.6)
```

Doesn't seem like it would be related to these changes (or maybe surfacing some latent bug), but any idea what's going on?

EugenHotaj:
^ @RdoubleA any thoughts on the failure I mentioned above?

RdoubleA:
@EugenHotaj Sorry I've been meaning to investigate this more. I do think it's related to changing the seed though, the failing test is using DCP's resume from checkpoint. DCP was only recently added, so my guess is it's not setting the seed correctly when resuming training.

Will update here once I debug more.

ebsmothers:
Quick update on the deadlock: I noticed it only occurs for the 8B model and not the 7B one, and for the 8B config we shard a couple extra layers by default. So by commenting out [this line](https://github.com/pytorch/torchtune/blob/23896c30dc10c97f8baaed8841e61bc3b8f2a61c/recipes/configs/llama3/8B_full.yaml#L73) the test actually passes on your PR. I am not exactly sure why this sharding by itself is OK, but combining it with the seed is not. I've been digging around the DCP code a bit and found that by setting `planner=DefaultPlanner(flatten_state_dict=False)` in both `async_save` and `load` we get by the deadlock (but the outputs incorrect). Admittedly I'm a bit of a DCP noob so will tag experts @saumishr and @ankitageorge for their thoughts here.

EugenHotaj:
@ebsmothers thanks for taking a look!

> Many of our configs set seed: null and in that case we will actually pass seed=None [here](https://github.com/EugenHotaj/torchtune/blob/eugen/patch-1/recipes/full_finetune_distributed.py#L662), triggering an error in the sampler.

Let me verify this, my assumption was that this would return `0` when `seed=null` (i.e. `seed=None`) but you might be right.

Issue #2082, opened 2024-11-27
Full-finetune DPO single device recipe


salmanmohammadi:
This should be straightforward. The main issue I see coming up is with compile - similar to how we attempt to compile the reference and policy model in our single device PPO recipe. Since the `SelfAttentionLayer` block is inlined and shared across the models, we're going to hit recompiles due to `param.requires_grad`. This might be acceptable in this case, since the recompiles won't be as severe as with PPO in it's current state #2066.

We might want to offer some kind of customization around the choice of reference policy model. The only constraint I can think of here is ensuring that both of the reference and policy models share a tokenizer - otherwise users should be able to freely experiment here.

PR #1966, opened 2024-11-07
full dpo

jxmsML:
#### Context
What is the purpose of this PR? Is it to
- [x] add a new feature
- [ ] fix a bug
- [ ] update tests and/or documentation
- [ ] other (please add here)

Please link to any issues this PR addresses.

#### Changelog
What are the changes made in this PR?
*

#### Test plan
Please make sure to do each of the following if applicable to your PR. If you're unsure about any one of these just ask and we will happily help. We also have a [contributing page](https://github.com/pytorch/torchtune/blob/main/CONTRIBUTING.md) for some guidance on contributing.

- [ ] run pre-commit hooks and linters (make sure you've first installed via `pre-commit install`)
- [ ] add [unit tests](https://github.com/pytorch/torchtune/tree/main/tests/torchtune) for any new functionality
- [ ] update [docstrings](https://github.com/pytorch/torchtune/tree/main/docs/source) for any new or updated methods or classes
- [ ] run unit tests via `pytest tests`
- [ ] run recipe tests via `pytest tests -m integration_test`
- [ ] manually run any new or modified recipes with sufficient proof of correctness
- [ ] include relevant commands and any other artifacts in this summary (pastes of loss curves, eval results, etc.)

#### UX
If your function changed a public API, please add a dummy example of what the user experience will look like when calling it.
Here is a [docstring example](https://github.com/pytorch/torchtune/blob/6a7951f1cdd0b56a9746ef5935106989415f50e3/torchtune/modules/vision_transformer.py#L285)
and a [tutorial example](https://pytorch.org/torchtune/main/tutorials/qat_finetune.html#applying-qat-to-llama3-models)

- [ ] I did not change any public API
- [ ] I have added an example to docs or docstrings

PR #2275, opened 2025-01-17
Full DPO Distributed

sam-pi:
#### Context

Adapted from the great work in #1966 

- [x] add a new feature

Please link to any issues this PR addresses: relates to #2082 

#### Changelog
* Adds full DPO distributed training configs and recipes, adapting from the lora DPO training
* Includes integration tests
* Includes configs for llama3.1 8B and 70B models

#### Test plan

#### UX

### Commands and Sample Outputs
Full DPO Config
```yaml
output_dir: .../Meta-Llama-3.1-8B-Instruct/full_dpo
model:
tokenizer:
  path: .../Meta-Llama-3.1-8B-Instruct/original/tokenizer.model
  max_seq_len: 1024
checkpointer:
  checkpoint_dir: .../Meta-Llama-3.1-8B-Instruct
  checkpoint_files:
  - model-00001-of-00004.safetensors
  - model-00002-of-00004.safetensors
  - model-00003-of-00004.safetensors
  - model-00004-of-00004.safetensors
  recipe_checkpoint: null
  output_dir: ${output_dir}
  model_type: LLAMA3
resume_from_checkpoint: false
ref_checkpointer:
  checkpoint_dir: .../Meta-Llama-3.1-8B-Instruct
  checkpoint_files:
  - model-00001-of-00004.safetensors
  - model-00002-of-00004.safetensors
  - model-00003-of-00004.safetensors
  - model-00004-of-00004.safetensors
  recipe_checkpoint: null
  output_dir: ${output_dir}
  model_type: LLAMA3
dataset:
  _component_: torchtune.datasets.stack_exchange_paired_dataset
seed: null
shuffle: true
batch_size: 4
optimizer:
  fused: true
  weight_decay: 0.05
  lr: 1.0e-06
lr_scheduler:
  num_warmup_steps: 100
loss:
  _component_: torchtune.rlhf.loss.DPOLoss
  beta: 0.05
  label_smoothing: 0
epochs: 1
max_steps_per_epoch: 2000
gradient_accumulation_steps: 4
compile: false
metric_logger:
  _component_: torchtune.training.metric_logging.WandBLogger
  log_dir: ${output_dir}/logs
  project: torchtune
  name: llama3.1-8B-dpo_3605
log_every_n_steps: 1
log_peak_memory_stats: true
device: cuda
dtype: bf16
```

Lora DPO Config
```yaml
output_dir: .../Meta-Llama-3.1-8B-Instruct/lora_dpo
model:
  _component_: torchtune.models.llama3_1.lora_llama3_1_8b
  lora_attn_modules:
  - q_proj
  - v_proj
  - output_proj
  apply_lora_to_mlp: true
  apply_lora_to_output: false
  lora_rank: 256
  lora_alpha: 256
  lora_dropout: 0.0
tokenizer:
  path: .../Meta-Llama-3.1-8B-Instruct/original/tokenizer.model
  max_seq_len: 1024
checkpointer:
  checkpoint_dir: .../Meta-Llama-3.1-8B-Instruct
  checkpoint_files:
  - model-00001-of-00004.safetensors
  - model-00002-of-00004.safetensors
  - model-00003-of-00004.safetensors
  - model-00004-of-00004.safetensors
  recipe_checkpoint: null
  output_dir: ${output_dir}
  model_type: LLAMA3
resume_from_checkpoint: false
save_adapter_weights_only: false
dataset:
  _component_: torchtune.datasets.stack_exchange_paired_dataset
seed: null
shuffle: true
batch_size: 4
optimizer:
  fused: true
  weight_decay: 0.05
  lr: 1.0e-05
lr_scheduler:
  num_warmup_steps: 100
loss:
  _component_: torchtune.rlhf.loss.DPOLoss
  beta: 0.1
  label_smoothing: 0
epochs: 1
max_steps_per_epoch: 100
gradient_accumulation_steps: 4
compile: false
metric_logger:
  _component_: torchtune.training.metric_logging.WandBLogger
  log_dir: ${output_dir}/logs
  project: torchtune
  name: llama3.1-8Blora-dpo_3603
log_every_n_steps: 1
log_peak_memory_stats: true
device: cuda
dtype: bf16
```
Screenshot 2025-01-16 at 12 39 23 PM


sam-pi:
@joecummings Please take a look and let me know if you have feedback!

recipes/full_dpo_distributed.py
```diff
@@ -0,0 +1,952 @@
...
+                del policy_chosen_logits, policy_rejected_logits
+
+                with torch.no_grad():
+                    (
+                        reference_chosen_log_probs,
+                        reference_rejected_log_probs,
+                        _,
+                        _,
+                    ) = self.concatenated_forward(self._ref_model, batch)
+                assert not reference_chosen_log_probs.requires_grad
```
SalmanMohammadi, line 837: I'm not sure if we need these here.

recipes/full_dpo_distributed.py
```diff
@@ -0,0 +1,952 @@
...
+        if self._is_rank_zero:
+            log.info(
+                f"Instantiating model and loading checkpoint took {time.perf_counter() - init_start:.2f} secs"
+            )
+            memory_stats = training.get_memory_stats(device=self._device)
+            training.log_memory_stats(memory_stats)
+
+        # synchronize before training begins
+        torch.distributed.barrier()
+
```
SalmanMohammadi, line 532: You'll also want to make sure dropout is disabled in both models.

SalmanMohammadi:
Hey @sam-pi! Thanks so much for adding this. I had a quick skim through and it looked good to me. I'll have a closer look soon. First, a couple of high level points.

Did you manage to train using configs? If so, could you attach some evidence of successful runs (e.g. WandB links)? 

I'm particularly interested in the hardware requirements for the 70B config. We may want to think about offering some additional memory performance improvements for this recipe in particular, such as different parallelization configurations for the reference model (which doesn't need gradients or activations to be sharded), offloading the entire reference model to CPU, etc.

recipes/full_dpo_distributed.py
```diff
@@ -0,0 +1,952 @@
...
+            enable_activation_checkpointing=self._enable_activation_checkpointing,
+            enable_activation_offloading=self._enable_activation_offloading,
+            custom_sharded_layers=cfg.get("custom_sharded_layers", None),
+            fsdp_cpu_offload=cfg.get("fsdp_cpu_offload", False),
+            reshard_after_forward=cfg.get("fsdp_reshard_after_forward", True),
+            model_state_dict=checkpoint_dict[training.MODEL_KEY],
+            ac_mode=cfg.get("ac_mode", None),
+            ac_option=cfg.get("ac_option", None),
+        )
+        self._ref_model = self._setup_model(
```
SalmanMohammadi, line 292: We may want to consider a more bespoke way to setup the reference model here. For example, we don't need to do things like `enable_activation_offloading=self._enable_activation_offloading`, as this should always be set to `False` for the reference model, which won't be storing intermediate activations under `no_grad`. 

I've left a more general comment about this - the gist is that we don't necessarily wish to apply the same parallelization and memory optimization strategies to both the reference and policy models.

sam-pi: Ah yes good point, do you think then I can just update to the following, or should there be more generalization?

```python
        self._ref_model = self._setup_model(
            cfg_model=cfg.model,
            custom_sharded_layers=cfg.get("custom_sharded_layers", None),
            fsdp_cpu_offload=cfg.get("fsdp_cpu_offload", False),
            reshard_after_forward=cfg.get("fsdp_reshard_after_forward", True),
            model_state_dict=self.load_ref_states(cfg.ref_checkpointer),
        )
```

Also, I found that for 70B it was necessary to set `reshard_after_forward=True` or else the entire model seemed to be loaded per node/rank.

recipes/full_dpo_distributed.py
sam-pi: Ok will do, following this pattern in the PPO recipe? https://github.com/pytorch/torchtune/blob/main/recipes/ppo_full_finetune_single_device.py#L569

sam-pi:
@SalmanMohammadi Please take a look at my training run screenshots and configs at the bottom of the PR summary. I tried showing a comparison of a rank/alpha 256 lora dpo run against a full dpo run (only 100 iterations).
For Llama3.1-70B-Instruct, I was able to run using 2 nodes with 8x H100 GPUs (I think this is just 2x the HW requirements for running a single non-quantized 70B).

recipes/full_dpo_distributed.py
SalmanMohammadi: Exactly! Though if you have a neater way of doing it feel free : )

recipes/full_dpo_distributed.py
SalmanMohammadi: > Also, I found that for 70B it was necessary to set reshard_after_forward=True or else the entire model seemed to be loaded per node/rank.

Was this for the reference or the policy model?

> Ah yes good point, do you think then I can just update to something like the following, or should there be more generalization?

I think having a seperate method is good! I'll try get round to testing out some different configs this weekend. Would you mind if I pushed changes to your branch?

recipes/configs/llama3_1/70B_full_dpo.yaml
```diff
@@ -0,0 +1,92 @@
...
+# this run:
+#   tune download meta-llama/Meta-Llama-3.1-70B-Instruct --output-dir /tmp/Meta-Llama-3.1-70B-Instruct --ignore-patterns "original/consolidated.00.pth"
+#
+# To launch on 2 devices, run the following command from root:
+#   tune run --nnodes 1 --nproc_per_node 2 full_dpo_distributed --config llama3_1/70B_full_dpo
+#
+# You can add specific overrides through the command line. For example
+# to override the checkpointer directory while launching training
+# you can run:
+#   tune run --nnodes 1 --nproc_per_node 2 full_dpo_distributed --config llama3_1/70B_full_dpo checkpointer.checkpoint_dir=<YOUR_CHECKPOINT_DIR>
```
SalmanMohammadi, line 14: Since you mentioned you trained on 2 nodes it'd be good to add the command you used here. 

Seperately, I'm going to try see if I can find a config that can train on a single node with reasonable speeds.

recipes/full_dpo_distributed.py
sam-pi: That is for the reference model. Ok thanks, and no problem for you to push changes to my branch.

recipes/full_dpo_distributed.py
sam-pi: I updated to remove these assertions!

recipes/full_dpo_distributed.py
sam-pi: Thanks, I added in that same approach

recipes/full_dpo_distributed.py
sam-pi: I added a first cut of this, just re-using the _setup_model method wrapped in _setup_reference_model.

recipes/configs/llama3_1/70B_full_dpo.yaml
sam-pi: I looked into running this on 1 node and I couldn't find a way to get it to fit - if you do please feel free to update. Otherwise, maybe it's not worth including this in the PR since technically I only got this working using sbatch and torchrun with --nnodes 2.

EugenHotaj:
Any updates on merging this to main? Really excited to use it 😄

recipes/full_dpo_distributed.py
```diff
@@ -0,0 +1,970 @@
...
+        """
+        self._checkpointer = config.instantiate(
+            cfg_checkpointer,
+            resume_from_checkpoint=self._resume_from_checkpoint,
+        )
+        checkpoint_dict = self._checkpointer.load_checkpoint()
+
+        if self._resume_from_checkpoint:
+            self._update_recipe_state(checkpoint_dict)
+        return checkpoint_dict
```
EugenHotaj, line 205: Heads up that this is never called so checkpoint saving will fail since `self._checkpointer` will never be set.

recipes/full_dpo_distributed.py
```diff
@@ -0,0 +1,970 @@
...
+                            .cpu(),
+                            "log_probs/rejected": policy_rejected_log_probs.detach()
+                            .mean()
+                            .cpu(),
+                            "log_probs/chosen": policy_chosen_log_probs.detach()
+                            .mean()
+                            .cpu(),
+                            "logits/rejected": policy_rejected_logits_mean.cpu(),
+                            "logits/chosen": policy_chosen_logits_mean.cpu(),
+                        }
```
EugenHotaj, line 909: I think we need to `allreduce` these across the data ranks before logging. Right now we're just logging rank_0 metrics afaict.

recipes/full_dpo_distributed.py
SalmanMohammadi: Thanks! I was in the middle of updating the checkpointing logic. Will wrap it up soon.

recipes/full_dpo_distributed.py
```diff
@@ -0,0 +1,970 @@
...
+
+                with torch.no_grad():
+                    (
+                        reference_chosen_log_probs,
+                        reference_rejected_log_probs,
+                        _,
+                        _,
+                    ) = self.concatenated_forward(self._ref_model, batch)
+
+                loss, chosen_rewards, rejected_rewards = self._loss_fn(
```
EugenHotaj, line 856: Another heads up: we log these below but we're not taking GAS into account.

(lmk if these comments are unhelpful btw and I'll stop 🙂  -- just trying to get this PR to run / verify on our setup and commenting as I find discrepancies)

recipes/full_dpo_distributed.py
```diff
@@ -0,0 +1,970 @@
...
+            self._log_peak_memory_stats = False
+
+        _, rank = get_world_size_and_rank()
+        self._is_rank_zero = rank == 0
+
+        # Training cfg
+        self._resume_from_checkpoint = cfg.resume_from_checkpoint
+        self._gradient_accumulation_steps = cfg.gradient_accumulation_steps
+        self._optimizer_in_bwd = cfg.get("optimizer_in_bwd", False)
+        self._clip_grad_norm = cfg.get("clip_grad_norm", None)
```
EugenHotaj, line 142: Looks like we're missing the actual grad clipping logic in the train step.

EugenHotaj:
With the changes I mentioned in the comments I was able to get parity with [NeMo](https://github.com/NVIDIA/NeMo)'s DPO using the same data / hparams. E.g. here's the loss curves:

Screenshot 2025-01-29 at 5 06 40 PM

Really awesome work! Pretty excited to use this.

recipes/full_dpo_distributed.py
SalmanMohammadi: > (lmk if these comments are unhelpful btw and I'll stop 🙂 -- just trying to get this PR to run / verify on our setup and commenting as I find discrepancies)

Not at all, your comments are incredibly helpful and more than welcome! Thanks for taking the time to help review.

SalmanMohammadi: > Another heads up: we log these below but we're not taking GAS into account.

noob q: what's GAS?

SalmanMohammadi:
> Any updates on merging this to main? Really excited to use it 😄

I'm going to try out some more efficient sharding strategies for the reference model (we shouldn't need FSDP), and see if I can get single-node training working for 70B. Will update soon. @sam-pi would you be up for looking into @EugenHotaj's comments above?

recipes/full_dpo_distributed.py
EugenHotaj: Gradient Accumulation Steps

SalmanMohammadi:
OK so we're not blocking this PR I'm going to leave exploring different parallelism strategies for a follow-up. Let's make the necessary fixes to this recipe and bring it in line with our other distributed recipes.

@sam-pi If the 70B config doesn't work on a single node, I'd also suggest we remove it for now and add it back in after patching in the changes from #2301. What do you think?

recipes/full_dpo_distributed.py
SalmanMohammadi: Yeah you're totally right. We should update to correct for gradient accumulation steps.

recipes/full_dpo_distributed.py
```diff
@@ -0,0 +1,1037 @@
...
+            Tuple of chosen log probs, rejected log probs, chosen logits, rejected logits.
+        """
+        concatenated_input_ids, concatenated_labels = batch
+        concatenated_input_ids = concatenated_input_ids.to(self._device)
+        concatenated_labels = concatenated_labels.to(self._device)
+
+        # formed by concatenating an equal number of "chosen" and "rejected".
+        len_chosen = concatenated_input_ids.shape[0] // 2
+
+        all_logits = model(concatenated_input_ids)
```
EugenHotaj, line 845: One way to reduce memory and potentially fit this on a single node is to call `model(...)` twice. Right now we're effectively doubling the batch size here, and might be causing OOMs.

sam-pi:
> OK so we're not blocking this PR I'm going to leave exploring different parallelism strategies for a follow-up. Let's make the necessary fixes to this recipe and bring it in line with our other distributed recipes.
> 
> @sam-pi If the 70B config doesn't work on a single node, I'd also suggest we remove it for now and add it back in after patching in the changes from #2301. What do you think?

Thanks, I will look into all these fixes today and also remove the 70B config for now

recipes/full_dpo_distributed.py
SalmanMohammadi: One scenario I can think of is if a user is OOMing and can't go below an effective batch size of 2 (by configuring a batch size of 1). I'd be interested in seeing the tradeoff here vs. the additional computation from two extra model forward passes (both the policy and reference models) - though I have a feeling the memory savings may not be worth it.

recipes/full_dpo_distributed.py
```diff
@@ -0,0 +1,1037 @@
...
+            training.log_memory_stats(memory_stats)
+
+        # disabling dropout if found - non-determinism leads to issues in e.g. comparing logprobs
+        # between ref policy and current policy
+        for module in model.modules():
+            if isinstance(module, torch.nn.Dropout):
+                warn(
+                    f"Dropout found in {module}. This is likely to cause issues during training. Disabling."
+                )
+                module.p = 0
```
ebsmothers, line 520: This might be useful as a training utility, e.g. `disable_dropout(model)`

recipes/full_dpo_distributed.py
```diff
@@ -0,0 +1,1037 @@
...
+        """
+        if self._is_rank_zero:
+            self._metric_logger = config.instantiate(cfg.metric_logger)
+
+            # log config with parameter override
+            self._metric_logger.log_config(cfg)
+
+        # Load the base model
+        checkpoint_dict = self._load_checkpoint(cfg.checkpointer)
+        ref_checkoint_dict = self._load_ref_checkpoint(cfg.ref_checkpointer)
```
ebsmothers, line 268: nit: typo (and used again in L288)
```suggestion
        ref_checkpoint_dict = self._load_ref_checkpoint(cfg.ref_checkpointer)
```

recipes/full_dpo_distributed.py
```diff
@@ -0,0 +1,1037 @@
...
+            model_state_dict=checkpoint_dict[training.MODEL_KEY],
+            ac_mode=cfg.get("ac_mode", None),
+            ac_option=cfg.get("ac_option", None),
+        )
+
+        # TODO (@SalmanMohammadi) investigate TP for ref model
+        self._ref_model = self._setup_reference_model(
+            cfg_model=cfg.model,
+            fsdp_cpu_offload=cfg.get("fsdp_cpu_offload", False),
+            reshard_after_forward=cfg.get("fsdp_reshard_after_forward", True),
```
ebsmothers, line 287: It may be worth running some memory profiling on this recipe (especially since it's already enabled). Like it seems to me that by setting `reshard_after_forward=True` here, we never have both the reference model and the policy model weights gathered at the same time. Is that the correct understanding? If so, worth confirming that it happens in practice (especially given the discussion around fitting on a single node)

recipes/full_dpo_distributed.py
```diff
@@ -0,0 +1,1037 @@
...
+                        policy_rejected_logits,
+                    ) = self.concatenated_forward(self._model, batch)
+
+                policy_chosen_logits_mean = policy_chosen_logits.detach().mean()
+                policy_rejected_logits_mean = policy_rejected_logits.detach().mean()
+
+                # deleting logits here helps reduce (peak) memory usage - we only need them for metric logging
+                del policy_chosen_logits, policy_rejected_logits
+
+                with torch.no_grad():
```
ebsmothers, line 915: I wonder whether we can just run in inference mode? Cause reference model never has grad updates so stuff like view tracking etc afforded by `no_grad` shouldn't be relevant, right? Lmk if I'm way off base, otherwise worth a try imo.

recipes/full_dpo_distributed.py
ebsmothers: In that case I assume the same holds for the LoRA DPO recipe too, right?

recipes/full_dpo_distributed.py
```diff
@@ -0,0 +1,1037 @@
...
+                loss = loss.mean()
+                reward_accuracies = (chosen_rewards > rejected_rewards).float()
+
+                loss = loss / self._gradient_accumulation_steps
+                running_loss += loss
+                loss.backward()
+
+                # Step with optimizer
+                if (idx + 1) % self._gradient_accumulation_steps == 0:
+                    self._optimizer.step()
```
ebsmothers, line 939: I don't think we've actually enabled optimizer in backward here either

recipes/full_dpo_distributed.py
```diff
@@ -0,0 +1,1037 @@
...
+
+                    # Log per-step metrics
+                    if (
+                        self.global_step % self._log_every_n_steps == 0
+                        and self._is_rank_zero
+                    ):
+                        time_per_step = time.perf_counter() - t0
+                        log_dict = {
+                            "loss": loss_to_log,
+                            "lr": self._optimizer.param_groups[0]["lr"],
```
ebsmothers, line 960: I believe we also have [this utility](https://github.com/pytorch/torchtune/blob/6764618d2f148b28d2cb506f5ce70bf213fa1c3a/torchtune/training/lr_schedulers.py#L61) now in case that's helpful here

ebsmothers (commented):
This is looking great, thanks so much for adding this @sam-pi! Aside from my inline comments it'd be good to confirm that various features like compile, optimizer-in-backward, etc are working and doing what we'd expect (we can even add e.g. compile to the recipe test)

tests/recipes/test_full_dpo_distributed.py
```diff
@@ -0,0 +1,209 @@
...
+        tokenizer_path = Path(TOKENIZER_PATHS["llama3"])
+
+        # Config file needed for model conversion.
+        # Create a second copy for training resume
+        write_hf_ckpt_config(ckpt_dir)
+        write_hf_ckpt_config(tmpdir)
+
+        # Train for two epochs
+        cmd_1 = f"""
+        tune run --nnodes 1 --nproc_per_node 1 full_dpo_distributed \
```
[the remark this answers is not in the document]
sam-pi: Ah good point, I more or less copied this from the lora DPO testing - I will look into updating it

tests/recipes/test_full_dpo_distributed.py
```diff
@@ -0,0 +1,209 @@
...
+        runpy.run_path(TUNE_PATH, run_name="__main__")
+
+        expected_loss_values = get_loss_values_from_metric_logger(log_file)
+
+        resumed_log_dir = (tmpdir / "resumed/").mkdir()
+        resumed_log_file = gen_log_file_name(resumed_log_dir)
+
+        # Resume training
+        # epoch_folder = get_largest_iter_folder(tmpdir)
+        # epoch_folder_minus_one = f"epoch_{int(epoch_folder.split('_')[-1]) - 1}"
```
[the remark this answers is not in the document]
sam-pi: I will remove, thanks for catching that. These are copied from the lora DPO tests

tests/recipes/test_full_dpo_distributed.py
```diff
@@ -0,0 +1,209 @@
...
+
+        # Second epoch only
+        resumed_loss_values = get_loss_values_from_metric_logger(resumed_log_file)
+
+        torch.testing.assert_close(
+            resumed_loss_values[:2], expected_loss_values[2:], rtol=1e-5, atol=1e-5
+        )
+
+    @pytest.mark.integration_test
+    def test_save_and_load_weights(self, tmpdir, monkeypatch):
```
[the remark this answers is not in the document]
sam-pi: I was guessing it's good practice to make sure save/load works in general, but happy to remove if it doesn't make sense.

recipes/full_dpo_distributed.py
sam-pi: I thought that by running model.eval() we are setting inference mode - am I mistaken?

recipes/full_dpo_distributed.py
```diff
@@ -0,0 +1,1037 @@
...
+            ac_mode=cfg.get("ac_mode", None),
+            ac_option=cfg.get("ac_option", None),
+        )
+
+        # TODO (@SalmanMohammadi) investigate TP for ref model
+        self._ref_model = self._setup_reference_model(
+            cfg_model=cfg.model,
+            fsdp_cpu_offload=cfg.get("fsdp_cpu_offload", False),
+            reshard_after_forward=cfg.get("fsdp_reshard_after_forward", True),
+            model_state_dict=ref_checkoint_dict,
```
SalmanMohammadi, line 288:
```suggestion
            model_state_dict=ref_checkpoint_dict,
```
my b

recipes/full_dpo_distributed.py
SalmanMohammadi: @sam-pi you mentioned you had to set this to True, right? What did you find?

recipes/full_dpo_distributed.py
```diff
@@ -0,0 +1,1037 @@
...
+        # the older version of AC and this behavior is unchanged
+        # ac_mode and ac_option together control selective AC. This is only enabled
+        # when these are set AND ``enable_activation_checkpointing`` is set to False
+        # We'll clean this up as soon as testing of AC is complete
+        if (not enable_activation_checkpointing) and (ac_mode is not None):
+            apply_selective_activation_checkpointing(
+                model,
+                ac_mode,
+                ac_option,
+            )
```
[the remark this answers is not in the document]
SalmanMohammadi: I agree. Though if we're kind-of staging early APIs in our "canonical" recipes it makes copying them to add new recipes slightly more confusing.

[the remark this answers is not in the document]
SalmanMohammadi: I think we'll need to remove all references to `ac_mode` (including in the configs, if it's there) to make this change @sam-pi.

recipes/full_dpo_distributed.py
ebsmothers: > I've had issues with inference mode causing additional recompiles in the inlined compiled transformer layers.

Ah fair enough. Yeah mainly I was thinking it could potentially give some minor speedups. But if it messes with compile then agree it's not worth it

tests/recipes/test_full_dpo_distributed.py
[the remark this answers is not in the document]
ebsmothers: Yeah mainly I am cognizant of us not having too too many recipe tests (they take a bit of time to run and run on every PR). In this case I claim save and load is already pretty well-covered by `test_training_state_on_resume`

recipes/full_dpo_distributed.py
sam-pi: I was finding if this was set to False I was getting OOM issues. I didn't yet debug further than note that it just works for me when set to True.

recipes/configs/llama3_1/70B_full_dpo.yaml
sam-pi: Thanks, I removed the 70B config for now! Fair point on using 3.3 - I stuck to 3.1 to keep it simple for now and I hope it could be adapted relatively easily to 3.3.

recipes/full_dpo_distributed.py
sam-pi: I think that's a great idea - but I will plan to leave it out of this PR if that's alright. That seems like a useful addition to all DPO recipes

recipes/full_dpo_distributed.py
```diff
@@ -0,0 +1,1033 @@
...
+                    f"Dropout found in {module}. This is likely to cause issues during training. Disabling."
+                )
+                module.p = 0
+
+        # synchronize before training begins
+        torch.distributed.barrier()
+
+        return model
+
+    def _setup_reference_model(
```
EugenHotaj, line 527: Is there any reason we're not doing ac for the ref model?

BTW this also seems like an almost exact duplicate with `_setup_model()` except for a few extra calls at the bottom, maybe we can reuse much of the code?

sam-pi: That may be a misunderstanding on my part by removing ac, and I can put it back if that makes sense. @SalmanMohammadi may have thoughts on keeping it separate/duplicated to allow for future updates like running the ref model with TP instead of FSDP.

SalmanMohammadi: > Is there any reason we're not doing ac for the ref model?

Activation checkpointing is primarily used to save memory for models which require gradient updates. Rather than keep the activations alive until backprop, activation checkpointing re-computes the intermediate activations on the backward pass.

Since we're not computing gradients for the reference model, it's activations will be discarded when they're no longer needed in the forward pass, so activation-based memory saving techniques don't apply here. Correct me if I've misunderstood anything here though : )

I also found the separate function helpful when I was trying out different sharding strategies, but I agree it's pretty similar now. I'm fine to merge it back together for now and hardcode the additional parameters - it'll be neater once we remove all the `ac_mode` stuff.

EugenHotaj: > it's activations will be discarded when they're no longer needed in the forward pass, so activation-based memory saving techniques don't apply here

Yea good point, I guess this is the main thing I am concerned about. Does PyTorch automatically do this when you don't need to compute grads?

SalmanMohammadi: I believe so. The garbage collector will kick in when the ref count for the activation tensors hits zero.

recipes/full_dpo_distributed.py
```diff
@@ -0,0 +1,1037 @@
...
+                raise RuntimeError(
+                    "enable_activation_offloading should only be True when training on CUDA"
+                )
+            if not self._enable_activation_checkpointing:
+                raise RuntimeError(
+                    "enable_activation_offloading should only be True when enable_activation_checkpointing is True"
+                )
+        elif (
+            self._enable_activation_checkpointing
+            and cfg.checkpointer.model_type != "LLAMA3_VISION"
```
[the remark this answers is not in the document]
SalmanMohammadi: I think we can try and fail early here if the users attempts to use this recipe with a vision model.

recipes/full_dpo_distributed.py
```diff
@@ -0,0 +1,1033 @@
...
+
+                # deleting logits here helps reduce (peak) memory usage - we only need them for metric logging
+                del policy_chosen_logits, policy_rejected_logits
+
+                with torch.no_grad():
+                    (
+                        reference_chosen_log_probs,
+                        reference_rejected_log_probs,
+                        _,
+                        _,
```
EugenHotaj, line 916: One final heads up: I was getting OOMs on 70B unless I deleted these logits as well. I also had to call `gc.collect(); torch.cuda.clear_cache()` otherwise I'd OOM mid-training sometimes. 

Since all we're doing with the logits is calling `.mean()` below, we could return the means directly from `concatenated_forward`. Then you don't need to do any explicit deletion (but maybe still have to call collect(), clear_cache().

sam-pi:
@ebsmothers @EugenHotaj @SalmanMohammadi Please take a look at the updates from @bogdansalyp to fix metric syncing/averaging across ranks and accounting for gradient accumulation in metrics.

recipes/full_dpo_distributed.py
SalmanMohammadi: @sam-pi relevant code snippet to enable https://github.com/pytorch/torchtune/blob/64870293c7e2c195166a85e179638cb41d848d9a/recipes/ppo_full_finetune_single_device.py#L1007

recipes/full_dpo_distributed.py
sam-pi: Thanks, I added in deleting the logits for now at least

torchtune/_recipe_registry.py
```diff
@@ -325,6 +325,21 @@ class Recipe:
...
     ),
+    Recipe(
+        name="full_dpo_distributed",
+        file_path="full_dpo_distributed.py",
+        configs=[
+            Config(
+                name="llama3_1/8B_full_dpo",
+                file_path="llama3_1/8B_full_dpo.yaml",
+            ),
+            Config(
```
SalmanMohammadi, line 336: This can now be removed.

sam-pi: Done!

recipes/full_dpo_distributed.py
```diff
@@ -0,0 +1,1060 @@
...
+            our checkpointer deepdive (https://pytorch.org/torchtune/main/deep_dives/checkpointer.html).
+
+        - Gradient Clipping. Gradient clipping is supported using the ``clip_grad_norm`` flag. By default,
+            ``clip_grad_norm`` is set to ``None``. If you only want to log the grad norm, you can set
+            ``clip_grad_norm='inf'``.
+
+        - Logging. Terminal, Disk, WandB and TensorBoard are all supported.
+
+    The following losses are supported in this recipe:
+        - :class:`~torchtune.modules.rlhf.loss.DPOLoss`: Direct Preference Optimization (DPO).
```
SalmanMohammadi, line 98:
```suggestion
        - :class:`~torchtune.modules.rlhf.loss.DPOLoss`: Direct Preference Optimization (DPO).
        - :class:`~torchtune.rlhf.loss.RSOPLoss`: Rejection Sampling Optimization (RSO).
        
```

recipes/full_dpo_distributed.py
```diff
@@ -0,0 +1,1060 @@
...
+                )
+        elif (
+            self._enable_activation_checkpointing
+            and cfg.checkpointer.model_type != "LLAMA3_VISION"
+        ):
+            utils.log_rank_zero(
+                log,
+                "Hint: enable_activation_checkpointing is True, but enable_activation_offloading isn't. "
+                "Enabling activation offloading should reduce memory further.",
+            )
```
SalmanMohammadi, line 180:
```suggestion
        elif self._enable_activation_checkpointing:
            utils.log_rank_zero(
                log,
                "Hint: enable_activation_checkpointing is True, but enable_activation_offloading isn't. "
                "Enabling activation offloading should reduce memory further.",
            )
```

SalmanMohammadi:
@sam-pi Last few comments but this looks great! 

Did you happen to run the 8B recipe long enough to see convergence?

recipes/configs/llama3_1/8B_full_dpo.yaml
```diff
@@ -0,0 +1,99 @@
...
+# To launch on 4 devices, run the following command from root:
+#   tune run --nnodes 1 --nproc_per_node 4 full_dpo_distributed --config llama3_1/8B_full_dpo
+#
+# You can add specific overrides through the command line. For example
+# to override the checkpointer directory while launching training
+# you can run:
+#   tune run --nnodes 1 --nproc_per_node 4 full_dpo_distributed --config llama3_1/8B_full_dpo checkpointer.checkpoint_dir=<YOUR_CHECKPOINT_DIR>
+#
+# This config works best when the model is being fine-tuned on 2+ GPUs.
+# For single device full DPO alignment please use  llama3_1/8B_full_dpo_single_device
```
SalmanMohammadi, line 17:
```suggestion
# This config works best when the model is being fine-tuned on 2+ GPUs.
# For single device full DPO alignment please use  llama3_1/8B_full_dpo_single_device
```
We don't have a single device DPO FFT recipe, right?

sam-pi: That's right, although I believe with some minor tweaks this same recipe could be used for single device (likely with a smaller model)

bogdansalyp:
> @sam-pi Last few comments but this looks great!
> 
> Did you happen to run the 8B recipe long enough to see convergence?
> 
> edit: could you also try the recipe out with `compile: True` and compare the tokens/s charts vs. baseline? I have a suspicion the cold-start time might be higher than usual.

Yes, it looks good with the config @sam-pi just added to the PR

image
image

sam-pi:
> @sam-pi Last few comments but this looks great!
> 
> Did you happen to run the 8B recipe long enough to see convergence?
> 
> edit: could you also try the recipe out with `compile: True` and compare the tokens/s charts vs. baseline? I have a suspicion the cold-start time might be higher than usual.

Ok I have updated the docs with a small change, and please let me know if there is anything else! Here is a direct comparison of the run Bogdan showed (using the config as-is with compile=False) and setting compile=True:
Screenshot 2025-02-04 at 11 06 10 AM

docs/source/recipes/dpo.rst
```diff
@@ -13,7 +13,7 @@ To see the best results when using this recipe, it may be helpful to first fine-
 on-distribution for the domain you're interested in. To do this, check out our other fine-tuning recipes in the :ref:`recipe overview <recipes_overview_label>` which
 support a variety of SFT paradigms.
 
-After supervised fine-tuning, here is an example of DPO with Llama 3.1 8B:
+After supervised fine-tuning, here is an example of Lora DPO and Full DPO with Llama 3.1 8B:
```
SalmanMohammadi, line 16:
```suggestion
After supervised fine-tuning, here is an example of using either LoRA-based finetuning, or full-finetuning Llama 3.1 8B with DPO:
```

recipes/full_dpo_distributed.py
SalmanMohammadi: bump

recipes/full_dpo_distributed.py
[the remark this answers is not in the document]
SalmanMohammadi: bump

recipes/full_dpo_distributed.py
SalmanMohammadi: bump

sam-pi: I started looking into this but can see the `if self._optimizer_in_bwd:` in a few places (especially in full_finetune_distributed) - could this make sense in a new PR which also fixes the lora DPO recipes?

recipes/full_dpo_distributed.py
sam-pi: I updated to account for this!

recipes/full_dpo_distributed.py
sam-pi: Fixed!

EugenHotaj:
The rewards being different between compile and non-compile is also a bit suspicious. I'm surprised we're not able to get fully deterministic results between compile / non-compile. Maybe we're forgetting to set the seed somewhere?

ebsmothers:
@EugenHotaj compile is awesome 😃. 2x is definitely not unheard of, in some cases I've seen gains even larger than that. I agree that the large delta in rewards is suspicious. Compile tends to not achieve exact numerical parity, but we should also rule out something like #2335. Once we verify the delta is actually coming from compile, we can loop in some compiler folks to help out here.

sam-pi:
> The loss starting at 5.5 seems off? We might not be normalizing correctly somewhere.
> 
> Also wow, 2x improvement in tps with compile! Is this pretty common / expected? I thought compile would give maybe 10-20% extra boost but 2x is awesome!

Note we were running these with 8xH100 GPUs - otherwise the same default config. With 4xH100 GPUs the loss starts closer to 2.7 and with 2xH100 the loss starts around 1.4.

EugenHotaj:
> Note we were running these with 8xH100 GPUs - otherwise the same default config. With 4xH100 GPUs the loss starts closer to 2.7 and with 2xH100 the loss starts around 1.4.

Yea so we probably forgot to normalize by dp rank. Loss should not change with the number of nodes we add.

sam-pi:
> > Note we were running these with 8xH100 GPUs - otherwise the same default config. With 4xH100 GPUs the loss starts closer to 2.7 and with 2xH100 the loss starts around 1.4.
> 
> Yea so we probably forgot to normalize by dp rank. Loss should not change with the number of ranks we have.

Ok I added in all_reduce with average, and it works! I can see starting loss of 0.69 for 2x and 8x H100s

recipes/full_dpo_distributed.py
bogdansalyp: @SalmanMohammadi wait but it's already accounted for
metrics are scaled for grad_acc steps -> https://github.com/pytorch/torchtune/pull/2275/files#diff-ca64e06ed8af5f389cf3ecfc1fa0eda295e21c4248d44bdfb69992d57368033cR728
and loss is also averaged out -> https://github.com/pytorch/torchtune/pull/2275/files#diff-ca64e06ed8af5f389cf3ecfc1fa0eda295e21c4248d44bdfb69992d57368033cR724

recipes/full_dpo_distributed.py
sam-pi: Thanks for adding it in!

SalmanMohammadi (approved):
LGTM. I'll let @ebsmothers have one final look - please seek his approval before merging.

Thanks so much for this contribution and your patience through the review process @sam-pi @EugenHotaj @bogdansalyp!

ebsmothers (commented):
Thanks @sam-pi, @bogdansalyp and others for your patience on this one! I do have a few more questions but I think we're close here. The two main ones are making sure we're applying optimizer-in-backward and activation offloading in the right way. Especially for optimizer-in-backward I'm not sure the loss is being calculated in a consistent way. I mentioned this in an inline comment, but open to dropping this from the recipe altogether if you feel it will make things more straightforward. Modulo these last few comments I think we are in good shape here.

recipes/full_dpo_distributed.py
```diff
@@ -0,0 +1,1151 @@
...
+            for idx, batch in enumerate(self._dataloader):
+                if (
+                    self.max_steps_per_epoch is not None
+                    and (idx // self._gradient_accumulation_steps)
+                    == self.max_steps_per_epoch
+                ):
+                    break
+
+                # batch is input_ids, labels
+                with self.activations_handling_ctx:
```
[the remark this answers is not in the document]
sam-pi: Ah good point, that should be updated I think. I can add that now

recipes/full_dpo_distributed.py
```diff
@@ -0,0 +1,1151 @@
...
+                    torch.distributed.all_reduce(num_tokens)
+                    torch.distributed.all_reduce(
+                        running_loss, op=torch.distributed.ReduceOp.AVG
+                    )
+                    for key in running_metrics:
+                        torch.distributed.all_reduce(
+                            running_metrics[key], op=torch.distributed.ReduceOp.AVG
+                        )
+                    # We multiply by world_size to undo FSDP2 gradient normalization.
+                    loss = loss * (world_size / num_tokens)
```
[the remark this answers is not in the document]
sam-pi: Note I reverted commits 893398e b293ce3 9c71083 4fd18bf 6781894 46c59ec eef1b01 716efca

This removes optimizer_in_bwd updates for now, along with a few minor update to use log_rank_zero.

recipes/full_dpo_distributed.py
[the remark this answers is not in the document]
sam-pi: I just updated in my latest commit so activations_handling_ctx is optionally applied with an additional optional bool arg. This appears to show ~20% speedup in tokens per second @ebsmothers

ebsmothers (approved):
Thanks so much @sam-pi, @bogdansalyp, @SalmanMohammadi for the team effort on this one! Once CI is green I think we're good to go here.

PR #2276, opened 2025-01-17
Construct EarlyFusion's encoder_token_ids on correct device

ebsmothers:
Fix a bug where encoder_token_ids are not placed on the proper device


acisseJZhong (approved):
thanks for fixing!

RdoubleA (approved):

Issue #1708, opened 2024-09-28
ConcatDataset allows Packed sub datasets


pbontrager:
Currently we don’t allow data packing with concat dataset. We can discuss enabling that, but in the meantime there’s a big in recipes wheee a user can set packed =True to each dataset in the list . This will individually pack each one but not use the right collate function, also, all of the datasets in the list should probably be packed together. To fix this for now, we can have ConcatDataset check that there are no packed datasets in the list.

krammnic:
I see that this issue without "community welcome" label, but because it became little bit stale here are few thoughts. Probably something like:

```
for dataset in self._datasets:
    if isinstance(dataset, PackedDataset):
        raise ...
```

While initializing ConcatDataset will solve it, but knowing that we can import PackedDataset to ConcatDataset due to circular import, probably the better idea is to do something from this list: 
1. Introduce some is_packed: bool property
2. Add utility to check this condition in utils(looks like bad idea)

So final solution will consist of:
1. Do some of the checks described above
2. Add unit test 
3. Add some information in docs?

If you not mind, I might try to do a PR.

RdoubleA:
There shouldn't be a circular import if ConcatDataset imports PackedDataset, as `_packed.py` does not currently depend on `_concat.py`. Would prefer doing the check within ConcatDataset, otherwise you would have to update all recipes.

krammnic:
Ok, will recheck this, as I see torchtune does not have any custom exceptions, so I assume that raising some `ValueError` will be fine. Will create PR soon for this

PR #2278, opened 2025-01-17
Sample packing for ConcatDataset

ebsmothers:
Currently we error when any individual datasets in `ConcatDataset` have `packed=True`. This goes back to #1708:  because packed and unpacked datasets require different collators and packing is done on the individual datasets rather than the `ConcatDataset`, we can't really guarantee a single collator for `ConcatDataset` is well-defined in all cases. 

Fortunately, it seems pretty likely that someone who is enabling packing on one dataset would want to do so on all of them. And it's actually trivial to support this case. So this PR relaxes the check in `ConcatDataset` to also allow the case that every dataset is packed.

A couple things that can be revisited here:

1) Given that dataset packing in `ConcatDataset` is either all or nothing, it's probably more natural to define the `packed` attribute there instead. However, then we get into questions of whether we pack before or after merging. Relatedly, 
2) this implementation packs each individual dataset and then merges. This means that we do not support packs combining samples from different datasets. 

## Test plan:

Added a unit test:

```
pytest tests/torchtune/datasets/test_concat_dataset.py
...
======== 5 passed in 0.10s ==========
```

Also ran one of the updated recipes with the following config updates:

```
tokenizer:
	max_seq_len: 512


dataset:
		packed: True
	- _component_: torchtune.datasets.alpaca_cleaned_dataset
		packed: True
```

(Per HF datasets page, Alpaca dataset has 52k samples, Alpaca cleaned has 51.8k samples)

Printing the length of different dataset properties at different points:

```
# Inside PackedDataset, prior to packing
>>> len(self.ds) # alpaca dataset
52002
>>> len(self.ds) # alpaca-cleaned dataset
51760

# Inside PackedDataset, after packing
>>> len(self) # alpaca dataset
13348
>>> len(self) # alpaca-cleaned dataset
 25891

# Inside ConcatDataset, after construction
>>> len(self)
39239
>>> self._indexes
[(0, 13348, 0), (13348, 39239, 1)]
```

### DIFF
--- a/tests/torchtune/datasets/test_concat_dataset.py
+++ b/tests/torchtune/datasets/test_concat_dataset.py
@@ -80,7 +80,7 @@ class TestConcatDataset:
         with pytest.raises(TypeError):
             multi_dataset["invalid_type"]  # Non-integer index
 
-    def test_packed_dataset(self, torch_datasets):
+    def test_single_packed_dataset(self, torch_datasets):
         torch_datasets[0] = PackedDataset(
             torch_datasets[0],
             max_seq_len=25,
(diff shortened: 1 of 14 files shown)


tests/torchtune/datasets/test_concat_dataset.py
```diff
@@ -90,3 +90,33 @@ def test_packed_dataset(self, torch_datasets):
 
         with pytest.raises(ValueError):
             concated_dataset = ConcatDataset(torch_datasets)
+
+    def test_all_packed_datasets(self, torch_datasets):
```
RdoubleA, line 94: should also test error is caught when some datasets are unpacked?

RdoubleA (approved):

tests/torchtune/datasets/test_concat_dataset.py
ebsmothers: Oh yeah it already exists in the previous unit test

Issue #2270, opened 2025-01-16
Don't use _get_clones


ebsmothers:
We've used the [_get_clones](https://github.com/pytorch/torchtune/blob/28c9ec20a9c8a27687302268672cfcd43d263262/torchtune/modules/transformer.py#L307) utility a lot for cleaner instantiation of our `TransformerDecoder` from a single layer. However, we shouldn't use it for cases that require random initialization and don't subsequently override their params with a state dict load (i.e. what we do for LoRA when we're not resuming from an intermediate checkpoint). The following script demonstrates why: the initialized values for cloned modules will be the same across layers, so if we use `_get_clones` (and don't subsequently load in a weight to override the init values), all our layers have identical values.

```
import torch
from torchtune.modules.transformer import _get_clones
from torchtune.modules.peft import LoRALinear

def main():
    loras_loop = [None] * 4
    for i in range(4):
        loras_loop[i] = LoRALinear(in_dim=16, out_dim=8, rank=4, alpha=1.0)

    loras_cloned = _get_clones(
        LoRALinear(in_dim=16, out_dim=8, rank=4, alpha=1.0), 4
    )

    loop_max_diff = torch.max(torch.abs(loras_loop[0].lora_a.weight - loras_loop[3].lora_a.weight))
    cloned_max_diff = torch.max(torch.abs(loras_cloned[0].lora_a.weight - loras_cloned[3].lora_a.weight))

    print(f"Max diff in for loop: {loop_max_diff}")
    print(f"Max diff with _get_clones: {cloned_max_diff}")

if __name__ == "__main__":
    main()


...

Max diff in for loop: 0.4612632691860199
Max diff with _get_clones: 0.0
```

RdoubleA:
@ebsmothers what would be the recommended alternative here? Could we just call `reset_parameters` or similar on all the layers in `_get_clones`?

Ankur-singh:
Dumb question: why can't we use `for-loop`? Does it have to do something with performance?

RdoubleA:
The layer is already instantiated outside of `TransformerDecoder`/`VisionTransformer`, or wherever `get_clones` is used. So we can't use a for loop to instantiate multiple layers, unless we change the `__init__` signatures to accept a `ModuleList` instead.

ebsmothers:
[...]

PR #2282, opened 2025-01-19
Update model builders

Ankur-singh:
#### Context
- [ ] add a new feature
- [ ] fix a bug
- [X] other (Refactor)

Please link to any issues this PR addresses #2270 

#### Changelog
* Updated all `component_builder` to pass `nn.ModuleList` to `TransformerDecoder` instead of (`layer` + `num_layers`).
* Updated test for `T5` model

#### Test plan

#### UX


Ankur-singh:
@ebsmothers for now, I have only updated the `component_builders` for all the models. In my local testing, removing this `else` block https://github.com/pytorch/torchtune/blob/779569e49c6a7c9988d0109841180b6e69ea2027/torchtune/modules/transformer.py#L375

passes all the tests. If this PR looks good, I will:
1. Update `VisionTransformer` class and component builders for vision models
2. Delete the `_get_clones` function

torchtune/models/gemma2/_component_builders.py
```diff
@@ -115,8 +114,7 @@ def gemma2(
     """
     rope = RotaryPositionalEmbeddings(dim=head_dim, max_seq_len=max_seq_len, base=rope_base)
     
-    layers = torch.nn.ModuleList()
-    
+    layers = nn.ModuleList()
```
RdoubleA, line 117: super nit: let's keep `torch.nn.ModuleList()` as is, there was some big PR a long time ago to keep torch.nn

torchtune/models/gemma/_component_builders.py
```diff
@@ -76,33 +76,37 @@ def gemma(
...
+        self_att = MultiHeadAttention(
+            embed_dim=embed_dim,
+            num_heads=num_heads,
+            num_kv_heads=num_kv_heads,
+            head_dim=head_dim,
+            q_proj=nn.Linear(embed_dim, num_heads * head_dim, bias=False),
+            k_proj=nn.Linear(embed_dim, num_kv_heads * head_dim, bias=False),
+            v_proj=nn.Linear(embed_dim, num_kv_heads * head_dim, bias=False),
+            output_proj=nn.Linear(num_heads * head_dim, embed_dim, bias=False),
+            pos_embeddings=rope,
```
RdoubleA, line 91: rope doesn't have any trainable params, but I'm worried that assigning every MHA to the same rope module could lead to silent bugs. before `_get_clones` would make a copy of each layers rope, so let's retain that behavior

torchtune/models/llama2/_component_builders.py
```diff
@@ -212,45 +215,48 @@ def lora_llama2(
...
         intermediate_dim if intermediate_dim else scale_hidden_dim_for_mlp(embed_dim)
     )
-    if apply_lora_to_mlp:
-        mlp = lora_llama2_mlp(
-            dim=embed_dim,
-            hidden_dim=hidden_dim,
+
+    layers = nn.ModuleList()
+    for _ in range(num_layers):
+        self_attn = lora_llama2_self_attention(
```
RdoubleA, line 224: Here, you are creating a separate rope for every MHA. So we should make sure all the other updates also do the same (referring to my other comment)

RdoubleA:
Just one comment about make sure RoPE is instantiated correctly, but the rest looks good

torchtune/models/gemma/_component_builders.py
ebsmothers: I previously had a similar concern but I think it should be OK. This was already done by @felipemello1 in #1554 for Llama 3.1 and we stress tested it there.

torchtune/models/gemma2/_component_builders.py
ebsmothers: Yeah looks like we aren't even using `nn.ModuleList` vs `torch.nn.ModuleList` consistently today. If you want to standardize on `torch.nn.ModuleList` that's great but not a blocker for this PR

tests/torchtune/models/t5/test_t5_encoder.py
```diff
@@ -48,27 +48,28 @@ def inputs(self):
 
     def test_forward(self, model, inputs):
         actual = model(inputs)
+        print(actual)
```
ebsmothers, line 51: remove

tests/torchtune/models/t5/test_t5_encoder.py
```diff
@@ -48,27 +48,28 @@ def inputs(self):
...
                 [
-                    [0.3670, 0.2938],
-                    [0.3692, 0.2921],
-                    [0.3611, 0.2984],
-                    [0.4207, 0.2437],
-                    [0.3447, 0.3106],
-                    [0.3383, 0.3150],
-                    [0.3727, 0.2892],
-                    [0.3996, 0.2653],
+                    [0.4958, 0.4845],
```
ebsmothers, line 55: I'm a bit surprised that the expected values need to change here. I would have thought that uniform initialization iterating over `model.parameters()` shouldn't be affected by whether we use `_get_clones` or not

torchtune/models/llama2/_component_builders.py
ebsmothers: Yeah I am now realizing that Felipe's changes only propagated to some, but not all, of our builders. E.g. the [Llama 3.1 self-attention builder](https://github.com/pytorch/torchtune/blob/779569e49c6a7c9988d0109841180b6e69ea2027/torchtune/models/llama3_1/_component_builders.py#L277) takes RoPE embeddings as an argument, allowing us to build RoPE a single time across all layers. However, the corresponding components for [Llama2](https://github.com/pytorch/torchtune/blob/779569e49c6a7c9988d0109841180b6e69ea2027/torchtune/models/llama2/_component_builders.py#L412) and [Llama3](https://github.com/pytorch/torchtune/blob/779569e49c6a7c9988d0109841180b6e69ea2027/torchtune/models/llama3/_component_builders.py#L413-L415) do not do this -- they instead construct RoPE for every single layer. 

My advice is to not worry too much about whether we use a single global RoPE vs one RoPE per layer in this particular PR. We should standardize this but I don't want to introduce scope creep on this PR. Let's file an issue and do it separately, for now you can try to stick with whatever the existing APIs support.

ebsmothers (commented):
Thanks for making these changes! I left a few comments (mainly in response to some of @RdoubleA's points about RoPE -- TLDR is don't worry too much for this PR provided everything works). Otherwise no major concerns though

tests/torchtune/models/t5/test_t5_encoder.py
Ankur-singh: After making all the changes, I tried `pytest tests` with:
- with `_get_clones`: success 
- Without `_get_clones`: failed

Hence, I thought the difference is because of `_get_clones`. Ran the following script to test the hypothesis:

```python
import torch
from torchtune.modules.transformer import _get_clones
from torchtune.modules.peft import LoRALinear

def main():
    loras_loop = [None] * 4
    for i in range(4):
        loras_loop[i] = LoRALinear(in_dim=16, out_dim=16, rank=4, alpha=1.0)

    loras_cloned = _get_clones(
        LoRALinear(in_dim=16, out_dim=16, rank=4, alpha=1.0), 4
    )

    loop_max_diff = torch.max(torch.abs(loras_loop[0].lora_a.weight - loras_loop[3].lora_a.weight))
    cloned_max_diff = torch.max(torch.abs(loras_cloned[0].lora_a.weight - loras_cloned[3].lora_a.weight))

    print(f"Max diff between layers using for-loop: {loop_max_diff}")
    print(f"Max diff between layers using _get_clones: {cloned_max_diff}")

    input = torch.randn(1, 16)
    
    output1 = input.clone()
    for layer in loras_loop:
        output1 = layer(output1)

    output2 = input.clone()
    for layer in loras_cloned:
        output2 = layer(output2)

    cloned_max_diff = torch.max(torch.abs(output1 - output2))
    print(f"Max diff between outputs from two approach: {cloned_max_diff}")


if __name__ == "__main__":
    main()

# ----
# Max diff between layers using for-loop: 0.4660979211330414
# Max diff between layers using _get_clones: 0.0
# Max diff between outputs from two approach: 0.3515825569629669
```

torchtune/models/t5/_encoder.py
```diff
@@ -37,17 +36,18 @@ def __init__(
...
         rel_pos_num_buckets: int,
         rel_pos_max_dist: int,
         max_seq_len: int,
     ):
         super().__init__()
         self.token_embedding = token_embedding
-        self.layers = nn.ModuleList([copy.deepcopy(layer) for i in range(num_layers)])
+        self.layers = (
+            layers if isinstance(layers, nn.ModuleList) else nn.ModuleList(layers)
+        )
```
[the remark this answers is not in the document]
Ankur-singh: Thanks for pointing it out. I will re-introduce support for `num_layers` and add `deprecated` decorator with message "`num_layers` argument will be deprecated in upcoming release."

torchtune/models/llama2/_component_builders.py
Ankur-singh: Thanks @RdoubleA for bringing it to my attention. I totally missed it, TBH its quite subtle. 

@ebsmothers I'll go ahead a create an issue. We can discuss the best way forward before implementing.

torchtune/models/t5/_encoder.py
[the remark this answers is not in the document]
Ankur-singh: @ebsmothers Do you want me to add `deprecated` decorator to `TransformerDecoder` class as well?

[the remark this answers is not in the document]
ebsmothers: @Ankur-singh actually I wouldn’t worry about adding the decorator to the class. I think we need a separate utility to log only when the to-be-deprecated argument is passed. Otherwise everyone will see the warning about deprecation of num_layers even if they aren’t using it

tests/torchtune/models/t5/test_t5_encoder.py
ebsmothers: Yeah I suspect this is due to something with the random seed. Let me take a closer look to confirm but otherwise I think updating the expected values is fine

ebsmothers:
Hi @Ankur-singh thanks for your patience, I wanted to sanity check the failing test to make sure it's expected. In our `generate_v2` recipe we set the seed during init [here](https://github.com/pytorch/torchtune/blob/d7afc40c4098335340f6ca3683123117b346fb90/recipes/dev/generate_v2.py#L81). This means we construct the model after setting seed, and so when you change how we initialize the model (i.e. we now do some random initialization for every single layer, instead of just one layer that we then copy), the RNG state will be in a different place by the time we get to our call to `generate`, hence our sampling for generation will yield different results. You can confirm this by moving the call to `training.set_seed` e.g. [here](https://github.com/pytorch/torchtune/blob/d7afc40c4098335340f6ca3683123117b346fb90/recipes/dev/generate_v2.py#L126) to force model initialization to happen first -- you will see that whether you run on main or on this PR you will get the same result.

Anyways TLDR for this PR is that you are not breaking anything, you can safely just update the expected value for this test. Separately we can think about whether it's clearer to call `set_seed` at the beginning of `generate` to ensure that we get deterministic behavior irrespective of model initialization. cc @joecummings in case he has any thoughts on this.

tests/torchtune/models/t5/test_t5_encoder.py
ebsmothers: Also dug into this one a bit more: the changes to T5 change the order in which modules are registered, so when we do

```
for param in model.parameters():
	param.data.uniform_(0, 1)
```

we again wind up with different state for the rng. So again it's fine to change the expected values here

Ankur-singh:
Hi @ebsmothers thanks for clarifying. I'm a bit confused, after initializing the model, we load the `state_dict` from checkpoint. This should basically overwrite any previous initialization, right? Furthermore, during testing, we are loading some dummy model weights. This should make the model weights deterministic, irrespective of the initialization method and `random_seed`. Is my understanding correct so far?

`random_seed` only comes to picture in `generate` method when sampling the next token. So at least during testing, as long as we have the same `random_seed` and `set_seed` is called before the generation starts, we should be getting the same output. Only thing that could lead to different generation would be change in checkpoint weights.

I also tried out this small script to see if calling a random operation before setting the seed affects the outcome (looks like it does not):
```python
 import torch

# Case 1: No random operations before setting seed
torch.manual_seed(42)
x1 = torch.randn(3)
print("Case 1:", x1)

_ = torch.randn(5)  # Consumes RNG state
_ = torch.randn(5) 

# Case 2: Random operation before setting seed
torch.manual_seed(42)
x2 = torch.randn(3)
print("Case 2:", x2)
print(f"Case 1 and Case 2 are equal: {torch.allclose(x1, x2)}")

-----
# Output:
# Case 1: tensor([0.3367, 0.1288, 0.2345])
# Case 2: tensor([0.3367, 0.1288, 0.2345])
# Case 1 and Case 2 are equal: True
```

TLDR, as long as we are loading the model weight from same checkpoint and calling `set_seed` before start of generation (either in `__init__` or `generate` method) we should be getting the same output. Am I missing something here?

ebsmothers:
@Ankur-singh yeah it's a bit tricky, let me try to address a couple of these points.

> I'm a bit confused, after initializing the model, we load the state_dict from checkpoint. This should basically overwrite any previous initialization, right? Furthermore, during testing, we are loading some dummy model weights. This should make the model weights deterministic, irrespective of the initialization method and random_seed. Is my understanding correct so far?

This is the correct understanding. The model weights will wind up being the same regardless of the random seed because the state dict weights override any random initialization. The one caveat is that if the initialization differs, the random state will not be the same. This is why the generate_v2 test results change. We currently do things in the following order: (1) set the seed, (2) initialize the model, (3) perform some other operation requiring randomness (in this case sampling from logits). 

Looking at your code snippet, I think things are not quite in the right order -- hence why you're getting the same results both times. Instead it should look like this:

```
 import torch

# Case 1: No random operations *AFTER* setting seed
torch.manual_seed(42)
x1 = torch.randn(3)
print("Case 1:", x1)


# Case 2: Random operation *AFTER* setting seed
torch.manual_seed(42)
_ = torch.randn(5)  # Consumes RNG state
_ = torch.randn(5) # *THESE NOW OCCUR AFTER TORCH.MANUAL_SEED*
x2 = torch.randn(3)
print("Case 2:", x2)
print(f"Case 1 and Case 2 are equal: {torch.allclose(x1, x2)}")

-----
# Output:
# Case 1: tensor([0.3367, 0.1288, 0.2345])
# Case 2: tensor([0.5349, 0.8094, 1.1103])
# Case 1 and Case 2 are equal: False
```

Anytime you call `torch.manual_seed` you are basically resetting the RNG state. So if you generate random numbers, then call `torch.manual_seed`, the RNG state will then be as though you had never generated said numbers. That's why your snippet shows the same results, whereas if I call `randn` after `manual_seed`, the results differ.

Going back to the order of (1), (2), (3) for `generate_v2.py` I described above: the implications of this are that any random operation after calling `set_seed` (even if its results are later overwritten) will impact the RNG state, and without a subsequent call to `set_seed` to reset things we will get different results for any subsequent random operations. This is why I mentioned moving the `set_seed` call [here](https://github.com/pytorch/torchtune/blob/d7afc40c4098335340f6ca3683123117b346fb90/recipes/dev/generate_v2.py#L126) -- this basically swaps (1) and (2), so that there is no longer any model initialization (i.e. there are no longer any random operations) after the call to `set_seed`. By doing this, you will see that the generation results look the same regardless of how the model is initialized.

Hope this helps clarify things a bit! Please let me know if anything remains unclear.

Ankur-singh:
@ebsmothers thank you so much. My mental model for `random_state` is completely shattered. I can see how this seemingly minor change (swapping 1 & 2) can leading to different results. I really appreciate you taking the time and sharing these insightful nuggets. 

Regarding this PR at hand, do you want me to change the `generatev2` output in the test file or do you want me wait until we finalize the approach?

ebsmothers:
@Ankur-singh you can go ahead and change the output, we can decide whether it’s worth moving our call to set_seed at a later date (since it’s not a bug, more of a quality-of-life thing)

Ankur-singh:
@ebsmothers I have updated the value of [`expected_output`](https://github.com/pytorch/torchtune/blob/d7afc40c4098335340f6ca3683123117b346fb90/tests/recipes/dev/test_generate_v2.py#L59) with `"Pietroместkap щotimes rivers cache НиtringindexPathNAME"`. Tested it by running `pytest tests/recipes/dev/test_generate_v2.py --ignore tests/torchtune/modules/_export --with-integration --durations=20 -vv`

```bash
(tune) ankur@nuc:~/github/torchtune$ pytest tests/recipes/dev/test_generate_v2.py --ignore tests/torchtune/modules/_export --with-integration --durations=20 -vv
small-ckpt-tune-03082024.pt
small-ckpt-meta-03082024.pt
small-ckpt-hf-03082024.pt
small-ckpt-tune-llama3-05052024.pt
small-ckpt-hf-reward-07122024.pt
small-ckpt-meta-vision-10172024.pt
small-ckpt-hf-vision-10172024.pt
tokenizer.model
tokenizer_llama3.model
File already exists locally: /tmp/test-artifacts/small-ckpt-tune-03082024.pt
File already exists locally: /tmp/test-artifacts/small-ckpt-meta-03082024.pt
File already exists locally: /tmp/test-artifacts/small-ckpt-hf-03082024.pt
File already exists locally: /tmp/test-artifacts/small-ckpt-tune-llama3-05052024.pt
File already exists locally: /tmp/test-artifacts/small-ckpt-hf-reward-07122024.pt
File already exists locally: /tmp/test-artifacts/small-ckpt-meta-vision-10172024.pt
File already exists locally: /tmp/test-artifacts/small-ckpt-hf-vision-10172024.pt
File already exists locally: /tmp/test-artifacts/tokenizer.model
File already exists locally: /tmp/test-artifacts/tokenizer_llama3.model
================================================================= test session starts =================================================================
platform linux -- Python 3.10.16, pytest-7.4.0, pluggy-1.5.0 -- /home/ankur/miniforge3/envs/tune/bin/python3.10
cachedir: .pytest_cache
rootdir: /home/ankur/github/torchtune
configfile: pyproject.toml
plugins: cov-6.0.0, mock-3.14.0, integration-0.2.3
collected 2 items                                                                                                                                     

tests/recipes/dev/test_generate_v2.py::TestGenerateV2::test_llama2_generate_results PASSED                                                      [ 50%]
tests/recipes/dev/test_generate_v2.py::TestGenerateV2::test_llama2_fail_on_bad_input PASSED                                                     [100%]

================================================================ slowest 20 durations =================================================================
0.55s call     tests/recipes/dev/test_generate_v2.py::TestGenerateV2::test_llama2_generate_results
0.00s setup    tests/recipes/dev/test_generate_v2.py::TestGenerateV2::test_llama2_generate_results
0.00s setup    tests/recipes/dev/test_generate_v2.py::TestGenerateV2::test_llama2_fail_on_bad_input
0.00s teardown tests/recipes/dev/test_generate_v2.py::TestGenerateV2::test_llama2_fail_on_bad_input
0.00s teardown tests/recipes/dev/test_generate_v2.py::TestGenerateV2::test_llama2_generate_results
0.00s call     tests/recipes/dev/test_generate_v2.py::TestGenerateV2::test_llama2_fail_on_bad_input
================================================================== 2 passed in 0.58s ==================================================================
```

ebsmothers (approved):
Thanks @Ankur-singh! This one looks good as well

Issue #2261, opened 2025-01-13
[RFC] Additional chat loss masking strategies


RdoubleA:
## Context
Continuing the discussion from #2214.

When LLMs are trained on chat data, you typically need control over which messages the model is actually trained on and which ones are masked from the loss, and thereby do not affect loss calculation. The most common setting is to only train on the model responses in the dataset:
```
messages = [
	Message(role="user", content="What is the greatest TV show of all time?", masked=True),
	Message(role="assistant", content="The Sopranos, c'mon Tony", masked=False),
]
```
In this example, we don't need to train to model to predict the next tokens in the user prompt, only the answer to the actual prompt. Thus, the user message is masked. Another alternative is to train the model on all messages including the user prompt. This is the approach taken by Stanford Alpaca.

In torchtune, we provide this control to the user via `train_on_input`, which is supported by all our [built-in message transforms](https://pytorch.org/torchtune/main/basics/message_transforms.html#example-message-transforms). This lets you easily configure the masking strategies from the config or command line: 
```
dataset:
  _component_: torchtune.datasets.alpaca_dataset
  train_on_input: True
```
or add `dataset.train_on_input=True` to your launch command: `tune run full_finetune_single_device --config my_config.yaml dataset.train_on_input=True`

However, as brought up in #2214, there may be more common masking strategies that we should provide strong support for. The most requested one is to mask out all messages except the last assistant message in a multi-turn conversation. Currently, a [custom message transform](https://github.com/pytorch/torchtune/issues/2214#issuecomment-2583743052) needs to be created to support this, which is not intuitive to beginner users.

## Proposal
Instead of the binary `train_on_input` parameter, we can replace this with a string parameter that specifies the masking strategy. We can add mask all but last since this was requested, and any other masking strategies in the future if the community requires it.
```
masking_strategy (str): string parameter indicating the masking strategy to use. Must be one of ``train_on_all``, ``train_on_assistant``, ``train_on_last``
	- ``train_on_all``: both user and assistant messages are unmasked and included in loss calculation
	- ``train_on_assistant``: user messages are masked, only assistant messages are unmasked and included in loss calculation
	- ``train_on_last``: only the last assistant message is included in loss calculation
```
The names of the strategies can be different if folks feel strongly (like none, user, except_last if that's more clear to what is being masked, but I do like being explicit about what is being trained).

We will need to do the following:
[...]

PR #2284, opened 2025-01-20
Add masking strategies to message transforms

supreethmanyam:
#### Context
- [x] add a new feature
- [x] fix a bug
- [ ] other (please add here)

https://github.com/pytorch/torchtune/issues/2261

#### Changelog
* Added `mask_messages` utility in `torchtune/data/_messages.py` to mask each message given a list of messages based on user provided masking_strategy
* Fixes in-built message transforms to always mask system message
* Added tests for calling message transforms with masking strategies

#### Test plan

#### UX


supreethmanyam:
@RdoubleA Added `mask_messages` utility in `torchtune/data/_messages.py` along with relevant unit tests.
I will update the documentation once the changes look good.

torchtune/data/_messages.py
```diff
@@ -176,12 +185,24 @@ class InputOutputToMessages(Transform):
 
     def __init__(
         self,
-        train_on_input: bool = False,
+        masking_strategy: Optional[str] = "train_on_all",
```
RdoubleA, line 188: if the default value for `train_on_input` was False, this would correspond to `train_on_assistant`. Let's keep that same default behavior.

torchtune/data/_messages.py
```diff
@@ -305,11 +332,23 @@ class ChosenRejectedToMessages(Transform):
...
```
RdoubleA, line 335: same comment here

torchtune/data/_messages.py
```diff
@@ -406,20 +445,35 @@ class ShareGPTToMessages(Transform):
...
+            Default is None.
 
     Raises:
         ValueError: If ``column_map`` is provided and ``conversations`` not in ``column_map``.
     """
 
     def __init__(
         self,
-        train_on_input: bool = False,
+        masking_strategy: Optional[str] = "train_on_all",
```
RdoubleA, line 458: same comment here

torchtune/data/_messages.py
```diff
@@ -623,55 +691,12 @@ def __call__(self, sample: Mapping[str, Any]) -> Mapping[str, Any]:
...
                     content=content,
-                    masked=masked,
                 ),
             )
-
+        mask_messages(updated_messages, self.masking_strategy)
         return {"messages": updated_messages}
 
 
-def validate_messages(
```
RdoubleA, line 633: not sure if this is still used but let's keep this here for now

torchtune/data/_messages.py
```diff
@@ -682,17 +707,38 @@ class AlpacaToMessages(Transform):
...
             keeping the default column names.
+        train_on_input (Optional[bool]): whether the model is trained on the user prompt or not.
+            Deprecated parameter and will be removed in a future release.
+            Default is None.
     """
 
     def __init__(
-        self, train_on_input: bool = True, column_map: Optional[Dict[str, str]] = None
+        self,
+        masking_strategy: Optional[str] = "train_on_all",
```
RdoubleA, line 726: contrary to the other comment, this default value should remain `train_on_all` since `train_on_input` was defaulted to True. so let's keep this

torchtune/data/_messages.py
```diff
@@ -726,14 +772,94 @@ def __call__(self, sample: Mapping[str, Any]) -> Mapping[str, Any]:
...
+            Must be one of: `train_on_all`, `train_on_assistant`, `train_on_last`.
+            - ``train_on_all``: both user and assistant messages are unmasked
+            - ``train_on_assistant``: user messages are masked, only assistant messages are unmasked
+            - ``train_on_last``: only the last assistant message is unmasked
+
+    Raises:
+        ValueError: If the masking strategy is not one of the supported strategies:
+                    `train_on_all`, `train_on_assistant`, `train_on_last`.
+    """
+    if masking_strategy not in ["train_on_all", "train_on_assistant", "train_on_last"]:
```
RdoubleA, line 845: One suggestion that I didn't mention in the issue is that we could make `masking_strategy` an Enum so that it can be type checked more easily. users can still specify as a string in the config and in the message transforms above, but here in `mask_messages` you could convert to the Enum, and the ValueError will automatically be captured.

RdoubleA (commented):
This looks great, really like the tests 👌 left a couple of comments

torchtune/data/_messages.py
supreethmanyam: Updated default strategy

torchtune/data/_messages.py
supreethmanyam: Updated default strategy

torchtune/data/_messages.py
supreethmanyam: Updated default strategy

torchtune/data/_messages.py
supreethmanyam: Converted masking_strategy to Enum and raises ValueErroe automatically if doesn't match with mapping

supreethmanyam:
@RdoubleA 

In addition to converting masking_strategy to Enum and fixing default strategy in built-in transforms, I made following small changes:
1. I modified structure of `column_map` usage in `InputOutputToMessages` and `AlpacaToMessages` to make it consistent across transforms
2. I realized there were no unit tests for `AlpacaToMessages`, I added a few tests.

torchtune/data/_messages.py
```diff
@@ -177,31 +193,42 @@ class InputOutputToMessages(Transform):
 
     def __init__(
         self,
-        train_on_input: bool = False,
+        masking_strategy: Optional[str] = "train_on_assistant",
```
ebsmothers, line 196: Why do we switch the order of these arguments? If someone is passing `InputOutputToMessages(False)` this will break them, right? (Since we unfortunately don't enforce keyword-only args on these APIs). Personally I would leave train_on_input in the first position and put masking_strategy last

tests/torchtune/data/test_messages.py
```diff
@@ -267,6 +296,55 @@ def test_call_train_on_input(self, sample):
         ]
         assert_dialogue_equal(actual["rejected"], expected_rejected)
 
+    @pytest.mark.parametrize(
+        "masking_strategy, expected_masks",
+        [
+            ("train_on_all", [[False, False], [False, False]]),
+            ("train_on_assistant", [[True, False], [True, False]]),
+            ("train_on_last", [[True, False], [True, False]]),
```
ebsmothers, line 304: Ideally we should explicitly test a case where the results of `train_on_assistant` and `train_on_last` differ

tests/torchtune/data/test_messages.py
```diff
@@ -479,6 +609,200 @@ def test_call_image_messages(self, mock_load_image):
         mock_load_image.assert_called_once_with("https://example.com")
 
 
+class TestAlpacaToMessages:
```
ebsmothers, line 612: What's the difference between this and the existing `TestAlpacaToMessages` [here](https://github.com/pytorch/torchtune/blob/67a8706abd993d4b03c70506075a2a9804919148/tests/torchtune/datasets/test_alpaca_dataset.py#L93)? They look fairly similar

torchtune/data/_messages.py
```diff
@@ -709,32 +782,102 @@ def __init__(
...
-                masked=False,
+                content=sample[self._column_map["output"]],
                 eot=True,
             ),
         ]
+        mask_messages(messages, self.masking_strategy)
         return {"messages": messages}
+
+
+def validate_messages(
```
ebsmothers, line 812: Similar q here: don't we have an existing `validate_messages` function?

ebsmothers (commented):
Sorry @supreethmanyam, I think this PR fell through the cracks a bit. I have a couple comments but after that I think it looks good. cc @RdoubleA to confirm

torchtune/data/_messages.py
RdoubleA: agreed, let's keep the positions of the arguments the same until we deprecate

torchtune/data/_messages.py
RdoubleA: it was just moved down here I believe

torchtune/data/_messages.py
```diff
@@ -683,18 +714,60 @@ class AlpacaToMessages(Transform):
...
     def __init__(
-        self, train_on_input: bool = True, column_map: Optional[Dict[str, str]] = None
+        self,
+        masking_strategy: Optional[str] = "train_on_all",
+        column_map: Optional[Dict[str, str]] = None,
+        train_on_input: Optional[bool] = None,
     ):
-        self.train_on_input = train_on_input
-        self.column_map = column_map
+        if train_on_input is not None:
```
RdoubleA, line 742: I thought we had a deprecated param utility, but it looks like it never got reviewed: https://github.com/pytorch/torchtune/pull/2321. should make a note there to also replace this code if this PR merges first @ebsmothers

tests/torchtune/data/test_messages.py
supreethmanyam: I have added test cases using multi turn samples where results `train_on_assistant` and `train_on_last` differ where applicable (ChosenRejectedToMessages, ShareGPTToMessages, OpenAIToMessages)

tests/torchtune/data/test_messages.py
supreethmanyam: I did not know we had `TestAlpacaToMessages` in test_alpaca_dataset.py and added one in test_messages.py. 

Both are similar and the one in test_messages.py runs more tests. Please let me know if I can remove TestAlpacaToMessages from [here](https://github.com/pytorch/torchtune/blob/67a8706abd993d4b03c70506075a2a9804919148/tests/torchtune/datasets/test_alpaca_dataset.py#L93)

torchtune/data/_messages.py
supreethmanyam: I moved `validate_messages` down to keep all the message transforms at one place.

supreethmanyam:
@RdoubleA @ebsmothers 
I have made all the changes and updated the documentation. Requesting another review.

Current references to `train_on_input` in the docs comes from dataset APIs, hence I haven't made any changes yet. Please let me know if I should create another issue for dataset API changes. I will be happy to make changes to dataset APIs to reflect using `masking_strategy` instead of `train_on_input` and update docs accordingly.

ebsmothers:
Thanks @supreethmanyam for your patience on this one. Since we have not been great about timely reviews here, I took the liberty of pushing some changes directly to the PR (hope that's alright). Mainly I changed the deprecation logic slightly: when `train_on_input` and `masking_strategy` are both passed, we should still use `train_on_input` but warn the user (otherwise it's possible for silent behavior changes without the user actually migrating). While doing this, I noticed that our SlimOrca test case seems incorrect -- it appears as though we were never masking the system message (see [here](https://github.com/pytorch/torchtune/blob/9b05ac46ded4424bcecf3984ace42704af6a7422/torchtune/data/_messages.py#L490-L492)). cc @RdoubleA to sanity check me here. After this (and green CI) I think this should be good to land. Thanks again for your patience!

RdoubleA:
That's a tough one... it does look like system is being masked, but if train_on_input is true it will not be masked, same as the user message. So yeah, big oof

ebsmothers (approved):
Really appreciate you adding this @supreethmanyam!

PR #2291, opened 2025-01-22
Proper prefix handling in EarlyFusion sd hooks

ebsmothers:
Thanks to @mkrainin for pointing this out and making the initial fix.

Our `EarlyFusion` state dict hooks don't work when an `EarlyFusion` module is wrapped as part of a larger model (as is the case with e.g. DDP). This is because state dict keys in the hooks have global key names, so we can't just use e.g. `del state_dict["decoder.tok_embeddings.weight"]` since `decoder` may not be a top-level module.

The fix is to use the `prefix` arg to state dict hooks, which gets recursively prepended while traversing the module tree ([ref](https://github.com/pytorch/pytorch/blob/a56af029133ba0e87ed4c099e7e4efc40a24e20c/torch/nn/modules/module.py#L1892-L1894)).

## Test plan

Added a unit test wrapping `EarlyFusion` into an `nn.Sequential` so that the state dict keys have integer prefixes. This plus the existing test of state dict hooks pass after these changes:

```
pytest tests/torchtune/modules/model_fusion/test_early_fusion.py
...
================== 12 passed in 0.18s =============
```

### DIFF
--- a/torchtune/modules/model_fusion/_early_fusion.py
+++ b/torchtune/modules/model_fusion/_early_fusion.py
@@ -130,23 +130,25 @@ class EarlyFusionModel(nn.Module):
         set_trainable_params(self, trainable_params)
 
     @staticmethod
-    def _state_dict_hook(module, state_dict, *args, **kwargs):
+    def _state_dict_hook(module, state_dict, prefix, *args, **kwargs):
         """
         Keep tok_embeddings inside of decoder state_dict
 
         [!Note] This update changes the order of the OrderedDict
         """
         for n, p in module.tok_embeddings.named_parameters():
-            state_dict[f"decoder.tok_embeddings.{n}"] = p
-            del state_dict[f"tok_embeddings.{n}"]
+            state_dict[f"{prefix}decoder.tok_embeddings.{n}"] = p
+            del state_dict[f"{prefix}tok_embeddings.{n}"]
 
     @staticmethod
-    def _load_state_dict_hook(module, state_dict, *args, **kwargs):
+    def _load_state_dict_hook(module, state_dict, prefix, *args, **kwargs):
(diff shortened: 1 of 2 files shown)


torchtune/modules/model_fusion/_early_fusion.py
```diff
@@ -130,23 +130,25 @@ def __init__(
...
+    def _state_dict_hook(module, state_dict, prefix, *args, **kwargs):
         """
         Keep tok_embeddings inside of decoder state_dict
 
         [!Note] This update changes the order of the OrderedDict
         """
         for n, p in module.tok_embeddings.named_parameters():
-            state_dict[f"decoder.tok_embeddings.{n}"] = p
-            del state_dict[f"tok_embeddings.{n}"]
+            state_dict[f"{prefix}decoder.tok_embeddings.{n}"] = p
```
RdoubleA, line 140: prefix will include the period?

RdoubleA (approved):
Good find with the prefix argument, do we need to update DeepFusion as well?

torchtune/modules/model_fusion/_early_fusion.py
ebsmothers: Yep, see [here](https://github.com/pytorch/pytorch/blob/a56af029133ba0e87ed4c099e7e4efc40a24e20c/torch/nn/modules/module.py#L1892-L1894)

ebsmothers:
@RdoubleA DeepFusion should be good. State dict hooks are not defined at the model level there, but at least individual layers seem to already be using the prefix argument in a similar fashion. [E.g.](https://github.com/pytorch/torchtune/blob/9b5062889b9205ab70c336d1c453ea98074b8edd/torchtune/modules/model_fusion/_fusion_layers.py#L216-L226)

PR #2063, opened 2024-11-25
Deprecate `SimpoLoss`

SalmanMohammadi:
#### Context
What is the purpose of this PR? Is it to
- [ ] add a new feature
- [ ] fix a bug
- [ ] update tests and/or documentation
- [x] other (deprecation)

Closes #2062. We could link to an example repo in a follow up.

#### Changelog

Added a deprecation message. To be removed in 0.6.

PR #2292, opened 2025-01-22
Custom DPO losses support

krammnic:
#### Context
- [X] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

#### Changelog
* #2035 

#### Test plan

#### UX

Was able to forward with external loss and concatenated_forward.

### DIFF
--- a/recipes/lora_dpo_distributed.py
+++ b/recipes/lora_dpo_distributed.py
@@ -5,6 +5,7 @@
 # LICENSE file in the root directory of this source tree.
 
 import sys
+import omegaconf
 import time
 
 from functools import partial
@@ -289,6 +290,16 @@ class LoRADPORecipeDistributed(FTRecipeInterface):
         self._loss_fn = config.instantiate(cfg.loss)
 
         utils.log_rank_zero(log, "Loss is initialized.")
+
+        # reference: true is not exposed in the configs.
+        try:
+            self._reference = config.instantiate(cfg.reference)
+            log.info("Concatenated forward is initialized.")
+        except omegaconf.errors.ConfigAttributeError:
+            log.info(
+                "reference: false was not selected, going with true."
+            )
+            self._reference = True
 
         # sampler and dataloader depend on the tokenizer and loss_fn and should be
         # setup after all of these are setup
@@ -611,39 +622,6 @@ class LoRADPORecipeDistributed(FTRecipeInterface):
                 adapter_only=self._save_adapter_weights_only,
             )
 
-    def concatenated_forward(
-        self, model: nn.Module, batch: Tuple[torch.Tensor, torch.Tensor]
-    ) -> Tuple[torch.Tensor, torch.Tensor, torch.Tensor, torch.Tensor]:
-        """
-        Run forward pass of the model with chosen and rejected samples concatenated.
-
-        Args:
-            model (nn.Module): The model to be used for the forward pass.
-            batch (Tuple[torch.Tensor, torch.Tensor]): Tuple of input_ids and labels.
-
-        Returns:
-            Tuple of chosen log probs, rejected log probs, chosen logits, rejected logits.
-        """
-        concatenated_input_ids, concatenated_labels = batch
-        concatenated_input_ids = concatenated_input_ids.to(self._device)
-        concatenated_labels = concatenated_labels.to(self._device)
-
-        # formed by concatenating an equal number of "chosen" and "rejected".
-        len_chosen = concatenated_input_ids.shape[0] // 2
-
-        with self.activations_handling_ctx:
-            all_logits = model(concatenated_input_ids)
-
-        all_log_probs = rlhf.get_batch_log_probs(all_logits, concatenated_labels)
-
-        chosen_log_probs = all_log_probs[:len_chosen]
-        rejected_log_probs = all_log_probs[len_chosen:]
-
-        chosen_logits = all_logits[:len_chosen]
-        rejected_logits = all_logits[len_chosen:]
-
-        return (chosen_log_probs, rejected_log_probs, chosen_logits, rejected_logits)
-
     def train(self) -> None:
         """
         The core training loop.
@@ -696,7 +674,10 @@ class LoRADPORecipeDistributed(FTRecipeInterface):
                     policy_rejected_log_probs,
                     policy_chosen_logits,
                     policy_rejected_logits,
-                ) = self.concatenated_forward(self._model, batch)
+                    *args,
+                ) = self._loss_fn.concatenated_forward(
+                    self._model, batch, self._device, self.activations_handling_ctx
+                )
 
                 policy_chosen_logits_mean = policy_chosen_logits.detach().mean()
                 policy_rejected_logits_mean = policy_rejected_logits.detach().mean()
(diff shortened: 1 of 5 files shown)


krammnic:
@SalmanMohammadi Hey! Can you take a look please?

recipes/lora_dpo_distributed.py
```diff
@@ -290,6 +292,16 @@ def setup(self, cfg: DictConfig) -> None:
 
         utils.log_rank_zero(log, "Loss is initialized.")
 
+        try:
+            self._forward = config.instantiate(cfg.forward)
```
RdoubleA, line 296: What's the motivation for this change?

RdoubleA (commented):
I am a bit confused on why a custom forward class needs to be created, can't everything be contained in the custom loss module, as long as it follows a certain contract?

krammnic:
> I am a bit confused on why a custom forward class needs to be created, can't everything be contained in the custom loss module, as long as it follows a certain contract?

Do you want to make it in one class?

SalmanMohammadi:
Hey @krammnic. Thanks for opening this.

> What's the motivation for this change?

@RdoubleA to try and provide some context (@krammnic feel free to correct me here since this is your PR) - the core issue is that our DPO recipe(s) as they are today are not easily extensible to other DPO-esque losses. As the recipes are now, we could start to see that adding new losses would overall be harmful to the code quality. This is why we removed SimPO in #2063 - it was a "reference-free" loss, which meant the [logic for calculating the loss was slightly different](https://github.com/SalmanMohammadi/torchtune/blob/0140dc51289b0408eca1417d874f7920fde4fcb3/recipes/lora_dpo_distributed.py#L650). 

If a user wishes to add SimPO in today, they'd have to duplicate the recipe file and make the same changes which we removed. If you'd like to estimate the space of possible DPO-style losses and the degree of branching logic they require, I'd reccomend taking a look at [TRL's code](https://github.com/huggingface/trl/blob/fe4b5efe4e23f4331ba9c5b0c8bd92dc8302c287/trl/trainer/dpo_trainer.py#L911).

As it is, I'm not sure if this PR is heading in the right direction. In the example with `SimPO` which you've provided, we're still making the extra forward pass in the recipe to obtain the reference logprobs, and passing these into the `SimPOLoss` which would be instantiated, which would error out.

I think it would be helpful to enumerate the different cases we would need to handle to support the most commonly used DPO-style losses, which would allow us to generalize in the right place. I would need to spend a little more time to come up with a suitable design, but right now I'm leaning towards pulling most of the logic for forward passes and loss calculation (e.g. [from here](https://github.com/pytorch/torchtune/blob/90fd2d3afaab25cfcf63dd5ff160b6c31a283093/recipes/lora_dpo_single_device.py#L533) to [here](https://github.com/pytorch/torchtune/blob/90fd2d3afaab25cfcf63dd5ff160b6c31a283093/recipes/lora_dpo_single_device.py#L559)) out of the recipe, and into the losses. I'd rather duplicate this logic for losses that are very similar than over-generalize.

I'd love to hear your thoughts @krammnic (and @RdoubleA and anyone else who has stuck with my ramblings).

krammnic:
We can handle this case related to reference logprobs with adding parameter related to such possibility in the recipe. Speaking about the design overall, I actually could not find better alternative to save adaptiveness and code quality

> Hey @krammnic. Thanks for opening this.
> 
> > What's the motivation for this change?
> 
> @RdoubleA to try and provide some context (@krammnic feel free to correct me here since this is your PR) - the core issue is that our DPO recipe(s) as they are today are not easily extensible to other DPO-esque losses. As the recipes are now, we could start to see that adding new losses would overall be harmful to the code quality. This is why we removed SimPO in #2063 - it was a "reference-free" loss, which meant the [logic for calculating the loss was slightly different](https://github.com/SalmanMohammadi/torchtune/blob/0140dc51289b0408eca1417d874f7920fde4fcb3/recipes/lora_dpo_distributed.py#L650).
> 
> If a user wishes to add SimPO in today, they'd have to duplicate the recipe file and make the same changes which we removed. If you'd like to estimate the space of possible DPO-style losses and the degree of branching logic they require, I'd reccomend taking a look at [TRL's code](https://github.com/huggingface/trl/blob/fe4b5efe4e23f4331ba9c5b0c8bd92dc8302c287/trl/trainer/dpo_trainer.py#L911).
> 
> As it is, I'm not sure if this PR is heading in the right direction. In the example with `SimPO` which you've provided, we're still making the extra forward pass in the recipe to obtain the reference logprobs, and passing these into the `SimPOLoss` which would be instantiated, which would error out.
> 
> I think it would be helpful to enumerate the different cases we would need to handle to support the most commonly used DPO-style losses, which would allow us to generalize in the right place. I would need to spend a little more time to come up with a suitable design, but right now I'm leaning towards pulling most of the logic for forward passes and loss calculation (e.g. [from here](https://github.com/pytorch/torchtune/blob/90fd2d3afaab25cfcf63dd5ff160b6c31a283093/recipes/lora_dpo_single_device.py#L533) to [here](https://github.com/pytorch/torchtune/blob/90fd2d3afaab25cfcf63dd5ff160b6c31a283093/recipes/lora_dpo_single_device.py#L559)) out of the recipe, and into the losses. I'd rather duplicate this logic for losses that are very similar than over-generalize.
> 
> I'd love to hear your thoughts @krammnic (and @RdoubleA and anyone else who has stuck with my ramblings).

krammnic:
@SalmanMohammadi @RdoubleA Hey! What I need to do to proceed on this? Currently, I mainly use slightly modified torchtune from this PR as the feature is pretty important.

krammnic:
@SalmanMohammadi I see that full DPO was merged. Let me update this to align with it and our new design that we have discussed

krammnic:
@SalmanMohammadi Pushed the design that we discussed. If it is OK, then will run some tests with each recipe.

krammnic:
@ebsmothers As Salman is unavailable, after small discussion with Felipe, we concluded that you can review this relatively small but important feature. Would love to here some comments!

ebsmothers (commented):
@krammnic admittedly I am a bit out of the loop on these changes so pardon my ignorance here. But personally I don't think that `model` should be an argument to the forward method of a loss module. This is introducing some coupling that shouldn't exist -- namely any loss should be usable as a standalone component. 

Again I am probably missing some important background here, if you can share more details on what you're trying to support I'm happy to brainstorm some alternative paths here.

krammnic:
> @krammnic admittedly I am a bit out of the loop on these changes so pardon my ignorance here. But personally I don't think that `model` should be an argument to the forward method of a loss module. This is introducing some coupling that shouldn't exist -- namely any loss should be usable as a standalone component.
> 
> Again I am probably missing some important background here, if you can share more details on what you're trying to support I'm happy to brainstorm some alternative paths here.

Yep, let me clarify the idea. Currently we support only `DPOLoss` as a contrastive loss for our recipes. In some prevous PRs we have dropped the support of other contrastive losses due to the overcomplecation of the recipe. The key point is that we don't want to include all losses directly in the framework, because SOTA can be changed frequently and because recipe should be clean. Key observation: to support almost any contrastive loss, custom `forward` of the loss and `concatenated_forward` in the recipe is enough. Firstly, I wanted to pass them separately, but then we concluded that the better idea is to put both forwards directly in the loss. The idea of this PR to support possibility of such pass. Therefore, support any contrastive losses that users then may take from TRL or any other source.

Issue #2281, opened 2025-01-19
[RFC] Proposal for `tune cat` Command


Ankur-singh:
First of all, thank you very much for the wonderful package. I’ve started actively looking at the source code, and I must say it’s an absolute pleasure to read. It was difficult to stop myself from contributing.

### Context: 
Currently, the `tune ls` command allows users to list all available recipes and their corresponding configurations. However, it doesn’t provide an easy way to view the actual content of a configuration file itself. Before modifying or overriding parameters, users often need to inspect the contents of a specific configuration to understand its structure and settings.

### Proposal:
A new command to the `torchtune` framework: `tune cat`.

The `tune cat` command would fill this gap by allowing users to view the contents of a specific configuration file. It would work similarly to [`kubectl describe`](https://kubernetes.io/docs/reference/kubectl/generated/kubectl_describe/), providing a clear and detailed view of the configuration content.

The ideal workflow could be as follows:

1. Use `tune ls` to list all available configurations.
2. Use `tune cat <config>` to view the contents of a specific config file.
3. After inspecting the config, users can:
    - Override existing parameters directly via the command line, or
    - Copy the config using `tune cp` and modify it locally.

This feature would make the workflow more intuitive and user-friendly.

I’d love to hear your thoughts on this proposal. Do you think `tune cat` would be a valuable addition to the framework? Are there any other suggestions or improvements you would recommend?

Thank you for your time and feedback!

joecummings:
@Ankur-singh Thanks for your kind words and even kinder contributions to the torchtune library! They are much appreciated :)

On the whole, I'm very supportive of `tune cat`. Couple of (minor) follow ups:

1. Do you imagine extending this functionality to work with recipes as well? The `tune ls` and `tune cp` commands cover both recipes and configs, so `tune cat` only working with configs would break that paradigm a little bit. However, the value of `cat` for configs is clear whereas the value of `cat` for recipes is not. Personally, I don't see this extending to recipes, but that means that you should provide a clear error message if someone wants to `cat` a recipe file, which also means you'll need to know what is a recipe, a config, and is not in the torchtune system.
2. Should a user be able to `tune cat` a local file? 

[...]

PR #2298, opened 2025-01-24
[WIP] 'tune cat' command for pretty printing configuration files

Ankur-singh:
#### Context
- [X] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

Please link to any issues this PR addresses. #2281 

#### Changelog
* Implemented `cat` subcommand for the CLI
* Wrote basic test-cases for `tune cat` 

#### Test plan

- [X] include relevant commands and any other artifacts in this summary (pastes of loss curves, eval results, etc.)

When file doesn't exist:
```bash
(tune) ankur@nuc:~/github/torchtune$ tune cat llama4/8B_full.yaml
usage: tune cat [-h] config_name
tune cat: error: Config 'llama4/8B_full.yaml' not found.
```

When some other file is passed:
```bash
(tune) ankur@nuc:~/github/torchtune$ tune cat llama4/8B_full.json
usage: tune cat [-h] config_name
tune cat: error: Invalid config format: 'llama4/8B_full.json'. Must be YAML (.yaml/.yml)
```

When recipe name is passed:
```bash
(tune) ankur@nuc:~/github/torchtune$ tune cat quantize
'quantize' is a recipe, not a config. Please use a config name.
```

When correct config is passed:
```bash
(tune) ankur@nuc:~/github/torchtune$ tune cat llama2/7B_full
output_dir: /tmp/torchtune/llama2_7B/full
tokenizer:
    path: /tmp/Llama-2-7b-hf/tokenizer.model
    max_seq_len: null
dataset:
    packed: false
seed: null
shuffle: true
model:
    _component_: torchtune.models.llama2.llama2_7b
checkpointer:
    checkpoint_dir: /tmp/Llama-2-7b-hf
    checkpoint_files:
    - pytorch_model-00001-of-00002.bin
    - pytorch_model-00002-of-00002.bin
    recipe_checkpoint: null
    output_dir: ${output_dir}
    model_type: LLAMA2
resume_from_checkpoint: false
batch_size: 2
epochs: 1
optimizer:
    fused: true
    lr: 2e-5
loss:
    _component_: torchtune.modules.loss.CEWithChunkedOutputLoss
max_steps_per_epoch: null
compile: false
optimizer_in_bwd: false
device: cuda
dtype: bf16
metric_logger:
    log_dir: ${output_dir}/logs
log_every_n_steps: 1
log_peak_memory_stats: true
profiler:
    _component_: torchtune.training.setup_torch_profiler
    enabled: false
    output_dir: ${output_dir}/profiling_outputs
    cpu: true
    cuda: true
    profile_memory: false
    with_stack: false
    record_shapes: true
    with_flops: false
    wait_steps: 5
    warmup_steps: 3
    active_steps: 2
    num_cycles: 1
```

#### UX

### DIFF
--- a/torchtune/_cli/cat.py
+++ b/torchtune/_cli/cat.py
@@ -10,12 +10,11 @@ import textwrap
 from pathlib import Path
 from typing import List, Optional
 
-import torchtune
 import yaml
 from torchtune._cli.subcommand import Subcommand
 from torchtune._recipe_registry import Config, get_all_recipes
 
-ROOT = Path(torchtune.__file__).parent.parent
+ROOT = Path(__file__).parent.parent.parent
 
 
 class Cat(Subcommand):
@@ -26,22 +25,29 @@ class Cat(Subcommand):
         self._parser = subparsers.add_parser(
             "cat",
             prog="tune cat",
-            help="Pretty print a config",
-            description="Pretty print a config",
+            help="Pretty print a config, making it easy to override parameters when using `tune run`",
+            description="Pretty print a config, making it easy to override parameters when using `tune run`",
             epilog=textwrap.dedent(
                 """\
                 examples:
                     $ tune cat llama2/7B_full
-                    model: llama2
-                    size: 7B
-                    task: full_finetune
+                    output_dir: /tmp/torchtune/llama2_7B/full
+                    tokenizer:
+                        _component_: torchtune.models.llama2.llama2_tokenizer
+                        path: /tmp/Llama-2-7b-hf/tokenizer.model
+                        max_seq_len: null
                     ...
 
-                    $ tune cat non_existent_config
-                    Config 'non_existent_config' not found.
+                    # Pretty print the config in sorted order
+                    $ tune cat llama2/7B_full --sort
 
-                    $ tune cat some_recipe
-                    'some_recipe' is a recipe, not a config. Please use a config name.
+                    # Pretty print the contents of LOCALFILE.yaml
+                    $ tune cat LOCALFILE.yaml
+
+                    # Example of launching a run overriding a key found via the `tune cat` command:
+                    $ tune run full_finetune_distributed --config llama2/7B_full output_dir=./
+
+                    # Note: You can find all the cat-able configs via the `tune ls` command.
                 """
             ),
             formatter_class=argparse.RawTextHelpFormatter,
@@ -50,6 +56,9 @@ class Cat(Subcommand):
             "config_name", type=str, help="Name of the config to print"
         )
         self._parser.set_defaults(func=self._cat_cmd)
+        self._parser.add_argument(
+            "--sort", action="store_true", help="Print the config in sorted order"
+        )
 
     def _get_all_recipes(self) -> List[str]:
         return [recipe.name for recipe in get_all_recipes()]
@@ -61,7 +70,7 @@ class Cat(Subcommand):
                 if config.name == config_str:
                     return config
 
-    def _print_file(self, file: str) -> None:
+    def _print_yaml_file(self, file: str, sort_keys: bool) -> None:
         try:
             with open(file, "r") as f:
                 data = yaml.safe_load(f)
@@ -70,7 +79,7 @@ class Cat(Subcommand):
                         yaml.dump(
                             data,
                             default_flow_style=False,
-                            sort_keys=False,
+                            sort_keys=sort_keys,
                             indent=4,
                             width=80,
                             allow_unicode=True,
@@ -116,4 +125,4 @@ class Cat(Subcommand):
             self._parser.error(f"Config '{config_str}' not found.")
             return
 
-        self._print_file(str(config_path))
+        self._print_yaml_file(str(config_path), args.sort)
--- a/torchtune/_cli/cat.py
+++ b/torchtune/_cli/cat.py
@@ -25,8 +25,8 @@ class Cat(Subcommand):
         self._parser = subparsers.add_parser(
             "cat",
             prog="tune cat",
-            help="Pretty print a config, making it easy to override parameters when using `tune run`",
-            description="Pretty print a config, making it easy to override parameters when using `tune run`",
+            help="Pretty print a config, making it easy to know which parameters you can override with `tune run`.",
+            description="Pretty print a config, making it easy to know which parameters you can override with `tune run`.",
             epilog=textwrap.dedent(
                 """\
                 examples:
@@ -44,10 +44,10 @@ class Cat(Subcommand):
                     # Pretty print the contents of LOCALFILE.yaml
                     $ tune cat LOCALFILE.yaml
 
-                    # Example of launching a run overriding a key found via the `tune cat` command:
+                You can now easily override a key based on your findings from `tune cat`:
                     $ tune run full_finetune_distributed --config llama2/7B_full output_dir=./
 
-                    # Note: You can find all the cat-able configs via the `tune ls` command.
+                Need to find all the "cat"-able configs? Try `tune ls`!
                 """
             ),
             formatter_class=argparse.RawTextHelpFormatter,
--- a/torchtune/_cli/cat.py
+++ b/torchtune/_cli/cat.py
@@ -1,0 +1,119 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+import argparse
+import textwrap
+
+from pathlib import Path
+from typing import List, Optional
+
+import torchtune
+import yaml
+from torchtune._cli.subcommand import Subcommand
+from torchtune._recipe_registry import Config, get_all_recipes
+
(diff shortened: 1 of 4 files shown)


Ankur-singh:
Still a work in progress, but I’d love to gather as much feedback as possible while it's still in the oven!

cc @RdoubleA @joecummings

SalmanMohammadi:
Thanks so much for the awesome RFC and for putting this up @Ankur-singh!

Would you be able to attach some example output(s) from using the command to the PR description?

Ankur-singh:
@SalmanMohammadi Thank you for the comment. I have added the command outs for different scenarios. This should make it easy for other as well. Again, thank you very much.

torchtune/_cli/cat.py
```diff
@@ -0,0 +1,119 @@
...
+                    $ tune cat llama2/7B_full
+                    model: llama2
+                    size: 7B
+                    task: full_finetune
+                    ...
+
+                    $ tune cat non_existent_config
+                    Config 'non_existent_config' not found.
+
+                    $ tune cat some_recipe
```
joecummings, line 43: I don't think we need an example for recipe

torchtune/_cli/cat.py
```diff
@@ -0,0 +1,119 @@
...
+                    'some_recipe' is a recipe, not a config. Please use a config name.
```
joecummings, line 44: Can you add an additional line down here saying that they can find all the `cat`-able configs via the `tune ls` command? You can also provide an example of launching a run overriding a key that was found via the new `tune cat` command.

torchtune/_cli/cat.py
```diff
@@ -0,0 +1,119 @@
...
+
+from pathlib import Path
+from typing import List, Optional
+
+import torchtune
+import yaml
+from torchtune._cli.subcommand import Subcommand
+from torchtune._recipe_registry import Config, get_all_recipes
+
+ROOT = Path(torchtune.__file__).parent.parent
```
joecummings, line 18: Ugh, we need to find a different way to do this (laziness on my part, I apologize). I suspect the call to `import torchtune` significantly slows down the execution of the CLI. 

The key part here is that we want to find the underlying project in which the torchtune package is installed. For a local installation AND for a PyPI installation, it *should* be two levels up from the __init__.py file. So theoretically it's three levels up from any of the CLI files.

torchtune/_cli/cat.py
```diff
@@ -0,0 +1,119 @@
...
+        return [recipe.name for recipe in get_all_recipes()]
+
+    def _get_config(self, config_str: str) -> Optional[Config]:
+        # Search through all recipes
+        for recipe in get_all_recipes():
+            for config in recipe.configs:
+                if config.name == config_str:
+                    return config
+
+    def _print_file(self, file: str) -> None:
```
joecummings, line 64: `_print_yaml_file` is a little more descriptive.

joecummings (commented):
Command looks good, just a few comments on docstrings and impl

torchtune/_cli/cat.py
```diff
@@ -0,0 +1,119 @@
...
+class Cat(Subcommand):
+    """Holds all the logic for the `tune cat` subcommand."""
+
+    def __init__(self, subparsers: argparse._SubParsersAction):
+        super().__init__()
+        self._parser = subparsers.add_parser(
+            "cat",
+            prog="tune cat",
+            help="Pretty print a config",
+            description="Pretty print a config",
```
joecummings, line 30: Can you add something related to WHY someone might want to do this? E.g. pretty print a config, making it easy to know which keys you can override when launching a training run (or something a little more concise)

torchtune/_cli/cat.py
```diff
@@ -0,0 +1,119 @@
...
+            epilog=textwrap.dedent(
+                """\
+                examples:
+                    $ tune cat llama2/7B_full
+                    model: llama2
+                    size: 7B
+                    task: full_finetune
+                    ...
+
+                    $ tune cat non_existent_config
```
joecummings, line 40: No need for negative examples

torchtune/_cli/cat.py
```diff
@@ -0,0 +1,119 @@
...
+        super().__init__()
+        self._parser = subparsers.add_parser(
+            "cat",
+            prog="tune cat",
+            help="Pretty print a config",
+            description="Pretty print a config",
+            epilog=textwrap.dedent(
+                """\
+                examples:
+                    $ tune cat llama2/7B_full
```
joecummings, line 34: Can you add an example of `tune cat LOCALFILE.yaml`

torchtune/_cli/cat.py
```diff
@@ -0,0 +1,119 @@
...
+    def _print_file(self, file: str) -> None:
+        try:
+            with open(file, "r") as f:
+                data = yaml.safe_load(f)
+                if data:
+                    print(
+                        yaml.dump(
+                            data,
+                            default_flow_style=False,
+                            sort_keys=False,
```
SalmanMohammadi, line 73: I personally find the sorted configs easier to read when printed.

SalmanMohammadi:
Would you be up for adding a little section for your shiny new command in the docs? : )

https://github.com/pytorch/torchtune/blob/main/docs/source/tune_cli.rst

torchtune/_cli/cat.py
Ankur-singh: Can you please it further? What do you mean by "descriptive"?

torchtune/_cli/cat.py
Ankur-singh: Thats a great idea. In fact, this is was the original motivation.

torchtune/_cli/cat.py
SalmanMohammadi: cc @ebsmothers @joecummings thoughts?

torchtune/_cli/cat.py
joecummings: It's a nit for sure, but this doesn't just print *any* file. It only takes in and prints a YAML file, therefore the name should be something like `print_yaml_file`.

torchtune/_cli/cat.py
joecummings: I like the idea of a `--sort` argument. I agree that sorted configs *are* a little easier to read; however, they wouldn't look like our configs on Github, which would be confusing so the default should be to match what we show on Github.

torchtune/_cli/cat.py
Ankur-singh: Yes, importing `torchtune` is slow. I will address this issue. We will need to update other CLI files that import `torchtune` as well, but it might be better to handle that in a separate PR to avoid scope creep.

torchtune/_cli/cat.py
Ankur-singh: Yes, thats a good point. Will update the function name.

Ankur-singh:
@SalmanMohammadi and @joecummings Thank you so much for all the amazing feedback. I have made all the requested changes.

torchtune/_cli/cat.py
joecummings: Yep, let's do this in another PR.

torchtune/_cli/cat.py
```diff
@@ -0,0 +1,128 @@
...
+                        max_seq_len: null
+                    ...
+
+                    # Pretty print the config in sorted order
+                    $ tune cat llama2/7B_full --sort
+
+                    # Pretty print the contents of LOCALFILE.yaml
+                    $ tune cat LOCALFILE.yaml
+
+                    # Example of launching a run overriding a key found via the `tune cat` command:
```
joecummings, line 47: super nit, but take a look at how I did this for `tune ls`: https://github.com/pytorch/torchtune/blob/d4465c8b21629949a404f821528badb35962502f/torchtune/_cli/ls.py#L37

It should be unindented OUT of the examples section and instead be considered a followup. 

e.g.

```
examples:
	$ tune cat llama2/7B_full
	...

	$ tune cat LOCALFILE.yaml
	...

You can now easily override a key based on your findings from `tune cat`:
	$ tune run full_finetune_distributed ...

Need to find all the "cat"-able configs? Try `tune ls`!
```

torchtune/_cli/cat.py
```diff
@@ -0,0 +1,128 @@
...
+
+class Cat(Subcommand):
+    """Holds all the logic for the `tune cat` subcommand."""
+
+    def __init__(self, subparsers: argparse._SubParsersAction):
+        super().__init__()
+        self._parser = subparsers.add_parser(
+            "cat",
+            prog="tune cat",
+            help="Pretty print a config, making it easy to override parameters when using `tune run`",
```
joecummings, line 28: nit: "making it easy to *know* which parameters you can override with `tune run`.

joecummings (commented):
This looks awesome! Just two more nits I swear :)

torchtune/_cli/cat.py
Ankur-singh: Done!

torchtune/_cli/cat.py
Ankur-singh: Done

joecummings (approved):
🫡🫡🫡

PR #2302, opened 2025-01-27
Flux Model

calvinpelletier:
#### Context
This adds the main Flux flow-matching model to TorchTune.

NOTE: @pbontrager had mostly finished an implementation of this model before going on leave, so this implementation is temporary to unblock the rest of the Flux PRs. I just copied the code from the official Flux repo with some minimal changes. We can replace it with @pbontrager 's version when he returns.

More Flux PRs coming soon.

#### Changelog
* Flux flow model implementation and builders
* LoRA builders
* Unit tests
* Removed input image shape assertion from autoencoder (the autoencoder can be used with many image resolutions)
* Some utility functions for predicting the noise in an image latent using the flow model
* Added checkpoint loading/saving logic for the flow model (which is currently a no-op since the temporary torchtune implementation is the same as the huggingface one, see note above) 

#### Usage
```bash
tune download black-forest-labs/FLUX.1-dev --output-dir /tmp/flux
```

```python
from torchtune.models.flux import flux_1_dev_flow_model
from torchtune.training.checkpointing import FullModelHFCheckpointer

model = flux_1_dev_flow_model()
checkpointer = FullModelHFCheckpointer(
    "/tmp/flux",
    ["flux1-dev.safetensors"],
    "FLUX",
    "/tmp/flux_",
)
sd = checkpointer.load_checkpoint()["model"]
model.load_state_dict(sd)
```

#### Test plan

Manual testing:
- [x] verified that the torchtune implemention of the flow model has perfect parity with the official implementation and same performance
- [x] verified that the LoRA trains properly

Checklist:

### DIFF
--- a/torchtune/models/flux/_flow_model.py
+++ b/torchtune/models/flux/_flow_model.py
@@ -1,0 +1,443 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+import math
+from dataclasses import dataclass
+
+import torch
+from torch import nn, Tensor
+
+
+class FluxFlowModel(nn.Module):
+    """
+    Flow-matching model for Flux v1 from https://github.com/black-forest-labs/flux
+
+    Args:
+        in_channels (int): The number of channels of the "img" input (the latent from the autoencoder).
+        out_channels (int): The number of channels of the output.
+        vec_in_dim (int): The number of dimensions of the "y" input (the CLIP embedding).
+        context_in_dim (int): The number of dimensions of the "txt" input (the T5 text encoding).
+        hidden_size (int): The inner dimension of the model.
+        mlp_ratio (float): The size of the dimension relative to the hidden size of the MLPs in the attention blocks.
+        num_heads (int): The number of attention heads.
+        depth (int): The number of double stream blocks.
+        depth_single_blocks (int): The number of single stream blocks.
+        axes_dim (list[int]): The dimension of the rope embeddings.
+        theta (int): The theta of the rope embeddings.
+        qkv_bias (bool): Enable bias for the QKV projections.
+        use_guidance (bool): Enable guidance (True for dev models and False for schnell).
+
+    Raises:
+        ValueError: If hidden_size, num_heads, axes_dim, or pe_dim is invalid.
+    """
+
+    def __init__(
+        self,
+        in_channels: int,
+        out_channels: int,
+        vec_in_dim: int,
+        context_in_dim: int,
+        hidden_size: int,
+        mlp_ratio: float,
+        num_heads: int,
+        depth: int,
+        depth_single_blocks: int,
+        axes_dim: list[int],
+        theta: int,
+        qkv_bias: bool,
+        use_guidance: bool,
+    ):
+        super().__init__()
+
+        self.in_channels = in_channels
+        self.out_channels = out_channels
+        if hidden_size % num_heads != 0:
+            raise ValueError(
+                f"Hidden size {hidden_size} must be divisible by num_heads {num_heads}"
+            )
+        pe_dim = hidden_size // num_heads
+        if sum(axes_dim) != pe_dim:
+            raise ValueError(f"Got {axes_dim} but expected positional dim {pe_dim}")
+        self.hidden_size = hidden_size
+        self.num_heads = num_heads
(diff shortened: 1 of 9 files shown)


torchtune/models/flux/_flow_model.py
```diff
@@ -0,0 +1,443 @@
...
+        super().__init__()
+        self.in_layer = nn.Linear(in_dim, hidden_dim, bias=True)
+        self.silu = nn.SiLU()
+        self.out_layer = nn.Linear(hidden_dim, hidden_dim, bias=True)
+
+    def forward(self, x: Tensor) -> Tensor:
+        return self.out_layer(self.silu(self.in_layer(x)))
+
+
+class RMSNorm(torch.nn.Module):
```
ebsmothers, line 191: Same here.. we already have this, right?

torchtune/models/flux/_flow_model.py
```diff
@@ -0,0 +1,443 @@
...
+        super().__init__()
+        self.num_heads = num_heads
+        head_dim = dim // num_heads
+
+        self.qkv = nn.Linear(dim, dim * 3, bias=qkv_bias)
+        self.norm = QKNorm(head_dim)
+        self.proj = nn.Linear(dim, dim)
+
+    def forward(self, x: Tensor, pe: Tensor) -> Tensor:
+        qkv = self.qkv(x)
```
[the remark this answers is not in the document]
ebsmothers: Separately, I may be missing something.. is there a gap in our existing `MultiHeadAttention` that necessitates the use of these? I think we should support Q norm, K norm, RoPE, all of that there. But admittedly I'm kind unfamiliar with some of the Flux details

torchtune/models/flux/_model_builders.py
```diff
@@ -67,7 +69,154 @@ def flux_1_autoencoder(
...
+        context_in_dim=4096,
+        hidden_size=3072,
+        mlp_ratio=4.0,
+        num_heads=24,
+        depth=19,
+        depth_single_blocks=38,
+        axes_dim=[16, 56, 56],
+        theta=10_000,
+        qkv_bias=True,
+        use_guidance=False,
```
ebsmothers, line 121: Noob q: the only diff between dev and schnell is that dev uses guidance and schnell doesn't?

torchtune/models/flux/_model_builders.py
```diff
@@ -67,7 +69,154 @@ def flux_1_autoencoder(
...
+        rank=lora_rank,
+        alpha=lora_alpha,
+        dropout=lora_dropout,
+        quantize_base=quantize_base,
+        use_dora=use_dora,
+    )
+    return model
+
+
+def _replace_linear_with_lora(
```
ebsmothers, line 193: 👀 @felipemello1

ebsmothers: I think we've talked about this one for a bit anyways, so great to see it finally materializing. Maybe let's put in `modules/peft/_utils.py` or something?

ebsmothers: One other thing we'll have to watch out for here.. for NF4-quantized layers we also add state dict hooks to prevent higher memory on checkpoint save. It may be worth including that logic in this utility

torchtune/training/checkpointing/_checkpointer.py
```diff
@@ -709,6 +718,8 @@ def save_checkpoint(
                     dim=self._config["hidden_size"],
                     head_dim=self._config.get("head_dim", None),
                 )
+            elif self._model_type == ModelType.FLUX:
+                pass  # the torchtune Flux model state dict is identical to the huggingface one
```
ebsmothers, line 722: Not a problem for this PR, but in this big if/elif/.../elif/else I think this (no weight conversion) should actually be the else, not Llama2 (for whatever reason)

torchtune/models/flux/_util.py
```diff
@@ -0,0 +1,165 @@
...
+
+from typing import Optional
+
+import torch
+from torch import Tensor
+
+from torchtune.models.flux._flow_model import FluxFlowModel
+
+PATCH_HEIGHT, PATCH_WIDTH = 2, 2
+POSITION_DIM = 3
```
ebsmothers, line 15: Noob q: these are constant in all cases?

ebsmothers (commented):
I'm excited to see this PR! This is just a first pass, most of my comments are around the Flux model components. I don't have a ton to say about the noise prediction yet (I may just need to do more research/see it in action). Still need to look at the single/double stream blocks later, so will probably come back with a few more comments

PR #2318, opened 2025-01-29
[WIP] 2D parallelism for training

joecummings:


recipes/full_finetune_distributed.py
```diff
@@ -137,8 +139,11 @@ def __init__(self, cfg: DictConfig) -> None:
             )
             self._log_peak_memory_stats = False
 
-        _, rank = utils.get_world_size_and_rank()
-        self._is_rank_zero = rank == 0
+        # Distributed variables
+        self.world_size, self.rank = utils.get_world_size_and_rank()
+        self._is_rank_zero = self.rank == 0
+        self.parallelize_plan = config.instantiate(cfg.get("parallelize_plan"))
```
SalmanMohammadi, line 145: FYI the default parallelize plan we have didn't work when I was using it with the smaller Llama3 models with tied embeddings.

PR #2321, opened 2025-01-31
Refactor validate missing for LoRA + deprecate param utility

RdoubleA:
#### Context
- [x] add a new feature
- [x] fix a bug
- [ ] other (please add here)

`validate_missing_and_unexpected_for_lora` takes two sets of keys: base_missing and lora_missing. These are produced when loading the base model weights and adapter model weights separately with load state dict. When loading the base weights, the lora weights will be missing, and vice versa. It then checks against the lora config parameters to make sure everything was loaded correctly.

This is a bit rigid, because 1) it's tied to config params and not the actual model and 2) it checks for a hardcoded list of lora params (like w1, w2, w3 for MLP). When adding new lora parameters, such as for Mixture-of-Experts, this validation utility gets angry.

Since we have to deprecate existing parameters, I've also added a quick decorator similar to `@deprecated` to warn when a deprecated parameter is used. It also logs only once, only on rank zero! Hooray~
```
FSDP is enabled. Instantiating model and loading checkpoint on Rank 0 ...
NCCL version 2.21.5+cuda12.4
/data/users/rafiayub/torchtune-rafiayub/recipes/lora_finetune_distributed.py:513: FutureWarning: lora_attn_modules is deprecated for validate_missing_and_unexpected_for_lora and will be removed in future versions. Please use state_dict_keys instead.
  validate_missing_and_unexpected_for_lora(
/data/users/rafiayub/torchtune-rafiayub/torchtune/utils/_logging.py:143: FutureWarning: apply_lora_to_mlp is deprecated for validate_missing_and_unexpected_for_lora and will be removed in future versions. Please use state_dict_keys instead.
  return obj(*args, **kwargs)
/data/users/rafiayub/torchtune-rafiayub/torchtune/utils/_logging.py:143: FutureWarning: apply_lora_to_output is deprecated for validate_missing_and_unexpected_for_lora and will be removed in future versions. Please use state_dict_keys instead.
  return obj(*args, **kwargs)
Instantiating model and loading checkpoint took 2.85 secs
Memory stats after model init:
        GPU peak memory allocation: 1.24 GiB
        GPU peak memory reserved: 1.26 GiB
        GPU peak memory active: 1.24 GiB
```

#### Approach
Use set comparisons to deduce what is missing and check against the model's actual lora state dict (done by filtering by "lora" or "magnitude" in the param name. Still a bit hardcoded but more flexible). This does involve passing the model state dict to the function. Previously, the function assumed that querying model state dict incurred memory hits, hence the workarounds to determine lora keys in the model state dict. However, the model state dict is already loaded before you want to validate it anyway, so there is no memory hit. We can just pass in the model state dict directly to make the logic easier.

#### Changelog
* Refactor `validate_missing_and_unexpected_for_lora`
* Update unit test
* Update callsites in all recipes

#### Test plan

#### UX


ebsmothers (approved):

RdoubleA:
I don't even remember what this is for

PR #2301, opened 2025-01-27
Multinode support in torchtune

joecummings:
### Context

This has been an explicit ask by several users (#2161, #2142) and although things *should* work fairly easily OOTB, we hadn't been able to test it and therefore didn't want to make any claims that we couldn't back up. Therefore, I sent myself on a quest to set up my own SLURM cluster, waste a lot of money, and test our multi-node scripts!

### Changes
* Fully deprecate `get_world_size_and_rank` from `training` in recipes
* Add `get_distributed_backend` method that mirrors torchtitan's
* Update `full_finetune_distributed` to utilize `get_distributed_backend` instead of setting "cuda:nccl,cpu:gloo"
* Add `full_finetune_multinode.slurm` script
* Add tutorial on multi node in torchtune

### Testing

Experiments were run with a SLURM cluster w/ 2 worker nodes (H100 w/ Infiniband) set up on Nebius AI, following this tutorial: https://docs.nebius.com/compute/clusters/slurm.

```
Node IP: slurm-worker-1
I0129 19:29:59.590000 8654 torch/distributed/run.py:675] Using nproc_per_node=8.
W0129 19:29:59.591000 8654 torch/distributed/run.py:792]
W0129 19:29:59.591000 8654 torch/distributed/run.py:792] *****************************************
W0129 19:29:59.591000 8654 torch/distributed/run.py:792] Setting OMP_NUM_THREADS environment variable for each process to be 1 in default, to avoid your system being overloaded, please further tune the variable for optimal performance in your application as needed.
W0129 19:29:59.591000 8654 torch/distributed/run.py:792] *****************************************
I0129 19:29:59.591000 8654 torch/distributed/launcher/api.py:195] Starting elastic_operator with launch configs:
I0129 19:29:59.591000 8654 torch/distributed/launcher/api.py:195]   entrypoint       : /mnt/slurm/torchtune/lib/python3.10/site-packages/recipes/full_finetune_distributed.py
I0129 19:29:59.591000 8654 torch/distributed/launcher/api.py:195]   min_nodes        : 2
I0129 19:29:59.591000 8654 torch/distributed/launcher/api.py:195]   max_nodes        : 2
I0129 19:29:59.591000 8654 torch/distributed/launcher/api.py:195]   nproc_per_node   : 8
I0129 19:29:59.591000 8654 torch/distributed/launcher/api.py:195]   run_id           : 101
I0129 19:29:59.591000 8654 torch/distributed/launcher/api.py:195]   rdzv_backend     : c10d
[...]

PR #2326, opened 2025-02-01
R1-Style distributed GRPO

RedTachyon:
#### Context
- [x] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

After some discussions on another PR and on Discord, this is the current state of my distributed GRPO implementation. I'm still iterating on this, prioritizing checking whether it *actually* works.

I have some early successes, but it's too soon to proclaim victory. Soon I'll probably also adapt it to a multinode workflow when #2301 is merged (or just snatch some code from there), because RL is sufficiently resource-hungry that single-node training isn't really an option for anything even moderately serious.

Right now the repo/PR is very messy and in a researchy state, to find something that works. Once it does work, I'll start cleaning it up to meet OSS standards. I'm putting it here to be able to keep track of the diffs and for potential discussions.

The rest will be filled when possible/relevant:

---

#### Changelog
*

#### Test plan

#### UX


musabgultekin:
Im extremely interested in this. 
One extra suggestion is that we could technically run the frozen reference policy on another device through something like SGLang or vLLM. That way, we only hold the policy in the VRAM.
One potential drawback would be the inner implementations of these engines might cause subtle differences and that might cause issues on advantage&loss calculation.

akashc1:
@RedTachyon thanks for the implementation! Please let me know if there's anything you'd like support on, I'm happy to help!

For my own use, could you please share how you set up the environment/commands to launch training?

RedTachyon:
Hi everyone, glad to see people interested in contributing to this implementation! I'm happy to say that the core implementation "just works" - on a recent run I did the following process:

1. Take base model Llama 3B
2. Train it with SFT for 1 epoch on 1/10th of the GSM8k train set, using the R1 prompt template (note: extremely quick training, took like 3 minutes on a single node)
3. Continue training the resulting model on the remaining 9/10ths of the GSM8k train set

At the moment, everything follows the R1 paper relatively closely - format is <think>cot</think> <answer>42</answer>, and reward computation is done by XML parsing the response, and checking the answer within <answer> tags for a perfect match (failed parse = 0 reward). There's some

Base model by itself struggles with the format (and knowing when to output <|eos|>), so its performance is just bad. The SFT-trained model follows the format well, but kinda sucks at math, so it tends to get about 10% success rate on GSM8k. 

Continuing the training with GRPO, the model climbs up to a ~60% success rate! (and can probably go higher if it keeps running for longer)

image


Caveats: this estimate is based on the training set questions, without a separate eval - but a majority of the improvement happens before the first epoch finishes.

Which brings me to the list of things that still need to be done. I'll probably move it to the top comment on the PR later, but for now here it is:

- [ ] A proper eval workflow, so that every once in a while we run a full eval on the test set
- [ ] Adding proper (unit) tests, compliant with the normal torchtune testing workflow
- [ ] Adding proper documentation to everything
- [ ] General cleanup of messy research code
- [ ] Refactoring of the GRPO losses (there's also a research-y question of what loss should be used - there's some ambiguity in the paper and reference implementations)
- [ ] More modular approach to reward computation (probably as a component with the regular OmegaConf setup)
- [ ] Probably more frequent checkpointing, not just tied to epochs (since epochs are sloooow with RL)
- [ ] Memory profiling and experiments on "controlled" hardware (a recipe tuned to work ~optimally on a node of 8xH100, or on a single H100, or on smaller hardware with e.g. LoRA)
- [ ] Dataset cleanup (gsm8k is functional but could use some polish, there's also the MATH dataset that's more challenging
- [ ] Maybe a single-device version? (might be memory-sensitive)
- [ ] Try to improve generation speed by using vLLM (or something else)?
- [ ] Probably more to be found soon

I'll start going through some of these things, and if someone wants to contribute, please mention it so that we don't duplicate the effort.

Later today I'll also push my sbatch files to run the full pipeline so that anyone can run it on slurm - for single-node experiments, regular torchtune runner should work fine.

Regarding the organization: it's probably best to keep everything under a shared PR (i.e. this), and make other intermediate PRs into this branch. So various people can contribute parts of the GRPO setup, we merge it into this branch, and then when it's all done, the maintainers can review the complete thing and merge it into main. 
It might be better to move the code to a branch on this repo instead of my fork - in any case, I'm happy to adapt it to whatever the maintainers think is best (@SalmanMohammadi @felipemello1 - not sure who's the right person to bother about the managementy things)


CC: @musabgultekin @akashc1 @ianbarber

felipemello1:
It would be easier for people to contribute if we could land a simplified version of this PR in /dev. We can then have the other items in this list as new PRs. It will be much easier to review and get things going. What do you think?

musabgultekin:
I would love to contribute. 

Like Felipe mentioned, it would be actually easier for me or anyone else to run/test/contribute, if we simplified this to its bare minimum first, and then add features as we go.

Some simplifying suggestions:
- No need for mps support for now.
- Starting with only single node or distributed recipe, rather than both.
- Instead of both math and gsm8k datasets, only landing one of them.

Etc

RedTachyon:
@felipemello1 What do you mean exactly by /dev? Is that some "centralized" dev branch, or maybe just a custom branch under this repo for this effort?

felipemello1:
> @felipemello1 What do you mean exactly by /dev? Is that some "centralized" dev branch, or maybe just a custom branch under this repo for this effort?

here: https://github.com/pytorch/torchtune/tree/main/recipes/dev

Lets get some simple recipe here that works. The recipe doesnt need to be perfect.

However, the other modules that are not in /dev need a bit more scrutiny (e.g. datasets). 

If you can simplify this PR (maybe make a new one if thats easier), we can approve in smaller chunks and folks can start building on top of it. Whatever works best for you.

felipemello1:
> Everything should generally be backwards compatible, but not sufficiently tested - is it fine to just merge it, if the only recipes that use it are ones in /dev?

If its backwards compatible and blast radius is small and we review, yes! Its just that this PR is 4000 lines long already. I imagine that a lot of it is the recipe+configs though. Let us know when you have a simple version that we can review, and lets land it :)

nazrak-atlassian:
Very excited for this - on the multi-node note, it looks like https://github.com/pytorch/torchtune/pull/2330 is close to landing. I'm not sure if there's any reason FSDP + TP wouldn't be compatible for GRPO, but wanted to flag (and might be nice to include in the multi-node recipe, if it lands before the GRPO PR)

nathan-az:
Very excited for this - on the multi-node note, it looks like https://github.com/pytorch/torchtune/pull/2330 is close to landing. 

I'm not sure if there's any reason FSDP + TP wouldn't be immediately compatible with GRPO, but wanted to flag this! (And might be nice to include in the multi-node recipe, if it lands before the GRPO PR)

RedTachyon:
@felipemello1 this shooooould be ready for review now. I cleaned everything up a bit, validated that it (still) runs, and removed whatever old code that doesn't need to be here yet. There's still 2500 sloc, but the vast majority is in the dev recipe/configs.

The parts that I think need the most scrutiny are:
- Changes to `generate` - an extra argument `return_logits` (defaults to original behavior), and a multi-device-aware stop token processing
- SFTDataset getting an extra argument of `filter_kwargs`, to make dataset filtering a bit more flexible (in particular, filtering by index). Note that in a very strict sense, this is a breaking change, because before this, `filter_kwargs` would automatically get consumed by `**load_dataset_kwargs`. Which would be insane if anyone tried to do that. But good to keep this in mind. (I'd probably also suggest getting rid of the `**load_dataset_kwargs` in favor of a normal dictionary input, but that's a separate issue/PR)

Just about everything else is stand-alone and shouldn't interfere with old code.

ianbarber:
@RedTachyon I'm happy to drop in the LoRA support when this lands re your todo list, great work getting it cleaned up!

RedTachyon:
@joecummings Are you sure this isn't possible right now? On the right I have a checked button that says "If checked, users with write access to pytorch/torchtune can add new commits to your main branch.", so in principle you should be able to just do it now.

I remember mucking about with this setting initially when I wasn't sure what would be the balance of research/OSS of this specific codebase, so I might have messed something up - in which case I'll need to dig into GH settings to figure it out

---

@ianbarber afaik the plan now is that this will get merged to main, and after that we can collaboratively start refining it and adding new features. I'll probably make a new issue with the checklist for tracking, but then you should be able to add LoRA whenever with a new PR.

felipemello1:
fyi Joe, Evan and I will be reviewing it between today and tomorrow

felipemello1 (commented):
initial review, will come back later. Mostly nits and questions for now.

torchtune/datasets/_rl.py
```diff
@@ -0,0 +1,169 @@
...
+    def _prepare_sample(self, sample: Mapping[str, Any]) -> Dict[str, Any]:
+        transformed_sample = self._problem_transform(
+            sample
+        )  # keys "question" and "answer"
+
+        question = BASE_PROMPT % transformed_sample["question"]
+
+        q_tokens = self._tokenizer.encode(question, add_eos=False)
+        mask = [1 for _ in q_tokens]
+        answer = transformed_sample["answer"]
```
ebsmothers, line 73: One very fundamental question here: why do we keep the answer as a string? This seems to break a pretty big assumption held by most of our data abstractions (and I assume this is why you've had to write so many custom components for data processing). I notice that you specify `self._tokenizer: ModelTokenizer` and call `self._tokenizer.encode()`, but I think this is conflating our `ModelTokenizer` and our `BaseTokenizer`. 

Our `BaseTokenizer`s have `encode` and `decode` methods defined on them, but in most cases these aren't really called directly. Instead we try to rely pretty heavily on our `ModelTokenizer`s, which typically wrap a `BaseTokenizer` with some additional logic for custom tokens, prompt templating, etc. These operate on lists of [Messages](https://github.com/pytorch/torchtune/blob/fb5255725e571e40147a32b939db32b6b3c79dad/torchtune/data/_messages.py#L22), which contain information about role, content, etc. So I am wondering if there's a world where e.g. `BASE_PROMPT`, `PREAMBLE_PROMPT`, `TRAINABLE_PROMPT`, become [PromptTemplates](https://github.com/pytorch/torchtune/blob/fb5255725e571e40147a32b939db32b6b3c79dad/torchtune/data/_prompt_templates.py#L16) and `ReasoningProblem` is covered by `Message` roles. Then we can define the right version of `tokenize_messages` to handle this (though admittedly this may require a little finagling).

Anyways it's possible this is all deliberate and for a good reason that I'm missing on a first pass, but want to make sure that (a) you're aware of the purposes of these different abstractions (cause admittedly it can be a bit unclear at first), and (b) we don't reinvent the wheel any more than we have to.

ebsmothers (commented):
Thanks for the PR! I'm really excited to see this (and I know the community is too). I left a lot of comments, but this is just a first pass. Admittedly I need to come back to the losses and core training logic, which I will do soon. Please let me know if any of the comments so far are unclear (some may require my reading the rest of the code, so please bear with me until then)

recipes/configs/dev/grpo/3B_full_rl.yaml
```diff
@@ -0,0 +1,139 @@
...
+grpo_samples: 16
+forward_batch_size: 1
+max_generated_tokens: 512
+top_k: null
+temperature: 1.0
+
+fsdp_reshard_after_forward: True
+
+ppo_epochs: 1
+ppo_batch_size: ${grpo_samples}  # single-step for the "simple" loss
```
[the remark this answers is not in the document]
RedTachyon: It probably makes sense to rename it, but it will be an awkward area of weird overlaps in naming and active research.

As an algorithm, GRPO is basically the same as PPO, but with a different advantage estimation method. Or at least that's how it's defined in the R1 paper. 
However, the implementation that's validated to run right now doesn't follow that math super closely, and instead follows the interpretation introduced in TRL, which has their own justification for that interpretation based on Deepseek's paper.

The general idea in PPO is that we can take multiple gradient steps with a single batch of data - `ppo_epochs` indicates how many times we go through the batch, and `ppo_batch_size` is a minibatch size inside of this (what we actually process at once).

The implementation that is currently running simplifies a few elements of the PPO/GRPO loss, which also requires only taking a single gradient step per data batch. This is what works in TRL, and what I validated to work here in torchtune. It's very likely that some other variations will also work, but I've yet to run those experiments.

recipes/configs/dev/grpo/3B_full_rl.yaml
```diff
@@ -0,0 +1,139 @@
...
+
+# Fine-tuning arguments
+batch_size: 1
+grpo_samples: 16
+forward_batch_size: 1
+max_generated_tokens: 512
+top_k: null
+temperature: 1.0
+
+fsdp_reshard_after_forward: True
```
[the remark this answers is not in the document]
RedTachyon: Probably not necessary - I think my process was that I initially set it to False, which sped up the generation (by a decent margin too, like 25% in the data generation phase), but then it caused some irregular memory issues. I didn't want to deal with the memory profiling just yet, so I set it back to True to see if the pipeline even works. 

In any case, this will likely be an impactful parameter to check when optimizing the recipe for performance, and I'm guessing it's not very commonly used in other recipes, since here we're doing an unusually large amount of generation.

recipes/configs/llama3_2/3B_full_rl_single_device_mps.yaml
```diff
@@ -0,0 +1,110 @@
...
+  lr: 2e-5
+loss:
+  _component_: torchtune.modules.loss.CEWithChunkedOutputLoss
+max_steps_per_epoch: null
+gradient_accumulation_steps: 1  # Use to increase virtual batch size
+optimizer_in_bwd: True  # True saves memory. Requires gradient_accumulation_steps=1
+compile: False  # pytorch compile, set to true for better perf/memory
+
+# Training environment
+device: mps
```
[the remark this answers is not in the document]
RedTachyon: Somehow completely missed this file when I was cleaning up the code - this was an old prototype on single-device MPS, which I didn't end up developing any further, so I deleted the whole config

recipes/configs/dev/grpo/3B_full_rl.yaml
```diff
@@ -0,0 +1,139 @@
...
+# to override the checkpointer directory while launching training
+# you can run:
+#   tune run --nproc_per_node 4 dev/grpo_full_finetune_distributed --config dev/grpo/3B_full_rl checkpointer.checkpoint_dir=<YOUR_CHECKPOINT_DIR>
+#
+# This config works best when the model is being fine-tuned on 2+ GPUs.
+#
+# Furthermore, you can launch it on multiple nodes by going to recipes/dev/ and using
+#  sbatch multinode_grpo.sbatch
+
+base_dir: /tmp
```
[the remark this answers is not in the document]
RedTachyon: So my personal preference is actually keeping base_dir or something similar, but for the "official" codebase it probably makes sense to keep it more uniform.

My reasoning is that - putting everything in /tmp/ isn't a terrible assumption with no information about the user's filesystem, but at the same time, most of the time, I don't want to put everything in /tmp. So it's much easier for me to change this in one place instead of tracking down all /tmp instances in the config.

recipes/dev/grpo_full_finetune_distributed.py
```diff
@@ -0,0 +1,1326 @@
...
+        self._ppo_epochs = cfg.ppo_epochs
+        self._ppo_batch_size = cfg.ppo_batch_size
+        self._gradient_accumulation_steps = cfg.gradient_accumulation_steps
+        self._ppo_backward_batch_size = (
+            cfg.ppo_batch_size // self._gradient_accumulation_steps
+        )
+
+        self._save_every_n_epochs = cfg.save_every_n_epochs
+
+        self._total_steps = cfg.num_steps // self.batch_size
```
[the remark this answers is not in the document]
RedTachyon: Also agreed it would probably be better to just store the config. The recipe was based on an amalgamation of single-device PPO and multi-device SFT, which (I think?) both don't do `self.cfg = cfg`, so I assumed there might be some reason for that.

recipes/dev/grpo_full_finetune_distributed.py
```diff
@@ -0,0 +1,1326 @@
...
+            f"Instantiating model and loading checkpoint took {time.perf_counter() - init_start:.2f} secs",
+        )
+        if self._is_rank_zero:
+            memory_stats = training.get_memory_stats(device=self._device)
+            training.log_memory_stats(memory_stats)
+
+        # synchronize before training begins
+        torch.distributed.barrier()
+
+        ref_model = None
```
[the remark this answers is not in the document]
RedTachyon: This is another "open research question", or at least "we need to figure out what's the best way to do things".

The formula listed in the R1 paper has a KL divergence term with reference policy logits, but a simplified implementation that I partly based this on, omits the reference logprobs.

Basically, to make this a fully proper and fleshed-out GRPO implementation, we probably want to bring back ref_model and use its logprobs, but this will require more experiments and more memory profiling. For the time being, we can get reasonable results by omitting it.

recipes/dev/grpo_full_finetune_distributed.py
```diff
@@ -0,0 +1,1326 @@
...
+            self._optim_ckpt_wrapper = training.create_optim_in_bwd_wrapper(
+                model=self._model, optim_dict=optim_dict
+            )
+            # Load optimizer states for each param. If optimizer states are being restored in an optimizer in
+            # backward run, these need to have been saved with the same setting. Cannot restore from runs that
+            # did not use optimizer in backward.
+            if opt_state_dict is not None:
+                for param in opt_state_dict.keys():
+                    try:
+                        training.load_from_full_optimizer_state_dict(
```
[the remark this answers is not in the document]
RedTachyon: I'm guessing something changed about the API between the time I "forked" the recipe and now. Updated since the signature wasn't matching.

recipes/dev/gsm8k_sft.sbatch
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
RedTachyon: This recipe is only potentially useful to simplify experiments on the preformance of GRPO. In my experience so far (aligned with other reports), training GRPO immediately from the base model is doable, but not super efficient - instead, doing a small step of SFT beforehand can make it run much better. This script is for that quick SFT.

torchtune/data/_collate.py
```diff
@@ -216,6 +216,48 @@ def padded_collate_sft(
     return {"tokens": input_ids.long(), "labels": labels.long()}
 
 
+def padded_collate_rl(
```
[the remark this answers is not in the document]
RedTachyon: The general idea is that the tokens can be collated "normally", but ground truth answers should stay in "string space". 

Inside of the training loop, we need some way to score a completion, in a component that is largely independent from the main data loader. The simplest way is decoding the completion, parsing it for <answer>XYZ</answer>, and then checking for a perfect match with the ground truth answer. This naturally makes more sense keeping it as a string, instead of tokenizing and detokenizing back, when the "answer" tokens are never used themselves.

Notably, this is different from reward model-based approaches (e.g. RLHF), where answer tokens are relevant and go into that reward model. Here, the reward is a "normal" function, and those are easier to write for strings than for tokens.

recipes/dev/multinode_grpo.sbatch
```diff
@@ -0,0 +1,48 @@
...
+# You probably want to load in a virtual env w/ conda...
+# module load conda
+# conda activate torchtune
+# ...or venv
+# source torchtune/bin/activate
+
+source ../../.venv/bin/activate
+
+# Adjust sbatch --ntasks and sbatch --nodes above and --nnodes below to your specific node count
+srun --export=ALL,OMP_NUM_THREADS=8 tune run --nnodes 16 --nproc_per_node 8 --rdzv_id 101 --rdzv_backend c10d --rdzv_endpoint "$head_node_ip:29500" \
```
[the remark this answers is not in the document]
RedTachyon: I don't have any detailed profiling on this. One thing I noticed is that if I don't set it, there's a bunch of messages in the logs that this gets automatically set to 1, which by my understanding is not optimal. Not sure on the practical impact though.

recipes/dev/multinode_grpo.sbatch
```diff
@@ -0,0 +1,48 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
RedTachyon: I think the main value in these "redundant" files is ease of development - this is a setup that's validated to work, so if you want to e.g. run it yourself to debug/improve some part of the pipeline, this is a good starting point.

My guess is it might be fine while in dev/, and then it could get removed before being moved into the "core" part of the repo

recipes/dev/grpo_full_finetune_distributed.py
```diff
@@ -0,0 +1,1326 @@
...
+        RuntimeError: If ``enable_activation_offloading`` is True and device is not CUDA.
+        RuntimeError: If ``enable_activation_offloading`` is True and ``enable_activation_checkpointing`` is False.
+    """
+
+    def __init__(self, cfg: DictConfig) -> None:
+        self._device = utils.get_device(device=cfg.device)
+        self._dtype = training.get_dtype(cfg.dtype, device=self._device)
+        device_type = cfg.device
+
+        # if self._dtype == torch.float16:
```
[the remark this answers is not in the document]
RedTachyon: Early on in the development, I was fighting some memory issues (and some older hardware), so I tested all of fp16/32 and bf16. Fp16 pretty much fails completely, so it probably makes sense to readd this check

recipes/dev/grpo_full_finetune_distributed.py
```diff
@@ -0,0 +1,1326 @@
...
+                max_generated_tokens=self._max_generated_tokens,
+                temperature=self._temperature,
+                top_k=self._top_k,
+                pad_id=self._tokenizer.pad_id,
+                rng=self._rng,
+                return_logits=False,
+                stop_tokens=self._tokenizer.stop_tokens,
+            )
+
+        training.recursive_reshard(self._model)
```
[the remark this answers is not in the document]
RedTachyon: I think this will need to get some more attention when profiling/optimizing the recipe - in the current form, indeed this is redundant. It's possible that setting `reshard_after_forward=False` and keeping this manual reshard would make the data collection faster, without increasing the top memory spike, but I'm not sure at the moment.

torchtune/generation/_generation.py
```diff
@@ -377,19 +383,29 @@ def generate(
...
+                if all_done == world_size:
+                    # Multiple devices
+                    break
 
     # mask out generated tokens in seqs that already hit a stop token
     if stop_tokens is not None:
         generated_tokens *= stop_token_mask
-        generated_logits *= stop_token_mask[:, -generated_logits.shape[1] :, None]
+        if return_logits:
+            generated_logits *= stop_token_mask[:, : -generated_logits.shape[1], None]
```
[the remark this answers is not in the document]
RedTachyon: Oof, good catch, probably some mixup with code changed in the meantime; removed this change

torchtune/datasets/_rl.py
```diff
@@ -0,0 +1,169 @@
...
+
+def extract_tags(text: str) -> dict[str, list[str]]:
+    """
+    Parse XML-like tags from text. Returns a dictionary with keys 'think' and 'answer'.
+    The values are lists of strings, with each string being the content of a tag.
+    """
+    xml_string = f"<root>{text}</root>"
+    root = ET.fromstring(xml_string)
+
+    return {
```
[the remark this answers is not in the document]
RedTachyon: The template is based on the prompt from Deepseek-R1-Zero (https://arxiv.org/pdf/2501.12948, page 6, Table 1).

Parsing is just custom logic to parse XML. This isn't perfect, but if the model follows the general template of <think>CoT</think> <answer>XYZ</answer>, it works fine.

torchtune/datasets/_rl.py
```diff
@@ -0,0 +1,169 @@
...
+    if len(tags["think"]) == 1:
+        reward += 5.0
+
+    if any(attempt == answer for attempt in tags["answer"]):
+        # One of the answer tags has the right answer
+        reward += 20.0
+
+    if any((answer in attempt) for attempt in tags["answer"]):
+        # One of the answer tags contains the right answer (might be e.g. $20 instead of 20)
+        reward += 10.0
```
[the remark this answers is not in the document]
RedTachyon: Rewards are mostly shaped intuitively to ease the training a little bit, especially starting from a base model. The platonic form of the reward function would be binary, either it matches or it doesn't. However, this will be very painful when trying to train a model that's not good at following the template. Instead, providing a bit of a reward is helpful to guide it towards using the right format.

torchtune/datasets/_rl.py
ebsmothers: Thanks for the explanation, this helps a lot. I just want to push on one more detail here to make sure I fully understand. 

You mention that doing everything in token space isn't a good option. Why is that? As far as I can see, it shouldn't matter whether we are in string space or in token space, especially if we are just checking equality. Say that instead of doing XML parsing using `<think>` and `<answer>` tags, we are just splitting on the tokens that these correspond to. (Aside: do these correspond to special tokens in the vocabulary? If so I think it's easier, but from my initial search I couldn't confirm this.) I think string equality < - > token equality is pretty easy if wrapped in special tokens, i.e. `s1 == s2` if and only if `tokenizer.encode(s1) == tokenizer.encode(s2)`. The subset reward does seem a bit trickier, but again I think `s1` $\subset$ `s2` if and only if `tokenizer.encode(s1)` $\subset$ `tokenizer.encode(s2)` (modulo some off-by-one error, but this shouldn't be relevant if the reward doesn't actually depend on the length of the match).

Admittedly I am thinking about this from an engineering perspective quite a bit here -- if we can standardize formats it will make it much easier to generalize and reuse some of these things. 

Anyways, let me know if this makes sense. If not, totally understandable -- we can figure out a way to make the string approach work (may just need more refactoring on our end further down the line).

torchtune/_recipe_registry.py
```diff
@@ -98,6 +106,7 @@ class Recipe:
         name="full_finetune_distributed",
         file_path="full_finetune_distributed.py",
         configs=[
+            Config(name="dev/grpo/3B_sft", file_path="dev/grpo/3B_sft.yaml"),
```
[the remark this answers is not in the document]
felipemello1: oh, i missed that its _sft.

torchtune/datasets/_rl.py
felipemello1: @ebsmothers, because LLMs are probabilistic, isnt there a chance that 2 texts are encoding differently? e.g. say we have the tokens [a, b, c, ab, bc, abc]. I the model wants to output "abc", it actually has many ways of doing it. For the first token, it may have a high probability to "a" and "ab" and "abc".

Doing it in the string space seems safer, but maybe i am misunderstanding.

torchtune/rlhf/loss/grpo.py
```diff
@@ -0,0 +1,272 @@
...
+from torchtune.rlhf import masked_sum
+
+
+class GRPOLoss(nn.Module):
+    """
+    Group Relative Policy Optimization (GRPO) Loss module.
+    Introduced by https://arxiv.org/abs/2402.03300, popularized by https://arxiv.org/abs/2501.12948.
+
+    This loss implementation follows the usual formulation of GRPO with clipped ratios of token-wise logprobs.
+    Currently not validated to perform well.
```
SalmanMohammadi, line 21: What needs to be validated?

torchtune/rlhf/loss/grpo.py
```diff
@@ -0,0 +1,272 @@
...
+            policy_loss.detach(),
+            kl_loss.detach(),
+            ratios.mean().detach(),
+            clipfrac.detach(),
+        )
+
+
+class GRPOCompletionLoss(nn.Module):
+    """
+    Group Relative Policy Optimization (GRPO) Loss module.
```
SalmanMohammadi, line 103: Could we make it slightly clearer from the docstring why we have this loss in addition to `GRPOLoss`?

torchtune/rlhf/sequence_processing.py
```diff
@@ -101,6 +101,42 @@ def logits_to_logprobs(
     ).squeeze(-1)
 
 
+def batched_logits_to_logprobs(
```
SalmanMohammadi, line 104: I like this

torchtune/generation/_generation.py
```diff
@@ -377,19 +383,29 @@ def generate(
...
+            if world_size == 1:
+                # Single device
+                if stop_token_reached.all():
+                    break
+            else:
+                all_done = stop_token_reached.all().int()
+                torch.distributed.all_reduce(all_done)
+                if all_done == world_size:
+                    # Multiple devices
+                    break
```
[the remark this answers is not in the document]
SalmanMohammadi: Yeah this looks a bit weird sitting here to me, but I'm not immediately sure what the best way to do this is.

recipes/dev/grpo_full_finetune_distributed.py
```diff
@@ -0,0 +1,1326 @@
...
+        self.seed = training.set_seed(seed=cfg.seed)
+        self.total_epochs = cfg.epochs
+        self.max_steps_per_epoch = cfg.max_steps_per_epoch
+        self.global_step = 0
+        self._steps_run = 0
+        self._total_steps = 0
+        self._epochs_run = 0
+        self._rng = torch.Generator(self._device).manual_seed(self.seed)
+
+    def load_checkpoint(self, cfg_checkpointer: DictConfig) -> Dict[str, Any]:
```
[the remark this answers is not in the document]
SalmanMohammadi: `CheckpointClient` also needs to be slightly reworked. When we load a reference model we won't have `self._cfg.checkpointer,` [here](https://github.com/pytorch/torchtune/blob/8c9235e2b10cc9e1aa381aaec4b992dbf52c48da/torchtune/training/checkpointing/_checkpoint_client.py#L89), but rather `self._cfg.ref_checkpointer`.

recipes/dev/grpo_full_finetune_distributed.py
```diff
@@ -0,0 +1,1289 @@
...
+        self._compile = cfg.get("compile", False)
+        self._model, self._ref_model = self._setup_model(
+            cfg_model=cfg.model,
+            enable_activation_checkpointing=self._enable_activation_checkpointing,
+            enable_activation_offloading=self._enable_activation_offloading,
+            custom_sharded_layers=cfg.get("custom_sharded_layers", None),
+            fsdp_cpu_offload=self.fsdp_cpu_offload,
+            reshard_after_forward=cfg.get("fsdp_reshard_after_forward", True),
+            model_state_dict=checkpoint_dict[training.MODEL_KEY],
+            ac_mode=cfg.get("ac_mode", None),
```
SalmanMohammadi, line 287: Did you try activation checkpointing here? When I last tried to implement it in the PPO recipe it didn't work, as we needed to disable the activation checkpointing hooks during generation - I wrote a utility to do this but didn't really see much performance gains anyway, so I dropped it.

recipes/dev/grpo_full_finetune_distributed.py
```diff
@@ -0,0 +1,1326 @@
...
+                model,
+                ac_mode,
+                ac_option,
+            )
+
+            # apply_selective_activation_checkpointing(
+            #     ref_model,
+            #     ac_mode,
+            #     ac_option
+            # )
```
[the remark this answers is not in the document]
SalmanMohammadi: I also had a comment above about generally using activation checkpointing in this recipe.

recipes/dev/grpo_full_finetune_distributed.py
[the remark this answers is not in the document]
SalmanMohammadi: In this case I think we should remove the references and comments here and make it clear that the recipe doesn't support a reference model, and that we plan to add this in.

recipes/dev/grpo_full_finetune_distributed.py
```diff
@@ -0,0 +1,1289 @@
...
+        self._is_rank_zero = rank == 0
+
+        # Training cfg
+        self._resume_from_checkpoint = cfg.resume_from_checkpoint
+        self._gradient_accumulation_steps = cfg.gradient_accumulation_steps
+        self._optimizer_in_bwd = cfg.get("optimizer_in_bwd", False)
+        self._clip_grad_norm = cfg.get("clip_grad_norm", None)
+
+        # Optimizer in backward is not compatible with gradient accumulation or gradient clipping
+        if self._optimizer_in_bwd:
```
SalmanMohammadi, line 158: Shall we just remove this from the recipe? cc @ebsmothers

recipes/dev/grpo_full_finetune_distributed.py
```diff
@@ -0,0 +1,1289 @@
...
+        #     strict=True,
+        #     cpu_offload=fsdp_cpu_offload,
+        # )
+
+        # activation offloading
+        self.policy_activations_handling_ctx = training.get_act_offloading_ctx_manager(
+            model, enable_activation_offloading
+        )
+
+        # self.ref_activations_handling_ctx = training.get_act_offloading_ctx_manager(
```
SalmanMohammadi, line 622: side note - we wouldn't need to offload/checkpoint any activations for the reference model since it's not being trained

recipes/dev/grpo_full_finetune_distributed.py
```diff
@@ -0,0 +1,1289 @@
...
+            )
+            log.info(f"Saving checkpoint took {time.perf_counter() - start:.2f} secs")
+
+        torch.distributed.barrier()
+
+    def generate_trajectory(
+        self, input_ids: torch.Tensor, answers: List[str]
+    ) -> GRPOTrajectory:
+        """
+        Generates a trajectory given the current policy model, the reference policy model, the reward model,
```
SalmanMohammadi, line 843:
```suggestion
        Generates a trajectory given the current policy model, the reward function(s)
```
We don't support a reward model or reference model right?

recipes/dev/grpo_full_finetune_distributed.py
```diff
@@ -0,0 +1,1289 @@
...
+                temperature=self._temperature,
+                top_k=self._top_k,
+                pad_id=self._tokenizer.pad_id,
+                rng=self._rng,
+                return_logits=False,
+                stop_tokens=self._tokenizer.stop_tokens,
+            )
+
+        training.recursive_reshard(self._model)
+        torch.cuda.empty_cache()
```
SalmanMohammadi, line 890: Did you find this helped?

torchtune/datasets/_rl.py
```diff
@@ -0,0 +1,169 @@
...
+            elem.text if elem.text is not None else "" for elem in root.findall("think")
+        ],
+        "answer": [
+            elem.text if elem.text is not None else ""
+            for elem in root.findall("answer")
+        ],
+    }
+
+
+def shaped_correctness_reward(
```
SalmanMohammadi, line 97: nit: I think I would rather see these reward functions in the `rlhf` module - maybe as `grpo_rewards`, and then import them under the `rlhf` namespace.

i.e. add these to `grpo_rewards.py` in `rlhf/`, import them in `rlhf.__init__`, and in the recipe use `rlhf.shaped_correctness_reward`.

torchtune/datasets/_rl.py
```diff
@@ -0,0 +1,169 @@
...
+    question: str, answer: str, completion: str
+) -> tuple[float, float]:
+    """
+    Reward function for verifiable rewards with some mild shaping.
```
SalmanMohammadi, line 101: Let's add docstrings here : )

torchtune/datasets/_rl.py
```diff
@@ -0,0 +1,169 @@
...
+
+    if any((answer in attempt) for attempt in tags["answer"]):
+        # One of the answer tags contains the right answer (might be e.g. $20 instead of 20)
+        reward += 10.0
+
+    # total_think_length = sum(len(c) + len("<think></think>") for c in tags["think"])
+    # total_answer_length = sum(len(a) + len("<answer></answer>") for a in tags["answer"])
+    # total_extra_length = len(only_completion) - total_think_length - total_answer_length
+
+    # reward -= 0.1 * total_extra_length
```
SalmanMohammadi, line 134:
```suggestion
```
Can this be removed?

torchtune/datasets/_rl.py
```diff
@@ -0,0 +1,169 @@
...
+
+    if len(tags["answer"]) > 0 and tags["answer"][-1] == answer:
+        reward = 100.0
+        success = 1
+
+    _, rank = utils.get_world_size_and_rank()
+    if rank == 0:
+        print()
+        print(f"{question=}\n{answer=}\n{reward=}\n{only_completion=}")
+        print()
```
SalmanMohammadi, line 144: also
```suggestion
```

torchtune/datasets/_rl.py
```diff
@@ -0,0 +1,169 @@
...
+    rewards = torch.zeros(batch_size, grpo_size, dtype=torch.float32)
+    successes = torch.zeros(batch_size, grpo_size, dtype=torch.float32)
+    # completions :: [B, G, L]
+    for b in range(batch_size):
+        for g in range(grpo_size):
+            text_completion = tokenizer.decode(
+                completions[b, g].tolist()
+            )  # skips special tokens, stops at eos
+            reward, success = shaped_correctness_reward(
+                question="", answer=answers[b], completion=text_completion
```
SalmanMohammadi, line 164: Sorry if I'm missing something here, but why is `question=""`? At first glance it's not clear.

recipes/dev/grpo_full_finetune_distributed.py
```diff
@@ -0,0 +1,1289 @@
...
+        masks = generation.get_causal_mask_from_padding_mask(
+            query_response_padding_masks
+        )
+        position_ids = generation.get_position_ids_from_padding_mask(
+            query_response_padding_masks
+        )
+
+        del query_response_padding_masks
+
+        logits = self._model(query_responses, input_pos=position_ids, mask=masks)
```
SalmanMohammadi, line 907: How come you're recalculating the logits?

recipes/dev/grpo_full_finetune_distributed.py
```diff
@@ -0,0 +1,1289 @@
...
+        trajectories: List[GRPOTrajectory] = []
+        with torch.no_grad():
+            for batch_start in range(0, self.batch_size, self._forward_batch_size):
+                batch_input_ids = input_ids[
+                    batch_start : batch_start + self._forward_batch_size
+                ]
+                batch_answers = answers[
+                    batch_start : batch_start + self._forward_batch_size
+                ]
+                torch.cuda.empty_cache()
```
SalmanMohammadi, line 1000: I'm seeing a lot of these. When I profiled the PPO recipe (which this is based on) I didn't find I needed it, did you find something specific that meant this helped?

recipes/dev/grpo_full_finetune_distributed.py
```diff
@@ -0,0 +1,1289 @@
...
+        return GRPOTrajectory(*map(torch.cat, zip(*trajectories)))
+
+    def _grpo_step(
+        self,
+        trajectory: GRPOTrajectory,
+        context_length: int,
+    ) -> GRPOStats:
+        """
+        Perform a single GRPO optimisation step over a batch of trajectories and corresponding advantages and returns.
```
SalmanMohammadi, line 1013:
```suggestion
        Perform a single GRPO optimization step over a batch of trajectories and corresponding advantages and returns.
```
americans -_-

recipes/dev/grpo_full_finetune_distributed.py
```diff
@@ -0,0 +1,1289 @@
...
+
+        Args:
+            trajectory (Trajectory): a batch of trajectories
+            context_length (int): input ids sequence length
+
+        Returns:
+            GRPOStats: An instance of :class:`~torchtune.rlhf.PPOStats`, a NamedTuple containing:
```
SalmanMohammadi, line 1020:
```suggestion
            GRPOStats: An instance of :class:`~torchtune.rlhf.GRPOStats`, a NamedTuple containing:
```

recipes/dev/grpo_full_finetune_distributed.py
```diff
@@ -0,0 +1,1289 @@
...
+                                    partial(
+                                        torch.index_select,
+                                        dim=0,
+                                        index=backward_batch_idxs,
+                                    ),
+                                    trajectory,
+                                )
+                            )
+                            batch_grpo_stats.append(
+                                self._grpo_step(  # .backward() happens here
```
SalmanMohammadi, line 1160: nit, we don't need to make these private
```suggestion
                                self.grpo_step(  # .backward() happens here
```

recipes/dev/grpo_full_finetune_distributed.py
```diff
@@ -0,0 +1,1289 @@
...
+
+                _, context_length = tokens.shape
+
+                trajectory = self.generate_trajectory_batched(tokens, answers)
+
+                # Calculate the number of unmasked tokens in the current batch
+                # and increment the total number of tokens seen in the step
+
+                effective_batch_size = self.batch_size * self.grpo_samples  # = B x G
+                # torch.distributed.breakpoint()
```
SalmanMohammadi, line 1133:
```suggestion
```

recipes/dev/grpo_full_finetune_distributed.py
```diff
@@ -0,0 +1,1289 @@
...
+                answers = batch["answers"]  # type: ignore
+                tokens = tokens.to(self._device)  # [B, P]
+
+                _, context_length = tokens.shape
+
+                trajectory = self.generate_trajectory_batched(tokens, answers)
+
+                # Calculate the number of unmasked tokens in the current batch
+                # and increment the total number of tokens seen in the step
+
```
SalmanMohammadi, line 1131: Are we doing this here?

SalmanMohammadi:
Thanks for your updates @RedTachyon. I've left a bunch of inline comments as a first pass. I'll collect some of my high level thoughts here.

Regarding the reference model - my suggestion here would be:

- Remove commented out/hacks in the recipe for using a reference model (e.g. logprobs.clone). Let's be explicit that this recipe doesn't support reference models at a high level, and link to the reference implementation you use that follows this design.

I also notice there's multiple losses you've added, and the one you've tested is GRPOSimpleLoss - is that right? If this is the case, I'd suggest removing the KL logic from this loss and updating the docstring to indicate that it doesn't use a reference model. If the other losses haven't been validated let's just remove them for now.

If this follows the TRL implementation I'm slightly confused w.r.t. the usage of "ppo_epochs" - I haven't tried implementing this myself so I'd appreciate your thoughts here.

recipes/dev/grpo_full_finetune_distributed.py
ebsmothers: Yeah I would be in favor of this

RedTachyon:
So one meta-level question when it comes to deciding on the version that will be merged into /dev - how should we handle things that theoretically make sense, but might require some more experimental validation?

This applies to a few different things that were brought up in discussion, off the top of my mind:
- ppo_epochs is set to 1, but GRPOLoss/GRPOCompletionLoss could plausibly work with ppo_epochs>1
- as mentioned above, GRPOSimpleLoss works, and the other two losses plausibly should also work, but I don't have results for that yet
- I just ran a few experiments with ref_policy reenabled, and it actually still works, though not as well in one random experiment I did

Basically, there's a whole bunch of settings that make sense theoretically, but there is still ongoing "research" to check whether they make sense. This is what I mean when I write that something is not validated to work - it should run without crashing, but at this moment, I can't guarantee that it won't tank your performance (either due to a hidden bug, or due to the natural difficulties of RL/research).

My thinking is that it makes sense that while it's in /dev, we try to support everything that could plausibly work. Then various people will keep experimenting with different configurations (including me right now), which eventually we can distill into a "proper" form of the recipe that works well.

---

My overall message here is that this style of training is very much an active area of research - it's not as neat and tidy as SFT, but rather we have one main inspiring paper (R1) with a little bit of detail, a bunch of reimplementations (this, trl, some other libraries), and various independent experimental results. We probably want something solid that sorta works, but then the details will be ironed out by the wider community over the upcoming weeks/months.

recipes/dev/grpo_full_finetune_distributed.py
RedTachyon: Do you mean remove grad norm clipping, or optimizer_in_bwd?

SalmanMohammadi: `optimizer_in_bwd`

RedTachyon: (I'll interpret this as removing optimizer_in_bwd, because clip_grad_norm was useful in some of my experiments)

SalmanMohammadi:
> My overall message here is that this style of training is very much an active area of research - it's not as neat and tidy as SFT, but rather we have one main inspiring paper (R1) with a little bit of detail, a bunch of reimplementations (this, trl, some other libraries), and various independent experimental results. We probably want something solid that sorta works, but then the details will be ironed out by the wider community over the upcoming weeks/months.

My view here (I'll let the other maintainers chip in) is that we must commit to a level of reasonable support for every feature we land in torchtune. Partly, this means onboarding features which are tried-and-tested and which have reference implementations in order to provide a stable, hackable, it-just-works baseline for our users. Note: we haven't always been consistent about this (e.g. we regularly onboard nightly features from core). Removing features which are broken/untested are a less-than-ideal scenario for us, which means we exercise a relatively high level of caution when considering more research-y features.

In this case I think we'd definitely like to land your work, and land it soon. For this PR, I'd be hesitant to onboard features which you're not quite sure work OOTB right now empirically, and/or which lack precedent.

recipes/dev/grpo_full_finetune_distributed.py
```diff
@@ -0,0 +1,1289 @@
...
+        # ref_logits = self._ref_model(
+        #     query_responses, input_pos=position_ids, mask=masks
+        # )
+        # ref_logits = rlhf.truncate_sequence_for_logprobs(ref_logits, context_length)
+        # ref_logprobs = rlhf.batched_logits_to_logprobs(ref_logits, responses, self._temperature)
+
+        # del ref_logits
+        # torch.cuda.empty_cache()
+
+        ref_logprobs = logprobs.clone()
```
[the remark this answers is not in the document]
RedTachyon: Note: I reenabled the reference policy logprobs, and checked that it also learns. On n=1, the performance is worse, but it will take much more experiments (and some more hyperparameter search) to really check that.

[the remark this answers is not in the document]
SalmanMohammadi: Performance is worse in what regard?

recipes/dev/grpo_full_finetune_distributed.py
RedTachyon: I've been hitting various OOMs, so I tried to be a bit defensive and just make sure it runs with excessive cache emptying. Some of them can probably be deleted, but that will likely best be done during general memory profiling/optimization of the recipe

recipes/dev/grpo_full_finetune_distributed.py
RedTachyon: This was copied verbatim from the corresponding PPO recipe, just sayin

recipes/dev/grpo_full_finetune_distributed.py
RedTachyon: Not necessarily with 3B, initially I was experimenting with 8B, and also at some point I realized that slurm was sometimes giving me 16GB VRAM GPUs. So *probably* most of these are redundant, but I don't want to ship it and then find out that this specific one wasn't.

RedTachyon:
Update: yesterday I removed a bunch of dead features, fixed some bugs, ran it again (now properly with the reference policy and KL loss), and it all works nicely, steadily reaching the 50-60% success rates like before. 

I think the main thing to clean up today is how steps are being counted, and maybe sub-epoch saving, basically "Counting how many times we did something, and making decisions based on that", after that this miiiight be ready?

RedTachyon:
I made some new changes, notably:
- Removed some unused settings like gradient accumulation and minibatches within PPO/GRPO epochs. 
- Cleaned up iteration counting; checkpointing is still done by epochs, which will be pretty infrequent, but I didn't want to get into rewinding 
- Reorganized recipes and configs out of their own /dev/grpo directory into just /dev

Some general notes:
- I'm keeping the SFT recipe (for now?), mainly for development simplicity. Llama base models (from what I've seen) don't learn well from pure RL, so if someone wants to contribute and/or run their own experiments, the SFT recipe is an extremely helpful quick step to do before running the RL loop
- I simplified some of the step/epoch management by removing most of the sub-batching in backward etc. The only minibatching happening right now is the one in data generation. The current recipe is running nicely on a node with 8x32GB GPUs, so a more modern card with like 80GB should handle all of this easily.


Right now I'm running another batch of experiments to make sure it still learns well (as of right now, I can confirm that it at least doesn't crash), I should have results this evening/tomorrow morning. Please let me know if there's still something important I missed, and many thanks for helping bring this to a good state!

recipes/dev/grpo_full_finetune_distributed.py
```diff
@@ -0,0 +1,1099 @@
...
+            apply_selective_activation_checkpointing(
+                model,
+                ac_mode,
+                ac_option,
+            )
+
+            apply_selective_activation_checkpointing(ref_model, ac_mode, ac_option)
+
+        # original activation checkpointing (full) - flip the condition above
+        if enable_activation_checkpointing and ac_mode is None:
```
SalmanMohammadi, line 477:
```suggestion
        if enable_activation_checkpointing:
```
@RedTachyon let's remove all references to `ac_mode` from this file and the configs too, please.

recipes/dev/grpo_full_finetune_distributed.py
```diff
@@ -0,0 +1,1099 @@
...
+            shard_conditions=fsdp_shard_conditions,
+            cpu_offload=fsdp_cpu_offload,
+            reshard_after_forward=reshard_after_forward,
+        )
+
+        training.shard_model(
+            model=ref_model,
+            shard_conditions=fsdp_shard_conditions,
+            cpu_offload=fsdp_cpu_offload,
+            reshard_after_forward=reshard_after_forward,
```
SalmanMohammadi, line 504:
```suggestion
            reshard_after_forward=True
```
cc @ebsmothers since there's no backwards pass happening here the gathered parameters during the forward pass would never be freed, right?

recipes/dev/grpo_full_finetune_distributed.py
SalmanMohammadi: bump @RedTachyon

recipes/dev/grpo_full_finetune_distributed.py
```diff
@@ -0,0 +1,1099 @@
...
+        training.validate_no_params_on_meta_device(ref_model)
+
+        utils.log_rank_zero(
+            log,
+            f"Instantiating model and loading checkpoint took {time.perf_counter() - init_start:.2f} secs",
+        )
+        if self._is_rank_zero:
+            memory_stats = training.get_memory_stats(device=self._device)
+            training.log_memory_stats(memory_stats)
+
```
SalmanMohammadi, line 546: I think you'll need to add the `disable_drouput` calls here @RedTachyon

torchtune/training/_distributed.py
```diff
@@ -568,6 +568,17 @@ def shard_model(
...
+    """
+    Manually reshard all modules in the model.
+    This might be useful for memory management when a model isn't automatically resharded after forward.
+    """
+    for n, m in reversed(list(module.named_modules())):
+        module.reshard()
+
+    module.reshard()
+
+
```
SalmanMohammadi, line 581: As per my comment above, we could just enable `reshard_after_forward` here for the reference model and remove this utility. cc @ebsmothers

RedTachyon: The intention for this was a bit different. For the policy model, `reshard_after_forward==False` significantly speeds up generation. After generation is done, there's a bunch of computation to be done (converting logits to logprobs etc.), and only after that do we reach a backward call, when the model would get resharded. Instead, I wanted to manually reshard the model to free up memory immediately after generation, so that we have more memory to spare for the intermediate computation.

SalmanMohammadi: Ah that makes more sense now, sorry. Unfortunately I think this is a nasty side effect of having to use FSDP for generation, as generation steps would effectively just use DDP. This is probably fine for a dev recipe until we have better infra in place.

SalmanMohammadi:
Thanks so much for all the updates and sticking with this @RedTachyon. Some last few outstanding comments, but this is looking pretty good.

I have two outstanding questions:
-  `reshard_after_forward` - I think the behaviour of this feature differs substantially from our other recipes and I would almost want to remove it from this recipe, since it's always `False` for the reference model, and the optimal strategy is to set it to `False` for the policy model and manually reshard.
- `generation` changes: this is a public API, and the changes primarily seem to stem from a line in the code where you're recalculating logits - I'm curious what motivated this decision?

RedTachyon:
1. To be sure, are you suggesting to simply keep it as False and remove it from the config? I think I'm fine with that - I don't really see a use-case for ever switching it to True since all it does is prevent immediate unshard-reshard cycles.
2. Yea, this is probably the most questionable thing. I initially did it during some memory debugging, but in the end it's most likely unnecessary. Now, on the other hand - I started (separately, not in this PR (yet?)) writing a proper evaluation routine for RL tasks, and for that, this change is actually pretty important. For evaluation, we only want to generate completions and don't need logits - and logits can take up a ton of VRAM. So I'm leaning towards keeping it, and will definitely be keeping it in my "personal use" fork, but if you think it's unnecessary/premature, I can remove it from the PR.

SalmanMohammadi:
> 1. To be sure, are you suggesting to simply keep it as False and remove it from the config? I think I'm fine with that - I don't really see a use-case for ever switching it to True since all it does is prevent immediate unshard-reshard cycles.

Yeah, this makes sense to me. Could you add a comment explaining why this is the case? Linking to my comment above on this PR is also fine.


>     2. Yea, this is probably the most questionable thing. I initially did it during some memory debugging, but in the end it's most likely unnecessary. Now, on the other hand - I started (separately, not in this PR (yet?)) writing a proper evaluation routine for RL tasks, and for that, this change is actually pretty important. For evaluation, we only want to generate completions and don't need logits - and logits can take up a ton of VRAM. So I'm leaning towards keeping it, and will definitely be keeping it in my "personal use" fork, but if you think it's unnecessary/premature, I can remove it from the PR.

If you wouldn't mind I think if we aren't using it at the moment, we should take it out. I think such a change would be better made if it was contextualized in e.g. a follow up to add evaluation functionality, rather than generalizing prematurely.

Thanks again for all your work here. 

P.S. I've started kicking CI off and it looks like some tests are failing which you might need to fix. You should be able to run the unit tests pretty painlessly by using `pytest tests` locally (see https://github.com/pytorch/torchtune/blob/main/CONTRIBUTING.md#testing).

torchtune/training/__init__.py
```diff
@@ -20,6 +20,7 @@
     load_from_full_model_state_dict,
     load_from_full_optimizer_state_dict,
     prepare_mha_for_tp,
+    recursive_reshard,
```
SalmanMohammadi, line 23: I don't think this should be a public API - cc @joecummings

RedTachyon:
@SalmanMohammadi Actually, I'm already observing something which I don't know if it makes sense or not, after the return_logits change - the data generation starts out fast, but slows down significantly towards the end of the batch (in this case, with maximum length of 512 tokens). In the next iteration, it again starts out fast and slows down over time. Think 5 (16x batched) tokens per second in the beginning, and 1.5 token per second by step 500.

To recap, before just about now, the data generation process completely discarded logits, simply recording the emitted tokens. This is basically free in terms of VRAM. After collecting a batch of data, we explicitly compute the logits for the whole batch of completions, compute logprobs etc.

Now, I reverted it to how generation was done "originally", that is - at each step, record the emitted token and the logits of all possible tokens in that slot, which is 128k floats. This uses more memory during generation, but saves us from having to recompute the logits again afterwards.


When I look at the generation logic, logits are collected using 
```python
generated_logits = torch.cat([generated_logits, logits], dim=1)
```

I'm far from an expert on low-level memory management stuff, but my barely educated guess is that this way, we consecutively allocate larger and larger arrays from scratch - first [16, 1, 128k], then [16, 2, 128k], ..., [16, 500, 128k]. Each time we want to add another logit, we have allocate a ton of memory ([16, 500, 128k] is 2GB of memory), which cannot be good for performance, and probably contributes to some fragmentation if that's even a thing.

When logits were disabled, this wasn't a problem - that part of the code was simply skipped.

As for a solution, the simple one is to allow disabling it again :). The correct solution is probably allocating the entire logits array beforehand and progressively filling it as we progress. 
This gets a bit more subtle with stop tokens, considering that we might not need to generate the full maximum length. The array probably needs to get initialized with some stop tokens, but even then we're allocating a bunch of extra memory, which might or might not be impactful during training (for example - we give the model up to 20k tokens for a long CoT, but early on in the training it only knows how to think for 200 tokens, we're allocating a ton of redundant memory).

So I think I looped all the way back around and I think it would be best to still allow `return_logits=False`? And to change the way logits are computed within the generation loop, but this can be either this or another PR.

SalmanMohammadi:
Nice sleuthing @RedTachyon! 

>The correct solution is probably allocating the entire logits array beforehand and progressively filling it as we progress. 

I think you're probably on the ball here. My general wariness about landing changes to our generation functionality in this PR is due to other parts of our codebase relying on this functionality - the testing and review necessary for these changes would warrant an entirely separate PR. Ideally I'd like to investigate this as a follow-up (also partially to make reviewing this PR slightly easier). How does that sound?

Overall you've highlighted several gaps in our inference infra with this PR which weren't on our radar, which is incredibly helpful - I'm sorry you had to partially guinea pig this.

RedTachyon:
So to be sure, either way we fix it up properly in a subsequent PR, but for the time being, which option do you prefer?
a) Keep generation as-is, i.e. no changes to the public API and existing user-facing functions, at the cost of a performance loss in the current GRPO (on average, generation takes 2x as much for me with this bug, making this easily the biggest bottleneck)
b) Add a relatively non-intrusive opt-in to disable recording logits

I'm obviously leaning b) because I hate to leave so much performance on the table. The biggest argument I see against is if we want to keep the public-facing API stable, i.e. we're not sure if we actually want to have a `return_logits` setting long-term. At the same time, we probably need some sort of a generation revamp either way? So it can't stay *that* stable

---

Side note - the algorithm is still learning with the other losses and ppo_epochs=2, though I can't make any definitive claims on how the performances compare - my initial experiments indicate that GRPOLoss works comparably well, and GRPOCompletionLoss is worse, but all of this could be down to the hyperparameters. All three variants have some level of theoretical justification, so I'm inclined to keep them and over time the community will figure out which approaches work best.

ebsmothers:
@RedTachyon, @SalmanMohammadi, just to weigh in on the conversation of generation APIs: I think this is just part of a broader comment I have about these changes. Personally I was pushing on it in the data abstractions (and thanks for the answers there, the design makes a lot more sense now), but e.g. we do need to be careful about exposing `RLDataset` as a core abstraction before we are sure how well it will generalize (not to pick on that class specifically). I think we need to balance landing important research innovations with not becoming too much of a franken-library (and being able to properly maintain everything that we claim to support, so that people can have a great experience using it). Tbh this is on us -- we haven't really provided the proper infrastructure or guidance on landing researchy changes. But one obvious thing I think we can do is create an analogous `torchtune/dev` folder for any (non-recipe) APIs that we consider not yet solidified. Then the division of APIs should roughly follow:

1) torchtune public modules: unit-tested, general API, documented in live API docs, guaranteed backwards compatibility
2) other modules under torchtune: still unit-tested, maybe more for internal use, relatively general, but no other hard guarantees 
3) torchtune/dev modules: no backwards compatibility guarantees, not unit-tested, 

(To clarify the main difference between (1) and (2) is whether we actually expose in `__all__` of our `__init__.py` files). I would hope that making this separation eliminates some of the questions around breaking existing generation APIs etc, and gives you the flexibility to try out the optimizations you need in a more lightweight way.

Lmk if this makes sense at a high level, if so happy to start diving into the weeds a bit more on what can go where. For things that fall into (1) and (2) we can also help with unit tests as needed, just want to make sure that we do a proper audit of everything before we start writing them.

torchtune/generation/_generation.py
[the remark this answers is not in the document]
sam-pi: I can at least vouch this appears to work for me!

recipes/dev/grpo_full_finetune_distributed.py
```diff
@@ -0,0 +1,1074 @@
...
+            self._checkpointer.save_checkpoint(
+                checkpoint_dict,
+                epoch=epoch,
+                intermediate_checkpoint=intermediate_checkpoint,
+            )
+            log.info(f"Saving checkpoint took {time.perf_counter() - start:.2f} secs")
+
+        torch.distributed.barrier()
+
+    def generate_trajectory(
```
sam-pi, line 683: In a way, this is effectively applying a "transform" to the input dataset, right? input_ids go in, and multiple generated responses come out (potentially with reference model logprobs). Could it potentially make sense to have a portion of these be applied as part of the dataloader/transforms?

torchtune/rlhf/sequence_processing.py
```diff
@@ -101,6 +101,42 @@ def logits_to_logprobs(
...
+    for chunk_start in range(0, batch_size, chunk_size):
+        chunk_end = min(chunk_start + chunk_size, batch_size)
+
+        # Process log_softmax for this batch chunk
+        chunk_log_probs = F.log_softmax(
+            logits[chunk_start:chunk_end] / temperature, dim=-1
+        )
+
+        # Gather for this chunk
+        result[chunk_start:chunk_end] = torch.gather(
```
TAplutos, line 133: I'm running into an error when running 
```
tune run --nproc_per_node 2 dev/grpo_full_finetune_distributed --config torchtune/recipes/configs/dev/3B_full_grpo.yaml
```
I tried fixing it by adding 
```
if sequences.size(1) != logits.size(1):
        sequences = sequences[:, -logits.size(1):]
```
at the start of the function but this causes more issues down the line.  Anyone else running into this same problem?

Also apologies if this isn't supposed to go directly here.  New to the OSS community

RedTachyon: Good catch, thanks! I had a bit of a mix-up with different versions of the generation, I'll push out a patch in a bit along with the general refactor

recipes/configs/dev/3B_sft_for_grpo.yaml
```diff
@@ -0,0 +1,109 @@
...
+
+# Tokenizer
+tokenizer:
+  _component_: torchtune.models.llama3.llama3_tokenizer
+  path: /tmp/Llama-3.2-3B/original/tokenizer.model
+  max_seq_len: null
+
+# Dataset
+dataset:
+  _component_: torchtune.dev.grpo.data.gsm8k_sft
```
TAplutos, line 31: nit but should be: torchtune.dev.grpo.gsm8k.gsm8k_sft

ebsmothers:
@RedTachyon thanks for doing that! At least at first glance this looks much easier for us to land, will give a proper review tomorrow.

RedTachyon:
@ebsmothers Gentle nudge - I'm hoping to have this merged sooner rather than later, so that I can start iterating on some more experimental stuff without getting into a branching nightmare

ebsmothers (commented):
Thanks for your patience @RedTachyon! A few more small comments and I think this is good to go (now that everything is in dev I will not be as pedantic about some of the design considerations). Main thing is to make sure that everything is runnable out-of-the-box. Also if you're able to share some of the logged metrics (successes, rewards, etc) on your latest runs that would be great as well.

felipemello1:
hey @RedTachyon , just wanted to thank you for being so responsive and putting the effort into this. We all appreciate it :)

RedTachyon:
Happy to contribute, and thanks for all the help in bringing this to a publishable state!

I applied the final fixes and launched another run to make sure it still learns - so far it's just about the same, and I don't really expect anything to be different.

As for some existing metrics - right now I don't have anything very pretty, but just to give a sense of what to expect, here are some wandb graphs (which I unfortunately can't share as actual wandb reports at the moment).

There are 8 curves, varying across training from the base model or from an SFT-initialized model, and then across the number of nodes (so effectively batch size - 2 nodes have an effective batch size of 16, 4 nodes - 32, 8 nodes - 64).

Success rate vs steps
image


Reward vs steps
image



Success rate vs time
image


Reward vs time
image


Note that the hardware isn't super consistent, so the time graphs are meant as a very rough estimate

recipes/dev/grpo_full_finetune_distributed.py
```diff
@@ -0,0 +1,1077 @@
...
+class FullGRPOFinetuneRecipeDistributed(FTRecipeInterface):
+    """
+    Full finetuning recipe for dense transformer-based LLMs such as Llama2, trained with GRPO. This recipe supports
+    distributed training and can be run on a single node (1 to 8 GPUs).
+
+    Features:
+        - FSDP. Supported using PyTorch's FSDP APIs. CPU offload of parameters, gradients, and optimizer states
+            is supported via ``fsdp_cpu_offload``. Resharding of parameters after the forward pass is
+            disabled for faster generation (corresponding to FULL_SHARD sharding strategy).
+            DDP is currently not supported. Training on CPU is not supported.
```
abdalgader-a, line 44: does DDP not supported?

PR #2328, opened 2025-02-01
[WIP] Add distributed inference for llama3.2 vision

acisseJZhong:
#### Context
- [ ] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

#### Changelog
*

#### Test plan

#### UX

### DIFF
--- a/recipes/configs/llama3_2_vision/90B_generation_distributed.yaml
+++ b/recipes/configs/llama3_2_vision/90B_generation_distributed.yaml
@@ -1,0 +1,51 @@
+# Config for running the InferenceRecipe in dev/generate_v2.py to generate output
+# from a Llama3.2 90B Vision Instruct model
+#
+# This config assumes that you've run the following command before launching:
+#   tune download meta-llama/Llama-3.2-90B-Vision-Instruct --output-dir /tmp/Llama-3.2-90B-Vision-Instruct --ignore-patterns "original/consolidated*"
+#
+# To launch, run the following command from root torchtune directory:
+#    tune run --nproc_per_node 8 dev/generate_v2_distributed --config llama3_2_vision/90B_generation_distributed
+
+output_dir: ./
+
+# Model arguments
+model:
+  _component_: torchtune.models.llama3_2_vision.llama3_2_vision_90b
+
+parallelize_plan:
+  _component_: torchtune.models.llama3_2_vision.llama3_2_vision_tp_plan
+
+# Transform arguments
+tokenizer:
+  _component_: torchtune.models.llama3_2_vision.llama3_2_vision_transform
+  path: /tmp/Llama-3.2-90B-Vision-Instruct/original/tokenizer.model
+  prompt_template: null
+  max_seq_len: 8192
+
+# Checkpointer
+checkpointer:
+  _component_: torchtune.training.FullModelHFCheckpointer
+  checkpoint_dir: /tmp/Llama-3.2-90B-Vision-Instruct/
+  checkpoint_files:
+    filename_format: model-{}-of-{}.safetensors
+    max_filename: "00037"
+  recipe_checkpoint: null
+  output_dir: ${output_dir}
+  model_type: LLAMA3_VISION
+
+# Device
+device: cuda
+dtype: bf16
+seed: 1234
+log_level: INFO
+
+# Generation arguments
+prompt:
+  system: You are a helpful assistant who responds like the author Shakespeare.
+  user:
+    image: https://cdn.britannica.com/61/93061-050-99147DCE/Statue-of-Liberty-Island-New-York-Bay.jpg
+    text: What is in this image?
+max_new_tokens: 200
+temperature: 0.6 # 0.8 and 0.6 are popular values to try
+top_k: 300
--- a/torchtune/_recipe_registry.py
+++ b/torchtune/_recipe_registry.py
@@ -446,6 +446,10 @@ _ALL_RECIPES = [
                 file_path="llama3_1/70B_generation_distributed.yaml",
             ),
             Config(
+                name="llama3_2/90B_generation_distributed",
+                file_path="llama3_2/90B_generation_distributed.yaml",
+            ),
+            Config(
                 name="llama3_3/70B_generation_distributed",
                 file_path="llama3_3/70B_generation_distributed.yaml",
             ),
--- a/torchtune/models/llama3_2_vision/_parallelism.py
+++ b/torchtune/models/llama3_2_vision/_parallelism.py
@@ -1,0 +1,49 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+from typing import Dict
+
+from torch.distributed._tensor import Replicate, Shard
+from torch.distributed.tensor.parallel import ColwiseParallel, RowwiseParallel
+from torch.distributed.tensor.parallel.style import ParallelStyle
+
+
+# Define the Tensor Parallel plan for Llama3.2 vision model
+LLAMA3_2_VISION_TP_PLAN = {
+    "decoder.tok_embeddings.embedding": RowwiseParallel(
+        input_layouts=Replicate()
+    ),
+    "decoder.tok_embeddings.fusion_embedding": RowwiseParallel(
+        input_layouts=Replicate()
+    ),
+    "decoder.output": ColwiseParallel(
+        output_layouts=Replicate()
+    ),
+    "decoder.layers.*.layer.attn.q_proj": ColwiseParallel(),
+    "decoder.layers.*.layer.attn.k_proj": ColwiseParallel(),
+    "decoder.layers.*.layer.attn.v_proj": ColwiseParallel(),
+    "decoder.layers.*.layer.attn.output_proj": RowwiseParallel(),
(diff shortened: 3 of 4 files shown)


torchtune/models/llama3_2_vision/_parallelism.py
```diff
@@ -0,0 +1,70 @@
...
+    "decoder.tok_embeddings.embedding": RowwiseParallel(
+        input_layouts=Replicate()
+    ),
+    "decoder.tok_embeddings.fusion_embedding": RowwiseParallel(
+        input_layouts=Replicate()
+    ),
+    "decoder.output": ColwiseParallel(
+        output_layouts=Replicate()
+    ),
+    "decoder.layers.*.attn.q_proj": ColwiseParallel(input_layouts=Replicate()),
```
SalmanMohammadi, line 39: very noob q: should we be sharding the kv-cache buffers too? are they automatically replicated across ranks?

acisseJZhong: great question! I believe the primary reason is that parallelization is typically applied during training rather than inference. Therefore most parallelism plan I saw (from Torchtitan and Torchtune) does not include the KV cache in their parallelism plans.

recipes/dev/generate_v2_distributed.py
```diff
@@ -15,8 +15,9 @@
 
 from torchtune import config, training, utils
 from torchtune.data import load_image, Message, padded_collate_tiled_images_and_mask
-
 from torchtune.generation import sample
+from torchtune.models.llama3_2_vision import parallelize_llama3_2_vision
```
pbontrager, line 19: A recipe shouldn't have model specific imports

recipes/dev/generate_v2_distributed.py
```diff
@@ -105,12 +106,18 @@ def setup(self, cfg: DictConfig) -> None:
...
-        parallelize_module(
-            model,
-            tp_device_mesh,
-            parallelize_plan=config.instantiate(cfg.parallelize_plan),
-        )
+        model = training.prepare_mha_for_tp(model, tp_device_mesh)
+
+        # Special handling for 3.2 vision model, we need to parallelize the model differently as wildcard is not
+        # working for now. Need to change back to wildcard once the issue is fixed.
+        if isinstance(model, DeepFusionModel):
```
pbontrager, line 113: This logic should be pushed into the "parallelize_module" utility and not exposed in the recipe.

torchtune/models/llama3_2_vision/_parallelism.py
```diff
@@ -0,0 +1,141 @@
...
+    RowwiseParallel,
+)
+from torch.distributed.tensor.parallel.style import ParallelStyle
+from torchtune.modules import TransformerSelfAttentionLayer
+from torchtune.modules.model_fusion import FusionLayer
+
+
+# Define the Tensor Parallel plan for Llama3.2 vision model
+# TODO: wildcard is not working for the 3.2 deep fusion model. Use `parallelize_llama3_2_vision` for now.
+LLAMA3_2_VISION_TP_PLAN = {
```
pbontrager, line 25: The encoder doesn't get parallelized?

torchtune/models/llama3_2_vision/_parallelism.py
```diff
@@ -0,0 +1,141 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
```
pbontrager, line 1: Since this is a patch to make deepfusion models work until the wildcard issue is fixed, I'd be inclined to move this out of llama3.2 and directly into the distributed file to get called when given a DeepFusion model within parallelize_model. It could be kept at the bottom of the file after some "###### TITLE #####" style separator

torchtune/training/_distributed.py
```diff
@@ -299,16 +298,15 @@ def load_from_full_model_state_dict(
                     ),
                     requires_grad=sharded_meta_param.requires_grad,
                 )
-
-            elif not hasattr(sharded_meta_param, "device_mesh"):
-                # In cases where parts of the model aren't sharded, some parameters will be plain tensors
-                sharded_tensor = full_tensor
-            else:
+            elif isinstance(sharded_meta_param, DTensor):
```
pbontrager, line 301: I don't remember why we were using device_mesh to infer DTensor's but this code cleanup does make me nervous that there was an edge case where checking for DTensor correctly didn't work. Are you sure there's no such case?

torchtune/training/_distributed.py
```diff
@@ -624,4 +622,6 @@ def prepare_mha_for_tp(
             m.num_heads = m.num_heads // tp_size
             m.num_kv_heads = m.num_kv_heads // tp_size
             m.embed_dim = m.embed_dim // tp_size
+    if is_fusion_model:
```
pbontrager, line 625: So before we were only returning the decoder and now we return the full model? Was this a bug?

pbontrager (commented):
Thanks for enabling this. I left a few comments around where we want to put and expose these changes.

torchtune/models/llama3_2_vision/_parallelism.py
acisseJZhong: yes, I think encoder doesn't needs to be parallelized as it's not that big compared to decoder layers

torchtune/training/_distributed.py
acisseJZhong: it actually took a while for me to debug this; https://github.com/pytorch/pytorch/blob/main/torch/distributed/tensor/_api.py#L239 DTensor doesn't seems to have this device_mesh field, so this elif block was never activated. 

Curious why we can't just do isinstance DTensor?

torchtune/training/_distributed.py
acisseJZhong: this is already fixed on the public repo already i believe. I need to rebase.

torchtune/models/llama3_2_vision/_parallelism.py
acisseJZhong: are you suggesting moving this hack logic into _distributed.py? I prefer keeping it here as it's 3.2 model specific?

recipes/dev/generate_v2_distributed.py
acisseJZhong: yeah I can have another `parallelize_module` function in `_distributed.py`. But that wouldn't solve the problem(having this if else check)? it just makes the recipe looks cleaner.

torchtune/_recipe_registry.py
```diff
@@ -449,6 +449,10 @@ class Recipe:
                 name="llama3_1/70B_generation_distributed",
                 file_path="llama3_1/70B_generation_distributed.yaml",
             ),
+            Config(
+                name="llama3_2/90B_generation_distributed",
+                file_path="llama3_2/90B_generation_distributed.yaml",
```
ebsmothers, line 454:
```suggestion
                name="llama3_2_vision/90B_generation_distributed",
                file_path="llama3_2_vision/90B_generation_distributed.yaml",
```

recipes/dev/generate_v2_distributed.py
ebsmothers: Bumping this comment, can we take this out?

torchtune/models/llama3_2_vision/_parallelism.py
ebsmothers: To cast a tiebreaking vote, I also prefer to keep it here. (I don't wanna add even more code to `_distributed.py` than we have already)

recipes/dev/generate_v2_distributed.py
ebsmothers: Yeah the if/else is not going to go away. But we shouldn't put it in the recipe (more boilerplate that the user shouldn't have to care about)

torchtune/training/_distributed.py
ebsmothers: Sorry I'm also a bit confused by this.. what is the actual change that's happening here? I think DTensor does have this field, see [here](https://github.com/pytorch/pytorch/blob/main/torch/distributed/tensor/_api.py#L580). Is there some case where it's null?

acisseJZhong: Yeah we when using the `not hasattr(sharded_meta_param, "device_mesh")`, its not entering this block. I changed the logic to using `isinstance(DTesnor)`

recipes/configs/llama3_2_vision/90B_generation_distributed.yaml
```diff
@@ -0,0 +1,51 @@
...
+#    tune run --nproc_per_node 8 dev/generate_v2_distributed --config llama3_2_vision/90B_generation_distributed
+
+output_dir: ./
+
+# Model arguments
+model:
+  _component_: torchtune.models.llama3_2_vision.llama3_2_vision_90b
+
+parallelize_plan:
+  _component_: torchtune.models.llama3_2_vision.llama3_2_vision_tp_plan
```
ebsmothers, line 17: I know it's due to the wildcard issue but it's confusing to expose this in the config and then use `parallelize_llama3_2_vision` in the recipe. We should simplify this (personally I would just rip out the wildcards until they're actually supported, but up to you)

ebsmothers (commented):
A few more comments, the main thing is to understand why the fix in `load_from_full_model_state_dict` is the correct one. Two other general things: 

(1) Can you share tok/s and peak memory numbers in the PR summary? 
(2) Does the job actually shutdown after generation completes? At least for me it does the generation and then just hangs until NCCL timeout. (Maybe it's a problem unique to my machine, but just want to confirm.)

Issue #2280, opened 2025-01-19
Roadmap for other parallelisms


rahul-sarvam:
First of all, thanks for all the great work you guys are doing! I find the code a pleasure to read and extend 👍 

I am writing to ask whether there are plans to expand torchtune to include DDP, TP, and other parallelisms out there. I think the ability to run multinode finetuning will make torchtune even more widely adopted. I know it is trivial to extended the current code to run on multiple nodes since `tune run` extends `torchrun`. However, as you know, for bigger models, FSDP recomputations and communication overhead kills any scaling we can get. 

I would gladly contribute to any ongoing work already going on in this direction. Thanks!

RdoubleA:
Hi @rahul-sarvam, thanks for the questions!
- DDP APIs should work out of the box for our models, we just have not yet tested these or provided any utilities in the library for it, since it has not yet been requested. If you are planning to explore this direction, we are happy to help with any roadblocks you run into, and there could be opportunities to contribute here.
- We just landed support for TP and will be socializing them soon and providing documentation! They should be available in nightlies.

EugenHotaj:
>  They should be available in nightlies. See the distributed inference script for how it can be used: https://github.com/pytorch/torchtune/blob/main/recipes/dev/generate_v2_distributed.py

Are there plans to enable this for the distributed training recipes?

acisseJZhong:
> > They should be available in nightlies. See the distributed inference script for how it can be used: https://github.com/pytorch/torchtune/blob/main/recipes/dev/generate_v2_distributed.py
> 
> Are there plans to enable this for the distributed training recipes?

Hi Eugen! Nice to meet you here :)) 
Yes we are already working on extending to multinode support and bringing 2D parallelism (FSDP + TP) into the training script. It should become available in the next week or so, and I will tag you!

EugenHotaj:
@acisseJZhong nice to see you again! Thanks so much, looking forward to using tp once it's ready 🙂

dz1iang:
> > > They should be available in nightlies. See the distributed inference script for how it can be used: [`main`/recipes/dev/generate_v2_distributed.py](https://github.com/pytorch/torchtune/blob/main/recipes/dev/generate_v2_distributed.py)
> > 
> > 
> > Are there plans to enable this for the distributed training recipes?
> 
[...]

PR #2204, opened 2024-12-24
[WIP] Add 2D Parallelism (FSDP + Tensor Parallel) LoRA

matthew-hippocratic:
#### Context
What is the purpose of this PR?
- [X] add a new feature - support for 2D parallelism using FSDP+TP with LoRA
- [X] other (Request help from the community to productionize this minimal working example. This code works with single and multinode setups and I have provided the run commands with configs below. However, it is hacky is isn't intended to merge at this point. I thought this feature would be useful and I'm running out of time to work on it so I wanted to see if others would like to help bring it to the finish line! If interested, see **Limitations** below)

#### Changelog
What are the changes made in this PR?
* Add support for 2D parallelism (FSDP+TP) of LoRA Llama models in `lora_finetune_distributed.py`
    * Add 2D device_mesh that shards along "dp" (FSDP) and "tp" dimensions
    * Utilize `DTensor` like in [torchtitan](https://github.com/pytorch/torchtitan/blob/main/torchtitan/parallelisms/parallelize_llama.py) for easy-to-understand device_mesh sharding
    * Modify the `DistributedSampler` to shard data along "dp" dimension
* Modify `LoRALinear` to remap naming so that the Tensor Parallel layer_plan can target the main model weights
    * This involves renaming when training, then renaming back to original names for saving the checkpoint(s)

#### Test plan
Please make sure to do each of the following if applicable to your PR. If you're unsure about any one of these just ask and we will happily help. We also have a [contributing page](https://github.com/pytorch/torchtune/blob/main/CONTRIBUTING.md) for some guidance on contributing.

- [X] run pre-commit hooks and linters (make sure you've first installed via `pre-commit install`)
- [X] update [docstrings](https://github.com/pytorch/torchtune/tree/main/docs/source) for any new or updated methods or classes
- [X] manually run any new or modified recipes with sufficient proof of correctness
- [X] include relevant commands

Successful run commands:
_Llama-3.1-8B-Instruct on a single 8xH100 Node (FSDP=1, TP=8)_
```
tune run --rdzv-endpoint <node_ip> --nnodes 1 --nproc-per-node 8 --rdzv-id <slurm_job_id> --rdzv-backend=c10d \
    lora_finetune_distributed --config recipes/configs/llama3_1/8B_lora.yaml \
    tokenizer.path=<path_to_/meta-llama/Llama-3.1-8B-Instruct/original/tokenizer.model> \
[...]

PR #2330, opened 2025-02-02
TP + FSDP distributed training (full finetuning)

acisseJZhong:
#### Context
- [x] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

#### Changelog
Enabled TP training on top of FSDP. This is both from user request https://github.com/pytorch/torchtune/issues/2280, and also to improve memory so that we can get rid of `fsdp_cpu_offload` when training larger model. 

* **Reference PRs for this work: https://github.com/pytorch/torchtune/pull/2204, https://github.com/pytorch/torchtune/pull/2318**

Note: 
- TP + FSDP doesn't work with fused optimizer yet(`optimizer.fused` needs to be False). 
- Applying TP needs to happen after AC since AC changes the fqn. (this causes a few hours debugging :( )

#### Test plan

Running TP = 2, FSPD = 4:
image
image
Running TP = 1, FSDP = 8:
image
image
#### UX

### DIFF
--- a/recipes/full_finetune_distributed.py
+++ b/recipes/full_finetune_distributed.py
@@ -12,6 +12,7 @@ from typing import Any, Dict, List, Optional, Tuple, Union
 from warnings import warn
 
 import torch
+import torch.distributed as dist
 from omegaconf import DictConfig, ListConfig
 
 from torch import nn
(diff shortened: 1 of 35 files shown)


recipes/full_finetune_distributed.py
```diff
@@ -502,6 +512,27 @@ def _setup_model(
         if self._compile:
             training.compile_model(model, verbose=self._is_rank_zero)
 
+        self.device_mesh = dist.init_device_mesh(
```
joecummings, line 515: I would rather explicitly save what we need here. So instead of putting the whole device_mesh on the object, later on if data_parallel_dim > 1, save self.local_size = dp_mesh.size() and self.local_rank = dp_mesh.get_local_rank() else self.local_size = self.world_size and self.local_rank = self.rank.

Does that make sense?

joecummings (approved):
Looks awesome! Just need to make sure the merge changes are correct.

PR #2350, opened 2025-02-05
HF tokenizers: initial base tokenizer support

### LINKED ISSUE #2212: [Question] what to do when model doesn't have `tokenizer.model`?
while `tokenizer.model` is required in yaml config, but there are many models that doesn't have `tokenizer.model` (example: [unsloth/Llama-3.2-1B](https://huggingface.co/unsloth/Llama-3.2-1B/tree/main))

In these cases, how can we use `tokenizer.json` or `tokenizer_config.json` that are included in almost all model instead of `tokenizer.model`?

ebsmothers:
Fixes #2212

This is an initial PR to support general tokenizers from Hugging Face via a tokenizer.json file. This is just a starting point to parse relevant JSON files, infer BOS and EOS, and define the encode/decode methods. Special tokens are not handled here, nor is lining things up with our `tokenize_messages` abstraction. The goal here is to unblock initial usage of models that contain only tokenizer.json on the HF hub. With this PR, it should be possible to define a `ModelTokenizer` containing `HFTokenizer` as a component, similar to how our other `ModelTokenizers` contain either `SentencePieceBaseTokenizer` or   `TikTokenBaseTokenizer`. This also means that we add a dependency on the tokenizers library, which I think is a necessity to support a broader set of models.

## Test plan

Added a unit test to verify parity on our toy tiktoken tokenizer. The HF tokenizer.json version of this tokenizer was generated with [this script](https://gist.github.com/ebsmothers/55b2f177f5ed15a3b81508f8f8b91159).

```
pytest tests/torchtune/modules/transforms/tokenizers/test_hf_tokenizer.py
...
========== 4 passed, 1 warning in 0.08s ============
```

Will also provide a script comparing a proper model tokenizer across torchtune and HF implementations (but this shouldn't go in our CI due to the larger tokenizer model size).


torchtune/modules/transforms/tokenizers/_hf_tokenizer.py
```diff
@@ -0,0 +1,119 @@
...
+class HFTokenizer(BaseTokenizer):
+    """
+    A wrapper around Hugging Face tokenizers. This class can be used to load from a
+    Hugging Face tokenizer.json file into a torchtune BaseTokenizer.
+
+    Args:
+        path (str): Path to tokenizer.json file
+        config_path (Optional[str]): Path to tokenizer_config.json file. Default: None
+        generation_config_path (Optional[str]): Path to generation_config.json file.
+            Default: None
```
felipemello1, line 23: we have 3 paths. For the last 2 paths, they are more specific, which is good. I think we could rename path -> tokenizer_json_path, to make it explicit

torchtune/modules/transforms/tokenizers/_hf_tokenizer.py
```diff
@@ -0,0 +1,119 @@
...
+            self.bos_id = self.generation_config.get("bos_token_id")
+            self.eos_id = self.generation_config.get("eos_token_id")
+
+        if self.config:
+            bos_token = self._get_token_from_config(self.config, "bos_token")
+            eos_token = self._get_token_from_config(self.config, "eos_token")
+            if bos_token is not None and self.bos_id is None:
+                self.bos_id = self.hf_tokenizer.token_to_id(bos_token)
+            if eos_token is not None and self.eos_id is None:
+                self.eos_id = self.eos_id or self.hf_tokenizer.token_to_id(eos_token)
```
felipemello1, line 81: I dont know much about the files, should we add a sanity check to see if config.eos == generation_config.bos_token_id?

torchtune/modules/transforms/tokenizers/_hf_tokenizer.py
```diff
@@ -0,0 +1,119 @@
...
+            with open(config_path, "rb") as f:
+                self.config = json.load(f)
+        else:
+            self.config = None
+        if generation_config_path:
+            with open(generation_config_path, "rb") as f:
+                self.generation_config = json.load(f)
+        else:
+            self.generation_config = None
+        self._infer_bos_eos_tokens()
```
felipemello1, line 43: lets raise an error if self.generation_config is None and self.config is None

Also, apparently we always use one or the other. We could simplify the rest of the code by just making something like:

```
self.config = generation_config or config
```

felipemello1 (commented):
just checked _hf_tokenizer. Didnt look at the tests yet

torchtune/modules/transforms/tokenizers/_hf_tokenizer.py
ebsmothers: Sorry maybe I'm confused.. why would we expect eos to equal bos? But if you mean check on eos == eos and bos == bos across config and generation config, I think we can add that check

torchtune/modules/transforms/tokenizers/_hf_tokenizer.py
ebsmothers: Yeah I think it's simplest if we just use config. At the same time, I read (on some random forum) that generation config should take precedence over config? So currently I'm doing that. One thing that's nice about generation config is that it directly gives the int value for eos and bos, whereas the regular config tends to give the string value, which then needs to be converted. Aside from that, config has chat template, special tokens, and all that other stuff we would need. So yeah right now I am hedging a bit and keeping both around because I suspect we may need both at different times

joecummings: Cite your sources.

ebsmothers: [Source](https://github.com/huggingface/transformers/issues/25395#issuecomment-1671863257) 

Like I said, maybe not the actual source-of-truth here. Especially as I look through the `PreTrainedTokenizerBase` class -- they seem to really only rely on `tokenizer_config.json` ([ref](https://github.com/huggingface/transformers/blob/main/src/transformers/tokenization_utils_base.py#L2106)). For what it's worth, litgpt also gives `tokenizer_config.json` [precedence](https://github.com/Lightning-AI/litgpt/blob/f6031e3a88e272ec86ad8f412573699589f4d41b/litgpt/tokenizer.py#L30-L40), and it seems to work OK for them (also shout-out to @Andrei-Aksionov who helped give me some pointers here).

So yeah perhaps I will switch the order to read from tokenizer config, then generation config. I do trust the code itself + litgpt's implementation more than that one specific issue comment.

Andrei-Aksionov: Oh, thanks for the shout-out 😊.

It was long time since I worked with tokenizers, but yeah, we decided to rely on `tokenizer.json` and if it's in the repo - use HF Tokenizers. Before `tokenizer.model` and thus SentencePiece had a precedence.

If to take a closer look at the init process of AutoTokenizer, one can see that if there is only a `tokenizer.model` file (a vocab for SP), then this vocab is loaded and then the SP tokenizer is converted to a HF Tokenizers format. The latter allows to extend the vocab with special tokens (often comes with instruct versions of models). With SP to perform the same you will need a much more fiddling (i.e. operate with protobuf format).

So, basically, HF never works with SP tokenizer, even if only it's vocab is provided.
(I [asked](https://huggingface.co/infly/OpenCoder-8B-Instruct/discussions/4) in the repo of OpenCoder what to do if there is only `tokenizer.model`, additional tokens in `tokenizer_config.json` and I don't want to use AutoTokenizer, but never got an answer. Do they all rely on AutoTokenizer?)

In other words, it's good that you are working on supporting `tokenizer.json`.

-----

> For what it's worth, litgpt also gives tokenizer_config.json [precedence](https://github.com/Lightning-AI/litgpt/blob/f6031e3a88e272ec86ad8f412573699589f4d41b/litgpt/tokenizer.py#L30-L40)

FYI, after looking at the codebase, previously `generation_config.json` had a precedence, but it was changed in this  [commit](https://github.com/Lightning-AI/litgpt/pull/552/commits/219662c8672ada6a3b991bb405f9ca89ea79651b).

pyproject.toml
```diff
@@ -23,6 +23,7 @@ dependencies = [
     "sentencepiece",
     "tiktoken",
     "blobfile>=2",
+    "tokenizers",
```
RdoubleA, line 26: I had thought we would make this a dev dependency and conditionally import this in hf_tokenizer

ebsmothers: Yeah I was on the fence about this. But actually I think the upstream dependencies of tokenizers are already a subset of our upstream dependencies. So in that case it's not like we are pulling in a bunch of new transitive deps, this is really just tokenizers and nothing else. Because of that I feel OK about having it as a core dep, but happy to take the approach you're suggesting if you (or others) strongly disagree

torchtune/modules/transforms/tokenizers/_hf_tokenizer.py
ebsmothers: Update: I opted not to add the check, mainly because it's just some extra logic to validate something that shouldn't happen (the user has config and generation config that don't line up). In that case, we should just be explicit in our docs that we prefer config.json whenever it's passed. Lmk if any concerns with that though

pyproject.toml
joecummings: Just documenting. Tokenizer package is roughly 3 MB (https://pypi.org/project/tokenizers/#files)

tests/torchtune/modules/transforms/tokenizers/test_hf_tokenizer.py
```diff
@@ -0,0 +1,126 @@
...
+            special_tokens={
+                "<|test_token_0|>": 2000,
+                "<|test_token_1|>": 2001,
+            },
+        )
+
+    @pytest.fixture
+    def texts(self):
+        return [
+            "I can see the sun. But even if I cannot see the sun, I know that it exists.",
```
joecummings, line 38: "I am not a cultivated man, brother, but I've thought a lot about this."

torchtune/modules/transforms/tokenizers/__init__.py
```diff
@@ -4,6 +4,7 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
+from ._hf_tokenizer import HFTokenizer
```
joecummings, line 7: nit: it's weird that we spell out SentencePiece and TikToken, but not HuggingFace...

torchtune/modules/transforms/tokenizers/_hf_tokenizer.py
```diff
@@ -0,0 +1,132 @@
...
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+import json
+from typing import Any, Dict, List, Optional
+
+from tokenizers import Tokenizer
+from torchtune.modules.transforms.tokenizers._utils import BaseTokenizer
```
joecummings, line 11: If we add one more submodule, Satan gives us a cookie and a "good job" sticker

torchtune/modules/transforms/tokenizers/_hf_tokenizer.py
```diff
@@ -0,0 +1,132 @@
...
+
+
+class HFTokenizer(BaseTokenizer):
+    """
+    A wrapper around Hugging Face tokenizers. This class can be used to load from a
```
joecummings, line 16: nit: point to the repo.

torchtune/modules/transforms/tokenizers/_hf_tokenizer.py
```diff
@@ -0,0 +1,132 @@
...
+    Hugging Face tokenizer.json file into a torchtune BaseTokenizer.
+
+    This class will load the tokenizer.json file from tokenizer_json_path. It will
+    attempt to infer BOS and EOS token IDs from config.json if possible, and if not
+    will fallback to inferring them from generation_config.json.
+
+    Args:
+        tokenizer_json_path (str): Path to tokenizer.json file
+        config_path (Optional[str]): Path to tokenizer_config.json file. Default: None
```
joecummings, line 25: Should be called `tokenizer_config_json_path`. Yes, I know that's a mouthful, but it pattern matches with the name of the file and there's already other "configs" in a Hugging Face model dir.

torchtune/modules/transforms/tokenizers/_hf_tokenizer.py
```diff
@@ -0,0 +1,132 @@
...
+            Default: None
+
+    Raises:
+        ValueError: If neither config_path or generation_config_path are specified.
+    """
+
+    def __init__(
+        self,
+        tokenizer_json_path: str,
+        config_path: Optional[str] = None,
```
joecummings, line 36: kwargs only

torchtune/modules/transforms/tokenizers/_hf_tokenizer.py
```diff
@@ -0,0 +1,132 @@
...
+            with open(config_path, "rb") as f:
+                self.config = json.load(f)
+        else:
+            self.config = None
+        if generation_config_path:
+            with open(generation_config_path, "rb") as f:
+                self.generation_config = json.load(f)
+        else:
+            self.generation_config = None
+        self._infer_bos_eos_tokens()
```
joecummings, line 54: Should we give them an option to explicitly set the BOS and EOS if it can't be inferred from the file? Or are they just screwed?

torchtune/modules/transforms/tokenizers/_hf_tokenizer.py
```diff
@@ -0,0 +1,132 @@
...
+        ValueError: If neither config_path or generation_config_path are specified.
+    """
+
+    def __init__(
+        self,
+        tokenizer_json_path: str,
+        config_path: Optional[str] = None,
+        generation_config_path: Optional[str] = None,
+    ):
+        self.hf_tokenizer = Tokenizer.from_file(tokenizer_json_path)
```
joecummings, line 39: Is there a reason this has to be `self.hf_tokenizer` instead of just `tokenizer`? You already know it's a HFTokenizer class.

joecummings (commented):
neat!

torchtune/modules/transforms/tokenizers/_hf_tokenizer.py
ebsmothers: Yeah it can just be `self.tokenizer`

torchtune/modules/transforms/tokenizers/_hf_tokenizer.py
ebsmothers: I am hedging a bit here for future work -- eventually I would like to have a `ModelTokenizer` wrapping this. In that case I think we should require at least `tokenizer_config.json` and explicitly giving EOS and BOS will probably not be enough

torchtune/modules/transforms/tokenizers/__init__.py
ebsmothers: OK yeah we can spell it out

pyproject.toml
ebsmothers: Isn't torch close to 1 GB? Anyways if everyone would prefer to make this an optional dep I am fine with it

joecummings: Yeah, 3 MB is nothing. I'm fine with it as a core dep.

joecummings (approved):
bueno

torchtune/modules/transforms/tokenizers/_hf_tokenizer.py
```diff
@@ -0,0 +1,133 @@
...
+            Default: None
+
+    Raises:
+        ValueError: If neither tokenizer_config_json_path or generation_config_path are specified.
+    """
+
+    def __init__(
+        self,
+        *,
+        tokenizer_json_path: str,
```
joecummings, line 36: Could've sworn I made a comment here, but b/c tokenizer_json_path is required, it can be a positional argument. My thinking is that kwargs would apply to tokenizer_config_json_path and generation_config_path

ebsmothers: Ah duh, good point. I should've thought about it more carefully anyways

PR #2356, opened 2025-02-07
Generic classifier builder

SalmanMohammadi:
This PR adds support for a generic classifier builder. Currently we require explicit classifier builders to be defined for every model which we wish to use as a classifier. This is costly to maintain. 

We can instead define a generic classifier builder which simply adapts the final layer of any decoder-based model (how strict should this contract be?) to the desired number of classes.

This PR does not include support for PEFT-based classifier model builders. This is because there are no use-cases in the library for such model builders at this time.


torchtune/models/common.py
```diff
@@ -0,0 +1,33 @@
...
+
+    Args:
+        num_classes (int): The number of classes for the classifier.
+        base_model (str): The path to the base model.
+        base_model_kwargs (Dict[str, Any]): Keyword arguments for the base model.
+
+    Returns:
+        nn.Module: The classifier model.
+    """
+    model = _get_component_from_path(base_model)(**base_model_kwargs)
```
RdoubleA, line 25: if this is bound by the models in our library, then we would need to test that we can successfully create a classifier model for every model in our library, and any future models that are added. Can users pass in a custom model path? If that's not the intention, then we'll need to add some more gating. For example, if we check for if it's a TransformerDecoder instance that would easily keep this constrained to models that we know with confidence can be made into classifiers.

RdoubleA: Yep agreed, if we test against TransformerDecoder that should satisfy all the conditions I mentioned

torchtune/models/common.py
```diff
@@ -0,0 +1,43 @@
...
+
+    """
+    model = _get_component_from_path(base_model_path)(**base_model_kwargs)
+
+    if isinstance(model.output, nn.Linear):
+        del model.output.weight
+        if hasattr(model.output, "bias"):
+            del model.output.bias
+    model.output = nn.Linear(model.head_dim * model.num_heads, num_classes, bias=False)
+    return model
```
felipemello1, line 43: this wont work for TiedEmbedding layer.

Why use:
```
if isinstance(model.output, nn.Linear)
```

Instead of:

```
if hasattr(model, 'output')
```

also, for multimodal, it would have to be model.decoder.output, instead of model.output

recipes/configs/mistral/7B_full_ppo_low_memory.yaml
```diff
@@ -46,17 +46,18 @@ policy_model:
 # we need to manually build the mistral classifier model
 # because our reward model checkpoint has a larger vocabulary size (due to an added padding token)
 reward_and_value_model:
-  _component_: torchtune.models.mistral._component_builders.mistral_classifier
+  _component_: torchtune.models.common.classifier_model
+  base_model_path: torchtune.models.mistral.mistral_7b
```
felipemello1, line 50: this works, but if we have nested instantiation (which we dont), maybe a better design would be:

reward_and_value_model:
  _component_: torchtune.models.common.classifier_model
  model:
          _component_: torchtune.models.mistral.mistral_7b
          args

SalmanMohammadi: Do you mean changing `base_model_path` to `model` here? This sounds good to me : )

torchtune/models/common.py
SalmanMohammadi: > Should we use
> `if hasattr(model, 'output')`

Hmm so the logic would be something like -

```python

    if isinstance(model.output, nn.Linear):
        del model.output.weight
        if hasattr(model.output, "bias"):
            del model.output.bias
            
```

In the `TiedEmbedding` case we have no weight/bias parameters to delete, so we won't need to enter this block right? Let me know if I'm missing something or misunderstanding here : )

> also, for multimodal, it would have to be model.decoder.output, instead of model.output

I agree, but similar to PEFT, we don't support any use-cases for MM classifier models at the moment, so I thought I'd keep the design super simple and directed. What do you think?

recipes/configs/mistral/7B_full_ppo_low_memory.yaml
felipemello1: edited so that it renders a bit better

torchtune/models/common.py
felipemello1: sorry, still not clear to me why not just replace the output head if there is one, instead of deleting the weights first.

Just do:
```
if hasattr(model, 'output'):
     model.output = new_head()
```

> I agree, but similar to PEFT, we don't support any use-cases for MM classifier models at the moment, so I thought I'd keep the design super simple and directed. What do you think?

I dont think it hurts to add and extra "else" to cover this case, if someone tries to experiments with it. If you are strongly against it, maybe we could raise an warning?

```
if hasattr(model, 'output'):
     do_something:
else:
     raise "Could not find model.output"
```

pbontrager:
Instead of adding a common.py function here, could this be a classifier.py file in torchtune/modules?

recipes/configs/mistral/7B_full_ppo_low_memory.yaml
```diff
@@ -46,17 +46,18 @@ policy_model:
 # we need to manually build the mistral classifier model
 # because our reward model checkpoint has a larger vocabulary size (due to an added padding token)
 reward_and_value_model:
-  _component_: torchtune.models.mistral._component_builders.mistral_classifier
+  _component_: torchtune.modules.classifier.classifier_model
```
joecummings, line 49: NEAT

torchtune/modules/classifier.py
```diff
@@ -0,0 +1,55 @@
...
+            model.head_dim * model.num_heads, num_classes, bias=False
+        )
+    elif hasattr(model, "decoder") and hasattr(model.decoder, "output"):
+        del model.decoder.output
+        model.decoder.output = nn.Linear(
+            model.decoder.head_dim * model.decoder.num_heads, num_classes, bias=False
+        )
+    else:
+        raise ValueError("Could not find a valid output layer to adapt.")
+    return model
```
felipemello1, line 55: looks good! Not for this PR, but this wont work with PEFT. I dont think we want to go there, just mentioning it. Maybe we could add a warning: "found LoRA module, this is not supported, replacing it with a new head", but this might be an overkill. TLDR: its fine to keep as is.

felipemello1 (commented):
will approve after CLI is green. Thanks for this PR!!

torchtune/modules/classifier.py
SalmanMohammadi: I can update the docstring but I think adding checks to cover PEFT cases might be overkill : )

PR #2358, opened 2025-02-07
Add mistral small

AndrewMead10:
#### Context
- [x] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

#### Changelog
* Added mistral small (24B) model builders 

#### Test plan

#### UX

### DIFF
--- a/torchtune/models/mistral/_model_builders.py
+++ b/torchtune/models/mistral/_model_builders.py
@@ -48,7 +48,13 @@ def mistral_7b() -> TransformerDecoder:
     )
 
 
-def mistral_tokenizer(path: str, max_seq_len: Optional[int] = None, prompt_template: Optional[_TemplateType] = "torchtune.models.mistral.MistralChatTemplate") -> MistralTokenizer:
+def mistral_tokenizer(
+    path: str,
+    max_seq_len: Optional[int] = None,
+    prompt_template: Optional[
+        _TemplateType
+    ] = "torchtune.models.mistral.MistralChatTemplate",
+) -> MistralTokenizer:
     """
     Tokenizer for Mistral models.
 
@@ -64,7 +70,15 @@ def mistral_tokenizer(path: str, max_seq_len: Optional[int] = None, prompt_templ
     Returns:
         MistralTokenizer: Instantiation of the Mistral tokenizer
     """
-    return MistralTokenizer(path=path, max_seq_len=max_seq_len, prompt_template=_get_prompt_template(prompt_template) if prompt_template is not None else None)
+    return MistralTokenizer(
+        path=path,
+        max_seq_len=max_seq_len,
+        prompt_template=(
+            _get_prompt_template(prompt_template)
+            if prompt_template is not None
+            else None
+        ),
+    )
 
 
 def lora_mistral_7b(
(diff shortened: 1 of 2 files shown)


torchtune/models/mistral/_model_builders.py
```diff
@@ -214,3 +228,149 @@ def lora_mistral_reward_7b(
...
+        vocab_size=131_072,
+        num_layers=40,
+        num_heads=32,
+        num_kv_heads=8,
+        embed_dim=5120,
+        intermediate_dim=32768,
+        max_seq_len=32768,
+        attn_dropout=0.0,
+        norm_eps=1e-5,
+        rope_base=10_000_000,
```
ebsmothers, line 248: I think this should be 100M, not 10M? [Ref](https://huggingface.co/mistralai/Mistral-Small-24B-Base-2501/blob/main/config.json#L19) (And similar comment in the other builders)

ebsmothers (commented):
Thanks @AndrewMead10 for the PR! Overall this looks good to me, just one minor comment. Do you also want to add some configs under `recipes/configs/mistral` for folks to use this model?

joecummings (commented):
Can we actually move this to it's own section and own folder? Mistral Small (even though it's the same architecture) is a different model and under a different release.

torchtune/models/mistral/_model_builders.py
AndrewMead10: misread from the config, fixed

AndrewMead10:
Added some recipes, just followed the mistral 7B ones. Moved the model into its own folder, and updated rope base. 

I am currently getting the following error when trying to run 24B_qlora_single_device

```
size mismatch for layers.28._checkpoint_wrapped_module.attn.q_proj.weight: copying a param with shape torch.Size([4096, 5120]) from checkpoint, the shape in current model is torch.Size([5120, 5120])
size mismatch for layers.28._checkpoint_wrapped_module.attn.k_proj.weight: copying a param with shape torch.Size([1024, 5120]) from checkpoint, the shape in current model is torch.Size([1280, 5120]).
size mismatch for layers.28._checkpoint_wrapped_module.attn.v_proj.weight: copying a param with shape torch.Size([1024, 5120]) from checkpoint, the shape in current model is torch.Size([1280, 5120]).
size mismatch for layers.28._checkpoint_wrapped_module.attn.output_proj.weight: copying a param with shape torch.Size([5120, 4096]) from checkpoint, the shape in current model is torch.Size([5120, 5120]).
```

i am unsure if this is due to a diff in the modeling code or if I did something wrong in the model setup. My guess is the former, but I dont have the time rn to go and compare the diff from your impl to the hf one.

ebsmothers:
Oh I see the issue @AndrewMead10 -- the Mistral Small config sneakily uses `head_dim=128`, which takes precedence over `hidden_size // num_attention_heads` [here](https://github.com/huggingface/transformers/blob/main/src/transformers/models/mistral/modeling_mistral.py#L140). Actually I think this should be a pretty easy fix: you can follow our Gemma builders, which take in `head_dim` as an explicit arg and construct `MultiHeadAttention` using that (instead of setting `head_dim = embed_dim // num_heads` as we do in some other places). See [here](https://github.com/pytorch/torchtune/blob/504cbea31e9a4f222deaf379a7292a0f7af895e7/torchtune/models/gemma/_component_builders.py#L86-L90) for how that can be done.

AndrewMead10:
Thanks for the tip @ebsmothers, added that in, and it fixed a bunch of the shape errors, still getting the following mismatch tho for the output proj

        size mismatch for layers.0._checkpoint_wrapped_module.attn.output_proj.weight: copying a param with shape torch.Size([5120, 4096]) from checkpoint, the shape in current model is torch.Size([5120, 5120]).
        size mismatch for layers.1._checkpoint_wrapped_module.attn.output_proj.weight: copying a param with shape torch.Size([5120, 4096]) from checkpoint, the shape in current model is torch.Size([5120, 5120]).
        etc
        
I looked at the HF impl and the gemma impl and wasnt able to see what or why this mismatch is there. HF just the hidden size, which should be 5120 based on the config, right?

ebsmothers:
@AndrewMead10 I suspect you will also need to make [this change](https://github.com/pytorch/torchtune/blob/504cbea31e9a4f222deaf379a7292a0f7af895e7/torchtune/models/gemma/_component_builders.py#L330) to the output proj in each self-attention layer (that example is for LoRA, but really any self-attention builder should be using `num_heads * head_dim` instead of the current `embed_dim`)

joecummings:
> @AndrewMead10 I suspect you will also need to make [this change](https://github.com/pytorch/torchtune/blob/504cbea31e9a4f222deaf379a7292a0f7af895e7/torchtune/models/gemma/_component_builders.py#L330) to the output proj in each self-attention layer (that example is for LoRA, but really any self-attention builder should be using `num_heads * head_dim` instead of the current `embed_dim`)

@AndrewMead10 Did this work? Let's get this landed!!

AndrewMead10:
@joecummings model can now be loaded, the issue now is that mistral changed the tokenizer to tekken (tiktoken based) so we need make a new tokenizer class for mistral small. I was going to look into it more today, but if you or @ebsmothers have any tips or pointers for how to implement it that would be great. I was looking through the code base and didn't see any other tiktoken tokenizers being used, i assume setting it up is similar to the sentence piece tokenizers?

joecummings:
> @joecummings the model can now be loaded, the issue now is that mistral changed the tokenizer to tekken (tiktoken based) so we need make a new tokenizer class for mistral small. I was going to look into it more today, but if you or @ebsmothers have any tips or pointers for how to implement it that would be great. I was looking through the code base and didn't see any other tiktoken tokenizers being used, i assume setting it up is similar to the sentence piece tokenizers?

Llama3 Tokenizer is TikToken based! https://github.com/pytorch/torchtune/blob/main/torchtune/models/llama3/_tokenizer.py

joecummings:
Hiya @AndrewMead10 - just following up on this!

AndrewMead10:
Sorry for the delay on this, it ended up being a bit more than I initially expected lol. I have not had time to finish work on the tokenizer, and probably wont have time for at least the next week to work on it. I pushed the current code that I have for it if someone wants to finish it up. Currently I have gone through the initialization and special tokens for it, but from encode onwards the code is just AI gen placeholders and needs to be done properly. Everything still needs to be tested as well. I also havent gone super deep on the prompt template, so there may be missing spaces or something weird like that for the formatting of it (models tend to be super sensitive to those sorts of things).

@joecummings

PR #2359, opened 2025-02-07
Resume from checkpoint with distributed optimizer-in-backward repro

ebsmothers:
This is just a repro for the bug described in #2360


ebsmothers:
Closing this now that #2390 has landed

Issue #2211, opened 2024-12-28
[Feature] add `mlflow` metric_logging


steveepreston:
Please add `mlflow` metric_logging.
The lack of `mlflow` integration in `torchtune` is so painful.

ebsmothers:
@steveepreston thanks for creating the issue. Based on [this discussion](https://github.com/pytorch/torchtune/discussions/2103) seems like @fabiogeraci will take this on once he returns from leave

fabiogeraci:
Back to work on the 8/01, I will start then

steveepreston:
@ebsmothers Thanks for note
@fabiogeraci Awesome!

PR #2365, opened 2025-02-08
Implements MLFlowLogger

nathan-az:
#### Context
- [x] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

#### Changelog
Implements a basic MLFlowLogger in-line with the functionality of the other loggers.

#### Test plan

#### UX

----------

Note the MLFlowLogger's init was heavily inspired by the HF trainer's [MLflowCallback](https://github.com/huggingface/transformers/blob/v4.48.2/src/transformers/integrations/integration_utils.py#L1229). In-line with the other loggers, key identifiers (experiment name, run ID) can be overridden, but are often set in MLFlow via environment variables. 

Fixes https://github.com/pytorch/torchtune/issues/2211, interest shown in https://github.com/pytorch/torchtune/discussions/2103

#### Images

Logging the config shows, nested under the same path as the config output directory (this looks like it aligns to other loggers, but it's a bit ugly - do we want to keep it or should we always log in the artifact root as `torchtune_config.yaml`?)

image

Parameters and metrics show in the run overview, with nested parameters period-separated (e.g. `profiler.profile_memory`)

image

Metrics also display over-time in the Model Metrics tab according to the logging steps

image

### DIFF
--- a/torchtune/training/metric_logging.py
+++ b/torchtune/training/metric_logging.py
@@ -48,12 +48,30 @@ def save_config(config: DictConfig) -> Path:
         log.warning(f"Error saving config.\nError: \n{e}.")
 
 
-def flatten_dict(d, parent_key="", sep="."):
+def flatten_dict(d: Dict[str, Any], sep: str = ".", parent_key: str = ""):
+    """Recursively flattens a nested dictionary into one level of key-value pairs.
+
+    Args:
+        d (Dict[str, Any]): Any dictionary to flatten
+        sep (str): Desired separator for flattening nested keys
+        parent_key (str): Key prefix for children (nested keys), containing parent key
+            names
+
+    Example:
+        >>> flatten_dict({"foo": {"bar": "baz"}, "qux"; "quux"}, parent_key="--")
+        {"foo--bar": "baz", "qux", "quux"}
+
+    Returns:
+        Flattened dictionary
+
+    Note:
+        Does not unnest dictionaries within list values (i.e. {"foo": [{"bar": "baz}]}})
+    """
     items = []
     for k, v in d.items():
         new_key = f"{parent_key}{sep}{k}" if parent_key else k
         if isinstance(v, dict):
-            items.extend(flatten_dict(v, new_key, sep=sep).items())
+            items.extend(flatten_dict(v, sep=sep, parent_key=new_key).items())
         else:
             items.append((new_key, v))
     return dict(items)
--- a/torchtune/training/metric_logging.py
+++ b/torchtune/training/metric_logging.py
@@ -530,11 +530,9 @@ class MLFlowLogger(MetricLoggerInterface):
 
         self._mlflow = mlflow
 
-        self._tracking_uri = tracking_uri or os.getenv("MLFLOW_TRACKING_URI", None)
-        self._experiment_name = experiment_name or os.getenv(
-            "MLFLOW_EXPERIMENT_NAME", None
-        )
-        self._run_id = run_id or os.getenv("MLFLOW_RUN_ID", None)
+        self._tracking_uri = tracking_uri or os.getenv("MLFLOW_TRACKING_URI")
+        self._experiment_name = experiment_name or os.getenv("MLFLOW_EXPERIMENT_NAME")
+        self._run_id = run_id or os.getenv("MLFLOW_RUN_ID")
 
         if self.rank == 0:
             if not self._mlflow.is_tracking_uri_set():
--- a/torchtune/training/metric_logging.py
+++ b/torchtune/training/metric_logging.py
@@ -483,7 +483,7 @@ class CometLogger(MetricLoggerInterface):
 
 
 class MLFlowLogger(MetricLoggerInterface):
-    """Logger for use w/ MLFlow.
+    """Logger for use w/ MLFlow (https://mlflow.org/).
 
     Args:
         experiment_name (Optional[str]): MLFlow experiment name. If not specified, will
@@ -497,7 +497,7 @@ class MLFlowLogger(MetricLoggerInterface):
             to MLFLOW_RUN_ID environment variable if set, or a new run will be created.
 
     Example:
-        >>> logger = MLFlowLogger(experiment_name="my_experiment", run_name="run1", log_dir="./mlruns")
+        >>> logger = MLFlowLogger(experiment_name="my_experiment", run_name="run1")
         >>> logger.log("accuracy", 0.95, step=1)
         >>> logger.log_dict({"loss": 0.1, "accuracy": 0.95}, step=1)
         >>> logger.log_config(config)
--- a/torchtune/training/metric_logging.py
+++ b/torchtune/training/metric_logging.py
@@ -482,49 +482,6 @@ class CometLogger(MetricLoggerInterface):
         self.close()
 
 
-class MetricLoggerInterface(Protocol):
-    """Abstract metric logger."""
-
-    def log(
-        self,
-        name: str,
-        data: Scalar,
-        step: int,
-    ) -> None:
-        """Log scalar data.
-
-        Args:
-            name (str): tag name used to group scalars
-            data (Scalar): scalar data to log
-            step (int): step value to record
-        """
-        pass
-
-    def log_config(self, config: DictConfig) -> None:
-        """Logs the config as file
-
-        Args:
-            config (DictConfig): config to log
-        """
-        pass
-
-    def log_dict(self, payload: Mapping[str, Scalar], step: int) -> None:
-        """Log multiple scalar values.
-
-        Args:
-            payload (Mapping[str, Scalar]): dictionary of tag name and scalar value
-            step (int): step value to record
-        """
-        pass
-
-    def close(self) -> None:
-        """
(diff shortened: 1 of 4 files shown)


fabiogeraci:
Very similar approach but I found that self.rank pulls the local rank, prefer to log the experiment on global_rank and implement custom metrics for system log. But if the guys are happy good for me

nathan-az:
> I found that self.rank pulls the local rank, prefer to log the experiment on global_rank and implement custom metrics for system log

I considered this, but for the sake of uniformity aligned it to the other logger implementations.

> I think you might encounter issue in setting the eun_id when running multi GPUs as well as multi nodes multi GPUs.In my custom class I also included auto_log, with true/false, for user friendly approach. But if the guys are happy good for me.

Multi-GPU should be ok due ot the `self._mlflow.active_run()` checks, and only setting the run on rank 0.

Regarding multi-node, my solution generally is to have a setup script before my trainer which sets the `MLFLOW_RUN_ID` env var on all nodes. This means there is some redundant logging on each node, but at least it aligns and doesn't create more runs.

This approach has also been useful when using MLFlow as an artifact store for multi-node runs, where different nodes have different checkpoint ranks that need to be logged together.

fabiogeraci:
> Multi-GPU should be ok due ot the `self._mlflow.active_run()` checks, and only setting the run on rank 0.
> 
> Regarding multi-node, my solution generally is to have a setup script before my trainer which sets the `MLFLOW_RUN_ID` env var on all nodes. This means there is some redundant logging on each node, but at least it aligns and doesn't create more runs.
> 
> This approach has also been useful when using MLFlow as an artifact store for multi-node runs, where different nodes have different checkpoint ranks that need to be logged together.

Personally I prefer to let MlFlow setting the RUN_ID, simply because experiment_name and run_name can be the same and if the user forgets to change the run_id, i think MlFlow raises an error (I did not check, from memory).

I deployed my implementation on a internal server of my company, at the moment all user feed from the same front-end, therefore for me it is not an option to have redundant runs (under default), it would be confusing and messy, as well as spamming the logger database (the database team would not be happy). 

I am not a core developer therefore take this as my personal opinion

Ankur-singh:
@lulmer cc'ing you because you request support for MLFlow logging in last office hour.

fabiogeraci:
> @lulmer cc'ing you because you request support for MLFlow logging in last office hour.

I am not sure i understand!!

steveepreston:
@fabiogeraci I think he mentioned you because [your message](https://github.com/pytorch/torchtune/issues/2211#issuecomment-2568064850)

joecummings (commented):
This is a super quick turnaround - thanks! Can you also include some screenshots using the MLFlow logger from torchtune for additional "testing"?

Ankur-singh:
@lulmer (Louis Ulmer) requested MLFlow logger during the office hour. Hence tagged him to this PR, so he can follow the progress. 

@fabiogeraci & @steveepreston sorry for the confusion.

nathan-az:
No worries @joecummings - added some images! Let me know if we want to change the behaviour around the artifact naming/path. I tried to align it to the wandb logger, which just takes `output_config_fname.parent`.

My deeply nested absolute path is an unlikely case - I assume most users use relative paths for their config e.g. `configs/my_config.yaml`, which will show more cleanly.

That said, we can always just store in the artifact root `torchtune_config.yaml`. Happy to make any changes, and glad to see it's all working well 🙂

torchtune/training/metric_logging.py
```diff
@@ -469,3 +480,116 @@ def close(self) -> None:
...
+            raise ImportError(
+                "``mlflow`` package not found. Please install mlflow using `pip install mlflow` to use MLFlowLogger."
+                "Alternatively, use the ``StdoutLogger``, which can be specified by setting metric_logger_type='stdout'."
+            ) from e
+
+        _, self.rank = get_world_size_and_rank()
+
+        self._mlflow = mlflow
+
+        self._tracking_uri = tracking_uri or os.getenv("MLFLOW_TRACKING_URI", None)
```
joecummings, line 533: super nit: `os.getenv` should return None as the default, so no need to specify in these calls.

torchtune/training/metric_logging.py
```diff
@@ -469,3 +480,116 @@ def close(self) -> None:
...
+        self._experiment_name = experiment_name or os.getenv(
+            "MLFLOW_EXPERIMENT_NAME", None
+        )
+        self._run_id = run_id or os.getenv("MLFLOW_RUN_ID", None)
+
+        if self.rank == 0:
+            if not self._mlflow.is_tracking_uri_set():
+                if self._tracking_uri:
```
joecummings, line 541:
```suggestion
                if self._tracking_uri is not None:
```

torchtune/training/metric_logging.py
```diff
@@ -469,3 +480,116 @@ def close(self) -> None:
...
+                    self._mlflow.set_tracking_uri(self._tracking_uri)
+
+            if self._mlflow.active_run() is None or self._nested_run or self._run_id:
+                if self._experiment_name:
```
joecummings, line 545:
```suggestion
                if self._experiment_name is not None:
```

torchtune/training/metric_logging.py
joecummings: Also, what would happen if the tracking URI *is* None? Would the run fail? If so, maybe we should raise a descriptive error.

torchtune/training/metric_logging.py
joecummings: Same question here about expected behavior for a null experiment name.

torchtune/training/metric_logging.py
```diff
@@ -469,3 +480,116 @@ def close(self) -> None:
 
     def __del__(self) -> None:
         self.close()
+
+
+class MLFlowLogger(MetricLoggerInterface):
+    """Logger for use w/ MLFlow.
```
joecummings, line 486: Can you link to the MLFlow site here?

torchtune/training/metric_logging.py
```diff
@@ -48,6 +48,17 @@ def save_config(config: DictConfig) -> Path:
         log.warning(f"Error saving config.\nError: \n{e}.")
 
 
+def flatten_dict(d, parent_key="", sep="."):
```
joecummings, line 51: Please add a small comment here explaining the expected input format and output format. Also kwargs only after `d` param.

joecummings (commented):
Mostly minor comments!

torchtune/training/metric_logging.py
nathan-az: Added a proper docstring to align to `save_config` - let me know if it's not clear!

> Also kwargs only after d param.

I wasn't exactly sure what this meant. Did you mean to explicitly use `**kwargs` then parse the dict in the function?

For now I placed `sep` before `parent_key` since `sep` is more likely to be changed by a user (although setting `parent_key` may be useful if we ever wanted to prefix the config params with `torchtune_config` or similar)

torchtune/training/metric_logging.py
```diff
@@ -469,3 +480,116 @@ def close(self) -> None:
...
+        tracking_uri (Optional[str]): MLFlow tracking uri. If not specified, will default
+            to MLFLOW_TRACKING_URI environment variable if set, or default.
+        run_id (Optional[str]): MLFlow run name. If not specified, will default
+            to mlflow-generated HRID. Unused if run_id is specified or MLFLOW_RUN_ID
+            environment variable is found.
+        run_name (Optional[str]): MLFlow run ID. If not specified, will default
+            to MLFLOW_RUN_ID environment variable if set, or a new run will be created.
+
+    Example:
+        >>> logger = MLFlowLogger(experiment_name="my_experiment", run_name="run1", log_dir="./mlruns")
```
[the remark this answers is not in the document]
nathan-az: Good catch - removed.

torchtune/training/metric_logging.py
nathan-az: If `experiment_name` isn't set, MLFlow logs runs to a `Default` experiment.

I've don't use it since I prefer the experiment name groupings, but I think it's ok not to raise an error since it's handled behaviour. That said, I'm ambivalent.

torchtune/training/metric_logging.py
joecummings: Ahh, sorry I wasn't being clear. In general, for any optional params we like to use the wildcard operator to force the user to specify the param using keyword only. An example is like so:

```python
def foo(a: int, *, b: Optional[int] = None) -> int:
   ...
```

This forces the user to refer to the parameter `b` in function `foo` by name:

```python
foo(1, b=2)
```

Not only does this improve readability of the code, reduce the possibility for silent errors by sending parameters in with the wrong ordering, it also makes it easier to deprecate any kwargs that we might want to change down the line. So in this case, I'd like to see:

```python
def flatten_dict(d: Dict[str, Any], *, parent_key: str = "", sep: str = ".") -> Dict[str, Any]:
    ...
```

Does that make sense?

nathan-az: Ahh gotcha, love it, done.

joecummings (approved):
Awesome work - thanks! 🙏

Issue #2303, opened 2025-01-28
Move from PIL to torchvision.io.decode_image


ebsmothers:
Instead of using `PIL.open`, @NicolasHug has pointed out that we can move to `torchvision.io.decode_image` to speed up image processing by doing everything on pure tensors ([see also](https://pytorch.org/vision/main/io.html#image-decoding). This would also allow us to drop our explicit [PIL requirement](https://github.com/pytorch/torchtune/blob/main/pyproject.toml#L34)). This should entail:

1) changing [load_image](https://github.com/pytorch/torchtune/blob/23896c30dc10c97f8baaed8841e61bc3b8f2a61c/torchtune/data/_utils.py#L47) to use [torchvision.io.decode_image](https://pytorch.org/vision/main/generated/torchvision.io.decode_image.html#torchvision.io.decode_image)
2) updating [CLIPImageTransform](https://github.com/pytorch/torchtune/blob/23896c30dc10c97f8baaed8841e61bc3b8f2a61c/torchtune/models/clip/_transform.py#L26) to accept tensors instead of PIL.Image (in the short term we can keep PIL.Image support for backwards compatibility)

Ankur-singh:
@ebsmothers Can you please assign this to me? 

Thank you 😁

ebsmothers:
@Ankur-singh just assigned it to you, thanks for taking it on!

NicolasHug:
Feel free to ping me for help / reviews / best practices!

RdoubleA:
If we do this we should update all the documentation on Messages that image content is expected to be a tensor instead of a PIL image @ebsmothers @Ankur-singh

Ankur-singh:
@RdoubleA  Just `torch.Tensor` or `Union[torch.Tensor, PIL.Image]`?

Vision transformer as of now supports both.

pbontrager:
PIL was used as the only input type intentionally to enforce a consistent image API. A PIL image contains all the necessary meta data about an image to ensure the right transforms are being applied to an image (e.g. RGB channels vs BGR) whereas tensors don't have this information and we have to put the responsibility on the user to input the data in the expected format. If we're able to get that info from torchvision though that changes things.

Ankur-singh:
Keeping backward compatibility in mind, should be support PIL images as well? We can add a deprecation warning and drop PIL image support in the next release. 

OR we can drop the support right away and update all the docs to use `torch.Tensor`.

NicolasHug:
[...]

PR #2366, opened 2025-02-08
Refactor load_image to return torch.Tensor instead of PIL.Image

Ankur-singh:
#### Context
- [ ] add a new feature
- [ ] fix a bug
- [X] other (refactor)

Please link to any issues this PR addresses. #2303 

#### Changelog
* Updated `load_image` to return `torch.Tensor` instead of `PIL.Image`
* Updated `CLIPImageTransform` to support both `torch.Tensor` and `PIL.Image`
* Added 4x new tests for `CLIPImageTransform` with `torch.Tensor` input

#### Test plan

#### UX

### DIFF
--- a/torchtune/data/_utils.py
+++ b/torchtune/data/_utils.py
@@ -8,6 +8,8 @@ from pathlib import Path
 from typing import Any, Callable, Dict, List, Literal, Optional, TypeVar, Union
 from urllib import request
 
+import torch
+import torchvision
 from datasets import load_dataset
 from datasets.distributed import split_dataset_by_node
 from torch.utils.data import default_collate, DistributedSampler
@@ -44,9 +46,9 @@ def truncate(
     return tokens_truncated
 
 
-def load_image(image_loc: Union[Path, str]) -> "PIL.Image.Image":
+def load_image(image_loc: Union[Path, str]) -> "torch.Tensor":
     """
-    Convenience method to load an image in PIL format from a local file path or remote source.
+    Convenience method to load an image in torch.Tensor format from a local file path or remote source.
 
     Args:
         image_loc (Union[Path, str]): Local file path or remote source pointing to the image
@@ -59,7 +61,7 @@ def load_image(image_loc: Union[Path, str]) -> "PIL.Image.Image":
     Raises:
         ValueError:
             If the image cannot be loaded from remote source, **or**
-            if the image cannot be opened as a :class:`~PIL.Image.Image`.
+            if the image cannot be opened as a :class:`~torch.Tensor`.
 
     Examples:
         >>> # Load from remote source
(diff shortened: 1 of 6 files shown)


joecummings (commented):
cc @NicolasHug just for a quick glance.

torchtune/data/_utils.py
```diff
@@ -44,9 +46,9 @@ def truncate(
     return tokens_truncated
 
 
-def load_image(image_loc: Union[Path, str]) -> "PIL.Image.Image":
+def load_image(image_loc: Union[Path, str]) -> "torch.Tensor":
```
joecummings, line 49: You don't need to quote the return type as `torch.Tensor` is definitely already importable.

torchtune/data/_utils.py
```diff
@@ -69,25 +71,24 @@ def load_image(image_loc: Union[Path, str]) -> "PIL.Image.Image":
...
+            )
         except Exception as e:
-            raise ValueError(f"Failed to load image from {image_loc}") from e
-
-    # Open the local image as a PIL image
-    try:
-        image = Image.open(image_loc)
-    except Exception as e:
-        raise ValueError(f"Failed to open image as PIL Image from {image_loc}") from e
+            raise ValueError("Failed to load image as torch.Tensor") from e
```
joecummings, line 84: Can you be more specific with this error? Something like failed to load image from remote.

torchtune/data/_utils.py
```diff
@@ -69,25 +71,24 @@ def load_image(image_loc: Union[Path, str]) -> "PIL.Image.Image":
...
 
+    # Open the local image as a Tensor image
+    else:
+        try:
+            image = torchvision.io.decode_image(image_loc)
+        except Exception as e:
+            raise ValueError("Failed to open image as torch.Tensor") from e
```
joecummings, line 91: Same here, lets be specific that this fails to load from a local image.

torchtune/models/clip/inference/_transform.py
```diff
@@ -252,9 +252,13 @@ def __init__(
             antialias=self.antialias,
         )
 
-    def __call__(self, *, image: Image.Image, **kwargs) -> Mapping[str, Any]:
+    def __call__(
+        self, *, image: Union[Image.Image, torch.Tensor], **kwargs
```
joecummings, line 256: cc @pbontrager Do we actually want to allow PIL images through here? Or limit to just tensor?

joecummings (commented):
Couple of comments, but super happy to see this working!

NicolasHug:
> cc @NicolasHug just for a quick glance.

This SGTM

torchtune/data/_utils.py
```diff
@@ -69,25 +71,24 @@ def load_image(image_loc: Union[Path, str]) -> "PIL.Image.Image":
...
-    # TODO: Fix this
-    from PIL import Image
-
     # If pointing to remote source, try to load to local
     if isinstance(image_loc, str) and image_loc.startswith("http"):
         try:
-            image_loc = request.urlopen(image_loc)
+            image_loc = request.urlopen(image_loc).read()
+            image = torchvision.io.decode_image(
+                torch.frombuffer(image_loc, dtype=torch.uint8)
```
NicolasHug, line 81: Chatting with @pbontrager about this, I think you'll want to ensure that you get RGB images from `decode_image` by passing `mode="RGB"`. https://pytorch.org/vision/main/generated/torchvision.io.ImageReadMode.html#torchvision.io.ImageReadMode

This is typically needed for images that are encoded as grayscale, or have some transparency channel.

(Here and below)

torchtune/models/clip/inference/_transform.py
Ankur-singh: hi @pbontrager, any comments?

Ankur-singh:
Sorry guys, I was using python 3.11 and everything seemed to work. It was a typing issue, learnt that in python 3.9 one can't use subscripted generics like `Union[. . .]` in runtime checks. I was using 

```python
if isinstance(x, Union[Image.Image, torch.Tensor]):
    ... 
```

I change it to 

```python
if isinstance(x, (Image.Image, torch.Tensor)):
    ... 
```

this seems to work in both python 3.9 and 3.11. 

> Also, any suggestions on how to manage multiple python versions? What does the workflow look like? Docker, uv, etc.

Ankur-singh:
@ebsmothers its read for review. Should work 🤞

joecummings (approved):
lgtm

joecummings:
Thanks for all the hard work @Ankur-singh !!!@

PR #2367, opened 2025-02-09
feat: Added cfg.cudnn_deterministic_mode flag

### LINKED ISSUE #2335: Seed is not applied for DPO recipes
# TL;DR

Launching same config twice with `seed: 42` results in two different loss curves

Image

# Affected recipes

### `full_dpo_distributed` - seed is not set

Full DPO is taken from #2275 

Image

### `lora_dpo_distributed` - seed is not set

Image

# Not affected recipes

### `full_finetune_distributed` - works fine

Image

### `lora_finetune_distributed` - works fine

Image

bogdansalyp:
#### Context
- [x] add a new feature
- [x] fix a bug -> closes #2335
- [ ] other (please add here)

#### Changelog

`cfg.deterministic_mode` added for reproducibility

#### Test plan

#### UX

### DIFF
--- a/recipes/dev/early_exit_finetune_distributed.py
+++ b/recipes/dev/early_exit_finetune_distributed.py
@@ -234,7 +234,7 @@ class EarlyExitFinetuneRecipeDistributed(FTRecipeInterface):
         # These are public properties which are updated by the checkpoint loader
         # when ``resume_from_checkpoint`` is `True` or validated in tests
         self.seed = training.set_seed(
-            seed=cfg.seed, debug_mode=cfg.get("deterministic_mode", None)
+            seed=cfg.seed, debug_mode=cfg.get("cudnn_deterministic_mode", None)
         )
         self.epochs_run = 0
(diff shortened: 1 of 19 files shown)


bogdansalyp:
image

pbontrager (commented):
This looks good, my only concern is that the name is too generic as this only affects cudnn backends. Maybe cuda_deterministic_mode?

bogdansalyp:
> This looks good, my only concern is that the name is too generic as this only affects cudnn backends. Maybe cuda_deterministic_mode?

hmmmm it's not just CUDA but all [torch randomness](https://pytorch.org/docs/stable/notes/randomness.html) settings
ideally you also use this `local_seed` for Generators as well

bogdansalyp:
> This looks good, my only concern is that the name is too generic as this only affects cudnn backends. Maybe cuda_deterministic_mode?

agree, that may be a better naming

pbontrager (approved):

PR #2368, opened 2025-02-09
fix: torch and torchvision import check

### LINKED ISSUE #2369: Misleading import error message for torchao
`torch`, `torchao`, and `torchvision` are not a part of `pyproject.toml`. According to [docs](https://github.com/pytorch/torchtune/blob/main/torchtune/__init__.py#L12), it's done because "it is not currently possible to properly support both nightly and stable installs of PyTorch + torchao in pyproject.toml."

Small issue is that if `torchao` is installed and `torch` is not, the error still says `torchao` is not installed. May waste some time for developers who encounter with this issue.

Also, there's no check for the third package, `torchvision`.

bogdansalyp:
#### Context
- [ ] add a new feature
- [x] fix a bug - closes #2369 
- [ ] other (please add here)

#### Changelog

Additional check added that `torch` and `torchvision` packages are installed with corresponding error messages.

#### Test plan

#### UX

### DIFF
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -4,11 +4,21 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
-# Check at the top-level that torchao is installed.
+# Check at the top-level that torch, torchao, and torchvision are installed.
 # This is better than doing it at every import site.
 # We have to do this because it is not currently possible to
 # properly support both nightly and stable installs of PyTorch + torchao
 # in pyproject.toml.
+try:
+    import torch  # noqa
+except ImportError as e:
+    raise ImportError(
+        """
+        torch not installed.
+        Please follow the instructions at https://pytorch.org/torchtune/main/install.html#pre-requisites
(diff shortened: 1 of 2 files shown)


ebsmothers (commented):
Thanks for creating the PR. I agree that the try/except ImportError just on torchao was a bit too much of a hammer to solve this specific problem (and I am to blame.. if you haven't already, you can see the backstory in #1452). So I think checking the torch import first makes just so that we can give a more helpful error message. 

I am on the fence about doing the same for torchvision though -- it is actually not a hard dependency for many of our components (e.g. `from torchtune import training` should succeed even if you don't have torchvision installed). The main reason for taking this approach with ao is that it's pretty tightly-integrated into our low-level components, and it's a painful experience to only hit the import error when you're ready train a model or something. (E.g. the same import will fail without ao installed because of our LoRA components.) For torchvision I claim it's actually OK to not be so aggressive about it -- it's only relevant for those who are training multimodal models, and most components should work without it, so we don't need to push installation up front to the same degree.

Issue #2052, opened 2024-11-22
improve _optim_ckpt_wrapper to its a drop in replacement of optimizer


felipemello1:
Our recipes are cluttered with logic that checks "if optim_in_bwd".

With a bit of engineering, we can make it a drop in replacement of optimizer, and avoid code like [this](https://github.com/pytorch/torchtune/blob/a9aadf5d24696fa8790c52739f8edd5ebd298881/recipes/full_finetune_distributed.py#L772):

```
if not self._optimizer_in_bwd:
    self._optimizer.zero_grad()
else:
    for opt in self._optim_ckpt_wrapper.optim_map.values():
        opt.zero_grad()
```

That can be replaced with:
```
class MyOptWrapper:
	def __init__(self, optimizers):
		self.optimizers = optimizers

	def zero_grad():
		for opt in self.optimizers.optim_map.values():
        	opt.zero_grad()

optimizer = MyOptWrapper(optimizers)
optimizer.zero_grad()
```

It may break things from time to time, but good testing should avoid errors hitting prod. For overly complex situations, we can still do if/else, but we definitely don't need every if/else that we have today: A total of 8.

felipemello1:
> wouldn't it require all optimizers to be wrapped in this regardless of if they're using optimizer in bwd or not?

I dont think so. The implementation would be something like this:

```
optimizer = config.instantiate(my_opt)
if opt_in_bwd:
	optimizer = MyOptWrapper(optimizer)
```

The idea is that, for example, when you call `optimizer.zero_grad()`, it doesn't need to know if its the optimizer or the wrapper, because the wrapper behave like the optimizer.

krammnic:
So the point is to avoid all conditions about `opt_in_bwd` except checkpointing one?

PR #2370, opened 2025-02-09
[RFC]: Get rid of optim_bwd checks via wrapper.

krammnic:
#2052 

We have lots of similar codes related to theoptim_bwd  optimizer setup in our recipes. Basically, we can create a wrapper/abstraction of the basic optimizer, where we will put all checks, making the recipes cleaner. One of the possible ways of doing such a cleanup is presented in this RFC. Basically, we should discuss the balance between abstraction and simplicity of the given solution. Would love to hear some comments on it.

### DIFF
--- a/torchtune/training/_optimizer_wrapper.py
+++ b/torchtune/training/_optimizer_wrapper.py
@@ -77,7 +77,7 @@ class OptimizerWrapper:
         else:
             return get_lr(self._optimizer)
         
-    def set_lr(self, lr_scheduler):
+    def set_learning_rate_scheduler
         if self._optimizer_in_bwd:
             self._optim_ckpt_wrapper.set_lr_scheduler(lr_scheduler)
 
--- a/torchtune/training/_optimizer_wrapper.py
+++ b/torchtune/training/_optimizer_wrapper.py
@@ -1,0 +1,118 @@
+from typing import Any, Dict, Optional
+
+from omegaconf import DictConfig
+
+from torch import nn
+from torchtune import config, training
+from torchtune.training.lr_schedulers import get_lr
+
+
+class OptimizerWrapper:
+    """
+    Abstraction on optimizer to get rid of _optimizer_in_bwd checks.
+    """
+
+    def __init__(
+        self,
+        model: nn.Module,
+        cfg_optimizer: DictConfig,
+        optimizer_in_bwd: bool = False,
+        opt_state_dict: Optional[Dict[str, Any]] = None,
+    ):
+        self._optimizer_in_bwd = optimizer_in_bwd
+        self._model = model
+
+        if self._optimizer_in_bwd:
+            self._optim_dict = {
+                param: config.instantiate(cfg_optimizer, [param])
+                for param in self._model.parameters()
+            }
+            training.register_optim_in_bwd_hooks(model=self._model, optim_dict=self._optim_dict)
+            self._optim_ckpt_wrapper = training.create_optim_in_bwd_wrapper(
+                model=self._model, optim_dict=self._optim_dict
+            )
+            if opt_state_dict is not None:
+                for param in opt_state_dict.keys():
+                    try:
+                        training.load_from_full_optimizer_state_dict(
+                            self._model,
+                            self._optim_ckpt_wrapper.state_dict()[param],
+                            opt_state_dict[param],
+                            self._device,
+                        )
+                    except BaseException as e:
+                        raise RuntimeError(
+                            "Failed loading in-backward optimizer checkpoints."
+                            "Please make sure run being restored from was using in-backward optimizer."
+                        ) from e
+        else:
+            # Standard setup
+            self._optimizer = config.instantiate(cfg_optimizer, self._model.parameters())
+            if opt_state_dict:
+                training.load_from_full_optimizer_state_dict(
+                    self._model,
+                    self._optimizer,
+                    opt_state_dict,
+                    self._device,
+                )
+
+    def zero_grad(self) -> None:
+        if self._optimizer_in_bwd:
+            for opt in self._optim_ckpt_wrapper.optim_map.values():
+                opt.zero_grad()
+        else:
+            self._optimizer.zero_grad()
+
+    def step(self) -> None:
+        if self._optimizer_in_bwd:
+            # No explicit step needed here.
+            pass
+        else:
+            self._optimizer.step()
+
+    def get_lr(self) -> float:
+        if self._optimizer_in_bwd:
+            # Return the learning rate of the first optimizer in the wrapper
+            return get_lr(next(iter(self._optim_ckpt_wrapper.optim_map.values())))
+        else:
+            return get_lr(self._optimizer)
+        
+    def set_lr(self, lr_scheduler):
+        if self._optimizer_in_bwd:
+            self._optim_ckpt_wrapper.set_lr_scheduler(lr_scheduler)
+
+
+    def state_dict(self) -> Dict[str, Any]:
+        if self._optimizer_in_bwd:
+            return self._optim_ckpt_wrapper.state_dict()
+        else:
+            return self._optimizer.state_dict()
+    
+    def get_optimizer(self) -> tuple:
+        if self._optimizer_in_bwd:
+            return next(iter(self._optim_ckpt_wrapper.optim_map.values()))
+        else:
+            return self._optimizer
+
+    def load_state_dict(self, state_dict: Dict[str, Any]) -> None:
+        """
+        Load optimizer state dict.
+
+        Args:
+            state_dict (Dict[str, Any]): optimizer state_dict.
+        """
+        if self._optimizer_in_bwd:
+            for param in state_dict.keys():
+                training.load_from_full_optimizer_state_dict(
+                    self._model,
+                    self._optim_ckpt_wrapper.state_dict()[param],
+                    state_dict[param],
+                    self._device,
+                )
+        else:
+            training.load_from_full_optimizer_state_dict(
+                self._model,
+                self._optimizer,
+                state_dict,
+                self._device,
+            )
--- a/recipes/full_finetune_distributed.py
+++ b/recipes/full_finetune_distributed.py
@@ -25,6 +25,7 @@ from torchtune.data import padded_collate_packed
 from torchtune.datasets import ConcatDataset
 from torchtune.recipe_interfaces import FTRecipeInterface
 from torchtune.training import DummyProfiler, PROFILER_KEY
+from torchtune.training._optimizer_wrapper import OptimizerWrapper
 from torchtune.training.activations import apply_selective_activation_checkpointing
 from torchtune.training.checkpointing._checkpoint_client import (
     CheckpointClient,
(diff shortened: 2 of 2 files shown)


krammnic:
@felipemello1 Hey! Require review. I haven't found a better design at the point of the wrapper.

SalmanMohammadi:
Hey @krammnic. We're currently seperately working through a separate issue with our optimizer-in-backward feature. I wonder if it'd be best to investigate those fixes first? cc @ebsmothers

krammnic:
@SalmanMohammadi Hey! Sure, will check it out.

felipemello1 (commented):
hey Mark, thanks for this PR! At first glance, i love it :)

There are a few things that worry me. Let me know your thoughts about them:

1. Initially when i thought about the opt_in_bwd wrapper, was for it to mimic the expected behavior of the optimizer. But in this PR, we are also wrapping the regular optimizer, for example:

```
def state_dict(self) -> Dict[str, Any]:
        if self._optimizer_in_bwd:
            return self._optim_ckpt_wrapper.state_dict()
        else:
            return self._optimizer.state_dict()
```

My fear is that we are adding layers on something that may not need layers, hurting hackability.

On the other hand, i like how you removed a lot of utilities/boilerplate from the recipe and added here.

Q: should we wrap the optimizer?

2. I believe that in the training loop, the optimizer is called differently when its opt_in_bwd and when it isnt. This happens because of grad_clipping and normalizing loss by number of tokens.

check lines 814, 825, 868: https://github.com/pytorch/torchtune/blob/main/recipes/full_finetune_distributed.py#L814

Q: How should we handle it? Probably still check `if self._optimizer_in_bwd`?

3) Some recipes dont have optimizer_in_bwd, like RL and lora recipes. It may be worth sanity checking that using this implementation will also work for them. But I am ok with doing it in a follow up PR.

torchtune/training/_optimizer_wrapper.py
joecummings, line 1: Can we have this live in memory with the other memory optimizations?

torchtune/training/_optimizer_wrapper.py
```diff
@@ -0,0 +1,118 @@
...
+            self._optimizer.zero_grad()
+
+    def step(self) -> None:
+        if self._optimizer_in_bwd:
+            # No explicit step needed here.
+            pass
+        else:
+            self._optimizer.step()
+
+    def get_lr(self) -> float:
```
joecummings, line 73: I'm in favor of slightly more descriptive names like `get_learning_rate`

torchtune/training/_optimizer_wrapper.py
```diff
@@ -0,0 +1,118 @@
...
+        if self._optimizer_in_bwd:
+            # Return the learning rate of the first optimizer in the wrapper
+            return get_lr(next(iter(self._optim_ckpt_wrapper.optim_map.values())))
+        else:
+            return get_lr(self._optimizer)
+        
+    def set_lr(self, lr_scheduler):
```
joecummings, line 80:
```suggestion
    def set_learning_rate_scheduler
```

krammnic:
> hey Mark, thanks for this PR! At first glance, i love it :)
> 
> There are a few things that worry me. Let me know your thoughts about them:
> 
> 1. Initially when i thought about the opt_in_bwd wrapper, was for it to mimic the expected behavior of the optimizer. But in this PR, we are also wrapping the regular optimizer, for example:
> 
> ```
> def state_dict(self) -> Dict[str, Any]:
>         if self._optimizer_in_bwd:
>             return self._optim_ckpt_wrapper.state_dict()
>         else:
>             return self._optimizer.state_dict()
> ```
> 
> My fear is that we are adding layers on something that may not need layers, hurting hackability.
> 
> On the other hand, i like how you removed a lot of utilities/boilerplate from the recipe and added here.
> 
> Q: should we wrap the optimizer?
> 
> 2. I believe that in the training loop, the optimizer is called differently when its opt_in_bwd and when it isnt. This happens because of grad_clipping and normalizing loss by number of tokens.
> 
> check lines 814, 825, 868: https://github.com/pytorch/torchtune/blob/main/recipes/full_finetune_distributed.py#L814
> 
> Q: How should we handle it? Probably still check `if self._optimizer_in_bwd`?
> 
> 3. Some recipes dont have optimizer_in_bwd, like RL and lora recipes. It may be worth sanity checking that using this implementation will also work for them. But I am ok with doing it in a follow up PR.

@felipemello1 Thanks for the answer! My opinion that there is no problem to make point 3 done during this PR, it is actually pretty small procedure, isn't it? (Correct me if I'm missing something, please). Speaking about the design overal... To be honest, for me it is hard to choose either. For instance, I love how some parts of the recipes has gone in this optimizer abstraction. But it is abstraction! Therefore, it is a complication. Will think more about it.

torchtune/training/_optimizer_wrapper.py
krammnic: Sure!

felipemello1:
lets get the approval for one (or two recipes, maybe a LoRA one). And when everyone is ok with it, we can look towards updating the others, so it minimizes the work for you. Thanks again!

torchtune/training/_optimizer_wrapper.py
krammnic: <del>Meanwhile, `lr: 5e-8` in the recipes.</del> Yeah, sounds reasonable!

pbontrager (commented):
I left some comments and then realized that we already have OptimizerInBackwardWrapper. I should have started with that. But can't we just extend OptimizerInBackwardWrapper to have all the safety checks that are currently inside of the recipe instead of wrapping it in an additional layer?

felipemello1:
Here is what i am understanding:

Step 1: It sounds like @pbontrager thinks we should try to keep only a single wrapper and avoid wrapper or wrappers. I agree. If the opt_in_bwd can mimic the optimizer API, then its a drop in for any optimizer. I do not think it will work to inherit the optimizer class though :/

Step 2: Decide how to implement aux functions that are not part of the optimizer API, e.g. checkpoint related functions. Maybe those can just be utilities instead of class? Not sure.

Step 3: update the rest of the recipe (lines 814, 825, 868: [main/recipes/full_finetune_distributed.py#L814](https://github.com/pytorch/torchtune/blob/main/recipes/full_finetune_distributed.py#L814))

What do you think?

torchtune/training/_optimizer_wrapper.py
```diff
@@ -0,0 +1,118 @@
...
+            self._optim_ckpt_wrapper.set_lr_scheduler(lr_scheduler)
+
+
+    def state_dict(self) -> Dict[str, Any]:
+        if self._optimizer_in_bwd:
+            return self._optim_ckpt_wrapper.state_dict()
+        else:
+            return self._optimizer.state_dict()
+    
+    def get_optimizer(self) -> tuple:
```
felipemello1, line 91: where is this used?

torchtune/training/_optimizer_wrapper.py
```diff
@@ -0,0 +1,118 @@
...
+            self._optimizer.step()
+
+    def get_lr(self) -> float:
+        if self._optimizer_in_bwd:
+            # Return the learning rate of the first optimizer in the wrapper
+            return get_lr(next(iter(self._optim_ckpt_wrapper.optim_map.values())))
+        else:
+            return get_lr(self._optimizer)
+        
+    def set_learning_rate_scheduler
```
felipemello1, line 80: where is this used?

torchtune/training/_optimizer_wrapper.py
```diff
@@ -0,0 +1,118 @@
...
+        else:
+            return self._optimizer.state_dict()
+    
+    def get_optimizer(self) -> tuple:
+        if self._optimizer_in_bwd:
+            return next(iter(self._optim_ckpt_wrapper.optim_map.values()))
+        else:
+            return self._optimizer
+
+    def load_state_dict(self, state_dict: Dict[str, Any]) -> None:
```
felipemello1, line 97: this seems to be the same code as in __init__, should we just call it instead?

torchtune/training/_optimizer_wrapper.py
```diff
@@ -0,0 +1,118 @@
+from typing import Any, Dict, Optional
+
+from omegaconf import DictConfig
+
+from torch import nn
+from torchtune import config, training
+from torchtune.training.lr_schedulers import get_lr
+
+
+class OptimizerWrapper:
```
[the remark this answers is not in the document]
felipemello1: I dont think that inheriting would help. Even for regular optimizer, we still need it when `resuming_from_checkpoint` so we can call:

```
training.load_from_full_optimizer_state_dict(
                    model,
                    optimizer,
                    opt_state_dict,
                    self._device,
                )
```

torchtune/training/_optimizer_wrapper.py
```diff
@@ -0,0 +1,118 @@
...
+    """
+    Abstraction on optimizer to get rid of _optimizer_in_bwd checks.
+    """
+
+    def __init__(
+        self,
+        model: nn.Module,
+        cfg_optimizer: DictConfig,
```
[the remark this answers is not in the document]
felipemello1: We use it like this:

```
param: config.instantiate(cfg_optimizer, [param]) for param in model.parameters()
or
param: config.instantiate(cfg_optimizer, model.parameters())
```

we also pass a DictConfig to `setup_optimizer`

Can you share how you are thinking about it?

```
self._optimizer = OptimizerWrapper(
            model=self._model,
            cfg_optimizer=<INSERT HERE>,
            optimizer_in_bwd=self._optimizer_in_bwd,
            opt_state_dict=(
                checkpoint_dict[training.OPT_KEY]
                if self._resume_from_checkpoint
                else None
            ),
        )
 ```

felipemello1:
Here is what i am understanding:

a) There are optimizer specific functions (zero grad, step, state_dict)
b) There are utilities / boiler plate that needs to call the optimizer differently based on opt_in_bwd (get_lr, load_state_dict, set_learning_rate_scheduler)

For (a), i feel like these should belong to `training.create_optim_in_bwd_wrapper`, so that it has the same API as other optimizers.

For (b), IMO your `OptimizerWrapper` makes sense: It removes from the recipe the burden of if/else, and puts it in a separate class that handles it, working for both regular optimizer and opt_in_bws

@krammnic @pbontrager  do you agree?

@krammnic , when you have time, can you also address the other parts of the recipe (lines 814, 825, 868: [main/recipes/full_finetune_distributed.py#L814](https://github.com/pytorch/torchtune/blob/main/recipes/full_finetune_distributed.py#L814))

krammnic:
@felipemello1 Yep, reasonable comments! Let me fix then

krammnic:
Speaking about `optimizer_wrapper.optimizer.zero_grad()` or `optimizer_wrapper.zero_grad()` I assume that common `.zero_grad()` without `.optimizer` is less confusing.

krammnic:
@pbontrager Hey Philip! I reformat this as an RFC and would love to hear your comments on desired design for this (except the changes the @felipemello1 already mentioned). I assume that such clean up should be done, but it mostly depends on chosen design.

PR #2384, opened 2025-02-11
Implement step based checkpointing

joecummings:
- **Add helper functions for pruning old checkpoints**
- **Step based checkpointing and tests**

#### Context
- [x] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

Closes #2105. This is a widely requested feature that allows users to have greater control over checkpointing frequency in torchtune. 

TODO: Add commentary on design decisions. Acknowledge spaghetti code. Beg forgiveness.

#### Changelog
* Update ``FullModelHFCheckpointer`` to accept a step parameter when saving a checkpoint. Use that step to designate the checkpoint folder name. Keep ``epoch_{}`` as a fall-back for BC.
* Add helper function ``prune_surplus_checkpoints`` that removes extra checkpoints in order to free up space for the user. This is controlled by the ``keep_n_last_checkpoints``.
* Add helper function ``get_all_checkpoints_in_dir`` that scans a directory (meant for the output_dir) and returns all checkpoint directories.
* Add tests for corresponding helper functions
* Modify the ``full_finetune_single_device.py`` recipe to utilize step-based checkpointing.

#### Test plan

**Ran single device recipe w/ step-based checkpointing**
*Evidence of succesful run*

*Evidence of correct number of checkpoints being saved*

#### UX

### DIFF
--- a/recipes/full_finetune_single_device.py
+++ b/recipes/full_finetune_single_device.py
@@ -23,6 +23,7 @@ from torchtune.data import padded_collate_packed
 from torchtune.datasets import ConcatDataset
 from torchtune.recipe_interfaces import FTRecipeInterface
 from torchtune.training import DummyProfiler, PROFILER_KEY
+from torchtune.training.checkpointing.constants import CURR_STEP_KEY
 from torchtune.training.lr_schedulers import get_lr
 
 from tqdm import tqdm
@@ -139,6 +140,7 @@ class FullFinetuneRecipeSingleDevice(FTRecipeInterface):
 
         # Training cfg
         self._resume_from_checkpoint = cfg.resume_from_checkpoint
+        self.save_every_n_steps = cfg.get("save_every_n_steps")
         self._gradient_accumulation_steps = cfg.gradient_accumulation_steps
         self._optimizer_in_bwd = cfg.optimizer_in_bwd
         self._clip_grad_norm = cfg.get("clip_grad_norm", None)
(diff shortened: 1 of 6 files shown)


bogdansalyp:
- `recipe_state` is still saved to `${output_dir}`, not `${output_dir}/step_XXX`
- `resume_from_checkpoint` logic should be updated
  - RN it looks for `${output_dir}` root, not `step_XXX`
  - maybe replace top level `cfg.resume_from_checkpoint` to have `cfg.checkpointer.resume_from` which is either "latest" (default) or the path to the checkpoint to resume from. Or separate mutually exclusive `resume_from: /path/` and `resume_from_latest: True`
  - offtopic but `cfg.resume_from_checkpoint` is mentioned in code as deprecated and replaced by `should_load_recipe_state` but de facto `resume_from_checkpoint` is mandatory and `should_load_recipe_state` doesn't work
- lr schedulers don't work when resuming from `recipe_state`
- step and epoch are recovered properly from `recipe_state` but not used to update `current_step` - as a result logging breaks and

recipes/full_finetune_single_device.py
```diff
@@ -627,14 +663,20 @@ def train(self) -> None:
                             log_dict["grad_norm"] = grad_norm
                         self._metric_logger.log_dict(log_dict, step=self.global_step)
 
+                    # Save checkpoint if specified by user
+                    if self.global_step % self.save_every_n_steps == 0:
+                        self.save_checkpoint(
+                            epoch=curr_epoch, step=self.global_step, full_tensors=False
```
felipemello1, line 669: var "step" is missing in distributed recipe

bogdansalyp:
rebased to #2869

PR #2389, opened 2025-02-12
check if log_dir is not none

felipemello1:
#### Context
- [ ] add a new feature
- [x] fix a bug
- [ ] other (please add here)

if log_dir was not passed ,it would crash. This makes log_dir optonal

#### Test plan

```
tune run full_finetune_single_device --config llama3_2/1B_full_single_device dataset.packed=True tokenizer.max_seq_len=512 dataset.split=train[:5%] metric_logger=torchtune.training.metric_logging.WandBLogger ~metric_logger.log_dir
```

Before this change:
```
TypeError: stat: path should be string, bytes, os.PathLike or integer, not NoneType
```

After this change:
It works!

### DIFF
--- a/torchtune/training/metric_logging.py
+++ b/torchtune/training/metric_logging.py
@@ -243,7 +243,7 @@ class WandBLogger(MetricLoggerInterface):
         self.log_dir = kwargs.pop("dir", log_dir)
 
         # create log_dir if missing
-        if self.log_dir and not os.path.exists(self.log_dir):
+        if self.log_dir is not None and not os.path.exists(self.log_dir):
             os.makedirs(self.log_dir)
 
         _, self.rank = get_world_size_and_rank()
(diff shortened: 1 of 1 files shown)


torchtune/training/metric_logging.py
```diff
@@ -243,7 +243,7 @@ def __init__(
         self.log_dir = kwargs.pop("dir", log_dir)
 
         # create log_dir if missing
-        if not os.path.exists(self.log_dir):
+        if self.log_dir and not os.path.exists(self.log_dir):
```
joecummings, line 246: is not None

joecummings: A new test would be nice ...

felipemello1: `if self.log_dir`  works, no? Why add "is not None"? readability or something else?

boooo for the test

joecummings: If there's a possibility it can be None, then the proper check is `if x is not None`

joecummings: I'm fine with filing a follow up for test :)

joecummings (approved):
stampy stamp

Issue #2405, opened 2025-02-18
Tool Calling Fine-Tuning Fails because of validation logic in messages


init27:
Hi Team, 

First, please correct me if my understanding is wrong or if I missed something. 

However, I believe the validation logic in torch.data.messages [here](https://github.com/pytorch/torchtune/blob/main/torchtune/data/_messages.py#L634) doesn't account for tool calling flow. 

I've a dataset which follows the following flow: system -> user -> assistant -> tool -> assistant -> user and so on and so forth.

I've setup the correct role mappings for this dataset and also tried extending `SFTDataset` by creating a Transform which ensures that `eot=False` logic is followed for a) Assistant call before Tool call, b) Tool Call 

However both fail the validation. 

Please let me know if I missed something super obvious. 

Thanks!

RdoubleA:
Thanks for flagging this. This function is quite outdated and definitely needs to be updated to account for tool calling logic. At a quick glance it looks like the `last_turn` logic needs to account for potential tool calls that will involve a non user role in between assistant messages.

musabgultekin:
I had the same problem before and fixed it locally.

Just opened a quick PR for that support: https://github.com/pytorch/torchtune/pull/2407

pbontrager:
@init27 does #2407 resolve this issue if it lands?

init27:
@pbontrager thanks and thanks @musabgultekin for quick PR! Actually I need some clarification from the team: 

Context: When we do tool call, a single flow is:

>  (step-1) user (with ask requiring tool call) -> (step-2) Assistant (responds with tool call arguments) -> (step-3) Tool (responds with Tool response) -> (step-4) Assistant (answers user query)

Now, my Q is regarding the `eot` token-which denotes if the turn of one role (assistant/user/system is over)

For the above flow, do we set `eot=False` in both (step-2) and (step-3) or is it `False` JUST in (step-3)? 

From @musabgultekin's PR: 
> `eot=False if message["role"] in ["tool", "ipython"] else True`

My understanding is `eot` should ALSO be set to `False` when Assistant is returning the tool call, step-2 in above example (before the tool call happens)?

cc: @RdoubleA

musabgultekin:
@init27 

`eot` should be true when assistant is calling a custom tool.

See:
> While builtin-tool calls end with <|eom_id|>, notice the <|eot_id|> for zero-shot function calls.
[...]

PR #2407, opened 2025-02-18
Tool role support

musabgultekin:
#### Context
role=tool not supported. This PR adds support for OpenAI messages dataset

Related Issue: https://github.com/pytorch/torchtune/issues/2405

#### Changelog
Added role=tool
Set eot=False if the role is tool or ipython

#### Test plan

Ran pre-commit checks.
Added tests both for validation function and OAI transformation function

#### UX

 I did not change any public API

### DIFF
--- a/torchtune/data/_messages.py
+++ b/torchtune/data/_messages.py
@@ -626,7 +626,7 @@ class OpenAIToMessages(Transform):
                     role=message["role"],
                     content=content,
                     masked=masked,
-                    eot=False if message["role"] in ["tool", "ipython"] else True,
+                    eot=False if message["role"] in ["tool"] else True,
                 ),
             )
 
--- a/torchtune/data/_messages.py
+++ b/torchtune/data/_messages.py
@@ -613,7 +613,8 @@ class OpenAIToMessages(Transform):
                     role="system", content=self.new_system_prompt, masked=True, eot=True
                 )
             )
-        for message in sample[self._column_map["messages"]]:
+        messages = sample[self._column_map["messages"]]
+        for i, message in enumerate(messages):
             if message["role"] == "system" and self.new_system_prompt is not None:
                 continue
             masked = (message["role"] != "assistant") and (not self.train_on_input)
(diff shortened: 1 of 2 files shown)


torchtune/data/_messages.py
```diff
@@ -625,6 +626,7 @@ def __call__(self, sample: Mapping[str, Any]) -> Mapping[str, Any]:
                     role=message["role"],
                     content=content,
                     masked=masked,
+                    eot=False if message["role"] in ["tool"] else True,
```
ebsmothers, line 629: Also commented on the issue [here](https://github.com/pytorch/torchtune/issues/2405#issuecomment-2670102451), but is this definitely the condition we want?

torchtune/data/_messages.py
```diff
@@ -658,9 +660,9 @@ def validate_messages(
 
     last_turn = "assistant"
     for i, message in enumerate(messages):
-        if message.role == "assistant" and last_turn != "user":
+        if message.role == "assistant" and last_turn not in ["user", "tool"]:
```
RdoubleA, line 663: if we still support ipython and tool then we need to also include ipython in here

torchtune/data/_messages.py
```diff
@@ -658,9 +660,9 @@ def validate_messages(
...
             raise ValueError(
-                f"Assistant message before expected user message at index {i} in messages"
+                f"Assistant message before expected user or tool message at index {i} in messages"
```
RdoubleA, line 665: I think we will need to add another check to make sure tool returns always come after assistant messages with `ipython=True` (not the role, the attribute in the message)

RdoubleA (commented):
Was discussing with @Ankur-singh and @pbontrager that we should also begin replacing all mentions of `role=ipython` with `role=tool` and deprecate ipython if we feel that tool is more clear and explicit. Perhaps doesn't have to be in this PR but defintiely before we close #2405.

I'm also questioning the value of `validate_messages`. It is a bit strict and users have complained about it in the past. My assessment is that restricting double user messages or assistant messages is kinda loose these days, and some frameworks enforce not having doubles or just coalesce doubles together for you. Don't know if others have thoughts on if we want to remove some of these checks.

torchtune/data/_messages.py
RdoubleA: yeah assistant message with `ipython=True` should have `eot=False` as well

ebsmothers:
@musabgultekin are you still interesting in finishing this one up? Let us know if you don't have bandwidth and we can push some changes to take it over the finish line

torchtune/data/_messages.py
musabgultekin: Done. Added `if message.role in ["tool", "ipython"] and not last_message.ipython: raise ValueError`

torchtune/data/_messages.py
musabgultekin: Added

torchtune/data/_messages.py
musabgultekin: Added next message conditions.

musabgultekin:
Made all updates and added comments for the reasoning behind the decisions. I kept the "ipython" role. But as Rafi mentioned, we could handle removing `ipython` role in another PR

musabgultekin:
@ebsmothers Would you mind checking it again if it looks good

joecummings (approved):
Will make sure tests pass, but this LGTM

joecummings:
@musabgultekin Can you check out these failing tests?

musabgultekin:
Tests passes now sorry about that. Thanks!

PR #2410, opened 2025-02-18
Add support for ``StatefulDataLoader``

joecummings:
#### Context
- [x] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

This PR adds support for the `StatefulDataLoader` class from the PyTorch library torchdata.

**FAQs**
1. **Why?** This is necessary for resuming from checkpoints mid-epoch - a feature set we will need to support for step-based checkpointing.
2. **Only full finetune single device?** Yeah, this is more to see how we'll integrate this as a POC and make sure that the tests can pass. Once this is merged, I'll add support for the rest of the recipes. 
3. **Hardcoding iterator_finished?** Yeah, we'll change this when we _actually_ move to step-based checkpointing, but right now we expect to save on the epoch boundaries so if the epoch is cut short, then we expect that the dataloader will restart its shuffling and data provided as if the iterator has finished going through all the samples. Huge, huge thanks to @ramanishsingh for helping me debug this last issue.
4. **You removed the sampler?!!** The `StatefulDataLoader` creates a batched random sampler of it's own so there's no need for us to be creating a new one in this context. Less code means my life is easier.
5. **WTH, you removed the check for max_steps is None. Won't that break?** Nope, Python is dumb. Check it:
```
>>> 1 == None
False
>>> 0 == None
False
>>> 5 == None
False
>>> 1000000000000000 == None
False
```

#### Changelog
* Import `StatefulDataloader` and replace it's usage in the `_setup_data` method
* Checkpoint the dataloader state dict
* Load the dataloader state dict if we are trying to resume from the checkpoint
* Update tests to match (yes the numbers changed b/c we are using a new random state - it's fine)

#### Test plan

#### UX

### DIFF
--- a/recipes/full_finetune_single_device.py
+++ b/recipes/full_finetune_single_device.py
@@ -770,10 +770,8 @@ class FullFinetuneRecipeSingleDevice(FTRecipeInterface):
 
                 # Check if we should stop training for this epoch
                 if (
-                    self.max_steps_per_epoch is not None
-                    and (idx + 1 // self._gradient_accumulation_steps)
-                    == self.max_steps_per_epoch
-                ):
+                    (idx + 1) // self._gradient_accumulation_steps
+                ) == self.max_steps_per_epoch:
                     break
 
             self.epochs_run += 1
--- a/recipes/full_finetune_single_device.py
+++ b/recipes/full_finetune_single_device.py
@@ -554,11 +554,11 @@ class FullFinetuneRecipeSingleDevice(FTRecipeInterface):
         batch_size: int,
         collate_fn: str,
         dataloader_state_dict: Optional[Dict[str, Any]] = None,
-    ) -> Tuple[DistributedSampler, DataLoader]:
+    ) -> StatefulDataLoader:
         """
-        All data related setup happens here. Currently this recipe only supports the
-        DistributedSamplers with Map-style Datasets which fit into memory. Other samplers,
-        iterable datasets and streaming datasets are not supported.
+        All data related setup happens here. This recipe currently supports only
+        map-style datasets. If a state_dict is provided (meaning we are resuming a training run),
+        it is loaded into the dataloader.
         """
         if isinstance(cfg_dataset, ListConfig):
             datasets = [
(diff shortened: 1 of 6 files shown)


recipes/full_finetune_single_device.py
```diff
@@ -302,11 +302,16 @@ def setup(self, cfg: DictConfig) -> None:
...
         # setup after both of these are initialized
         collate_name = cfg.get("collate_fn", "torchtune.data.padded_collate_sft")
-        self._sampler, self._dataloader = self._setup_data(
+        self._dataloader = self._setup_data(
             cfg_dataset=cfg.dataset,
             shuffle=cfg.shuffle,
             batch_size=cfg.batch_size,
             collate_fn=collate_name,
+            dataloader_state_dict=(
+                ckpt_dict[training.DATALOADER_KEY]
```
RdoubleA, line 311: should check if this key even exists for BC

tests/recipes/test_full_finetune_single_device.py
```diff
@@ -56,7 +56,7 @@ def _get_test_config_overrides(self):
 
     def _fetch_expected_loss_values(self, model_type):
         loss_values_map = {
-            "llama2": [10.5201, 10.5217, 10.4945, 10.5136],
+            "llama2": [10.5130, 10.5215, 10.5764, 10.5046],
```
RdoubleA, line 59: is this because the sampler is different??

recipes/full_finetune_single_device.py
```diff
@@ -570,19 +576,11 @@ def _setup_data(
             raise RuntimeError("left_pad_sequence collator is only for inference.")
         collate_fn = _get_component_from_path(collate_fn)
 
-        sampler = DistributedSampler(
-            ds,
-            num_replicas=1,
-            rank=0,
-            shuffle=shuffle,
-            seed=0,
```
RdoubleA, line 578: so how is seed passed to StatefulDataLoader in this case?

recipes/full_finetune_single_device.py
joecummings: This might just have to break BC b/c without this, the user will not be able to successfully resume training at any point.

SalmanMohammadi (approved):

PR #2414, opened 2025-02-20
Simplify early fusion behavior

ebsmothers:
This is a very WIP PR to take the state dict hooks out of our early fusion. 

Why? [These two lines](https://github.com/pytorch/torchtune/blob/a65d23caf448ccad0261e6738f67539dbade1ab7/torchtune/modules/model_fusion/_early_fusion.py#L105-L106) patching the decoder's token embeddings into EarlyFusion are fundamentally incompatible with the ability to call `load_state_dict` directly on `EarlyFusion`'s `decoder`. This is something we claim to support for our multimodal models, so we need to fix it. 

I discussed a few different ways we could support this with @pbontrager. Unfortunately all approaches besides this one require some extra indirection or trickery. I claim doing the dumb/obvious thing is the most amenable to maintainable and debuggable code.

**Question:** Why are you adding another argument to `TransformerDecoder`?

I know we do not want to be one of those libraries with 20 different indecipherable forward parameters. And we also really try not to have conditionals in our forward passes. In this case I claim this is the lesser of several evils. `EarlyFusion` needs to be easy to use and flexible for loading different subsets of params, and the current implementation is not. Separately, we are (very belatedly) addressing a pretty common feature request that we received in the past -- the ability to bypass the token embeddings in our transformer implementation.

## Test plan

Updated tests in `test_early_fusion.py`. Added a new test case to `test_transformer_decoder.py`.

```
pytest tests/torchtune/modules/test_transformer_decoder.py
...
=================== 14 passed in 8.97s ==========
```

```
pytest tests/torchtune/modules/model_fusion/test_early_fusion.py
...
================= 10 passed in 0.10s ==========
```


joecummings (commented):
tests

RdoubleA:
+++++++1, down with state dict hooks

joecummings (commented):
@ebsmothers  to pick this up again soon.

torchtune/modules/transformer.py
```diff
@@ -502,29 +502,40 @@ def chunked_output(self, last_hidden_state: torch.Tensor) -> List[torch.Tensor]:
...
         Raises:
             ValueError:
+                If neither tokens nor input_embeds are passed **or**
                 If seq_len of x is bigger than max_seq_len, **or**
                 if the model has caches which have been setup with self-attention layers and ``mask`` is not provided, **or**
                 if the model has caches which have been setup with encoder layers and ``encoder_mask`` is not provided, **or**
                 if the model has caches which have been setup ``input_pos`` is not provided.
         """
 
+        if tokens is None and input_embeds is None:
```
RdoubleA, line 531: Should also check if both are not none - only one should be passed in

RdoubleA (approved):
So elegant

joecummings (approved):
He's a master of his craft

torchtune/modules/transformer.py
ebsmothers: Lol @joecummings too fast on the trigger finger merged before I even saw this. Can do as an easy follow-up

Issue #2344, opened 2025-02-05
Discussion: Update dataloader to skip rows that dont require training


felipemello1:
https://github.com/pytorch/torchtune/issues/2341

When a) `train_on_input=False` and b) message is too long that output is truncated, there may be a batch without trainable tokens, raising an error on the loss because of division by zero.

Beyond raising an inconvenient bug, this is a waste of compute.

In the dataloader, should we skip rows that does have trainable embeddings?

RdoubleA:
It's difficult to know if this is the case fro the DataLoader perspective. You would have to catch this and raise/skip in the recipe or add additional preprocessing to your dataset to prevent this. We don't have the ability to skip an item within the `SFTDataset`'s `__getitem__`.

felipemello1:
I see. It could be added to the recipe, like "check_batch_requires_grad(batch)". I am sure @ebsmothers will love the idea!

EugenHotaj:
Maybe slightly orthogonal but it would also be great to expose a way to do left truncation instead of right truncation. I think this is even more important once the "mask all previous turns" strategy is introduced.

(I'm actually a bit surprised to see that torchtune supports left truncation only since I thought right truncation was the standard practice for llama models).

pocca2048:
> You would have to catch this and raise/skip in the recipe

I agree. Actually, I implemented skipping logic in the recipe w/o fixing the loss.

PR #2419, opened 2025-02-21
[RFC] truncation and skipping

krammnic:
#2344 Mention two important points related to our data loading and processing. This RFC works on both of these aspects.

### Truncation

Currently, we don't support truncation in both right and left directions. Unfortunately, this is not the best strategy, as both of the truncation types might be relevant in different situations (assume that important information in the sample is presented in further context, then right truncation is better and vice versa). First of all, will show losses for both experiments (Qwen2.5 3B, LoRA, 2048 max_seq_len, `packed: True`). 

Right truncation:

W B Chart 2_21_2025, 10_06_52 PM

Left truncation:

W B Chart 2_21_2025, 9_53_48 PM

We see that with left truncation we have better convergence. Notice, that it happens because of the dataset structure, not because that it is always better to have such truncation!

Therefore, there are 2 options:

1. `truncation_type` as tokenizer option in config;
 
Pros:
Adaptiveness to datasets with different sample structures.

Cons:
One more argument in config!

2. `truncation_type` is default to right/left

Pros and Cons are opposite to 1. 

I like the idea of setting `truncation_type: "left"` by default without exposing it to all configs, but saving possibility to change truncation type to left.

### Skipping 

It is simple check directly in recipe, the only possible question is - where to put the `check_batch_requires_grad` function.


torchtune/data/_utils.py
```diff
@@ -24,6 +24,7 @@ def truncate(
     tokens: List[Any],
     max_seq_len: int,
     eos_id: Optional[Any] = None,
+    truncation_type: str = "right",
```
RedTachyon, line 27: This seems to change the default behavior - is that intended? My intuition is that it's better to not break backwards compatibility unless necessary when introducing new options like this.

krammnic: Yep, it should be left

recipes/lora_finetune_single_device.py
```diff
@@ -629,6 +629,13 @@ def save_checkpoint(self, epoch: int) -> None:
             intermediate_checkpoint=intermediate_checkpoint,
             adapter_only=self._save_adapter_weights_only,
         )
+    
+    def check_batch_requires_grad(self, batch: dict) -> bool:
+        labels = batch.get("labels")
+        if labels is None or not isinstance(labels, torch.Tensor):
+            return False  # No labels, no training
+    
+        return (labels != -100).any().item()
```
felipemello1, line 638: I may be wrong, but i dont think that this is correct.

We would have to do something like:
```
requires_grad = any(batch["labels"] != self._loss_fn.ignore_index)
```

this requires an unit test

torchtune/data/_utils.py
```diff
@@ -34,13 +35,27 @@ def truncate(
...
     """
-    tokens_truncated = tokens[:max_seq_len]
-    if eos_id is not None and tokens_truncated[-1] != eos_id:
+    if truncation_type not in ("left", "right"):
+        raise ValueError(f"truncation_type must be 'left' or 'right', got {truncation_type}")
+    
+    if truncation_type == "right":
+        tokens_truncated = tokens[-max_seq_len:]  # Take the last max_seq_len tokens
+    else:
+        tokens_truncated = tokens[:max_seq_len]  # Take the first max_seq_len tokens
```
felipemello1, line 53: maybe change it to:

```
if truncation_type == "right":
elif truncation_type == "left":
else: raise ....
```

torchtune/data/_utils.py
```diff
@@ -34,13 +35,27 @@ def truncate(
         max_seq_len (int): maximum length of the list
         eos_id (Optional[Any]): token to replace the last token with. If None, the
             last token will not be replaced. Default is None.
+        truncation_type (str): type of truncation to apply, either "left" or "right".
+            Default is "right".
```
felipemello1, line 39: i find it a bit confusing what truncate left and right is. My understand is that truncate "right" is x[:max_seq_len], but you coded it the opposite.

torchtune/models/qwen2_5/_tokenizer.py
```diff
@@ -186,9 +190,9 @@ def tokenize_messages(
         # Finally, truncate if necessary
         if self.max_seq_len:
             tokenized_messages = truncate(
-                tokenized_messages, self.max_seq_len, self.eos_id if add_eos else None
+                tokenized_messages, self.max_seq_len, self.eos_id if add_eos else None, truncation_type=self.truncation_type
             )
-            mask = truncate(mask, self.max_seq_len, True if add_eos else None)
+            mask = truncate(mask, self.max_seq_len, True if add_eos else None, truncation_type=self.truncation_type)
```
felipemello1, line 195: it would be easier to understand if we used key arguments here, eg:

```
truncate(arg1=arg1, arg2=arg2, etc)
```

torchtune/data/_utils.py
```diff
@@ -34,13 +35,27 @@ def truncate(
...
+    if truncation_type not in ("left", "right"):
+        raise ValueError(f"truncation_type must be 'left' or 'right', got {truncation_type}")
+    
+    if truncation_type == "right":
+        tokens_truncated = tokens[-max_seq_len:]  # Take the last max_seq_len tokens
+    else:
+        tokens_truncated = tokens[:max_seq_len]  # Take the first max_seq_len tokens
+    
+    # Replace the last token with eos_id if necessary
+    if eos_id is not None and tokens_truncated and tokens_truncated[-1] != eos_id:
```
felipemello1, line 56: why do we need 'and tokens_truncated` if we didnt need it before?

recipes/lora_finetune_single_device.py
```diff
@@ -629,6 +629,13 @@ def save_checkpoint(self, epoch: int) -> None:
             intermediate_checkpoint=intermediate_checkpoint,
             adapter_only=self._save_adapter_weights_only,
         )
+    
+    def check_batch_requires_grad(self, batch: dict) -> bool:
```
felipemello1, line 633: maybe rename to 'check_has_trainable_tokens'

recipes/lora_finetune_single_device.py
felipemello1: We also shouldnt hardcode -100. Lets use self._loss_fn.ignore_index and pass as an arg.

recipes/lora_finetune_single_device.py
felipemello1: instead of passing a dict, which can be anything, lets just pass only what we need, which is the Optional[torch.tensor] = batch.get("labels")

felipemello1: food for thought: we could check if ANY ROW has missing trainable tokens. But this increases complexity and we dont want to remove the rows anyway because it would mess up with compile. A warning could be nice, but then it increases code complexity.

recipes/lora_finetune_single_device.py
felipemello1: Lets add a warning in this utility

We can use [log_once](https://github.com/pytorch/torchtune/blob/cf0142bd2eb5eabe33a3d43eca4f07ec48a8749f/torchtune/utils/_logging.py#L42). The warning can be something like this: "Found batch with no trainable tokens. Consider changing to tokenizer.truncation='left'".

felipemello1 (commented):
hey, thanks for the PR1 Left a few comments.

I think that we should keep the default the same, for bc.
You said that left truncation is better, but the graph labeled "right truncation" looks better.

Just to confirm, which graph do you think is better, the top one or the bottom one?

lets make the changes, i will give a last pass, then lets add unit tests and add to all recipes. Let me know if it sounds like a good plan.

recipes/lora_finetune_single_device.py
krammnic: Hmm, I though that it is current, after some manual testing, but let's go with your approach then.

felipemello1: Sorry, i made a mistake in my first comment. I didnt see "(labels != -100).any().item()". The way you did works fine.

but we should still add the log and not hardcode -100.

recipes/qat_lora_finetune_distributed.py
```diff
@@ -781,6 +782,22 @@ def save_checkpoint(
...
+            labels (Optional[torch.tensor]): labels for the current batch.
+
+        Returns:
+            bool: True if there are trainable tokens in batch, otherwise False.
+        """
+        log_once(
+            log,
+            "Found batch with no trainable tokens. Consider changing tokenizer.truncation direction!",
+        )
+        return any(labels != self._loss_fn.ignore_index)
```
felipemello1, line 799: this should be an utility, and ignore_index should be an argument

recipes/full_dpo_distributed.py
```diff
@@ -838,6 +839,22 @@ def concatenated_forward(
...
+
+        Args:
+            labels (Optional[torch.tensor]): labels for the current batch.
+
+        Returns:
+            bool: True if there are trainable tokens in batch, otherwise False.
+        """
+        log_once(
+            log,
+            "Found batch with no trainable tokens. Consider changing tokenizer.truncation direction!",
```
felipemello1, line 854: I would remove the "!", maybe do: "Consider changing to tokenizer.truncation="left" or increase tokernizer.max_seq_len."

recipes/full_dpo_distributed.py
```diff
@@ -892,7 +876,7 @@ def train(self) -> None:
 
             pbar = tqdm(total=self._steps_per_epoch, disable=not (self.rank == 0))
             for idx, batch in enumerate(self._dataloader):
-                if not self.check_has_trainable_tokens(batch):
+                if not check_has_trainable_tokens(batch.get("labels"), self._loss_fn.ignore_index, log):
```
felipemello1, line 879: can we call it using argnames? fun(arg1=arg1, arg2=arg2), and make the log optional

tests/torchtune/data/test_data_utils.py
```diff
@@ -17,24 +17,23 @@ def test_truncate():
...
-    
+
     # Test right truncation
-    truncated_masks = truncate(tokens=masks, max_seq_len=4, eos_id=False, truncation_type="right")
-    
+    truncated_masks = truncate(
+        tokens=masks, max_seq_len=4, eos_id=False, truncation_type="left"
+    )
+
     assert truncated_masks == [True, False, True, False]
```
felipemello1, line 36: n00b question: why does [True, True, False, True, False] become [True, True, False, False]?

Are we sure that in right truncation the outcome should be [True, False, True, False]?

nit: the comment is says "right truncation", but the command is "truncation_type=left"

torchtune/modules/transforms/tokenizers/_utils.py
```diff
@@ -195,3 +199,25 @@ def parse_hf_tokenizer_json(tokenizer_json_path: str) -> Dict[str, int]:
         tokenizer_json = json.load(f)
 
     return {token["content"]: token["id"] for token in tokenizer_json["added_tokens"]}
+
+
+def check_has_trainable_tokens(
+    labels: Optional[torch.tensor], ignore_index: int, log: Optional[logging.Logger]
+) -> bool:
```
felipemello1, line 206: two nits:

lets rename to 'has_trainable_tokens'

and lets remove the log from the args. We can add this to the top of the file:

```
from torchtune.utils import get_logger
_log: logging.Logger = get_logger()
```

felipemello1 (commented):
cool, thank you! Just left one last comment about `check_has_trainable_tokens`

Also had one comment on the unit testing. After that, its good to be added to every recipe/tokenizer, run tests, merge

torchtune/modules/transforms/tokenizers/_utils.py
krammnic: oh there is one more nit

tests/torchtune/data/test_data_utils.py
krammnic: We have eos_id and:
```
if eos_id is not None and tokens_truncated and tokens_truncated[-1] != eos_id:
        tokens_truncated[-1] = eos_id
```

torchtune/modules/transforms/tokenizers/_utils.py
```diff
@@ -195,3 +202,28 @@ def parse_hf_tokenizer_json(tokenizer_json_path: str) -> Dict[str, int]:
...
+    Returns:
+        bool: True if there are trainable tokens in batch, otherwise False.
+    """
+    if labels is None:
+        return False
+    
+    if not any(labels != ignore_index):
+        log_once(
+            _log,
+            'Consider changing to tokenizer.truncation="left" or increase tokernizer.max_seq_len.',
```
felipemello1, line 226:
```suggestion
            'Consider changing to tokenizer.truncation="left" or increasing tokenizer.max_seq_len.',
```

torchtune/models/qwen2_5/_tokenizer.py
```diff
@@ -75,6 +75,8 @@ class Qwen2_5Tokenizer(Qwen2Tokenizer):  # noqa: N801
             large for long running processes (esp. for texts of language that do not use space between
             word, e.g. Chinese); technically not a memory leak but appears as one.
             By default, we set the cache size equals to size of the official Qwen2 tokenizer.
+        truncation_type (str): type of truncation to apply, either "left" or "right".
+            Default is "right".
```
felipemello1, line 79: please double check if every tokenizer has docstring and exposes the arg

recipes/full_dpo_distributed.py
```diff
@@ -875,6 +876,12 @@ def train(self) -> None:
 
             pbar = tqdm(total=self._steps_per_epoch, disable=not (self.rank == 0))
             for idx, batch in enumerate(self._dataloader):
+                if not has_trainable_tokens(
+                    labels=batch.get("labels"),
+                    ignore_index=self._loss_fn.ignore_index,
```
felipemello1, line 881: i am ok with letting this stay like this, but its possible that some instantiated loss doesnt have ignore_index. We could do something like `self._loss_fn.ignore_index is hasatrr(self._loss_fn, 'ignore_index') else -100`. This would be more robust.

recipes/full_finetune_distributed.py
```diff
@@ -732,6 +733,22 @@ def _setup_data(
 
         return sampler, dataloader
 
+    def check_has_trainable_tokens(self, labels: Optional[torch.tensor]) -> bool:
```
felipemello1, line 736: this has to be deleted from every recipe

recipes/full_finetune_distributed.py
```diff
@@ -760,6 +777,9 @@ def train(self) -> None:
 
             pbar = tqdm(total=self._steps_per_epoch, disable=not self._is_rank_zero)
             for idx, batch in enumerate(self._dataloader):
+                if not self.check_has_trainable_tokens(batch):
```
felipemello1, line 780: this has to be updated on every recipe

tests/torchtune/modules/transforms/tokenizers/test_utils.py
```diff
@@ -55,3 +59,35 @@ def test_tokenize_no_special_tokens(self, tokenizer, messages, add_bos, add_eos)
             assert tokens[-1] == tokenizer.eos_id
         else:
             assert tokens[-1] != tokenizer.eos_id
+
+class TestHasTrainableTokens:
```
felipemello1, line 63: looks good

torchtune/models/qwen2_5/_tokenizer.py
felipemello1: what i meant is that other tokenizers dont have it. I think you only did it for qwen2_5

torchtune/data/_utils.py
```diff
@@ -24,6 +24,7 @@ def truncate(
     tokens: List[Any],
     max_seq_len: int,
     eos_id: Optional[Any] = None,
+    truncation_type: str = "left",
```
felipemello1, line 27:
```suggestion
    truncation_type: str = "right",
```

torchtune/models/llama3/_tokenizer.py
```diff
@@ -322,10 +322,18 @@ def tokenize_messages(
             mask = mask + [True]
 
         if self.max_seq_len:
-            tokens = truncate(
-                tokens, self.max_seq_len, self.eos_id if add_end_tokens else None
+            tokenized_messages = truncate(
+                tokens=tokenized_messages,
```
felipemello1, line 326:
```suggestion
            tokens = truncate(
                tokens=tokens,
```

krammnic:
Screenshot 2025-02-26 at 01 44 37
DPO LoRA single device qwen3b 20 steps, truncation_type: right

krammnic:
Screenshot 2025-02-26 at 01 53 18

DPO LoRA single device qwen3b 20 steps, truncation_type: left

krammnic:
Gemma2b, lora distributed, truncation_type: "right"
<details>
Step 1 | loss:26.85302734375 lr:2.0000000000000003e-06 tokens_per_second_per_gpu:209.4905242919922 peak_memory_active:5.214685440063477 >
Step 2 | loss:27.898555755615234 lr:4.000000000000001e-06 tokens_per_second_per_gpu:399.8436279296875 peak_memory_active:5.3756670951843>
Step 3 | loss:26.375633239746094 lr:6e-06 tokens_per_second_per_gpu:589.7933959960938 peak_memory_active:5.548027515411377 peak_memory_a>
Step 4 | loss:25.86777687072754 lr:8.000000000000001e-06 tokens_per_second_per_gpu:457.12811279296875 peak_memory_active:5.3763642311096>
Step 5 | loss:23.593978881835938 lr:1e-05 tokens_per_second_per_gpu:502.33099365234375 peak_memory_active:5.376002788543701 peak_memory_>
Step 6 | loss:20.89067268371582 lr:1.2e-05 tokens_per_second_per_gpu:652.4540405273438 peak_memory_active:5.844363689422607 peak_memory_>
Step 7 | loss:20.04551887512207 lr:1.4e-05 tokens_per_second_per_gpu:482.68695068359375 peak_memory_active:5.376347064971924 peak_memory>
Step 8 | loss:17.218624114990234 lr:1.6000000000000003e-05 tokens_per_second_per_gpu:551.7702026367188 peak_memory_active:6.474055767059>
Step 9 | loss:17.941665649414062 lr:1.8e-05 tokens_per_second_per_gpu:546.103515625 peak_memory_active:5.376846790313721 peak_memory_all>
Step 10 | loss:16.08778953552246 lr:2e-05 tokens_per_second_per_gpu:485.8663024902344 peak_memory_active:5.376261234283447 peak_memory_a>
Step 11 | loss:15.600129127502441 lr:1.9999998828397348e-05 tokens_per_second_per_gpu:557.3656616210938 peak_memory_active:5.37738084793>
Step 12 | loss:14.559475898742676 lr:1.999999531358965e-05 tokens_per_second_per_gpu:532.0314331054688 peak_memory_active:5.375916957855>
Step 13 | loss:13.859354019165039 lr:1.9999989455577743e-05 tokens_per_second_per_gpu:672.8965454101562 peak_memory_active:5.41393947601>
Step 14 | loss:12.657991409301758 lr:1.999998125436299e-05 tokens_per_second_per_gpu:654.568115234375 peak_memory_active:5.4139394760131>
Step 15 | loss:11.866594314575195 lr:1.9999970709947322e-05 tokens_per_second_per_gpu:413.97149658203125 peak_memory_active:5.3771395683>
Step 16 | loss:11.371137619018555 lr:1.9999957822333203e-05 tokens_per_second_per_gpu:567.3087158203125 peak_memory_active:5.37627840042>
Step 17 | loss:11.43862247467041 lr:1.9999942591523652e-05 tokens_per_second_per_gpu:556.8408203125 peak_memory_active:5.376760959625244>
Step 18 | loss:10.980806350708008 lr:1.9999925017522245e-05 tokens_per_second_per_gpu:574.8773193359375 peak_memory_active:5.37643289566>
Step 19 | loss:10.352394104003906 lr:1.999990510033309e-05 tokens_per_second_per_gpu:548.7708740234375 peak_memory_active:5.375951290130>
Step 20 | loss:10.502054214477539 lr:1.9999882839960865e-05 tokens_per_second_per_gpu:683.5306396484375 peak_memory_active:6.06643533706>
  </details>

krammnic:
Gemma 2b, lora_distributed, truncation: "left"

W B Chart 2_26_2025, 2_11_47 AM

krammnic:
W B Chart 2_26_2025, 2_41_05 AM
W B Chart 2_26_2025, 2_45_17 AM

krammnic:
Reverified dpo: it saving correctly, so if recipe will fail again - it is llama2 specific.

felipemello1 (approved):

felipemello1:
awesome work, thanks for implementing this!

345ishaan:
@felipemello1 this feature seems like did not get pick in latest stable release (i.e. 0.6.1)

felipemello1:
@pbontrager

Issue #2421, opened 2025-02-22
GRPO Improvement checklist


RedTachyon:
It's alive! It's alive! (#2326)

The recipe definitely works (as in, I can run it and reach like a 60% success rate on GSM8k with a 3B model), but it's somewhat barebones and underoptimized. Here, I want to keep track of all the most important features and improvements that I think are missing. I'll probably go through this at some point, at some pace, but if anyone else wants to contribute - you can grab something from this list.

## Improvements
- [ ] Figure out how to move things from dev into the main repository (decide on the final APIs etc.)
- [ ] A proper eval workflow, so that every once in a while we run a full eval on the test set. Alternatively, a separate evaluation recipe? (note: I actually have a working separate eval recipe that works with the same paradigm as GRPO, goes through the full test dataset, computes success/reward and saves it to a file inside the checkpoint - happy to make it a PR)
- [ ] Adding proper (unit) tests, compliant with the normal torchtune testing workflow
- [ ] Adding proper documentation to everything
- [ ] Refactoring of the GRPO losses (there's also a research-y question of what loss should be used - there's some ambiguity in the paper and reference implementations)
- [ ] More modular approach to reward computation (probably as a component with the regular OmegaConf setup)
- [ ] Step-based checkpointing (#2384) - this is pretty important, since one epoch can be very long, leading to very infrequent checkpoints
- [ ] Memory profiling and experiments on "controlled" hardware (a recipe tuned to work ~optimally on a node of 8xH100, or on a single H100, or on smaller hardware with e.g. LoRA)
- [ ] Optimization of the default recipe - maybe we can get a big performance boost e.g. by doing `ppo_epochs>1`?
- [ ] Dataset improvements (gsm8k is functional but could use some polish, there's also the MATH and DeepscaleR datasets with a similar structure that can be added)
- [ ] A single-device version - should be pretty simple, but probably also slow. Might require gradient accumulation to properly work, I tend to get bad results with small batch sizes.
- [ ] Try to improve generation speed by using vLLM (or something else)?
- [ ] Probably more to be found soon

## Bugs
Because of course I found a bug right after everything was finalized. There will likely be more, so this subsection might or might not be useful.
[...]

PR #2422, opened 2025-02-22
[WIP] GRPO datasets

krammnic:
According to #2421, the addition of 2 datasets, MATH and DeepScaleR is required. Unfortunately, there are some problems with the MATH dataset: https://huggingface.co/datasets/huggingface-legal/takedown-notices/blob/main/2025/2025-01-02-AoPS.md Therefore, it might not be the best idea to add, due to some ethical reasons, but it is still the object of the discussion.


krammnic:
UPD: MATH dataset from "EleutherAI/hendrycks_math" is added, but it still might be banned in future.

Issue #2263, opened 2025-01-13
adding support for LR schedule for full distributed finetune


tginart:
My understanding is that the full multi-gpu fine-tuning doesn't yet support learning rate schedules.

Would it be possible to add support for this? Even basic ones, such as linear warmup follow by cosine or linear decay?

I can also take a look at doing this myself if I could get a pointer to the code. 

Thank you!

felipemello1:
hey @tginart ,

We have it in our recipes here: https://github.com/pytorch/torchtune/blob/6923833cc70f6792049964939e5a61c859f2cba9/recipes/full_finetune_distributed.py#L341

It gets called in the training loop here: https://github.com/pytorch/torchtune/blob/6923833cc70f6792049964939e5a61c859f2cba9/recipes/full_finetune_distributed.py#L798

The issue seems that we don't have it exposed in some configs. I will fix that. You can add this to your config, and it should work:

```
lr_scheduler:
  _component_: torchtune.training.lr_schedulers.get_cosine_schedule_with_warmup
  num_warmup_steps: 100
```

Thanks for flagging that!

tginart:
Thank you!

Ankur-singh:
@tginart thanks for pointing it out!

> The issue seems that we don't have it exposed in some configs. I will fix that. You can add this to your config, and it should work:

@felipemello1, I apologize if I'm overstepping here, but I think this could be part of a bigger issue. For someone familiar with the codebase (at least the recipes), it's clear how to adapt the config to their needs. They can easily see the different components used in the recipe and the various arguments allowed for each one. However, for others, this process isn’t as obvious since it’s not documented. I believe documenting this "esoteric" knowledge would greatly improve the accessibility of the library for new users, while also offering more flexibility for advanced users.

felipemello1:
hey @Ankur-singh , i agree! We try to expose as much as possible through our configs and add comments so users feel ok exploring them. We also document most of our performance flags and have them in the readme. But i 100% agree that there is room to improve by a lot. This LR scheduler is an example of something that we added to the recipe and forgot to expose in the configs.

If you feel that there is something else in specific that we should prioritize adding, or some tutorial, please feel free to suggest or make a PR.

ionutmodo:
@felipemello1 I am reviving this thread with an error I encountered. I am trying to do full-finetuning-distributed with cosine learning rate with warmup using the default AdamW optimizer from pytorch and I get the following error (see below).

[...]

Issue #2064, opened 2024-11-25
Add LR scheduler to PPO recipe


SalmanMohammadi:

Seoley:
Hi @SalmanMohammadi,

I attempted to solve this issue and made some modifications to `ppo_full_finetune_single_device.py` I would like to verify whether my code is working correctly.

To ensure this, do I only need to pass the three tests (Unit test, Recipe test, Regression test)? If there are any additional steps I should take, I would appreciate your guidance on contributing to this issue.

Thanks in advance!

pbontrager:
Thank you for taking this on! You can push a draft PR and share it here so we can give better feedback. Ideally your changes should look similar to the LR scheduler solution in [SFT single device](https://github.com/pytorch/torchtune/blob/504cbea31e9a4f222deaf379a7292a0f7af895e7/recipes/full_finetune_single_device.py#L330). For testing you should match any tests that exist for it for the SFT version and include a sample run with it and show a similar or better loss curve than with it turned off.

Seoley:
@pbontrager Thank you for your advice!
I followed the procedures you suggested to validate the model using an A100, but I encountered the following issue.

```
  File "/root/torchtune/recipes/ppo_full_finetune_single_device.py", line 1277, in recipe_main
    recipe.train()
  File "/root/torchtune/recipes/ppo_full_finetune_single_device.py", line 1007, in train
    trajectory = self.generate_trajectory_batched(batch)
  File "/root/torchtune/recipes/ppo_full_finetune_single_device.py", line 967, in generate_trajectory_batched
    trajectories.append(self.generate_trajectory(batch_input_ids))
  File "/root/torchtune/recipes/ppo_full_finetune_single_device.py", line 843, in generate_trajectory
    with self.cache_ctx_manager(self.enable_kv_cache):
  File "/usr/lib/python3.10/contextlib.py", line 135, in __enter__
    return next(self.gen)
  File "/root/torchtune/torchtune/modules/common_utils.py", line 389, in local_kv_cache
    model.setup_caches(
  File "/root/torchtune/torchtune/modules/transformer.py", line 441, in setup_caches
    layer.setup_caches(
  File "/root/torchtune/torchtune/modules/transformer.py", line 62, in setup_caches
    self.attn.setup_cache(batch_size, dtype, max_seq_len=decoder_max_seq_len)
  File "/root/torchtune/torchtune/modules/attention.py", line 165, in setup_cache
    self.kv_cache = KVCache(
[...]

PR #2423, opened 2025-02-22
[RFC] Proposal to Update PPO Test to Add LR Scheduler

Seoley:
#### Context
- [x] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

#### Changelog
Related Issue: https://github.com/pytorch/torchtune/issues/2064#issuecomment-2668489405

I refered to https://github.com/pytorch/torchtune/issues/2263 and add an lr_scheduler option to the Mistral PPO recipe. Then, run the recipe using the following command:
`tune run ppo_full_finetune_single_device --config mistral/7B_full_ppo_low_memory`

Here are the loss curves for reference:
PPO_loss_curve

And then, I need to update the test code in `recipes/test_ppo_full_finetune_single_device.py`

This commit includes changes where I added the lr_scheduler option to recipes/test_ppo_full_finetune_single_device.py, but the test has failed. I suspect the failure might be due to using LLaMA.
How should I resolve this issue? I would like to ask for some advice.

#### Test plan

#### UX


joecummings (commented):
@SalmanMohammadi  Can you take a look?

recipes/ppo_full_finetune_single_device.py
```diff
@@ -257,6 +258,18 @@ def setup(self, cfg: DictConfig) -> None:
             * (self.batch_size // self._ppo_batch_size)
         )
 
+        self._steps_per_epoch = (
+            len(self._dataloader) // self._gradient_accumulation_steps
+        )
+        self.global_step = self._epochs_run * self._steps_per_epoch
```
SalmanMohammadi, line 264: `global_step` is already defined here https://github.com/pytorch/torchtune/blob/80da6a5dae23a201595d07041c12ffde830332d7/recipes/ppo_full_finetune_single_device.py#L253 - how come you're re-defining it here?

recipes/ppo_full_finetune_single_device.py
```diff
@@ -257,6 +258,18 @@ def setup(self, cfg: DictConfig) -> None:
             * (self.batch_size // self._ppo_batch_size)
         )
 
+        self._steps_per_epoch = (
```
SalmanMohammadi, line 261: I think we should name this `lr_steps` and the correct value would be the total number of optimizer steps being taken, which should be `self._total_steps * self._ppo_epochs * (self.batch_size // self._ppo_batch_size)`, right?

SalmanMohammadi:
Hi @Seoley! Thanks so much for taking this on. Sorry for the delay in getting round to a review here.

I think this generally looks good - I think the only comments I have are how we calculate the number of steps for the scheduler, which I've left inline. Let me know what you think : )

Seoley:
Hi @SalmanMohammadi! Thanks for the review. Sorry for the late reply, I was on vacation. I will check and reply this weekend!

SalmanMohammadi:
Hi @Seoley - let me know if you still plan on finishing this, otherwise I'm happy to help get it over the line.

Seoley:
Hi @SalmanMohammadi
Sorry for the delay, I’ve been a bit tied up recently so I lost track of this thread. I'm looking into it now and will get back to you soon! I haven’t had the chance to fully look into this yet, but it’s still on my radar. 
Thank you, as always, for your support!

recipes/ppo_full_finetune_single_device.py
Seoley: You're right. I had redefined `global_step` in `full_finetune_single_device.py` to use it, but it seems I wrote it incorrectly. I've deleted the redefined code and I'm testing code now.

recipes/ppo_full_finetune_single_device.py
Seoley: You're absolutely right! My code didn't properly reflect the structure of PPO. After your suggestion, I added `lr_step` and I'm also updating the definition of the `lr_scheduler` as follows:

```python
# one "step" is a single gradient update over a minibatch of trajectories
self.global_step = (
    self._steps_run
    * self._ppo_epochs
    * (self.batch_size // self._ppo_batch_size)
)

self.lr_step = self._total_steps * self._ppo_epochs * (self.batch_size // self._ppo_batch_size)

# Setup lr scheduler
self._lr_scheduler = self._setup_lr_scheduler(
    cfg_lr_scheduler=cfg.get("lr_scheduler", None),
    num_training_steps=self.lr_step,
    last_epoch=self.global_step - 1,
)
```
Did I understand your suggestion correctly?

SalmanMohammadi: This sounds good to me!

Seoley:
@SalmanMohammadi I updated lr_step and global_step as you suggested and pushed the changes again.
Also, I accidentally wrote the wrong commit message while applying some formatting changes. Thanks for your understanding.

recipes/ppo_full_finetune_single_device.py
```diff
@@ -257,6 +258,19 @@ def setup(self, cfg: DictConfig) -> None:
...
+        self.lr_step = (
+            self._total_steps
+            * self._ppo_epochs
+            * (self.batch_size // self._ppo_batch_size)
+        )
+
+        # Setup lr scheduler
+        self._lr_scheduler = self._setup_lr_scheduler(
+            cfg_lr_scheduler=cfg.get("lr_scheduler", None),
+            num_training_steps=self.lr_step,
```
SalmanMohammadi, line 270:
```suggestion
            num_training_steps=lr_step,
```

recipes/ppo_full_finetune_single_device.py
```diff
@@ -257,6 +258,19 @@ def setup(self, cfg: DictConfig) -> None:
             * (self.batch_size // self._ppo_batch_size)
         )
 
+       lr_steps = (
```
SalmanMohammadi, line 261:
```suggestion
        lr_steps = (
```

SalmanMohammadi (approved):

PR #2427, opened 2025-02-23
Custom DPO losses support

krammnic:
#### Context
- [X] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

#### Changelog
* #2292 

#### Test plan

#### UX

### DIFF
--- a/recipes/lora_dpo_single_device.py
+++ b/recipes/lora_dpo_single_device.py
@@ -539,8 +539,6 @@ class LoRADPORecipeSingleDevice(FTRecipeInterface):
                 policy_chosen_rejected_outputs = self.concatenated_forward(
                     self._model, batch
                 )
-                
-                print("forwared")
 
                 policy_chosen_logits_mean = policy_chosen_rejected_outputs.chosen_logits.detach().mean()
                 policy_rejected_logits_mean = policy_chosen_rejected_outputs.rejected_logits.detach().mean()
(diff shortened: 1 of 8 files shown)


krammnic:
Pushed small fixes and I was able to forward!

torchtune/rlhf/loss/dpo.py
```diff
@@ -85,15 +80,15 @@ def forward(
...
         )
         rejected_rewards = (
-            self.beta * (policy_rejected_logps - reference_rejected_logps).detach()
+            self.beta * (policy_inputs.rejected_logps - reference_inputs.rejected_logps).detach()
         )
 
         return losses, chosen_rewards, rejected_rewards
 
-
+@deprecated(msg="RSOLoss will be deprecated in an upcoming release.")
```
ebsmothers, line 91: Why are we marking as deprecated? Seems this will not be usable with the current changes anyways?

recipes/lora_dpo_single_device.py
```diff
@@ -533,31 +536,25 @@ def train(self) -> None:
...
-                    policy_chosen_log_probs,
-                    policy_rejected_log_probs,
-                    policy_chosen_logits,
-                    policy_rejected_logits,
-                ) = self.concatenated_forward(self._model, batch)
+                policy_chosen_rejected_outputs = self.concatenated_forward(
+                    self._model, batch
+                )
+                
+                print("forwared")
```
ebsmothers, line 543: remove

ebsmothers (commented):
One question and one nit. After that we're good to go!

recipes/lora_dpo_single_device.py
krammnic: Oops

ebsmothers (approved):
Thanks for patiently addressing all the comments. This looks good to me!

Issue #2424, opened 2025-02-22
Token 128011 is not defined in any way in the Llama tokenizer


RedTachyon:
This is a fun one. In the [Llama 3 tokenizer](https://github.com/pytorch/torchtune/blob/main/torchtune/models/llama3/_tokenizer.py), we generally have 128257 tokens. Tokens 0-127999 are all "normal" BPE, and tokens 128000-128256 are special tokens - some are fairly standardized, like <|begin_of_text|>, others are left largely meaningless, like <|reserved_special_token_2|>, so that people can use them for some custom fine-tuning or whatever else.

The way it's handled in torchtune is that we have a dictionary of [meaningful special tokens](https://github.com/pytorch/torchtune/blob/cf0142bd2eb5eabe33a3d43eca4f07ec48a8749f/torchtune/models/llama3/_tokenizer.py#L20) and then we create another dictionary with automatically-named [meaningless special tokens](https://github.com/pytorch/torchtune/blob/cf0142bd2eb5eabe33a3d43eca4f07ec48a8749f/torchtune/models/llama3/_tokenizer.py#L38)

However, one of them was forgotten. Like the one ring, many thought it to be a legend. Alas, token 128011 is nowhere to be found, and if a naive ~hobbit~ LLM with a high temperature decides to sample it, evil will triumph (the program will crash).

The meaningful special tokens cover tokens (128000+)0-10, 256, 12. The meaningless tokens start from token 128013. Leaving token 128011 undefined, causing tiktoken to crash whenever it appears.

---

After writing all of this, I decided to look it up and indeed, there already is an issue for this: #1995. Still, I figure it's better to open a new issue than try to necropost on that one. The key question that was asked there by @RdoubleA was "But do you find this token id in a real use case?", and unfortunately, RL is that use case. When we train an LLM with RL, it will sample a lot of tokens and try out a lot of weird stuff, because who knows, perhaps token 128011 is the key to AGI. So while it doesn't really matter what 128011 is detokenized into, what matters is that the program doesn't crash in the process.

The code change is obviously trivial, the less trivial part is deciding on the specific fix. In my opinion, it should just become <|reserved_special_token_245|>. The non-consecutiveness is annoying, but probably better than relabeling all other tokens. I'll be happy to submit a PR if this is acceptable.

krammnic:
[...]

PR #2428, opened 2025-02-24
Add token 128011 to the Llama tokenizer

RedTachyon:
#### Context
- [ ] add a new feature
- [x] fix a bug
- [ ] other (please add here)

See issue #2424 

For now I just defined this token as a <|reserved_special_token_244|> for simplicity. The alternative approach is reindexing all special tokens, but this might require some extra attention.

My two concerns about this are:
1. How exactly should we reindex them? Tokens that don't have a specific definition are tokens (128000+) 2, 3, 11, and then 13-255. I don't think there's a coherent way in which we can match up with the HF tokenizer that has 128011 <-> <|reserved_special_token_6|>. Do we just do 128011 as special token 2, and then tokens 128013+i become special tokens 3+i?
2. Are we absolutely sure that this won't break anyone's workflows or trained models? I never actually used special tokens, but the way I understand they could be used, is that someone could define a format, e.g. "Question <|reserved_special_token_0|>cot<|reserved_special_token_1|> <|reserved_special_token_2|>answer<|reserved_special_token_3|>". A model trained on this today would learn to associate <|reserved_special_token_2|> with token 128013. If we reindex it, then <|reserved_special_token_2|> becomes token token 128011, so everything falls apart. But is this how the special tokens are actually used?

The way I see it, defining it as special token 244 won't break anything, because trying to decode token 128011 would have resulted in a crash. Reindexing all special tokens miiiiiiiiiight cause some issues in some weird cases, but I'm very much uncertain.

As a side note - NUM_RESERVED_TOKENS was set to 256, which seems reasonable, it's a nice round number. But in fact there are 257 special tokens: 0-256, inclusive on both ends. This bug canceled itself out with the omission of token 128011, but now I needed to fix this as well.

#### Changelog
* Added token 128011 to the Llama tokenizer

### DIFF
--- a/torchtune/models/llama3/_tokenizer.py
+++ b/torchtune/models/llama3/_tokenizer.py
@@ -31,7 +31,7 @@ SPECIAL_TOKENS = {
     "<|python_tag|>": 128010,
     "<|image|>": 128256,
     "<|video|>": 128012,
-    "<|reserved_special_token_255|>": 128011,
+    "<|reserved_special_token_244|>": 128011,
 }
 
 NUM_RESERVED_SPECIAL_TOKENS = 256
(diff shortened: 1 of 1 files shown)


joecummings (approved):
I'm okay with this as a stop gap while we look into re-indexing

PR #2431, opened 2025-02-24
Add `StatefulDataLoader` to select other recipes

joecummings:
**What?** This expands the use of `StatefulDataloader` to the following four recipes:
- [x] Full distributed
- [x] LoRA single device
- [x] LoRA distributed
- [x] GRPO

**What did you change?**
I copied pretty much the exact same changes as I did in #2410 into the above four recipes. I modified the tests for single device recipes b/c we are not relying on a different state. For the distributed recipes, I did not modify the tests at all. I had to change the hack we used to ensure the iterator finish even if we cut the epoch short b/c it was not robust and also looked ugly to manually modify the dataloader state dict. 

**Why these recipes?**
These include our most stable recipes + our newest one GRPO. By looking at the changes I did here, users will be able to propagate changes to everything else in the library easily. I will be creating an Issue to track this for all the rest of the recipes. 

**How did you test GRPO?**
Good question, b/c GRPO has no standardized tests in the torchtune library. It's also an interesting case b/c it does not follow the same format as our other recipes. For instance, there is NO option to cut an epoch short based on the number of steps in an individual epoch. (It's a weird variable anyways). However, there IS an option to reach a total number of training steps and stop. But wait, this wasn't included in the saved state dict!! There is no way to reasonably resume training for GRPO without *actually* implementing step-based checkpointing. This one is a wash right now :/

### DIFF
--- a/recipes/dev/grpo_full_finetune_distributed.py
+++ b/recipes/dev/grpo_full_finetune_distributed.py
@@ -599,6 +599,7 @@ class FullGRPOFinetuneRecipeDistributed(FTRecipeInterface):
         )
         if dataloader_state_dict is not None:
             dataloader.load_state_dict(dataloader_state_dict)
+            list(dataloader)  # Hack to force dataloader to finish last iteration
         return dataloader
 
     def save_checkpoint(
@@ -666,8 +667,6 @@ class FullGRPOFinetuneRecipeDistributed(FTRecipeInterface):
             # if training is in-progress, checkpoint the optimizer state and recipe state
             # as well.
             if intermediate_checkpoint:
-                dataloader_sd = self._dataloader.state_dict()
-                dataloader_sd["_iterator_finished"] = True
                 checkpoint_dict.update(
                     {
                         training.OPT_KEY: opt_state_dict,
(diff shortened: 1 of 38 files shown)


recipes/dev/grpo_full_finetune_distributed.py
```diff
@@ -571,30 +577,30 @@ def _setup_data(
...
+            # dropping last avoids shape issues with compile + flex attention
+            drop_last=True,
         )
-
-        utils.log_rank_zero(log, "Dataset and Sampler are initialized.")
-
-        return sampler, dataloader
+        if dataloader_state_dict is not None:
+            dataloader.load_state_dict(dataloader_state_dict)
+            list(dataloader)  # Hack to force dataloader to finish last iteration
```
ramanishsingh, line 602: Thanks for including StatefulDataloader in these recipes. 
I have a couple of comments.
1. Let's add a more descriptive comment, something like "Since we break early with complete the last epoch and we want to start a new epoch when restarting train, we need to yield the remaining batches from the last epoch break breaking". Doesn't need to be this long, but something like this, as someone who doesn't work with StatefulDataLoader might get confused here.
2. Do you think we need a flag (like `finish_current_epoch` or some other name) that we set when we break early from an epoch, and if that flag is `True` (we set it if we break early) only then we complete the epoch before restarting training.

joecummings: Re 1 - yeah definitely makes sense let me add it.

Re 2 - b/c right now we always save at epoch boundaries, there will never be a time when we don't want to complete the epoch before restarting training. This will change very very soon as we start mid-epoch checkpointing.

felipemello1 (approved):

Issue #2371, opened 2025-02-09
Adding new role "Tool" to support Llama 3.3 models


init27:
Right now in chat/SFT datasets, supported [roles](https://pytorch.org/torchtune/stable/generated/torchtune.data.Role.html#torchtune.data.Role) are "[‘system’, ‘user’, ‘assistant’, ‘ipython’]". 

With the latest Llama models 3.3 onwards, we have a new [role: `tool`](https://github.com/meta-llama/llama-models/blob/main/models/llama3_3/prompt_format.md#llama-33---prompt-formats) which serves the same purpose as "`ipython`"

It would be great to have this added to support fine-tuning for Tool-calling purposes using Torchtune. 

Thanks in advance!

RdoubleA:
Yes, this is a leftover item we needed to do after the release of 3.3. Renaming ipython to tool is a lot more clear. Feel free to open a PR to make this change (replace all mentions of ipython role with tool in code and docs), otherwise we can keep this issue open until it gets picked up.

init27:
@RdoubleA Thanks for the reply! 

Shouldn't we also keep the role `ipython` given that we might still want to keep support for 3.1 and 3.2 available. I leave the decision to the team ofcourse it might be better to just support the latest 70B however for the smaller model(s) we still need to keep the `ipython` role if I understand correctly.

Ankur-singh:
Thanks for pointing it out @init27. 

After a brief investigation, I see that only `Qwen2_5` and `Mistral` tokenizer and chat template that make user of `ipython` role. As tool calling is becoming more and more important, it would be great idea to generalize it and also add some example config.

> One thing to consider is that this would be BC breaking, so we should make the change in a way that doesn't break older versions or recipes.

Totally! @RdoubleA I'll be happy to pick it up. Can you please assign it to me? 

I'm working with this part of the codebase for the first time, might need some help.

init27:
Thanks for taking a look, @Ankur-singh!

> After a brief investigation, I see that only Qwen2_5 and Mistral tokenizer and chat template that make user of ipython role. As tool calling is becoming more and more important, it would be great idea to generalize it and also add some example config.

Llama 3.1 and 3.2 models do as well so it's an interesting split for 3.3 onwards using `Tool`

It might be really cool to include an example config using a dataset like [Nous-func-calling](https://huggingface.co/datasets/NousResearch/hermes-function-calling-v1/tree/main) for the demo config if you'd like. 

[...]

PR #2438, opened 2025-02-26
[WIP] Add Llama3.3 tokenizer and update message roles to 'tool' (instead of 'ipython')

Ankur-singh:
#### Context
- [X] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

Please link to any issues this PR addresses #2371

#### Changelog
- Updated `Role` to use `tool` instead of `ipython`. Also updated all tokenizers and tests to make sure all tests pass
- Updated `Llama3Tokenizer` to continue using `ipython` for backward compatibility
- Created `Llama3_3Tokenizer` that inherits from `Llama3Tokenizer` overriding `_tokenize_header` method

#### Test plan

#### UX

---

I'm not sure whether this is the right approach to handle Llama3 and Llama3.3 tokenizer? Based on my discussion with @pbontrager & @RdoubleA there are two possible ways:
    - Having a single tokenizer, and taking care of tool roles (`ipython` or `tool`) based on the model. This requires us to pass the model detail to the tokenizer.
    - Having two separate tokenizers 

This PR implements the 2nd approach, while its easier to implement it can lead to confusion among users. Looking forward to the discussion.

### DIFF
--- a/tests/torchtune/data/test_messages.py
+++ b/tests/torchtune/data/test_messages.py
@@ -57,7 +57,7 @@ class TestMessage:
             ValueError,
             match="Only assistant messages can be tool calls. Found role user in message: hello world",
         ):
-            message = Message(role="user", content="hello world", ipython=True)
+            message = Message(role="user", content="hello world", tool=True)
 
         with pytest.raises(
             ValueError,
@@ -66,7 +66,7 @@ class TestMessage:
             message = Message(
                 role="user",
                 content=[{"type": "image", "content": test_image}],
-                ipython=True,
+                tool=True,
             )
 
     def test_from_dict(self):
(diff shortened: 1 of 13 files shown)


felipemello1:
hey @Ankur-singh , i was not part of the discussions and i went this pr very quickly. Just two thoughts:

1. Does it overlap with https://github.com/pytorch/torchtune/pull/2407 ?
2. Does it preserve backward compatbility?

Ankur-singh:
> 1. Does it overlap with [Tool role support #2407](https://github.com/pytorch/torchtune/pull/2407)?
> 2. Does it preserve backward compatibility?

1. #2407 was only about `tool` role in OpenAI messages. This PR is about deprecating `ipython` role and replacing it with `tool` role. The discussing started with #2371. I think there is certaining some scope creep in this PR, as its trying to do two things: replacing `ipython` role, and adding llama3.3 tokenizer. Probably, I should split this into two separate PRs.
2. No, its not backward compatible. Handling `ipython` role (i.e. replacing `ipython` with `tool` and showing a deprecation warning) during `Message` initialization should be sufficiant. Please let me know if I'm missing something.
 
Also, currently we don't have any recipe that makes use of tool dataset. Tool calling reciepe would be great a great addition.

PS: Really sorry for delayed response, I was travelling.

PR #2447, opened 2025-02-28
[WIP][RFC] Refactored and Simplified Recipes

pbontrager:
This is initial work to refactor the recipes and update the training utils to have simpler, more maintainable recipes.

**This is a first pass**

Please review the new_sft_distributed file for directional feedback. The first pass mostly avoids adding new utils (besides for profiler and gradient accumulation) but we could choose to more aggressively include some.

More context and a writeup will be included here later. This requires more updates to pass all tests.

### DIFF
--- a/recipes/dev/new_sft_distributed.py
+++ b/recipes/dev/new_sft_distributed.py
@@ -1,0 +1,357 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+import sys
+import time
+from tqdm import tqdm
+from functools import partial
+from typing import Any, Dict, List, Optional, Tuple, Union
+
+import torch
+from torch.distributed import (
+    destroy_process_group,
+    init_device_mesh,
+    init_process_group,
+)
+from torch.distributed._tensor import DTensor
+from torch.utils.data import DataLoader, DistributedSampler
+from torchtune import config, modules, training, utils
+from torchtune.config._utils import _get_component_from_path
+from torchtune.data import padded_collate_packed
+from torchtune.datasets import ConcatDataset
+
+
+log = utils.get_logger("DEBUG")
+
+@dataclass
+class RecipeState:
+    device = "cuda"
+    dtype = "bf16"
+    seed = None
+    compile = False
+
+    # distributed attributes
+    enable_async_checkpointing = False
+    fsdp_cpu_offload = False
+    fsdp_reshard_after_forward = True
+    custom_sharded_layers = None
+    tensor_parallel_plan = None
+    tensor_parallel_dim = 1
+
+    # logging attributes
+    output_dir = "."
+    log_every_n_steps = 1
+    log_peak_memory_stats = False
+    eval_every_n_steps = 1
+    save_every_n_steps = 1
+
+    # training attributes
+    resume_from_checkpoint = False
+    gradient_accumulation_steps = 1
+    clip_grad_norm = None
+    enable_activation_checkpointing = False
+    enable_activation_offloading = False
+    shuffle = False
+    batch_size = 1
+    num_training_steps = 1
+
+    # artifacts
+    model: Any
+    tokenizer: Any
+    optimizer: Any
+    loss: Any
+    checkpointer: Any
+    metric_logger: Any
+    dataset: Any
+    eval_dataset: Any = None
+    collate_fn: Any = torchtune.data.padded_collate_sft
+    profiler: Any = None
+
+
+def setup_env(state):
+    state.device = training.get_device(state.device)
+    state.dtype = training.get_dtype(state.dtype, device=state.device, options=["fp32", "bf16"])
+
+    # ---- Initialize distributed ---- #
+    cpu_offload = state.fsdp_cpu_offload or state.enable_async_checkpointing
+    distributed_backend = training.get_distributed_backend(state.device, offload_ops_to_cpu=cpu_offload)
+    init_process_group(distributed_backend)
+    if state.fsdp_cpu_offload:
+        training.set_torch_num_threads()
+
+    state.world_size, state.rank = training.get_world_size_and_rank()
+    state.data_parallel_dim = state.world_size // state.tensor_parallel_dim
+    state.device_mesh = training.init_device_mesh(
+        state.device.type,
+        mesh_shape=(state.data_parallel_dim, state.tensor_parallel_dim),
+        mesh_dim_names=("dp", "tp"),
+    )
+    state.data_parallel_size = device_mesh["dp"].size()
+    state.data_parallel_rank = device_mesh["dp"].get_local_rank()
+
+    state.seed = training.set_seed(state.seed, debug_mode=state.cudnn_deterministic_mode)
+
+    # ---- Setup Checkpointing ---- #
+    state.checkpointer = training.Checkpointer(state.checkponter)
+    state.checkpoint_dict = state.checkpointer.load_base_checkpoint()
+    state.total_steps = state.num_training_steps
+    state.step = 0
+
+    # ---- Initialize logging ---- #
+    state.metric_logger = config.instantiate(state.metric_logger, log_rank=0)
+    state.metric_logger.log_config(state)
+    state.tqdm = tqdm(total=state.num_training_steps, disable=state.rank > 0)
+    state.profiler = training.get_profiler(**state.profiler)
+
+
+def setup_model(state):
+    utils.log_rank_zero(log, "Distributed training is enabled. Instantiating model and loading checkpoint on Rank 0",)
+    init_start = time.perf_counter()
+
+    with training.set_default_dtype(state.dtype), torch.device("meta"):
+        state.model = config.instantiate(state.model)
+
+    if state.compile:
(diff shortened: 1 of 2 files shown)


recipes/dev/new_sft_distributed.py
```diff
@@ -0,0 +1,357 @@
...
+            else padded_collate_packed
+        ),
+    )
+
+    utils.log_rank_zero(log, "Dataset and Sampler are initialized.")
+
+
+def resume_training(state)
+    state.checkpointer.update_state(state)
+    state.checkpointer.load_distributed_checkpoint(state.model, state.optimizer)
```
Darktex, line 240: This needs to be extended for dataloader state too

recipes/dev/new_sft_utils.py
```diff
@@ -0,0 +1,228 @@
...
+            self.torch_profiler.start()
+        return self
+
+    def __exit__(self, type, value, traceback):
+        if torch_profiler is not None:
+            self.torch_profiler.stop()
+
+
+
+class Checkpointer: # formerly CheckpointClient
```
Darktex, line 96: Data state needs to be checkpointed here too, no?

recipes/dev/new_sft_utils.py
```diff
@@ -0,0 +1,228 @@
...
+        loss = state.loss(logits, labels)
+        grad_acc.scale_loss(loss, num_tokens)
+
+        loss.backward()
+
+
+
+# Option 2
+@dataclass
+class AccumulationStats:
```
Darktex, line 194: For just gradient accumulation, I would probably go with Option 1. What I like about this option is that we can extend it to cover more training stats. Indeed, these values can be reuse to compute tokens per second and other useful performance metrics

recipes/dev/new_sft_distributed.py
```diff
@@ -0,0 +1,357 @@
...
+def train(state):
+    training.cleanup_before_training()
+    state.optimizer.zero_grad()
+
+    with state.profiler() as profiler:
+        for state.step in range(state.step, state.total_steps):
+            # ---- train batch ---- #
+            with GradientAccumulation(state.gradient_accumulation_steps, state.model, state.data_parallel_size) as grad_acc:
+                for _ in range(grad_acc.steps):
+                    batch = next(state.dataloader)
```
Darktex, line 256: `batch` could be a TensorDict (or better yet a TensorClass), see comment below

recipes/dev/new_sft_distributed.py
```diff
@@ -0,0 +1,357 @@
...
+                    utils.batch_to_device(batch, state.device)
+
+                    # Shape [b, s], needed for the loss not the model
+                    labels = batch.pop("labels")
+                    num_tokens = (labels != loss.ignore_index).sum()
+
+                    with self.activation_offloading:
+                        logits = state.model(**batch)
```
Darktex, line 264: Tune supports packing today. Don't we need to send the mask to the model here to enable it? (Can be optional if you pad).

This is also a great example on why TensorDict/TensorClass can be ergonomical!

recipes/dev/new_sft_distributed.py
```diff
@@ -0,0 +1,357 @@
...
+    with state.profiler() as profiler:
+        for state.step in range(state.step, state.total_steps):
+            # ---- train batch ---- #
+            with GradientAccumulation(state.gradient_accumulation_steps, state.model, state.data_parallel_size) as grad_acc:
+                for _ in range(grad_acc.steps):
+                    batch = next(state.dataloader)
+                    utils.batch_to_device(batch, state.device)
+
+                    # Shape [b, s], needed for the loss not the model
+                    labels = batch.pop("labels")
```
Darktex, line 260: This is SFT, so `labels` is just the same as `x`, offset by 1 and with additional masking so that system prompt and user prompt (as well as any delimiters) do not contribute to the loss.

We may wanna build a custom loss function for this, which takes only the input sequence and a mask for what tokens contribute to the loss, as this avoids making a copy of the input tokens. 

We can get to this when we start working on efficient data loading, just pointing it out as an item we should have on the radar

recipes/dev/new_sft_distributed.py
```diff
@@ -0,0 +1,357 @@
...
+
+
+def train(state):
+    training.cleanup_before_training()
+    state.optimizer.zero_grad()
+
+    with state.profiler() as profiler:
+        for state.step in range(state.step, state.total_steps):
+            # ---- train batch ---- #
+            with GradientAccumulation(state.gradient_accumulation_steps, state.model, state.data_parallel_size) as grad_acc:
```
Darktex, line 254: I think we should put a fp8 context manager around here too potentially?

Darktex: Design question: do we want to have a big context manager class that will enable certain things based on config such as grad checkpointing, fp8, and whatever else we may need vs having multiple single-purpose context managers?

recipes/dev/new_sft_distributed.py
Darktex: (note that with packing we would need to carry with us _two_ masks: one for the forward, where you mask only the sequences that do not belong to the current sequence, and one for the backward where in addition to that masking, you _also_ mask out special tokens, user prompts and system prompt)

felipemello1:
Thanks for taking the time to think about this! First impressions:

1.  I like that the "state" is exposed (maybe the equivalent to our config?)
2. I like that the functions are more clean and organized.

3. Implementation Concerns: I’m unclear about how this will work in practice. Will these functions be included in the recipe file that users might copy? If so, then I don't think that it address the issue of stale code when users duplicate recipes. Alternatively, if the functions are separated from the recipe, I worry it might make it harder for users to understand the whole picture.

4. Cognitive Load:  Developers are familiar with functions and classes, but using a state object feels like an attempt to mimic class-like behavior without actually using a class. I’d need more clarity on why this is better than a well-designed class.
    4.1 Utility Functions: For example, if the goal is to make setup_loss a reusable utility function, we could still achieve that within a class:
    ```
    class MyClass:
        def __init__(self):
            self.loss = utilities.setup_loss(*args)
    ```
    4.2 Customization: If the idea is to let users replace setup_loss(state) easily, inheritance or a custom recipe could also accomplish it
    
5. Coding Practices: Passing a state object to functions seems to be different from our standards. We typically prefer functions with specific arguments (e.g., fun(arg1, arg2)) over passing a config or state object (e.g., fun(my_dict)), as it makes dependencies clearer. I’m worried this could complicate debugging, since it’s less obvious which parts of the state each function uses or changes.

----
I’m completely aligned with the goal of cleaner, more modular code (points (1) and (2)), but I’m not yet convinced that sharing a state object between functions is the best way to go. I am more inclined to seeing our recipes as independent trainers, keeping them as they are, but adding the ability to import and inherit, so that minimal recipe changes do not result in having to copy the entire recipe.

recipes/dev/new_sft_distributed.py
```diff
@@ -0,0 +1,392 @@
...
+        - Overwritten by arguments from the command-line
+    """
+    config.log_config(recipe_name="FullFinetuneRecipeDistributed", cfg=cfg)
+    trainer = TrainingRecipe(cfg=cfg)
+    trainer.setup_env()
+    trainer.setup_model()
+    trainer.setup_optimizer()
+    trainer.setup_loss()
+    trainer.setup_data(trainer.dataset)
+    trainer.setup_data(trainer.eval_dataset)
```
felipemello1, line 384: is this preference or is there a reason to not have trainer.setup()

recipes/dev/new_sft_utils.py
```diff
@@ -0,0 +1,228 @@
...
+        return scaled_loss
+
+    @property
+    def avg_loss(self):
+        if self._count > -1:
+            raise ValueError(f"avg_loss was accessed before gradient accumulation steps were completed")
+        return self.running_loss.item() / self.num_tokens
+
+    def __enter__(self):
+        return self
```
pbontrager, line 164: Add reduce_all_reduce settings if the user is using fsdp

recipes/dev/new_sft_utils.py
```diff
@@ -0,0 +1,228 @@
...
+            torch.distributed.all_reduce(stats.num_tokens)
+            torch.distributed.all_reduce(stats.running_loss)
+        training.scale_grads(model, dp_size / stats.num_tokens)
+    return grad_accumulator
+
+
+for acc_step in grad_accumulator(grad_steps, model, dp_size):
+    batch = next(dataloader)
+    labels = batch.pop("labels")
+    acc_step.token_count((labels != loss.ignore_index).sum())
```
pbontrager, line 220: If we give the dataloader to the grad_accumulator, then it can return each batch and grab the token count internally without the need for calling acc_step. This would leave the inside of the loop completely clean.

recipes/dev/new_sft_utils.py
```diff
@@ -0,0 +1,228 @@
...
+    running_loss: int = 0
+
+    def scale_loss(self, loss, num_tokens):
+        scaled_loss = loss * num_tokens
+        self.num_tokens += num_tokens
+        self.running_loss += scale_loss
+        return scaled_loss
+
+
+def get_gradient_accumulator(steps, model, dp_size=1):
```
pbontrager, line 205: Make this a class to instantiate once with model and dp_size and steps. Then just call the generator method in train.

Issue #1042, opened 2024-06-01
Recommendations for obtaining validation dataset loss after each epoch


dcsuka:
For finetuning using a custom dataset, message converter function, and csv column format, how do we obtain validation losses on a separate csv with the same format at the end of each epoch? Do we need to wait until after training to run on all the checkpointed files?

Also, how can we generate outputs using the same message converter function and `tune run generate`, using a csv file with a single row as input?

pbontrager:
None of our recipes currently support doing a validation check each epoch. The easiest way to get this functionality would be to copy the recipe you want, duplicate _setup_data with something like _setup_val_data, call it, and then setup a for loop after the training for loop to run validation. 

You can also make a request for this to be in our recipes by default, but we'd have to discuss whether it's worth the extra complexity. As for the generate recipe, it's not meant for generating from a csv, that seems more like an evaluation flow but we can discuss that too.

MaxFrax:
@dcsuka Where you able to add the validation loss to your recipe?

MaxFrax:
I've started working on this for lora_singledevice_finetune https://github.com/MaxFrax/torchtune/tree/add_validation_set_lorasingledevice

If anybody wants to contribute or share tips, feel free to reach out

felipemello1:
hey @MaxFrax , thats so nice! :)

We are aware of this gap in torchtune, and its something that we will focus on next quarter. We do support lm harness using eleuther AI, but are docs don't cover yet what users should do when evaluating on custom datasets.

We would love to take a look at your fork when you are done with your changes!

Issue #883, opened 2024-04-26
Early stopping during training


kinggongzilla:
Is there a way to evaluate the model performance during training on a validation dataset and only save a new checkpoint if it achieves lower validation loss?

rohan-varma:
Hi @kinggongzilla, thanks for filing this issue!

Currently we only support an early stopping that's based on the # of steps taken in an epoch, i.e. you can set `max_steps_per_epoch` flag in the [configuration](https://github.com/pytorch/torchtune/blob/cc2dd05f47f5de5737e8d8cc3666f8c65946c78d/recipes/configs/llama3/8B_full.yaml#L59) to early stop your model based on a # of steps.

However this doesn't satisfy your use case of only early stopping / saving a checkpoint based on some validation results. 

In training evaluation + stopping criteria based on evaluation is a large space we haven't looked deeply into, what do you folks think @ebsmothers @RdoubleA? I could see a future in which we allow users to specify a validation dataset or validation split, and incorporate validation metrics into our checkpointer for whether to save a checkpoint or not. This is definitely something we could look at enabling in the future if there's sufficient interest.

kinggongzilla:
Thanks for the quick reply!
Being able to define a validation dataset and do early stoppingbased on the validation loss would definitely be super helpful.

optimass:
+1 this would be super useful.

Some-random:
+1 Would be super useful!

ebsmothers:
Thanks all for the comments. This feature (along with general validation loops) are fairly high on our wishlist right now. We still need to do a bit of design to make sure it's not too intrusive into our recipes, but definitely hear you on the need for this feature. We will keep you posted here!

Tandon-A:
Hi @ebsmothers, 

Is there any update on this?

karrtikiyer:
+1 Would be super useful!

PR #2464, opened 2025-03-06
Add validation dataset loss to distributed SFT recipies

bzz:
#### Context
- [x] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

Addresses #1042 / part of the #883 for distributed recipies

#### Changelog
* distributed SFT recipeis (full_finetune_distributed and lora_finetune_distributed) support computing loss on validation dataset (no early stopping)
* example of configuring validation dataset in [recipes/configs/llama3_2/1B_qlora_validation.yaml](recipes/configs/llama3_2/1B_qlora_validation.yaml)

New [configuration](https://github.com/pytorch/torchtune/pull/2238#pullrequestreview-2553490828):
```
dataset_validation: null
run_val_every_n_steps: null
max_validation_batches: -1
```

#### Test plan

As [suggested](https://github.com/pytorch/torchtune/pull/2238/files#r1917299528), I run with  compile + opt_in_bwd + activation ckpt + activation offloading on 2xH100 using [mhenrichsen/alpaca_2k_test](https://huggingface.co/datasets/mhenrichsen/alpaca_2k_test) as a validation dataset.

#### LoRA Distributed

```
pip install -U git+https://github.com/bzz/torchtune.git@validate_distributed

# with validation
tune run --nnodes 1 --nproc_per_node 2 lora_finetune_distributed --config 1B_qlora_validation.yaml \
	compile=True \
	enable_activation_offloading=True \
	optimizer_in_bwd=True \
	tokenizer.max_seq_len=4096 \
	epochs=1 \
	batch_size=16 \
	metric_logger._component_=torchtune.training.metric_logging.WandBLogger \
	output_dir=/tmp/torchtune/llama3_2_1B/qlora_val
```

Without validation
```
tune run --nnodes 1 --nproc_per_node 2 lora_finetune_distributed --config 1B_qlora_validation.yaml \
	run_val_every_n_steps=None \
	compile=True \
	enable_activation_offloading=True \
	optimizer_in_bwd=True \
	tokenizer.max_seq_len=4096 \
	epochs=1 \
	batch_size=16 \
	metric_logger._component_=torchtune.training.metric_logging.WandBLogger \
	output_dir=/tmp/torchtune/llama3_2_1B/qlora
```

Screenshot 2025-03-06 at 08 05 19

<details>
<description>llama3_2/1B_qlora_validation.yaml</description>

```
diff recipes/configs/llama3_2/1B_qlora_validation.yaml recipes/configs/llama3_2/1B_lora.yaml

20c20
< output_dir: /tmp/torchtune/llama3_2_1B/qlora # /tmp may be deleted by your system. Change it to your preference.
---
> output_dir: /tmp/torchtune/llama3_2_1B/lora # /tmp may be deleted by your system. Change it to your preference.
30c30
<   _component_: torchtune.models.llama3_2.qlora_llama3_2_1b
---
>   _component_: torchtune.models.llama3_2.lora_llama3_2_1b
57,65d56
< # Validation
< dataset_validation:
<   _component_: torchtune.datasets.alpaca_dataset
<   source: mhenrichsen/alpaca_2k_test
<   split: train
<   packed: False  # True increases speed
< run_val_every_n_steps: 100
< max_validation_batches: -1
<
82c73
< gradient_accumulation_steps: 1  # Use to increase effective batch size
---
> gradient_accumulation_steps: 8  # Use to increase effective batch size
84c75
< compile: True  # torch.compile the model + loss, True increases speed + decreases memory
---
> compile: False  # torch.compile the model + loss, True increases speed + decreases memory
88c79
<   _component_: torchtune.training.metric_logging.WandBLogger
---
>   _component_: torchtune.training.metric_logging.DiskLogger
```
</details>

The difference in throughput is due to both runs using the same 2 GPUs.

#### Full Distributed

```
tune cp llama3_2/1B_full 1B_full_validation

# Validation
dataset_validation:
  source: mhenrichsen/alpaca_2k_test
  split: train
run_val_every_n_steps: 100
max_validation_batches: -1

metric_logger:
	_component_: torchtune.training.metric_logging.WandBLogger
```

Validation
```
tune run --nnodes 1 --nproc_per_node 2 \
    full_finetune_distributed --config 1B_full_validation.yaml \
	compile=True \
	enable_activation_offloading=True \
	optimizer_in_bwd=True \
	tokenizer.max_seq_len=4096 \
	epochs=1 \
	batch_size=16 \
	output_dir=/tmp/torchtune/llama3_2_1B/full_dist_val
```

And without
```
tune run --nnodes 1 --nproc_per_node 2 \
    full_finetune_distributed --config 1B_full_validation.yaml \
	run_val_every_n_steps=None \
	compile=True \
	enable_activation_offloading=True \
	optimizer_in_bwd=True \
	tokenizer.max_seq_len=4096 \
	epochs=1 \
	batch_size=16 \
	output_dir=/tmp/torchtune/llama3_2_1B/full_dist
```

Screenshot 2025-03-06 at 11 34 48

#### UX

The output is really simplistic and looks like 
```
1|100|Loss: 3.8272886276245117:   6%|████                                                               | 100/1625 [01:05<05:46,  4.41it/s]
INFO:torchtune.utils._logging:Validation loss: 4.8242
1|200|Loss: 1.7956730127334595:  12%|████████▏                                                          | 200/1625 [01:30<05:11,  4.58it/s]
INFO:torchtune.utils._logging:Validation loss: 1.8639
1|300|Loss: 1.4479581117630005:  18%|████████████▎                                                      | 300/1625 [01:52<03:54,  5.66it/s]
INFO:torchtune.utils._logging:Validation loss: 1.7377
1|400|Loss: 1.433119535446167:  25%|████████████████▋                                                   | 400/1625 [02:13<03:51,  5.28it/s]
INFO:torchtune.utils._logging:Validation loss: 1.7247
1|500|Loss: 1.5169874429702759:  31%|████████████████████▌                                              | 500/1625 [02:34<03:13,  5.81it/s]
INFO:torchtune.utils._logging:Validation loss: 1.7714
1|600|Loss: 1.4446262121200562:  37%|████████████████████████▋                                          | 600/1625 [02:55<02:54,  5.89it/s]
INFO:torchtune.utils._logging:Validation loss: 1.7542
1|700|Loss: 1.5118173360824585:  43%|████████████████████████████▊                                      | 700/1625 [03:18<02:54,  5.32it/s]
INFO:torchtune.utils._logging:Validation loss: 1.7290
1|800|Loss: 1.2981687784194946:  49%|████████████████████████████████▉                                  | 800/1625 [03:40<02:09,  6.35it/s]
INFO:torchtune.utils._logging:Validation loss: 1.7309
1|900|Loss: 1.533372163772583:  55%|█████████████████████████████████████▋                              | 900/1625 [04:01<02:22,  5.08it/s]
INFO:torchtune.utils._logging:Validation loss: 1.7264
```

Please if you think it needs to be somehow improved.

### DIFF
--- a/recipes/configs/llama3_2/1B_qlora_validation.yaml
+++ b/recipes/configs/llama3_2/1B_qlora_validation.yaml
@@ -17,7 +17,7 @@
 # For single device LoRA finetuning please use 1B_lora_single_device.yaml
 # or 1B_qlora_single_device.yaml
 
-output_dir: /tmp/torchtune/llama3_2_1B/lora # /tmp may be deleted by your system. Change it to your preference.
+output_dir: /tmp/torchtune/llama3_2_1B/qlora # /tmp may be deleted by your system. Change it to your preference.
 
 # Tokenizer
 tokenizer:
@@ -27,7 +27,7 @@ tokenizer:
 
 # Model Arguments
 model:
-  _component_: torchtune.models.llama3_2.lora_llama3_2_1b
+  _component_: torchtune.models.llama3_2.qlora_llama3_2_1b
   lora_attn_modules: ['q_proj', 'v_proj', 'output_proj']
   apply_lora_to_mlp: True
   lora_rank: 64  # higher increases accuracy and memory
@@ -85,7 +85,7 @@ compile: True  # torch.compile the model + loss, True increases speed + decrease
 
 # Logging
 metric_logger:
-  _component_: torchtune.training.metric_logging.DiskLogger
+  _component_: torchtune.training.metric_logging.WandBLogger
   log_dir: ${output_dir}/logs
 log_every_n_steps: 1
 log_peak_memory_stats: True
--- a/recipes/full_finetune_distributed.py
+++ b/recipes/full_finetune_distributed.py
@@ -750,6 +750,30 @@ class FullFinetuneRecipeDistributed(FTRecipeInterface):
             list(dataloader)
         return dataloader
 
+    def _loss_step(self, batch: Dict[str, torch.Tensor]) -> torch.Tensor:
+        # Shape [b, s], needed for the loss not the model
+        labels = batch.pop("labels")
+
+        with self.activations_handling_ctx:
+            logits = self._model(**batch)
+
+        # Shift labels to compute loss
+        # equivalent to doing labels[..., 1:] and logits[..., :-1, :]
+        # But this way we dont need to slice the logits. We just add an ignore index to labels.
+        labels = torch.hstack(
+            (labels[..., 1:], self.ignore_labels_cache[: labels.shape[0]])
+        )
+        if not isinstance(logits, list):
+            labels = labels.reshape(-1)
+            logits = logits.reshape(-1, logits.size(-1))
+
+        # Compute loss
+        loss = self._loss_fn(logits, labels)
+        # free logits otherwise it peaks backward memory
+        del logits
+
+        return loss
+
     def validate(self) -> float:
         """
         Run validation loop and return average validation loss.
@@ -774,21 +798,8 @@ class FullFinetuneRecipeDistributed(FTRecipeInterface):
                     batch["labels"] != self._loss_fn.ignore_index
                 ).sum()
 
-                labels = batch.pop("labels")
-
-                with self.activations_handling_ctx:
-                    logits = self._model(**batch)
-
-                # Shift labels to compute loss
-                labels = torch.hstack(
-                    (labels[..., 1:], self.ignore_labels_cache[: labels.shape[0]])
-                )
-                if not isinstance(logits, list):
-                    labels = labels.reshape(-1)
-                    logits = logits.reshape(-1, logits.size(-1))
-
                 # Compute loss
-                val_loss = self._loss_fn(logits, labels) * current_num_tokens
+                val_loss = self._loss_step(batch) * current_num_tokens
 
                 total_val_loss += val_loss
                 total_val_tokens += current_num_tokens
@@ -850,30 +861,9 @@ class FullFinetuneRecipeDistributed(FTRecipeInterface):
                 ).sum()
                 num_tokens += current_num_tokens
 
-                # Shape [b, s], needed for the loss not the model
-                labels = batch.pop("labels")
-
-                with self.activations_handling_ctx:
-                    logits = self._model(**batch)
-
-                # Shift labels to compute loss
-                # equivalent to doing labels[..., 1:] and logits[..., :-1, :]
-                # But this way we dont need to slice the logits. We just add an ignore index to labels.
-                labels = torch.hstack(
-                    (labels[..., 1:], self.ignore_labels_cache[: labels.shape[0]])
-                )
-                if not isinstance(logits, list):
-                    labels = labels.reshape(-1)
-                    logits = logits.reshape(-1, logits.size(-1))
-
-                # Compute loss
                 # Loss is normalized by default so we multiply by the number of tokens
                 # This way we can normalize by the total number of tokens if we're accumulating gradients
-                current_loss = self._loss_fn(logits, labels) * current_num_tokens
-
-                # free logits otherwise it peaks backward memory
-                del logits
-
+                current_loss = self._loss_step(batch) * current_num_tokens
                 running_loss += current_loss
 
                 # For optimizer in backward, we need to normalize before calling backward
--- a/recipes/full_finetune_distributed.py
+++ b/recipes/full_finetune_distributed.py
@@ -355,6 +355,16 @@ class FullFinetuneRecipeDistributed(FTRecipeInterface):
             collate_fn=collate_name,
         )
 
+        # Setup validation dataloader if validation dataset is provided
+        self._val_sampler, self._val_dataloader = None, None
+        if cfg.get("dataset_validation") is not None:
+            self._val_sampler, self._val_dataloader = self._setup_data(
+                cfg_dataset=cfg.dataset_validation,
+                batch_size=cfg.batch_size,
+                collate_fn=collate_name,
+                shuffle=False,
+            )
+
         # Finally update the recipe state which can only be correctly set after all of the
         # other components have been initialized and updated.
         #
@@ -382,6 +392,10 @@ class FullFinetuneRecipeDistributed(FTRecipeInterface):
         # Set up profiler, returns DummyProfiler (nullcontext object with no-op `step` method)
         # if cfg is missing profiler key or if `cfg.profiler.enabled = False`
         self._profiler = self._setup_profiler(cfg.get(PROFILER_KEY, None))
+
+        # Add validation config parameters
+        self._run_val_every_n_steps = cfg.get("run_val_every_n_steps", None)
+        self._max_validation_batches = cfg.get("max_validation_batches", -1)
 
         # Used to ignore labels for loss computation
         self.ignore_labels_cache = torch.full(
@@ -736,6 +750,63 @@ class FullFinetuneRecipeDistributed(FTRecipeInterface):
             list(dataloader)
         return dataloader
 
+    def validate(self) -> float:
+        """
+        Run validation loop and return average validation loss.
+        """
+        if self._val_dataloader is None:
+            return 0.0
+
+        self._model.eval()
+        total_val_loss = torch.tensor(0.0, device=self._device)
+        total_val_tokens = torch.tensor(0.0, device=self._device)
+
+        with torch.no_grad():
+            for batch_idx, batch in enumerate(self._val_dataloader):
+                if (self._max_validation_batches > 0 and
+                    batch_idx >= self._max_validation_batches):
+                    break
+
+                utils.batch_to_device(batch, self._device)
+
+                # Count tokens excluding padding
+                current_num_tokens = (
+                    batch["labels"] != self._loss_fn.ignore_index
+                ).sum()
+
+                labels = batch.pop("labels")
+
+                with self.activations_handling_ctx:
+                    logits = self._model(**batch)
+
+                # Shift labels to compute loss
+                labels = torch.hstack(
+                    (labels[..., 1:], self.ignore_labels_cache[: labels.shape[0]])
+                )
+                if not isinstance(logits, list):
+                    labels = labels.reshape(-1)
+                    logits = logits.reshape(-1, logits.size(-1))
+
+                # Compute loss
+                val_loss = self._loss_fn(logits, labels) * current_num_tokens
+
+                total_val_loss += val_loss
+                total_val_tokens += current_num_tokens
+
+        # Aggregate validation metrics across all ranks
+        if torch.distributed.is_initialized():
+            torch.distributed.all_reduce(total_val_loss)
+            torch.distributed.all_reduce(total_val_tokens)
+
+        avg_val_loss = (
+            (total_val_loss / total_val_tokens).item()
+            if total_val_tokens > 0
+            else float('inf')
+        )
+
+        self._model.train()
+        return avg_val_loss
+
     def train(self) -> None:
         """
         The core training loop.
@@ -901,6 +972,18 @@ class FullFinetuneRecipeDistributed(FTRecipeInterface):
                     # Note that this is called within gradient accumulation block, hence
                     # will include multiple forward / backward passes if gradient accumulation > 1
                     self._profiler.step()
+
+                    # Run validation after gradient update
+                    if (self._run_val_every_n_steps is not None and
+                        self.global_step % self._run_val_every_n_steps == 0):
+                        pbar.refresh()
+                        val_loss = self.validate()
+                        if self._is_rank_zero:
+                            log.info(f"Validation loss: {val_loss:.4f}")
+                            self._metric_logger.log_dict(
+                                {"val_loss": val_loss},
+                                step=self.global_step,
+                            )
 
                 if (
                     (idx + 1) // self._gradient_accumulation_steps
--- a/recipes/full_finetune_distributed.py
+++ b/recipes/full_finetune_distributed.py
@@ -356,9 +356,9 @@ class FullFinetuneRecipeDistributed(FTRecipeInterface):
         )
 
         # Setup validation dataloader if validation dataset is provided
-        self._val_sampler, self._val_dataloader = None, None
+        self._val_dataloader = None
         if cfg.get("dataset_validation") is not None:
-            self._val_sampler, self._val_dataloader = self._setup_data(
+            self._val_dataloader = self._setup_data(
                 cfg_dataset=cfg.dataset_validation,
                 batch_size=cfg.batch_size,
                 collate_fn=collate_name,
@@ -787,8 +787,10 @@ class FullFinetuneRecipeDistributed(FTRecipeInterface):
 
         with torch.no_grad():
             for batch_idx, batch in enumerate(self._val_dataloader):
-                if (self._max_validation_batches > 0 and
-                    batch_idx >= self._max_validation_batches):
+                if (
+                    self._max_validation_batches > 0
+                    and batch_idx >= self._max_validation_batches
+                ):
                     break
 
                 utils.batch_to_device(batch, self._device)
@@ -812,7 +814,7 @@ class FullFinetuneRecipeDistributed(FTRecipeInterface):
         avg_val_loss = (
             (total_val_loss / total_val_tokens).item()
             if total_val_tokens > 0
-            else float('inf')
+            else float("inf")
         )
 
         self._model.train()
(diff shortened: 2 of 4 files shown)


felipemello1:
nice! Let me review it this monday

recipes/lora_finetune_distributed.py
```diff
@@ -899,6 +925,67 @@ def train(self) -> None:
...
+        total_val_loss = torch.tensor(0.0, device=self._device)
+        total_val_tokens = torch.tensor(0.0, device=self._device)
+
+        with torch.no_grad():
+            for batch_idx, batch in enumerate(self._val_dataloader):
+                if (
+                    self._max_validation_batches > 0
+                    and batch_idx >= self._max_validation_batches
+                ):
+                    break
```
felipemello1, line 945: this is ok, but if `self._max_validation_batches > 0`,  i wonder if for every epoch we are comparing the same batches. Not sure how the dataloader behaves here. We would want `shuffle=False`, which you did, and if the dataloader is an iterator, it should be restarted. Would need to confirm if thats how the torch dataloader behaves.

Maybe a simple way is to skip the training step, and confirm that we get the same eval.

recipes/configs/llama3_2/1B_qlora_validation.yaml
```diff
@@ -0,0 +1,123 @@
...
+resume_from_checkpoint: False
+save_adapter_weights_only: False
+
+# Dataset and Sampler
+dataset:
+  _component_: torchtune.datasets.alpaca_cleaned_dataset
+  packed: False  # True increases speed
+seed: null
+shuffle: True
+batch_size: 4
```
felipemello1, line 55: We prob want to have a larger batch_size for validation, since we dont need the extra memory for activations + grads. Maybe `batch_size_validation`?

Not for this PR: We need to introduce a dataloader config, to manage bsz, shuffle, etc. I am working on it in a different PR.

recipes/lora_finetune_distributed.py
```diff
@@ -899,6 +925,67 @@ def train(self) -> None:
...
+                utils.batch_to_device(batch, self._device)
+
+                # Count tokens excluding padding
+                current_num_tokens = (
+                    batch["labels"] != self._loss_fn.ignore_index
+                ).sum()
+
+                labels = batch.pop("labels")
+
+                with self.activations_handling_ctx:
```
felipemello1, line 956: We dont need to offload activations because there is no back propagation. We would need to double check if this is a no-op or if we are wasting resources by having it here. Either way, i think we can delete.

recipes/lora_finetune_distributed.py
```diff
@@ -899,6 +925,67 @@ def train(self) -> None:
...
+            torch.distributed.all_reduce(total_val_loss)
+            torch.distributed.all_reduce(total_val_tokens)
+
+        avg_val_loss = (
+            (total_val_loss / total_val_tokens).item()
+            if total_val_tokens > 0
+            else float("inf")
+        )
+
+        self._model.train()
```
felipemello1, line 986: this is ok here, but i wonder if it should be in the training loop

felipemello1 (commented):
this looks very good! Just left a few comments, let me know what you think

recipes/lora_finetune_distributed.py
felipemello1: Edit: Do we need `_max_validation_batches`? I can see how convenient it is, but i wonder if the user shouldnt just sample a smaller dataset, e.g. you can do that by using the config `dataset.split=valid[:5%]`

recipes/configs/llama3_2/1B_qlora_validation.yaml
bzz: That makes sense, but it would impede re-use of a single `_loss_step()` as-is, as it relies on `self.ignore_labels_cache` sized and setup up with `cfg.batch_size` that would either need to be different between training and validation, or skipped during the validation  🤔 

If you still think it it is worth adding it - I'll appreciate an advice on the best way to go about it: passing in a flag to loss_step(), or make recipe stateful? Have 2 caches or skip cache and slice logits?

recipes/lora_finetune_distributed.py
bzz: > Edit: Do we need `_max_validation_batches`? I can see how convenient it is, but i wonder if the user shouldnt just sample a smaller dataset, e.g. you can do that by using the config `dataset.split=valid[:5%]`

I agree, it doesn't seem necessary. It was included only because of https://github.com/pytorch/torchtune/pull/2238#pullrequestreview-2553490828

Will remove to reduce the number of new configuration parameters.

recipes/lora_finetune_distributed.py
bzz: I added it here on a premises that it's easier to reason about the model state "by inspection" when one can see it being changed before/after validation, in a single place.

Would [96bb4f4](https://github.com/pytorch/torchtune/pull/2464/commits/96bb4f48f5ede8c978c16ec5c2ee00a94df95ee3) do?

recipes/configs/llama3_2/1B_qlora_validation.yaml
felipemello1: >If you think it is worth adding it here - I'll appreciate an advice on the best way to go about it: passing in a flag to loss_step() (e.g. train=True), or add it as a state/property? Have 2 caches or check shapes & skip cache and slice logits?

good question! I dont think we need 2 caches. We can just get the max of both:

```
bsz_cache = self.batch_size if self.batch_size_validation is None else max(self.batch_size, self.batch_size_validation)
self.ignore_labels_cache = torch.full(
            (bsz_cache, 1), self._loss_fn.ignore_index, device=self._device
        )
```

Since we slice it like this later:
```
labels = torch.hstack(
                    (labels[..., 1:], self.ignore_labels_cache[: labels.shape[0]])
                )
```

felipemello1 (commented):
sorry for the delay, added a few comments. The PR is looking good! :)

recipes/lora_finetune_distributed.py
felipemello1: i see your point, i does bother me a bit that we do model.train() multiple times in the loop without necessity, and having it right after validation only does it when necessary. I think that what you had before was better, thanks for giving it a try!

recipes/full_finetune_distributed.py
```diff
@@ -903,6 +960,20 @@ def train(self) -> None:
...
+                        and self.global_step % self._run_val_every_n_steps == 0
+                    ):
+                        pbar.refresh()
+                        val_loss = self.validate()
+                        if self._is_rank_zero:
+                            log.info(f"Validation loss: {val_loss:.4f}")
+                            self._metric_logger.log_dict(
+                                {"val_loss": val_loss},
+                                step=self.global_step,
+                            )
```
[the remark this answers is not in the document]
bzz: I went with 1 in [42c2bac](https://github.com/pytorch/torchtune/pull/2464/commits/42c2bac70f46e6747c4b7003e9ca543fd45e8cb3) as it makes training loop code a bit shorter.
For 1, the only difference seems to be that .validate() will have to access self.global_step.

bzz:
Thanks for the review!

I'll address the [last item](https://github.com/pytorch/torchtune/pull/2464#discussion_r2023136718) on adding _loss_step() to lora distributed recipe tomorrow, re-run the same experiments and post new training curves.

recipes/full_finetune_distributed.py
```diff
@@ -777,6 +866,8 @@ def train(self) -> None:
                     and self._device.type == "cuda"
                 ):
                     torch.cuda.memory._record_memory_history()
+
+                self._model.train()
```
ebsmothers, line 870: Should we put this at the end of the validate method instead of here? It feels pretty redundant to call this for every single batch

recipes/full_finetune_distributed.py
```diff
@@ -743,6 +762,76 @@ def _setup_data(
...
+                ).sum()
+
+                # Compute loss
+                val_loss = self._loss_step(batch) * current_num_tokens
+
+                total_val_loss += val_loss
+                total_val_tokens += current_num_tokens
+
+        # Aggregate validation metrics across all ranks
+        if torch.distributed.is_initialized():
```
ebsmothers, line 816: Is there a case where distributed would not be initialized in this recipe?

recipes/lora_finetune_distributed.py
```diff
@@ -351,10 +362,17 @@ def setup(self, cfg: DictConfig) -> None:
         self._profiler = self._setup_profiler(cfg.get(PROFILER_KEY, None))
 
         # Used to ignore labels for loss computation
+        bzs_cache = (
```
ebsmothers, line 365:
```suggestion
        bsz_cache = (
```

recipes/full_finetune_distributed.py
felipemello1: Thats what he had initially, i suggested moving it to train loop, so the train loop would be responsible for making sure that the model is in train mode, but then it became obvious that it would be called all the time unnecessarily

recipes/lora_finetune_distributed.py
ebsmothers: The existing activation offloading hooks inherit from torch's [saved_tensors_hooks](https://github.com/pytorch/pytorch/blob/203a27e0cecce5b9050218c9d6a56c8cd2ebd0a5/torch/autograd/graph.py#L238). From the docstring:

>  In that context, the ``pack_hook`` function will be called everytime an 
> operation saves a tensor for backward (this includes intermediary results \
> saved using
> :func:`~torch.autograd.function._ContextMethodMixin.save_for_backward` but
> also those recorded by a PyTorch-defined operation). 

So if I understand correctly, the usage of no_grad should ensure that the first case does not occur. It is possible that PyTorch is saving some other intermediary results that are not strictly needed for backward but offhand I'm not sure what those would be.

ebsmothers (commented):
Thanks for adding this! A few small comments, but other than that no major concerns from me. Looking forward to finally having proper validation support. As a follow-up we should think about how we can pull out `validate` into a general utility that can be used across all our recipes

recipes/full_finetune_distributed.py
ebsmothers: Ah I see, sorry for the thrash then. Personally I prefer treating the val loop like a context manager wrt the model but won't block on it

felipemello1: i think its fine to revert it to the validation method or put at right after the validate call. Cool with moving it back to validate, since both of you had the same idea

recipes/full_finetune_distributed.py
bzz: Addressed in [5d23a06](https://github.com/pytorch/torchtune/pull/2464/commits/5d23a06305c1d85040d7b0f9bfa76d95ada6f214)

recipes/lora_finetune_distributed.py
bzz: Thanks, addressed in [4436592](https://github.com/pytorch/torchtune/pull/2464/commits/4436592dd01a9fa329b0ce65b0e529e869240352)

recipes/full_finetune_distributed.py
bzz: Moved model.train() back to .validate() in [eb22bb6](https://github.com/pytorch/torchtune/pull/2464/commits/eb22bb6d2e0ace1ad9e12198bf21a0749ef311fa).

Let me know if that works, otherwise will be happy to move both into train e.g.

```
self._model.eval()
self.validate()
self._model.train()
```

bzz:
lora

```
        run_val_every_n_steps=100
	batch_size=16 
	batch_size_validation=64 
```

Screenshot 2025-04-02 at 23 27 33

bzz:
@ebsmothers @felipemello1 thanks for prompt reviews, ready for another round.

recipes/configs/llama3_2/1B_qlora_validation.yaml
```diff
@@ -0,0 +1,123 @@
...
+
+# Validation
+dataset_validation:
```
joecummings, line 58: super nit: let's be consistent between val and validation. I vote for val, but don't feel strongly.

recipes/lora_finetune_distributed.py
```diff
@@ -894,6 +899,77 @@ def train(self) -> None:
...
+        # free logits otherwise it peaks backward memory
+        del logits
+
+        return loss
+
+    def validate(self) -> Dict[str, float]:
+        """
+        Run validation loop and return average validation loss.
+        """
+        if self._val_dataloader is None:
```
joecummings, line 930: I'd rather throw an error upfront if the user tries to specify a "run_val_every_n_steps" without specifying a valdiation_dataset. This block of code should never be hit unless they're definitely doing validation.

joecummings (approved):
I'll let @felipemello1 and @ebsmothers  do a final approval, but this looks great - only a couple suggestions.

Overall, I'm curious if we've tried compiling the validation step or this new larger loss step to see any speedups?

felipemello1:
will do a last pass this afternoon. Thank you!

recipes/configs/llama3_2/1B_qlora_validation.yaml
felipemello1, line 1: I would prefer to avoid adding a config just for validation. I think we should add the validation to maybe all of the llama 3_2 full/lora distributed, maybe with default `run_val_every_n_steps=None # Change to an integer to enable validation every N steps`. Wdyt?

felipemello1 (commented):
the PR looks ready to go! There are just a few minor comments from Joe, and changing the config. There would be extra extra points if we had an unit test, but i am ok with adding it as a follow up task. Lets get it merged :)

recipes/configs/llama3_2/1B_qlora_validation.yaml
bzz: Sounds good. Would changing config for every supported recipe in 3_2 to something like
```
# Dataset and Sampler
dataset:
  packed: False  # True increases speed
  split: train[:95%]
seed: null
shuffle: True
batch_size: 4

# Validation
run_val_every_n_steps=None # Change to an integer to enable validation every N steps
dataset_val:
  split: train[95%:]
batch_size_val: 16
```

work?

I just realize that that `  source: mhenrichsen/alpaca_2k_test` that was included is actually part of the train 😮‍💨  (and an example encouraging the use of training data for validation can be.. misleading)

felipemello1: does alpaca have a "valid" split?

But that works! I also like how you are demonstrating the use of [:95%].

bzz: Sadly it doesn't. And yes, a lack of example where exactly to put the % was the reason to add that other dataset.

recipes/configs/llama3_2/1B_qlora_validation.yaml
bzz: Addressed in [368cf7c](https://github.com/pytorch/torchtune/pull/2464/commits/368cf7c619fbc5f02b355609fa77c0e0f3b4d3b8)

recipes/lora_finetune_distributed.py
bzz: I found myself setting `run_val_every_n_steps` in config but then overriding `valdiation_dataset` to null from CLI (as a forgetful user who decided that's a way to disable validation temporarily).
But failing fast with an error message sounds alright (although I wish those errors were more readable in distributed cases).

Addressed in [6b15365](https://github.com/pytorch/torchtune/pull/2464/commits/6b15365413187e9065ad8fe7694d7fe7ffdbfe3a)

bzz:
I believe all the feedback (but a unit test) should be addressed now. 

@joecummings 

> Overall, I'm curious if we've tried compiling the validation step or this new larger loss step to see any speedups?

I'm not sure I follow - do you suggest to compare training times with validation loop enabled between compile/no compile experiments to see the speedup?

felipemello1 (approved):
LGTM! Awesome work! thank you so much for implementing this. I started the tests. Once CI is green, i will merge.

recipes/configs/llama3_2/1B_qlora_validation.yaml
mariozupan: This isn't clear. Shouldn't be the colon in yaml file  "run_val_every_n_steps=None". However still I can't get val loss with my yaml:
# Dataset and Sampler
dataset:
  packed: False  # True increases speed
  split: train[:95%]
seed: null
shuffle: True
batch_size: 4

# Validation
run_val_every_n_steps: None  # Change to an integer to enable validation every N steps
#max_validation_batches: -1
dataset_val:
  split: train[95%:]
batch_size_validation: 8


# Fine-tuning arguments
epochs: 2
max_steps_per_epoch: 300
...
Name: torchtune
Version: 0.5.0

mariozupan:
Check this: pip install -U git+https://github.com/bzz/torchtune.git@validate_distributed
Could  you clarify the procedure?

PR #2467, opened 2025-03-09
(draft/discussion) GRPO LoRA

ianbarber:
#### Context
- [x ] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

#2421 - exploring a LoRA recipe. 

#### Changelog
* Add in new configs and recipe for lora distributed (though tested on a single device)
* Futz with the checkpointer

Putting this up early to get some input. Following the pattern from @RedTachyon I added the SFT config and the GRPO recipe. One immediate challenge was that checkpointing loading for adapters is only expected to be used when resuming: this case is more like a pipeline with SFT to lora first then a separate recipe with the same lora adapters. I ended up hacking around the checkpoint loading issues, but it doesn't feel particularly clean. 

On the positive side, by reducing the number of generations I got the recipe running in a single 24GB card, which is a good baseline for ease of access, not the fastest thing in the world, but it goes! 

#### Test plan

#### UX

### DIFF
--- a/recipes/configs/dev/3B_lora_grpo.yaml
+++ b/recipes/configs/dev/3B_lora_grpo.yaml
@@ -1,0 +1,129 @@
+# Config for multi-node GRPO in dev/grpo_lora_finetune_distributed.py
+# using a Llama3.2 3B Base model
+#
+# This config assumes that you've run the following command before launching
+# this run:
+#   tune download meta-llama/Llama-3.2-3B --output-dir /tmp/Llama-3.2-3B --ignore-patterns "original/consolidated.00.pth"
+#
+# It can be beneficial to first train the base model with SFT using the 3B_sft recipe.
+#
+# To launch on 1 device, run the following command from root:
+# tune run --nproc_per_node 1 dev/grpo_lora_finetune_distributed --config dev/3B_lora_grpo
+#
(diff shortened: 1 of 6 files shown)


ianbarber:
Updated this - I validated the recipe for single device, so figured it was just cleaner to do a single device diff. I reverted the checkpointing changes as the LoRA from SFT gets merged in and its cleaner just start it again. The recipe gets to a decent success rate during training - testing on 4 epochs running the gms8k task from lm evals gives:

```
|Tasks|Version|     Filter     |n-shot|  Metric   |   |Value |   |Stderr|
|-----|------:|----------------|-----:|-----------|---|-----:|---|-----:|
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.5436|±  |0.0137|
|     |       |strict-match    |     5|exact_match|↑  |0.5421|±  |0.0137|
```
This is (clearly) eval-on-the-training data, but for reference the base llama 3B gets 0.2646 on flexible, and after one epoch .48. I was a little dubious whether LoRA would work here and it appears it does!

pbontrager:
I just saw this PR has been hanging here since March! @ianbarber is this recipe not ready for review? Or did we just miss it?

PR #2472, opened 2025-03-10
Add missing `--output_dir` to config download commands

SalmanMohammadi:
```bash
$ grep -r --include="*.yaml" "tune download" . | grep -v "output-dir"
...
$ grep -r --include="*.yaml" "tune download" . | grep -v "output-dir" | wc -l
33
```


SalmanMohammadi:
> Should we put --output-dir before the hf-token?

you got it boss

felipemello1 (approved):

Issue #2460, opened 2025-03-05
No pad_id passed in _LLMEvalWrapper


xueyan-lii:
In recipes/eleuther_eval.py, _LLMEvalWrapper's line [411](https://github.com/pytorch/torchtune/blob/cf0142bd2eb5eabe33a3d43eca4f07ec48a8749f/recipes/eleuther_eval.py#L411), _tokenizer.pad_id is not passed in the generate function which means [torchtune/generation/_generate.py](https://github.com/pytorch/torchtune/blob/cf0142bd2eb5eabe33a3d43eca4f07ec48a8749f/torchtune/generation/_generation.py#L194) will always set pad_id to default 0. Is this expected behavior?

SalmanMohammadi:
I think @RedTachyon raised this in Discord too. We should update the eval recipe to pass in the correct token.

krammnic:
It is an extremely simple fix, but it became a little bit annoying. Let me open PR then, because I don't see the opened one for this...

PR #2476, opened 2025-03-11
Pass pad_id, to the generate

krammnic:
#### Context
- [ ] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

#### Changelog
* #2460

#### Test plan

#### UX

### DIFF
--- a/recipes/eleuther_eval.py
+++ b/recipes/eleuther_eval.py
@@ -315,7 +315,7 @@ class _LLMEvalWrapper(HFLM):
         self._batch_size = batch_size
         self._dtype = dtype
         self._enable_kv_cache = enable_kv_cache
-        
+
         if hasattr(self._tokenizer, "pad_id") and self._tokenizer.pad_id is not None:
             self.pad_id = self._tokenizer.pad_id
         else:
@@ -413,7 +413,6 @@ class _LLMEvalWrapper(HFLM):
             dtype=self._dtype,
             decoder_max_seq_len=self.max_length,
         ):
-           
 
             toks, _ = generate(
                 self.model,
--- a/recipes/eleuther_eval.py
+++ b/recipes/eleuther_eval.py
@@ -408,12 +408,18 @@ class _LLMEvalWrapper(HFLM):
             dtype=self._dtype,
             decoder_max_seq_len=self.max_length,
         ):
+            if hasattr(self.tokenizer, "pad_id") and self.tokenizer.pad_id is not None:
+                pad_id = self.tokenizer.pad_id
+            else:
+                pad_id = 0
+
             toks, _ = generate(
                 self.model,
                 maybe_padded_context,
                 max_generated_tokens=self.max_gen_toks,
                 temperature=temperature,
                 top_k=None,
+                pad_id=pad_id,
                 stop_tokens=self._tokenizer.stop_tokens,
             )
         return toks[:bsz]
--- a/recipes/eleuther_eval.py
+++ b/recipes/eleuther_eval.py
@@ -316,11 +316,6 @@ class _LLMEvalWrapper(HFLM):
         self._dtype = dtype
         self._enable_kv_cache = enable_kv_cache
 
-        if hasattr(self._tokenizer, "pad_id") and self._tokenizer.pad_id is not None:
-            self.pad_id = self._tokenizer.pad_id
-        else:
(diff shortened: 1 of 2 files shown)


recipes/eleuther_eval.py
```diff
@@ -408,12 +408,18 @@ def _model_generate(
             dtype=self._dtype,
             decoder_max_seq_len=self.max_length,
         ):
+            if hasattr(self.tokenizer, "pad_id") and self.tokenizer.pad_id is not None:
+                pad_id = self.tokenizer.pad_id
+            else:
+                pad_id = 0
```
felipemello1, line 414: is it `self.tokenizer` or `self._tokenizer`? This also can be in the __init__, since its a constant.

Can you please test it to confirm it works? thank you!

recipes/eleuther_eval.py
```diff
@@ -408,12 +408,18 @@ def _model_generate(
...
+
             toks, _ = generate(
```
felipemello1, line 416: do we also need changes in `_model_multimodal_generate`?

krammnic: I think we don't as we don't use `generate` there

krammnic:
It seems to me that both wrappers are not really covered with tests

krammnic:
`TestEleutherEval` has passed locally

felipemello1:
> `TestEleutherEval` has passed locally

can you please run the recipe?

recipes/eleuther_eval.py
felipemello1: true, but we are still generating. Does it means that no padding is happening or the arg is just not exposed?

krammnic:
> > `TestEleutherEval` has passed locally
> 
> can you please run the recipe?

It is fine: 

Max memory allocated: 9.61 GB


|    Tasks     |Version|Filter|n-shot|Metric|   |Value |   |Stderr|
|--------------|------:|------|-----:|------|---|-----:|---|-----:|
|truthfulqa_mc2|      2|none  |     0|acc   |↑  |0.4175|±  |0.0146|

Furthermore, I've noticed that qwen2_5 eval config is incorrect, so pushed fix for it here.

recipes/eleuther_eval.py
```diff
@@ -316,6 +316,11 @@ def __init__(
         self._dtype = dtype
         self._enable_kv_cache = enable_kv_cache
 
+        if hasattr(self._tokenizer, "pad_id") and self._tokenizer.pad_id is not None:
```
pbontrager, line 319: We really need to an an abstract class for our tokenizers, but I think it's safe to assume in the meantime that tokenizer.pad_id exists for each tokenizer as our recipes use this. I think we can skip this hasattr check.

recipes/eleuther_eval.py
pbontrager: padded_collate_tiled_images_and_mask used in the _VLMEvalWrapper also takes padding_idx as an input

recipes/eleuther_eval.py
```diff
@@ -361,7 +366,7 @@ def tok_batch_encode(
         x = left_pad_sequence(
             [torch.tensor(x) for x in tokenized_text],
             batch_first=True,
-            padding_value=self._tokenizer.pad_id,
+            padding_value=self.pad_id,
```
pbontrager, line 369: I actually think self._tokenizer.pad_id is correct here. I don't think this is the location where the issue is.

recipes/eleuther_eval.py
pbontrager: @felipemello1 we don't do batched inference for multimodal

recipes/eleuther_eval.py
krammnic: Then it will be inconsistent. As the solution, I assume that we can just not define it in `__init__`, dropping `hasattr` check according to the previous comment, but we still need `is not None` check don't we?

felipemello1: since .pad_id is in every tokenizer, it makes sense to remove it from the __init__, without hasattr, and just doing `self._tokenizer.pad_id` directly. Thanks for asking!

After it I will merge

krammnic:
@felipemello1 lint is fine + integration test passed, so we can start CI

felipemello1 (approved):

Issue #2459, opened 2025-03-05
No chat template in evaluation


xueyan-lii:
In LLAMA3 tokenizer, chat template is applied with the option to add system instruction. However, when using eleuther_eval.py, the tokenization process do not apply chat template. How can this be done?

felipemello1:
@duduyi2013 , mind taking a look?

krammnic:
@felipemello1 It should be simple fix. We need to expose `apply_chat_template` argument in our wrapper (notice that it is not bool!). According to lm_eval: `chat_template=getattr(lm, "apply_chat_template")`.

pbontrager:
I explored using torchtune's apply_chat_template [here](#2482) but since Eleuther does it's own formatting, they'd conflict with each other. To echo what @krammnic already said, we need to 
- override  HFLM's "apply_chat_template" method in _LLMEvalWrapper shown [here](https://github.com/EleutherAI/lm-evaluation-harness/blob/0f944779d5f37f74480664da71a4ee301f62d013/lm_eval/models/huggingface.py#L1393). This should stay mostly the same but grab the chat_template function that's inside the torchtune tokenizer [here](https://github.com/pytorch/torchtune/blob/dab36d2e5dcc2192c93172f9236cda2c8b6ca41c/torchtune/models/llama3/_tokenizer.py#L301).
- We also need to verify that the apply_chat_template from HF tokenizer is doing the same thing as torchtune's apply template.

joecummings:
> [@pbontrager](https://github.com/pbontrager) Can I take this now?

fire the lasers

PR #2482, opened 2025-03-11
Update Eleuther Eval to use Message format

pbontrager:
#### Context
- [ ] add a new feature
- [x] fix a bug
- [ ] other (please add here)

Please link to any issues this PR addresses. #2459 

#### Changelog
* Update tok_encode in eluther_eval _LLMEvalWrapper to wrap text as messages and use the tokenizer __calll__ method with inference=True
* Update eleuther eval test for new accuracy figure

#### Test plan
**This is a backwards breaking change** in that the eval numbers will look different. By using the __call__ method you get properly formatted text based on the model tokenizer (including any prompt template). But before now, we were running text evals without any message formatting. This should make eval numbers go up for instruct models but potentially go down for base models, as seen in the regression test.

[TODO] manually run eval with an instruct model to see if accuracy goes up

### DIFF
--- a/recipes/eleuther_eval.py
+++ b/recipes/eleuther_eval.py
@@ -345,12 +345,13 @@ class _LLMEvalWrapper(HFLM):
         return self._enable_kv_cache
 
     def tok_encode(self, text: str, **kwargs) -> List[int]:
-        # Note on add_bos flag: setting to False as this gives better results, for example
-        # +1% on truthfulqa_mc2 with a LoRA finetune. lit-gpt also sets this to False,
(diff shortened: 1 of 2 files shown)


ebsmothers (approved):
Pending your TODO this looks good to me. Let's run a proper eval with both base and instruct model just to verify your hypothesis here

pbontrager:
Digging deeper into Eleuther internals, it appears that Eleuther handles its own formatting and doesn't consistently add the message formatting. If we apply message formatting it would be wrapped around what they're doing and often it would wrap multiple question/answers as a single message. Therefore I'm closing this direction as formatting needs to be handled within Eleuther.

PR #2486, opened 2025-03-12
Fix MPS `get_device`

SalmanMohammadi:
#### Context
- [ ] add a new feature
- [x] fix a bug
- [ ] other (please add here)

On main:
```python
>>> from torchtune.utils import get_device
>>> get_device("mps")
  File "<stdin>", line 1, in <module>
  File "/Users/salmanmohammadi/projects/torchtune/torchtune/utils/_device.py", line 170, in get_device
    device = _setup_device(device)
             ^^^^^^^^^^^^^^^^^^^^^
  File "/Users/salmanmohammadi/projects/torchtune/torchtune/utils/_device.py", line 89, in _setup_device
    torch_device.set_device(device)
    ^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: module 'torch.mps' has no attribute 'set_device'
>>> 
```

On this branch:
```python
>>> from torchtune.utils import get_device
>>> get_device("mps")
device(type='mps')
>>> import torch
>>> torch.ones((3,3)).to(device)
tensor([[1., 1., 1.],
        [1., 1., 1.],
        [1., 1., 1.]], device='mps:0')
```

### DIFF
--- a/torchtune/utils/_device.py
+++ b/torchtune/utils/_device.py
@@ -67,8 +67,8 @@ def _setup_device(device: torch.device) -> torch.device:
         device (torch.device): The device to set.
 
     Raises:
-        RuntimeError: If device index is not available, or if ``set_device`` is not
-            supported for the device type (e.g. on MPS).
+        RuntimeError: If device index is not available.
+        AttributeError: If ``set_device`` is not supported for the device type (e.g. on MPS).
 
     Returns:
         device
@@ -86,6 +86,10 @@ def _setup_device(device: torch.device) -> torch.device:
     if device.index >= torch_device.device_count():
         raise RuntimeError(
             f"The local rank is larger than the number of available {device_name}s."
+        )
+    if not hasattr(torch_device, "set_device"):
+        raise AttributeError(
+            f"The device type {device_type} does not support the `set_device` method."
         )
     torch_device.set_device(device)
     return device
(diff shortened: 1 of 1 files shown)


felipemello1 (commented):
This is a bit confusing to me. You removed 'mps' from 'get_device' docstring, but in your example you show how 'get_device("mps")' works now.

Should we use 'get_device(mps)'? If so, i think we should keep it in the args and add a small note on why we cant do `device = _setup_device('mps')`.

Maybe `_setup_device` needs fixing to also work with mps?

SalmanMohammadi:
> This is a bit confusing to me. You removed 'mps' from 'get_device' docstring, but in your example you show how 'get_device("mps")' works now.

Ah, good catch. I shouldn't have removed it from the docstring. 

> Should we use 'get_device(mps)'? If so, i think we should keep it in the args and add a small note on why we cant do `device = _setup_device('mps')`.

Agreed!

> Maybe `_setup_device` needs fixing to also work with mps?

CUDA/XPU/NPU backends all support `set_device` (presumably for multi-device accelerator setups). This isn't supported on MPS. I've indicated this in the docstring now.

felipemello1 (approved):
I am fine with merging it like this. Another solution would be to **not** hardcode the device list, i.e. `if device.type in ["cuda", "npu", "xpu"]:`

Instead, just check `hasattr(torch_device, "set_device")`, before calling `torch_device.set_device(device)`

This would be easier for future devices, but this would be less explicit. Since i dont think we will add many more device types, its probably ok to just merge as is

PR #2441, opened 2025-02-26
[WIP] Add `StatefulDataLoader` to all recipes except knowledge_single

krammnic:
#### Context
What is the purpose of this PR? Is it to
- [X] add a new feature
- [ ] fix a bug
- [ ] update tests and/or documentation
- [ ] other (please add here)

Please link to any issues this PR addresses.
#2431  
#2439

#### Changelog
What are the changes made in this PR?
Added `StatefulDataLoader` to all recipes except `knowledge_single` (this recipe will be updated by @jxtngx)

Notice, that I haven't done verifying runs yet. Will add them gradually asap.

#### Test plan
Please make sure to do each of the following if applicable to your PR. If you're unsure about any one of these just ask and we will happily help. We also have a [contributing page](https://github.com/pytorch/torchtune/blob/main/CONTRIBUTING.md) for some guidance on contributing.

- [X] run pre-commit hooks and linters (make sure you've first installed via `pre-commit install`)
- [ ] add [unit tests](https://github.com/pytorch/torchtune/tree/main/tests/torchtune) for any new functionality
- [ ] update [docstrings](https://github.com/pytorch/torchtune/tree/main/docs/source) for any new or updated methods or classes
- [ ] run unit tests via `pytest tests`
- [ ] run recipe tests via `pytest tests -m integration_test`
- [ ] manually run any new or modified recipes with sufficient proof of correctness
- [ ] include relevant commands and any other artifacts in this summary (pastes of loss curves, eval results, etc.)

#### UX
If your function changed a public API, please add a dummy example of what the user experience will look like when calling it.
Here is a [docstring example](https://github.com/pytorch/torchtune/blob/6a7951f1cdd0b56a9746ef5935106989415f50e3/torchtune/modules/vision_transformer.py#L285)
and a [tutorial example](https://pytorch.org/torchtune/main/tutorials/qat_finetune.html#applying-qat-to-llama3-models)

- [ ] I did not change any public API
- [ ] I have added an example to docs or docstrings

PR #2490, opened 2025-03-12
Remove dataloader state dict loading

felipemello1:
#### Context
- [ ] add a new feature
- [x] fix a bug
- [ ] other (please add here)

Our current dataloader is epoch based. We do not need to load the state dict, as long as we call 'set_epoch', which sets the seed to the epoch number + original seed:
```
            self._dataloader.sampler.set_epoch(curr_epoch)
```

The current implementation loads the state dict, then does a full pass on the dataloader, i.e. list(dataloader). For epoch based ckpt, this doesnt help, but make resuming extremely slow, since you have to go over all of it. 

For step based ckpt, we do NOT need list(dataloader). We just need to load the ckpt. 

NOTE: Loading ckpt state dict will need to be reintroduced when we change to step based ckpt.

#### Test plan

unit tests + running multimodal. Notice that grads are not the same, but what matters for this PR is that the batches are the same, and they are, given the loss shape.

```
tune run --nproc_per_node 2 lora_finetune_distributed --config llama3_2_vision/11B_lora epochs=2 max_steps_per_epoch=20 metric_logger=torchtune.training.metric_logging.WandBLogger resume_from_checkpoint=True checkpointer.adapter_checkpoint="epoch_0/adapter_model.pt"
```

image

```
tune run full_finetune_single_device --config llama3_2/1B_full_single_device epochs=2 max_steps_per_epoch=20 optimizer=torch.optim.SGDresume_from_checkpoint=True checkpointer.checkpoint_files=["epoch_0/model-00001-of-00001.safetensors"] 
```
image

### DIFF
--- a/recipes/full_finetune_distributed.py
+++ b/recipes/full_finetune_distributed.py
@@ -729,11 +729,7 @@ class FullFinetuneRecipeDistributed(FTRecipeInterface):
             # dropping last avoids shape issues with compile + flex attention
             drop_last=True,
         )
-        if dataloader_state_dict is not None:
-            dataloader.load_state_dict(dataloader_state_dict)
-            # B/c we currently only save at epoch boundaries, if we cut the previous epoch short
-            # we need to force the dataloader to finish the last iteration before it's actually used
-            list(dataloader)
+
         return dataloader
 
     def train(self) -> None:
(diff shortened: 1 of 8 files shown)


felipemello1:
cc: @krammnic @ankitageorge @ebsmothers @joecummings @pbontrager

krammnic:
@felipemello1 Looks like I need to spread this on #2431

felipemello1:
> @felipemello1 Looks like I need to spread this on #2431

I guess you meant #2441? yeah, thats why I tagged  you :D

ebsmothers (approved):

PR #2491, opened 2025-03-12
Move multidata to dev + fix bugs

felipemello1:
#### Context
- [ ] add a new feature
- [x] fix a bug
- [ ] other (please add here)

#### Changelog
- moved recipe to dev
- Updated checkpointing, to make it work

ISSUES: The dataloader does NOT resume from ckpt. This means that the order of the shuffle for a new epoch will be different vs checkpoint. This is NOT a blocking issue. When moving to step based checkpoint, we can resume the dataloader from ckpt. This is currently not possible because `loader` does not have method "set_epoch". We could go around it by adding StatefulDataLoader, but this is beyond the scope of this PR.

#### Test plan

```
tune run --nproc_per_node 2 dev/lora_finetune_distributed_multi_dataset --config dev/11B_lora_multi_dataset epochs=2 max_steps_per_epoch=50 metric_logger=torchtune.training.metric_logging.WandBLogger
```

image

### DIFF
--- a/recipes/dev/lora_finetune_distributed_multi_dataset.py
+++ b/recipes/dev/lora_finetune_distributed_multi_dataset.py
@@ -314,6 +314,11 @@ class LoRAFinetuneRecipeDistributed(FTRecipeInterface):
             cfg_dataloader=cfg.dataloader,
             cfg_datasets=cfg.datasets,
             batch_size=cfg.batch_size,
+            dataloader_state_dict=(
+                checkpoint_dict[training.DATALOADER_KEY]
+                if self._resume_from_checkpoint
+                else None
+            ),
         )
 
         # Finally update the recipe state which can only be correctly set after all of the
@@ -571,6 +576,7 @@ class LoRAFinetuneRecipeDistributed(FTRecipeInterface):
         cfg_dataloader: DictConfig,
         cfg_datasets: ListConfig,
         batch_size: int,
+        dataloader_state_dict: Optional[Dict[str, Any]] = None,
     ) -> Loader:
         """
         Torchdata related setup happens here. Currently this recipe supports
--- a/recipes/dev/lora_finetune_distributed_multi_dataset.py
+++ b/recipes/dev/lora_finetune_distributed_multi_dataset.py
@@ -18,10 +18,10 @@ from torch import nn
 from torch.distributed import destroy_process_group, init_process_group
 
 from torch.optim import Optimizer
-
 from torchdata.nodes import Loader, StopCriteria
 from torchtune import config, modules, training, utils
 from torchtune.config._utils import _get_component_from_path
+
 from torchtune.data import padded_collate_packed
 from torchtune.data._utils import get_dataloader, get_multi_dataset, load_hf_dataset
 from torchtune.datasets._sft import SFTTransform
@@ -37,7 +37,6 @@ from torchtune.modules.peft import (
 )
 from torchtune.recipe_interfaces import FTRecipeInterface
 from torchtune.training import DummyProfiler, PROFILER_KEY
-
 from tqdm import tqdm
 
 log = utils.get_logger("DEBUG")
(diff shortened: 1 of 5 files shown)


recipes/dev/lora_finetune_distributed_multi_dataset.py
```diff
@@ -489,11 +489,9 @@ def _setup_model(
                 ) and not lora_weights_state_dict:
                     # lora may not be covered in state dict
                     # if finetune for the 1st time
-                    m.lora_a.to_empty(device=lora_device)
-                    m.lora_b.to_empty(device=lora_device)
                     m.to_empty(device=lora_device)
                     m.initialize_parameters()
-                # RoPE is not covered in state dict
```
felipemello1, line 496: copied from full_finetune_distributed

recipes/dev/lora_finetune_distributed_multi_dataset.py
```diff
@@ -689,14 +689,11 @@ def save_checkpoint(
...
         # To prevent GPU memory from spiking during checkpoint save,
         # we consolidate the full model and optim state dicts on CPU for rank 0
-        state_dict = self._model.state_dict()
-        if self._save_adapter_weights_only:
-            state_dict = get_adapter_state_dict(state_dict, device=None)
-
         cpu_state_dict = training.gather_cpu_state_dict(
-            state_dict,
+            self._model,
             self._is_rank_zero,
```
felipemello1, line 694: copied from full_finetune_distributed

torchtune/_recipe_registry.py
```diff
@@ -438,12 +438,12 @@ class Recipe:
...
-        name="lora_finetune_distributed_multi_dataset",
-        file_path="lora_finetune_distributed_multi_dataset.py",
+        name="dev/lora_finetune_distributed_multi_dataset",
+        file_path="dev/lora_finetune_distributed_multi_dataset.py",
         configs=[
             Config(
-                name="llama3_2_vision/11B_lora_multi_dataset",
-                file_path="llama3_2_vision/11B_lora_multi_dataset.yaml",
+                name="dev/11B_lora_multi_dataset",
+                file_path="dev/11B_lora_multi_dataset.yaml",
```
ebsmothers, line 446: nit: it might be worth keeping some version of `llama3_2_vision` in the filename (otherwise it's not immediately clear what this config is)

recipes/dev/lora_finetune_distributed_multi_dataset.py
```diff
@@ -314,6 +314,11 @@ def setup(self, cfg: DictConfig) -> None:
             cfg_dataloader=cfg.dataloader,
             cfg_datasets=cfg.datasets,
             batch_size=cfg.batch_size,
+            dataloader_state_dict=(
+                checkpoint_dict[training.DATALOADER_KEY]
+                if self._resume_from_checkpoint
+                else None
+            ),
```
ebsmothers, line 321: Shouldn't this be removed?

ebsmothers (approved):
A couple small comments, otherwise looks good

torchtune/_recipe_registry.py
felipemello1: i agree with the logic, but its not the pattern we use for any configs in /dev. Change it still?

recipes/dev/lora_finetune_distributed_multi_dataset.py
felipemello1: i thought about it. Decided to leave it as a placeholder for when we merge step based ckpt, but it can be deceiving that we save the state dict but dont use it. It is just a small dict though.

I can remove if you think its better

ebsmothers: Personally I would remove, but it's not a blocker

torchtune/_recipe_registry.py
ebsmothers: OK then it's fine to leave as is

PR #2494, opened 2025-03-13
Use DCP in HFCheckpointer to read/write directly to HuggingFace

ankitageorge:
#### Context
- [ x] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

#### Changelog
* Add the ability to read and write checkpoints directly in HF format to HF, behind the enable_dcp flag. 

#### Test plan

- [ x] add [unit tests](https://github.com/pytorch/torchtune/tree/main/tests/torchtune) for any new functionality
- [ x] update [docstrings](https://github.com/pytorch/torchtune/tree/main/docs/source) for any new or updated methods or classes
- [ x] run recipe tests via `pytest tests -m integration_test`
- [ x] manually run any new or modified recipes with sufficient proof of correctness
- [ x] include relevant commands and any other artifacts in this summary (pastes of loss curves, eval results, etc.)

#### UX

- [x ] I have added an example to docs or docstrings


torchtune/training/checkpointing/_utils.py
```diff
@@ -581,15 +600,21 @@ def validate_checkpoint_files(
...
+) -> bool:
     """
     Checks that the output directory is not equal to or a subdirectory of the checkpoint directory.
     This is necessary to avoid making copies of copies when geting config files from ckpt_dir.
     """
-
     # Resolve the absolute paths to avoid issues with relative paths
-    _ckpt_dir = ckpt_dir.resolve()
-    _out_dir = out_dir.resolve()
+    if isinstance(ckpt_dir, Path):
```
joecummings, line 611: Should probably just convert to path to start and then resolve.

torchtune/training/checkpointing/_utils.py
```diff
@@ -236,22 +241,31 @@ def safe_torch_load(
...
-                weights_only=weights_only,
-            )
+            if isinstance(fs, HfFileSystem):
+                # HfFileSystem does not support mmap
+                mmap = False
+                with fs.open(checkpoint_path, "rb") as checkpoint_file:
+                    state_dict = torch.load(
+                        checkpoint_file,
+                        map_location="cpu",
+                        mmap=mmap,
```
joecummings, line 259: Will this be much slower without mmap then?

joecummings (commented):
Very small comments - would you mind just manually verifying that you can download directly from a HF Hub repo and upload to a new Repo as a test?

torchtune/training/checkpointing/_utils.py
ankitageorge: for remote paths this won't work. they don't have a resolve method, that's why I did this. if it's a remote path like hf, it would be a string and not call resolve

torchtune/training/checkpointing/_utils.py
ankitageorge: yes, for large files. But this code path would be for the extra files that don't go through dcp like, adapter checkpoint and recipe checkpoint which would be smaller files right?

ankitageorge:
> Very small comments - would you mind just manually verifying that you can download directly from a HF Hub repo and upload to a new Repo as a test?

Yes, this is working

PR #2495, opened 2025-03-13
Phi4 Updates

pbontrager:
#### Context
- [ ] add a new feature
- [x] fix a bug
- [ ] other (please add here)

Fixed some config bugs, added docs, and updated naming.

#### Changelog
* phi4_14b_tokenizer -> phi4_tokenizer
* PHI4MiniTokenizer -> Phi4Tokenizer
* added PHI4 model type
* added model to the docs
* fixed bad evaluation configs and added evaluation to registery

#### Test plan

Ran all configs to ensure they start

### DIFF
--- a/docs/source/api_ref_models.rst
+++ b/docs/source/api_ref_models.rst
@@ -299,6 +299,26 @@ To download the Qwen2 1.5B model, for example:
     qwen2.lora_qwen2_7b
     qwen2.qwen2_tokenizer
 
+phi-4
+-----
+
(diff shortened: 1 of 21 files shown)


ebsmothers (approved):
While we're at it, thoughts on renaming the configs to include the model size in their filenames? I know there's only one model size but it does feel a bit weird to see just `full.yaml` when all our other configs would be e.g. `14B_full.yaml`. Stamping to unblock

PR #2496, opened 2025-03-13
update teacher checkpointer paths in KD config

ebsmothers:
Fix checkpoint path in one of our KD configs

### DIFF
--- a/recipes/configs/llama3_2/8B_to_1B_KD_lora_distributed.yaml
+++ b/recipes/configs/llama3_2/8B_to_1B_KD_lora_distributed.yaml
@@ -51,7 +51,7 @@ save_adapter_weights_only: False
 # Teacher checkpoint
 teacher_checkpointer:
   _component_: torchtune.training.FullModelHFCheckpointer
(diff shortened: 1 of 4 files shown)


recipes/configs/llama3_2/8B_to_1B_KD_lora_distributed.yaml
```diff
@@ -51,7 +51,7 @@ save_adapter_weights_only: False
 # Teacher checkpoint
 teacher_checkpointer:
   _component_: torchtune.training.FullModelHFCheckpointer
-  checkpoint_dir: /tmp/Meta-Llama-3.1-8B-Instruct/
+  checkpoint_dir: /tmp/torchtune/llama3_1_8B/lora/epoch_0
```
pbontrager, line 54: Why is epoch 0 the right default here?

pbontrager (commented):
I'm uncomfortable with this PR because it's making a lot of assumptions about where things are saved during fine-tuning that might break in the future. It also makes the recipe dependent on having run another recipe which is new. Could you add more context on how common it is to finetune the teacher first?

felipemello1 (approved):

PR #2498, opened 2025-03-13
Update concat dataset example

ebsmothers:
Update example in ConcatDataset to something that works (and also uses packed dataset now that we support it)

### DIFF
--- a/torchtune/datasets/_concat.py
+++ b/torchtune/datasets/_concat.py
@@ -54,11 +54,11 @@ class ConcatDataset(Dataset):
             source: vicgalle/alpaca-gpt4
             split: train
             train_on_input: True
-          - _component_: torchtune.datasets.instruct_dataset
-            source: samsum
-            column_map: {"output": "summary"}
-            split: train
+            packed: True
+          - _component_: torchtune.datasets.grammar_dataset
+            split: train[:1%]
             train_on_input: False
+            packed: True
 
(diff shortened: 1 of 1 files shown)


torchtune/datasets/_concat.py
```diff
@@ -54,11 +54,11 @@ class ConcatDataset(Dataset):
...
             train_on_input: True
-          - _component_: torchtune.datasets.instruct_dataset
-            source: samsum
-            column_map: {"output": "summary"}
-            split: train
+            packed: True
+          - _component_: torchtune.datasets.grammar_dataset
+            split: train[:1%]
             train_on_input: False
+            packed: True
```
pbontrager, line 61: Should this be packed twice? Also I didn't realize concat supported packed.

felipemello1: It will raise an error

https://github.com/pytorch/torchtune/blob/main/torchtune/datasets/_concat.py#L71-L73

ebsmothers: In the example both datasets are packed. And yes, I added the support for packing ConcatDatasets, provided that we are packing all individual datasets. That's exactly why I made this update

felipemello1: sorry, can you point to the PR where you made these changes?

ebsmothers: #2278

felipemello1 (approved):

PR #2501, opened 2025-03-14
Raise error with tensor parallel + fused optimizer

ebsmothers:
Our current tensor parallel implementation is incompatible with fused Adam. I think this is because we are not currently applying sequence parallel to our norm layers. However, enabling that requires a bit more work. Until then, let's just raise an error to make the behavior explicit.

### DIFF
--- a/recipes/full_finetune_distributed.py
+++ b/recipes/full_finetune_distributed.py
@@ -160,7 +160,7 @@ class FullFinetuneRecipeDistributed(FTRecipeInterface):
             )
         if self.tensor_parallel_dim > 1 and cfg.optimizer.get("fused", False):
             raise ValueError(
-                "2D parallelism is currently incompatible with fused optimizer."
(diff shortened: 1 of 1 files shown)


felipemello1 (approved):
i imagine that this is also relevant for KD and DPO distributed too. Does it matter for LoRA?

we need to have some "config_parser" fn, so this type of change is applied to every recipe

pbontrager:
> i imagine that this is also relevant for KD and DPO distributed too. Does it matter for LoRA?
> 
> we need to have some "config_parser" fn, so this type of change is applied to every recipe

This is a bandaid really for now. The refactor work is what would allow us to get these safety checks out of the recipe.

pbontrager:
Good point, an recipe where we added TP

ebsmothers:
This is the only recipe with TP (besides generation where optimizer is irrelevant anyways)

PR #2054, opened 2024-11-22
Switch to PyTorch's built-in RMSNorm

calvinpelletier:
### Context
What is the purpose of this PR? Is it to
- [ ] add a new feature
- [ ] fix a bug
- [ ] update tests and/or documentation
- [ ] other (please add here)
- [x] zoom zoom 

### Results
**Inference**
Uncompiled: 31x faster
Compiled:  compilation and first batch = 5x faster, post-first batches = 1.5x faster

**Train (forward + backward)**:
Uncompiled: 22x faster
Compiled:  compilation and first batch = 1.5x faster, post-first batches = 1.2x faster

**Parity**:
Uncompiled: perfect
Compiled: 2.76e-6 MSE

### Test plan

Test code (output is below):
1 = original RMSNorm, 2 = new RMSNorm
```python
 from time import time

import torch
import torch.nn.functional as F

from torchtune.modules.rms_norm import RMSNorm, RMSNorm2


class Time:
    def __init__(s, name):
        s.name = name

    def __enter__(s):
        s.start = time()

    def __exit__(s, *a, **kw):
        print(f"TIME_{s.name}", time() - s.start)


shape = [1024, 512, 4096]
xs = [torch.randn(shape, dtype=torch.bfloat16, device="cuda") for _ in range(10)]

n1 = RMSNorm(shape[-1], eps=1e-6).cuda()
n2 = RMSNorm2(shape[-1], eps=1e-6).cuda()

with torch.no_grad():
    n1 = RMSNorm(shape[-1], eps=1e-6).cuda().eval()
    n2 = RMSNorm2(shape[-1], eps=1e-6).cuda().eval()

    with Time("inference_uncompiled1"):
        for x in xs:
            y1 = n1(x)
    with Time("inference_uncompiled2"):
        for x in xs:
            y2 = n2(x)
    print("inference uncompiled mse", F.mse_loss(y1, y2).item())

    with Time("inference_initial_compile1"):
        n1 = torch.compile(n1)
        n1(xs[0])
    with Time("inference_initial_compile2"):
        n2 = torch.compile(n2)
        n2(xs[0])

    with Time("inference_compiled1"):
        for x in xs:
            y1 = n1(x)
    with Time("inference_compiled2"):
        for x in xs:
            y2 = n2(x)
    print("inference compiled mse", F.mse_loss(y1, y2).item())


n1 = RMSNorm(shape[-1], eps=1e-6).cuda().train()
[...]

PR #2503, opened 2025-03-14
Enable SequenceParallel in 2D training

ebsmothers:
Currently we are unable to train with TP and fused optimizer (see #2501) because our norm layers were not parallelized. This PR changes that. To do that, we need to 

(1) Revert #2054, which switched our RMSNorm to `torch.nn.functional.rms_norm`. That PR did demonstrate some performance improvements and we should probably get 
(2) Pad to a multiple of `tensor_parallel_dim` by adding a `pad_to_multiple_of` option to `padded_collate_sft` -- otherwise we get unequal sequence lengths when calculating our loss. This is slightly inefficient but hopefully tp_dim << seq_len in most cases. We should probably do this for other collate functions too, but this is the most widely used, so starting here for now. (Really we should just direct people to use packed dataset, in that case this is kinda implicit anyways.)

(2) is currently done super hackily, I will do some more testing and eventually clean it up.

### DIFF
--- a/torchtune/data/_collate.py
+++ b/torchtune/data/_collate.py
@@ -161,6 +161,7 @@ def padded_collate_sft(
     batch: List[Dict[str, List[int]]],
     padding_idx: int = 0,
     ignore_idx: int = CROSS_ENTROPY_IGNORE_IDX,
+    pad_to_multiple_of: Optional[int] = None,
 ) -> Dict[str, torch.Tensor]:
     """Pad a batch of sequences to the longest sequence length in the batch, and
     convert integer lists to tensors.
@@ -169,6 +170,8 @@ def padded_collate_sft(
         batch (List[Dict[str, List[int]]]): A list of dictionaries containing input, label pairs.
         padding_idx (int): Padding index for input ids. Defaults to 0.
         ignore_idx (int): Padding index for labels. Defaults to -100.
+        pad_to_multiple_of (Optional[int]): If not None, pad the sequence to a multiple of this number.
+            This is useful for proper sharding with e.g. SequenceParallel.
 
     Returns:
         Dict[str, torch.Tensor]: Collated input and label tensors.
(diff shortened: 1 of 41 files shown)


torchtune/models/llama3/_parallelism.py
```diff
@@ -6,21 +6,41 @@
...
 # Define the Tensor Parallel plan for Llama3 model, which will also be shared with 3.1, 3.2, and 3.3 models
 BASE_LLAMA_TP_PLAN = {
-    "tok_embeddings": RowwiseParallel(input_layouts=Replicate()),
-    "output": ColwiseParallel(output_layouts=Replicate()),
+    "tok_embeddings": RowwiseParallel(
+        input_layouts=Replicate(), output_layouts=Shard(1)
+    ),
+    "norm": SequenceParallel(),
+    "output": ColwiseParallel(
+        input_layouts=Shard(1), output_layouts=Replicate(), use_local_output=True
```
acisseJZhong, line 26: I believe `use_local_output` is by default True?

torchtune/data/_collate.py
```diff
@@ -169,6 +170,8 @@ def padded_collate_sft(
         batch (List[Dict[str, List[int]]]): A list of dictionaries containing input, label pairs.
         padding_idx (int): Padding index for input ids. Defaults to 0.
         ignore_idx (int): Padding index for labels. Defaults to -100.
+        pad_to_multiple_of (Optional[int]): If not None, pad the sequence to a multiple of this number.
+            This is useful for proper sharding with e.g. SequenceParallel.
```
acisseJZhong, line 175: nice catch! thanks for that!

joecummings:
Screenshot 2025-03-26 at 12 51 12 PM
Confirmation that this works as expected. Let's take a look:

1. **Loss is the same**: cool, but wait you're testing with two different GPU configurations - one has two GPUs and is all DP and the other has DP + TP on 4 GPUs? Yes, this is b/c the data split and batch sizes should remain the same in order to properly test that the loss is the same. If we did full DP on 4 GPUs, then the data would be sharded 4 times, not twice and the order of data shown to the model would be different. 
2. **Tokens per second is lower w/ 'hybrid'**: Yep, this makes sense as we are still on single node and there will be more small communication synchronizations w/ TP than pure FSDP. 
3. **Memory usage is lower w/ 'hybrid'**: Duh, each GPU holds an even smaller fraction of parameters at all times

pbontrager (approved):
Thank you! I still think it would be cleaner to set the default for pad_to_multiple_of=1 instead of None since a multiple of 1 means no change and doesn't require a special None type. I'll approve anyway.

PR #2507, opened 2025-03-17
Linear Cross Entropy

Andrei-Aksionov:
Hi there 👋 

This PR adds support for Linear Cross Entropy loss (a.k.a. Cut Cross Entropy) proposed by Apple in [CUT YOUR LOSSES
IN LARGE-VOCABULARY LANGUAGE MODELS](https://arxiv.org/pdf/2411.09009) paper by adding support of the official [implementation](https://github.com/apple/ml-cross-entropy/tree/main).

So, from now on, we have:
- `CCE`: chunked cross entropy
- `LCE`: linear cross entropy

The benefit of LCE is that it reduces VRAM usage, especially for models with large vocab sizes.

This is done with 3 things:
1) **Custom Kernels**.
LCE uses custom CUDA kernels to perform matrix multiplications and the log-sum-exp reduction in shared memory. 
Usually you need to first calculate logits by multiplying embeddings (B, T, hidden_dim) with the output layer (hidden_dim, vocab_size) and then calculate loss with these logits and target labels.
LCE doesn't materialize this matrix with logits in the global memory (this matrix might be huge with modern LLMs that have large vocab size), but rather multiplies only a part (at a time) of output layer weights with logits in the shared (fast) memory and immediately calculates loss there (flash attention playbook).
This avoids materializing logits matrix (saves memory) and saves memory bandwidth (no need to move data from the global to shared memory and back multiple times).

2) **Gradient filtering**.
LCE leverages the sparsity of the softmax gradient to skip elements of the gradient computation that have a negligible contribution. This improves the throughput of LCE by reducing unnecessary computation.

3) **Vocabulary sorting**.
LCE uses vocabulary sorting to group tokens with similar average logits together. This increases the block-level sparsity and improves the effectiveness of gradient filtering.

-------

I ran a quick single device LoRA recipe with `gemma 2 2b` model.
This is the best possible case for LCE, since this model is relatively small, but has the same size of the vocab as larger models of gemma 2 family.

I ran 4 experiments:
- `CCE (non-compiled)`: chunked cross-entropy without compilation of model and loss
- `CCE`: chunked cross-entropy with compilation
- `LCE`: linear cross entropy with compilation
- `LCE (torch_compile impl)`: an implementation for older GPUs and MPS devices

> [!NOTE]
> To do this, one needs to change loss section of a config
> ```yaml
> loss:
>     _component_: cut_cross_entropy.LinearCrossEntropy
>     impl: torch_compile # (older than Ampere GPUs or MPS devices)
> ```

| Name                     | Peak Memory Reserved (GB) | Time  |
|--------------------------|---------------------------|-------|
| CCE (non-compiled)       | 21.6                      | 10:15 |
| CCE                      | 12.3                      | 09:18 |
| LCE                      | 7.73                      | 04:51 |
| LCE (torch_compile impl) | 9.1                       | 04:43 |

Screenshot 2025-03-17 at 6 55 01 PM

As one can see, the loss chart is identical in all cases, yet LCE significantly reduced reserved memory size and reduced time for training.

-----------

#### Context
- [x] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

#### Test plan

#### UX

### DIFF
--- a/recipes/lora_finetune_single_device.py
+++ b/recipes/lora_finetune_single_device.py
@@ -297,9 +297,38 @@ class LoRAFinetuneRecipeSingleDevice(FTRecipeInterface):
         if self._compile:
             self._loss_fn = training.compile_loss(self._loss_fn)
 
-        if self._loss_fn.__class__.__name__ == "CEWithChunkedOutputLoss":
+        # save the name of the loss class to reuse later
+        self.loss_class_name = (
+            self._loss_fn._orig_mod.__class__.__name__
+            if hasattr(self._loss_fn, "_orig_mod")
+            else self._loss_fn.__class__.__name__
+        )
+
+        if self.loss_class_name == "CEWithChunkedOutputLoss":
             # set num_output_chunks for model
             self._model.set_num_output_chunks(self._loss_fn.num_output_chunks)
+        elif self.loss_class_name == "LinearCrossEntropy":
+            # LCE doesn't call forward method of the class, but rather multiples embeddings with output weights.
+            # This is why PEFT methods (that use additional weights with custom logic) are not supported in LCE.
+            if isinstance(
+                self._model.output, (modules.peft.LoRALinear, modules.peft.DoRALinear)
+            ):
+                raise ValueError(
+                    "PEFT (LoRA/DoRA) applied to output is not supported with LinearCrossEntropy."
+                )
+
+            # LCE does multiplication of embeddings with output weights by itself in an efficient way.
+            # Thus we need to disable the forward method of the output layer.
+            if isinstance(self._model.output, nn.Linear):
+                self._model.output.forward = lambda x: x
+                self._output_weights = self._model.output.weight
+            elif isinstance(self._model.output, modules.TiedLinear):
+                self._model.output.linear.forward = lambda x, _: x
+                self._output_weights = self._model.output.tied_module.weight
+            else:
+                raise ValueError(
+                    f"`{type(self._model.output)}` for output layer is not supported with LinearCrossEntropy."
+                )
 
         log.info("Loss is initialized.")
 
@@ -653,7 +682,14 @@ class LoRAFinetuneRecipeSingleDevice(FTRecipeInterface):
             labels = labels.reshape(-1)
             logits = logits.reshape(-1, logits.size(-1))
 
-        loss = self._loss_fn(logits, labels)
+        if self.loss_class_name == "LinearCrossEntropy":
+            # LCE doesn't materialize logits in the global memory, but instead multiples embeddings with output weights
+            # in the shared memory and calculates loss. This is done to save memory.
+            # This is why we need to provide embeddings (logits), output weights and labels.
+            logits = logits.type_as(self._output_weights)
+            loss = self._loss_fn(logits, self._output_weights, labels)
+        else:
+            loss = self._loss_fn(logits, labels)
 
         # free logits otherwise it peaks backward memory
         del logits
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -66,7 +66,7 @@ you can also install the package locally with the following command.
     pip install -e .
 
     # or for a developer installation
-    pip install -e .["dev"]
+    pip install -e ".[dev]"
 
 |
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -61,6 +61,7 @@ dev = [
     "urllib3<2.0.0",
     "wandb",
     "expecttest",
+    "cut-cross-entropy @ git+https://github.com/apple/ml-cross-entropy.git",
 ]
 
 [tool.setuptools.dynamic]


felipemello1:
hey @Andrei-Aksionov , thanks for the PR! impressive numbers. Do you have any intuition for why tokens per second would double? From their paper, torch.compile is actually the fastest. I wouldnt imagine that changing the the output_layer + CE takes much longer than computing N transformer layers.

https://arxiv.org/pdf/2411.09009
image

If you have bandwidth, could you try with llama 8b for a few steps CCE+compile vs cutloss compile?

Andrei-Aksionov:
Hello @felipemello1 

> Do you have any intuition for why tokens per second would double?

Well, I guess that's because it's Gemma 2 with only 2b parameters, but the vocab size is the same as for larger models from Gemma 2 family, so applying classification head and calculating loss has a significant effect on training speed. Plus it's a LoRA variant, so it makes it even more pronounced.
I would imagine for other models of a bigger size and smaller vocab size the effect might be smaller.

As I understand, in case of a regular cross-entropy (chunked or not), the process is:
1. Calculate logits by multiplying embeddings with the output layer. If the vocab size is large, there are a lot of copying operations from shared memory (SM) to global memory (GB or HBM).
2. We need to convert those logits to probabilities, so we need:
    a) load logits block-by-block from HBM into SM to exponentiate values and then return them back.
    b) calculate sum of logits
    c) load exponentiated logits back from HBM into SM to divide them by the sum value, so we can convert them to probabilities
3. Load them again to calculate the loss

As you can see, a lot of copying operations compared to number of floating point operations.

LCE, on the other hand:
1. Copies a portion of embeddings and output weights (that can fit into a thread block), HBM --> SM
2. Multiplies them to get logits, but keeps the result in SM
3. Converts logits into probabilities without a need of knowing the sum of all exponentiated values, thanks to `online softmax normalization` trick (the one used in flash attention).
4. Probably calculates the loss there, but I'm not sure.
5. Copies the result back SM --> HBM.

As you can see, the proportion of `(number of floating point ops)` / `(bits to copy)`  is higher than for the regular CE calculation.
That's exactly what you want when you are writing a custom kernel to make everything faster (considering that the default kernel is memory bound).

------

The most interesting part to me, is that the `torch_compile impl` version is not that far off, actually.
It's as fast, but consumes a bit more memory (albeit smaller than CCE).
And this is just your regular [pytorch code](https://github.com/apple/ml-cross-entropy/blob/main/cut_cross_entropy/torch_compile.py), where in one function embeddings @ output layer + CE calculation, all wrapped in `torch.compile` decorator. No custom triton kernels.
I would imagine, that `torch.compile` can fuse these operations together and for softmax operations it uses `online softmax normalization` trick, which is already implemented in SDPA anyway. So, there is high chance for it.

That said, it could be a better option, since no additional dependencies required and no custom triton kernels.


-----

> I dont know if in their paper, when they compared vs Torchtune, they used compile=True.

I initially thought it might be unfair to compare a regular, non-compiled model with LCE. To address this, I included a compiled version in the comparison as well. 
For consistency, I also decided to compile the LCE variant.  I'm planning to add a non-compiled LCE version for completeness, although I expect the loss function to be compiled automatically anyway (only the model itself will stay non-compiled).

> If you have bandwidth, could you try with llama 8b for a few steps CCE+compile vs cutloss compile?

I'll try to.

Andrei-Aksionov:
Hey @felipemello1 

I've rerun test runs without compilation for both implementations of LCE (for completeness).
(Forgot to mention that it's all done on a single L4.)

| Loss Name                | Peak Memory Reserved (GB) | Time  | Compiled |
|--------------------------|---------------------------|-------|----------|
| CCE                      | 21.6                      | 10:15 | False    |
| CCE                      | 12.3                      | 09:18 | True     |
|-------------------------|-------------------------|-------------------------|-------------------------|
| LCE                      | 8.2                       | 05:53 | False    |
| LCE                      | 7.73                      | 04:51 | True     |
|-------------------------|-------------------------|-------------------------|-------------------------|
| LCE (torch_compile impl) | 8.7                       | 05:40 | False    |
| LCE (torch_compile impl) | 9.1                       | 04:43 | True     |

As you can see, even without a compilation, it's still ~2x faster than non-compiled CCE.

-----------------------

I wasn't able to run Llama 3.1 8b, since I don't have access to the repo. Tried to request access a couple of times and now I'm blocked foreeeeeeveeer.
So I tried Qwen 2.5 7b.  It has a larger vocab size than llama (152k vs 128k) and 1b less parameters, but at least it's something somewhat similar (kinda 🙃).

| Loss Name                | Peak Memory Reserved (GB) | Time  |
|--------------------------|---------------------------|-------|
| CCE                      | 17.5                      | 05:14 |
| LCE                      | 16.2                      | 04:40 |
| LCE (torch_compile impl) | 16.9                      | 04:26 |

Screenshot 2025-03-19 at 8 33 24 PM

Here the difference is less pronounced, since now the vocab size doesn't dwarf everything else, so doesn't have such a significant impact.

Again, what's interesting, is the `torch_compile impl`:
1. As fast as LCE implemented with custom Triton kernels
2. Provides some memory savings, but smaller than the proper Triton variant. (Due to not supporting Gradient filtering.)

-------

With larger and larger models the memory savings will become negligible, but the improvements in the speed of training is a nice thing.
I recommend spending time on researching the impact of a function, that, apparently, makes it easier to fuse last layer matmul and loss calculations. Again, look at the [code](https://github.com/apple/ml-cross-entropy/blob/24fbe4b5dab9a6c250a014573613c1890190536c/cut_cross_entropy/torch_compile.py#L14-L35) for `torch_compile impl`: it's just a function with two operations in it. The benefit is that you can replace `F.cross_entropy` with all the custom loss functions from this repo.


What do you think, @felipemello1?
And @ebsmothers, of course 😊

felipemello1:
hey @Andrei-Aksionov , this is amazing. Thanks for all the insightful experiments and comments. 

I think that, as a rule of thumb, we would be way more comfortable to add a torch-only implementation to torchtune than take dependency on a new repo. I don't know if their license allows copying it, but horace implemented an early version of pytorch-only CCE and shared it on twitter. Maybe we could use that?

We are a bit swamped now with some other tasks, but I dont want to keep you waiting. Can i get back you next week after i meet with the team and check if/how we want to add this loss?

The way you did works, but a) it adds some extra logic to the recipe (which is already a bit bloated), b) it modifies the model.
What I thought about doing originally was to add to the transformer a flag "skip_output_layer", but this is also bad, as it adds yet another flag to the model.

So let me check what others think and we can get back to it. Does that sound like a plan?

felipemello1:
Also, food for thought: Many other losses in torchtune ended up implementing the chunked version (e.g. knowledge distilation and GRPO). Although the fused version is more efficient, i wonder how easy it would be for someone to try to repurpose it. Maybe we would need to keep the current chunked version around.

recipes/lora_finetune_single_device.py
```diff
@@ -297,9 +297,38 @@ def setup(self, cfg: DictConfig) -> None:
         if self._compile:
             self._loss_fn = training.compile_loss(self._loss_fn)
 
-        if self._loss_fn.__class__.__name__ == "CEWithChunkedOutputLoss":
+        # save the name of the loss class to reuse later
+        self.loss_class_name = (
+            self._loss_fn._orig_mod.__class__.__name__
+            if hasattr(self._loss_fn, "_orig_mod")
+            else self._loss_fn.__class__.__name__
+        )
```
nathan-az, line 305: Noob question: what's the issue with always using `self._loss_fn.__class__.__name__`? 

<details>

```py
from torchtune import config

# Create an empty dictionary
short_dict = {
    "loss1":
        {
            "_component_": "torchtune.modules.loss.CEWithChunkedOutputLoss",
        },
    "loss2":
        {
            "_component_": "cut_cross_entropy.LinearCrossEntropy",
        }
}

# Convert the empty dictionary to an OmegaConf DictConfig object

cfg = OmegaConf.create(short_dict)
loss1 = config.instantiate(cfg.loss1)
loss2 = config.instantiate(cfg.loss2)

print(loss1.__class__.__name__)
print(loss2.__class__.__name__)
```

</details>

The above toy example prints
```
CEWithChunkedOutputLoss
LinearCrossEntropy
```

Andrei-Aksionov: Hello, @nathan-az 
Unfortunately, I didn't fully understand the question 🙃.
Are you asking why do we need to know the name of the loss function in the first place?
If so, then we need to know it since different loss functions require specific changes in the model.
If not, then a clarification would be nice 😊.

Andrei-Aksionov: `orig_mod` is needed because if I compile the loss with `torch.compile`, then the class will be wrapped into `OptimizedModule` (for LinearCrossEntropy). So thus I need to go deeper to retrieve the proper name.

There are many ways to make this PR nicer.
The whole logic could be wrapped in a new loss class, and in the training recipe only something like will be added
```python
if loss_class_name == "LinearCrossEntropy":
    self.loss_fn.prepare_model(model)
```
and the rest will be the same. Including `loss_step` function.

Or, if the core team decides to keep a fused version of the loss (like in `torch_compile impl`) variant, then we could have `FusedLoss` class that will contain any loss function and in the forward call will do the trick, the yaml file will have `fused` argument:
```yaml
loss:
    __component__: ...
    fused: True
```

I just need to know what is the decision.

nathan-az: Thanks for clarifying @felipemello1 and @Andrei-Aksionov - that does clarify and answer my question.

Agreed that a nicer abstraction is desired here. The manual handling of the output layer, logits and losses is an unfortunate side-effect too. I wonder if there's a nicer pattern for handling that.

nathan-az:
FYI: I pulled your branch to test whether the gains were significant with larger models, and made the same adjustments as this PR to the `full_finetune_distributed` recipe since I only have access to multi-GPU nodes currently. I consistently got `Illegal Memory Access` with Triton. I don't know how to look into this further, but details are below.

Docker: `ghcr.io/pytorch/pytorch-nightly:2.7.0.dev20250221-cuda12.4-cudnn9-devel`

Hardware: 8x Nvidia H100 80GB

**Error (rank0)**

<details>

```
[rank0]:   File "/app/cce_test.py", line 984, in recipe_main
[rank0]:     recipe.train()
[rank0]:   File "/app/cce_test.py", line 835, in train
[rank0]:     current_loss = self._loss_fn(logits, self._output_weights, labels)
[rank0]:                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank0]:   File "/opt/conda/lib/python3.11/site-packages/torch/nn/modules/module.py", line 1751, in _wrapped_call_impl
[rank0]:     return self._call_impl(*args, **kwargs)
[rank0]:            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank0]:   File "/opt/conda/lib/python3.11/site-packages/torch/nn/modules/module.py", line 1762, in _call_impl
[rank0]:     return forward_call(*args, **kwargs)
[rank0]:            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank0]:   File "/opt/conda/lib/python3.11/site-packages/cut_cross_entropy/linear_cross_entropy.py", line 126, in forward
[rank0]:     return linear_cross_entropy(
[rank0]:            ^^^^^^^^^^^^^^^^^^^^^
[rank0]:   File "/opt/conda/lib/python3.11/site-packages/cut_cross_entropy/linear_cross_entropy.py", line 71, in linear_cross_entropy
[rank0]:     return cce_linear_cross_entropy(
[rank0]:            ^^^^^^^^^^^^^^^^^^^^^^^^^
[rank0]:   File "/opt/conda/lib/python3.11/site-packages/cut_cross_entropy/cce.py", line 196, in cce_linear_cross_entropy
[rank0]:     return linear_cross_entropy_apply(
[rank0]:            ^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank0]:   File "/opt/conda/lib/python3.11/site-packages/cut_cross_entropy/cce.py", line 146, in linear_cross_entropy_apply
[rank0]:     loss = LinearCrossEntropyFunction.apply(e, c, bias, params)
[rank0]:            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank0]:   File "/opt/conda/lib/python3.11/site-packages/torch/autograd/function.py", line 575, in apply
[rank0]:     return super().apply(*args, **kwargs)  # type: ignore[misc]
[rank0]:            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank0]:   File "/opt/conda/lib/python3.11/site-packages/cut_cross_entropy/cce.py", line 67, in forward
[rank0]:     neg_dot = indexed_neg_dot_forward_kernel(
[rank0]:               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank0]:   File "/opt/conda/lib/python3.11/site-packages/cut_cross_entropy/indexed_dot.py", line 128, in indexed_neg_dot_forward_kernel
[rank0]:     _indexed_neg_dot_forward_kernel[grid](
[rank0]:   File "/opt/conda/lib/python3.11/site-packages/triton/runtime/jit.py", line 330, in <lambda>
[rank0]:     return lambda *args, **kwargs: self.run(grid=grid, warmup=False, *args, **kwargs)
[rank0]:                                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank0]:   File "/opt/conda/lib/python3.11/site-packages/triton/runtime/autotuner.py", line 385, in run
[rank0]:     return self.fn.run(*args, **kwargs)
[rank0]:            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank0]:   File "/opt/conda/lib/python3.11/site-packages/triton/runtime/autotuner.py", line 385, in run
[rank0]:     return self.fn.run(*args, **kwargs)
[rank0]:            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank0]:   File "/opt/conda/lib/python3.11/site-packages/triton/runtime/jit.py", line 653, in run
[rank0]:     kernel.run(grid_0, grid_1, grid_2, stream, kernel.function, kernel.packed_metadata, launch_metadata,
[rank0]:     ^^^^^^^^^^
[rank0]:   File "/opt/conda/lib/python3.11/site-packages/triton/compiler/compiler.py", line 395, in __getattribute__
[rank0]:     self._init_handles()
[rank0]:   File "/opt/conda/lib/python3.11/site-packages/triton/compiler/compiler.py", line 390, in _init_handles
[rank0]:     self.module, self.function, self.n_regs, self.n_spills = driver.active.utils.load_binary(
[rank0]:                                                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank0]: RuntimeError: Triton Error [CUDA]: an illegal memory access was encountered
```

</details>

**Config**

<details>

```yaml
batch_size: 1
checkpointer:
  checkpoint_files:
    max_filename: '00004'
  model_type: LLAMA3
  output_dir: ${output_dir}
  recipe_checkpoint: null
  checkpoint_dir: inputs/model
clip_grad_norm: null
compile: true
cudnn_deterministic_mode: false
custom_sharded_layers: []
dataset:
  _component_: torchtune.datasets.chat_dataset
  source: parquet
  conversation_column: messages
  conversation_style: openai
  split: train
  packed: true
  train_on_input: true
  data_dir: inputs/dataset
device: cuda
dtype: bf16
enable_activation_offloading: true
epochs: 1
fsdp_cpu_offload: false
log_every_n_steps: 1
log_peak_memory_stats: true
loss:
  _component_: cut_cross_entropy.LinearCrossEntropy
max_steps_per_epoch: 20
metric_logger:
  _component_: torchtune.training.metric_logging.MLFlowLogger
optimizer:
  _component_: torchao.prototype.low_bit_optim.AdamW8bit
  lr: 5.0e-05
optimizer_in_bwd: false
output_dir: outputs
resume_from_checkpoint: false
seed: 100
shuffle: true
tokenizer:
  max_seq_len: 2048
  path: inputs/model/original/tokenizer.model
model:
```

</details>

This may be a side-effect that I don't understand of being in a multi-device setting. I recall that `unsloth` has had issues related to memory in a DDP/FSDP multi-GPU setting which I believe are related to triton (although it puzzles me why this would be the case given the **data** parallel part).

Andrei-Aksionov:
Hey @nathan-az 
Thanks for checking it.
In the PR I only proposed changes to a single device regime, since I have access to a single device only.
So it's great to know what are implications with multi-device setting.

From the error log I can see that the issue is in the `indexed_neg_dot_forward_kernel` function, which is a part of a custom Triton kernel.

Could you change the config to the following
```yaml
loss:
    _component_: cut_cross_entropy.LinearCrossEntropy
    impl: torch_compile
```

and try it one more time?

With this implementation, a special function of only PyTorch code wrapped in `torch.compile` will be used. So it has higher chances of working properly.

nathan-az:
@Andrei-Aksionov I was able to make a pretty minimal repro so created https://github.com/apple/ml-cross-entropy/issues/31. I also tried with compiling the loss but there was no change. Maybe it'll be obvious to the maintainers what's going wrong! 

The above even fails with `--nproc_per_node=1` (so either my repro is very wrong, or something funky happens with FSDP). I wonder if this is even a PyTorch Core issue, or some funky interaction between FSDP and triton.

felipemello1:
> EDIT: In the repro, this is "fixed" by calling full_tensor on the output module, i.e. output.weight.full_tensor()

@nathan-az , could you try setting reshard_after_forward=False? If you could create a simple repro with just the loss and some dummy tensor, that would make it easier to ping someone from torch distributed. Thanks for investigating!

Andrei-Aksionov:
Thanks @nathan-az for investigating this 🎉 
The only thing to note is that
```yaml
loss:
    _component_: cut_cross_entropy.LinearCrossEntropy
    impl: torch_compile
```
is basically 
```python
loss = LinearCrossEntropy(imlp="torch_compile")
```
and not 
```python
loss = torch.compile(LinearCrossEntropy)
```

The difference is that the default implementation uses custom Triton kernels, while `torch_compile` implementation just fuses logits calculation and loss calculation in a [single function](https://github.com/apple/ml-cross-entropy/blob/main/cut_cross_entropy/torch_compile.py#L14-L35) wrapped in `torch.compile` decorator. 

I believe we are more interested in this implementation to work, since it is more aligned with the goal of the repo to provide PyTorch native code. In addition, it will make it easier to swap a loss function if needed.

felipemello1:
@Andrei-Aksionov , thanks again for all of the work on this! I am trying to think about the requirements to land it, but i think it will need some refactoring on torchtune side. 

requirements:
0. We should NOT need to add a new dependency (lets use pure pytorch maybe?), without violating licenses.

1. Recipes cannot be bloated with if/else. The complexity has to be mitigated by maybe checking if the loss has a method, e.g. "is_chunked" or "is_fused", or needs to be handled by some utility.

2. The model definition cannot be changed, e.g. model.output = new_output. This makes it harder to debug, because you read the code and its one thing, but its executing another.

3. It should work with distributed

Let me refactor some things today and it should make it easier for you to land your PR. Do you mind working on item 0, i.e. making sure that we don't add new dependencies?

nathan-az:
@Andrei-Aksionov using the `torch_compile` impl didn't correct the issue unfortunately, although it did produce a more useful error, explicitly stating that the issue was from mixing DTensor and Tensor.

@felipemello1 neither did disabling reshard, although this was a good suggestion that I expected to work. However, I believe this would cause the entire model to accumulate in memory during the forward pass (correct me if I'm wrong on how that arg works), which seems counter to the benefit of the merged kernel.

On the license note - the Liger Kernel repo also has a linear cross entropy implementation. Their repo is BSD-2 so very permissive. I'm not sure if there are any performance difference. From what I've seen on the Issue I posted, the Apple CCE maintainers are very responsive and super helpful, but the Liger implementation could be a backup.

Best of luck getting this going! Apologies for raising the hassle in the distributed case 😅

felipemello1:
hey @Andrei-Aksionov @nathan-az , i proposed a refactor here: https://github.com/pytorch/torchtune/pull/2531

@Andrei-Aksionov , this should help with trying liger or cut CE. I wonder if with this setup the user can directly instantiate the loss and use it, or if we need some sort of LossAdapter to avoid changing the recipe and all_gather the weights / redirect the inputs (eg weight,outputs.label) --> (output, weight, label)

In the config it could be something like:

```
loss:
    _component: torchtune.LossAdapter
    module: Liger.my_function
    new_order: [output, weight, label]
```

But I am afraid that this may be an overkill and would rather not go down that path. Would prefer to have a torchtune version and/or work with the compile team to match the CE perf by doing torch.compile(loss)

btw: someone used liger with tune and wrote a blog https://pytorch.org/blog/peak-performance-minimized-memory/

Andrei-Aksionov:
Hello @felipemello1 

Since in #2531 a [LinearCrossEntropyLoss](https://github.com/pytorch/torchtune/blob/main/torchtune/modules/loss/cross_entropy_loss.py#L14) was introduced, which basically repeats logic of `LinearCrossEntropy(imlp="torch_compile")` (and we've discussed that it's better to stick to it instead of bringing cut cross entropy as an additional dependency) I guess I can close this draft PR.

pbontrager:
Thanks for all the great work you did here which helped us get the final version across the line. To make distributed work with cross entropy, we've had to further refactor our losses. [This](https://github.com/pytorch/torchtune/pull/2667) should land today. Once this lands, we'll need an adapter, like @felipemello1 mentioned to work with Liger. It should probably be something like `LigerCrossEntropyLoss` and will have to handle processing DTensor's before handing off to the liger module and in `setup_model_output` it will need to ensure the model output head isn't sharded or quantized. I'll put up an issue for this once the other PR lands.

Andrei-Aksionov:
> Thanks for all the great work you did here which helped us get the final version across the line.

Glad that it was helpful 😊.

> [This](https://github.com/pytorch/torchtune/pull/2667) should land today.

Cool.

PR #2510, opened 2025-03-18
REMOVE `recursive_reshard` UTILITY

### LINKED ISSUE #2483: recursive_reshard
https://github.com/pytorch/torchtune/blob/main/torchtune/training/_distributed.py#L571C1-L572C1

The logic seems strange, shoud this be?:
```
    for n, m in reversed(list(module.named_modules())):
        m.reshard()
```

Image

ebsmothers:
At first, this  PR was supposed to fix #2483. Upon further inspection of this utility, it became clear that it wasn't needed. And if it's not needed, why keep it around?

### How do you know it's not needed?

Well, the intuition behind doing this is that on the model we don't utilize the `reshard_after_forward` param in order to speed up generation but then manually reshard to save memory.

Unfortunately, while this does *briefly* save memory, it's not meaningful due to the high memory peaks immediately afterwards. Here's the memory trace w/ the resharding utility.
Screenshot 2025-04-01 at 3 21 04 PM

And here's the memory trace without the resharding.
Screenshot 2025-04-01 at 3 21 13 PM

**Key point: This does not meaningfully save memory.**

### Okay, but does it hurt to keep around?

Well, let's think about it. Resharding ostensibly takes time. 

Here's a view of the call trace w/ resharding. The function housing the call is `generate_trajectory` and here it takes a **whopping** 2s649ms. The extra empty cache call takes 207 ms and doing a forward pass on a newly resharded model takes 396 ms. 
Screenshot 2025-04-01 at 4 18 04 PM

Compare this to no resharding. The total `generate_trajectory` call only takes 1s887ms!!! There's no extra empty cache call and the model forward takes 170ms. Plus, the other empty cache call only takes 28ms. Need to investigate this further, but my guess is that there are some extra ops hanging around after our manual resharding that are cleaned up. Without manual resharding, there's nothing really to clean up. Just a theory here. 
Screenshot 2025-04-01 at 4 18 20 PM

Not convinced? Okay, let's say in GRPO we are really dumb and have a batch size of 32 and a forward batch size of 1. Thus, we call `generate_trajectory` 32 times in one overall GRPO step. 32 x 0.762ms = 24s384ms EXTRA TIME PER STEP. Now let's assume a very basic 500 steps for GRPO. THAT'S ALMOST AN EXTRA 3 HOURS TO YOUR TRAINING RUN. HOLY HELL. WHAT ARE YOU DOING.

Okay so in conclusion, manually resharding the model:
1. Does not meaningfully save us memory
2. Hurts us a LOT in performance

Hence, we remove :)

We might need this utility again in the future, but if so we can always add it back. I'm not in favor of keeping a function around that isn't used ATM.


joecummings (commented):
@ebsmothers Can I just get a vibe check here? Can this be landed?

ebsmothers:
@joecummings yeah I believe it can be landed, just needs a final test

torchtune/training/_distributed.py
```diff
@@ -26,7 +26,7 @@
     StateDictOptions,
 )
 from torch.distributed.device_mesh import DeviceMesh, init_device_mesh
-from torch.distributed.fsdp import ShardingStrategy
+from torch.distributed.fsdp import FSDPModule, ShardingStrategy
```
ebsmothers, line 29: I think we don't need FSDPModule now, right?

joecummings: Eh, it's technically used as a signature in some of our utils - I'd rather just keep it in.

joecummings (approved):
Great work me

PR #2513, opened 2025-03-18
Remove old torchao imports, require 0.7.0+

andrewor14:
These imports have been deprecated for 3 torchao versions already. Since torchtune only supports the latest torchao version, we remove these imports.


andrewor14:
@ebsmothers

PR #2514, opened 2025-03-18
[WIP] System Prompt Mapping for Instruct DS

jaysonfrancis:
To support instruct datasets that may have system prompts defined per row, e.g; `(system, input, output)`, otherwise, fall back to a fixed `new_system_prompt`, or `None`.

Are there alternatives to this I may have overlooked? Otherwise, I can proceed with adding tests/docs. Thanks!

---

#### Context
- [x] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

#### Changelog
*

#### Test plan

#### UX

---

### DIFF
--- a/torchtune/data/_messages.py
+++ b/torchtune/data/_messages.py
@@ -209,6 +209,7 @@ class InputOutputToMessages(Transform):
             )
 
         self.image_dir = image_dir
+        self.has_system = "system" in self.column_map
 
     def __call__(self, sample: Mapping[str, Any]) -> Mapping[str, Any]:
         is_multimodal = "image" in sample or (
@@ -254,12 +255,21 @@ class InputOutputToMessages(Transform):
                 eot=True,
             ),
         ]
-        if self.new_system_prompt is not None:
+
+        system_prompt = None
+        if self.has_system and self.column_map["system"] in sample and sample[self.column_map["system"]]:
+            system_prompt = sample[self.column_map["system"]]
+
+        elif self.new_system_prompt is not None:
+            system_prompt = self.new_system_prompt 
+
+        if system_prompt is not None:
             messages = [
                 Message(
-                    role="system", content=self.new_system_prompt, masked=True, eot=True
+                    role="system", content=system_prompt, masked=True, eot=True
                 )
             ] + messages
+
         return {"messages": messages}
 
 


torchtune/data/_messages.py
```diff
@@ -209,6 +209,7 @@ def __init__(
             )
 
         self.image_dir = image_dir
+        self.has_system = "system" in self.column_map
```
RdoubleA, line 212: So if a user wants to use a dataset that has a system column, they will need to specify `{"system": "system"}` in column map? It feels a bit unintuitive, ideally they don't need to do this and if there is a system column and it's named system, they shouldn't need to do anything because it follows a default pattern. Still thinking about the best logic for this...

torchtune/data/_messages.py
```diff
@@ -254,12 +255,21 @@ def __call__(self, sample: Mapping[str, Any]) -> Mapping[str, Any]:
                 eot=True,
             ),
         ]
-        if self.new_system_prompt is not None:
+
+        system_prompt = None
+        if self.has_system and self.column_map["system"] in sample and sample[self.column_map["system"]]:
```
RdoubleA, line 260: I think we should do something to the effect of:
- If "system" is in column map, user has specified a system column that is not the default "system" -> use that to key into sample dict
- If "system" is in the sample dict, then the user is using the default system column name -> use that to key into sample dict directly. Column map can be none 
- If neither of these are true, there is no system column

This will keep the behavior I described above. Let me know if this makes sense or it is unnecessarily complicating things.

RdoubleA (commented):
Could you share some HF datasets that have a system column?

jaysonfrancis:
> Sorry I know this was a draft but had some early thoughts. Could you share some HF datasets that have a system column?

Thanks for feedback.

There aren't many, but here are a few examples; [here](https://huggingface.co/datasets/sujet-ai/Sujet-Finance-Instruct-177k), [here](https://huggingface.co/datasets/Josephgflowers/Finance-Instruct-500k), [here](https://huggingface.co/datasets/rombodawg/Everything_Instruct_Multilingual). It may also be useful for consolidated datasets.

Yes your comments make sense. Any thoughts on the case if "system" is in sample dict and _not_ in column map, but user has defined a fixed, "new_system_prompt" ?

PR #2518, opened 2025-03-19
Enable instantiation from globals

felipemello1:
#### Context
- [x] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

Before this PR, you couldn't instantiate 'my_local_function', .e.g. in a script or a jupyter notebook. We only supported functions that were importable. Now you can do this

```
def my_local_function():
    return 0
config = {"foo": "_component_": "my_local_function"}
foo = instantiate(config)
```

### DIFF
--- a/torchtune/config/_utils.py
+++ b/torchtune/config/_utils.py
@@ -61,34 +61,38 @@ def _get_component_from_path(
         >>> _get_component_from_path("torch")
         <module 'torch' from '...'>
         >>> # Assuming FooBar is in caller's globals
-        >>> _get_component_from_path("FooBar", globals())
+        >>> _get_component_from_path("FooBar")
         <class 'FooBar'>
     """
     if not path or not isinstance(path, str):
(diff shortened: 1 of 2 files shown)


torchtune/config/_utils.py
```diff
@@ -34,79 +34,74 @@ def _has_component(node: Union[Dict[str, Any], DictConfig]) -> bool:
...
+        >>> # Importing a class from a module
+        >>> _get_component_from_path("torch.nn.Linear")
+        <class 'torch.nn.modules.linear.Linear'>
+
+        >>> _get_component_from_path("torch")
+        <module 'torch' from '...'>
+
+        >>> # Assuming FooBar is defined in caller's globals
+        >>> _get_component_from_path("FooBar", globals())
+        <class 'SetupDataset'>
```
felipemello1, line 67: todo: fix this

joecummings (approved):
Kewl

PR #2519, opened 2025-03-19
nested instantiation

felipemello1:
#### Context
- [x] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

Enable nested instantiation, e.g.

```
from torchtune.config import instantiate


class Spice:
    def __init__(self, heat_level):
        self.heat_level = heat_level


class Food:
    def __init__(self, seed, ingredient):
        self.seed = seed
        self.ingredient = ingredient


config = OmegaConf.create(
    {
        "_component_": "Food",
        "seed": 0,
        "ingredient": {"_component_": "Spice", "heat_level": 5},
    }
)
food = instantiate(config, seed=42)
print(food.seed)  # 42
print(food.ingredient.heat_level)  # 5
new_spice = {"_component_": "Spice", "heat_level": 10}
food = instantiate(config, ingredient=new_spice)
print(food.ingredient.heat_level)  # 10
```

#### Test plan
unit tests


ebsmothers:
I understand the motivation here, but at this point why not just use Hydra?

felipemello1:
> I understand the motivation here, but at this point why not just use Hydra?

bigger PR

RdoubleA:
> bigger PR, but i think we should

I'm bout to flip

joecummings (approved):
Can we add a warning that you really *shouldn't* have multiple levels of instantiation though?

PR #2531, opened 2025-03-27
Refactor losses instantiation and chunked CE

felipemello1:
#### Context
- [ ] add a new feature
- [ ] fix a bug
- [x] other (please add here)

IMPORTANT: Recipes do NOT work with older version of ChunkedCrosEntropy anymore, because we dont expect transformer to chunk the outputs.

Problem:
1) We have seen many chunked losses being added to torchtune. The current setup put the chunking burden on the model.
2) Users have interest in using losses that require model.output.weight as input, e.g. liger losses

Solution:
1) Enable the recipe to call loss(weight, input, targets)
2) Reimplement ChunkedCE, so that chunking and projection happens in the loss.
3) Adds protocol so that new losses can follow the same pattern

PROFILING: https://drive.google.com/drive/folders/1jHOCuOF74F9lmmJv7wxbcK-i_wtB2stf?usp=sharing

#### Changelog
- Updated full_distributed and lora_distributed
- Tested with lora llama 3.2 distributed (TiedLinear)
- Implemented new ChunkedCE

TODO: when approved, will implement it to the other recipes/losses/update configs

#### Test

ChunkedCrossEntropyLoss
```
tune run --nproc_per_node 2 lora_finetune_distributed --config /data/users/felipemello/torchtune/recipes/configs/llama3_2/1B_lora.yaml \
metric_logger=torchtune.training.metric_logging.WandBLogger \
dataset.packed=True \
dataset.split=train[:50%] \
tokenizer.max_seq_len=4096 \
batch_size=4 \
max_steps_per_epoch=20 \
compile=True \
use_output_weight_in_loss=True \
loss=torchtune.modules.loss.sft_losses.ChunkedCrossEntropyLoss
```

image

### To reproduce

```
fork ----> https://github.com/pytorch/torchtune
git clone https://github.com/<YOUR_GITHUB_USER>/torchtune.git

cd torchtune
conda create -n torchtune python=3.11
conda activate torchtune
pip install --pre --upgrade torch torchvision torchao --index-url https://download.pytorch.org/whl/nightly/cu124
pip install -e ".[dev]"
pre-commit install

git remote add felipemello1 https://github.com/felipemello1/torchtune.git
git checkout -b loss_refactor felipemello1/loss_refactor
```


torchtune/modules/loss/ce_chunked_output_loss.py
```diff
@@ -8,8 +8,239 @@
...
+        target_chunks = targets.chunk(self.num_output_chunks, dim=1)
+
+        total_loss = 0.0
+        for hidden_chunk, target_chunk in zip(hidden_chunks, target_chunks):
+            total_loss += self.compute_cross_entropy(weight, hidden_chunk, target_chunk)
+
+        return total_loss / total_elements
+
+
+class ChunkedCEAutograd(torch.autograd.Function):
```
felipemello1, line 85: just another version, we can prob delete, but its more memory efficient

torchtune/modules/loss/ce_chunked_output_loss.py
```diff
@@ -8,8 +8,239 @@
...
+        """
+        # Create mask for non-ignored tokens: [batch_size, chunk_size]
+        mask = target_chunk != self.ignore_index
+
+        # Select hidden states and targets where mask is True
+        hidden_selected = hidden_chunk[mask]  # [num_valid, embed_dim]
+        target_selected = target_chunk[mask]  # [num_valid]
+
+        # Project only selected hidden states: [num_valid, embed_dim] @ [embed_dim, vocab_size]
+        logits_selected = F.linear(hidden_selected, weight)  # [num_valid, vocab_size]
```
felipemello1, line 49: mask before projection to avoid waste, call the projection here

torchtune/modules/transformer.py
```diff
@@ -667,11 +672,13 @@ def forward(
...
         else:
-            # shape: [b, seq_len, out_dim]
-            output = self.output(h).float()
+            if self.num_output_chunks > 0:
+                output = self.chunked_output(h)
+            else:
+                # shape: [b, seq_len, out_dim]
+                output = self.output(h).float()
 
         return output
```
felipemello1, line 684: we can remove chunking logic when after next release

torchtune/training/_compile.py
```diff
@@ -57,6 +57,7 @@ def compile_model(
             m.compile(backend=backend)
 
 
+# @deprecate
 def compile_loss(loss: nn.Module, verbose: bool = True) -> nn.Module:
```
felipemello1, line 61: we can keep the utility, but it can be much simpler now. No need to have a different compilation strategy per loss.

recipes/full_finetune_distributed.py
```diff
@@ -784,25 +784,48 @@ def train(self) -> None:
...
+                        )
+
+                if self.skip_unembedding:
+                    weight = get_model_proj_weight(self._model)
+                    current_loss = self._loss_fn(weight, outputs, labels)
+                else:
                     labels = labels.reshape(-1)
-                    logits = logits.reshape(-1, logits.size(-1))
+                    outputs = outputs.reshape(-1, outputs.size(-1))
+                    current_loss = self._loss_fn(outputs, labels)
```
felipemello1, line 821: if skip_unembedding, let the loss handle the output projection

recipes/full_finetune_distributed.py
```diff
@@ -344,9 +344,9 @@ def setup(self, cfg: DictConfig) -> None:
         if self._compile:
             training.compile_loss(self._loss_fn, verbose=self._is_rank_zero)
 
-        if self._loss_fn.__class__.__name__ == "CEWithChunkedOutputLoss":
-            # set num_output_chunks for model
-            self._model.set_num_output_chunks(self._loss_fn.num_output_chunks)
+        # skip final projection, since the loss takes hidden input instead of logits
+        self.skip_unembedding = cfg.get("loss_takes_embeddings", False)
+        self._model.set_skip_unembedding(self.skip_unembedding)
```
felipemello1, line 349: an alternative to "loss_takes_embeddings' is to have some loss protocol and let the losses have this arg, but it doesnt seem necessary

recipes/full_finetune_distributed.py
```diff
@@ -784,25 +784,48 @@ def train(self) -> None:
...
+                        return output.weight
+                    elif isinstance(output, modules.TiedLinear):
+                        return output.tied_module.weight
+                    else:
+                        raise ValueError(
+                            f"Unsupported output module type: {type(output)}"
+                        )
+
+                if self.skip_unembedding:
+                    weight = get_model_proj_weight(self._model)
```
felipemello1, line 816: might need to run weight.full_tensor for it to work with apples implementation

torchtune/modules/transformer.py
```diff
@@ -478,7 +489,14 @@ def reset_caches(self):
         for layer in self.layers:
             layer.reset_cache()
 
-    @torch.compiler.disable
+    def get_output_weight(self) -> torch.Tensor:
+        """Returns the output weight matrix."""
+        return self.output.weight
```
SalmanMohammadi, line 494: Would this work for models with tied embeddings? Wouldn't it need to be `self.output.tied_module.weight` for those cases?

felipemello1: great question! I added @property weight to tied_module

torchtune/modules/loss/sft_losses.py
```diff
@@ -0,0 +1,233 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
```
SalmanMohammadi, line 1: I liked the more specific filename for the chunked CE loss - `chunked_ce.py` mayhaps?

felipemello1: i am ok either way. I did it like this because we have kd_losses.py.

but i agree that having a more descriptive filename is better

torchtune/modules/transformer.py
SalmanMohammadi: neat. I totally missed that  ( :

torchtune/modules/model_fusion/_deep_fusion.py
```diff
@@ -92,6 +92,7 @@ def __init__(
         else:
             trainable_params -= set(get_fusion_params(self))
         set_trainable_params(self, trainable_params)
+        self.skip_output_layer = False
```
SalmanMohammadi, line 95: nit: worth making this a property which returns `self.decoder.skip_output_layer`?

torchtune/modules/model_fusion/_deep_fusion.py
```diff
@@ -92,6 +92,7 @@ def __init__(
...
 
     def set_num_output_chunks(self, num_output_chunks: int) -> None:
```
SalmanMohammadi, line 97: Deprecate here too?

SalmanMohammadi, line 97: I know that if you call this you'll see the deprecation message from `TransformerDecoder`, do you think it's overkill to indicate that `DeepFusionModel.set_num_output_chunks` will also be deprecated?

felipemello1: if we add it, then the user would get 2 warnings, one from deepfusion, and one from self.decoder. I was thinking that we could keep it only on self.decoder.

torchtune/modules/transformer.py
```diff
@@ -478,7 +489,14 @@ def reset_caches(self):
...
+
+    def set_skip_output_layer(self, skip: bool) -> None:
```
joecummings, line 496: I would opt for a slightly longer description here of when and why you would want to use this.

torchtune/modules/transformer.py
```diff
@@ -389,6 +393,7 @@ def __init__(
         self.head_dim = head_dim
         self.causal_mask = None
         self.num_output_chunks = 0
+        self.skip_output_layer = False
```
joecummings, line 396: I would prefer the variable `skip_output_projection` to make it even more clear that this is just skipping the normal output projection and nothing else.

joecummings: Would need to update other methods and varaibles as well.

torchtune/modules/model_fusion/_deep_fusion.py
joecummings: Nah, not overkill.

torchtune/modules/loss/sft_losses.py
joecummings: I'd just be in favor of chunked_cross_entropy.py

SalmanMohammadi:
@felipemello1 this is awesome. Out of curioisity did you happen to benchmark against the existing `CEWithChunkedOutputLoss`?

I wonder if we could simplify the configuration further by removing the need for the user to also specify `use_output_weight_in_loss`? Could we define a 
```python
class BaseLoss(Protocol):
    is_chunked: bool
```
and do

```diff
-                if self.use_output_weight_in_loss:
+                if self.loss_fn.is_chunked:
                    weight = self._model.get_output_weight()
                    current_loss = self._loss_fn(weight, outputs, labels)
                else:
                    labels = labels.reshape(-1)
                    logits = logits.reshape(-1, logits.size(-1))
                    outputs = outputs.reshape(-1, outputs.size(-1))
                    current_loss = self._loss_fn(outputs, labels)
```
It would require either 1) requiring that all losses use this protocol (which tbh I wouldn't be opposed to as we start to support more custom losses without needing to modify recipes), or doing a `hasattr` check on `self._loss_fn` and relying on an identifying field on *just* the chunked losses.

wdyt?

torchtune/modules/loss/sft_losses.py
```diff
@@ -0,0 +1,233 @@
...
+        target_chunks = targets.chunk(self.num_output_chunks, dim=1)
+
+        total_loss = 0.0
+        for hidden_chunk, target_chunk in zip(hidden_chunks, target_chunks):
+            total_loss += self.compute_cross_entropy(weight, hidden_chunk, target_chunk)
+
+        return total_loss / total_elements
+
+
+class ChunkedCrossEntropywithAutograd(torch.autograd.Function):
```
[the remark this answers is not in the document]
felipemello1: this version is based off horace's code from a few months back. In this implementation, the chunks are not held up in memory. He coded it to show that you dont need trition.

I dont want to keep it in torchtune, because it would be hard to use to for KD/RL losses. This is more a reference for the compile folks. They are working on enabling the chunking on compile to match the autograd memory perf.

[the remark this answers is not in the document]
felipemello1: Without autograd
image

torchtune/modules/transformer.py
```diff
@@ -602,9 +626,9 @@ def forward(
                 and skip straight to the transformer layers. Shape ``[b x s x d]``. Default: None
 
         Returns:
-            Union[torch.Tensor, List[torch.Tensor]]: output tensor with shape ``[b x s x v]`` or a list of layer
-                output tensors defined by ``output_hidden_states`` with the
-                final output tensor appended to the list.
+            Union[torch.Tensor, List[torch.Tensor]]: output tensor with shape ``[b x s x v]`` if `skip_output_layer=False`
+            and ``[b x s x d]`` if true, or a list of layer output tensors defined by ``output_hidden_states`` with the
+            final output tensor appended to the list.
```
SalmanMohammadi, line 631: nit
```suggestion
            Union[torch.Tensor, List[torch.Tensor]]: output tensor with shape ``[b x s x v]`` if `self.skip_output_layer=False`
            and ``[b x s x d]`` otherwise, or a list of layer output tensors defined by ``output_hidden_states`` with the
            final output tensor appended to the list.
```

torchtune/modules/loss/sft_losses.py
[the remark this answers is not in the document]
felipemello1: with autograd

image

felipemello1:
> I wonder if we could simplify the configuration further by removing the need for the user to also specify use_output_weight_in_loss?

@SalmanMohammadi , i thought about it and even implemented, but then realized that it would be hard to support 3rd party libraries, unless we create some sort of loss adapter, which we may need to do anyway, because not all libraries follow the patten (weight, input, label). They may follow (label, weight, input), for example.

torchtune/modules/transformer.py
```diff
@@ -667,11 +691,13 @@ def forward(
...
+        if self.skip_output_layer:
+            output = h
         else:
-            # shape: [b, seq_len, out_dim]
-            output = self.output(h).float()
+            if self.num_output_chunks > 0:
+                output = self.chunked_output(h)
+            else:
+                # shape: [b, seq_len, out_dim]
+                output = self.output(h).float()
```
SalmanMohammadi, line 701: epic nit:
```python
         elif self.num_output_chunks > 0:
            output = self.chunked_output(h)
         else:
            # shape: [b, seq_len, out_dim]
```

torchtune/training/_compile.py
```diff
@@ -73,6 +73,8 @@ def compile_loss(loss: nn.Module, verbose: bool = True) -> nn.Module:
     backend = os.environ.get("TORCH_COMPILE_BACKEND", "inductor")
     if verbose:
         log.info("Compiling loss with torch.compile...")
+
+    # TODO: Delete all of this loss specific code for next release
     if isinstance(loss, CEWithChunkedOutputLoss):
```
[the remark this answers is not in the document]
felipemello1: i dont think we needed it anymore. It works well now just compiling the loss.forward, instead of the loss. I dont know why, but i dont question the gods.

in short: we can get rid of all if/else

torchtune/modules/loss/sft_losses.py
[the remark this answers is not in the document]
joecummings: Can you put a comment in the code to that effect?

torchtune/training/_compile.py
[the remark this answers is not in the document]
SalmanMohammadi: compile giveth and compile taketh

felipemello1:
> I can't think of any 3rd party losses which we claim to support which would fall into this category - do you have any examples?

yes, liger and apple:
https://github.com/linkedin/Liger-Kernel/tree/main/src/liger_kernel/chunked_loss
https://github.com/apple/ml-cross-entropy/tree/main/cut_cross_entropy

torchtune/modules/loss/loss_protocols.py
```diff
@@ -0,0 +1,67 @@
...
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+from typing import Protocol
+
+import torch
+
+
+class SFTLossWithProjection(Protocol):
```
joecummings, line 12: Probably needs to be ``SFTLossWithOutputProj`` or something. Projection is too vague.

torchtune/modules/loss/cross_entropy_loss.py
```diff
@@ -0,0 +1,114 @@
...
+# LICENSE file in the root directory of this source tree.
+
+import torch
+import torch.nn.functional as F
+from torch import nn
+
+from .loss_protocols import SFTLossWithProjection
+
+
+class ChunkedCrossEntropyLoss(nn.Module, SFTLossWithProjection):
```
joecummings, line 14: Can you add this to the docs? Also maybe include a slightly longer description of why we might want to use this. And how we might use this in a generic training loop.

recipes/full_finetune_distributed.py
```diff
@@ -347,9 +347,11 @@ def setup(self, cfg: DictConfig) -> None:
         if self._compile:
             training.compile_loss(self._loss_fn, verbose=self._is_rank_zero)
 
-        if self._loss_fn.__class__.__name__ == "CEWithChunkedOutputLoss":
-            # set num_output_chunks for model
-            self._model.set_num_output_chunks(self._loss_fn.num_output_chunks)
+        # The loss may handle the output projection. If true, the model should skip it.
+        self.use_output_weight_in_loss = getattr(
```
joecummings, line 351: output weight & output proj? Should stay consistent, no?

joecummings (approved):
Approving to unblock but there's a few things related to consistency and documentation that should be cleaned up.

torchtune/modules/loss/loss_protocols.py
```diff
@@ -0,0 +1,67 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
```
SalmanMohammadi, line 1: Our protocols for tokenizers live in `tokenizers/_utils.py`. Do you think it's worth keeping things consistent and renaming this to `_utils.py`?

torchtune/modules/loss/loss_protocols.py
```diff
@@ -0,0 +1,67 @@
...
+class SFTLossWithProjection(Protocol):
+    """Protocol for loss functions in torchtune used in Supervised Finetune recipes and that require
+    model output projection weights in loss computation."""
+
+    use_output_proj_in_loss: bool = True
+
+    def apply_compile_strategy(self, *args, **kwargs):
+        """Torch compiles the loss function. Can be useful when greater control is needed,
+        for example when only compiling a portion of the loss calculation."""
+        self.forward = torch.compile(self.forward, *args, **kwargs)
```
SalmanMohammadi, line 21: Should we be doing this here? I'd vote to add this into the docstring as an example

torchtune/modules/loss/loss_protocols.py
felipemello1: i could move it, but it doesnt feel intuitive that protocols are in utils. Do you think its a better choice, or is it just to keep things consistent?

torchtune/modules/loss/loss_protocols.py
```diff
@@ -0,0 +1,67 @@
...
+        """
+        ...
+
+
+class SFTLoss(Protocol):
+    """Protocol for loss functions in torchtune used in sft recipes."""
+
+    use_output_proj_in_loss: bool = False
+
+    def apply_compile_strategy(self, *args, **kwargs):
```
SalmanMohammadi, line 50: similar comment to above

torchtune/modules/transformer.py
```diff
@@ -609,9 +641,9 @@ def forward(
                 and skip straight to the transformer layers. Shape ``[b x s x d]``. Default: None
 
         Returns:
-            Union[torch.Tensor, List[torch.Tensor]]: output tensor with shape ``[b x s x v]`` or a list of layer
-                output tensors defined by ``output_hidden_states`` with the
-                final output tensor appended to the list.
+            Union[torch.Tensor, List[torch.Tensor]]: output tensor with shape ``[b x s x v]`` if `self.skip_output_projection=False`
```
SalmanMohammadi, line 644:
```suggestion
            Union[torch.Tensor, List[torch.Tensor]]: output tensor with shape ``[b x s x v]`` if ``self.skip_output_projection=False``
```
my bad

torchtune/modules/loss/loss_protocols.py
```diff
@@ -0,0 +1,67 @@
...
+import torch
+
+
+class SFTLossWithProjection(Protocol):
+    """Protocol for loss functions in torchtune used in Supervised Finetune recipes and that require
+    model output projection weights in loss computation."""
+
+    use_output_proj_in_loss: bool = True
+
+    def apply_compile_strategy(self, *args, **kwargs):
```
SalmanMohammadi, line 18: Thoughts on naming this `def compile`? Is that too vague?

torchtune/modules/loss/loss_protocols.py
```diff
@@ -0,0 +1,67 @@
...
+        outputs: torch.Tensor,
+        targets: torch.Tensor,
+        *args,
+        **kwargs,
+    ) -> torch.Tensor:
+        """
+        Args:
+            weight (torch.Tensor): Tensor with weights of the model output projection layer. Shape [vocab_size, emb_dim]
+            outputs (torch.Tensor): Hidden state of the model, pre projection. Shape [bsz, seq_len, emb_dim]
+            targets (torch.Tensor): Labels for the model. Shape [bsz, seq_len]
```
SalmanMohammadi, line 36:
```suggestion
            weight (torch.Tensor): Tensor with weights of the model output projection layer. Shape ``[vocab_size, emb_dim]``
            outputs (torch.Tensor): Hidden state of the model, pre projection. Shape ``[bsz, seq_len, emb_dim]``
            targets (torch.Tensor): Labels for the model. Shape ``[bsz, seq_len]``
```

torchtune/modules/loss/loss_protocols.py
```diff
@@ -0,0 +1,67 @@
...
+        self.forward = torch.compile(self.forward, *args, **kwargs)
+        return self
+
+    def forward(
+        self, outputs: torch.Tensor, targets: torch.Tensor, *args, **kwargs
+    ) -> torch.Tensor:
+        """
+        Args:
+            outputs (torch.Tensor): Logits of the model. Shape [bsz, seq_len, vocab_size]
+            targets (torch.Tensor): Labels for the model. Shape [bsz, seq_len]
```
SalmanMohammadi, line 61:
```suggestion
            outputs (torch.Tensor): Logits of the model. Shape ``[bsz, seq_len, vocab_size]``
            targets (torch.Tensor): Labels for the model. Shape ``[bsz, seq_len]``
```

recipes/full_finetune_distributed.py
```diff
@@ -795,24 +797,29 @@ def train(self) -> None:
...
+
                 # Shift labels to compute loss
                 # equivalent to doing labels[..., 1:] and logits[..., :-1, :]
                 # But this way we dont need to slice the logits. We just add an ignore index to labels.
                 labels = torch.hstack(
                     (labels[..., 1:], self.ignore_labels_cache[: labels.shape[0]])
                 )
-                if not isinstance(logits, list):
+
+                if self.use_output_weight_in_loss:
```
SalmanMohammadi, line 809: very nice

recipes/full_finetune_distributed.py
SalmanMohammadi, line 351: tangential point: if the contract is that SFT losses follow the protocols defined in `loss_protocols`, do we need to make this check?

torchtune/modules/loss/cross_entropy_loss.py
SalmanMohammadi: nit: 
```suggestion
class MegaProjChunkyLossinator(nn.Module, SFTLossWithProjection):
```

torchtune/modules/loss/cross_entropy_loss.py
```diff
@@ -0,0 +1,114 @@
...
+        outputs: torch.Tensor,
+        targets: torch.Tensor,
+        *args,
+        **kwargs,
+    ) -> torch.Tensor:
+        """
+        Args:
+            weight (torch.Tensor): Tensor with weights of the model output projection layer. Shape [vocab_size, emb_dim]
+            outputs (torch.Tensor): Hidden state of the model, pre projection. Shape [bsz, seq_len, emb_dim]
+            targets (torch.Tensor): Labels for the model. Shape [bsz, seq_len]
```
SalmanMohammadi, line 91:
```suggestion
            weight (torch.Tensor): Tensor with weights of the model output projection layer. Shape ``[vocab_size, emb_dim]``
            outputs (torch.Tensor): Hidden state of the model, pre projection. Shape ``[bsz, seq_len, emb_dim]``
            targets (torch.Tensor): Labels for the model. Shape ``[bsz, seq_len]``
```

torchtune/modules/loss/loss_protocols.py
```diff
@@ -0,0 +1,67 @@
...
+            *args: Additional positional arguments.
+            **kwargs: Additional keyword arguments.
+        Returns:
+            torch.Tensor: loss tensor
+        """
+        ...
+
+
+class SFTLoss(Protocol):
+    """Protocol for loss functions in torchtune used in sft recipes."""
```
SalmanMohammadi, line 46:
```suggestion
    """Protocol for loss functions in torchtune used in SFT recipes."""
```

torchtune/modules/loss/loss_protocols.py
```diff
@@ -0,0 +1,67 @@
...
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+from typing import Protocol
+
+import torch
+
+
+class SFTLossWithProjection(Protocol):
+    """Protocol for loss functions in torchtune used in Supervised Finetune recipes and that require
```
SalmanMohammadi, line 13:
```suggestion
    """Protocol for loss functions in torchtune used in SFT recipes and that require
```

torchtune/modules/loss/loss_protocols.py
SalmanMohammadi: Yeah I do like protocols better. Since we're already in the `loss` module maybe just `protocols.py`?

torchtune/modules/loss/loss_protocols.py
pbontrager: I agree that this name is confusing. I think we should just standardize on "fused" or "linear", or "chunked". All the names have issues which we've discussed but if we're consistent at least people should be able to learn the term quickly.

torchtune/modules/loss/cross_entropy_loss.py
```diff
@@ -0,0 +1,114 @@
...
+        # Compute cross-entropy loss for the chunks
+        total_loss = 0.0
+        for idx in range(len(hidden_chunks)):
+            total_loss += self.compute_cross_entropy(
+                weight,
+                hidden_chunks[idx],
+                target_chunks[idx],
+            )
+
+        return total_loss / total_elements
```
pbontrager, line 114: nit: it'd be nice to offer the same 'reduction' option as most pytorch losses to control returning the mean, sum, or no reduction

recipes/lora_finetune_distributed.py
```diff
@@ -301,9 +301,12 @@ def setup(self, cfg: DictConfig) -> None:
         if self._compile:
```
pbontrager, line 301: What's the plan for rolling this out to the other sft recipes?

torchtune/modules/transformer.py
```diff
@@ -396,6 +400,7 @@ def __init__(
         self.head_dim = head_dim
         self.causal_mask = None
         self.num_output_chunks = 0
+        self._skip_output_projection = False
```
pbontrager, line 404: You should enforce in init that the output module has the "weight" property

pbontrager (approved):
Thanks for this big effort. This looks good and I'm happy to approve it now. Please finish going through and resolving the open comments before landing.

torchtune/modules/loss/loss_protocols.py
joecummings: ^ this

torchtune/modules/loss/loss_protocols.py
felipemello1: i believe it would override the module.compile method. We probably dont want it.

torchtune/modules/loss/loss_protocols.py
felipemello1: I dont know if i like "SFT" here, since it may not be obvious for a new reader what it means

torchtune/modules/loss/loss_protocols.py
felipemello1: I prefer "SFTI dont know if i like "SFT" here, since it may not be obvious for a new reader what it means

recipes/lora_finetune_distributed.py
felipemello1: 1. Recipes NOT being updated should still work with configs NOT being updated
2. Recipes being updated should NOT work anymore with old ce_with_chunked_outputs_loss
3. So any recipe that is changed also requires the configs to be updated with the new loss

TODO: need to check if the deprecation warnings work fine. This can be checked by running a recipe/config that has not been updated.

recipes/full_finetune_distributed.py
felipemello1: someone may try to use a loss that is not from torchtune, e.g. vanilla F.cross_entropy_loss

felipemello1: we should

recipes/dev/early_exit_finetune_distributed.py
```diff
@@ -357,6 +357,10 @@ def setup(self, cfg: DictConfig) -> None:
         if self._loss_fn.__class__.__name__ == "CEWithChunkedOutputLoss":
             # set num_output_chunks for model
             self._model.set_num_output_chunks(self._loss_fn.num_output_chunks)
+        elif getattr(self._loss_fn, "linear_loss", False):
+            raise ValueError(
+                "Linear losses are not supported yet for KD. Please use the deprecated CEWithChunkedOutputLoss."
+            )
```
ebsmothers, line 363: Wrong error message? Also can we open a high-priority issue for this? I don't like being in a state where half our configs are on a deprecated API

torchtune/modules/loss/ce_chunked_output_loss.py
```diff
@@ -29,6 +34,11 @@ class CEWithChunkedOutputLoss(torch.nn.Module):
 
     def __init__(self, num_output_chunks: int = 8, ignore_index: int = -100):
         super().__init__()
+        msg = (
+            "'CEWithChunkedOutputLoss' is deprecated and will be removed in future versions. "
+            "Please use `torchtune.modules.loss.LinearCrossEntropyLoss` instead."
+        )
+        log_once(logger=logger, msg=msg, level=logging.WARNING)
```
ebsmothers, line 41: nit: isn't this what the deprecated decorator is for?

torchtune/modules/loss/ce_chunked_output_loss.py
```diff
@@ -42,6 +52,13 @@ def compute_cross_entropy(
             logits.float(), labels, ignore_index=self.ignore_index, reduction="sum"
         )
 
+    def apply_compile_strategy(self, *args, **kwargs):
+        """Applies compile only to the fkl_loss function."""
```
ebsmothers, line 56: this isn't fkl?

torchtune/modules/model_fusion/_deep_fusion.py
```diff
@@ -96,6 +100,12 @@ def __init__(
...
         """Used to save memory in combination with :class:`~torchtune.modules.loss.CEWithChunkedOutputLoss`.
         This should be called before the first forward pass, in the recipe."""
+        msg = (
```
ebsmothers, line 103: same here, can we use deprecated decorator?

torchtune/training/_profiler.py
```diff
@@ -114,12 +114,6 @@ def trace_handler(
     # Memory timeline sometimes fails to export
     if prof.profile_memory and torch.cuda.is_available():
         if rank == 0:
-            try:
```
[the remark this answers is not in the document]
ebsmothers: bump

torchtune/modules/transformer.py
```diff
@@ -485,7 +497,20 @@ def reset_caches(self):
...
 
-    @torch.compiler.disable
+    @property
+    def linear_projection_weight(self) -> torch.Tensor:
+        """Returns the output weight matrix. Useful when a finer control of the output projection is needed,
+        for example when using a custom loss function or when interested in applying it to only some tokens.
+        """
+        # Accessing the weight directly will not trigger FSDP hooks
+        # to gather the full tensor so we have to unshard manually
+        if isinstance(self.output, FSDPModule):
```
ebsmothers, line 507: So this is not relevant for early fusion or deep fusion?

torchtune/modules/transformer.py
```diff
@@ -485,7 +497,20 @@ def reset_caches(self):
...
+            self.output.unshard()
+            weight = self.output.weight.clone()
```
ebsmothers, line 509: so the fix was the removal of detach? also any memory implications of the clone here?

torchtune/modules/loss/loss_protocols.py
ebsmothers: Yeah I agree, I don't think we should be giving a concrete implementation in the protocol?

torchtune/modules/loss/loss_protocols.py
```diff
@@ -0,0 +1,67 @@
...
+from typing import Protocol
+
+import torch
+
+
+class SFTLinearLoss(Protocol):
+    """Protocol for loss functions in torchtune used in Supervised Finetune recipes that require
+    model output linear projection weights in loss computation."""
+
+    linear_loss: bool = True
```
ebsmothers, line 16: Can we leave a comment in these classes explaining what this field means? Also an example usage in the docstrings of both would help a lot imo

torchtune/modules/loss/cross_entropy_loss.py
```diff
@@ -0,0 +1,115 @@
...
+import torch.nn.functional as F
+from torch import nn
+
+from .loss_protocols import SFTLinearLoss
+
+
+class LinearCrossEntropyLoss(nn.Module, SFTLinearLoss):
+    """Memory efficient Cross-entropy loss that incrementally computes loss for chunks of tokens
+    by masking ignored tokens, calculating logits and then applying cross-entropy loss. Combines
+    the linear projection with the cross-entropy calculation for futher memory savings.
```
ebsmothers, line 17: nit
```suggestion
    the linear projection with the cross-entropy calculation for further memory savings.
```

torchtune/modules/loss/cross_entropy_loss.py
```diff
@@ -0,0 +1,115 @@
...
+        ignore_index: int = -100,
+        mask_pre_projection: bool = True,
+    ):
+        super().__init__()
+        """
+        Args:
+            num_output_chunks (int): Number of chunks to split the output tensor into. Default is 8.
+            ignore_index (int): Index to ignore in the target tensor. Default is -100.
+            mask_pre_projection (bool): Whether to mask the output tensor before projection, avoiding
+                computing it for tokens that will be ignored during CE anyway. Default is True.
```
ebsmothers, line 32: Maybe I'm out of the loop here, but why do we need to expose this? This doesn't seem like an intuitive parameter to me, is there a reason someone would want to modify it?

ebsmothers (commented):
Just a few more comments, but really happy with how this turned out. This addresses a longstanding problem of inflexibility in our losses with a clear UX and opens us up to other more memory-efficient CE implementations.

recipes/dev/lora_finetune_distributed_multi_dataset.py
```diff
@@ -825,19 +826,22 @@ def train(self) -> None:
                 labels = batch.pop("labels")
 
                 with self.activations_handling_ctx:
-                    logits = self._model(**batch)
+                    outputs = self._model(**batch)
 
-                if not isinstance(logits, list):
+                if self.linear_loss:
```
Andrei-Aksionov, line 831: I have a question: what if we can encapsulate the logic specific to linear cross entropy in the `LinearCrossEntropy` class itself?
When we check whether it's a linear ce or not, we can assign `self._loss_fn.output_weights=...` and then just inject it during forward pass without a need to provide it explicitly.
This way we don't need a custom logic for loss calculation, so it could be unified for all losses.

Maybe it can somehow affect compilations? 🤷 
Or there is a plan to have a functional version of this loss in the future?
Or it's just plain dumb? 😆

recipes/dev/lora_finetune_distributed_multi_dataset.py
```diff
@@ -825,19 +826,22 @@ def train(self) -> None:
...
+                    weight = self._model.linear_projection_weight
+                    current_loss = self._loss_fn(weight, outputs, labels)
+                else:
                     labels = labels.reshape(-1)
-                    logits = logits.reshape(-1, logits.size(-1))
+                    outputs = outputs.reshape(-1, outputs.size(-1))
+                    current_loss = self._loss_fn(outputs, labels)
 
-                # Compute loss
                 # Loss is normalized by default so we multiply by the number of tokens
```
Andrei-Aksionov, line 839: Nit: is it indeed normalized? 
For me, it's more like "aggregated". 
Normalization is a different thing, no?

recipes/configs/code_llama2/7B_full_low_memory.yaml
```diff
@@ -63,7 +63,7 @@ optimizer:
   lr: 2e-5
 optimizer_in_bwd: True  # True saves memory. Requires gradient_accumulation_steps=1
 loss:
-  _component_: torchtune.modules.loss.CEWithChunkedOutputLoss
+  _component_: torchtune.modules.loss.LinearCrossEntropyLoss
```
Andrei-Aksionov, line 66: Idea: `turn on` compilation for LinearCrossEntropyLoss by default.
Without compilation there shouldn't be any benefits like online softmax, simultaneous logits and loss calculation, ...
So it won't be, em, linear at all 🤓

torchtune/modules/loss/ce_chunked_output_loss.py
Andrei-Aksionov: No, it's not :)

torchtune/modules/loss/cross_entropy_loss.py
Andrei-Aksionov: I'm also curious.

torchtune/modules/loss/cross_entropy_loss.py
```diff
@@ -0,0 +1,115 @@
...
+            **kwargs: Additional keyword arguments.
+        Returns:
+            torch.Tensor: loss tensor
+        """
+        # Total number of non-ignored tokens across the entire batch
+        mask = targets != self.ignore_index
+        total_elements = mask.sum()
+
+        # Chunk along sequence dimension
+        hidden_chunks = outputs.tensor_split(self.num_output_chunks, dim=1)
```
Andrei-Aksionov, line 103: Out of curiosity: does it make a difference in peak memory consumption in case of linear CE?

If compilation works as I anticipate it in case of LinearCrossEntropyLoss class, where logits calculation and loss calculation are within the same method and thus compilation can produce basically the same kernels as custom kernels for cut cross entropy, so if it's true, than chunking is no longer needed.

Maybe only when someone uses LinearCrossEntropyLoss without compilation 🤔

torchtune/modules/loss/loss_protocols.py
Andrei-Aksionov: Well, I am a new reader, and if I see all capitalized letters I immediately think that it's an abbreviation, not just a word.
Actually, it's an `Initialism` as I've just learned.

Andrei-Aksionov:
Hello @felipemello1 
Sorry for the late review of this PR. Just left a couple of comments.
But in overall, great job!

PR #2542, opened 2025-04-01
add resize_token_embedding functionality

iamzainhuda:
This PR introduces a resize_token_embeddings method which allows users to add additional tokens to be trained as part of fine tuning. It supports both single and multi device setups. The function can be easily extended to support additional initialization strategies.

For the distributed case, we unshard() the token embedding and output layer before passing into resize_token_embedding(). This way we can use the same logic as the single device case as the parameters are gathered on to the rank. Once complete, we fully_shard() the new token embedding and output layers utilizing the same mesh used for the initial sharding. This is the recommended way to resize FSDP2 sharded parameters. Thanks to @weifengpy for his guidance on the FSDP2 sharding mechanism.

The usage is,

On single device and distributed setups:

```python
model = setup_model(...)
tokenizer = setup_tokenizer(...)
resize_token_embeddings(model, tokenizer.vocab_size)
```

When using a recipe, (right now enabled for `full_finetune_distributed` and `full_finetune_single_device`), add `resize_token_embeddings = True` in YAML.
Example,
```yaml
...
max_steps_per_epoch: null
clip_grad_norm: null
compile: False  # torch.compile the model + loss, True increases speed + decreases memory
optimizer_in_bwd: False  # True saves memory. Requires gradient_accumulation_steps=1
resize_token_embeddings: True
...
````

#### Follow up tasks:
- additional embedding weight init methods
- loading checkpoint from resized model requires model to be created with new vocab size
- FusionEmbedding resizing as part of TransformerDecoder
 
#### Context
- [x] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

#### Changelog
* Added resize_token_embedding() utility that allows users to flexibly change the size of the token embeddings and corresponding output projection.

#### Test plan

Unit Testing:

Unit tests cover both single device and distributed cases, we check the new nn.Embedding size as well as the mean values of the new tokens to ensure they were init correctly. The E2E case covers testing a training run. We can see training completes successfully in the resized model that is sharded across 8 devices.

Without resizing, we can see special token init values on LLama 3 8B base model are very small, taking a look at token `<|start_header_id|> = 128006`
image

With resizing, we can see special tokens are init to the mean of 128k tokens on LLama 8B base:
image

##### E2E Test

Resizing the last 256 tokens in Llama 3 tokenizer to mean embedding of previous 120k tokens. Added an additional 4 tokens to resize the entire layer. We don't see significant difference in loss (run for 2 epochs, 20 steps each), but we show with the resized embedding in distributed setup we're able to finish training with at least equivalent loss.

image

vs. baseline
image

Running,
```
tune run --nproc_per_node 8 full_finetune_distributed --config recipes/configs/llama3/8B_full.yaml max_steps_per_epoch=20 epochs=2 checkpointer.checkpoint_dir=/tmp/Meta-Llama-3-8B
 ```

#### UX

### DIFF
--- a/recipes/full_finetune_distributed.py
+++ b/recipes/full_finetune_distributed.py
@@ -227,6 +227,16 @@ class FullFinetuneRecipeDistributed(FTRecipeInterface):
         self._enable_activation_offloading = cfg.get(
             "enable_activation_offloading", False
         )
+        self._activation_offloading_use_streams = cfg.get(
+            "activation_offloading_use_streams", True
+        )
+        if self._activation_offloading_use_streams and self.parallel_dims.tp_enabled:
+            warn(
+                message=(
+                    "Using activation offloading with streams is not advised in tensor parallel, and may "
+                    "cause unstable training. It is advised to set activation_offloading_use_streams: False"
+                )
+            )
         if self._enable_activation_offloading:
             if device_type != "cuda":
                 raise RuntimeError(
@@ -339,6 +349,7 @@ class FullFinetuneRecipeDistributed(FTRecipeInterface):
             cfg_model=cfg.model,
             enable_activation_checkpointing=self._enable_activation_checkpointing,
             enable_activation_offloading=self._enable_activation_offloading,
+            activation_offloading_use_streams=self._activation_offloading_use_streams,
             custom_sharded_layers=cfg.get("custom_sharded_layers", None),
             fsdp_cpu_offload=self.fsdp_cpu_offload,
             reshard_after_forward=cfg.get("fsdp_reshard_after_forward", True),
@@ -541,6 +552,7 @@ class FullFinetuneRecipeDistributed(FTRecipeInterface):
         cfg_model: DictConfig,
         enable_activation_checkpointing: bool,
         enable_activation_offloading: bool,
+        activation_offloading_use_streams: bool,
         fsdp_cpu_offload: bool,
         reshard_after_forward: bool,
         model_state_dict: Dict[str, Any],
@@ -659,7 +671,7 @@ class FullFinetuneRecipeDistributed(FTRecipeInterface):
 
         # activation offloading
         self.activations_handling_ctx = training.get_act_offloading_ctx_manager(
-            model, enable_activation_offloading
+            model, enable_activation_offloading, activation_offloading_use_streams
         )
 
         # Ensure no params and buffers are on meta device
--- a/torchtune/training/_activation_offloading.py
+++ b/torchtune/training/_activation_offloading.py
@@ -378,7 +378,7 @@ class NoOpManager(saved_tensors_hooks):
 
 
 def get_act_offloading_ctx_manager(
-    model: nn.Module, enable_activation_offloading: bool
+    model: nn.Module, enable_activation_offloading: bool, use_streams: bool = True
 ) -> Union[OffloadActivations, contextlib.nullcontext]:
     """Returns the activation offloading context manager for the model, which will be
     a null context if enable_activation_offloading is False.
@@ -390,6 +390,7 @@ def get_act_offloading_ctx_manager(
         model (nn.Module): the model to wrap with the activation offloading context manager.
         enable_activation_offloading (bool): whether or not to enable activation offloading
             for the model.
+        use_streams (bool): whether or not to enable streams for overlapping communication.
 
     Returns:
         contextlib.ContextDecorator: the activation offloading context manager for the model.
@@ -398,7 +399,7 @@ def get_act_offloading_ctx_manager(
         NotImplementedError: If the model is a multimodal model and activation offloading is enabled.
     """
     if enable_activation_offloading:
-        activations_handling_ctx = OffloadActivations()
+        activations_handling_ctx = OffloadActivations(use_streams=use_streams)
 
         # Below is our hack to disable offloading the last output Linear in every
         # step, as the cost for offloading the activation and then soon after bringing


recipes/full_finetune_distributed.py
```diff
@@ -340,6 +343,25 @@ def setup(self, cfg: DictConfig) -> None:
         )
         self._tokenizer = config.instantiate(cfg.tokenizer)
 
+        if self._tokenizer.vocab_size != self._model.tok_embeddings.num_embeddings:
+            # Unshard the FSDP wrapped token embeddings and output projection
+            # tok_embeddings and output are wrapped by fully_shard, we can call .unshard()
+            # direclty on those layers only
+            self._model.tok_embeddings.unshard()
```
ebsmothers, line 350: It's possible that `self._model` is not an instance of `TransformerDecoder`. E.g. for multimodal models we provide both [EarlyFusionModel](https://github.com/pytorch/torchtune/blob/5417a29ba96be78f11b23a67657ab4c4530b7f78/torchtune/modules/model_fusion/_early_fusion.py#L16) and [DeepFusionModel](https://github.com/pytorch/torchtune/blob/5417a29ba96be78f11b23a67657ab4c4530b7f78/torchtune/modules/model_fusion/_deep_fusion.py#L20), so we should at least be able to handle both of these gracefully. One option is to just do all this directly in the `resize_token_embeddings` function, otherwise we start to bloat the recipe code with lots of if/else checks.

recipes/full_finetune_distributed.py
```diff
@@ -340,6 +343,25 @@ def setup(self, cfg: DictConfig) -> None:
         )
         self._tokenizer = config.instantiate(cfg.tokenizer)
 
+        if self._tokenizer.vocab_size != self._model.tok_embeddings.num_embeddings:
+            # Unshard the FSDP wrapped token embeddings and output projection
+            # tok_embeddings and output are wrapped by fully_shard, we can call .unshard()
+            # direclty on those layers only
```
ebsmothers, line 349: super nit
```suggestion
            # directly on those layers only
```

torchtune/modules/transforms/tokenizers/_utils.py
```diff
@@ -12,7 +12,7 @@
 from torchtune.data._utils import truncate
 from torchtune.utils import get_logger
 
-_log: logging.Logger = get_logger()
+log: logging.Logger = get_logger()
```
ebsmothers, line 15: Can also probably revert the changes to this file and transformer.py now to match main

torchtune/modules/common_utils.py
```diff
@@ -442,3 +447,95 @@ def delete_kv_caches(model: nn.Module):
         if hasattr(module, "kv_cache") and callable(module.kv_cache):
             module.cache_enabled = False
             module.kv_cache = None
+
+
+def resize_token_embeddings(model: TransformerDecoder, new_num_tokens: int) -> None:
+    """
+    Resizes the token embeddings and the final output projection layer of a TransformerDecoder model.
+
+    This function modifies the model in-place.
```
ebsmothers, line 456: Would also explain (somewhere) in the docstring that you're initializing new positions using the mean of the old embeddings

torchtune/modules/common_utils.py
```diff
@@ -442,3 +447,95 @@ def delete_kv_caches(model: nn.Module):
...
+    if new_num_tokens > old_num_tokens:
+        mean_embedding = new_embeddings.weight.data[:n].mean(dim=0)
+        new_embeddings.weight.data[n:] = mean_embedding.unsqueeze(0)
+
+    new_embeddings.weight.requires_grad_(old_embeddings.weight.requires_grad)
+    model.tok_embeddings = new_embeddings
+
+    output_layer = model.output
+    if isinstance(output_layer, TiedLinear):
+        if output_layer.tied_module is old_embeddings:
```
ebsmothers, line 512: Dumb q: is this just checking equality of the underlying data_ptr?

ebsmothers (commented):
Thanks for adding this and for the thorough testing! Left a bunch of comments but this is looking pretty good. Lmk if anything I suggested is unclear

recipes/full_finetune_distributed.py
```diff
@@ -340,6 +343,25 @@ def setup(self, cfg: DictConfig) -> None:
...
+                self._model.output.unshard()
+
+            log.info(
+                f"Tokenizer vocab size ({self._tokenizer.vocab_size}) differs from model "
+                f"embedding size ({self._model.tok_embeddings.num_embeddings}). Resizing model embeddings..."
+            )
+            resize_token_embeddings(self._model, self._tokenizer.vocab_size)
+
+            # Reshard with new values, use the same mesh that was used in initial sharding
+            fully_shard(self._model.tok_embeddings, mesh=self.world_mesh[("dp_shard",)])
```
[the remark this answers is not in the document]
iamzainhuda: Yeah agreed, much cleaner to handle it within function vs. bloating all the recipes/callsites with additional setup just to use function - I'll make the change to infer from DTensor mesh.

On your idea, I agree that type of utility would be very nice to have (almost akin to summon_full_params from FSDP1, https://docs.pytorch.org/docs/stable/fsdp.html#torch.distributed.fsdp.FullyShardedDataParallel.summon_full_params). I imagine this use case would be quite common

recipes/full_finetune_distributed.py
iamzainhuda: Got it - I'll move all of this within the function and handle for the various types of model's that can be passed in

torchtune/modules/common_utils.py
iamzainhuda: Added - I've also added "init_strategy" optional arg to the function. Abstracted the resize op to it's own helper as we add additional init strategies it will be cleaner to keep their implementations in a helper. I am thinking we can add this in a follow up PR and also allow for user's to pass in a callable for init strategy?

torchtune/modules/common_utils.py
iamzainhuda: this is checking if object points to the same memory as tok_embedding's nn.Embedding, underlying data_ptr() isn't checked. it is better check is to use data_ptr() instead if we have output and tok_embeddings as seperate nn.Embeddings for any case.

however, if output_layer is tied, we should always replace the tied_module with new tok_embedding - i feel like this check is not needed?

recipes/full_finetune_distributed.py
```diff
@@ -319,6 +322,24 @@ def setup(self, cfg: DictConfig) -> None:
             ac_option=cfg.get("ac_option", None),
         )
         self._tokenizer = config.instantiate(cfg.tokenizer)
+       
+        if self._tokenizer.vocab_size != self._model.tok_embeddings.num_embeddings:
+            # Unshard the FSDP wrapped token embeddings and output projection
+            # tok_embeddings and output are wrapped by fully_shard, we can call .unshard()
+            # direclty on those layers only
+            self._model.tok_embeddings.unshard()
+            if not isinstance(self._model.output, TiedLinear): self._model.output.unshard()
```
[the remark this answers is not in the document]
iamzainhuda: If I understand correctly, the context manager would essentially handle the loop of d_tensor.full_tensor() --> user op --> redistribute_tensor(tensor, mesh)?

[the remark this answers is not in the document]
ebsmothers: Yeah exactly. Doesn't have to be done in this PR anyways, just an idea for a future quality-of-life improvement

torchtune/modules/common_utils.py
```diff
@@ -442,3 +448,131 @@ def delete_kv_caches(model: nn.Module):
         if hasattr(module, "kv_cache") and callable(module.kv_cache):
             module.cache_enabled = False
             module.kv_cache = None
+
+
+def resize_token_embeddings(model: nn.Module, num_embeddings: int) -> None:
```
pbontrager, line 453: This currently wouldn't work with fusion models or models that use the [FusionEmbedding](https://docs.pytorch.org/torchtune/stable/generated/torchtune.modules.model_fusion.FusionEmbedding.html#torchtune.modules.model_fusion.FusionEmbedding) layer. For a fusion model you want to check for the "decoder" and grab that instead of model.

To support resizing FusionEmbeddings you would want resize the embedding parameter and num_embeddings. It would also be reasonable to throw an error and not support resizing FusionEmbeddings as they're already a different technique for resizing embeddings but I don't think it would be too much harder to support them.

ebsmothers: Ah yeah this is my fault. I was thinking that the usage of `FusionEmbedding` means that the model itself is a fusion model, but forgot that `FusionEmbedding` can be the embedding layer for a `TransformerDecoder`. In that case I agree, we should make a slight change here. Will leave a diff of what I think makes sense here

torchtune/modules/common_utils.py
```diff
@@ -442,3 +448,131 @@ def delete_kv_caches(model: nn.Module):
...
+    """
+    need_resize = _need_to_resize(model, num_embeddings)
+    if not need_resize:
+        _log.info("Embedding resize not needed, skipping")
+        return
+
+    if not isinstance(model, TransformerDecoder) and need_resize:
+        raise AssertionError(
+            f"Resize needed but model is of type {type(model)}, please pass in TransformerDecoder to function"
+        )
```
ebsmothers, line 493: Something like this? @pbontrager @iamzainhuda 

```suggestion
    if isinstance(model, DeepFusionModel) or isinstance(model, EarlyFusionModel):
        model = model.decoder
    if isinstance(model.tok_embeddings, FusionEmbedding) and resize_needed:
        raise AssertionError(
            f"Resize is not supported for FusionEmbeddings"
        )
```

torchtune/modules/common_utils.py
iamzainhuda: I've added `FusionEmbedding` resizing into the follow up tasks. I agree from looking at the module, we can extend the function to resize the fusion_embedding instead with ease

Issue #2515, opened 2025-03-19
Unnecessarily scaling gradients when gradient_accumulation_steps is 1


shunting314:
The `scale_grads` call [here](https://github.com/pytorch/torchtune/blob/124123149c0fedc59c889da64c385df0f1a4d830/recipes/full_finetune_single_device.py#L706) will need to read and write each parameter once. For a 8B parameters model with bfloat16 dtype, that results in 8B * 2byte * 2 = 32GB memory access. Assuming a 2TBGS memory bandwidth, this will translate to 16ms latency. Note that this estimation is a lower bound. In practice, due to kernel launch overhead, this latency can be as larger as 42.5ms:

<img width="1169" alt="Image" src="https://github.com/user-attachments/assets/f160a3b4-3540-46f0-8c5e-5859d7f9a31f" /> (check the long blue bar)

Simple change like
```
diff --git a/recipes/full_finetune_single_device.py b/recipes/full_finetune_single_device.py
index 39e596ae..83007343 100644
--- a/recipes/full_finetune_single_device.py
+++ b/recipes/full_finetune_single_device.py
@@ -692,14 +692,21 @@ class FullFinetuneRecipeSingleDevice(FTRecipeInterface):

                 # Loss is normalized by default so we multiply by the number of tokens
                 # This way we can normalize by the total number of tokens if we're accumulating gradients
-                current_loss = self._loss_step(batch) * current_num_tokens
-                running_loss += current_loss
-                current_loss.backward()
+
+                if self._gradient_accumulation_steps > 1:
+                    current_loss = self._loss_step(batch) * current_num_tokens
+                    running_loss += current_loss
+                    current_loss.backward()
+                else:
+                    current_loss = self._loss_step(batch)
+                    running_loss += current_loss * current_num_tokens
+                    current_loss.backward()

                 # Step with optimizer
                 if (idx + 1) % self._gradient_accumulation_steps == 0:
                     if not self._optimizer_in_bwd:
-                        training.scale_grads(self._model, 1 / num_tokens)
+                        if self._gradient_accumulation_steps > 1:
+                            training.scale_grads(self._model, 1 / num_tokens)
                         if self._clip_grad_norm is not None:
                             grad_norm = torch.nn.utils.clip_grad_norm_(
                                 self._model.parameters(),
```

can just skip this `scale_grads` call when gradient accumulation steps is 1.

nathan-az:
[...]

PR #2543, opened 2025-04-01
RFC: Update-based training loop refactor, scale loss and token count calculation rather than gradients

nathan-az:
#### Context
- [ ] add a new feature
- [ ] fix a bug
- [x] other (enhancement)

https://github.com/pytorch/torchtune/issues/2515

#### Changelog
This motivation behind this PR was to remove gradient scaling. `num_tokens` is corrected to calculate the actual number of tokens in unique training examples for the update step, and `loss` is scaled accordingly (and adjusted for over-counting from non-dp workers). This required a larger refactor of the training loop to count up to `update_steps` rather than iterating through the dataloader, then grouping the prefetch of gradient accumulation batches.

**Upsides**

* Less nesting
* More logical grouping of gradient accumulation / forward/backward steps before update
* Variables now hold the correct value (e.g. `num_tokens` now contains the actual number of tokens across the update step, on each rank)
* Implicitly corrects calculation of TPS per GPU
* Use of new context setters for methods like HSDP (e.g. `is_last_backwards_step`) can be localised to the forward/backward loop which should be quite clean
* Pattern can relatively painlessly be reverted
* Added a `num_tokens` metric - nice metric for gauging how effective packing a given dataset is

**Downsides**

* Currently does not account for future use of loss parallel, context parallel, etc.
* Profiler wait/warmup steps are now relative to _update_ steps rather than _forward_ steps. This may actually be desirable (I assumed this was the case until this refactor, and makes use of the term `steps` more consistent)
* Prefetching examples technically uses slightly more memory (pedantic note, but still)
* Likely to cause merge conflicts with in-flight work which includes changes to the training loop for this recipe

I recognise that this is beyond the scope of the initial motivation, but I think this format may be preferable in terms of readability and flexibility.

Losses are near identical (seeded with cudnn determinism mode enabled). There is very minor TPS benefit to avoiding the grad scaling. This is more substantial without gradient accumulation since the parameter read/write occurs more frequently. It may be more notable with much larger models.

image

#### Test plan

#### UX

### DIFF
--- a/recipes/full_finetune_distributed.py
+++ b/recipes/full_finetune_distributed.py
@@ -791,13 +791,13 @@ class FullFinetuneRecipeDistributed(FTRecipeInterface):
                     except StopIteration:
                         break
 
-                    utils.batch_to_device(batch, self._device)
-
                     # Calculate the number of unmasked tokens in the current batch
                     # and increment the total number of tokens seen in the step
                     current_num_tokens = (
-                        batch["labels"] != self._loss_fn.ignore_index
-                    ).sum()
+                        (batch["labels"] != self._loss_fn.ignore_index)
+                        .sum()
+                        .to(self._device)
+                    )
                     num_tokens += current_num_tokens
                     batches.append((batch, current_num_tokens.item()))
 
@@ -812,6 +812,7 @@ class FullFinetuneRecipeDistributed(FTRecipeInterface):
                 num_tokens = num_tokens / self.parallel_dims.non_data_parallel_size
 
                 for batch, token_count in batches:
+                    utils.batch_to_device(batch, self._device)
                     # Shape [b, s], needed for the loss not the model
                     labels = batch.pop("labels")
 
--- a/recipes/full_finetune_distributed.py
+++ b/recipes/full_finetune_distributed.py
@@ -767,147 +767,160 @@ class FullFinetuneRecipeDistributed(FTRecipeInterface):
         for curr_epoch in range(self.epochs_run, self.total_epochs):
             pbar = tqdm(total=self._steps_per_epoch, disable=not self._is_rank_zero)
             self._dataloader.sampler.set_epoch(curr_epoch)
-            for idx, batch in enumerate(self._dataloader):
+
+            dataloader_iter = iter(self._dataloader)
+
+            for update_step in range(self._steps_per_epoch):
                 # Start tracking CUDA memory for active steps for just the first epoch
                 if (
                     self._is_rank_zero
                     and curr_epoch == 0
                     and self.profiler_profile_memory
-                    and idx == self.profiler_wait_steps + self.profiler_warmup_steps
+                    # TODO: confirm we want these steps according to *update steps*, not *forward steps*
+                    and update_step
+                    == self.profiler_wait_steps + self.profiler_warmup_steps
                     and self._device.type == "cuda"
                 ):
                     torch.cuda.memory._record_memory_history()
-                utils.batch_to_device(batch, self._device)
 
-                # Calculate the number of unmasked tokens in the current batch
-                # and increment the total number of tokens seen in the step
-                current_num_tokens = (
-                    batch["labels"] != self._loss_fn.ignore_index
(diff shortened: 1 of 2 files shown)


nathan-az:
Thanks for the comment @pbontrager! I like the idea of separating the gradient accumulation logic per your branch (will leave that discussion for a separate thread). I have some thoughts on your points:

> I wonder if it'll always be minor if you're dealing with multimodal data

Agreed - in extreme cases inputs might consume an appreciable amount of VRAM (e.g. very large batches per device).

I've [delayed the move to GPU](https://github.com/pytorch/torchtune/pull/2543/commits/204f36e147ff7545c1737dbaac255ac445f2e638) to only include the example in the forward pass. I expect this is fine since RAM is more plentiful. I was concerned about the sum op on CPU, but testing on my very unimpressive home machine (Ryzen 5600x), performing this operation on batches of size (16, 2**17) indices required less than 2ms.

> Another concern is that prefetching puts more pressure on the dataloader as training is blocked until it prepares n batches

Do we not pay this either way (is the total time of blocked operations not equal whether we front-load the examples for an update step or do it between forward/backward passes?)

Even so - I expect the time cost here is much less than the parameter read/write time during gradient scaling, especially for large models (and my testing showed that TPS still increased).

If we did want to tackle this, could it be solved with a combination of`prefetch_factor` (I assume this is async), and `num_workers`? Again this might be overkill - I expect this is trivially fast, but I'm curious your thoughts.

Let me know what you think on the above, or if you have other thoughts!

ebsmothers:
Hey @nathan-az sorry for the delay in getting to this one. Tbh I am a bit on the fence here -- I took a look at HF's prefetching approach when first implementing our version and I didn't like certain aspects (e.g. they tied the total number of tokens into the loss). It seems like your approach avoids that, which is nice (and it gets us out of the usage of scale_grads, which can cause its own perf hit). Overall though I am a bit wary -- maybe let's have a broader design discussion about what the right UX is prior to moving forward with this PR. Personally I think the prefetching logic makes the training loop harder to parse (I acknowledge the current token normalization logic is also tricky though), but maybe there is a way we can sequence things here. E.g. what would a good training utility for prefetching look like? If we can agree on that, then maybe we could see what the experience is integrating that utility into the recipe. Lmk if that makes sense to you. And thanks for opening this RFC! I think it's a worthwhile discussion for us to have.

joecummings:
> Hey @nathan-az sorry for the delay in getting to this one. Tbh I am a bit on the fence here -- I took a look at HF's prefetching approach when first implementing our version and I didn't like certain aspects (e.g. they tied the total number of tokens into the loss). It seems like your approach avoids that, which is nice (and it gets us out of the usage of scale_grads, which can cause its own perf hit). Overall though I am a bit wary -- maybe let's have a broader design discussion about what the right UX is prior to moving forward with this PR. Personally I think the prefetching logic makes the training loop harder to parse (I acknowledge the current token normalization logic is also tricky though), but maybe there is a way we can sequence things here. E.g. what would a good training utility for prefetching look like? If we can agree on that, then maybe we could see what the experience is integrating that utility into the recipe. Lmk if that makes sense to you. And thanks for opening this RFC! I think it's a worthwhile discussion for us to have.

Also, please feel free to join our Discord if you haven't already: https://discord.gg/N4aVcgHS! Might be faster to communicate there.

nathan-az:
Thanks @joecummings (have joined the Discord server now!) and @ebsmothers for the comments!

I'll close this PR for now so it doesn't clog the list, but hopefully it can be used as a reference for a future Discussion thread (and if one is made I'll happily weigh in!)

To summarise my thoughts - I agree the prefetch is not the most intuitive, although I personally find the logical grouping of the forward/backward steps then update step more interpretable. In addition I think this loss scaling is a bit more interpretable than the grad scaling done currently (and it is more performant).

However I also agree that there is likely a better pattern that achieves all of this. And I understand it's likely not the highest priority with open PRs like LLaMA-4 support (including tp2ep, woo!) and fp8 training support in-flight!

Issue #2537, opened 2025-03-30
"Missing LoRA key from adapter state dict" when attempting to resume LoRA training from adapter checkpoint


prvnsmpth:
I'm fine-tuning Qwen2.5 7B with LoRA using the `lora_finetune_single_device.py` recipe. 

I trained on only 25% of the training set first, and saved the intermediate checkpoint. Here's what the output dir looks like:

```
❯ tree .
.
├── epoch_0
│   ├── adapter_config.json
│   ├── adapter_model.pt
│   ├── adapter_model.safetensors
│   ├── config.json
│   ├── generation_config.json
│   ├── LICENSE
│   ├── merges.txt
│   ├── model-00001-of-00004.safetensors
│   ├── model-00002-of-00004.safetensors
│   ├── model-00003-of-00004.safetensors
│   ├── model-00004-of-00004.safetensors
│   ├── model.safetensors.index.json
│   ├── original_repo_id.json
│   ├── README.md
│   ├── tokenizer_config.json
│   ├── tokenizer.json
│   └── vocab.json
...
...
├── logs
│   ├── log_1743138520.txt
│   ├── log_1743138825.txt
│   ├── log_1743138844.txt
│   ├── log_1743139054.txt
│   └── log_1743139370.txt
├── recipe_state
│   └── recipe_state.pt
``` 

Then, I decided to resume training from this checkpoint. As mentioned in the docs, I set `resume_from_checkpoint: True` in my config, and also set `adapter_checkpoint` to point to the safetensors file above. Here's the relevant portion of the config:
```
checkpointer:
  _component_: torchtune.training.FullModelHFCheckpointer
  checkpoint_dir: ${model_dir}
  checkpoint_files: [
    model-00001-of-00004.safetensors,
    model-00002-of-00004.safetensors,
    model-00003-of-00004.safetensors,
    model-00004-of-00004.safetensors,
  ]
  recipe_checkpoint: null
  adapter_checkpoint: epoch_0/adapter_model.safetensors
  output_dir: ${output_dir}
  model_type: QWEN2
resume_from_checkpoint: True
```

`model_dir` points to the location of the unmerged base model.

However, when attempting to resume training, `validate_missing_and_unexpected_for_lora` fails with the following error:
```
  File "/opt/conda/lib/python3.10/site-packages/torchtune/modules/peft/_utils.py", line 312, in validate_missing_and_unexpected_for_lora
[...]

PR #2544, opened 2025-04-01
improve ckpt error handling and docs

felipemello1:
#### Context
- [ ] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

In https://github.com/pytorch/torchtune/issues/2537 user had issues because they checked docs for 0.5, but there were still opportunities to make it crystal clear to the user.

Addressed a couple of TODOs, improved error messages and docs.

#### Testing

Resumed from ckpt:
a) normally
b) wrong adapter path
c) adapter.safetensor

### DIFF
--- a/docs/source/deep_dives/checkpointer.rst
+++ b/docs/source/deep_dives/checkpointer.rst
@@ -419,16 +419,17 @@ Resuming from checkpoint - LoRA Finetuning
 ------------------------------------------
 
 Similarly to full finetuning, we will also only need to modify two fields: ``resume_from_checkpoint``
-and ``adapter_checkpoint``, which will be loaded from output_dir. We do not have to modify ``checkpoint_files``,
-because the base model being loaded is still the same.
+and ``adapter_checkpoint``, which will be loaded from output_dir. We do NOT have to modify ``checkpoint_files``,
+because the base model being loaded is still the same. You can optionally leave ``adapter_checkpoint`` empty.
+In this case, we will look for it in the last saved epoch folder.
 
 .. code-block:: yaml
 
     checkpointer:
         # [... rest of the config...]
 
-        # adapter_checkpoint. Note that you will need to update this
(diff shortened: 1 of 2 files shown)


docs/source/deep_dives/checkpointer.rst
```diff
@@ -419,16 +419,17 @@ Resuming from checkpoint - LoRA Finetuning
 ------------------------------------------
 
 Similarly to full finetuning, we will also only need to modify two fields: ``resume_from_checkpoint``
-and ``adapter_checkpoint``, which will be loaded from output_dir. We do not have to modify ``checkpoint_files``,
-because the base model being loaded is still the same.
+and ``adapter_checkpoint``, which will be loaded from output_dir. We do NOT have to modify ``checkpoint_files``,
```
SalmanMohammadi, line 422:
```suggestion
and ``adapter_checkpoint``, which will be loaded from ``output_dir``. We do NOT have to modify ``checkpoint_files``,
```

torchtune/training/checkpointing/_utils.py
```diff
@@ -460,6 +460,9 @@ def get_adapter_checkpoint_path(
 
     Returns:
         Optional[Path]: Path to the adapter checkpoint file, or None if not applicable.
+
+    Raises:
+        ValueError: If the adapter checkpoint file is missing or If the adapter checkpoint file is not a .pt file.
```
SalmanMohammadi, line 465:
```suggestion
        ValueError: If the adapter checkpoint file is missing or if the adapter checkpoint file is not a .pt file.
```
Out of curiosity, when would adapter checkpoint not be  a .pt file?

felipemello1: we save both .safetensors (for HF) and .pt. Its ugly, but we dont have a function to load ckpt as .safetensors.

PR #2404, opened 2025-02-17
(WIP/RFC) FP8 full finetune distributed

nathan-az:
#### Context
What is the purpose of this PR? Is it to
- [x] add a new feature
- [ ] fix a bug
- [ ] update tests and/or documentation
- [ ] other (please add here)

This would solve https://github.com/pytorch/torchtune/issues/2201.

I'm far from an expert with `torchao` but tested this in an effort to further reduce memory usage, combining it with `torchao.prototype.low_bit_optim.AdamW8bit` to get LLaMA 3.3 70B training comfortably on 8x H100s with sequence length over 8k and only activation offloading.

I saw a significant (50%) improvement in tokens per second using this. A sense check shows near identical loss curves as well.

image

This PR should probably be treated equal parts as a PR and RFC, or a reference for a future one. Key points for discussion are:

* does this code make sense or conflict with any current features? (e.g. is it correct to do the conversion on the meta device before compilation)
* does the new functionality sit in the correct place in the `torchtune.memory` module, and are we happy with using the class as I have
* what is the best way to test this and ensure correctness and (near) equality versus the standard (bf16) training

#### Changelog
What are the changes made in this PR?
This implements a WIP version of fp8 training largely based on torchtitan's Float8Converter.

#### Test plan
Please make sure to do each of the following if applicable to your PR. If you're unsure about any one of these just ask and we will happily help. We also have a [contributing page](https://github.com/pytorch/torchtune/blob/main/CONTRIBUTING.md) for some guidance on contributing.

- [x] run pre-commit hooks and linters (make sure you've first installed via `pre-commit install`)
- [ ] add [unit tests](https://github.com/pytorch/torchtune/tree/main/tests/torchtune) for any new functionality
- [ ] update [docstrings](https://github.com/pytorch/torchtune/tree/main/docs/source) for any new or updated methods or classes
- [ ] run unit tests via `pytest tests`
- [ ] run recipe tests via `pytest tests -m integration_test`
- [ ] manually run any new or modified recipes with sufficient proof of correctness
- [ ] include relevant commands and any other artifacts in this summary (pastes of loss curves, eval results, etc.)

#### UX
If your function changed a public API, please add a dummy example of what the user experience will look like when calling it.
Here is a [docstring example](https://github.com/pytorch/torchtune/blob/6a7951f1cdd0b56a9746ef5935106989415f50e3/torchtune/modules/vision_transformer.py#L285)
[...]

PR #2546, opened 2025-04-01
Enable FP8 full finetune distributed

andrewor14:
**Summary:** This commit adds FP8 finetuning to the `full_finetune_distributed` recipe as an optional feature. For Llama3-8B, we saw up to 14.7% improvement in finetuning throughput with no degradation in memory usage or accuracy. This feature is currently gated on PyTorch nightlies since it depends on recent features added there. However, it will be available in the next torchtune release.

To use this feature, add the following to your config.yaml:
```
enable_fp8_training: true
fp8_recipe_name: tensorwise  # or rowwise, or rowwise_with_gw_hp
```

The default setting uses tensorwise scaling + `enable_fsdp_float8_all_gather=True`, which led to the largest speedups in our experiments.

Based on https://github.com/pytorch/torchtune/pull/2404 by @nathan-az

**Experimentation:** All experiments were run on 4x H100 GPUs with 94GB memory each. We finetune the model on the cleaned alpaca dataset for 1 epoch, using a batch size of 16 with torch.compile. We use the following commits from all 3 repos:

```
torchtune: b818e03 (https://github.com/andrewor14/torchtune/blob/fp8-finetuning)
torchao: 5a78b70
torch: 1017927
```

For Llama3-8B, fp8 finetuning saw 14.7% faster finetuning with no change in memory usage or quantized accuracy compared to the bf16 baseline:

```
experiment_name         tok/s                peak_mem_active    peak_mem_alloc    peak_mem_reserved
----------------------  -------------------  -----------------  ----------------  -------------------
full                    2773.473 (+0.000%)   18.481 (+0.000%)   18.481 (+0.000%)  34.291 (+0.000%)
fp8_noname              3182.220 (+14.738%)  18.484 (+0.014%)   18.484 (+0.014%)  34.325 (+0.097%)
fp8_tensorwise          3159.676 (+13.925%)  18.484 (+0.014%)   18.484 (+0.014%)  34.325 (+0.097%)
fp8_rowwise             2790.424 (+0.611%)   18.496 (+0.078%)   18.496 (+0.078%)  34.327 (+0.103%)

experiment_name         hellaswag_acc    wikitext_word_perplexity
----------------------  ---------------  --------------------------
full                    0.584 (+0.000)   9.419 (+0.000)
fp8_noname              0.585 (+0.000)   9.431 (+0.012)
fp8_tensorwise          0.584 (+0.000)   9.421 (+0.002)
fp8_rowwise             0.583 (-0.002)   9.421 (+0.002)
```

A few more observations here:
- The best tok/s improvement was from the default setting (`fp8_noname`)
- `fp8_rowwise` was the worst fp8 configuration, though still marginally better than the baseline

For Llama3.1-8B, we observed similar observations, with up to 14.3% faster finetuning and no change in quantized accuracy. However, memory usage did increase minimally (+2%) for most fp8 settings:

```
experiment_name         tok/s                peak_mem_active    peak_mem_alloc    peak_mem_reserved
----------------------  -------------------  -----------------  ----------------  -------------------
full                    2768.292 (+0.000%)   18.541 (+0.000%)   18.541 (+0.000%)  34.270 (+0.000%)
fp8_noname              3164.370 (+14.308%)  18.542 (+0.008%)   18.542 (+0.008%)  34.963 (+2.021%)
fp8_tensorwise          3136.952 (+13.317%)  18.542 (+0.008%)   18.542 (+0.008%)  34.963 (+2.021%)
fp8_rowwise             2790.672 (+0.808%)   18.554 (+0.073%)   18.554 (+0.073%)  34.389 (+0.348%)
fp8_rowwise_with_gw_hp  3144.678 (+13.596%)  18.551 (+0.056%)   18.551 (+0.056%)  34.966 (+2.032%)

experiment_name         hellaswag_acc    wikitext_word_perplexity
----------------------  ---------------  --------------------------
full                    0.594 (+0.000)   9.087 (+0.000)
fp8_noname              0.593 (-0.001)   9.070 (-0.017)
fp8_tensorwise          0.593 (-0.001)   9.061 (-0.026)
fp8_rowwise             0.593 (-0.000)   9.086 (-0.001)
fp8_rowwise_with_gw_hp  0.595 (+0.001)   9.087 (+0.000)
```

Llama3.2-3B saw up to 16.5% faster finetuning for rowwise with high precision `grad_weight`, which is a bigger improvement than just tensorwise. Similarly, there are no degradations in memory usage or quantized accuracy.
```
experiment_name         tok/s                peak_mem_active    peak_mem_alloc    peak_mem_reserved
----------------------  -------------------  -----------------  ----------------  -------------------
full                    6502.143 (+0.000%)   15.917 (+0.000%)   15.917 (+0.000%)  30.090 (+0.000%)
fp8_noname              7205.386 (+10.816%)  15.917 (+0.003%)   15.917 (+0.003%)  30.010 (-0.266%)
fp8_tensorwise          7222.198 (+11.074%)  15.917 (+0.003%)   15.917 (+0.003%)  30.010 (-0.266%)
fp8_rowwise             6387.968 (-1.756%)   15.916 (-0.002%)   15.916 (-0.002%)  29.158 (-3.096%)
fp8_rowwise_with_gw_hp  7573.698 (+16.480%)  15.917 (+0.001%)   15.917 (+0.001%)  29.516 (-1.908%)

experiment_name         hellaswag_acc    wikitext_word_perplexity
----------------------  ---------------  --------------------------
full                    0.533 (+0.000)   12.407 (+0.000)
fp8_noname              0.533 (+0.000)   12.414 (+0.007)
fp8_tensorwise          0.533 (+0.000)   12.412 (+0.005)
fp8_rowwise             0.533 (-0.000)   12.420 (+0.013)
fp8_rowwise_with_gw_hp  0.534 (+0.001)   12.416 (+0.009)
```

**Test Plan:**

Experiment command:
```
tune run --nnodes 1 --nproc_per_node 4 full_finetune_distributed --config llama3/8B_full \
    enable_fp8_training=true \
    fp8_recipe_name=tensorwise \
    epochs=1 \
    batch_size=16 \
    compile=true \
    dataset._component_=torchtune.datasets.alpaca_cleaned_dataset \
    checkpointer.output_dir="$LOG_DIR" \
    output_dir="${LOG_DIR}/metrics" \
    metric_logger.log_dir="${LOG_DIR}/metrics"
```
(full script:
https://github.com/andrewor14/torchtune/blob/fp8-finetuning-debug/run_it.sh)

Unit tests:
```
pytest tests -k test_convert_to_float8_training
pytest tests -k test_is_fp8_tensorwise_scaling
```

### DIFF
--- a/recipes/full_finetune_distributed.py
+++ b/recipes/full_finetune_distributed.py
@@ -19,6 +19,7 @@ from torch.distributed import destroy_process_group, init_process_group
 from torch.distributed._tensor import DTensor
 from torch.distributed.tensor.parallel import parallelize_module
 from torch.optim import Optimizer
+from torchao.float8 import precompute_float8_dynamic_scale_for_fsdp
 from torchdata.stateful_dataloader import StatefulDataLoader
 from torchdata.stateful_dataloader.sampler import StatefulDistributedSampler
 from torchtune import config, modules, training, utils
@@ -33,6 +34,10 @@ from torchtune.training.checkpointing._checkpoint_client import (
     TrainingProgress,
 )
 from torchtune.training.lr_schedulers import get_lr
+from torchtune.training.quantization import (
+    convert_to_float8_training,
+    is_fp8_tensorwise_scaling,
+)
 
 from tqdm import tqdm
 
@@ -184,6 +189,8 @@ class FullFinetuneRecipeDistributed(FTRecipeInterface):
         self._optimizer_in_bwd = cfg.get("optimizer_in_bwd", False)
         self._clip_grad_norm = cfg.get("clip_grad_norm", None)
         self._checkpoint_client = CheckpointClient(cfg)
+        self._enable_fp8_training = cfg.get("enable_fp8_training", False)
+        self._fp8_recipe_name = cfg.get("fp8_recipe_name", None)
 
         self._run_val_every_n_steps = cfg.get("run_val_every_n_steps", None)
         if self._run_val_every_n_steps is not None:
@@ -567,6 +574,19 @@ class FullFinetuneRecipeDistributed(FTRecipeInterface):
         if self._compile:
             training.compile_model(model, verbose=self._is_rank_zero)
 
+        if self._enable_fp8_training:
+            # Requires https://github.com/pytorch/pytorch/pull/148922
+            if torch.__version__ < "2.8.0.dev20250318":
+                raise RuntimeError(
+                    "Float8 fine-tuning requires PyTorch 2.8.0.dev20250318 or later."
+                )
+            if self.tp_plan is not None:
+                raise ValueError(
+                    "FP8 training does not support tensor parallelism yet. "
+                    "This will be enabled in the near future."
+                )
+            model = convert_to_float8_training(model, self._fp8_recipe_name)
+
         # Apply tensor parallelism to the model
         if self.parallel_dims.tp_enabled:
             if not self.parallel_dims.dp_enabled and self.fsdp_cpu_offload:
@@ -921,6 +941,16 @@ class FullFinetuneRecipeDistributed(FTRecipeInterface):
                     # Step the learning rate scheduler
                     if self._lr_scheduler is not None:
                         self._lr_scheduler.step()
+
+                    # If float8 training is enabled, perform a single all-reduce to compute the
+                    # scale for all float8 parameters efficiently instead of doing many small
+                    # all-reduces for each parameter
+                    if (
+                        self._enable_fp8_training
+                        and is_fp8_tensorwise_scaling(self._fp8_recipe_name)
+                        and self.dp_degree > 1
+                    ):
+                        precompute_float8_dynamic_scale_for_fsdp(self._model)
 
                     loss_to_log = running_loss.detach().item() / num_tokens
                     pbar.update(1)
--- a/torchtune/models/llama3/_parallelism.py
+++ b/torchtune/models/llama3/_parallelism.py
@@ -4,7 +4,7 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
-from typing import Dict
+from typing import Dict, Type
 
 from torch import nn
 
@@ -17,31 +17,56 @@ from torch.distributed.tensor.parallel import (
 )
 from torch.distributed.tensor.parallel.style import ParallelStyle
 
-# Define the Tensor Parallel plan for Llama3 model, which will also be shared with 3.1, 3.2, and 3.3 models
-BASE_LLAMA_TP_TRAINING_PLAN = {
-    "tok_embeddings": RowwiseParallel(
-        input_layouts=Replicate(), output_layouts=Shard(1)
-    ),
-    "norm": SequenceParallel(),
-    "output": ColwiseParallel(input_layouts=Shard(1), output_layouts=Replicate()),
-    "layers.*.attn": PrepareModuleInput(
-        input_layouts=(Shard(1), None),
-        desired_input_layouts=(Replicate(), None),
-    ),
-    "layers.*.mlp": PrepareModuleInput(
-        input_layouts=(Shard(1),),
-        desired_input_layouts=(Replicate(),),
-    ),
-    "layers.*.sa_norm": SequenceParallel(),
-    "layers.*.mlp_norm": SequenceParallel(),
-    "layers.*.attn.q_proj": ColwiseParallel(),
-    "layers.*.attn.k_proj": ColwiseParallel(),
-    "layers.*.attn.v_proj": ColwiseParallel(),
-    "layers.*.attn.output_proj": RowwiseParallel(output_layouts=Shard(1)),
-    "layers.*.mlp.w1": ColwiseParallel(),
-    "layers.*.mlp.w2": RowwiseParallel(output_layouts=Shard(1)),
-    "layers.*.mlp.w3": ColwiseParallel(),
-}
+from torchao.float8.float8_tensor_parallel import (
+    Float8ColwiseParallel,
+    Float8RowwiseParallel,
+    PrepareFloat8ModuleInput,
+)
+
+
+def _get_base_llama_tp_training_plan(
+    layerwise_colwise_parallel_cls: Type[ParallelStyle] = ColwiseParallel,
+    layerwise_rowwise_parallel_cls: Type[ParallelStyle] = RowwiseParallel,
+    layerwise_prepare_module_input_cls: Type[ParallelStyle] = PrepareModuleInput,
+) -> Dict[str, ParallelStyle]:
+    """
+    Define the Tensor Parallel plan for Llama3 model, which will also be shared with 3.1, 3.2, and 3.3 models.
+    """
+    return {
+        "tok_embeddings": RowwiseParallel(
+            input_layouts=Replicate(), output_layouts=Shard(1)
+        ),
+        "norm": SequenceParallel(),
+        "output": ColwiseParallel(input_layouts=Shard(1), output_layouts=Replicate()),
+        "layers.*.attn": layerwise_prepare_module_input_cls(
+            input_layouts=(Shard(1), None),
+            desired_input_layouts=(Replicate(), None),
+        ),
+        "layers.*.mlp": layerwise_prepare_module_input_cls(
+            input_layouts=(Shard(1),),
+            desired_input_layouts=(Replicate(),),
+        ),
+        "layers.*.sa_norm": SequenceParallel(),
+        "layers.*.mlp_norm": SequenceParallel(),
+        "layers.*.attn.q_proj": layerwise_colwise_parallel_cls(),
+        "layers.*.attn.k_proj": layerwise_colwise_parallel_cls(),
+        "layers.*.attn.v_proj": layerwise_colwise_parallel_cls(),
+        "layers.*.attn.output_proj": layerwise_rowwise_parallel_cls(
+            output_layouts=Shard(1)
+        ),
+        "layers.*.mlp.w1": layerwise_colwise_parallel_cls(),
+        "layers.*.mlp.w2": layerwise_rowwise_parallel_cls(output_layouts=Shard(1)),
+        "layers.*.mlp.w3": layerwise_colwise_parallel_cls(),
+    }
+
+
+BASE_LLAMA_TP_TRAINING_PLAN = _get_base_llama_tp_training_plan()
+
+FP8_LLAMA_TP_TRAINING_PLAN = _get_base_llama_tp_training_plan(
+    layerwise_colwise_parallel_cls=Float8ColwiseParallel,
+    layerwise_rowwise_parallel_cls=Float8RowwiseParallel,
+    layerwise_prepare_module_input_cls=PrepareFloat8ModuleInput,
+)
 
 BASE_LLAMA_TP_INFERENCE_PLAN = {
     "tok_embeddings": RowwiseParallel(input_layouts=Replicate()),
@@ -70,3 +95,16 @@ def base_llama_tp_plan(
         Dict[str, Any]: The tensor parallel plan for Llama3 model.
     """
     return BASE_LLAMA_TP_INFERENCE_PLAN if inference else BASE_LLAMA_TP_TRAINING_PLAN
+
+
+# TODO: expose this once tested
+def _fp8_llama_tp_plan() -> Dict[str, ParallelStyle]:
+    """
+    Return the tensor parallel plan for Llama3 model that uses float8 for all-gather for both
+    rowwise and colwise computation, currently only compatible with float8 fine-tuning with
+    "tensorwise" scaling. This tensor parallel plan is shared between 3.1, 3.2, and 3.3 models.
+
+    Returns:
+        Dict[str, Any]: The float8-enabled tensor parallel plan for Llama3 model.
+    """
+    return FP8_LLAMA_TP_TRAINING_PLAN
--- a/torchtune/training/quantization.py
+++ b/torchtune/training/quantization.py
@@ -4,18 +4,25 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
-from typing import Callable, Optional
+from typing import Callable, Dict, Optional
 
 from torch import nn
+from torch.distributed.tensor.parallel.style import ParallelStyle
 
 from torchao.dtypes import TensorCoreTiledLayout
-
+from torchao.float8 import (
+    convert_to_float8_training as _convert_to_float8_training_torchao,
+    Float8LinearConfig,
+)
+from torchao.float8.float8_tensor_parallel import (
+    Float8ColwiseParallel,
+    Float8RowwiseParallel,
+)
 from torchao.quantization import (
     int4_weight_only,
     int8_dynamic_activation_int4_weight,
     quantize_,
 )
-
 from torchao.quantization.qat import (
     Int4WeightOnlyQATQuantizer,
     Int8DynActInt4WeightQATQuantizer,
@@ -26,6 +33,7 @@ from torchao.quantization.qat.linear import (
     enable_4w_fake_quant,
     enable_8da4w_fake_quant,
 )
+
 from torchtune.modules.peft.lora import LoRALinear, QATLoRALinear
 
 
@@ -219,3 +227,61 @@ def swap_lora_linear_with_qat(
                 activation_qat_config,
                 weight_qat_config,
             )
+
+
+def convert_to_float8_training(
+    model: nn.Module,
+    fp8_recipe_name: Optional[str] = None,
+) -> nn.Module:
+    """
+    Prepare the model for float8 training by swapping all `nn.Linear` with `Float8Linear`.
+
+    Args:
+        model (nn.Module): The model to swap linear layers on
+        fp8_recipe_name (Optional[str]): name to identify one of the pre-made recipes,
+            one of "tensorwise", "rowwise", and "rowwise_with_gw_hp". If not specified,
+            defaults to "tensorwise" with "enable_fsdp_float8_all_gather=True". See
+            https://github.com/pytorch/ao/blob/v0.9.0/torchao/float8/config.py#L150
+            for more details.
+
+    Returns:
+        (nn.Module) The new model with `Float8Linear`.
+    """
+    if fp8_recipe_name is not None:
+        fp8_config = Float8LinearConfig.from_recipe_name(fp8_recipe_name)
+    else:
+        fp8_config = Float8LinearConfig(enable_fsdp_float8_all_gather=True)
+    return _convert_to_float8_training_torchao(
+        model,
+        config=fp8_config,
+        module_filter_fn=lambda mod, fqn: fqn != "output",
+    )
+
+
+# TODO: validate this in full_finetune_distributed recipe once FP8 + TP is enabled
+def _validate_float8_tp_plan(
+    tp_plan: Optional[Dict[str, ParallelStyle]],
+    fp8_recipe_name: Optional[str] = None,
+) -> None:
+    """
+    Validate that the provided tensor parallel plan is compatible with the
+    float8 settings. Specifically, float8 tensor parallel plans are only
+    supported when using 'tensorwise' float8 recipes.
+    """
+    if tp_plan is None or is_fp8_tensorwise_scaling(fp8_recipe_name):
+        return
+    for parallel_style in tp_plan.values():
+        if isinstance(parallel_style, Float8ColwiseParallel) or isinstance(
+            parallel_style, Float8RowwiseParallel
+        ):
+            raise ValueError(
+                "%s and %s are only compatible with 'tensorwise' float8 recipes"
+                % (Float8ColwiseParallel.__name__, Float8RowwiseParallel.__name__)
+            )
+
+
+def is_fp8_tensorwise_scaling(fp8_recipe_name: Optional[str]):
+    """
+    Return True if the fp8 recipe name refers to 'tensorwwise' scaling.
+    """
+    return fp8_recipe_name is None or fp8_recipe_name == "tensorwise"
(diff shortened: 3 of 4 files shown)


torchtune/training/quantization.py
```diff
@@ -219,3 +223,22 @@ def swap_lora_linear_with_qat(
                 activation_qat_config,
                 weight_qat_config,
             )
+
+
+def convert_to_float8_training(model: torch.nn.Module) -> torch.nn.Module:
+    """
+    Prepare the model for float8 training by swapping all `torch.nn.Linear` with `Float8Linear`.
+    """
+    print("doing fp8 quantized training")
```
andrewor14, line 232: will delete

recipes/full_finetune_distributed.py
```diff
@@ -547,6 +548,10 @@ def _setup_model(
 
         # Apply tensor parallelism to the model
         if self.parallel_dims.tp_enabled:
+            if self._enable_fp8_training:
+                raise ValueError(
+                    "Float8 training + tensor parallelism is not supported yet"
+                )
```
ebsmothers, line 554: Regarding this, are there still known gaps that need to be addressed? E.g. of the three issues from @vkuzo's [comment](https://github.com/pytorch/torchtune/pull/2404#issuecomment-2685709722) it seems like one is closed, one was a red herring, and one is not relevant to us today (as it's related to async TP). Wondering if we're good to just gate on nightly PyTorch instead of disabling entirely?

recipes/full_finetune_distributed.py
```diff
@@ -846,6 +854,12 @@ def train(self) -> None:
                     if self._lr_scheduler is not None:
                         self._lr_scheduler.step()
 
+                    # If float8 training is enabled, perform a single all-reduce to compute the
+                    # scale for all float8 parameters efficiently instead of doing many small
+                    # all-reduces for each parameter
+                    if self._enable_fp8_training and self.dp_degree > 1:
+                        precompute_float8_dynamic_scale_for_fsdp(self._model)
```
ebsmothers, line 861: Probably need to import this

recipes/full_finetune_distributed.py
```diff
@@ -559,6 +564,9 @@ def _setup_model(
                 parallelize_plan=self.tp_plan,
             )
 
+        if self._enable_fp8_training:
+            model = convert_to_float8_training(model)
```
ebsmothers, line 568: Noob clarification q but based on [this comment block](https://github.com/pytorch/ao/blob/c9b1490971ab3b95c20f076e8dd3a679e27883dc/torchao/float8/fsdp_utils.py#L96-L113) in ao it would seem that float8 conversion occurs prior to TP parallelization. Just wanna confirm that this isn't necessary (iiuc from how titan does it the two are basically done concurrently)

ebsmothers (commented):
This is looking good! I'm excited to see some initial runs

torchtune/training/quantization.py
```diff
@@ -219,3 +223,22 @@ def swap_lora_linear_with_qat(
                 activation_qat_config,
                 weight_qat_config,
             )
+
+
+def convert_to_float8_training(model: torch.nn.Module) -> torch.nn.Module:
```
vkuzo, line 228: nit: it's confusing to have `convert_to_float8_training` call `_convert_to_float8_training`, which then calls a different `convert_to_float8_training`.  Can we make the naming cleaner?

torchtune/training/quantization.py
```diff
@@ -219,3 +223,22 @@ def swap_lora_linear_with_qat(
...
+    """
+    Prepare the model for float8 training by swapping all `torch.nn.Linear` with `Float8Linear`.
+    """
+    print("doing fp8 quantized training")
+    fp8_config = Float8LinearConfig(
+        enable_fsdp_float8_all_gather=True,
```
vkuzo, line 234: is FSDP always on in torchtune?

torchtune/training/quantization.py
```diff
@@ -219,3 +223,22 @@ def swap_lora_linear_with_qat(
...
+        force_recompute_fp8_weight_in_bwd=True,
```
vkuzo, line 235: if we are gating on PT nightlies we can delete this, as a change landed to PT core recently to make us not need this config option anymore.  We plan to delete this option in the future.

recipes/full_finetune_distributed.py
vkuzo: the main issue I was worried about earlier turned out to be due to bad hardware, so we are all good!
 
are there plans / needs to support asyncTP in torchtune?

torchtune/training/quantization.py
```diff
@@ -219,3 +223,22 @@ def swap_lora_linear_with_qat(
...
+        cast_config_input=CastConfig(scaling_type=ScalingType.DYNAMIC),
+        cast_config_weight=CastConfig(scaling_type=ScalingType.DYNAMIC),
+        cast_config_grad_output=CastConfig(scaling_type=ScalingType.DYNAMIC),
+    )
```
vkuzo, line 239: it would be good to also enable the rowwise scaling recipes here for increased accuracy! `Float8LinearConfig.from_recipe_name(...)`, https://github.com/pytorch/ao/blob/c9b1490971ab3b95c20f076e8dd3a679e27883dc/torchao/float8/config.py#L114

Note that only the all-tensorwise recipe needs the custom FSDP and TP things.

torchtune/training/quantization.py
andrewor14: I think so, @ebsmothers?

torchtune/models/llama3/_parallelism.py
```diff
@@ -50,4 +59,19 @@ def base_llama_tp_plan() -> Dict[str, ParallelStyle]:
...
+    Return the tensor parallel plan for Llama3 model that uses float8 for all-gather for both
+    rowwise and colwise computation, currently only compatible with float8 fine-tuning with
+    "tensorwise" scaling. This tensor parallel plan is shared between 3.1, 3.2, and 3.3 models.
+
+    Returns:
+        Dict[str, Any]: The float8-enabled tensor parallel plan for Llama3 model.
+    """
+    return _get_base_llama_tp_plan(
+        _colwise_parallel_cls=Float8ColwiseParallel,
+        _rowwise_parallel_cls=Float8RowwiseParallel,
```
nathan-az, line 76: Are these imports missing?

recipes/full_finetune_distributed.py
```diff
@@ -846,6 +855,12 @@ def train(self) -> None:
                     if self._lr_scheduler is not None:
                         self._lr_scheduler.step()
 
+                    # If float8 training is enabled, perform a single all-reduce to compute the
+                    # scale for all float8 parameters efficiently instead of doing many small
+                    # all-reduces for each parameter
+                    if self._enable_fp8_training and self.dp_degree > 1:
+                        precompute_float8_dynamic_scale_for_fsdp(self._model)
```
vkuzo, line 862: I think this should only be enabled for tensorwise scaling, the other recipes do not support float8 all-gather for FSDP right now.

recipes/full_finetune_distributed.py
```diff
@@ -545,6 +549,11 @@ def _setup_model(
         if self._compile:
             training.compile_model(model, verbose=self._is_rank_zero)
 
+        if self._enable_fp8_training:
+            # TODO: gate on nightlies?
```
vkuzo, line 553: p.s. this is the PR in PT core I was talking about: https://github.com/pytorch/pytorch/pull/148922

if the PT core checkout has this PR, we don't need `Float8LinearConfig. force_recompute_fp8_weight_in_bwd` anymore as the min-cut partitioner now does the right thing automatically.

vkuzo:
this is great! Are there any performance benchmarks we can put in the PR summary?

recipes/full_finetune_distributed.py
```diff
@@ -184,6 +186,8 @@ def __init__(self, cfg: DictConfig) -> None:
         self._optimizer_in_bwd = cfg.get("optimizer_in_bwd", False)
         self._clip_grad_norm = cfg.get("clip_grad_norm", None)
         self._checkpoint_client = CheckpointClient(cfg)
+        self._enable_fp8_training = cfg.get("enable_fp8_training", False)
+        self._fp8_recipe_name = cfg.get("fp8_recipe_name", None)
```
andrewor14, line 190: @vkuzo does this level of configuration look good to you? Any other fields you think I should expose?

andrewor14:
> this is great! Are there any performance benchmarks we can put in the PR summary?

Thanks. Yeah these are still in progress, will add shortly.

recipes/full_finetune_distributed.py
vkuzo: I remember you mentioned a performance regression in previous experiments, and we then root caused it to the fact that Float8Linear was applied on parts of the model where torch.compile was not applied to.  I'm not familiar with how torch.compile is applied in torchtune, but it would be good to ensure Float8Linear is not applied to any regions which are not using compile.

recipes/full_finetune_distributed.py
andrewor14: I think the current state in this PR does them in the right order as described in the comment block you linked?

recipes/full_finetune_distributed.py
nathan-az: @vkuzo @andrewor14 From `torchao` source it appears rowwise scaling options do not support `enable_fsdp_float8_all_gather` or `enable_fsdp_float8_all_gather`.

Should we expose `enable_fsdp_float8_all_gather` in the case where a recipe name isn't provided? I imagine this is faster, especially for larger models.

In addition it looks like `force_recompute_fp8_weight_in_bwd` is always recommended in the FSDP case. Should we pair these two options, again, in the case where recipe name isn't provided?

torchtune/models/llama3/_parallelism.py
```diff
@@ -15,32 +15,53 @@
...
+    """
+    return {
+        "tok_embeddings": _rowwise_parallel_cls(
+            input_layouts=Replicate(), output_layouts=Shard(1)
+        ),
+        "norm": _sequence_parallel_cls(),
+        "output": _colwise_parallel_cls(
+            input_layouts=Shard(1), output_layouts=Replicate()
+        ),
+        "layers.*.attn": PrepareModuleInput(
```
nathan-az, line 40: @vkuzo any guidance on using `PrepareFloat8ModuleInput` here? Looks like [this is now used in `torchtitan`](https://github.com/pytorch/torchtitan/blob/a59ed30a08fb293015f0838b3dd8152e4a273f5f/torchtitan/models/llama/parallelize_llama.py#L178). 

The source mentions that this assumes tensorwise scaling, so not sure whether it plays nicely with the rowwise float8 recipes (or if that makes any difference).

recipes/full_finetune_distributed.py
andrewor14: I think @vkuzo mentioned that `force_recompute_fp8_weight_in_bwd` is no longer needed after a certain commit. This PR gates on the nightlies so we don't need this flag anymore

nathan-az:
Great work on this! 

> no change in memory usage

> However, memory usage did increase minimally (+2%) for most fp8 settings

I'm a bit surprised by the above - with all linear layers quantized, should we not see almost strictly decreasing memory usage (and quite substantially)? My understanding is that the weights and grads are both quantized in memory with these classes (in addition to their scaling factors) so I expected some throughput benefit as well as memory benefit to ~halving (noting that embedding layers aren't converted) the parameter sizes.

recipes/full_finetune_distributed.py
ebsmothers: > I'm not familiar with how torch.compile is applied in torchtune, but it would be good to ensure Float8Linear is not applied to any regions which are not using compile.

Regarding @vkuzo's comment, torchtune does per-layer compile with separate compilation of the loss. You can see the utility we use [here](https://github.com/pytorch/torchtune/blob/2ca4d522b8eb0f1212d42ac25aef5e70d8a8954f/torchtune/training/_compile.py#L30). This means that there may be two cases that are not covered:

1) output projection (i.e. LM head)
2) any linear layers in e.g. a vision encoder for a multimodal model

Given the numbers in the test plan it seems (1) is not a hard blocker. We could try out with e.g. Llama 3.2 Vision just to make sure there are no regressions there. Otherwise I don't have any huge concerns here

torchtune/models/llama3/_parallelism.py
```diff
@@ -15,32 +15,53 @@
...
+        "layers.*.attn.v_proj": _colwise_parallel_cls(),
+        "layers.*.attn.output_proj": _rowwise_parallel_cls(output_layouts=Shard(1)),
+        "layers.*.mlp.w1": _colwise_parallel_cls(),
+        "layers.*.mlp.w2": _rowwise_parallel_cls(output_layouts=Shard(1)),
+        "layers.*.mlp.w3": _colwise_parallel_cls(),
+    }
+
+
+_BASE_LLAMA_TP_PLAN = _get_base_llama_tp_plan()
+_FP8_LLAMA_TP_PLAN = _get_base_llama_tp_plan(
```
ebsmothers, line 61: I am doing something similar in #2570 (defining separate plans for inference vs training as opposed to float8 though). Mainly bringing it up as an fyi for potential merge conflicts. 

In that PR I define TP plan as a function taking in an nn.Module to get around some awkwardness with the parallelize_module API and Llama4 (even though it's not strictly needed for Llama3). I think we still need to figure out the right way to expose these TP plans, but no need to block this PR on that.

torchtune/training/quantization.py
```diff
@@ -219,3 +227,60 @@ def swap_lora_linear_with_qat(
...
+    model: nn.Module,
+    fp8_recipe_name: Optional[str] = None,
+) -> nn.Module:
+    """
+    Prepare the model for float8 training by swapping all `nn.Linear` with `Float8Linear`.
+
+    Args:
+        model (nn.Module): The model to swap linear layers on
+        fp8_recipe_name (Optional[str]): name to identify one of the pre-made recipes,
+            one of "tensorwise", "rowwise", and "rowwise_with_gw_hp". If not specified,
```
ebsmothers, line 242: Noob float8 q: seems like tensorwise gives the best perf in your runs, is that typically the case?  I assume the rowwise scaling will have less degradation, though doesn't seem to show up in the eval numbers at all. (Mainly want to gain some understanding of what a sensible default is for folks to use.)

torchtune/training/quantization.py
```diff
@@ -219,3 +227,60 @@ def swap_lora_linear_with_qat(
...
+            https://github.com/pytorch/ao/blob/v0.9.0/torchao/float8/config.py#L150
+            for more details.
+
+    Returns:
+        (nn.Module) The new model with `Float8Linear`.
+    """
+    if fp8_recipe_name is not None:
+        fp8_config = Float8LinearConfig.from_recipe_name(fp8_recipe_name)
+    else:
+        fp8_config = Float8LinearConfig(enable_fsdp_float8_all_gather=True)
```
ebsmothers, line 253: More of a nit cause I know you already mentioned it in the docstring, but it'd be nice if this config build explicitly referenced tensorwise (otherwise one has to trace through the ao code to understand which type of scaling is implied by the use of `enable_fsdp_float8_all_gather=True`)

torchtune/training/quantization.py
ebsmothers: Missed this before, but it is possible to have pure TP in torchtune. In that case FSDP would not be enabled

recipes/full_finetune_distributed.py
ebsmothers: @vkuzo asyncTP is not something that has really been on our radar. But given that it shows some nice speedups, I think it would be worth exploring.

ebsmothers (commented):
This is looking good! Left a few comments but no major concerns from me. We may want to test on a multimodal model (probably Llama 3.2 would be easiest) just to make sure there are no challenges around the compile point mentioned by @vkuzo. Otherwise I think this should be good to land.

torchtune/models/llama3/_parallelism.py
```diff
@@ -15,32 +15,53 @@
...
+def _get_base_llama_tp_plan(
+    _sequence_parallel_cls: Type[ParallelStyle] = SequenceParallel,
+    _colwise_parallel_cls: Type[ParallelStyle] = ColwiseParallel,
+    _rowwise_parallel_cls: Type[ParallelStyle] = RowwiseParallel,
+) -> Dict[str, ParallelStyle]:
+    """
+    Define the Tensor Parallel plan for Llama3 model, which will also be shared with 3.1, 3.2, and 3.3 models.
+    """
+    return {
+        "tok_embeddings": _rowwise_parallel_cls(
```
nathan-az, line 33: Does this work for embedding layers? `torchao` [source](https://github.com/pytorch/ao/blob/main/torchao/float8/float8_tensor_parallel.py#L158) looks to me like it would raise an error for non-Float8Linear objects (but I may misunderstand how this works)

torchtune/training/quantization.py
andrewor14: My understanding is rowwise is typically a bit slower but more accurate. I think we don't see the "more accurate" part in the experiments because for this workload tensorwise is already preserving accuracies very well, but it's possible we'll need rowwise for other workloads still

torchtune/models/llama3/_parallelism.py
andrewor14: Oh good point, let me double check

andrewor14: Oh good point, let me double check. When I ran the experiments at least there was no error even when using this TP plan

nathan-az:
I pulled this down to do some testing, but had trouble with tensor parallelism even after tweaking the plan.

> (full script: https://github.com/andrewor14/torchtune/blob/fp8-finetuning-debug/run_it.sh)

This script appears to set  `tensor_parallel_plan` but not `tensor_parallel_dim` - by default this be `1` so only FSDP is used. May need to do some testing setting the `tensor_parallel_dim` to confirm TP is working as expected. This may also be why you aren't seeing the same errors.

andrewor14:
@nathan-az I'm running into this error now, even with the original base TP plan (no float8 parallel styles). Do you have more context on this / how you worked around this in the past?

```
[rank0]:   File "/home/andrewor/local/pytorch/torch/distributed/_functional_collectives.py", line 215, in all_gather_tensor
[rank0]:     res = torch.cat(torch.chunk(res, group_size, dim=0), dim=gather_dim)
[rank0]:   File "/home/andrewor/local/ao/torchao/float8/float8_tensor.py", line 375, in __torch_dispatch__
[rank0]:     raise NotImplementedError(f"attempting to run {func}, this is not supported")
[rank0]: NotImplementedError: attempting to run aten.chunk.default, this is not supported
```
Full stack trace: https://gist.github.com/andrewor14/b4058adc32fcb1dafd70574e0eae335d

nathan-az:
> how you worked around this in the past

Perfect - that's the exact error I got stuck on after changing the TP plan. My best guess (absolute guess) if that error can be taken at face value is that the chunk op needs to be implemented in torchao. [See here](https://github.com/pytorch/ao/blob/a81322e4f58b828c5ddf42623fce809d170e4caa/torchao/float8/float8_ops.py#L101). CC @vkuzo does that seem likely?

andrewor14:
> Perfect - that's the exact error I got stuck on after changing the TP plan. My best guess (absolute guess) if that error can be taken at face value is that the chunk op needs to be implemented in torchao. [See here](https://github.com/pytorch/ao/blob/a81322e4f58b828c5ddf42623fce809d170e4caa/torchao/float8/float8_ops.py#L101)

I can implement that op. I'm just curious how we did it in torchtitan since we clearly have TP support there. Do you see any difference in the torchtune vs torchtitan setup that could be causing this?

nathan-az:
I expect there's something but it's not obvious to me. Unfortunately this is about where I hit a wall. I was able to change my error message by adding the chunk op to `ao`'s `split` (not sure if this is exactly correct or if there's any nuanced difference), but hit another error that I didn't know how to address.

One final note I can add is that I got a different error after disabling `compile` about `contiguous` not being implemented, but not sure if this is a red herring.

Just a thought - is it worth worth landing this with standard FP8 support for FSDP, raising an error that FP8 + TP isn't supported yet, then adding TP support into a follow-up?

andrewor14:
> Just a thought - is it worth worth landing this with standard FP8 support for FSDP, raising an error that FP8 + TP isn't supported yet, then adding TP support into a follow-up?

Yeah I think that's a good idea. @ebsmothers does this sound good to you?

ebsmothers:
> > Just a thought - is it worth worth landing this with standard FP8 support for FSDP, raising an error that FP8 + TP isn't supported yet, then adding TP support into a follow-up?
> 
> Yeah I think that's a good idea. @ebsmothers does this sound good to you?

@andrewor14 yeah this is fine with me. Let's file an issue to get things working with TP though

nathan-az:
Last note from me until we move to a separate issue for TP + FP8. By implementing `chunk` in torchao for the Float8Tensor (piggybacking `float8_split`), I had successful runs (no guarantee of correctness, just no errors), but only when `compile=false`. This did _not_ work for models with tied weights for the `output` layer (although this is mostly the case for smaller models for which it's less likely that TP + FP8 would be critical).

Compile created some metadata errors mentioning implementing `__force_to_same_metadata__`, but implementing this on the `Float8Tensor` class did not change the error for me. 

I created a [gist](https://gist.github.com/nathan-az/f23bfa49e24849a014b76badda813a0a) with a slightly more minimal and self-contained test script. It has very little dependency on `torchtune` imports so is hopefully everything that's happening is transparent, and it's useful for debugging. Right now it instantiates the full 8B, but could obviously be a custom/smaller model (even the same hyperparams but with only one layer).

andrewor14:
@nathan-az Great, thank you for all your investigation! I think we do want TP to work with compile so we probably need to investigate further. Will run some additional benchmarks for Llama 3.2 and let's aim to land this PR without TP.

torchtune/training/quantization.py
andrewor14: Tried to fix this but looks like `Float8LinearConfigs` are not mutable, so there's no easy way to associate this config with the name "tensorwise"

ebsmothers (approved):
Thanks for adding this! This looks good to me. Gonna re-run CI (I think there is some unrelated sporadic failure), after that let's land this!

PR #2362, opened 2025-02-07
[Fix Test] Fix failed generation test by pining pytorch nightlies 

acisseJZhong:
#### Context
What is the purpose of this PR? Is it to
- [ ] add a new feature
- [x] fix a bug
- [ ] update tests and/or documentation
- [ ] other (please add here)

Please link to any issues this PR addresses.

Some of the generation tests are failing with latest pytorch nightlies. See [link](https://github.com/pytorch/torchtune/actions/runs/13190328912/job/36821902764). We(thanks to @ebsmothers ) triaged the issue down to pytorch nightlies [2/2](https://github.com/pytorch/pytorch/commit/738ebb45cad7cef7fa0878ce4f835be59d186a94#diff-6b8e21a44da970b14c1d62eb1747af1b09d3914ff1f6354754c62f1f5784cb9d). To fix CI for now, we pinged pytorch nightlies to the 2/1 version. 

Need to look into what exactly in 2/2 causes the failure. 

#### Changelog
What are the changes made in this PR?
* 

#### Test plan
Please make sure to do each of the following if applicable to your PR. If you're unsure about any one of these just ask and we will happily help. We also have a [contributing page](https://github.com/pytorch/torchtune/blob/main/CONTRIBUTING.md) for some guidance on contributing.

CI for nightly green: 
image

- [ ] run pre-commit hooks and linters (make sure you've first installed via `pre-commit install`)
- [ ] add [unit tests](https://github.com/pytorch/torchtune/tree/main/tests/torchtune) for any new functionality
- [ ] update [docstrings](https://github.com/pytorch/torchtune/tree/main/docs/source) for any new or updated methods or classes
- [ ] run unit tests via `pytest tests`
- [ ] run recipe tests via `pytest tests -m integration_test`
- [ ] manually run any new or modified recipes with sufficient proof of correctness
- [ ] include relevant commands and any other artifacts in this summary (pastes of loss curves, eval results, etc.)

#### UX
If your function changed a public API, please add a dummy example of what the user experience will look like when calling it.
Here is a [docstring example](https://github.com/pytorch/torchtune/blob/6a7951f1cdd0b56a9746ef5935106989415f50e3/torchtune/modules/vision_transformer.py#L285)
and a [tutorial example](https://pytorch.org/torchtune/main/tutorials/qat_finetune.html#applying-qat-to-llama3-models)

- [ ] I did not change any public API
- [ ] I have added an example to docs or docstrings

PR #2549, opened 2025-04-02
Replace exponential with torch.log in generation sampling

felipemello1:
#### Context
- [ ] add a new feature
- [x] fix a bug
- [ ] other (please add here)

[Starting in torch 2.7.0.dev20250202, our tests began to fail.](https://github.com/pytorch/torchtune/pull/2362). After investigating, we found that [q = torch.empty_like(probs).exponential_(1) ](https://github.com/pytorch/torchtune/blob/3a2179cd22049ad873a81d6d5a1e7e3b5f8b8c80/torchtune/generation/_generation.py#L62) was the culprit.

```
import torch

torch.manual_seed(42)
print("torch.rand(5):", torch.rand(5))
print("torch.empty(5).exponential_(1):", torch.empty(5).exponential_(1))
```

In nightlies 2.8.0.dev20250401+cu126:
NOTE (pinning to 2.7.0.dev20250201 still works, so [dev20250202](https://github.com/pytorch/pytorch/commit/738ebb45cad7cef7fa0878ce4f835be59d186a94#diff-ecbd0bdf6ee8fc4eab27af1ec0b4369962dd340547451ac2f396b8f12f2f41bc) seems to be the cause):
```
torch.rand(5): tensor([0.8823, 0.9150, 0.3829, 0.9593, 0.3904])
torch.empty(5).exponential_(1): tensor([0.5809, 0.0436, 2.0548, 1.9680, 0.3315])
```

In stable 2.6.0:
```
torch.rand(5): tensor([0.8823, 0.9150, 0.3829, 0.9593, 0.3904])
torch.empty(5).exponential_(1): tensor([1.6459, 0.4294, 0.0677, 1.3809, 0.7803])
```

### Fix

To mitigate that, we can replace the function with rand_like+log to sample from an:
```
probs = torch.ones(5)
u = torch.rand_like(probs)  # Uniform random samples
q = -torch.log(1 - u)  # Transform to exponential distribution
```

this is [what exponential does](https://github.com/pytorch/pytorch/blob/e62d958f02afd74e4d4c023fa97d6c7966e00e26/torch/_refs/__init__.py#L6132), but for some reason random seed it not working on it anymore on nightlies.

### Test

1) Update stop tokens
2) Updated random gibberish generated
3) Remove FSDPTest, which was breaking in nigthlies


joecummings (approved):
kewl

PR #2571, opened 2025-04-07
Revert back to ``safe_open`` and don't use buffer

joecummings:
### Testing

| on main | pure revert to before | revert + remove buffer |
|--------|--------|--------|
| 12.84s | 0.0810s | 0.0020s |


ebsmothers (approved):
Great work

Issue #1716, opened 2024-09-30
mistral-nemo support


win4r:
https://youtu.be/Z3B-s1ix_RA

casper-hansen:
@joecummings Hi it's been a long time since this model was introduced. I find the 12-14B parameter range to be good and hope it could be supported in TorchTune.

PR #2588, opened 2025-04-12
[MODEL] Mistral Nemo 12B Support

casper-hansen:
#### Context
- [x] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

Issue: #1716

#### Changelog

* Added Mistral's "v3" tokenizer called Tekken (tiktoken based)
* Added support for specifying `head_dim` (default=None for backward compatibility) in the MistralModel for loading the model weights with correct shapes

#### Test plan

#### UX

### DIFF
--- a/recipes/configs/mistral/12B_full_low_memory.yaml
+++ b/recipes/configs/mistral/12B_full_low_memory.yaml
@@ -1,0 +1,117 @@
+# Config for single device full finetuning in full_finetune_single_device.py
+# using a Mistral 12B model
+#
+# This config uses hyperparameters based on small set of experiments and information
+# available on various forums. These are not meant to replicate the numbers
+# from the paper
+#
+# This config assumes that you've run the following command before launching
+# this run:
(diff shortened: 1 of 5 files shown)


casper-hansen:
Before I go onto add tests, align the config to other configs, implement LoRA etc., I would like to know if this PR is likely to be accepted to support the Mistral Nemo model if those things are implemented. CC @joecummings

pbontrager:
Thank you for this contribution! Could you add some context on how you verified this model and link to any gist where you compared the model outputs to HF? Ideally we could see that this tokenizer matches the official implementation exactly for some dummy inputs and that generation with the model weights (using the generate.py recipe) is coherent and that finetuning with the model seems to work.

casper-hansen:
@pbontrager I appreciate your review. I will not have time to go through more testing. You are free to take this code, modify it, etc. I tested the training and compared it to Mistral 7B with the same config and the loss was within the same ranges and seemed normal.

Issue #2598, opened 2025-04-15
Qwen Student and teacher logits mismatch during KD, needs padding


nouranali:
In this exact line, in kd_loss.py
![Image](https://github.com/user-attachments/assets/4255c8ac-7aca-41a1-b31e-5d9a044ba8d6)

ebsmothers:
Hi @nouranali thanks for creating the issue! I left a comment on #2599, happy to continue the discussion there.

PR #2599, opened 2025-04-15
Solves #2598

nouranali:


ebsmothers (commented):
Thanks for the PR! Before giving a more detailed review, one high-level comment: I wonder if we really want to handle this for people in general. E.g. with `pad_logits` it is possible for me to run KD on two totally different tokenizers (and vocab spaces) without seeing any errors, right? This goes beyond just the case of "vocab of teacher model is subset of vocab of student model" (or vice versa).

Anyways I agree that mismatched vocab is an challenge (especially for Qwen models) -- it's come up in a couple issues on the repo before. But I also want to make sure we don't do something behind the scenes that allows someone to accidentally shoot themselves in the foot. Would be interested to hear your thoughts on this though.

PR #2601, opened 2025-04-15
Revert "Padding bug in GRPO"

felipemello1:
Many parts of the code hardcode context_length truncation. Padding breaks this assumption, since context length is different for bsz>1. Other parts of the code need to be fixed as well.

### DIFF
--- a/recipes/dev/grpo_full_finetune_distributed.py
+++ b/recipes/dev/grpo_full_finetune_distributed.py
@@ -819,60 +819,6 @@ class FullGRPOFinetuneRecipeDistributed(FTRecipeInterface):
             seq_lens=seq_lens,
         )
 
-    def _pad_tensor(
-        self,
-        tensor: torch.Tensor,
-        target_dim: int,
-        pad_value: float,
-        dim: int = 1,
-        pad_right: bool = True,
-    ) -> torch.Tensor:
-        """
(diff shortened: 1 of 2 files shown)


ebsmothers (approved):

krammnic:
I don't really see the connection, the padding was introduced to fix the rare errors on the concatination of the trajectories. `context_length` truncation seems to be a separate problem. Why do we need to fix some hardcoded context length in case if we pad? Why it can't be done in a follow up?

felipemello1:
when bsz=1, all of the prompts are the same, and therefore, you can do the masking as the algorithm currently does:
```
query_response_logits[:, context_length - 1 : -1]
```

where 
query_reseponse = Prompt + Answer
context_length = len(prompt) (calculated as context_length = query_responses.shape[1] - responses.shape[1])

When bsz>1, len(prompt1) != len(prompt2)

example:

"P1 P2 P3 P4 | A1 A2 A3 AA"
"P1 P2 | A1 A2 A3 A4 A5 PAD"

after concat + padding, you can see that the truncation doesnt work anymore, because `context_length` is different for both. Maybe masking will take care of it, but at this point i dont know, and this needs more investigation. There may be other parts that also require refactoring. So looking at rewards going up is not enough, unfortunately.

I am fixing it in the other repository, but for this one, I would rather keep the constraint of bsz=1 for now until we are sure about correctness.

@krammnic

PR #2603, opened 2025-04-16
Fix Llama tokenizer failing to decode some tokens, but this time correctly

RedTachyon:
#### Context
- [ ] add a new feature
- [x] fix a bug
- [ ] other (please add here)

See #2424
I tried fixing it in #2428 but failed miserably, so trying to make it right. This time actually testing that it works.

#### Changelog
*Llama 3 tokenizer will no longer crash when trying to decode token 128011

#### Test plan

I added a test that tries decoding every token individually, and then all of them together. I don't want to mess with the test suite too much, so I'm restricting it to the "small" tokenizer included in the assets, i.e. 2000 normal tokens and then all the special tokens.


RedTachyon:
Note: in the meantime I actually found a similar bug in the Qwen tokenizer. It's not as clear to me what is happening, but also a "clean" fix is not as obvious since different Qwens have different tokenizer sizes.

The short story is that I was running RL on Qwen, and got a familiar error with the tokenizer freaking out because of an unknown token. 

The fix I'm using (and proposing here) is just skipping decoding tokens that don't have a defined meaning, but that fit inside the model's vocab size. For example, Qwen 2.5 32B has 152064 possible tokens, but the registered special tokens end at 151664. Anything above that gets sampled? Crash. 

Skipping this in decoding might have some odd interactions in that the model can output a bunch of tokens that don't do anything in string space, but I don't expect that to actually meaningfully impact anything (famous last words). A simple alternative would be adding something like <|unk|>. A more high-effort alternative would be checking how original Qwen decodes these (probably largely unused?) tokens, or giving them unique names ourselves.

Let me know if this should be included in this PR as a general tokenizer cleanup, or maybe factored out (or done differently)

ebsmothers (commented):
@RedTachyon thanks for doing this! I stole your fix (for Llama3 anyways) and put it into #2595 because I wanted to finally get our CI green again. Do you mind merging main into your branch and letting CI run? Thanks for adding these tests btw, this is just the kind of stuff we need to avoid these tricky bugs in the future

joecummings (approved):
This is good to go, right? @ebsmothers

ebsmothers (approved):
Thanks for the fix and the tests!

PR #2604, opened 2025-04-16
fix tag extraction

felipemello1:
#### Context
- [ ] add a new feature
- [x] fix a bug
- [ ] other (please add here)

### DIFF
--- a/torchtune/dev/grpo/rewards.py
+++ b/torchtune/dev/grpo/rewards.py
@@ -4,28 +4,31 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
-from xml.etree import ElementTree as ET
+import re
+from typing import Dict
 
 import torch
 
 from torchtune.modules.transforms.tokenizers import ModelTokenizer
 
 
-def extract_tags(text: str) -> dict[str, list[str]]:
+def extract_tags(text: str) -> Dict:
     """
     Parse XML-like tags from text. Returns a dictionary with keys 'think' and 'answer'.
     The values are lists of strings, with each string being the content of a tag.
     """
-    xml_string = f"<root>{text}</root>"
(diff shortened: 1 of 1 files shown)


RedTachyon:
One thing - while I acknowledge the clunkiness of that XML parsing, some of it was intentional. If the response is well-formatted, i.e. <think>cot</think><answer>42</answer><|eot|>, then it will be the same (or arguably better).

However, if we're training a small/dumb model, it will sometimes output almost-correctly-formatted responses like
<think>cot1</think><think>cot2</think><answer>1</answer><answer>2</answer>

Sometimes it's not even complete insanity, it can be like

<think>partial cot</think><answer>partial answer</answer><think>more cot</think><answer>final answer</answer>

This approach seems to match only the first think and answer tags, discarding the previous ones. This might or might not make some difference for reward shaping, but we should at least be aware of that.

---

Side note - I have a version that uses math-verify for slightly more flexible verification, think allowing `\boxed{42}` if the gold answer is `42` (and much more) - happy to upstream it here or to the async version (or both)

felipemello1:
hey @RedTachyon , thanks for commenting on the PR! 

>while I acknowledge the clunkiness of that XML parsing

The motivation for changing it is because i logged answers, saw that they were correct, but reward was 0. After i replaced them with regex, things worked fine. So I imagine that there is some issue with the current xml code, but i didnt investigate it further.

>This approach seems to match only the first think and answer tags, discarding the previous ones.

Regarding missing out on multiple <think< tags, i thought about it, but if the intended behavior is to have a single <think> and a single <answer> then i think that its fine for the reward to be smaller on the cases that dont follow thing pattern.

>  I have a version that uses math-verify for slightly more flexible verification [...] happy to upstream it here or to the async version (or both)

That would be great! I think its fine to put it only in the async version

PR #2607, opened 2025-04-17
Change README badge to follow gpu_test instead of recipe_test

joecummings:
recipe_test.yml no longer exists as of #2595


ebsmothers (approved):
good call

PR #2610, opened 2025-04-18
Fix: Ensure ViT includes the final hidden state in its output

tkdcjf159:
#### Context
-  fix a bug

When using the `out_indices` parameter to select specific hidden states from the Vision Transformer (ViT) based on their layer index, the implementation incorrectly fails to return the hidden state from the final transformer layer. This occurs even if the index corresponding to the final layer is explicitly included in the `out_indices` list.

#### Changelog
*

#### Test plan

#### UX


torchtune/modules/vision_transformer.py
```diff
@@ -376,11 +376,16 @@ def forward(
 
         # transformer with optional hidden layer outputs
         x = x.reshape(bsz_and_n_imgs, n_tiles * n_tokens, embed_dim)
-        for layer_idx, transformer_layer in enumerate(self.layers):
+        layer_idx = 0
+        for transformer_layer in self.layers:
```
pbontrager, line 380: You can still keep the enumerate, the layer_idx variable is still accessible after the for loop

pbontrager (approved):
Thanks for adding this. It was assumed that you'd get the last hidden layer from the x output, but it makes sense that you might want it before the layer norm and positional embeddings. Please put the enumerate back to keep things a bit more compact but otherwise it's good to go.

torchtune/modules/vision_transformer.py
tkdcjf159: @pbontrager Fix it! 
Initially, I misunderstood that the range of out_indices was 1 ~ num_layer instead of 0 ~ num_layer-1, so I made modifications based on that misconception. If the range of out_indices is 0~num_layer-1, then after the for loop, I should just check if layer_idx is included in self.out_indices and add it if it is. I've made the corrections according to your advice.

pbontrager:
You need to fix the test and this should be good to land.

PR #2613, opened 2025-04-18
[Fix] Incorrect Labels Shift for Packed Data

pbontrager:
#### Context
- [ ] add a new feature
- [x] fix a bug
- [ ] other (please add here)

@felipemello1 discovered that how we compute our shifted labels in the [recipe](https://github.com/pytorch/torchtune/blob/1075d9cddfd28d5f1d4ab4e154af6fa80d537fa5/recipes/full_finetune_single_device.py#L642) breaks when we do packed training. Instead of shifting the whole sequence over one and then appending an ignore token at the end of the whole packed sequence, we need to do it on the individual sequences before packing.

This would only affect a single token at each boundary, so it's unclear how much it affected performance in practice. 

#### Changelog
* Updated 10 sft based recipes `git grep hstack -> 10 recipes (full, lora, qat, kd, dev)`
* For each recipe removed ignore_index_cache and label shift logic
* Updated SFTDataset and TextCompletionDataset to handle shifting the labels
* Updated 7 dataset unit tests to have correct exepected_labels

#### Test plan

To compare loss I ran `tune run full_finetune_single_device --config llama3_2/1B_full_single_device tokenizer.max_seq_len=8092 dataset.packed=True compile=True optimizer=torch.optim.AdamW metric_logger=torchtune.training.metric_logging.WandBLogger metric_logger.project="Fix Packed"` on main and on this branch.

Screenshot 2025-04-18 at 4 58 33 PM

The bug doesn't seem to have made a noticeable impact on Alpaca loss.

### DIFF
--- a/recipes/dev/early_exit_finetune_distributed.py
+++ b/recipes/dev/early_exit_finetune_distributed.py
@@ -392,11 +392,6 @@ class EarlyExitFinetuneRecipeDistributed(FTRecipeInterface):
         # if cfg is missing profiler key or if `cfg.profiler.enabled = False`
         self._profiler = self._setup_profiler(cfg.get(PROFILER_KEY, None))
 
-        # Used to ignore labels for loss computation
-        self.ignore_labels_cache = torch.full(
-            (cfg.batch_size, 1), self._loss_fn.ignore_index, device=self._device
-        )
-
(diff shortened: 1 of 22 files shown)


joecummings (approved):
unblockdd

PR #2619, opened 2025-04-21
Add `padded_collate_packed` to api ref

SalmanMohammadi:
https://pytorch.org/torchtune/main/api_ref_data.html#collaters

is no der


joecummings (approved):
👀

PR #2622, opened 2025-04-22
Fix tok per sec

nathan-az:
#### Context
- [ ] add a new feature
- [x] fix a bug
- [ ] other (please add here)

#### Changelog
* Adds an attribute to ParallelDims to count non-dp workers
* Adjusts tokens per second calculation to adjust for over-counting of non-dp workers

#### Test plan

#### UX

### DIFF
--- a/recipes/full_finetune_distributed.py
+++ b/recipes/full_finetune_distributed.py
@@ -965,8 +965,9 @@ class FullFinetuneRecipeDistributed(FTRecipeInterface):
                                     else self._optim_ckpt_wrapper
                                 ),
                             ),
-                            "tokens_per_second_per_gpu": num_tokens
-                            / (time_per_step * self.world_size),
+                            "tokens_per_second_per_gpu": (
+                                num_tokens / self.parallel_dims.non_data_parallel_size
+                            ) / (time_per_step * self.world_size),
                         }
                         if self._log_peak_memory_stats:
                             log_dict.update(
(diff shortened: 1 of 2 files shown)


recipes/full_finetune_distributed.py
```diff
@@ -965,7 +965,9 @@ def train(self) -> None:
                                     else self._optim_ckpt_wrapper
                                 ),
                             ),
-                            "tokens_per_second_per_gpu": num_tokens
+                            "tokens_per_second_per_gpu": (
+                                num_tokens / self.parallel_dims.non_data_parallel_size
+                            )
                             / (time_per_step * self.world_size),
```
joecummings, line 971: Couldn't we just divide by self.dp_size here instead ?

nathan-az: Hmm I'm not following. An example to explain my motivation (and elucidate any misunderstanding on my behalf).

Suppose we have batch size 1, with 1000 tokens in the sample, and TP dim=4.

Due to the all-reduce, this will report as 4000 tokens. But functionally, each device processed only 250 tokens.

* `num_tokens / world_size=1000` over-reports due to the TP workers
* `num_tokens / dp_size` would report 4000 tokens (since dp size is 1) - over-reporting *twice* by the TP workers.

The motivation behind my adjustment is:
* In the denominator, dividing by world size gets us to the actual number of GPUs
* In the numerator, the adjustment corrects the number of tokens actually processed during the update step

joecummings: Okay, I see - thank you for your very clear thinking!

joecummings (approved):
👍

PR #2623, opened 2025-04-22
compile optimizer

IvanKobzarev:
* #2644
* #2624
* __->__ #2623

Compiling optimizer helps perf of Llama4 Scout Model
3.8 tokens_per_second -> 9 tokens_per_second (max value of tokens per second in the first ~10 iterations)
peak memory is the same

```
tune run --nproc_per_node 8 \
  full_finetune_distributed \
```

PS:
Current repo compilation fails if to set `skip_rope_interval=4,`, have to test with `skip_rope_interval=None,`

### DIFF
--- a/recipes/configs/llama4/scout_17B_16E_full.yaml
+++ b/recipes/configs/llama4/scout_17B_16E_full.yaml
@@ -71,6 +71,11 @@ enable_activation_offloading: False
 fsdp_cpu_offload: True
 compile: False # torch.compile, set to true for perf/memory improvement
 
+compile_components:
+  model: True
+  loss: True
+  optimizer_step: False
+
 # Reduced precision
 dtype: bf16
 
--- a/recipes/configs/llama4/scout_17B_16E_full.yaml
+++ b/recipes/configs/llama4/scout_17B_16E_full.yaml
@@ -69,12 +69,14 @@ device: cuda
 enable_activation_checkpointing: True
 enable_activation_offloading: False
 fsdp_cpu_offload: True
-compile: False # torch.compile, set to true for perf/memory improvement
-
-compile_components:
-  model: True
-  loss: True
-  optimizer_step: False
+# compile True means use torch.compile for all components
+# compile False means no torch.compile
+# compile Dictionary with keys: "model", "loss", "optimizer_step"
+# enables torch.compile only for specified components.
+compile: False
+#    model: True
+#    loss: True
+#    optimizer_step: False
 
 # Reduced precision
 dtype: bf16
--- a/recipes/full_finetune_distributed.py
+++ b/recipes/full_finetune_distributed.py
@@ -313,7 +313,7 @@ class FullFinetuneRecipeDistributed(FTRecipeInterface):
         self._compile_model = compile_bool
         self._compile_loss = compile_bool
         self._compile_optimizer_step = compile_bool
-        if isinstance(compile, dict):
+        if isinstance(compile, DictConfig):
             self._compile_model = compile.get("model", True)
             self._compile_loss = compile.get("loss", True)
             self._compile_optimizer_step = compile.get("optimizer_step", False)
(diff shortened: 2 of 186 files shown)


torchtune/training/_compile.py
```diff
@@ -86,3 +86,10 @@ def compile_loss(loss: nn.Module, verbose: bool = True) -> nn.Module:
     else:
         loss = torch.compile(loss, backend=backend)
     return loss
+
+
+def compile_optimizer_step(optimizer_step_fn, verbose: bool = True):
```
joecummings, line 91: I appreciate you wanting to keep this similar to how we were currently doing things; however, we only needed to this for the loss function b/c we were doing funky things with chunking. 

We should just compile this directly in the recipe. Same goes for the other PR you have up.

recipes/full_finetune_distributed.py
```diff
@@ -923,7 +923,16 @@ def train(self) -> None:
                             # If sharded, collect the DTensor here
                             if isinstance(grad_norm, DTensor):
                                 grad_norm = grad_norm.full_tensor()
-                        self._optimizer.step()
+                        optimizer_step_fn = self._optimizer.step
```
joecummings, line 926: See comment below, we can just compile the optimizer step in the recipe directly.

joecummings (approved):
Since we're now compiling several things independently, it might make sense logically to have a section of the recipe where we compile everything after instantiation.

joecummings:
Ah sorry, did not mean to approve :)

recipes/full_finetune_distributed.py
```diff
@@ -923,7 +925,13 @@ def train(self) -> None:
...
+                        if self._compile:
+                            optimizer_step_fn = torch.compile(
+                                optimizer_step_fn,
+                                backend=self._compile_backend,
+                            )
```
felipemello1, line 933: some optimizers might not work with this, if i remember it correctly, like torchao/bnb. May need some testing. The safest option might be to add a compile flag per area, e.g.:

```
compile:
     loss: True
     model: True
     optimizer_step: False
     ```

Comment (winglian) [reply recipes/full_finetune_distributed.py:933]: Should this get moved outside the epoch & dataloader loop? I assume this only needs to happen once?

File: recipes/full_finetune_distributed.py (hunk 2 above)
Comment (SalmanMohammadi) [reply recipes/full_finetune_distributed.py:928]: When I tried this out I found issues with setting up the LR scheduler which fails when attempting to wrap the optimizer step fn

**Comment by IvanKobzarev** [comment]
Tried direct compilation
```
+        if self._compile_optimizer_step:
+            self._optimizer.step = torch.compile(
+                self._optimizer.step, backend=self._compile_backend
+            )
```
And then just call self._optimizer.step() - this did not work for me, compilation of optimizer was not applied :( 
Smth that I do not understand too. 

Updating PR with extracting separate configs and setting compiled optimizer False by default.
Left compiling the same in separate function.

File: recipes/configs/llama4/scout_17B_16E_full.yaml [hunk 4]
```diff
@@ -71,6 +71,11 @@ enable_activation_offloading: False
 fsdp_cpu_offload: True
 compile: False # torch.compile, set to true for perf/memory improvement
 
+compile_components:
```
joecummings, line 74: nit: could we match the argument to just "compile"? Then valid arguments would be "True", "False", or the specific components. If "True", then we compile everything. If "False", we compile nothing. If the argument has a dictionary with each component, then we follow those instructions.

joecummings (approved):
Just one nit on naming, but this looks good!

recipes/configs/llama4/scout_17B_16E_full.yaml
IvanKobzarev: Ok. Agree with this logic,   will update to it.

PR #2624, opened 2025-04-22
scale_grads with foreach + compile

IvanKobzarev:
* __->__ #2624

Scaling gradients with for_each, adding also compilation of it.
9 -> 9.8 (max value of tokens_per_second in the first 10 iterations)

Helps with tokens_per_second (for llama4 the parameters are on cpu, not super big wins, but should be better when parameters are on gpu)

```
tune run --nproc_per_node 8 \
  full_finetune_distributed \
```
PS:
Current repo compilation fails if to set skip_rope_interval=4,, have to test with skip_rope_interval=None,


recipes/full_finetune_distributed.py
```diff
@@ -913,8 +913,19 @@ def train(self) -> None:
                         torch.distributed.all_reduce(num_tokens)
                         # This will ensure that the logged loss matches what we're optimizing
                         torch.distributed.all_reduce(running_loss)
+
                         # Manually scale the gradients from unnormalized loss by total # of tokens
-                        training.scale_grads(self._model, self.dp_degree / num_tokens)
+                        def scale_grads_fn():
```
joecummings, line 918: Dumb q: Why can't we compile the function directly instead of having this extra function?

joecummings (commented):
Cool work! Just a couple comments.

recipes/full_finetune_distributed.py
joecummings: Also, will this be addressed in this PR?

```Current repo compilation fails if to set skip_rope_interval=4,, have to test with skip_rope_interval=None,```

Comment (IvanKobzarev) [reply recipes/full_finetune_distributed.py:918]: > skip_rope_interval

There is a problem in flex_attention mask_mod - "Cuda illegal memory access" - some bad calculation of indices.
This needs a separate fix, will ask Flex attention engineers to take a look at a minimized repro.

Comment (joecummings) [reply recipes/full_finetune_distributed.py:920]: > I have not tried compiling object.method before. I can try.

@IvanKobzarev Did this work?

File: torchtune/training/_grad_scaler.py [hunk 2]
```diff
@@ -29,3 +34,70 @@ def scale_grads(model: nn.Module, scaler: torch.Tensor) -> None:
...
+        parameters = list(parameters)
+    _scale_grad_(parameters, scaler, foreach)
+
+
+def _group_tensors_by_device_and_dtype(
+    tensors: list[torch.Tensor],
+) -> dict[tuple[torch.device, torch.dtype], list[Tensor]]:
+    ret = defaultdict(list)
+    for i, tensor in enumerate(tensors):
+        ret[(tensor.device, tensor.dtype)].append(tensor)
```
ebsmothers, line 73: Noob q: where are we actually using the dtype groupby? Is it implicit in the usage of `_foreach_mul_`?

recipes/full_finetune_distributed.py
ebsmothers: Separately I have a similar question to the one I left on the compiled optimizer step PR -- do we need to do this every step? If not I wonder if we could just do something like `self.grad_scaler = get_grad_scale_fn(self._compile)` once during init, returning the maybe-compiled scale function, then call it in a single line here.

torchtune/training/_grad_scaler.py
IvanKobzarev: Technically here it is not used. I planned to put this function into torch and that it can be useful for other functionality, but it can be a long process :) 
So we do not use it here and can remove dtype groupping :)

IvanKobzarev:
Updated diff to compile only during initialization and use `self.grad_scaler` in the loop.

Just FYI for testing: compilation at the moment needs workarounds for 2 different problems:
1. There is some problem with rng states preservation which can be workarounded
```
diff --git a/torch/_dynamo/convert_frame.py b/torch/_dynamo/convert_frame.py
index 668353867ab..493883542f9 100644
--- a/torch/_dynamo/convert_frame.py
+++ b/torch/_dynamo/convert_frame.py
@@ -249,8 +249,8 @@ def preserve_global_state(fn: Callable[_P, _T]) -> Callable[_P, _T]:
             prior_dtype = torch.get_default_dtype()
             torch_rng_state = torch.random.get_rng_state()
             cuda_rng_state = None
-            if torch.cuda.is_available():
-                cuda_rng_state = torch.cuda.get_rng_state()
+            # if torch.cuda.is_available():
+            #     cuda_rng_state = torch.cuda.get_rng_state()
             allow_tf32 = torch._C._get_cublas_allow_tf32()
             prior_fwd_from_src = torch.fx.graph_module._forward_from_src
             torch.fx.graph_module._forward_from_src = fx_forward_from_src_skip_result
@@ -281,8 +281,8 @@ def preserve_global_state(fn: Callable[_P, _T]) -> Callable[_P, _T]:
                 )
                 if prior_mobile_allocator_state != curr_mobile_allocator_state:
                     torch._C._unset_default_mobile_cpu_allocator()
-                if cuda_rng_state is not None:
-                    torch.cuda.set_rng_state(cuda_rng_state)
+                # if cuda_rng_state is not None:
+                #     torch.cuda.set_rng_state(cuda_rng_state)
                 torch._C._set_cublas_allow_tf32(allow_tf32)
                 torch.fx.graph_module._forward_from_src = prior_fwd_from_src
                 assert guards.check(), (
diff --git a/torch/_dynamo/utils.py b/torch/_dynamo/utils.py
index b75b1d6c39f..7ca67523704 100644
--- a/torch/_dynamo/utils.py
+++ b/torch/_dynamo/utils.py
@@ -2110,15 +2110,15 @@ def preserve_rng_state():
     with disable_current_modes(), disable_functorch():
         rng_state = torch.clone(torch.random.get_rng_state())
         skip_frame_if_in_functorch_mode(rng_state)
-        if torch.cuda.is_available():
-            cuda_rng_state = torch.clone(torch.cuda.get_rng_state())
+        # if torch.cuda.is_available():
+        #     cuda_rng_state = torch.clone(torch.cuda.get_rng_state())
     try:
         yield
     finally:
         with torch.utils._python_dispatch._disable_current_modes():
             torch.random.set_rng_state(rng_state)
-            if torch.cuda.is_available():
-                torch.cuda.set_rng_state(cuda_rng_state)  # type: ignore[possibly-undefined]
+            # if torch.cuda.is_available():
+            #     torch.cuda.set_rng_state(cuda_rng_state)  # type: ignore[possibly-undefined]
 
 
 def is_jit_model(model0):
 ```
 
 2. There is illegal memory access in Chunked flex Attention x Caching.
 
 If to remove /tmp/torchinductor_${USER} before every run - then it does not fires (or disable pt2 cache)

File: torchtune/training/_grad_scaler.py [hunk 3]
```diff
@@ -4,8 +4,13 @@
...
+from typing import Optional
+
 import torch
-from torch import nn
+from torch import nn, Tensor
+from torch.nn.utils.clip_grad import _no_grad, _tensor_or_tensors
+from torch.utils._foreach_utils import _device_has_foreach_support, _has_foreach_support
 
 
 def scale_grads(model: nn.Module, scaler: torch.Tensor) -> None:
```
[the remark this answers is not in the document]
joecummings: Talked with Evan, let's add the deprecation label here.

joecummings (approved):
LGTM, just deprecate the old scale_grads

recipes/full_finetune_distributed.py
```diff
@@ -314,10 +314,19 @@ def setup(self, cfg: DictConfig) -> None:
         self._compile_model = compile_bool
         self._compile_loss = compile_bool
         self._compile_optimizer_step = compile_bool
+        self._compile_scale_grads = compile_bool
         if isinstance(compile, DictConfig):
             self._compile_model = compile.get("model", True)
             self._compile_loss = compile.get("loss", True)
             self._compile_optimizer_step = compile.get("optimizer_step", False)
+            self._compile_scale_grads = compile.get("scale_grads", True)
```
ebsmothers, line 322: nit but I actually wonder whether it's worth having a separate flag for this. I think model, loss, optimizer step are all pretty clear but this one may be a bit niche for someone to look at and immediately understand (and also I think it's unlikely for someone to actually want to experiment with). Could even just use some heuristic like `self._compile_scale_grads = all([self._compile_model, self._compile_loss, self._compile_optimizer_step])`

PR #2655, opened 2025-04-30
Add lora weights to state dict conversions

lucylq:
#### Context
- [ x] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

#### Changelog
* Add LoRA weights to Meta<->torchtune state dict conversions

#### Test plan

- [ x] run unit tests via `pytest tests`

#### UX

- [ x] I did not change any public API

### DIFF
--- a/torchtune/models/convert_weights.py
+++ b/torchtune/models/convert_weights.py
@@ -24,6 +24,15 @@ _FROM_META = {
     "layers.{}.feed_forward.w1.weight": "layers.{}.mlp.w1.weight",
     "layers.{}.feed_forward.w2.weight": "layers.{}.mlp.w2.weight",
     "layers.{}.feed_forward.w3.weight": "layers.{}.mlp.w3.weight",
+    # lora weights
+    "layers.{}.attention.wq.lora_a.weight": "layers.{}.attn.q_proj.lora_a.weight",
+    "layers.{}.attention.wq.lora_b.weight": "layers.{}.attn.q_proj.lora_b.weight",
+    "layers.{}.attention.wk.lora_a.weight": "layers.{}.attn.k_proj.lora_a.weight",
+    "layers.{}.attention.wk.lora_b.weight": "layers.{}.attn.k_proj.lora_b.weight",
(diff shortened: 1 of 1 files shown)


joecummings (approved):

ebsmothers:
Hi @lucylq can you share a bit more context on this? Most of our state dict mappings do not require LoRA keys because we merge the LoRA weights back into the original model after training. And if you are trying to load the adapter weights separately from e.g. a model in the Meta format trained with torchtune, they should already be saved in torchtune's format ([ref](https://github.com/pytorch/torchtune/blob/0ddd4b93c83de60656fb3db738228b06531f7c1e/torchtune/training/checkpointing/_checkpointer.py#L1225-L1235))

lucylq:
> Hi @lucylq can you share a bit more context on this? Most of our state dict mappings do not require LoRA keys because we merge the LoRA weights back into the original model after training. And if you are trying to load the adapter weights separately from e.g. a model in the Meta format trained with torchtune, they should already be saved in torchtune's format ([ref](https://github.com/pytorch/torchtune/blob/0ddd4b93c83de60656fb3db738228b06531f7c1e/torchtune/training/checkpointing/_checkpointer.py#L1225-L1235))

Hi @ebsmothers yes for sure. I'm trying to enable LoRA as a use-case for program-data sharing in executorch. We'd like for separate LoRA models to share the same foundation weights to reduce on-device memory usage; this means we can't merge the lora weights in. The final file structure would look something like this:

- lora1.pte: model + lora weights
- lora2.pte: model + lora weights
- foundation.ptd: foundation weights only.

I'm currently using adapter weights from torchtune, and want to convert them back to meta, so we can use it with the meta llama checkpoint if that's what a user ends up having. I'd like to use torchtune's convert function, but I can also create a copy in the executorch repo if that's preferred?

ebsmothers:
Thanks @lucylq for the additional info. Given that we only call tune_to_meta/meta_to_tune in torchtune on base checkpoints (or checkpoints whose LoRA weights have already been merged back into the original model), I think this change should be fine from torchtune's save/load point of view. So this looks good to me.

PR #2659, opened 2025-05-02
llama4 distributed: compile optimizer

IvanKobzarev:
* #2644
* #2624
* __->__ #2659

Sorry, recreating PR https://github.com/pytorch/torchtune/pull/2623 that I merged into ghstack branch [gh/IvanKobzarev/1/base](https://github.com/pytorch/torchtune/tree/gh/IvanKobzarev/1/base) instead of main

Fully copied PR with the latest state:

Compiling optimizer helps perf of Llama4 Scout Model
3.8 tokens_per_second -> 9 tokens_per_second (max value of tokens per second in the first ~10 iterations)
peak memory is the same

tune run --nproc_per_node 8 \
  full_finetune_distributed \

### DIFF
--- a/recipes/full_finetune_distributed.py
+++ b/recipes/full_finetune_distributed.py
@@ -4,6 +4,7 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
+import os
 import sys
 import time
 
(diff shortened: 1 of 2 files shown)


recipes/full_finetune_distributed.py
```diff
@@ -306,7 +307,18 @@ def setup(self, cfg: DictConfig) -> None:
...
+        compile = cfg.get("compile")
+        compile_bool = bool(compile)
+        self._compile_backend = os.environ.get("TORCH_COMPILE_BACKEND", "inductor")
+
+        self._compile_model = compile_bool
+        self._compile_loss = compile_bool
+        self._compile_optimizer_step = compile_bool
+        if isinstance(compile, DictConfig):
+            self._compile_model = compile.get("model", True)
+            self._compile_loss = compile.get("loss", True)
```
ebsmothers, line 319: A minor thing, but I wonder whether we should default to False here just to match the default when the user is not passing a dict. Not a huge deal since users will likely need to opt-in by passing the dict anyways

ebsmothers (approved):
One small (non-blocking) comment. Thanks for adding this!

recipes/full_finetune_distributed.py
IvanKobzarev: I look at this values as a default pieces that we recommend user to compile if user specified compilation as dict and did not specify exactly model, loss. 
I think this will be similar alignment what was before for "compile: True", where it was compiling model and loss.

PR #2667, opened 2025-05-02
Re-refactor Linear Loss

pbontrager:
#### Context
- [ ] add a new feature
- [x] fix a bug
- [ ] other (please add here)

The new linear loss just set as the replacement for CEWithChunkedOutputLoss exposed some issues with how we support DTensor inputs with our losses. The implementation also didn't work with TP as the linear_projection_weight needed to handle sharding and movement of data. Finally all of the recipe changes made to support it, broke CEWithChunkedOutputLoss, which while being deprecated should still work.

This PR simplifies the type system to only have one SFTLoss and introduces the "update_model_output" method for losses. This allows losses to contain any custom code that they need for controlling the model output without all of the recipes needing to control for that. CE and LinearCE use this new method now so that all the recipes can support both and we can more easily accommodate future losses. We also now directly use model.output instead of the weight to better support distributed modules.

This also fixes an outstanding issues where tied embeddings would fail for TP since their outputs would be DTensors.

Unfortunately, due to this [DTensor issue](https://github.com/pytorch/pytorch/issues/152717) we cannot support LinearCE with TP models. To fix this I've reverted all configs to CE loss for now and removed the deprecate warnings.

***EDIT***
@XilunWu helped to find a workaround for DTensor so we're able to keep LinearCE as the default

The new api here looks like:
```python
loss = config.instantiate(cfg.loss)
if isinstance(loss, SFTLoss):
	loss.set_model_output(model)

...

loss(outputs, labels)
```

#### Changelog
* reverted all config defaults to CEWithChunkedOutputLoss
* Converted loss types to ABC as protocols don't offer efficient run time checks
* Removed LinearLoss for set_model_output method
* Updated recipes for new loss interface
* Added additional output/label processing for non-SFTLoss losses
* Updated CEWithChunkedOutputLoss to handle DTensor inputs
* Added error to LinearLoss if input is DTensor to not use this loss for now
* Added more descriptive doc strings

#### Test plan

To test tied embeddings with CE loss run
`tune run --nproc_per_node 8 full_finetune_distributed --config llama3_2/3B_full max_steps_per_epoch=8 compile=False tensor_parallel_dim=4 tensor_parallel_plan._component_=torchtune.models.llama3.base_llama_tp_plan`


torchtune/modules/loss/ce_chunked_output_loss.py
```diff
@@ -4,18 +4,21 @@
...
 
-from torchtune.utils import get_logger, log_once
+# from torchtune.utils import deprecated
 
-logger = get_logger("DEBUG")
+from .loss_types import SFTLoss
 
 
-class CEWithChunkedOutputLoss(torch.nn.Module):
+# TODO: switch to linear_loss
```
joecummings, line 19: ?

torchtune/modules/transformer.py
```diff
@@ -4,17 +4,14 @@
...
 
 import torch
 from torch import nn
-from torch.distributed.fsdp import FSDPModule
 from torchtune.modules import MultiHeadAttention
 from torchtune.modules.attention_utils import _MaskType
-from torchtune.utils import get_logger, log_once
 
-logger = get_logger("DEBUG")
+# from torchtune.utils import deprecated
```
joecummings, line 14: ?

torchtune/modules/transformer.py
```diff
@@ -401,21 +398,17 @@ def __init__(
...
         self.num_output_chunks = 0
-        self.skip_linear_projection = False
+        self.skip_output = False
 
         # attributes for KV caches during inference
         self.encoder_max_cache_seq_len = None
         self.decoder_max_cache_seq_len = None
 
+    # TODO: switch to linear_loss
+    # @deprecated("Please use LinearCrossEntropyLoss instead")
```
joecummings, line 408: ?

torchtune/modules/transformer.py
```diff
@@ -401,21 +398,17 @@ def __init__(
         self.head_dim = head_dim
         self.causal_mask = None
         self.num_output_chunks = 0
-        self.skip_linear_projection = False
+        self.skip_output = False
```
joecummings, line 401: nit: skip_output is a little vague for my taste.

torchtune/modules/transformer.py
```diff
@@ -497,20 +490,8 @@ def reset_caches(self):
...
-        # Accessing the weight directly will not trigger FSDP hooks
-        # to gather the full tensor so we have to unshard manually
-        if isinstance(self.output, FSDPModule):
-            self.output.unshard()
-            weight = self.output.weight.clone()
-            self.output.reshard()
-            return weight
-        return self.output.weight
-
+    # TODO: switch to linear_loss
```
joecummings, line 493: ?

joecummings (commented):
Where are the tests for LinearLoss?

torchtune/modules/model_fusion/_deep_fusion.py
```diff
@@ -97,15 +94,10 @@ def __init__(
             trainable_params -= set(get_fusion_params(self))
         set_trainable_params(self, trainable_params)
 
+    @deprecated("Please use self.skip_output=True and use a linear loss instead")
```
ebsmothers, line 97: nit but I feel like "use a linear loss" is not really sufficient to describe how someone should migrate off of this

torchtune/modules/loss/loss_types.py
```diff
@@ -0,0 +1,36 @@
...
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+from abc import ABC, abstractmethod
+
+import torch
+from torch import nn
+
+
+class SFTLoss(ABC):
```
ebsmothers, line 13: What's our plan for non-SFT losses? E.g. for forward KL loss as used in KD -- do we need to define another abstract class for those too? (Also nit but why not protocol?)

torchtune/modules/loss/loss_types.py
```diff
@@ -0,0 +1,36 @@
...
+    """Interface for loss functions in torchtune used in sft recipes."""
+
+    def apply_compile_strategy(self, *args, **kwargs):
+        """Compile the loss function for inference."""
+        self.forward = torch.compile(self.forward, *args, **kwargs)
+        return self
```
ebsmothers, line 19: Why bother defining this here? We seem to define it in the child classes anyways

recipes/dev/lora_finetune_distributed_multi_dataset.py
```diff
@@ -828,13 +827,15 @@ def train(self) -> None:
...
-                if self.linear_loss:
-                    weight = self._model.linear_projection_weight
-                    current_loss = self._loss_fn(weight, outputs, labels)
-                else:
+                # post process for third party loss functions
+                if not isinstance(self._loss_fn, modules.loss.SFTLoss):
                     labels = labels.reshape(-1)
                     outputs = outputs.reshape(-1, outputs.size(-1))
-                    current_loss = self._loss_fn(outputs, labels)
+                    if isinstance(outputs, DTensor):
```
ebsmothers, line 834: nit but would at least add a comment explaining when/why we need this

torchtune/modules/loss/cross_entropy_loss.py
```diff
@@ -7,14 +7,24 @@
...
+from .loss_types import SFTLoss
 
 
-class LinearCrossEntropyLoss(nn.Module, SFTLinearLoss):
+class LinearCrossEntropyLoss(nn.Module, SFTLoss):
     """Memory efficient Cross-entropy loss that incrementally computes loss for chunks of tokens
     by masking ignored tokens, calculating logits and then applying cross-entropy loss. Combines
     the linear projection with the cross-entropy calculation for futher memory savings.
+
+    Linear cross entropy entropy masks out ignored tokens before the projection layer to save memory.
```
ebsmothers, line 20:
```suggestion
    Linear cross entropy masks out ignored tokens before the projection layer to save memory.
```

torchtune/modules/loss/cross_entropy_loss.py
```diff
@@ -43,32 +54,52 @@ def apply_compile_strategy(self, *args, **kwargs):
...
+                hidden_chunk = hidden_chunk[mask_chunk]  # [num_valid, embed_dim]
         else:
             hidden_chunk = hidden_chunk.reshape(-1, hidden_chunk.shape[-1])
             target_chunk = target_chunk.reshape(-1)
 
         # [num_valid, embed_dim] @ [embed_dim, vocab_size]
-        logits = F.linear(hidden_chunk, weight)  # [num_valid, vocab_size]
+        if self.linear_projection is None:
+            raise AttributeError("forward called before update_model")
+        logits = self.linear_projection(hidden_chunk)  # [num_valid, vocab_size]
```
ebsmothers, line 100: Dumb q: did we confirm that with compile the linear projection and cross_entropy ops definitely get fused? (Separately would be useful to pick a model like 3.2 1B or something and show peak memory for each of the different options we support)

torchtune/modules/loss/cross_entropy_loss.py
```diff
@@ -79,19 +110,14 @@ def compute_cross_entropy(
 
     def forward(
         self,
-        weight: torch.Tensor,
         outputs: torch.Tensor,
```
ebsmothers, line 113: I guess outputs is a bit of a misnomer here? Do we have to keep it this way for BC?

pbontrager: These are model outputs. I'm not sure what would be a better name. They used to be called logits, but they aren't logits in the linear loss case.

torchtune/modules/loss/loss_types.py
pbontrager: Yes, we have to define one per recipe type at this point. I'd be open to finding a different access to define these on but this is what there was and it seems reasonable.

re net: protocols don't work well for interfaces. They're very inefficient for runtime checks and not as clear a definition for users to extend (ABCs can be traced directly by the user: see an example class -> go to abc to see how to write your own). At this point I'm in favor of phasing protocols out of torchtune.

torchtune/modules/transformer.py
pbontrager: Used skip_output instead of skip_output_layer since the layer is literally called "output"

recipes/dev/lora_finetune_distributed_multi_dataset.py
ebsmothers: I guess.. as a reader of that my next question would be "why is the tensor sharded to begin with?" In normal cases of distributed training it's not required to do this manually, right?

torchtune/modules/loss/loss_types.py
ebsmothers: hmm ok iNheRItAnCe

torchtune/modules/classifier.py
```diff
@@ -41,16 +42,14 @@ def classifier_model(
 
     """
     model = _get_component_from_path(base_model_path)(**base_model_kwargs)
+    if hasattr(model, "decoder"):
+        decoder = model.decoder
+    else:
+        decoder = model
```
Andrei-Aksionov, line 48:
```suggestion
    decoder = getattr(model, "decoder", model)
```

torchtune/modules/loss/cross_entropy_loss.py
```diff
@@ -7,33 +7,40 @@
...
 
-from .loss_protocols import SFTLinearLoss
+from .loss_types import SFTLoss
 
 
-class LinearCrossEntropyLoss(nn.Module, SFTLinearLoss):
+class LinearCrossEntropyLoss(nn.Module, SFTLoss):
     """Memory efficient Cross-entropy loss that incrementally computes loss for chunks of tokens
     by masking ignored tokens, calculating logits and then applying cross-entropy loss. Combines
     the linear projection with the cross-entropy calculation for futher memory savings.
```
Andrei-Aksionov, line 18:
```suggestion
    the linear projection with the cross-entropy calculation for further memory savings.
```

torchtune/modules/classifier.py
```diff
@@ -41,16 +42,14 @@ def classifier_model(
...
 
     if hasattr(model, "output"):
```
Andrei-Aksionov, line 50: How likely that any given model doesn't have `output`?
LinearCrossEntropy pretty much depends on it.

torchtune/modules/loss/cross_entropy_loss.py
Andrei-Aksionov: Definitely needs such confirmation.
If everything matches expectations, I would go even further and advise using compilation for this loss. Otherwise, without compilation, it's not a Linear CE at all, just your regular CE.

torchtune/modules/transformer.py
Andrei-Aksionov: With `skip_output` this one looks odd to me:
```python
        if self.skip_output:
            output = h
```

If we want to skip an output, then why we're assigning to a variable called `output` 🤔 ?

So it's not like you're skipping the output, just skipping some action.

`self.skip_output_projection`?

recipes/dev/lora_finetune_distributed_multi_dataset.py
```diff
@@ -828,13 +827,15 @@ def train(self) -> None:
...
+                        outputs = outputs.full_tensor()
+
+                # Compute loss
+                current_loss = self._loss_fn(outputs, labels)
 
                 # Loss is normalized by default so we multiply by the number of tokens
```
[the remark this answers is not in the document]
pbontrager: I'm way too lazy to update that comment in every recipe right now, haha

torchtune/modules/classifier.py
pbontrager: Fair, I left that check in as it was already there. But every model has to have an output in torchtune currently.

torchtune/modules/loss/cross_entropy_loss.py
pbontrager: The main benefit of this implementation is that we get the masking before the logits which will save a lot of memory.  Getting the linear and CE is out of scope for now. The planned followup was to add support for the liger implementation.

torchtune/modules/loss/loss_types.py
```diff
@@ -0,0 +1,74 @@
...
+        pass
+
+    @abstractmethod
+    def forward(
+        self,
+        pi_old_outputs: torch.Tensor,  # [B x G, L]
+        pi_outputs: torch.Tensor,  # [B x G, L]
+        ref_outputs: torch.Tensor,  # [B x G, L]
+        advantages: torch.Tensor,  # [B x G]
+        padding_masks: Optional[torch.Tensor] = None,  # [B x G, L]
```
ebsmothers, line 61: Not blocking but are we sure this is generic enough to call this our `RLLoss` ABC? I think we may wanna revisit this

torchtune/modules/loss/cross_entropy_loss.py
```diff
@@ -43,32 +50,52 @@ def apply_compile_strategy(self, *args, **kwargs):
...
-            target_chunk = target_chunk.reshape(-1)
+            hidden_chunk = hidden_chunk[mask_chunk]  # [num_valid, embed_dim]
 
         # [num_valid, embed_dim] @ [embed_dim, vocab_size]
-        logits = F.linear(hidden_chunk, weight)  # [num_valid, vocab_size]
+        if self.linear_projection is None:
+            raise AttributeError("forward called before update_model")
+        logits = self.linear_projection(hidden_chunk)  # [num_valid, vocab_size]
+        if isinstance(logits, DTensor):
+            logits = logits.full_tensor()
```
nathan-az, line 98: Just a note - materialisation of the full logit chunk on all TP workers can be avoided with either:
1. using PyTorch's [loss_parallel](https://docs.pytorch.org/docs/stable/distributed.tensor.parallel.html#torch.distributed.tensor.parallel.loss_parallel) - this would require slightly more change since the ctx manager needs to wrap the `backward` step, or
2. applying sequence-dim sharding, which should play nicely since CELoss is independent on that dimension

(2) might actually be relatively easy:

```py
"output": ColwiseParallel(input_layouts=Shard(1), output_layouts=Replicate()),
```

Pretty sure dim `1` is the sequence dimension already, so if we output on that dimension as well I think we save on an all-to-all and the duplicated work. 

I understand that might be out of scope for this PR, and we may simply not want to force mixing of parallelism schemes (especially if applying more wide-scale context parallel is eventually on the roadmap), just noting here.

pbontrager: Thank you for sharing this! I'll include this context in an issue for some followup work as compile doesn't always work here either. 

> applying sequence-dim sharding, which should play nicely since CELoss is independent on that dimension
The reason I went down the `outputs.full_tensor()` route instead of playing with the parallelize output direction is that I wanted this to work with tied embeddings two and this seemed more robust. Though option 2 could likely be accomplished using dtensor.redistribute.

joecummings (approved):
Unblocking to unbreak main

PR #2668, opened 2025-05-02
1M+ context length (context parallel integration)

ebsmothers:
Screenshot 2025-05-28 at 4 38 57 PM

Initial implementation of context parallelism in torchtune. We add a utility `get_context_parallel_context` to return a context manager whenever context parallelism is enabled (otherwise it's basically nullcontext). We also update our collate function to pad to 2 * context_parallel_dim (needed due to [this](https://github.com/pytorch/pytorch/blob/4f62dcc/torch/distributed/tensor/experimental/_attention.py#L1246)).

### Tested
- [x] CP only
- [x] CP + TP
- [x] CP + DP shard
- [x] CP + DP replicate
- [x] Composability with activation checkpointing + offloading
- [x] Composability with optimizer in backward

### Still needs work (i.e. not addressed in this PR)
🚧 Composability with flex attention ([PyTorch PR](https://github.com/pytorch/pytorch/pull/151497), [torchtitan PR](https://github.com/pytorch/torchtitan/pull/1160) shared by @XilunWu)
🚧 Composability with fp8
🚧 Multimodal models

### Test plan

[Sweep script](https://gist.github.com/ebsmothers/d93df961c06563c4d8f16c0905f7df12) to test a bunch of different configurations across Llama3 8B and Llama 3.2 1B models.

[WandB project](https://wandb.ai/ebs/context-parallel-final-testing) with all the runs

Also used [this script](https://gist.github.com/ebsmothers/8040d43dd76d2fa636ad33e058a2eb7b) to generate a synthetic dataset of concatenated Alpaca samples with most samples at ~1M context length. 

Using that synthetic dataset, the screenshot above can be reproduced by running

```
tune run --nproc_per_node 8 full_finetune_distributed --config llama3_2/1B_full \
context_parallel_dim=8 batch_size=1 enable_activation_checkpointing=True \
enable_activation_offloading=True loss.num_output_chunks=128 \
model._component_=torchtune.models.llama3_2.llama3_2 model.vocab_size=128_256 \
model.num_layers=16 model.num_heads=32 model.num_kv_heads=8 model.embed_dim=2048 \
model.intermediate_dim=8192 model.max_seq_len=10_000_000 gradient_accumulation_steps=1 \
dataset._component_=torchtune.datasets.text_completion_dataset \
dataset.source=/tmp/long_alpaca_dataset/train max_steps_per_epoch=50
```


torchtune/training/_distributed.py
```diff
@@ -718,3 +743,108 @@ def prepare_mha_for_tp(
     if is_fusion_model:
         model.decoder = decoder
     return model
+
+
+def _get_sdpa_context() -> (
```
pbontrager, line 748: Does this mean CP doesn't work with FlexAttention?

recipes/full_finetune_distributed.py
```diff
@@ -888,6 +892,18 @@ def train(self) -> None:
 
                 utils.batch_to_device(batch, self._device)
 
+                # Define optional context manager for context parallelism
+                model_inputs = list(batch.values())
+                buffers = list(self._model.buffers())
+                optional_context_parallel_context_manager = (
```
pbontrager, line 898: Is this the naming we're using for other optional ctx managers? We have "activations_handling_ctx", though I'd prefer to consolidate on something like "context_parallel" or "maybe_context_parallel", I think the "with" statement says it's a context manager.

ebsmothers: Yeah I'm good taking out the "optional" here and matching what we do for activation offloading

torchtune/training/_distributed.py
ebsmothers: Yes, at least until https://github.com/pytorch/pytorch/pull/151497 lands

ebsmothers: But I also think this is somewhat orthogonal. Like flex does not have its own backend (see [here](https://github.com/pytorch/pytorch/blob/172a7c942edd4394f81fd0e889f10822af916b9e/aten/src/ATen/SDPBackend.h#L7)). My assumption is that it should be using the flash attention backend (but need to confirm)

XilunWu: curious if torchtune already has non-CP flex_attention? cc @ebsmothers

ebsmothers: @XilunWu yes we do, see e.g. [here](https://github.com/pytorch/torchtune/blob/dedc209c503b9ddcc3dbb62aeb5ae0dd9093b46f/torchtune/modules/attention_utils.py#L195-L227) for how we apply it for sample packing

torchtune/training/_distributed.py
```diff
@@ -726,3 +751,112 @@ def prepare_mha_for_tp(
...
+                stack.enter_context(cp_context)
+
+            yield
+
+    return context
+
+
+def get_context_parallel_context(
+    *,
+    cp_enabled: bool = False,
```
pbontrager, line 788: Should just be "enabled", cp seems redundant in this function

torchtune/training/_distributed.py
```diff
@@ -726,3 +751,112 @@ def prepare_mha_for_tp(
...
+    """
+
+    if cp_enabled and "cp" not in world_mesh.mesh_dim_names:
+        raise ValueError(
+            "Context parallel is enabled but no context parallel device mesh is provided."
+        )
+    # TODO: context parallel for multimodal models requires extra work
+    if cp_enabled and not isinstance(model, TransformerDecoder):
+        raise ValueError(
+            "Context parallel not supported for models other than TransformerDecoder"
```
pbontrager, line 834: I think this error message should just state that "Only text models are supported" as a user might not understand why they're getting this error.

torchtune/training/_distributed.py
```diff
@@ -726,3 +751,112 @@ def prepare_mha_for_tp(
...
+        )
+    model_buffers = list(model.buffers())
+
+    @contextlib.contextmanager
+    def context():
+        # Create context parallel context if enabled
+        cp_context = None
+        if (
+            cp_enabled
```
pbontrager, line 843: Why not just return a NullContext if not enabled?

torchtune/training/_distributed.py
```diff
@@ -57,26 +61,28 @@ class ParallelDims:
     dp_replicate: int
     dp_shard: int
     tp: int
+    cp: int
```
pbontrager, line 64: More of a meta comment, but there has to be a better way to reason about all of these interactions than building a literal tree of if statements

torchtune/training/__init__.py
```diff
@@ -11,6 +11,7 @@
 from torchtune.training._compile import compile_loss, compile_model
 from torchtune.training._distributed import (
     gather_cpu_state_dict,
+    get_context_parallel_context,
```
pbontrager, line 14: get_context_parallel_manager would read nicer

torchtune/training/_distributed.py
```diff
@@ -726,3 +751,112 @@ def prepare_mha_for_tp(
...
+
+    return context
+
+
+def get_context_parallel_context(
+    *,
+    cp_enabled: bool = False,
+    world_mesh: torch.distributed.DeviceMesh,
+    model: TransformerDecoder,
+    model_inputs: list[torch.Tensor],
```
pbontrager, line 791: This should take a dict, since all our recipes use a nested dict for inputs. Also, you can't just call list(batch.values) since there are nested dicts. Since we don't support multimodal for now, you can just explicitly check that tokens are inside the batch in this function.

pbontrager (commented):
Thank you for enabling this!  Left some feedback about how the manager is structured, and some questions on whether we're safely handling all of our edge cases.

torchtune/training/_distributed.py
```diff
@@ -726,3 +751,112 @@ def prepare_mha_for_tp(
...
+        )
+    model_buffers = list(model.buffers())
+
+    @contextlib.contextmanager
+    def context():
+        # Create context parallel context if enabled
+        cp_context = None
+        if (
+            cp_enabled
+            and world_mesh is not None
```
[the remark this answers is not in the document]
ebsmothers: Good point, will remove

torchtune/training/_distributed.py
ebsmothers: We can do that too, but this gives us the option to explicitly order the SDPBackends as we prefer them, so personally I think it's useful.

torchtune/training/_distributed.py
```diff
@@ -726,3 +751,112 @@ def prepare_mha_for_tp(
...
+                world_mesh["cp"],
+                buffers=model_inputs + model_buffers,
+                buffer_seq_dims=[1] * len(model_inputs) + [0] * len(model_buffers),
+                no_restore_buffers=set(model_inputs),
+            )
+
+        # Create and enter the train context with the optional cp_context
+        sdpa_context = _get_sdpa_context()
+
+        with sdpa_context(cp_context):
```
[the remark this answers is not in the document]
ebsmothers: The latest updates should address this. For packing we don't know until forward, so unfortunately have to check inside of `context`. For Llama4 we check straightaway whether any layer's `mask_mod` is not None (though also we will error anyways since it's an early fusion model).

recipes/full_finetune_distributed.py
```diff
@@ -158,6 +158,7 @@ def __init__(self, cfg: DictConfig) -> None:
             raise ValueError(
                 "Tensor Parallel plan needs to be provided when tensor parallel is enabled."
             )
+        self.cp_degree = cfg.get("context_parallel_dim", 1)
```
joecummings, line 161: Can you put this as a default in one of the full finetune distributed recipes?

torchtune/models/llama3_2/_model_builders.py
```diff
@@ -4,6 +4,7 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 from functools import partial
+from typing import List
```
joecummings, line 7: no. bad. use ``list``.

torchtune/models/llama3_2/_model_builders.py
```diff
@@ -67,6 +72,8 @@ def llama3_2_3b(
         scale_factor=32,
         tie_word_embeddings=tie_word_embeddings,
     )
+
```
joecummings, line 75: nonsense

torchtune/training/_distributed.py
```diff
@@ -99,17 +106,36 @@ def build_mesh(self, device_type):
...
+            mesh[tuple(dp_shard_cp_mesh_dim_names)]._flatten(
+                mesh_dim_name="dp_shard_cp"
+            )
+        if dp_cp_mesh_dim_names != []:
+            mesh[tuple(dp_cp_mesh_dim_names)]._flatten(mesh_dim_name="dp_cp")
+
         return mesh
 
+    @property
+    def enabled(self):
```
joecummings, line 136: shouldn't this be `cp_enabled`?

torchtune/training/_distributed.py
```diff
@@ -726,3 +767,110 @@ def prepare_mha_for_tp(
...
+        ValueError: if enabled is True but world_mesh does not contain a "cp" dimension
+
+    Example:
+        ```python
+        batch = {"inputs": inputs, "labels": labels}
+        with get_context_parallel_context(
+            enabled=parallel_dims.enabled,
+            cp_mesh=world_mesh["cp"] if parallel_dims.enabled else None,
+            model_inputs=list(batch.values()),
+            model_buffers=model.buffers(),
```
joecummings, line 832: What is this example?

joecummings (approved):
stamping to unblock but please fix the comments

torchtune/training/_distributed.py
ebsmothers: blargh i did find and replace all in the file, sorry. will fix

recipes/full_finetune_distributed.py
```diff
@@ -920,17 +933,17 @@ def train(self) -> None:
...
-                current_loss = self._loss_step(batch) * current_num_tokens
-                running_loss += current_loss
-
-                # For optimizer in backward, we need to normalize before calling backward
-                # This case and gradient accumulation are mutually exclusive
-                if self._optimizer_in_bwd:
-                    torch.distributed.all_reduce(num_tokens)
-                    torch.distributed.all_reduce(running_loss)
-                    current_loss = current_loss * (self.dp_degree / num_tokens)
+                with self.context_parallel_manager(list(batch.values())):
```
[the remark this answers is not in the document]
ebsmothers: I guess? But list is a more natural datatype, no? And also that's ultimately what core's context manager expects anyways so I would prefer to stay close to that

torchtune/models/llama3_2/_model_builders.py
ebsmothers: these clearly should have been here already, no?

SKRohit:
@ebsmothers Why are we not sharding fsdp model on `dp_shard_cp` dimension instead of `dp_shard` like in torchtitan? We could fit a larger seq length and training could become faster.

ebsmothers:
Hi @SKRohit nice catch, we should be doing this. Would you be interested in opening a PR with the fix?

SKRohit:
@ebsmothers sure. I am working on it. Will push a pr soon.

PR #2637, opened 2025-04-25
Async RL prototype

Darktex:
#### Context
What is the purpose of this PR? Is it to
- [X] add a new feature
- [ ] fix a bug
- [ ] update tests and/or documentation
- [ ] other (please add here)

This PR adds a prototype for async RL in TorchTune. Compared to the sync version of this code, we are seeing over 10x speedup. The API still has a few rough edges, and we are looking to make this more generic for the future, but this represents a good first step towards a fully async RL solution.

#### Changelog
What are the changes made in this PR? TLDR is:

#### Test plan
Please make sure to do each of the following if applicable to your PR. If you're unsure about any one of these just ask and we will happily help. We also have a [contributing page](https://github.com/pytorch/torchtune/blob/main/CONTRIBUTING.md) for some guidance on contributing.

- [X] run pre-commit hooks and linters (make sure you've first installed via `pre-commit install`)
- [X] add [unit tests](https://github.com/pytorch/torchtune/tree/main/tests/torchtune) for any new functionality
- [] update [docstrings](https://github.com/pytorch/torchtune/tree/main/docs/source) for any new or updated methods or classes
- [X] run unit tests via `pytest tests`
- [X] run recipe tests via `pytest tests -m integration_test`
- [X] manually run any new or modified recipes with sufficient proof of correctness
- [X] include relevant commands and any other artifacts in this summary (pastes of loss curves, eval results, etc.)

#### UX
If your function changed a public API, please add a dummy example of what the user experience will look like when calling it.
Here is a [docstring example](https://github.com/pytorch/torchtune/blob/6a7951f1cdd0b56a9746ef5935106989415f50e3/torchtune/modules/vision_transformer.py#L285)
and a [tutorial example](https://pytorch.org/torchtune/main/tutorials/qat_finetune.html#applying-qat-to-llama3-models)

- [ ] I did not change any public API
- [ ] I have added an example to docs or docstrings

PR #2675, opened 2025-05-05
Decouple rl test job

ebsmothers:
Separating out our async RL tests into their own runner, which runs only under certain conditions.

We now have two GPU jobs:

`rl_test.yaml` runs only if any of the `dev/rl` or `dev/grpo` files are modified. It will run only our async RL tests, which now need to be decorated by `@rl_test()`. The YAML file currently assumes that any such tests live under `tests/recipes/dev/rl` or `tests/torchtune/dev/rl`. If we move our async RL tests or add any in a new location we will need to update the runner's pytest command to include them.

`gpu_test.yaml` runs all other tests (i.e. our full test suite prior to #2637) on every PR. 

For local pytest runs, async RL tests will be skipped by default. Any pytest command can be supplemented with the flag `--run-rl-tests` in order to run them. (Note that RL tests are a subset of integration tests, so you will also need to use either `-m integration_test` or `--with-integration` as well.)

### DIFF
--- a/.github/workflows/rl_test.yaml
+++ b/.github/workflows/rl_test.yaml
@@ -1,13 +1,21 @@
 name: GPU tests
 
 on:
-  schedule:
-    # Runs at midnight every day
-    - cron:  '0 0 * * *'
   push:
-    branches: [ main ]
+    paths:
+      - 'torchtune/recipes/dev/**grpo**'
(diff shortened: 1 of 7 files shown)


joecummings (commented):
Eventually these should just be delineated by a name like "long running tests" rather than RL, but fine with this for now.

PR #2677, opened 2025-05-06
Loss Refactor - TP Fixes

nathan-az:
#### Context
- [ ] add a new feature
- [x] fix a bug
- [ ] other (please add here)

New loss refactor looks great, appears to encounter a couple of issues in TP:

* `outputs` comes out of the model as a `DTensor` but labels are `Tensor`, causing "mixed type" error
* loss comes out as a DTensor so does not play nicely with `all_reduce`

Solutions in this PR are minimally invasive (although somewhat lazy).

1. `labels` are cast to the same mesh and placements as outputs
2. `running_loss` retrieves the local tensor if it's a DTensor. This seems most correct since we do an all reduce across all ranks (as opposed to `full_tensor` which would likely cause loss to be over-counted)

Note that the new loss refactor does not appear to work with `mask_pre_projection`. I believe this is because masking doesn't play nicely with `DTensors` sharded on the index dimension where the shape may become unknown? This may be solved by having the `targets` and `labels` both replicated on all devices, then masked, then distributed. However that seemed like a lot of effort and would require the loss class to handle tensor sharding. 

CC @felipemello1 . I'm not particularly attached to this PR, just brings attention to the current issues and proposes a solution :) 

#### Changelog
*

#### Test plan

#### UX

### DIFF
--- a/recipes/full_finetune_distributed.py
+++ b/recipes/full_finetune_distributed.py
@@ -16,7 +16,7 @@ from omegaconf import DictConfig, ListConfig
 
 from torch import nn
 from torch.distributed import destroy_process_group, init_process_group
-from torch.distributed._tensor import DTensor
+from torch.distributed.tensor import distribute_tensor, DTensor
 from torch.distributed.tensor.parallel import parallelize_module
 from torch.optim import Optimizer
 from torchao.float8 import precompute_float8_dynamic_scale_for_fsdp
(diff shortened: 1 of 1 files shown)


pbontrager:
Thank you for getting this out @nathan-az. I've been working to fix this along with FSDP issues, and unbreaking the older loss as well. Here's my [PR](https://github.com/pytorch/torchtune/pull/2667)

nathan-az:
Brilliant, looks like you've got all bases covered @pbontrager - closing this.

Issue #2480, opened 2025-03-11
Add add_end_token to Phi tokenizers


pbontrager:
Llama3 Tokenizer was updated to support add_end_token in tokenize_messages to support correct generation https://github.com/pytorch/torchtune/pull/1494. These changes need to be made to phi 3 and 4.

Changes:
- Update tokenize_message to use add_start_tokens and add_end_tokens like in https://github.com/pytorch/torchtune/pull/1494
- Replace add_eos with add_end_tokens and update tokenize_messages as in https://github.com/pytorch/torchtune/pull/1494
- In call update tokens, mask = self.tokenize_messages(messages) to tokens, mask = self.tokenize_messages(messages, add_end_tokens=not inference)

Saurabh750:
Hi @pbontrager , I am new to OS and torchtune. I came across this issue and I think this will be a good start for me.
If this issue needs attention, can I please work on it?
I have went through #1494 and also read your comments on #2479. If you suggest, I can first look into #2479
Thanks!

ysurs:
Hey , I am also looking forward to making a contribution to torchtune and I am just starting out on OSS, can you recommend an issue to work on, does llama2 require these changes ?

Any suggestion would be helpful, thanks for your time.

pbontrager:
@ysurs take a look at the missing checkmarks on this one https://github.com/pytorch/torchtune/issues/1810

ysurs:
@pbontrager , thanks for the suggestion

PR #2679, opened 2025-05-06
Add add_end_token to Phi tokenizers #2480

Saurabh750:
#### Context
- [ ] add a new feature
- [x] fix a bug
- [ ] other (please add here)

#### Changelog
1. tokenize_message() method is not defined for phi3, Instead encode method is used directly. I have not used add_start_tokens as the model card definition of tokenizer is different for phi 3 & 4 compared to Llama3.

2. Used add_end_tokens in tokenize_messages(). Referred this: https://github.com/pytorch/torchtune/pull/1494 as mentioned in  #2480 

#### Test plan

#### UX

### DIFF
--- a/tests/torchtune/models/phi3/test_phi3_tokenizer.py
+++ b/tests/torchtune/models/phi3/test_phi3_tokenizer.py
@@ -5,6 +5,11 @@
 # LICENSE file in the root directory of this source tree.
 
 import pytest
+import sys, os
+
+sys.path.append(os.path.dirname(os.path.abspath(__file__)))
+
+print(sys.path)
 
 from tests.common import ASSETS
 from torchtune.data import Message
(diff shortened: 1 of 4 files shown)


Saurabh750:
Hi! this is my first OS issue :)
I have tried fixing the issue for Phi3 & 4. Also updated the respective tokenizer test files for both.
I have run these files using pytest: test_phi3_tokenizer.py, test_phi4_tokenizer.py

Please let me know if anything needs to be fixed or tested further. Thanks!

krammnic:
It doesn't seem to me that these changes are reasonable...

krammnic:
Let me clarify (Mostly duplicates the comment that I've written in discord):

If we consider similar PR that was  merged for the llama3 tokenizer we have 2 important differences:

`add_start_tokens` was required, here it isn't.
The idea of the argument was different, and the reason behind this name is described in the docstring.

`add_end_token` was originally messed up with `add_end_tokens`. And it is still messed up (consider PR name vs argument name in the PR). In case if we don't have `add_start_tokens` and we have only specific token (so the real name might have been `add_end_token`) I don't see why it is reasonable to make it more confusing with such renaming.

Let's generalize:
IMO it might be relevant to the different tokenizers, but here it will only cause confusion. Furthermore, I want to highlight that we can't standardize different tokenizers in really similar style and naming by different reasons. Hence, I don't think that different arguments naming in different tokenizers is a problem.

ebsmothers (approved):

PR #2680, opened 2025-05-06
Add assert for missing parameters

yizhang82:
Summary: This will make it easier to identify which parameter is missing when loading a dictionary. Otherwise you'll get a `'None type' object has no attribute 'dtype'` error which isn't super helpful.

Differential Revision: D74202965


ebsmothers (approved):
Thanks @yizhang82 for the PR! I have hacked a very similar line into my local code too many times.. this will definitely make it easier to debug model/checkpoint mismatches. Will let CI run and then merge

Issue #2641, opened 2025-04-28
Tensor parallelism accuracy + grad norm scale issues


nathan-az:
I noticed that when comparing TP to DP with the same seed and adjusting gradient accumulation steps, there's an odd spike after the second training step.

<img width="389" alt="Image" src="https://github.com/user-attachments/assets/778d4bba-300d-4f51-844a-08c96da5eac6" />

I made a custom recipe with some additional metrics - logging the number of tokens per update step, number of correctly predicted tokens (top 1 accuracy from the logits), and accuracy. I saw a steep drop-off in accuracy at the same point.

<img width="1178" alt="Image" src="https://github.com/user-attachments/assets/445e8be9-aca2-4baa-a6ea-71d2ce75a0fb" />

I did some further testing (on cheaper hardware, hence smaller sizes), comparing two custom recipes:

1. tweak of `full_finetune_distributed` which includes the accuracy metrics. Metrics groups prefixed `acc`
2. tweak of a recipe following the stepwise pattern from https://github.com/pytorch/torchtune/pull/2543. Metrics groups prefixed `stepwise`

<img width="1076" alt="Image" src="https://github.com/user-attachments/assets/c8a574bc-2751-4976-82fb-b75f086b75cd" />

(2) is really just a slight re-formulation of the training loop which scales losses rather than using the gradient scaling. Note that although the scales of the losses are different between the original method and the step-wise method, for stepwise TP, stepwise DP, and original TP, accuracies are almost identical, while original TP is near zero.

This might indicate there is an issue with the gradient scaling. Although given quirks with my hardware I'd love if anyone could try to reproduce the results.

DP config:

<details>

```yaml
batch_size: 1
checkpointer:
  _component_: torchtune.training.FullModelHFCheckpointer
  checkpoint_files:
    filename_format: model-{}-of-{}.safetensors
    max_filename: '00030'
  model_type: LLAMA3
  output_dir: ${output_dir}
  recipe_checkpoint: null
  checkpoint_dir: inputs/model
clip_grad_norm: null
compile: true
custom_sharded_layers: []

data_parallel_shard_dim: 8
data_parallel_replicate_dim: 1
tensor_parallel_dim: 1

tensor_parallel_plan:
  _component_: torchtune.models.llama3.base_llama_tp_plan

dataset:
  _component_: torchtune.datasets.slimorca_dataset
  packed: true
  train_on_input: true

device: cuda
dtype: bf16
enable_activation_checkpointing: true
[...]

PR #2687, opened 2025-05-07
Fix grad scaling in Tensor Parallelism

nathan-az:
#### Context
- [ ] add a new feature
- [x] fix a bug
- [ ] other (please add here)

[Please link to any issues this PR addresses.](https://github.com/pytorch/torchtune/issues/2641)

#### Changelog
* Forcefully disallows using `foreach` grad scaling implementation with TP
* If not using `foreach`, check grad type, and do not use inplace `mul_` on DTensor
* Rescale grads by world size, not DP size (not sure why this is required but it aligned loss and grad norms across all distributed setups)

image

#### Test plan

#### UX

### DIFF
--- a/torchtune/training/_grad_scaler.py
+++ b/torchtune/training/_grad_scaler.py
@@ -9,6 +9,7 @@ from typing import Optional
 
 import torch
 from torch import nn, Tensor
+from torch.distributed.tensor import DTensor
 from torch.nn.utils.clip_grad import _no_grad, _tensor_or_tensors
 from torch.utils._foreach_utils import _device_has_foreach_support, _has_foreach_support
 from torchtune.utils._logging import deprecated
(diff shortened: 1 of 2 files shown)


torchtune/training/_grad_scaler.py
```diff
@@ -102,4 +103,7 @@ def _scale_grad_(
         else:
             scaler_device = scaler.to(device)
             for g in device_grads:
-                g.mul_(scaler_device)
+                if isinstance(g, DTensor):
```
joecummings, line 106: @IvanKobzarev Any idea on why this would be happening?

nathan-az:
@joecummings assuming this (bug + fix) can be repro'd on your side, should we merge this in the short term? Just conscious that TP is recommended for at least one recipe (LLaMA-4) and currently appears incorrect, and a fix in Core for the in-place mul will probably take longer. Can revert in future.

ebsmothers (approved):
Thanks @nathan-az for figuring this out! I agree we should just land this to unblock TP training. Will merge once CI is green. Also cc @IvanKobzarev since we were talking about this earlier

PR #2698, opened 2025-05-08
Configurable reward functions

SalmanMohammadi:
Please see https://github.com/joecummings/r1-zero/pull/48

### DIFF
--- a/torchtune/dev/rl/rewards.py
+++ b/torchtune/dev/rl/rewards.py
@@ -5,151 +5,214 @@
 # LICENSE file in the root directory of this source tree.
 
 import re
-from typing import List, Tuple
-from xml.etree import ElementTree as ET
+from abc import ABC, abstractmethod
+from dataclasses import dataclass, field
+from typing import Dict, List, Optional
 
 import math_verify
-
 import torch
 
-from torchtune.modules.transforms.tokenizers import ModelTokenizer
 
-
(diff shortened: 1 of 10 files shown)


torchtune/dev/rl/rewards.py
```diff
@@ -5,151 +5,214 @@
...
+        completions: List[str],
+        answers: List[str],
+    ) -> RewardOutput:
+        rewards = []
+        import math_verify
+
+        for completion, answer in zip(completions, answers):
+            gold = math_verify.parse(answer)
+            attempt = math_verify.parse(completion)
+            if math_verify.verify(gold, attempt):
```
ebsmothers, line 128: Are we back on math_verify then? I thought we had gotten rid of this one

ebsmothers (commented):
One comment, but this looks good to me. I think this may just need a merge since CI is failing

torchtune/dev/rl/rewards.py
SalmanMohammadi: I thought it landed in main. I wouldn't be opposed to removing it here

ebsmothers: Oh I guess so, I totally missed that. Personally I would like to take it out but we don't have to block this PR on it

ebsmothers (approved):
Stamping to unblock, once merge conflicts are fixed and CI is green this looks good to me

PR #2703, opened 2025-05-08
Explicitly import SFTLoss in recipes

ebsmothers:
Before this change, e.g. 

```
tune run full_finetune_single_device --config llama3_2/1B_full_single_device loss=torch.nn.CrossEntropyLoss
```

fails with the error 

```
    if isinstance(self._loss_fn, modules.loss.SFTLoss):
                                 ^^^^^^^^^^^^
AttributeError: module 'torchtune.modules' has no attribute 'loss'
```


joecummings (approved):
bing bang bong

pbontrager (approved):

PR #2707, opened 2025-05-08
move device to training

omkar-334:
#### Context
- [ ] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

Addresses #1442 

#### Changelog
*

#### Test plan

#### UX


omkar-334:
The lint tests are failing because now we're importing the function `get_world_size_and_rank` two times, one from `torchtune.training._device` and the other from `torchtune.training._distributed`, which has a deprecated warning.   
```python
@deprecated(
    msg="`get_world_size_and_rank` will move to `torchtune.utils._device` in future releases. "
    "Please use `torchtune.utils.get_world_size_and_rank` instead."
)
```   
What would be the solution for this?

omkar-334:
@pbontrager @RdoubleA can you review my PR ? Thanks!

joecummings (commented):
Thanks so much for the contribution @omkar-334!

Unfortunately, I messed up :/ While I was going through and finding issues that would be good for first time contributors to take on, I found this one quickly updated it without thinking. I believe the thinking originally was that specifically the ``batch_to_device`` function did not belong in utils since it's only a training utility. After discussing with @pbontrager , the proper fix for something like that is *not* to just move device.py to the dir `training/` but rather use something like `tensordicts` to easily move data around on devices with the standard PyTorch `to` function. This will require a more in-depth piece of work and I'll write up an Issue at a later time for it.

We really do value your time and contributions, so I'm very sorry this one didn't work out! It looks like you're interested in contributing partially to #2715, which would also be a huge help.

omkar-334:
> After discussing with @pbontrager , the proper fix for something like that is _not_ to just move device.py to the dir `training/` but rather use something like `tensordicts` to easily move data around on devices with the standard PyTorch `to` function. This will require a more in-depth piece of work and I'll write up an Issue at a later time for it.

Understood... No worries. hopefully I can contribute more and have a good understanding by then.  
Closing this PR.  

> It looks like you're interested in contributing partially to #2715, which would also be a huge help.  

Yup, I'll get started with it.   
Thanks!

Issue #2683, opened 2025-05-07
TP + Activation Offloading produces bad results


nathan-az:
<img width="918" alt="Image" src="https://github.com/user-attachments/assets/8b04123b-4fbf-4dbd-ad9b-7a036e2b2e48" />

Note in above, that with offloading + streams (default currently), loss is higher, top-1 accuracy is much lower, and grad norms are much higher.

Using activation offloading with TP appears to produce much worse results. Note that due to an existing bug with TP (https://github.com/pytorch/torchtune/issues/2641), testing was done on a [custom branch](https://github.com/nathan-az/torchtune/tree/new-pattern-optims).

Note that setting a flag and disabling `use_streams` appears to re-align loss, grad norms, and top-1 token accuracy. This may point to some synchronisation issue between the offloading streams and collective schedule in TP.

joecummings:
cc @janeyx99

janeyx99:
Who is the TP expert who has context on how TP uses streams and backward hooks?

joecummings:
> Who is the TP expert who has context on how TP uses streams and backward hooks?

I think that might be @tianyu-l?

PR #2710, opened 2025-05-09
Allow disabling activation offloading streams in full finetune recipe

nathan-az:
#### Context
- [ ] add a new feature
- [x] fix a bug
- [ ] other (please add here)

https://github.com/pytorch/torchtune/issues/2683

#### Changelog
* Allows configuring `use_streams` in activation offloading

#### Test plan

#### UX

### DIFF
--- a/recipes/full_finetune_distributed.py
+++ b/recipes/full_finetune_distributed.py
@@ -227,6 +227,16 @@ class FullFinetuneRecipeDistributed(FTRecipeInterface):
         self._enable_activation_offloading = cfg.get(
             "enable_activation_offloading", False
         )
+        self._activation_offloading_use_streams = cfg.get(
+            "activation_offloading_use_streams", True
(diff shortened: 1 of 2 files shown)


nathan-az:
@joecummings @janeyx99 Let me know if this looks like a reasonable stopgap. I'm not certain that it completely solves the issue, but is a reasonable stopgap (if anyone can repro both the instability in the issue, and that this change fixes it. I assume that if the issue is the streams, fixing it may be quite a bit higher effort.

janeyx99:
@nathan-az I think allowing a nonstreams option is good! Your inclination is right that if it is streams that’s the problem, it will be nontrivial to fix.

ebsmothers (approved):
@nathan-az thanks for opening this! I agree, this looks like a reasonable fix to unblock TP with activation offloading

PR #2712, opened 2025-05-09
Cleanup usage of optimizer / optimizer in backwards

joecummings:
#### Context
- [ ] add a new feature
- [ ] fix a bug
- [x] other (please add here)

#### Changelog
* Optimizer and OptimizerInBwdWrapper both have similar APIs and can be used as such
* Instead of having a function that sets up the lr scheduler OR returns None if there's no config value, we just only call the function if there is a config value. Duh
* Other various cleanup like checking based on isinstance instead of a variable set at the top - this is more accurate

This removes 35 LOC from our recipe. If these changes look good, I'll open an issue to migrate these to the rest of the recipes.

#### Test plan

#### UX

### DIFF
--- a/recipes/full_finetune_single_device.py
+++ b/recipes/full_finetune_single_device.py
@@ -333,7 +333,7 @@ class FullFinetuneRecipeSingleDevice(FTRecipeInterface):
         lr_scheduler_cfg = cfg.get("lr_scheduler", None)
         if lr_scheduler_cfg is not None:
             self._lr_scheduler = self._setup_lr_scheduler(
-                cfg_lr_scheduler=cfg.get("lr_scheduler", None),
+                cfg_lr_scheduler=lr_scheduler_cfg,
                 num_training_steps=self.total_epochs * self._steps_per_epoch,
                 last_epoch=self.global_step - 1,
             )
@@ -532,9 +532,6 @@ class FullFinetuneRecipeSingleDevice(FTRecipeInterface):
             # dropping last avoids shape issues with compile + flex attention
             drop_last=True,
         )
-
-        if dataloader_state_dict:
-            dataloader.load_state_dict(dataloader_state_dict)
 
         return dataloader
 
--- a/recipes/full_finetune_single_device.py
+++ b/recipes/full_finetune_single_device.py
@@ -15,6 +15,7 @@ from omegaconf import DictConfig, ListConfig
 
 from torch import nn
 from torch.optim import Optimizer
+from torch.optim.lr_scheduler import LambdaLR
 from torchdata.stateful_dataloader import StatefulDataLoader
 from torchdata.stateful_dataloader.sampler import StatefulDistributedSampler
 from torchtune import config, modules, training, utils
@@ -25,6 +26,7 @@ from torchtune.modules.loss import SFTLoss
 from torchtune.recipe_interfaces import FTRecipeInterface
 from torchtune.training import DummyProfiler, PROFILER_KEY
 from torchtune.training.lr_schedulers import get_lr
+from torchtune.training.memory import OptimizerInBackwardWrapper
 
 from tqdm import tqdm
 
@@ -286,7 +288,6 @@ class FullFinetuneRecipeSingleDevice(FTRecipeInterface):
             ),
         )
 
-        # initialize loss
         self._loss_fn = config.instantiate(cfg.loss)
         if isinstance(self._loss_fn, SFTLoss):
             self._loss_fn.set_model_output(self._model)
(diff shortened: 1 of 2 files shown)


recipes/full_finetune_single_device.py
```diff
@@ -433,73 +438,44 @@ def _setup_optimizer(
...
+        return optimizer
 
     def _setup_lr_scheduler(
         self,
-        cfg_lr_scheduler: Optional[DictConfig],
+        cfg_lr_scheduler: DictConfig,
         num_training_steps: int,
         last_epoch: int,
-    ) -> Optional[Optimizer]:
+    ) -> LambdaLR:
```
joecummings, line 458: This was embarassing.

recipes/full_finetune_single_device.py
```diff
@@ -611,12 +587,7 @@ def train(self) -> None:
         The core training loop. Supports training on subsets of the dataset using the
         ``max_steps_per_epoch``.
         """
-        if self._compile:
-            self._logger.info(
-                "NOTE: torch.compile is enabled and model is compiled in first forward. Expect a relatively slow first iteration."
-            )
```
ebsmothers, line 617: Removing this is not strictly related, right? I'm not opposed, but we should at least log if we're compiling somewhere (I think we already do it in `compile_model`?)

recipes/full_finetune_single_device.py
```diff
@@ -433,73 +438,44 @@ def _setup_optimizer(
...
             cfg_lr_scheduler,
             optimizer,
             num_training_steps=num_training_steps,
             last_epoch=last_epoch,
         )
 
-        if self._optimizer_in_bwd:
-            # Modify the scheduler for optimizer_in_bwd case
-            self._optim_ckpt_wrapper.set_lr_scheduler(lr_scheduler)
+        if isinstance(self._optimizer, OptimizerInBackwardWrapper):
```
ebsmothers, line 477: Personally I like this like 5% more but I know @pbontrager hates `isinstance` checks

recipes/full_finetune_single_device.py
```diff
@@ -328,12 +329,16 @@ def setup(self, cfg: DictConfig) -> None:
...
-        self._lr_scheduler = self._setup_lr_scheduler(
-            cfg_lr_scheduler=cfg.get("lr_scheduler", None),
-            num_training_steps=self.total_epochs * self._steps_per_epoch,
-            last_epoch=self.global_step - 1,
-        )
+        # Setup lr scheduler if a component is included in the config
+        lr_scheduler_cfg = cfg.get("lr_scheduler", None)
+        if lr_scheduler_cfg is not None:
+            self._lr_scheduler = self._setup_lr_scheduler(
+                cfg_lr_scheduler=cfg.get("lr_scheduler", None),
```
ebsmothers, line 336: Why are we repeating the same thing from L333?

recipes/full_finetune_single_device.py
```diff
@@ -557,15 +533,18 @@ def _setup_data(
             drop_last=True,
         )
 
+        if dataloader_state_dict:
+            dataloader.load_state_dict(dataloader_state_dict)
```
ebsmothers, line 537: ❓

recipes/full_finetune_single_device.py
```diff
@@ -557,15 +533,18 @@ def _setup_data(
...
         return dataloader
 
     def save_checkpoint(self, epoch: int) -> None:
         """
         Save state dict to file. The recipe save_checkpoint method is responsible for
         correctly creating the checkpoint dict and passing to the checkpointer.
         """
-        ckpt_dict = {training.MODEL_KEY: self._model.state_dict()}
-        # if training is in-progress, checkpoint the optimizer state as well
+        ckpt_dict: dict[str, Any] = {training.MODEL_KEY: self._model.state_dict()}
```
ebsmothers, line 546: What's with the typing here? Do we need it for some reason?

recipes/full_finetune_single_device.py
ebsmothers: We should really generalize to generic `LRScheduler`..

recipes/full_finetune_single_device.py
joecummings: We already do this in compile model yeah.

recipes/full_finetune_single_device.py
joecummings: Linter was complaining.

joecummings: But no, not strictly necessary.

recipes/full_finetune_single_device.py
joecummings: Talk to me after we add any other LR scheduler besides cosine

recipes/full_finetune_single_device.py
joecummings: @pbontrager just told me it's fine and that I'm the best torchtune engineer.

recipes/full_finetune_single_device.py
joecummings: Wait nvm, I remember what happened here. Someone reverted my changes to all these files which technically is fine b/c we only save at epoch boundaries. But it's very weird and bad. I'll remove for now.

recipes/full_finetune_single_device.py
ebsmothers: Oh weird. Btw you wanna open a "good first issue" we should consolidate typehints on the non-typing versions

pbontrager (approved):
LGTM

Issue #2416, opened 2025-02-20
【not bug】how to distill qwen2.5 7b to qwen2.5 3b.


whk6688:
i want to distill qwen2.5 7b to qwen2.5 3b. but they have different vocab size. now i choice to  crop tensor (teacher_logits[:,:151936]).  is there better way to solve the issue?

thanks

ebsmothers:
Hi @whk6688 thanks for creating the issue. In general it's recommended to distill model with only the same vocab sizes, however in this case I believe you may be able to get away with the approach you're suggesting. You can see the following excerpt from [this comment](https://github.com/QwenLM/Qwen2.5/issues/466#issuecomment-2146759706):

> You can always resize the embedding, as the index over 151646 is meaningless.

So based on that I think your proposed approach should work.

PR #2714, opened 2025-05-09
add token resize embedding

iamzainhuda:
Redone from #2542 to fix git history

This PR introduces a resize_token_embeddings method which allows users to add additional tokens to be trained as part of fine tuning. It supports both single and multi device setups. The function can be easily extended to support additional initialization strategies.

For the distributed case, we unshard() the token embedding and output layer before passing into resize_token_embedding(). This way we can use the same logic as the single device case as the parameters are gathered on to the rank. Once complete, we fully_shard() the new token embedding and output layers utilizing the same mesh used for the initial sharding. This is the recommended way to resize FSDP2 sharded parameters. Thanks to @weifengpy for his guidance on the FSDP2 sharding mechanism.

The usage is,

On single device and distributed setups:

```python
model = setup_model(...)
tokenizer = setup_tokenizer(...)
resize_token_embeddings(model, tokenizer.vocab_size)
```

When using a recipe, (right now enabled for `full_finetune_distributed` and `full_finetune_single_device`), add `resize_token_embeddings = True` in YAML.
Example,
```yaml
...
max_steps_per_epoch: null
clip_grad_norm: null
compile: False  # torch.compile the model + loss, True increases speed + decreases memory
optimizer_in_bwd: False  # True saves memory. Requires gradient_accumulation_steps=1
resize_token_embeddings: True
...
````

#### Follow up tasks:
- [ ] additional embedding weight init methods
- [ ] loading checkpoint from resized model requires model to be created with new vocab size
- [ ] FusionEmbedding resizing as part of TransformerDecoder
 
#### Context
- [x] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

#### Changelog
* Added resize_token_embedding() utility that allows users to flexibly change the size of the token embeddings and corresponding output projection.

#### Test plan

Unit Testing:

Unit tests cover both single device and distributed cases, we check the new nn.Embedding size as well as the mean values of the new tokens to ensure they were init correctly. The E2E case covers testing a training run. We can see training completes successfully in the resized model that is sharded across 8 devices.

Without resizing, we can see special token init values on LLama 3 8B base model are very small, taking a look at token `<|start_header_id|> = 128006`
image

With resizing, we can see special tokens are init to the mean of 128k tokens on LLama 8B base:
image

##### E2E Test

Resizing the last 256 tokens in Llama 3 tokenizer to mean embedding of previous 120k tokens. Added an additional 4 tokens to resize the entire layer. We don't see significant difference in loss (run for 2 epochs, 20 steps each), but we show with the resized embedding in distributed setup we're able to finish training with at least equivalent loss.

image

vs. baseline
image

Running,
```
tune run --nproc_per_node 8 full_finetune_distributed --config recipes/configs/llama3/8B_full.yaml max_steps_per_epoch=20 epochs=2 checkpointer.checkpoint_dir=/tmp/Meta-Llama-3-8B
 ```

#### UX

View here:
https://docs-preview.pytorch.org/pytorch/torchtune/2714/generated/torchtune.modules.common_utils.resize_token_embeddings.html#torchtune.modules.common_utils.resize_token_embeddings


torchtune/modules/common_utils.py
```diff
@@ -442,3 +448,138 @@ def delete_kv_caches(model: nn.Module):
...
+        num_embeddings (int): The desired number of embeddings in the resized
+            embedding layer and output projection layer.
+
+    Returns:
+        None: The function modifies the `model` in-place.
+
+    Raises:
+        AssertionError: When model not of type ``TransformerDecoder`` is passed in
+            and is a resize is needed. User should pass in `TransformerDecoder` part
+            of the model only.
```
ebsmothers, line 483: Should update this comment to reflect the latest changes (i.e. now the AssertionError will only be raised if we are trying to resize a `FusionEmbedding`)

torchtune/modules/common_utils.py
```diff
@@ -442,3 +448,138 @@ def delete_kv_caches(model: nn.Module):
...
+    if not need_resize:
+        _log.info("Embedding resize not needed, skipping")
+        return
+
+    # Local import to break a circular dependency with TransformerDecoder module
+    from torchtune.modules.model_fusion import (
+        DeepFusionModel,
+        EarlyFusionModel,
+        FusionEmbedding,
+    )
```
ebsmothers, line 495: Wait what's the dependency chain that necessitates this? Fusion layers import TransformerDecoder, but I didn't think we were actually importing anything from common_utils for TransformerDecoder. Would moving this into a separate file (e.g. `embedding_utils.py`) resolve the circular dependency?

torchtune/modules/common_utils.py
```diff
@@ -442,3 +448,138 @@ def delete_kv_caches(model: nn.Module):
...
+        if num_embeddings > output_layer.out_features:
+            mean_weight = new_output.weight.data[:n_out].mean(dim=0)
+            new_output.weight.data[n_out:] = mean_weight.unsqueeze(0)
+
+        new_output.weight.requires_grad_(output_layer.weight.requires_grad)
+        model.output = new_output
+    else:
+        _log.warning(
+            f"Output layer is not tied and not a recognized Linear layer for resizing. Type: {type(output_layer)}. Skipping."
+        )
```
Andrei-Aksionov, line 585: Since there is some level of code duplication, may I propose a slight change?
```python
@torch.no_grad
def _resize_token_embeddings_helper(model: nn.Module, num_embeddings: int) -> None:
    """
    Helper utility that resizes the token embeddings and output layer to `num_embeddings`.
    """

    def _copy_weights(new_weights: torch.Tensor, old_weights: torch.Tensor) -> None:
        n = min(old_weights.shape[0], new_weights.shape[0])
        new_weights[:n] = old_weights[:n]
        if new_weights.shape[0] > old_weights.shape[0]:
            new_weights[n:] = old_weights[:n].mean(dim=0, keepdim=True)
        new_weights.requires_grad_(old_weights.requires_grad)

    old_embeddings = model.tok_embeddings
    _log.info(f"Resizing token embeddings from {old_embeddings.num_embeddings} to {num_embeddings}")

    model.tok_embeddings = nn.Embedding(
        num_embeddings=num_embeddings,
        embedding_dim=old_embeddings.embedding_dim,
        padding_idx=old_embeddings.padding_idx,
        device=old_embeddings.weight.device,
        dtype=old_embeddings.weight.dtype,
    )
    _copy_weights(model.tok_embeddings.weight.data, old_embeddings.weight.data)
    del old_embeddings  # free memory early

    if isinstance(model.output, TiedLinear):
        model.output.tied_module = model.tok_embeddings
    elif isinstance(model.output, nn.Linear):
        old_output = model.output
        model.output = nn.Linear(
            in_features=old_output.in_features,
            out_features=num_embeddings,
            bias=False,
            device=model.tok_embeddings.weight.device,
            dtype=model.tok_embeddings.weight.dtype,
        )
        _copy_weights(model.output.weight.data, old_output.weight.data)
        del old_output  # free memory early
    else:
        _log.warning(f"Output layer is not tied and not a recognized Linear layer for resizing.: {type(model.output)}. Skipping.")
 ```
 ---
 
 ```bash
 pytest tests/torchtune/modules/test_common_utils.py::TestResizeTokenEmbeddings
 ```
 
 showed all green

File: torchtune/modules/common_utils.py [hunk 4]
```diff
@@ -442,3 +448,138 @@ def delete_kv_caches(model: nn.Module):
...
+            fully_shard(model.output, mesh=mesh)
+
+
+def _need_to_resize(model: nn.Module, num_embeddings: int) -> bool:
+    try:
+        current_vocab = (
+            model.tok_embeddings.num_embeddings
+            if isinstance(model, TransformerDecoder)
+            else model.decoder.tok_embeddings.num_embeddings
+        )
```
Andrei-Aksionov, line 525:
```suggestion
                current_vocab = getattr(model, "decoder", model).tok_embeddings.num_embeddings
```

Just a proposal, perhaps for some it's more difficult to read.

Andrei-Aksionov:
I have a dummy question: the logic in this PR allows both increase and decrease of the embedding/output.
It's clear to me why we might want to increase the size.
But why someone might want to decrease the size, while keeping remaining trained weights? 
I am just curious. Perhaps there is a legit use case that I'm not aware of.

torchtune/modules/common_utils.py
iamzainhuda: Thanks! This is a great suggestion, updated the PR. Thanks for running the tests too

torchtune/modules/common_utils.py
iamzainhuda: Thanks - this was an old commit changed this function entirely, should be clearer

torchtune/modules/common_utils.py
iamzainhuda: Moved function to embedding_utils.py and force the import to be last in ``modules.__init__.py`` this fixes the circular import issues

Andrei-Aksionov:
>> I have a dummy question: the logic in this PR allows both increase and decrease of the embedding/output. It's clear to me why we might want to increase the size. But why someone might want to decrease the size, while keeping remaining trained weights? I am just curious. Perhaps there is a legit use case that I'm not aware of.

> For example, in knowledge distillation, we may want to crop the embedding size: https://github.com/pytorch/torchtune/issues/2416

Thank you for the explanation.

ebsmothers (approved):
Thanks @iamzainhuda for adding this!

Issue #2706, opened 2025-05-08
Integrate Hugging Face tokenizers with torchtune


ebsmothers:
## The problem

torchtune provides its own set of abstractions for tokenization and application of chat templates (details given below). Most popular open-source models are uploaded to the Hugging Face hub using the format from transformers. This typically includes at least: `tokenizer.json`, `tokenizer_config.json`, and `generation_config.json`. Details on each of these files and their contents are given below.

Consider [Qwen3-4B](https://huggingface.co/Qwen/Qwen3-4B/tree/main) as an example. Note that in this case, the hub contains `merges.txt` and `vocab.json` files, which can be used to build the torchtune tokenizer directly. But there are other cases, e.g. [DeepSeek's Qwen 7B distill](https://huggingface.co/deepseek-ai/DeepSeek-R1-Distill-Qwen-7B/tree/main), where only the `tokenizer.json` file is provided. This means that **in order to enable these models in general, we need an interface allowing us to build a torchtune tokenizer from an arbitrary Hugging Face model**.

### Tokenization in torchtune

In torchtune, we provide two levels of abstraction for logic related to tokenization and prompt/chat templating. First:

[BaseTokenizer](https://github.com/pytorch/torchtune/blob/1392863c6d1c18b8355ad7c4d13ecd7ce4d2a476/torchtune/modules/transforms/tokenizers/_utils.py#L18). Any base tokenizer should define `encode` and `decode` methods with the following signatures:

```
class MyBaseTokenizer(BaseTokenizer):

	def encode(self, token_ids: str, **kwargs) -> List[int]:
		...

	def decode(self, token_ids: List[int], **kwargs) -> str:
		...
```

Common examples of `BaseTokenizer`s include [TikTokenBaseTokenizer](https://github.com/pytorch/torchtune/blob/1392863c6d1c18b8355ad7c4d13ecd7ce4d2a476/torchtune/modules/transforms/tokenizers/_tiktoken.py#L18) and [SentencePiecBaseTokenizer](https://github.com/pytorch/torchtune/blob/1392863c6d1c18b8355ad7c4d13ecd7ce4d2a476/torchtune/modules/transforms/tokenizers/_sentencepiece.py#L15). Typically a new `BaseTokenizer` only needs to be implemented if the underlying tokenization algorithm itself changes. For changes that are specific to a given model family, prompting method, etc., we can instead rely on...

[...]

PR #2723, opened 2025-05-12
HuggingFaceModelTokenizer

krammnic:
#### Context
- [X] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

#### Changelog
* #2706 

#### Test plan

#### UX

Basically, this is a first pass (I'm thinking about how to add masking), but jinja render works surprisingly well.


torchtune/modules/transforms/tokenizers/_hf_tokenizer.py
```diff
@@ -146,3 +149,108 @@ def decode(self, token_ids: List[int]) -> str:
...
+    def tokenize_messages(
+        self,
+        messages: List[Message],
+        add_eos: bool = True,
+    ) -> Tuple[List[int], List[bool]]:
+        # This part is extremely hacky, but we need to handle case where we have variable access with jinja
+        special_tokens_mapping = {}
+        for token in self.special_tokens:
+            special_tokens_mapping[token] = self.base_tokenizer.encode(token)
+        rendered_template = self.template.render(
```
ebsmothers, line 239: Wow this actually wound up being quite easy lol

torchtune/modules/transforms/tokenizers/_hf_tokenizer.py
```diff
@@ -146,3 +149,108 @@ def decode(self, token_ids: List[int]) -> str:
...
+
+        if content:
+            special_tokens.add(content)
+
+    for token_info in config.get("added_tokens_decoder", {}).values():
+        if token_info.get("special", False):
+            if content := token_info.get("content"):
+                special_tokens.add(content)
+
+    # We sort lexicographically in order to get real tokens after all <|dummy_x|>
```
ebsmothers, line 191: Sorry I don't fully understand this comment. I assume this is referring to reserved special tokens? If so, why is string sort the thing to use here?

torchtune/modules/transforms/tokenizers/_hf_tokenizer.py
```diff
@@ -146,3 +149,108 @@ def decode(self, token_ids: List[int]) -> str:
...
+        tokenizer_json_path: str,
+        *,
+        tokenizer_config_json_path: Optional[str] = None,
+        generation_config_path: Optional[str] = None,
+    ):
+        self.base_tokenizer = HuggingFaceBaseTokenizer(
+            tokenizer_json_path=tokenizer_json_path,
+            tokenizer_config_json_path=tokenizer_config_json_path,
+            generation_config_path=generation_config_path,
+        )
```
ebsmothers, line 207: I know @joecummings had some thoughts on whether we should use generic `base_tokenizer` instead of constraining to use `HuggingFaceBaseTokenizer`. I suspect the latter is better for making sure everything works together, but I know at least `Qwen2Tokenizer` still relies on the merges + vocab files instead of the `tokenizer.json` file (I alluded to this at the very bottom of #2706). So we should figure out if this will work for that case

torchtune/modules/transforms/tokenizers/_hf_tokenizer.py
```diff
@@ -146,3 +149,108 @@ def decode(self, token_ids: List[int]) -> str:
...
+        # This part is extremely hacky, but we need to handle case where we have variable access with jinja
+        special_tokens_mapping = {}
+        for token in self.special_tokens:
+            special_tokens_mapping[token] = self.base_tokenizer.encode(token)
+        rendered_template = self.template.render(
+            messages=[
+                {"role": m.role, "content": m.content[0]["content"]} for m in messages
+            ],
+            add_generation_prompt=add_eos,
+            **special_tokens_mapping,  # We assume that the naming is consitent
```
ebsmothers, line 244: Yeah I think this should be a reasonable assumption (as long as we are also getting the special_tokens from the same place as the template)

ebsmothers (commented):
Thanks @krammnic for taking this one on! This will be huge for lowering the barrier to onboard new models. Let's definitely make sure to add unit tests for this one. (You can likely create some dummy `tokenizer_config.json` files and check them directly into the repo, since they should be pretty small.)

torchtune/modules/transforms/tokenizers/_hf_tokenizer.py
krammnic: Unfortunately, tool calling will still be quite tricky

divya-inflection: @krammnic Other than the lack of tool calls in tt Message class, is there any other reasons behind why tool calling will be tricky?

krammnic: Probably, no

torchtune/modules/transforms/tokenizers/_hf_tokenizer.py
```diff
@@ -146,3 +150,161 @@ def decode(self, token_ids: List[int]) -> str:
...
+
+        if add_eos and self.base_tokenizer.eos_id is not None:
+            tokenized_messages.append(self.base_tokenizer.eos_id)
+            mask.append(False)
+
+        # Finally, truncate if necessary
+        tokenized_messages = truncate(
+            tokens=tokenized_messages,
+            max_seq_len=max_seq_len,
+            eos_id=None,
```
ebsmothers, line 299: Should we also be adding eos here?

torchtune/modules/transforms/tokenizers/_hf_tokenizer.py
```diff
@@ -146,3 +150,161 @@ def decode(self, token_ids: List[int]) -> str:
...
+
+        tokenized_messages = []
+        mask = []
+        previous_tokens = []
+
+        for i, message in enumerate(messages):
+            current_messages = [
+                {"role": m.role, "content": m.content[0]["content"]}
+                for m in messages[: i + 1]
+            ]
```
ebsmothers, line 270: This seems a bit strange to me.. in my mind we should either be able to (a) render/tokenize all messages in one shot, or (b) loop over messages one at a time, render, tokenize, and concat. Why do we need to do this "cumulative tokenization"? (Is it because of the difficulties you mentioned with masking? If so I wonder whether there is an alternative)

ebsmothers (commented):
Thanks for your patience! Left a handful of comments. Personally I would just add a unit test now, it'll make it easier to reason about things and help validate that this is giving the expected results.

torchtune/modules/transforms/tokenizers/_hf_tokenizer.py
```diff
@@ -146,3 +150,161 @@ def decode(self, token_ids: List[int]) -> str:
...
+
+            rendered = self.template.render(
+                messages=current_messages,
+                add_generation_prompt=add_eos if i == len(messages) - 1 else False,
+                **special_tokens_mapping,  # We assume that the naming is consistent
+                **self.top_level_variables,
```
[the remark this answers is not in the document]
krammnic: Yes, in some configs the special token might be `bos_id`, for instance, but the `bos` is used in the chat_template, which is redefined as the top-level variable. Generally, we know that there is a possibility to pass some extra variables. Therefore, it is better to prevent some errors here with such trick.

torchtune/modules/transforms/tokenizers/_hf_tokenizer.py
```diff
@@ -146,3 +150,161 @@ def decode(self, token_ids: List[int]) -> str:
...
+
+        for i, message in enumerate(messages):
+            current_messages = [
+                {"role": m.role, "content": m.content[0]["content"]}
+                for m in messages[: i + 1]
+            ]
+
+            rendered = self.template.render(
+                messages=current_messages,
+                add_generation_prompt=add_eos if i == len(messages) - 1 else False,
```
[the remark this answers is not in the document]
krammnic: Good point!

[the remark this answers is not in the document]
krammnic: No, bad point, we should have this argument, in another case we will not be able to render.

torchtune/modules/transforms/tokenizers/_hf_tokenizer.py
```diff
@@ -146,3 +150,161 @@ def decode(self, token_ids: List[int]) -> str:
...
+
+            delta = current_tokens[len(previous_tokens) :]
+            previous_tokens = current_tokens
+
+            if message.masked:
+                tokenized_messages.extend([True] * len(delta))
+            else:
+                tokenized_messages.extend(delta)
+
+            mask.extend([message.masked] * len(delta))
```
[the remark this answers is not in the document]
krammnic: Oops

torchtune/modules/transforms/tokenizers/_hf_tokenizer.py
krammnic: We should

torchtune/modules/transforms/tokenizers/_hf_tokenizer.py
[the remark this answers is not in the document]
ebsmothers: Ah I think you're right, it is a bad point. I think I may have misread this line.. my point was that we should only add the generation prompt during inference, but I see that `add_eos` is basically a proxy for `inference`

krammnic:
@ebsmothers @joecummings fixed CI

tests/torchtune/modules/transforms/tokenizers/test_hf_model_tokenizer.py
```diff
@@ -0,0 +1,106 @@
...
+# LICENSE file in the root directory of this source tree.
+
+import pytest
+from tests.common import ASSETS
+from torchtune.data import Message
+from torchtune.modules.transforms.tokenizers import HuggingFaceModelTokenizer
+
+TOKENIZER_CONFIG_PATH = ASSETS / "tokenizer_config_gemma.json"
+GENERATION_CONFIG_PATH = ASSETS / "generation_config_gemma.json"
+TOKENIZER_PATH = ASSETS / "tokenizer_gemma_croped.json"
```
ebsmothers: nit (should also update the filename)
```suggestion
TOKENIZER_PATH = ASSETS / "tokenizer_gemma_cropped.json"
```

ebsmothers: Also looks like this one is quite large (over 8 MB)? Is there any way we can make a smaller version (e.g. most of our other tokenizer assets are  < 1 MB).

tests/torchtune/modules/transforms/tokenizers/test_hf_model_tokenizer.py
```diff
@@ -0,0 +1,106 @@
...
+            106,
+            1645,
+            108,
+            5049,
+            15827,
+            235336,
+            107,
+            108,
+        ]
+        assert mask[:-4] == [False] * 23
```
ebsmothers: Noob q: why do we only test up to -4?

torchtune/modules/transforms/tokenizers/_hf_tokenizer.py
```diff
@@ -34,8 +39,8 @@ def __init__(
         self,
         tokenizer_json_path: str,
         *,
-        tokenizer_config_json_path: Optional[str] = None,
-        generation_config_path: Optional[str] = None,
+        tokenizer_config_json_path: str | None = None,
+        generation_config_path: str | None = None,
```
ebsmothers: Unfortunately we still run Python 3.9 in our CI so the jobs will fail with this 😕 . You can just use Union for now until we upgrade

krammnic: Yeah, I've noticed this let me update the type hints then

tests/torchtune/modules/transforms/tokenizers/test_hf_model_tokenizer.py
krammnic: Yeah, let me build such

tests/torchtune/modules/transforms/tokenizers/test_hf_model_tokenizer.py
krammnic: It might be a bug, our tokenizer version adds extra "model" role at the and. Why have I done like this? This behavior seems to be desired by the chat template, but transformers do not add it. Will investigate more and update unit test

krammnic:
"Extra <|begin_of_text|> at the beginning" 

Not fully understand, it seems to me that we've discussed that this behavior is kind of expected for the gemma3 tokenizer and it is equivalent to the transformers version

ebsmothers:
> "Extra <|begin_of_text|> at the beginning"
> 
> Not fully understand, it seems to me that we've discussed that this behavior is kind of expected for the gemma3 tokenizer and it is equivalent to the transformers version

Oh my script was for Llama3, I see it there as well. Let me dig in a bit further though

torchtune/modules/transforms/tokenizers/_hf_tokenizer.py
```diff
@@ -146,3 +151,158 @@ def decode(self, token_ids: list[int]) -> str:
...
+                content = token.get("content")
+
+            if content:
+                special_tokens.add(content)
+
+    for token in config.get("additional_special_tokens", []):
+        if isinstance(token, str):
+            content = token
+        else:
+            content = token.get("content")
```
felipemello1:
```suggestion
            content = token.get("content", None)
```

torchtune/modules/transforms/tokenizers/_hf_tokenizer.py
```diff
@@ -146,3 +151,158 @@ def decode(self, token_ids: list[int]) -> str:
...
+
+    standard_keys = [
+        "bos_token",
+        "eos_token",
+        "pad_token",
+        "unk_token",
+        "sep_token",
+        "cls_token",
+        "mask_token",
+    ]
```
felipemello1, line 169: is there a list of special tokens in the config, so we dont have to hardcode this

torchtune/modules/transforms/tokenizers/_hf_tokenizer.py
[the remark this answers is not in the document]
felipemello1: I am a bit confused. Does render expect as inputs an arbritary set of arguments in the format: "token_str": "token_id"? I need to look at the api, but dropping the question here for now

tests/torchtune/modules/transforms/tokenizers/test_hf_model_tokenizer.py
```diff
@@ -0,0 +1,106 @@
...
+            ),
+            Message(
+                role="user",
+                content="whatsup?",
+                masked=False,
+            ),
+        ]
+
+    def test_no_mask(self, model_tokenizer, messages):
+        tokens, mask = model_tokenizer.tokenize_messages(messages)
```
felipemello1: can you please add a short docstring explaining what this is testing?

tests/torchtune/modules/transforms/tokenizers/test_hf_model_tokenizer.py
```diff
@@ -0,0 +1,106 @@
...
+from tests.common import ASSETS
+from torchtune.data import Message
+from torchtune.modules.transforms.tokenizers import HuggingFaceModelTokenizer
+
+TOKENIZER_CONFIG_PATH = ASSETS / "tokenizer_config_gemma.json"
+GENERATION_CONFIG_PATH = ASSETS / "generation_config_gemma.json"
+TOKENIZER_PATH = ASSETS / "tokenizer_gemma_cropped.json"
+
+
+class TestHuggingFaceModelTokenizer:
```
felipemello1: Could we add something like:

```assert text == decode(encode("text"))```

prob some other good ideas in other unit test files.

For the output of the tests, ideally it should be generated with the current implementation, so we are sure that it matches the new one. Maybe create a script to generate the ground truths and link the gist to the unit test file?

File: torchtune/modules/transforms/tokenizers/_hf_tokenizer.py [hunk 17]
```diff
@@ -146,3 +151,158 @@ def decode(self, token_ids: list[int]) -> str:
...
+
+        self.special_tokens = _infer_special_tokens_from_hf_config(config)
+        self.top_level_variables = self.extract_top_level_variables(config)
+
+        _env = jinja2.Environment(undefined=StrictUndefined)
+
+        # It is used sometimes in HF chat_templates
+        _env.globals["raise_exception"] = self._raise_helper
+
+        self.template = _env.from_string(config["chat_template"])
```
ebsmothers, line 243: There are cases where this field is not present, right? E.g. for google/gemma-2-2b (it's only present in the instruct-tuned version). I think for cases like these we should raise an explicit error here.

torchtune/modules/transforms/tokenizers/_hf_tokenizer.py
```diff
@@ -146,3 +151,158 @@ def decode(self, token_ids: list[int]) -> str:
...
+            ]
+
+            rendered = self.template.render(
+                messages=current_messages,
+                add_generation_prompt=add_eos if i == len(messages) - 1 else False,
+                **self.special_tokens_mapping,  # We assume that the naming is consistent
+                **self.top_level_variables,
+            )
+
+            current_tokens = self.base_tokenizer.encode(rendered, add_eos=False)
```
ebsmothers, line 283: Regarding the duplicate bos tokens, I suspect we just need to add an additional case [here](https://github.com/pytorch/torchtune/blob/20bdf1086e947fd010dd79faf1c78600062cb162/torchtune/modules/transforms/tokenizers/_hf_tokenizer.py#L132-L133) to deal with the fact that we are often passing a string like `"<bos>..."` to the base_tokenizer's encode method

ebsmothers:
In addition to @felipemello1's comments, let's sanity check that we get the same results with this as we get with HF's apply_chat_template. Similar to what I shared yesterday, if we can show a script like [this](https://gist.github.com/ebsmothers/7f82e9b7867c310ce706efb9aee16afd) gives the same results for a couple of models this should be good to go. (At least for this script, I get duplicate BOS at the beginning, which should be resolvable via my comment. I also get a newline in the HF one which ours does not have, not sure what that's about..)

Separately, please make sure to use a smaller version of tokenizer.json in your unit test when you get a chance.

krammnic:
@ebsmothers I don't like the fact that CI failed only for 3.9 and with:

`ValueError: Unable to load checkpoint from /tmp/pytest-of-ec2-user/pytest-0/test_training_state_on_resume_14/recipe_state/recipe_state.pt. `

It might be some separate issue....

salmanmohammadi:
@krammnic Did you get a chance to reduce the file size and also minify the tokenizer test assets? The diff should just be one line for those files.

krammnic:
> @krammnic Did you get a chance to reduce the file size and also minify the tokenizer test assets? The diff should just be one line for those files.

Have done it briefly, but reduced on 2 times. I think it is sufficient (under 5 mb)

krammnic:
I don't think that we've minified any other assets actually

ebsmothers (approved):
@krammnic thank you for sticking with this one! This will be a huge help towards getting new models in faster.

ebsmothers (approved):
lol i accidentally closed the pr. time to log off for the day

torchtune/modules/transforms/tokenizers/_hf_tokenizer.py
nathan-az: I know this is merged, but LLaMA-4 looks like it has a separate file - `chat_template.jinja` - rather than containing it in the config. This might be another pattern that needs supporting.

torchtune/modules/transforms/tokenizers/_hf_tokenizer.py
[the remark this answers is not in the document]
pocca2048: @krammnic What is the relationship between `add_eos` and `add_generation_prompt`?
Judging by the names, it seems like `add_eos` should be set to True and `add_generation_prompt` to False during training.

PR #2736, opened 2025-05-15
validate checkpoint is consistent with  meta_to_tune flag

lijia19:
Summary:
add a flag that do validation when load_checkpoint passed unexpected meta_to_tune flag.
i.e. if this flag is true but checkpoint is not in meta format, or
if this flag is flase but checkpoint is in meta format
Do validation early to avoid unexpected error later.

Differential Revision: D74784778

### DIFF
--- a/torchtune/models/convert_weights.py
+++ b/torchtune/models/convert_weights.py
@@ -73,6 +73,22 @@ def get_mapped_key(key: str, mapping_dict: Dict[str, str]) -> str:
     return new_key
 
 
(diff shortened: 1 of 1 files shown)


ebsmothers (changes requested):
I don't think we should land this in OSS torchtune. If I understand correctly this is only needed for internal checkpointer implementations? In torchtune we have a specific class for Meta format checkpoints, so adding this functionality can cause confusion. Also it will just be zombie code. Personally I would recommend to implement this inside of the internal integration layer instead

ebsmothers:
Closing to reduce backlog of PRs

Issue #2734, opened 2025-05-14
Pre-commit hooks for PEP 585 compliance


omkar-334:
### Precursors
Related Issue - #2715   
PR #2720 refactors the entire codebase to make it PEP585 compliant, which is  using the built-in syntax instead of `typing `module for the datatypes.  

```python
tuple # typing.Tuple  
list # typing.List  
dict # typing.Dict  
set # typing.Set  
```

### Goal
Add a pre-commit hook to enforce PEP 585 usage (check if typing.List, typing.Dict, typing.Set, typing.Tuple are imported).  
Import methods can very, for example - 
1. 
```python
from typing import List, Dict
```
2. 
```python
import typing
typing.List
```  
3.  
```python
from typing import (
    List,
    Dict
)
typing.List
```  
I don't think `flake8` and `flakepyi` enforce this by default. Possible approach would be to add a custom pre-commit hook using Regex/grep or a custom flake8 plugin.

PR #2720, opened 2025-05-12
refactor: migrate type hints to PEP 585

omkar-334:
#### Context
What is the purpose of this PR? Is it to
- [ ] add a new feature
- [ ] fix a bug
- [ ] update tests and/or documentation
- [.] other (refactor)

Please link to any issues this PR addresses.

#### Changelog
What are the changes made in this PR?
Refactored type hints to pep585

#### Test plan
Please make sure to do each of the following if applicable to your PR. If you're unsure about any one of these just ask and we will happily help. We also have a [contributing page](https://github.com/pytorch/torchtune/blob/main/CONTRIBUTING.md) for some guidance on contributing.

- [ .] run pre-commit hooks and linters (make sure you've first installed via `pre-commit install`)
- [ ] add [unit tests](https://github.com/pytorch/torchtune/tree/main/tests/torchtune) for any new functionality
- [ ] update [docstrings](https://github.com/pytorch/torchtune/tree/main/docs/source) for any new or updated methods or classes
- [ ] run unit tests via `pytest tests`
- [ ] run recipe tests via `pytest tests -m integration_test`
- [ ] manually run any new or modified recipes with sufficient proof of correctness
- [ ] include relevant commands and any other artifacts in this summary (pastes of loss curves, eval results, etc.)

#### UX
If your function changed a public API, please add a dummy example of what the user experience will look like when calling it.
Here is a [docstring example](https://github.com/pytorch/torchtune/blob/6a7951f1cdd0b56a9746ef5935106989415f50e3/torchtune/modules/vision_transformer.py#L285)
and a [tutorial example](https://pytorch.org/torchtune/main/tutorials/qat_finetune.html#applying-qat-to-llama3-models)

- [ ] I did not change any public API
- [ ] I have added an example to docs or docstrings

PR #2737, opened 2025-05-15
Add ruff pre-commit hook to check PEP 585

thomasjpfan:
#### Context
- [ ] add a new feature
- [ ] fix a bug
- [x] other (Adds pre-commit hook to check PEP 585)

Fixes https://github.com/pytorch/torchtune/issues/2734 Ruff can also replace other linters, but to keep this PR small, this PR only adds a single ruff rule.

#### Changelog
* Adds pre-commit hook to check PEP 585 using ruff

#### Test plan
To check this PR, you can `pre-commit run --all-files`, which shows all the errors. According to https://docs.astral.sh/ruff/rules/non-pep585-annotation/ this is a non-safe fix, so to fix it one needs to run:

```bash
ruff check . --unsafe-fixes --fix
```

This should reproduce the PR from https://github.com/pytorch/torchtune/pull/2720

#### UX

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -49,3 +49,10 @@ repos:
   hooks:
     - id: pydoclint
       args: [--config=pyproject.toml]
+
+repos:
+- repo: https://github.com/astral-sh/ruff-pre-commit
+  rev: v0.11.10
+  hooks:
+    - id: ruff-check
+    - id: ruff-format
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -55,4 +55,5 @@ repos:
   rev: v0.11.10
   hooks:
     - id: ruff-check
+      args: [ --fix ]
     - id: ruff-format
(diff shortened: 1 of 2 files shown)


omkar-334:
interesting, I didn;t know this existed in Ruff

thomasjpfan:
I think they can both be used at the same time. `A004` is about not defining a variable that is also a builtin name. `UP006` is about enforcing PEP 585 (i.e. `typing.List` -> `list`).

Note that lowercase `list` is a builtin while uppercase `List` is not a builtin. The builtins are listed here: https://docs.python.org/3/library/functions.html

omkar-334:
> I think they can both be used at the same time. `A004` is about not defining a variable that is also a builtin name. `UP006` is about enforcing PEP 585 (i.e. `typing.List` -> `list`).
> 
> Note that lowercase `list` is a builtin while uppercase `List` is not a builtin. The builtins are listed here: https://docs.python.org/3/library/functions.html

oh , yeah. I forgot I had replaced all `List` occurrences with `list` to trigger the `A004` warning. Understood now, makes much more sense

joecummings (approved):
Thanks for this!

PR #2741, opened 2025-05-16
Add feature ligerceloss

### LINKED ISSUE #2692: [Feature Request] Add Liger CE Loss
Add a new loss in the cross_entropy_loss.py file that inherits from SFT loss but calls the Liger [fused_linear_cross_entropy](https://github.com/linkedin/Liger-Kernel/blob/main/src/liger_kernel/ops/fused_linear_cross_entropy.py) loss. It will need to handle if the input is a DTensor and convert it before calling the liger loss.

Edge Case: if the model output is a tied embedding and TP sharded (DTensor). Then either we'll have to unshard and then reshard the weight every step, or throw an error for that case. (This assumes that liger losses don't work with sharded weights)

A good validation of this feature would be to see if this loss even further improves the numbers [here](https://github.com/pytorch/torchtune/blob/main/README.md#optimization-flags) over compiled linear cross entropy loss.

mananchawla2005:
# PR: Add LigerFusedCrossEntropyLoss

#### Context
- [x] add a new feature
- [ ] fix a bug
- [ ] other

Closes #2692

#### Changelog
- Added new `LigerFusedCrossEntropyLoss` class that provides memory-efficient cross entropy loss using fused CUDA kernels using `liger-kernels`
- Implemented proper handling of distributed tensors (DTensor) with gradient hooks
- Added comprehensive unit tests comparing against standard PyTorch cross entropy
- Added docstrings with usage examples

#### Test plan
- [x] Added unit tests in test_liger_ce_loss.py comparing against `F.cross_entropy`
- [x] Tests verify correctness with:
  - Regular tensor inputs
  - Batched inputs requiring reshaping
  - Ignored indices
- [x] Added proper docstrings with usage examples
- [x] Tests pass locally on CUDA device
- [x] Pre-commit hooks and linters pass

#### UX
Example usage in docstring:

```python
# Initialize model and loss
model = Transformer(...)  # model with skip_output_layer=True
loss = LigerFusedCrossEntropyLoss()
loss.set_model_output(model)  # This captures model's output layer

# Forward pass
hidden_states = model(inputs)  # [batch_size, seq_len, hidden_dim]
targets = labels  # [batch_size, seq_len]

# If needed, reshape to [batch_size*seq_len, hidden_dim]
hidden_states = hidden_states.view(-1, hidden_states.size(-1))
targets = targets.view(-1)

loss_value = loss(hidden_states, targets)
```

The implementation provides better performance and memory efficiency compared to the chunked `LinearCrossEntropyLoss` by:
1. Using fused CUDA kernels
2. Combining linear projection and cross entropy in a single operation
3. Properly handling distributed tensors

All tests verify numerical correctness against PyTorch's native cross entropy within expected tolerances.

<hr>

Hi PyTorch Team,

This is my first PR to a machine learning project, and I’ve tried to ensure the code is correct and well-structured. I’ve implemented the functionality for LigerCeLoss and included a test case that verifies its behavior with both masked and reshaped inputs.

Due to hardware limitations, I wasn't able to fully validate the distributed functionality or run the entire test suite across multiple GPUs. However, I’ve implemented support for DTensor-based weights, hidden states, and targets, and included the logic for gather → fuse → scatter to handle gradient computation for sharded weights in distributed settings.

Looking forward to your feedback! @pbontrager @joecummings

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -35,6 +35,12 @@ dependencies = [
 
     # Multimodal
     "Pillow>=9.4.0",
+
+    # Triton:
+    "triton>=2.3.1 ; platform_system != 'Windows'",
+
+    # Liger Kernels
+    "liger-kernel ; platform_system != 'Windows'"
 ]
 dynamic = ["version"]
 
--- a/tests/torchtune/modules/loss/test_liger_ce_loss.py
+++ b/tests/torchtune/modules/loss/test_liger_ce_loss.py
@@ -1,0 +1,108 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+import pytest
+import torch
+import torch.nn.functional as F
+from tests.test_utils import assert_expected
+from torch import nn
+from torchtune.modules.loss import LigerFusedCrossEntropyLoss
+from torchtune.training.seed import set_seed
+
+
+class Model(nn.Module):
+    def __init__(self, vocab_size, embed_dim):
+        super().__init__()
+        self.output = nn.Linear(embed_dim, vocab_size)
+
+    def forward(self, x):
+        return self.output(x)
+
+
+@pytest.fixture(autouse=True)
+def random():
+    set_seed(42)
+
+
+class TestLigerFusedCrossEntropyLoss:
+    def test_liger_fused_cross_entropy_loss(self):
+        """
+        Compares LigerFusedCrossEntropyLoss implementation vs standard F.cross_entropy
+        """
+        # Set up test parameters
+        batch_size = 2
+        seq_len = 8
+        embed_dim = 16
+        vocab_size = 100
+        ignore_index = -100
+
+        # Create dummy data
+        hidden = torch.randn(batch_size * seq_len, embed_dim, dtype=torch.float32)
+        targets = torch.randint(0, vocab_size, (batch_size * seq_len,), dtype=torch.long)
+        hidden = hidden.cuda()
+        targets = targets.cuda()
+        # Add some ignored indices
+        mask = torch.rand(batch_size * seq_len) < 0.2
+        targets[mask] = ignore_index
+
+        # Create a dummy model
+        model = Model(vocab_size, embed_dim).cuda()
+
+        # Compute fused CE loss
+        fused_ce_loss_fn = LigerFusedCrossEntropyLoss(ignore_index=ignore_index)
+        fused_ce_loss_fn.set_model_output(model)
+        fused_loss = fused_ce_loss_fn(hidden, targets)
+
+        # Compute standard cross entropy for comparison
+        logits = F.linear(hidden, model.output.weight, model.output.bias)  # [batch_size*seq_len, vocab_size]
+        standard_loss = F.cross_entropy(
+            logits, targets, reduction="sum", ignore_index=ignore_index
+        )
+
+        # Validate the results are close enough
+        assert_expected(fused_loss, standard_loss, rtol=1e-2, atol=1e-2)
+
+    def test_liger_fused_cross_entropy_loss_with_reshape(self):
+        """
+        Tests LigerFusedCrossEntropyLoss with batch x seq_len input that needs reshaping
+        """
+        # Set up test parameters
+        batch_size = 3
+        seq_len = 6
+        embed_dim = 32
+        vocab_size = 50
+        ignore_index = -100
+
+        # Create dummy data with batch x seq_len dimensions
+        hidden = torch.randn(batch_size, seq_len, embed_dim, dtype=torch.float32)
+        targets = torch.randint(0, vocab_size, (batch_size, seq_len), dtype=torch.long)
+        hidden = hidden.cuda()
+        targets = targets.cuda()
+        # Add some ignored indices
+        mask = torch.rand(batch_size, seq_len) < 0.2
+        targets[mask] = ignore_index
+
+        # Create a dummy model
+        model = Model(vocab_size, embed_dim).cuda()
+
+        # Reshape to match expected input shape for fused loss
+        hidden_reshaped = hidden.reshape(-1, embed_dim)
+        targets_reshaped = targets.reshape(-1)
+
+        # Compute fused CE loss
+        fused_ce_loss_fn = LigerFusedCrossEntropyLoss(ignore_index=ignore_index)
+        fused_ce_loss_fn.set_model_output(model)
+        fused_loss = fused_ce_loss_fn(hidden_reshaped, targets_reshaped)
+
+        # Compute standard cross entropy for comparison
+        logits = model(hidden_reshaped)  # [batch_size*seq_len, vocab_size]
+        standard_loss = F.cross_entropy(
+            logits, targets_reshaped, reduction="sum", ignore_index=ignore_index
+        )
+
+        # Validate the results are close enough
+        assert_expected(fused_loss, standard_loss, rtol=1e-2, atol=1e-2)
+
--- a/torchtune/modules/loss/cross_entropy_loss.py
+++ b/torchtune/modules/loss/cross_entropy_loss.py
@@ -9,9 +9,9 @@ import torch.nn.functional as F
 from torch import nn
 from torch.distributed.tensor import DTensor
 
+import liger_kernel.ops.fused_linear_cross_entropy
 from torchtune.modules.loss.loss_types import SFTLoss
 from torchtune.utils import get_logger
-
 log = get_logger()
 
 
@@ -143,3 +143,95 @@ class LinearCrossEntropyLoss(nn.Module, SFTLoss):
             return total_loss
         else:
             return total_loss / total_elements
+
+class LigerFusedCrossEntropyLoss(nn.Module, SFTLoss):
+    """Memory efficient Cross-entropy loss that uses fused CUDA kernels to compute the loss.
+    Combines the linear projection with the cross-entropy calculation for better performance
+    and memory efficiency.
+
+    Linear cross entropy is computed in a single fused operation. You need to skip the final 
+    projection layer in your model and pass it to the loss instead. You can setup the loss 
+    with the model as shown below.
+
+    >>> model = Transformer(...)
+    >>> loss = LigerFusedCrossEntropyLoss(...)
+    >>> loss.set_model_output(model)
+    """
+
+    def __init__(
+        self,
+        ignore_index: int = -100
+    ):
+        """Initialize the LigerFusedCrossEntropyLoss.
+
+        Args:
+            ignore_index (int): Index to ignore in the target tensor. Default is -100.
+        """
+        super().__init__()
+        self.linear_projection = None
+        self.ignore_index = ignore_index
+
+    def set_model_output(self, model: nn.Module) -> None:
+        """Modify model output to match the expected input for the loss function.
+        
+        Args:
+            model (nn.Module): The model whose output layer will be used for the loss computation.
+        """
+        model.skip_output_layer = True
+        self.linear_projection = model.output
+
+    def forward(
+        self,
+        hidden_states: torch.Tensor,
+        targets: torch.Tensor,
+    ) -> torch.Tensor:
+        """Compute the fused linear cross entropy loss.
+
+        Args:
+            hidden_states (torch.Tensor): Hidden state of the model, pre projection. Shape ``[bsz, seq_len, emb_dim]``
+            targets (torch.Tensor): Labels for the model. Shape ``[bsz, seq_len]``
+
+        Returns:
+            torch.Tensor: The computed loss value
+
+        Raises:
+            RuntimeError: If set_model_output() was not called before forward
+        """
+        if self.linear_projection is None:
+            raise RuntimeError("Must call set_model_output() before forward()")
+        if isinstance(hidden_states, DTensor):
+            hidden_states = hidden_states.full_tensor()
+        
+        orig_w = self.linear_projection.weight
+        if isinstance(orig_w, DTensor):
+            mesh, placements = orig_w.device_mesh, orig_w.placements
+            w = orig_w.full_tensor().detach().clone().requires_grad_(True)
+            def _scatter_w(grad):
+                orig_w.grad = DTensor.from_local(grad, mesh, placements)
+            w.register_hook(_scatter_w)
+        else:
+            w = orig_w
+        orig_b = self.linear_projection.bias
+        if isinstance(orig_b, DTensor):
+            mesh, placements = orig_b.device_mesh, orig_b.placements
+            b = orig_b.full_tensor().detach().clone().requires_grad_(True)
+            def _scatter_b(grad):
+                orig_b.grad = DTensor.from_local(grad, mesh, placements)
+            b.register_hook(_scatter_b)
+        else:
+            b = orig_b
+
+        loss, _ = liger_kernel.ops.fused_linear_cross_entropy.LigerFusedLinearCrossEntropyFunction.apply(
+                hidden_states, 
+                w,
+                targets,
+                b,
+                None,                # ce_weight
+                self.ignore_index,   # ignore_index
+                0.0,                 # lse_square_scale
+                0.0,                 # label_smoothing
+                "sum",               # reduction
+                None,                # softcap
+                False                # return_z_loss
+            )
+        return loss
--- a/docs/source/api_ref_modules.rst
+++ b/docs/source/api_ref_modules.rst
@@ -35,6 +35,7 @@ Losses
     :nosignatures:
 
     loss.LinearCrossEntropyLoss
+    loss.LigerFusedCrossEntropyLoss
     loss.ForwardKLLoss
     loss.ForwardKLWithChunkedOutputLoss
     loss.ReverseKLWithChunkedOutputLoss
--- a/torchtune/modules/loss/__init__.py
+++ b/torchtune/modules/loss/__init__.py
@@ -7,6 +7,7 @@
 from .ce_chunked_output_loss import CEWithChunkedOutputLoss
 
 from .cross_entropy_loss import LinearCrossEntropyLoss
+from .cross_entropy_loss import LigerFusedCrossEntropyLoss
 from .kd_losses import (
     ForwardKLLoss,
     ForwardKLWithChunkedOutputLoss,
@@ -26,6 +27,7 @@ __all__ = [
     "SymmetricKLLoss",
     "SymmetricKLWithChunkedOutputLoss",
     "LinearCrossEntropyLoss",
+    "LigerFusedCrossEntropyLoss"
     "SFTLoss",
     "RLLoss",
 ]


mananchawla2005:
CLA filled!

tests/torchtune/modules/loss/test_liger_ce_loss.py
```diff
@@ -0,0 +1,108 @@
...
+        return self.output(x)
+
+
+@pytest.fixture(autouse=True)
+def random():
+    set_seed(42)
+
+
+class TestLigerFusedCrossEntropyLoss:
+    def test_liger_fused_cross_entropy_loss(self):
```
pbontrager, line 31: Since this test requires cuda you should add the "@gpu_test(gpu_count=1)" decorator from "from tests.test_utils import gpu_test". Along with testing the loss value, I think it would be good to test a single forward and backward pass with opt step to ensure all the gradients are propagating back correctly too. You can use "fixed_init_model" (also from test_utils) as well to make it easier to initialize the model the same way each time

tests/torchtune/modules/loss/test_liger_ce_loss.py
```diff
@@ -0,0 +1,108 @@
...
+        # Compute standard cross entropy for comparison
+        logits = F.linear(hidden, model.output.weight, model.output.bias)  # [batch_size*seq_len, vocab_size]
+        standard_loss = F.cross_entropy(
+            logits, targets, reduction="sum", ignore_index=ignore_index
+        )
+
+        # Validate the results are close enough
+        assert_expected(fused_loss, standard_loss, rtol=1e-2, atol=1e-2)
+
+    def test_liger_fused_cross_entropy_loss_with_reshape(self):
```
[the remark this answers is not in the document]
pbontrager: I think we should have a second test, but it should be a distributed test. Same as the first test but with 4 gpus required and FSDP size 2 and TP size 2. If you need help on how to initialize the model that way I can give you the code.

pyproject.toml
```diff
@@ -35,6 +35,12 @@ dependencies = [
 
     # Multimodal
     "Pillow>=9.4.0",
+
+    # Triton:
+    "triton>=2.3.1 ; platform_system != 'Windows'",
```
pbontrager, line 40: I don't believe we'll want these as a primary dependency as we try to keep this list as lean as possible for users. I would move these to the optional dependencies for dev users and we'll include instructions for the extra dependencies in the loss doc string.

torchtune/modules/loss/cross_entropy_loss.py
```diff
@@ -9,9 +9,9 @@
 from torch import nn
 from torch.distributed.tensor import DTensor
 
+import liger_kernel.ops.fused_linear_cross_entropy
```
pbontrager, line 12: We should move this into the init of the model as they're an optional dependency

torchtune/modules/loss/cross_entropy_loss.py
```diff
@@ -143,3 +143,95 @@ def forward(
             return total_loss
         else:
             return total_loss / total_elements
+
+class LigerFusedCrossEntropyLoss(nn.Module, SFTLoss):
+    """Memory efficient Cross-entropy loss that uses fused CUDA kernels to compute the loss.
```
pbontrager, line 148: I think this should mention that this is an approximation of CrossEntropyLoss and as such will have some numerical differences. Further down there should also be a "Note" that using this loss requires pip install triton, liger_kernel.

torchtune/modules/loss/cross_entropy_loss.py
```diff
@@ -143,3 +143,95 @@ def forward(
             return total_loss
         else:
             return total_loss / total_elements
+
+class LigerFusedCrossEntropyLoss(nn.Module, SFTLoss):
```
pbontrager, line 147: I think LigerLinearCrossEntropy would match our naming convention more

torchtune/modules/loss/cross_entropy_loss.py
```diff
@@ -143,3 +143,95 @@ def forward(
...
+            hidden_states = hidden_states.full_tensor()
+        
+        orig_w = self.linear_projection.weight
+        if isinstance(orig_w, DTensor):
+            mesh, placements = orig_w.device_mesh, orig_w.placements
+            w = orig_w.full_tensor().detach().clone().requires_grad_(True)
+            def _scatter_w(grad):
+                orig_w.grad = DTensor.from_local(grad, mesh, placements)
+            w.register_hook(_scatter_w)
+        else:
```
pbontrager, line 212: nit: this else block doesn't seem necessary. There's no need for orig_w, you can just call it w and the if DTensor block would update that variable without affecting self.linear_projection.weight

torchtune/modules/loss/cross_entropy_loss.py
```diff
@@ -143,3 +143,95 @@ def forward(
...
+        """
+        if self.linear_projection is None:
+            raise RuntimeError("Must call set_model_output() before forward()")
+        if isinstance(hidden_states, DTensor):
+            hidden_states = hidden_states.full_tensor()
+        
+        orig_w = self.linear_projection.weight
+        if isinstance(orig_w, DTensor):
+            mesh, placements = orig_w.device_mesh, orig_w.placements
+            w = orig_w.full_tensor().detach().clone().requires_grad_(True)
```
pbontrager, line 208: Can you tell me more about what you're doing here? Does the liger loss require you to detach the weight? Why detach it only to manually register that gradients get reapplied? Also, I don't think you'd want to register a hook every forward pass.

pbontrager (commented):
Thank you for this contribution! Really clean first pass at the problem. I left some comments around testing and questions about how you're handling DTensors.

torchtune/modules/loss/cross_entropy_loss.py
```diff
@@ -143,3 +143,95 @@ def forward(
...
+        self.ignore_index = ignore_index
+
+    def set_model_output(self, model: nn.Module) -> None:
+        """Modify model output to match the expected input for the loss function.
+        
+        Args:
+            model (nn.Module): The model whose output layer will be used for the loss computation.
+        """
+        model.skip_output_layer = True
+        self.linear_projection = model.output
```
pbontrager, line 181: Since this loss requires that the linear_projection contains a weight and bias parameter. It might be a good idea to check for that here and raise an Exception if they're missing or if there are other parameters.

tests/torchtune/modules/loss/test_liger_ce_loss.py
pbontrager: I think it would be good to add `@pytest.mark.parametrize("compile", [False, True])` to the test and pass in compile as an argument on whether to call `apply_compile_strategy` on the loss

mananchawla2005:
@pbontrager Hey thanks for an indepth review! I have tried to resolve most of the issues that were coming earlier however for the distributed test would love if you can help a little with the code as well as testing it cause I dont have a distributed gpu setup.

torchtune/modules/loss/cross_entropy_loss.py
```diff
@@ -143,3 +143,140 @@ def forward(
...
+            )
+
+    def apply_compile_strategy(self, *args, **kwargs):
+        """Applies compile to the forward pass for fused kernel operations."""
+        log.warning("Skipping compile loss, as it is not supported at this time")
+        # TODO: Fix it later failing for charachter encoding
+        # self.forward = torch.compile(
+        #     self.forward, *args, **kwargs
+        # )
+        return self
```
Andrei-Aksionov, line 217: Do you need to compile a liger kernel at all?

Andrei-Aksionov: Oh, ok, now I see 🙂 

So the problem is that the kernel is automatically compiled.
It was obvious from the README of their repo - there was no mentioning that a user needs to call `torch.compile` manually, which leaves only one option. Yes, Liger provides custom optimized triton kernels, but without compilation they won't work.

So, after digging a bit of their codebase, here is how it works:
1. `LigerFusedLinearCrossEntropyFunction` has custom forward and backward implementations. (Let's focus on [forward](https://github.com/linkedin/Liger-Kernel/blob/main/src/liger_kernel/ops/fused_linear_cross_entropy.py#L207) variant for now.)
2. Inside it, a [fused_linear_cross_entropy_forward](https://github.com/linkedin/Liger-Kernel/blob/main/src/liger_kernel/ops/fused_linear_cross_entropy.py#L240) is called ... 
3. ... which calls a [triton kernel](https://github.com/linkedin/Liger-Kernel/blob/main/src/liger_kernel/ops/fused_linear_cross_entropy.py#L94), that has [triton.jit](https://github.com/linkedin/Liger-Kernel/blob/main/src/liger_kernel/ops/cross_entropy.py#L25) wrapper.

Of course, you can control even that with 
```python
with torch._dynamo.disable():
```

 and a flag that is disabled by default and enabled in `apply_compile_strategy`, but since Triton kernel code is not valid Python code for direct execution on a GPU without a compilation, there is no much sense in it.
Perhaps maybe control compilation of service code inside a loss code around this kernel, but I believe it won't make that much difference 🤔.

In other words, since it was never intended to use liger without a compilation, perhaps just skip this method without any warnings?

intervitens:
I think it makes more sense to reshape `hidden_states` and `targets` to `[batch_size*seq_len]` inside the loss forward and to set `reduction` to `mean` instead of `sum` in order to match the behavior of existing `LinearCrossEntropyLoss` and allow the new `LigerLinearCrossEntropy` to be used as a drop-in replacement in recipe configs.

I've tested a distributed LoRA finetune of llama3.1-8B with those changes, and it seems to work fine, amount of reserved memory was reduced and the difference in loss was minimal. 
https://wandb.ai/intervitens/8B-lora-dist-liger

torchtune/modules/loss/cross_entropy_loss.py
```diff
@@ -143,3 +143,140 @@ def forward(
...
+    >>> loss.set_model_output(model)
+    >>> loss.apply_compile_strategy()
+    """
+
+    def __init__(self, ignore_index: int = -100):
+        super().__init__()
+        try:
+            import liger_kernel.ops.fused_linear_cross_entropy
+
+            self.fused_linear_ce = liger_kernel.ops.fused_linear_cross_entropy
```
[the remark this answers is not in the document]
mananchawla2005: Hey, you are right to point out that! I chose the lower-level ops approach for bias handling as one under `transformers` don't have a option for bias parameter and during distributed training we need to handle DTensor. If thats not required I can replace it with the one in readme.

torchtune/modules/loss/cross_entropy_loss.py
```diff
@@ -238,6 +229,15 @@ def forward(
         if isinstance(hidden_states, DTensor):
             hidden_states = hidden_states.full_tensor()
 
+        batch_size, seq_len, emb_dim = hidden_states.shape
+        hidden_states = hidden_states.reshape(
+            -1, emb_dim
+        )  # [batch_size*seq_len, emb_dim]
+        targets = targets.reshape(-1)  # [batch_size*seq_len]
```
Andrei-Aksionov, line 236: Feels like since you don't reuse B, T, C values anywhere, it could be done simpler:

```suggestion
        hidden_states = hidden_states.flatten(0, 1) # (batch_size*seq_len, hidden_size)
        targets = targets.flatten() # (batch_size*seq_len)
```

torchtune/modules/loss/cross_entropy_loss.py
```diff
@@ -275,8 +275,10 @@ def _scatter_b(grad):
...
             0.0,  # lse_square_scale
             0.0,  # label_smoothing
-            "sum",  # reduction
+            "mean",  # reduction
             None,  # softcap
             False,  # return_z_loss
         )
+        if total_elements == 0:
+            return loss
         return loss
```
Andrei-Aksionov, line 284: So, basically, return `loss` regardless? 🙂

torchtune/modules/loss/cross_entropy_loss.py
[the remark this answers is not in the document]
Andrei-Aksionov: But it looks like forward method of the class also [accepts](https://github.com/linkedin/Liger-Kernel/blob/dd492af4cb156fe35bace711c1e6191915e5d52c/src/liger_kernel/transformers/fused_linear_cross_entropy.py#L36) bias, which by default is None.

Basically, there is nothing wrong with your approach, just the one with the class looks, at least to me, slightly cleaner.

torchtune/modules/loss/cross_entropy_loss.py
```diff
@@ -143,3 +143,142 @@ def forward(
...
+        targets = targets.reshape(-1)  # [batch_size*seq_len]
+
+        mask = targets != self.ignore_index
+        total_elements = mask.sum()
+
+        w = self.linear_projection.weight
+        if isinstance(w, DTensor):
+            mesh, placements = w.device_mesh, w.placements
+            w = w.full_tensor()
+            if not hasattr(self, "_w_hook_registered"):
```
pbontrager, line 245: I think "full_tensor" handles gradient placement and we don't need to do this ourselves [link](https://github.com/pytorch/pytorch/blob/35ddad284d350681073aa06a33904d1f31e6fff0/torch/distributed/tensor/_api.py#L571). I can test removing this though.

torchtune/modules/loss/cross_entropy_loss.py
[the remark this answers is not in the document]
pbontrager: Since the class version was just a thin wrapper around LigerFusedLinearCrossEntropyFunction, and our loss class is also a thin wrapper around the same functionality, it feels more right to me that we just directly call LigerFusedLinearCrossEntropyFunction and operate at the same abstraction level as the nn.Module you linked to.

tests/torchtune/modules/loss/test_liger_ce_loss.py
```diff
@@ -0,0 +1,146 @@
...
+import torch.nn.functional as F
+from tests.test_utils import assert_expected, fixed_init_model, gpu_test
+from torch import nn
+from torch.distributed.tensor import DTensor
+from torch.optim import SGD
+from torchtune.modules.loss import LigerLinearCrossEntropy
+from torchtune.training.seed import set_seed
+
+
+@gpu_test(gpu_count=1)
```
pbontrager, line 18: Let me edit and push some changes to these tests. I can run them myself then for you.

pbontrager (commented):
Thank you for doing all the updates! I think it's close now. I'm going to help with the unit tests and then once you have a chance to make any changes based on Andrei's comments, we should be good to land.

torchtune/modules/loss/cross_entropy_loss.py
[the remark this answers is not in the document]
Andrei-Aksionov: Ok, fair enough. 
Linear loss calls F.cross_entropy and this one calls a function - a lil bit of uniformity.

mananchawla2005:
@pbontrager I have simplifed distributed handling as well done the refactoring as per suggested changes. Would be greatful if you can help me add tests for distributed.

pbontrager:
@mananchawla2005 I've gotten the unit tests working to test FSDP + TP for a single training step with the Liger loss. I haven't pushed the changes to your PR yet because the DTensor backward hook doesn't seem to work correctly so I'm trying to fix that and get the test to pass. I should be able to get back to this and get something to you by the end of this week.

tests/torchtune/modules/loss/test_liger_ce_loss.py
```diff
@@ -0,0 +1,205 @@
...
+        # 1. Validate the results are close enough
+        assert_expected(fused_loss, standard_loss, rtol=1e-2, atol=1e-2)
+
+        # 2. Validate the weight updates are close enough
+        for p1, p2 in zip(fused_model.parameters(), model.parameters()):
+            assert_expected(p1, p2, rtol=1e-2, atol=1e-2)
+
+        dist.destroy_process_group()
+
+    @gpu_test(gpu_count=WORLD_SIZE)
```
ebsmothers, line 195: nit: just say 4 explicitly here. I like to be able to look directly at the test and see how many GPUs it needs

tests/torchtune/modules/loss/test_liger_ce_loss.py
```diff
@@ -0,0 +1,205 @@
...
+        for m in [model.layer, model.output, model]:
+            fully_shard(m, mesh=mesh["dp_shard"])
+
+        fused_loss, standard_loss = self.compute_loss(
+            fused_model, model, device, compile, mesh["dp_shard"].get_local_rank()
+        )
+
+        # Verify:
+        # 1. Validate the results are close enough
+        assert_expected(fused_loss, standard_loss, rtol=1e-2, atol=1e-2)
```
ebsmothers, line 187: Is this as close as we can get?

torchtune/modules/loss/cross_entropy_loss.py
```diff
@@ -143,3 +143,113 @@ def forward(
...
         else:
             return total_loss / total_elements
+
+
+class LigerLinearCrossEntropy(nn.Module, SFTLoss):
+    """Memory efficient Cross-entropy loss that uses fused CUDA kernels to compute the loss.
+    Combines the linear projection with the cross-entropy calculation for better performance
+    and memory efficiency. This is an approximation of CrossEntropyLoss and may have small
+    numerical differences compared to the standard implementation. This is a wrapper around
+    `LigerFusedLinearCrossEntropyFunction` from the `liger_kernel` package.
```
ebsmothers, line 153: Can we add a reference to the Liger paper or repo here?

tests/torchtune/modules/loss/test_liger_ce_loss.py
```diff
@@ -0,0 +1,205 @@
...
+        )
+        torch.cuda.set_device(device)
+        mesh = ParallelDims(
+            dp_replicate=1, dp_shard=WORLD_SIZE // 2, tp=2, cp=1, world_size=WORLD_SIZE
+        ).build_mesh("cuda")
+        plan = {
+            "output": ColwiseParallel(
+                input_layouts=Replicate(), output_layouts=Replicate()
+            )
+        }
```
ebsmothers, line 165: Is this actually a nontrivial application of TP? (Like should we at least have some shard for `layer` or something?)

ebsmothers (approved):
Thanks @mananchawla2005 for getting this most of the way! I have a few more small comments based on @pbontrager's changes, then this should be good to go. @pbontrager please make sure to test on post-6/4 nightlies given the changes in https://github.com/pytorch/pytorch/pull/154704. Stamping to unblock

torchtune/modules/loss/cross_entropy_loss.py
```diff
@@ -143,3 +143,113 @@ def forward(
...
+            torch.Tensor: The computed loss value
+
+        Raises:
+            RuntimeError: If set_model_output() was not called before forward
+        """
+        if self.linear_projection is None:
+            raise RuntimeError("Must call set_model_output() before forward()")
+
+        if isinstance(hidden_states, DTensor):
+            hidden_states = hidden_states.full_tensor()
```
nathan-az, line 234: So the compute isn't replicated across TP workers unnecessarily, could the hidden states and targets both be sharded on the sequence dimension, losses calculated per token, then correctly reduced after? (Basically to make better uses of the devices in a TP group).

Could be a follow-up PR, not MVP for liger. Just raising while I work on adding loss parallel for the standard `LinearCrossEntropyLoss` https://github.com/pytorch/torchtune/pull/2782/files

(EDIT: I can probably look to add this in future)

vadimkantorov:
Any plans for contributing this great fused linear cross entropy loss impl directly into PyTorch core?
- https://github.com/pytorch/pytorch/issues/124480
- https://github.com/pytorch/pytorch/issues/139908 (so far it shows that Inductor under-performs compared to Liger, so having the impl in torchtune run against this benchmark would be good)

As probably the same question is relevant for torchtitan, so would be best to have it upstreamed? (although, maybe a copy of Liger Triton (or CUDA?) kernel would probably need to be upstreamed, this PR is a wrapper around Liger package) And this way, it would also provide a good baseline for the Inductor codegen

I think it's been now quite some time that fused linear cross entropy has proved it's actually useful for warranting its inclusion in core (especially in the meanwhile Inductor underperforms codegen for this op)

ebsmothers:
@vadimkantorov actually this is something that @ngimel has brought up as well, she may have a better idea on the latest status of things here.

> Any plans for contributing this great fused linear cross entropy loss impl directly into PyTorch core?
> 
> * [Fused Linear and Cross-Entropy Loss `torch.nn.functional.linear_cross_entropy` pytorch#124480](https://github.com/pytorch/pytorch/issues/124480)
> * [Inductor vs. Liger Performance Track pytorch#139908](https://github.com/pytorch/pytorch/issues/139908) (so far it shows that Inductor under-performs compared to Liger, so having the impl in torchtune run against this benchmark would be good)
> 
> As probably the same question is relevant for torchtitan, so would be best to have it upstreamed? (although, maybe a copy of Liger Triton (or CUDA?) kernel would probably need to be upstreamed, this PR is a wrapper around Liger package) And this way, it would also provide a good baseline for the Inductor codegen
> 
> I think it's been now quite some time that fused linear cross entropy has proved it's actually useful for warranting its inclusion in core (especially in the meanwhile Inductor underperforms codegen for this op)

vadimkantorov:
I wonder if both triton-code and triton-produced cubin / ptx can be included in core distribution for populating local artifact cache . Like so we can both preserve triton's hackability and zero wait time for eager mode...

mikaylagawarecki:
Hey @vadimkantorov,

> Any plans for contributing this great fused linear cross entropy loss impl directly into PyTorch core?
> 
> * [Fused Linear and Cross-Entropy Loss `torch.nn.functional.linear_cross_entropy` pytorch#124480](https://github.com/pytorch/pytorch/issues/124480)
> * [Inductor vs. Liger Performance Track pytorch#139908](https://github.com/pytorch/pytorch/issues/139908) (so far it shows that Inductor under-performs compared to Liger, so having the impl in torchtune run against this benchmark would be good)
> 
> As probably the same question is relevant for torchtitan, so would be best to have it upstreamed? (although, maybe a copy of Liger Triton (or CUDA?) kernel would probably need to be upstreamed, this PR is a wrapper around Liger package) And this way, it would also provide a good baseline for the Inductor codegen
> 
> I think it's been now quite some time that fused linear cross entropy has proved it's actually useful for warranting its inclusion in core (especially in the meanwhile Inductor underperforms codegen for this op)

Wanted to let you know that a fused linear cross entropy in core is on our roadmap, we plan to work on it in the next month or so

vadimkantorov:
Any plans for a road forward for inclusion in core directly triton codes and torch.compile-d codes?

mikaylagawarecki:
> Any plans for a road forward for inclusion in core directly triton codes and torch.compile-d codes?

>  wonder if both triton-code and triton-produced cubin / ptx can be included in core distribution for populating local artifact cache . Like so we can both preserve triton's hackability and zero wait time for eager mode...

Not sure if this is what you're referring to, but I believe there is already infra to include triton code in core https://github.com/pytorch/pytorch/blob/a69e27ca5ad4287add73972ef1b34b469e3c7d23/torch/cuda/__init__.py#L1669-L1702 that is used for a very small number of ops

As to the triton-produced cubin / ptx I'm not sure how that would look like but I'll investigate when we look at upstreaming the fused linear cross entropy

vadimkantorov:
Exactly, I wonder what is needed for Triton code to be used for more ops in core. If it's cold start time for eager, then shipping pre-cached / pre-generated / pre-compiled ptx could be the solution (and still allow hackability if one wants to copy-paste and modify the triton code)

ngimel:
Precompiled triton is not really an option, triton is doing too much specialization, so startup time/unexpected reompiles  will always be a problem for using triton in eager.

vadimkantorov:
Is it possible to somehow precompile / pregenerate from triton some version of ptx which at least would run on all relevant hardwares?

Or are some new features in triton needed to control how it specializes / force it to specialize less? (I guess ideally one would like to have some sort of okay AOT ptx generation from triton which would at least run on many hardwares - even if not reaching the best possible perf. The PyTorch could ship ptx for a few popular important hardwares, and for others force the user to run triton compilation in their own machine if they want extra perf. This would allow users to write and maintain triton core where now they have to rewrite in cuda )

Or is it possible to lower triton into the cuda/c++ code?

nathan-az:
@mikaylagawarecki If there are any public issues/PRs that we could follow to track progress for a fused linear loss in Core, that would be great. I don't see it currently in the [2.9.0 milestones](https://github.com/pytorch/pytorch/milestone/55).

vadimkantorov:
> Wanted to let you know that a fused linear cross entropy in core is on our roadmap, we plan to work on it in the next month or so

@mikaylagawarecki If maybe vllm's impls could be upstreamed, vllm could transition to use more of core modules. Hopefully, vllm/sglang can transition to be using these new pytorch core modules (I wonder if this should inform the design/numerics in any ways) and thus reducing fragmentation...

Yet another recent impl of fused linear cross entropy (indicating fragmentation): https://github.com/volcengine/verl/blob/main/verl/utils/kernel/kernels.py

Also, a tutorial is needed for implementing forward+backward losses fusing Linear + some cross-entropy-like loss (e.g. important case is GRPO)

vadimkantorov:
Also curious if the pattern of fusing chunked Linear and some loss computation can be also implemented as a more generic/compilable higher-order op in PyTorch core. Seems the same pattern arises / is needed for GRPO loss and reducing peak VRAM for super-long reasoning traces (and maybe can be done with other loss variants as well)...

PR #2745, opened 2025-05-17
Fixing counting number of batches for accumulation through epoch

wesbz:
#### Context
- [ ] add a new feature
- [x] fix a bug
- [ ] other (please add here)

As I was running DPO and SFT, I noticed two surprising behaviours when running for several epochs:
image

1. The loss would drop at the beginning of a new epoch;
2. The accuracy (for DPO) would be >100% at the beginning of a new epoch.
I noticed that statistics accumulators re-initialization and zero-ing gradients would happen in this condition (e.g. `recipes/full_finetune_distributed.py`):
```python
    ...
    # Optimizer step (if not fused in backward call)
    if (idx+1) % self._gradient_accumulation_steps == 0:
        ...
```
Now the issue is that if your gradient accumulation parameter is set at say 8, but you only have 63 batches to process, it means that you process the last 7 batches, accumulating statistics and gradients, without re-initializing them before starting the new epoch, continuing accumulating stats and grads for the first 8 batches and then do a step.
You end up accumulating 15 batches instead of 8, messing up not only with the statistics report but also with the optimisation.
I see at least two possibilities:
1. dropping the last few batches so that `len(self._dataloader) % self._gradient_accumulation_steps == 0`;
2. counting the number of batches to accumulate not in terms of batch index but absolute number of processed batches.

#### Changelog
This PR implements the second option.
It should be done for all recipe but I thought discussing the solution first was better.

#### Test plan

#### UX

### DIFF
--- a/recipes/full_dpo_distributed.py
+++ b/recipes/full_dpo_distributed.py
@@ -210,6 +210,7 @@ class FullDPORecipeDistributed(FTRecipeInterface):
         self.total_epochs = cfg.epochs
         self.max_steps_per_epoch = cfg.max_steps_per_epoch
         self.global_step = 0
+        self.id_batch = 0
 
     def _load_ref_checkpoint(self, cfg_ref_checkpointer: DictConfig) -> dict[str, Any]:
         """
@@ -871,8 +872,10 @@ class FullDPORecipeDistributed(FTRecipeInterface):
 
                 loss.backward()
 
+                self.id_batch += 1
+
                 # Step with optimizer
-                if (idx + 1) % self._gradient_accumulation_steps == 0:
+                if self.id_batch % self._gradient_accumulation_steps == 0:
                     # Accumulate running metrics across all devices
                     torch.distributed.all_reduce(
                         running_loss, op=torch.distributed.ReduceOp.AVG
--- a/recipes/full_finetune_distributed.py
+++ b/recipes/full_finetune_distributed.py
@@ -266,6 +266,7 @@ class FullFinetuneRecipeDistributed(FTRecipeInterface):
         self.total_epochs = cfg.epochs
         self.max_steps_per_epoch = cfg.max_steps_per_epoch
         self.global_step = 0
+        self.id_batch = 0
 
     def _update_recipe_state(self, ckpt_dict: dict[str, Any]) -> None:
         """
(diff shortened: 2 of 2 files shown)


joecummings (commented):
Thanks for catching this @wesbz !

After some discussion with @pbontrager  and @ebsmothers , I think we actually prefer an approach like the one you outlined as option 1, where in we simply drop the extra grads / metrics for any data that has been accumulated before the end of an epoch. 

In our opinion, this breaks the least amount of assumptions around which data is trained in a certain epoch and also mirrors the behavior of the PyTorch dataloader which exposes an parameter called [`drop_last`](https://docs.pytorch.org/docs/stable/data.html#torch.utils.data.DataLoader), which will similarly drop any remaining data that does not divide evenly into the size of your batch.

wesbz:
Hey @joecummings , any update on this?

felipemello1 (commented):
hey @wesbz , thank you for explaining the error and submitting a PR!

In the recipes we do this: 
```
self._steps_per_epoch = (
            len(self._dataloader) // self._gradient_accumulation_steps
        )
if (
    self.max_steps_per_epoch is not None
    and self.max_steps_per_epoch < self._steps_per_epoch
):
    self._steps_per_epoch = self.max_steps_per_epoch
```

so it seems to me that all we need is to replace `max_steps_per_epoch` with `_steps_per_epoch` in the current condition. 

```
if (idx + 1) // self._gradient_accumulation_steps == self.steps_per_epoch:
    break
```

For example:
- with 63 batches and gradient_accumulation_steps = 8, _steps_per_epoch = 63 // 8 = 7
- The loop processes batches 0–55 (7 * 8 = 56), with optimizer steps after batches 7, 15, 23, 31, 39, 47, and 55, then breaks. 
- The last 7 batches are dropped, ensuring 8 batches per step and no carryover.
- If max_steps_per_epoch = 40, _steps_per_epoch = 40, it processes batches 0–39 (5 * 8 = 40), stopping after 5 steps.
- if max_steps_per_epoch = 100, we will still have _steps_per_epoch = 63

Let me know if that makes sense. @joecummings , wanna double check it?
Personally, i think that option 2) "counting the number of batches to accumulate not in terms of batch index but absolute number of processed batches." would be the best and avoid edge cases, e.g. num steps < num_accumulation. But those should be rare either way,

tip: Lets try to align on a single recipe first, and when you get a green light, do it for all other recipes, so its easier for you. Thanks again!

wesbz:
Hi @felipemello1 ,
Thanks for your answer. It seems to me like there is a confusion in the roles of `max_steps_per_epoch`/`_steps_per_epoch`. They're not here to count the number of processed batches but number of steps. So in your cases, if `max_steps_per_epoch=40 or 100`, it doesn't change anything.
The 2nd solution "counting the number of batches to accumulate not in terms of batch index but absolute number of processed batches." implied having to carryover the last few batches of an epoch. It seems to me like the solution you describe is actually the first one

PR #2748, opened 2025-05-18
Add support for a `.jsonl` logger

wesbz:
#### Context
- [x] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

Currently, the `DiskLogger` simply saves the logs into a `.txt` file in a quite raw format. While it's not too much work to parse it, it would be much more comfortable to simply have to work with `.jsonl` files (or csv files but I personally use `.jsonl`).

#### Changelog
* Implemented the `JsonLLogger` which inherits from the `DiskLogger` class and performs the exact same job except it does it in the `.jsonl` format instead of the `.txt` format.

#### Test plan

#### UX

### DIFF
--- a/tests/torchtune/training/test_metric_logging.py
+++ b/tests/torchtune/training/test_metric_logging.py
@@ -5,6 +5,7 @@
 # LICENSE file in the root directory of this source tree.
 
 
+import json
 import tempfile
 from io import StringIO
 from typing import cast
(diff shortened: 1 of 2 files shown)


joecummings (commented):
This will be very helpful, thanks @wesbz !

However, I think the functionality for outputting to JSONL instead of TXT could still be well contained in the DiskLogger interface. I might suggest instead just exposing a parameter in its init called `output_fmt` (or something similar, I'm not picky), and based on that output to JSONL or TXT. 

What do you think?

joecummings (approved):
LGTM, thank you!

PR #2753, opened 2025-05-21
Fix failing KD single device recipe

ankitageorge:
#### Context
- [ ] add a new feature
- [x] fix a bug
- [ ] other (please add here)

Fix bug introduced in https://github.com/pytorch/torchtune/pull/2726

#### Changelog
* In the PR that caused the issue (https://github.com/pytorch/torchtune/pull/2726) we're loading the teacher checkpoint by providing the checkpoint client with only the teacher checkpoint config, but this causes it to not have the values of variables that are set like $output_dir. This PR fixes the issue by providing the checkpoint client with the whole config when loading teacher checkpoints.

#### Test plan

- [x ] manually run any new or modified recipes with sufficient proof of correctness

#### UX

### DIFF
--- a/recipes/knowledge_distillation_single_device.py
+++ b/recipes/knowledge_distillation_single_device.py
@@ -156,7 +156,6 @@ class KDRecipeSingleDevice(FTRecipeInterface):
         del new_cfg["checkpointer"]
         teacher_checkpoint_val = new_cfg.pop("teacher_checkpointer")
         new_cfg["checkpointer"] = teacher_checkpoint_val
-        print("new config is: ", new_cfg)
 
         teacher_checkpoint_client = CheckpointClient(
             new_cfg,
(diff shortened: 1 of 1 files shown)


ebsmothers (approved):
Thanks @ankitageorge for the quick fix! After looking at things a bit more closely, I think the way we've hooked up `CheckpointClient` into the recipe may be a bit brittle. I'm glad you're able to make things work here, but the fact that we need to basically copy and modify the full config in order to use it doesn't smell right to me. Looking at the class, it seems we really only use `cfg` for `enable_async_checkpointing`, `optimizer_in_bwd`, `device`, `resume_from_checkpoint`, and `checkpointer`. So maybe we should just be explicit and require `CheckpointClient` to take those args instead of the full config? Then we don't need this extra config trickery to support multiple models (which I think will be a fairly common use case). Anyways, not a blocking comment for this PR, but we should think about how to redesign to make things more explicit and intuitive here.

PR #2754, opened 2025-05-21
Add async checkpointing for KD distributed and lora dpo distributed

ankitageorge:
#### Context
- [x] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

#### Changelog
* Add async checkpointing for KD distributed and lora dpo distributed recipes
* Also fix how teacher checkpoints are initialized with checkpoint client so that it doesn't involve hacky-ness of updating the config manually

#### Test plan

- [x ] run pre-commit hooks and linters (make sure you've first installed via `pre-commit install`)
- [x ] add [unit tests](https://github.com/pytorch/torchtune/tree/main/tests/torchtune) for any new functionality

#### UX

- [x ] I did not change any public API


recipes/knowledge_distillation_single_device.py
```diff
@@ -149,17 +149,11 @@ def load_teacher_checkpoint(self, cfg: DictConfig) -> dict[str, Any]:
...
-        teacher_checkpoint_val = new_cfg.pop("teacher_checkpointer")
-        new_cfg["checkpointer"] = teacher_checkpoint_val
-
-        teacher_checkpoint_client = CheckpointClient(
-            new_cfg,
+
+        teacher_checkpointer = config.instantiate(
+            cfg.teacher_checkpointer,
         )
+        teacher_checkpoint_client = CheckpointClient(cfg, teacher_checkpointer)
```
joecummings, line 156: IIUC, because we initialize with a checkpointer, when we go to load in a checkpoint, it will hit [this](https://github.com/pytorch/torchtune/blob/3e353259091ccdde439dbcc150812859319c4d81/torchtune/training/checkpointing/_checkpoint_client.py#L93) line and NOT load in a distributed checkpoint with a recipe state.

Is that right?

joecummings (commented):
Just a question

recipes/knowledge_distillation_single_device.py
ankitageorge: yes, you're right. But from my understanding we would never save an intermediate teacher checkpoint right? so it would never be saved with DCP anyways

joecummings (approved):

PR #2755, opened 2025-05-21
[llama4] use grouped_mm in moe for sm90

IvanKobzarev:
* (to be filled)

### DIFF
--- a/torchtune/models/llama4/_component_builders.py
+++ b/torchtune/models/llama4/_component_builders.py
@@ -38,6 +38,7 @@ from torchtune.modules.moe import (
     TokenChoiceTopKRouter,
 )
 from torchtune.modules.peft import DoRALinear, LORA_ATTN_MODULES, LoRALinear
+from torchtune.utils._device import has_cuda_capability
 
 """
 Component builders for the Llama4 model.
@@ -180,6 +181,7 @@ def llama4_decoder(
     num_experts: int = 16,
     experts_per_token: int = 1,
     use_shared_expert: bool = True,
+    use_grouped_mm: bool = True,
     use_qk_norm: bool = True,
     moe_every_n_layers: Optional[int] = None,
     mlp_hidden_dim: Optional[int] = None,
@@ -244,6 +246,11 @@ def llama4_decoder(
         raise ValueError(
             "Must pass local_chunk_size when enabling local chunked attention"
         )
+    if use_grouped_mm and not has_cuda_capability(9, 0):
+        torchtune.utils.get_logger("WARNING")(
+            "Failed to use grouped mm, which is only supported on SM90 or later",
+        )
+        use_grouped_mm = False
     head_dim = embed_dim // num_heads
     num_kv_heads = num_kv_heads if num_kv_heads else num_heads
 
@@ -263,7 +270,6 @@ def llama4_decoder(
         )
     layers = []
     for i in range(num_layers):
-
         mask_mod = None
         if skip_rope_interval is not None and (i + 1) % skip_rope_interval != 0:
             mask_mod = partial(
@@ -300,6 +306,7 @@ def llama4_decoder(
                 num_experts=num_experts,
                 experts_per_token=experts_per_token,
                 use_shared_expert=use_shared_expert,
+                use_grouped_mm=use_grouped_mm,
             )
         else:
             mlp_layer = llama4_mlp(dim=embed_dim, hidden_dim=mlp_hidden_dim)
@@ -355,6 +362,7 @@ def llama4_moe(
     num_experts: int = 8,
     experts_per_token: int = 1,
     use_shared_expert: bool = True,
+    use_grouped_mm: bool = True,
 ) -> MoE:
     """
     Build the MoE layer associated with the Llama model.
@@ -631,6 +639,7 @@ def lora_llama4_decoder(
         raise ValueError(
             "Must pass local_chunk_size when enabling local chunked attention"
         )
+
     head_dim = embed_dim // num_heads
     num_kv_heads = num_kv_heads if num_kv_heads else num_heads
     if use_scaled_rope:
@@ -649,7 +658,6 @@ def lora_llama4_decoder(
         )
     layers = []
     for i in range(num_layers):
-
         mask_mod = None
         if skip_rope_interval is not None and (i + 1) % skip_rope_interval != 0:
             mask_mod = partial(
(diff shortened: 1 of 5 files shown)


torchtune/models/llama4/_component_builders.py
```diff
@@ -180,6 +181,7 @@ def llama4_decoder(
     num_experts: int = 16,
     experts_per_token: int = 1,
     use_shared_expert: bool = True,
+    use_grouped_mm: bool = True,
```
pbontrager, line 184: Do we actually need to expose this as an option? Are there cases where it's supported and not the fastest solution?

torchtune/modules/moe/experts.py
```diff
@@ -85,7 +73,52 @@ def forward(
...
+            torch.Tensor: tensor with shape ``(bsz * seq_len * experts_per_token, dim)``
+        """
+        if not self.use_grouped_mm:
+            return self._forward_no_grouped_mm(x, num_tokens_per_expert)
+
+        # grouped mm implementation
+        if num_tokens_per_expert is not None:
+            # https://github.com/pytorch/pytorch/pull/150374
+            # NOTE: torch._gouped_mm requires bf16 dtypes
+            #       and shapes to be multiple of 8
```
ebsmothers, line 102: nit: elsewhere it says multiple of 16

torchtune/modules/moe/experts.py
```diff
@@ -85,7 +73,52 @@ def forward(
...
+            offsets = torch.cumsum(num_tokens_per_expert, dim=0, dtype=torch.int32)
+            # grouped mm between a 2D tensor and a 3D tensor
+            assert x.dim() == 2
+        else:
```
ebsmothers, line 106: So this case is for expert choice?

torchtune/modules/moe/moe.py
```diff
@@ -133,6 +136,34 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
...
+
+            with torch.no_grad():
+                (
+                    permuted_indices,
+                    num_tokens_per_expert,
+                    _,
+                ) = generate_permute_indices(
+                    num_tokens_per_expert,
+                    self.experts.num_experts,
+                    1,
```
ebsmothers, line 156: This is just because we are fully replicating the experts over all ranks, right?

torchtune/modules/moe/indices.py
```diff
@@ -0,0 +1,353 @@
...
+
+    Returns:
+        permuted_indices: Tensor of indices that map original token order to the expert-grouped order.
+        m_sizes: aligned number of tokens for each expert (padded to alignment boundary).
+        m_offsets: Cumulative sum of m_sizes. The exclusive ending position for each expert's tokens.
+
+    Explanatory details:
+        `tokens_per_expert_group` is of shape (num_ranks * experts_per_rank,), for example:
+        From: |       rank 0      |       rank 1      |
+        To:   | E0 | E1 | E2 | E3 | E0 | E1 | E2 | E3 |
```
ebsmothers, line 185: Noob q on this diagram: does this mean that we actually have the same expert on multiple ranks (i.e. there are only 4 experts total)? I ask because the arg `experts_per_rank` seems to imply that we actually have different experts on different ranks (I know it's moot for the implementation in its current form anyways so not a huge deal).

torchtune/modules/moe/indices.py
```diff
@@ -0,0 +1,353 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
```
ebsmothers, line 1: We should also add an acknowledgment to the original torchtitan code somewhere. @lessw2020 lmk if any preference on the best way to do this

ebsmothers (commented):
Looks good! Left a handful of comments, lmk if anything is unclear. Excited to get grouped_mm landed!

ebsmothers (commented):
Looks good! Left a handful of comments, lmk if anything is unclear. Excited to get grouped_mm landed!

torchtune/modules/moe/indices.py
```diff
@@ -0,0 +1,353 @@
...
+# LICENSE file in the root directory of this source tree.
+
+# flake8: noqa
+
+import torch
+import triton
+import triton.language as tl
+
+
+__all__ = ["generate_permute_indices"]
```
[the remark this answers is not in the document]
IvanKobzarev: Yes, I think better to keep it private.

torchtune/modules/moe/indices.py
```diff
@@ -0,0 +1,353 @@
...
+        total_tokens_per_expert,  # 'skip logic' check for zero tokens
+        experts_per_rank,
+        num_ranks,
+        BLOCK_SIZE=block_size,
+    )
+    return permuted_indices
+
+
+# reference
+def fill_indices_cpu(
```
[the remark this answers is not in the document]
IvanKobzarev: I think it is more for a reference and testing. It is not used in llama4 recipes. But I think better to keep it as it is easier to read and understand what kernel does.

torchtune/models/llama4/_model_builders.py
```diff
@@ -78,6 +78,8 @@ def llama4_scout_17b_16e(
         norm_eps=1e-5,
         num_experts=16,
         use_shared_expert=True,
+        # Main switch llama4 use_grouped_mm
+        use_grouped_mm=True,
```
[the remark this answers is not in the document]
IvanKobzarev: Yes, I have not seen NaNs after the fixes.  I will move use_grouped_mm to config. I think it is a good idea that modelling logic should not know about engineering implementation use_grouped_mm or not.

torchtune/modules/moe/indices.py
IvanKobzarev: In our case each rank runs all the experts, but all mms happen in parallel. 
So experts_per_rank is always 16 (all experts).

So we are not bucketing experts by ranks. I think this logic will be similar, to what TorchRec does with embeddings, basically each expert works as a lookup Table.

torchtune/modules/moe/config.py
```diff
@@ -0,0 +1,26 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
```
ebsmothers: super nit but would maybe put this in `modules/moe/utils.py` or something (mainly because torchtune also has its own config system so the naming could be a bit confusing from that perspective).

PR #2761, opened 2025-05-22
Fix TP, enable test, silence noisy logs

ebsmothers:
Fixing a few things:

1) Our TP training has been broken since https://github.com/pytorch/pytorch/pull/149764. We were not correctly redistributing the tensors on their way into our attention layers: since we pass both x and y [here](https://github.com/pytorch/torchtune/blob/fa92c96ffcc79aff3a91dcb542788f3206906406/torchtune/modules/attention.py#L183-L184), we also need to convert `y` from sharded on sequence dim to replicated.
2) We didn't catch this because of how we were checking PyTorch version in our TP test. Doing direct comparison we will actually get `torch.__version__ < "2.8.0"` on 2.8 nightlies. So instead we can use base_version to get this test to run on nightlies again.
3) Also along the way I got a bunch of angry logs yelling at me about using activation offloading with streams even though I wasn't. So fixing the check 

Test plan: 

On latest nightlies:

```
pytest tests -m integration_test tests/recipes/test_full_finetune_distributed.py -k 'test_loss_2d_parallel'
...
========== 2 passed, 14 deselected in 40.54s ========
```

Also patched our CI job to run against nightlies just to ensure everything works on the CI runners:

Screenshot 2025-05-22 at 5 31 14 PM

```
tune run --nproc_per_node 4 full_finetune_distributed --config llama3/8B_full \
tensor_parallel_plan._component_=torchtune.models.llama3.base_llama_tp_plan tensor_parallel_dim=4 \
metric_logger=torchtune.training.metric_logging.WandBLogger metric_logger.project=test-tp \
metric_logger.name=llama3-8b max_steps_per_epoch=50
```
Screenshot 2025-05-22 at 4 10 57 PM

```
tune run --nproc_per_node 8 full_finetune_distributed --config llama4/scout_17B_16E_full \
metric_logger=torchtune.training.metric_logging.WandBLogger metric_logger.project=test-tp \
metric_logger.name=llama4-scout max_steps_per_epoch=50
```

Screenshot 2025-05-22 at 5 04 55 PM

Note that our generation TP plans do not use `PrepareModuleInput` for attention layers so we don't need to make any analogous changes to those.

### DIFF
--- a/.github/workflows/gpu_test.yaml
+++ b/.github/workflows/gpu_test.yaml
@@ -30,9 +30,8 @@ jobs:
         python-version: ['3.9', '3.10', '3.11']
         torch-version: ["stable", "nightly"]
         # Do not run against nightlies on PR
-        # TODO: REMOVE (just for debug)
-        # exclude:
-        #   - torch-version: ${{ github.event_name == 'pull_request' && 'nightly' }}
+        exclude:
+          - torch-version: ${{ github.event_name == 'pull_request' && 'nightly' }}
     steps:
       - name: Check out repo
         uses: actions/checkout@v4
--- a/.github/workflows/gpu_test.yaml
+++ b/.github/workflows/gpu_test.yaml
@@ -30,8 +30,9 @@ jobs:
         python-version: ['3.9', '3.10', '3.11']
         torch-version: ["stable", "nightly"]
         # Do not run against nightlies on PR
-        exclude:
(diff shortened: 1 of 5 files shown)


joecummings (approved):

pbontrager (approved):
Thanks for fixing this! Maybe if we had a skipif version guard util we might not make this mistake again.

PR #2765, opened 2025-05-27
[compile] Fix compile producing NaNs

IvanKobzarev:
* __->__ #2765

1/ moe/experts.py - dynamo compilation fails because of unbacked symints from tolist().

This results in IllegalMemoryAccess in flex_attention - disabling it for now, until the compilation is fixed to avoid hard failure

2/ flex_attention supports dynamic shapes compilation, and this also gives perf wins.

In my test for llama4 (+grouped_mm) on 4 gpus I got 
tps_per_gpu 471 -> 530

Running llama4 distributed compile for 50 iterations, no NaNs

Screenshot 2025-05-28 at 19 06 58


ebsmothers (approved):
Thanks for figuring this out!

Issue #2768, opened 2025-05-28
Deprecate code_llama2


krammnic:
**Why?**

[Codellama inference repository ](https://github.com/meta-llama/codellama) is not really updated for more than a year. This model is not really used in the modern research (even in the comparisons) and there is no business value in this model post-training. 

**Why not to still save it?**

Extra configs to update! Furthermore, according to the torchtune philosophy we want to be hackable. Hence, we don't need massive model zoo.

krammnic:
cc @pbontrager

PR #2770, opened 2025-05-28
Remove code_llama2

krammnic:
#### Context
- [ ] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

#### Changelog
* #2768 

#### Test plan

#### UX


pbontrager:
Can you do a git grep for code llama? Any references in the ReadMe, docs, and unit tests should be removed too.

ebsmothers:
@pbontrager, @krammnic yeah sorry I was a bit imprecise in my comment. I agree that the proper thing to do here is do a formal deprecation. In this case I am OK to let it slide for two reasons: (1) this model was minimally used even when we first landed it, I suspect the risk of breakages due to an outright deletion is very small, and (2) we shouldn't make Mark suffer for my screwup. So let's go ahead with the deletion in this case, but in the future we should do a proper deprecation cycle.

krammnic:
Hopefully removed everything else

ebsmothers (approved):
Thanks for doing this! Once CI is green we can merge

Issue #2717, opened 2025-05-11
Improve `compile_model` logging output


joecummings:
## Context

In `torch.compile`, the kernel exploration and generation happens in the first iteration. Therefore, when you compile a model, the first iteration takes more time than the rest of the training.

In `compile_model` function we should improve the error message to read something like: "Compiling model layers with torch.compile. Expect a relatively slower first step". 

There may be some other mentions of this sprinkled throughout the codebase. These should be cleaned up so that the only reference happens in `compile_model`

## Acceptance Criteria

Update or add a unit test that checks for this output in the logs.

Muhammad-Rebaal:
Hi @joecummings !

Could you assign this issue to me ? I'd like to solve this one ?

Thank You !

joecummings:
> Hi [@joecummings](https://github.com/joecummings) !
> 
> Could you assign this issue to me ? I'd like to solve this one ?
> 
> Thank You !

@Muhammad-Rebaal you got it - thank yoU!

PR #2777, opened 2025-06-01
Improve torch.compile log message in compile_model

Vasist10:
#### Context
- [ ] add a new feature
- [x] fix a bug
- [ ] other (please add here)

Closes https://github.com/pytorch/torchtune/issues/2717

#### Changelog
-Updated the log message in torchtune/utils/compile_model.py to read:
   `"Compiling model layers with torch.compile. Expect a relatively slower first step."`

-Removed similar/sprinkled references elsewhere in the codebase to centralize this message.

*

#### Test plan

#### UX

### DIFF
--- a/recipes/knowledge_distillation_single_device.py
+++ b/recipes/knowledge_distillation_single_device.py
@@ -608,8 +608,6 @@ class KDRecipeSingleDevice(FTRecipeInterface):
         The core training loop.
         """
 
-        
-
         # Initialize tokens count and running loss (for grad accumulation)
         t0 = time.perf_counter()
         running_class_loss = 0
--- a/recipes/knowledge_distillation_single_device.py
+++ b/recipes/knowledge_distillation_single_device.py
@@ -570,7 +570,7 @@ class KDRecipeSingleDevice(FTRecipeInterface):
 
     def _loss_step(
         self, batch: dict[str, torch.Tensor]
-    ) ->tuple[torch.Tensor, torch.Tensor]:
+    ) -> tuple[torch.Tensor, torch.Tensor]:
         # Both are shape [b, s]
         tokens, labels = batch["tokens"], batch["labels"]
 
--- a/recipes/knowledge_distillation_single_device.py
+++ b/recipes/knowledge_distillation_single_device.py
@@ -8,7 +8,7 @@ import sys
 import time
 
 from functools import partial
-from typing import Any, Optional, Union, Tuple
+from typing import Any, Optional, Union
 from warnings import warn
 
 import torch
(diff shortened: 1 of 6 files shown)


salmanmohammadi:
Hi @Vasist10. Thank you for this contribution. I think this looks good, however, the test may be overkill here. If you could please remove the test I'm happy to merge this.

We can also remove the extra log [here](https://github.com/pytorch/torchtune/blob/5ecae863c099bb40a8758b3244cb17836bdf2623/recipes/ppo_full_finetune_single_device.py#L932), and in a similar place across a few of our other recipes. It seems like you mentioned this in your PR but didn't include the changes - could you take a pass at removing the additional logs in the recipes?

Vasist10:
Hi @SalmanMohammadi ,  
I've made the following changes :

- Removed the `torch.compile` log messages across files : 
- Removed the test from `test_utils.py`  
- Fixed the tuple return type in `_loss_step` to avoid the Pylance warning

Thank you

recipes/knowledge_distillation_single_device.py
```diff
@@ -8,7 +8,7 @@
 import time
 
 from functools import partial
-from typing import Any, Optional, Union
+from typing import Any, Optional, Union, Tuple
```
salmanmohammadi: You may need to update your pre-commit hooks as we have moved to PEP 585.

Vasist10:
Hey @SalmanMohammadi ,

Apologies for the failing check  in `_loss_step` — that was my mistake.

Would you prefer I revert this change and switch it back to the older version? Let me know, and I’ll update it right away. Thanks again for your guidance!

recipes/knowledge_distillation_single_device.py
Vasist10: should i change it the previous version which was like this :
`def _loss_step(
    self, batch: dict[str, torch.Tensor]
) -> (torch.Tensor, torch.Tensor):
`

or should i change it to this ? : `
def _loss_step(
    self, batch: dict[str, torch.Tensor]
) -> tuple[torch.Tensor, torch.Tensor]:
`

i just want to clarify it

SalmanMohammadi (approved):
@Vasist10 No worries at all! This looks good to me pending green CI.

salmanmohammadi:
@Vasist10 Could you run the pre-commit hooks locally and ensure they pass?

Vasist10:
hey @SalmanMohammadi ,
sorry for the disturbance, Earlier, I had added a test in test_compile_model.py to check the log output using caplog, but we later removed it during the review process. Right now, the PR description still mentions it.
Would you prefer I remove that line for clarity, or is it okay to leave it as-is?

PR #2778, opened 2025-06-02
Fix typing in _grad_scaler.py for nightly compatibility

nathan-az:
#### Context
- [ ] add a new feature
- [x] fix a bug
- [ ] other (please add here)

#### Changelog
Fixes the typing changes introduced in https://github.com/pytorch/pytorch/pull/154801

If we need `main` working with the stable pytorch 2.7.0 release, we will have to either not type and selectively disable type checking, or define the type in the file rather than importing.

#### Test plan

#### UX

### DIFF
--- a/torchtune/training/_grad_scaler.py
+++ b/torchtune/training/_grad_scaler.py
@@ -10,9 +10,15 @@ from typing import Optional
 import torch
 from torch import nn, Tensor
 from torch.distributed.tensor import DTensor
-from torch.nn.utils.clip_grad import _no_grad, _TensorOrTensors
+from torch.nn.utils.clip_grad import _no_grad
 from torch.utils._foreach_utils import _device_has_foreach_support, _has_foreach_support
 from torchtune.utils._logging import deprecated
+
+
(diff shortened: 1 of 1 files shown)


nathan-az:
Stable CI fails (as expected). @joecummings do we want to force this through and accept that `main` won't work with stable 2.7.0?

Alternative is I can just copy the definition of `_TensorOrTensors`, rather than relying on the import until 2.8.0 releases.

joecummings:
> Stable CI fails (as expected). @joecummings do we want to force this through and accept that `main` won't work with stable 2.7.0?
> 
> Alternative is I can just copy the definition of `_TensorOrTensors`, rather than relying on the import until 2.8.0 releases.

Since we are not on the same release cycle as PyTorch core, we have to ensure that our main at least _runs_ with stable and nightly. Although hacky, I would rely on a version check to do the import b/c that's easier to go back and crtl-find and replace when 2.8 is out.

nathan-az:
No worries, done. Let's see if they pass now.

Issue #2779, opened 2025-06-02
Proposal: reuse methods in recipes


Andrei-Aksionov:
Hi there 👋 

Back in March, I left a [comment](https://github.com/pytorch/torchtune/pull/2447#issuecomment-2710986457) under a PR that tries to refactor recipes, to make them simpler.

My idea (which is quite straightforward) is to avoid recipes that include complete code, essentially deviating from the single-file approach.
Let’s consider full-time fine-tuning and LoRA fine-tuning as examples. In this case, only the model instantiation and checkpoint saving should differ.
The proposal is to have full fine-tuning as the basis, and in the LoRA variant, only override the LoRA-specific methods.

The advantages and disadvantages, plus the example code, can be found in this [PR](https://github.com/pytorch/torchtune/pull/2780).

---

Now, let’s discuss another matter.

If we adhere to this approach and:
1. Maintain full fine-tuning as the basis
2. Override only the LoRA-specific methods in the LoRA fine-tuning variant
3. Override only the distributed-specific methods in the full fine-tuning variant

… what should we do with the LoRA fine-tuning distributed variant?

- Option A: Override both the LoRA- and distributed-specific methods, which essentially means duplicating the entire code, and that brings us back to square one.
- Option B: Reuse the LoRA-specific code from the LoRA recipe and the distributed-specific code from the distributed recipe. In this case, to read the LoRA distributed recipe, we would need to jump between three files, which is too cumbersome and not ideal.

Another option would be to create hardware environment-agnostic code. Regardless of whether there is a single GPU or multiple GPUs, the code will remain the same. All the logic will be encapsulated in a separate class.

This approach would eliminate the need for separate recipes for distributed cases.

----

The question is what the core team thinks about it?

pbontrager:
Hi @Andrei-Aksionov having single file implementations of our recipes is a key feature that we like to offer in order for people to be able to easily copy and edit the single file they're working on and not have to span and look through multiple files. That said, for ease of maintainability and to better keep features in sync across multiple recipes, it's good to revisit this from time to time. We plan to iterate on our recipe design later in the year, so please keep a look out for that and we'll definitely consider these proposals.

[...]

PR #2780, opened 2025-06-02
[WIP] Full fine-tune and LoRA fine-tune simplifications

Andrei-Aksionov:
Hey there 👋 

> [!NOTE]
> This is an implentation for the idea proposed in #2779 

This is an implementation of my [comment](https://github.com/pytorch/torchtune/pull/2447#issuecomment-2710986457) back from March.

*The objective is to enhance the maintainability and, in certain aspects, the readability of recipes.*

If we consider the differences between a full fine-tuning and LoRA variant, the primary distinctions should lie in:
1. The way a model is instantiated
2. The way a checkpoint is saved

This draft/PR proposes precisely this:
1. Utilizing the full fine-tuning recipe as a foundation
2. Overriding only the LoRA-specific methods

----

This approach offers several advantages:
1. **More maintainable code**: eliminating the need for identical modifications across multiple locations.
2. **Up-to-date code**: sometimes changes made to the full version may not have been incorporated into the LoRA variant.
5. **More readable code**: facilitating a clearer understanding of the unique aspects of LoRA fine-tuning (in terms of the recipe) for anyone interested in exploring it.

However, there are a few drawbacks to consider:
1. **Increased readability challenges**: while the single-file approach allows users to read the whole recipe as a single file, this approach requires opening two files (ideally side-by-side) to read the recipe.
2. **Reduced hackability**: if someone intends to use the LoRA recipe as a starting point and subsequently build upon it, the first two files must be merged. This process is straightforward, as methods need to be replaced as a whole, requiring minimal mental effort and taking a few minutes at most.

----

As with any solution, it has its own set of drawbacks. Despite these disadvantages, this approach may still be worthwhile.

### DIFF
--- a/recipes/configs/llama2/7B_lora_single_device.yaml
+++ b/recipes/configs/llama2/7B_lora_single_device.yaml
@@ -64,6 +64,7 @@ optimizer:
 lr_scheduler:
   _component_: torchtune.training.lr_schedulers.get_cosine_schedule_with_warmup
   num_warmup_steps: 100
+optimizer_in_bwd: False  # True saves memory. Requires gradient_accumulation_steps=1
 
 loss:
   _component_: torchtune.modules.loss.LinearCrossEntropyLoss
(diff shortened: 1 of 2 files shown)


Andrei-Aksionov:
Oops, it indented to be a draft 🙈

Andrei-Aksionov:
Closing, since in this [comment](https://github.com/pytorch/torchtune/issues/2779#issuecomment-2935967176) it was clearly stated that single-file implementation (with full copy of a code) is the way to go (for now, at least).

PR #2782, opened 2025-06-03
Enable loss parallel, Ungate FP8

nathan-az:
#### Context
- [x] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

#### Changelog
* Ungate FP8 + TP, and clean up LLaMA-3 TP plans
* Enables loss parallel
* Enables compiling autograd
    * Note that I have not been able to get this working so default to False, but leave it in for debugging purposes

Loss parallelism is the main feature. Loss curves look healthy, but memory utilisation is significantly lower now. This is also compatible (tested) with FP8.

## Peak active memory usage scales aggressively (linearly) has tensor parallelism increases without loss parallelism

image

## Peak active memory usage scales much more generously (still linear) with loss parallelism enabled

image

In the tp8 case (bs=8, seq len=8192), loss parallelism decreases active memory peak by about 30GB, about 35%. TPS remains the same.

## ~Compiling autograd doesn't work~ (removed autograd compile for now)

Unfortunately compiling autograd doesn't seem to work and throws an FSDP error that I don't know how to debug. The error seems to indicate it's caused by some graph break during checkpointing. We can remove this from the PR, or leave it in for future debugging, defaulted to `false`. Running with `TORCHDYNAMO_VERBOSE=1 TORCHINDUCTOR_AUTOGRAD_CACHE=0` yielded:

<details>

```
[rank7]: Traceback (most recent call last):
[rank7]:   File "/workspace/torchtune/recipes/full_finetune_distributed.py", line 1118, in <module>
[rank7]:     sys.exit(recipe_main())
[rank7]:              ^^^^^^^^^^^^^
[rank7]:   File "/workspace/torchtune/torchtune/config/_parse.py", line 99, in wrapper
[rank7]:     sys.exit(recipe_main(conf))
[rank7]:              ^^^^^^^^^^^^^^^^^
[rank7]:   File "/workspace/torchtune/recipes/full_finetune_distributed.py", line 1113, in recipe_main
[rank7]:     recipe.train()
[rank7]:   File "/workspace/torchtune/recipes/full_finetune_distributed.py", line 957, in train
[rank7]:     current_loss.backward()
[rank7]:   File "/opt/conda/lib/python3.11/site-packages/torch/_tensor.py", line 648, in backward
[rank7]:     torch.autograd.backward(
[rank7]:   File "/opt/conda/lib/python3.11/site-packages/torch/autograd/__init__.py", line 354, in backward
[rank7]:     _engine_run_backward(
[rank7]:   File "/opt/conda/lib/python3.11/site-packages/torch/autograd/graph.py", line 829, in _engine_run_backward
[rank7]:     return Variable._execution_engine.run_backward(  # Calls into the C++ engine to run the backward pass
[rank7]:            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank7]:   File "/opt/conda/lib/python3.11/site-packages/torch/_dynamo/compiled_autograd.py", line 1041, in runtime_wrapper
[rank7]:     out = compiled_fn(
[rank7]:           ^^^^^^^^^^^^
[rank7]:   File "/opt/conda/lib/python3.11/site-packages/torch/_dynamo/eval_frame.py", line 372, in __call__
[rank7]:     return super().__call__(*args, **kwargs)
[rank7]:            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank7]:   File "/opt/conda/lib/python3.11/site-packages/torch/nn/modules/module.py", line 1767, in _wrapped_call_impl
[rank7]:     return self._call_impl(*args, **kwargs)
[rank7]:            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank7]:   File "/opt/conda/lib/python3.11/site-packages/torch/nn/modules/module.py", line 1778, in _call_impl
[rank7]:     return forward_call(*args, **kwargs)
[rank7]:            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank7]:   File "/opt/conda/lib/python3.11/site-packages/torch/_dynamo/eval_frame.py", line 699, in compile_wrapper
[rank7]:     return fn(*args, **kwargs)
[rank7]:            ^^^^^^^^^^^^^^^^^^^
[rank7]:   File "/opt/conda/lib/python3.11/site-packages/torch/fx/graph_module.py", line 840, in call_wrapped
[rank7]:     return self._wrapped_call(self, *args, **kwargs)
[rank7]:            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank7]:   File "/opt/conda/lib/python3.11/site-packages/torch/fx/graph_module.py", line 416, in __call__
[rank7]:     raise e
[rank7]:   File "/opt/conda/lib/python3.11/site-packages/torch/fx/graph_module.py", line 403, in __call__
[rank7]:     return super(self.cls, obj).__call__(*args, **kwargs)  # type: ignore[misc]
[rank7]:            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank7]:   File "/opt/conda/lib/python3.11/site-packages/torch/nn/modules/module.py", line 1767, in _wrapped_call_impl
[rank7]:     return self._call_impl(*args, **kwargs)
[rank7]:            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank7]:   File "/opt/conda/lib/python3.11/site-packages/torch/nn/modules/module.py", line 1778, in _call_impl
[rank7]:     return forward_call(*args, **kwargs)
[rank7]:            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank7]:   File "/opt/conda/lib/python3.11/site-packages/torch/_dynamo/convert_frame.py", line 1469, in __call__
[rank7]:     return self._torchdynamo_orig_callable(
[rank7]:            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank7]:   File "/opt/conda/lib/python3.11/site-packages/torch/_dynamo/convert_frame.py", line 625, in __call__
[rank7]:     return _compile(
[rank7]:            ^^^^^^^^^
[rank7]:   File "/opt/conda/lib/python3.11/site-packages/torch/_dynamo/convert_frame.py", line 1092, in _compile
[rank7]:     guarded_code = compile_inner(code, one_graph, hooks, transform)
[rank7]:                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank7]:   File "/opt/conda/lib/python3.11/site-packages/torch/_utils_internal.py", line 97, in wrapper_function
[rank7]:     return function(*args, **kwargs)
[rank7]:            ^^^^^^^^^^^^^^^^^^^^^^^^^
[rank7]:   File "/opt/conda/lib/python3.11/site-packages/torch/_dynamo/convert_frame.py", line 779, in compile_inner
[rank7]:     return _compile_inner(code, one_graph, hooks, transform)
[rank7]:            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank7]:   File "/opt/conda/lib/python3.11/site-packages/torch/_dynamo/convert_frame.py", line 818, in _compile_inner
[rank7]:     out_code = transform_code_object(code, transform)
[rank7]:                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank7]:   File "/opt/conda/lib/python3.11/site-packages/torch/_dynamo/bytecode_transformation.py", line 1424, in transform_code_object
[rank7]:     transformations(instructions, code_options)
[rank7]:   File "/opt/conda/lib/python3.11/site-packages/torch/_dynamo/convert_frame.py", line 265, in _fn
[rank7]:     return fn(*args, **kwargs)
[rank7]:            ^^^^^^^^^^^^^^^^^^^
[rank7]:   File "/opt/conda/lib/python3.11/site-packages/torch/_dynamo/convert_frame.py", line 743, in transform
[rank7]:     tracer.run()
[rank7]:   File "/opt/conda/lib/python3.11/site-packages/torch/_dynamo/symbolic_convert.py", line 3484, in run
[rank7]:     super().run()
[rank7]:   File "/opt/conda/lib/python3.11/site-packages/torch/_dynamo/symbolic_convert.py", line 1359, in run
[rank7]:     while self.step():
[rank7]:           ^^^^^^^^^^^
[rank7]:   File "/opt/conda/lib/python3.11/site-packages/torch/_dynamo/symbolic_convert.py", line 1263, in step
[rank7]:     self.dispatch_table[inst.opcode](self, inst)
[rank7]:   File "/opt/conda/lib/python3.11/site-packages/torch/_dynamo/symbolic_convert.py", line 831, in wrapper
[rank7]:     return inner_fn(self, inst)
[rank7]:            ^^^^^^^^^^^^^^^^^^^^
[rank7]:   File "/opt/conda/lib/python3.11/site-packages/torch/_dynamo/symbolic_convert.py", line 2903, in CALL
[rank7]:     self._call(inst)
[rank7]:   File "/opt/conda/lib/python3.11/site-packages/torch/_dynamo/symbolic_convert.py", line 2897, in _call
[rank7]:     self.call_function(fn, args, kwargs)
[rank7]:   File "/opt/conda/lib/python3.11/site-packages/torch/_dynamo/symbolic_convert.py", line 1189, in call_function
[rank7]:     self.push(fn.call_function(self, args, kwargs))  # type: ignore[arg-type]
[rank7]:               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank7]:   File "/opt/conda/lib/python3.11/site-packages/torch/_dynamo/variables/lazy.py", line 201, in realize_and_forward
[rank7]:     return getattr(self.realize(), name)(*args, **kwargs)
[rank7]:            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank7]:   File "/opt/conda/lib/python3.11/site-packages/torch/_dynamo/variables/functions.py", line 516, in call_function
[rank7]:     return super().call_function(tx, args, kwargs)
[rank7]:            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank7]:   File "/opt/conda/lib/python3.11/site-packages/torch/_dynamo/variables/functions.py", line 291, in call_function
[rank7]:     return tx.inline_user_function_return(self, [*self.self_args(), *args], kwargs)
[rank7]:            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank7]:   File "/opt/conda/lib/python3.11/site-packages/torch/_dynamo/symbolic_convert.py", line 1206, in inline_user_function_return
[rank7]:     return InliningInstructionTranslator.inline_call(self, fn, args, kwargs)
[rank7]:            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank7]:   File "/opt/conda/lib/python3.11/site-packages/torch/_dynamo/symbolic_convert.py", line 3678, in inline_call
[rank7]:     return tracer.inline_call_()
[rank7]:            ^^^^^^^^^^^^^^^^^^^^^
[rank7]:   File "/opt/conda/lib/python3.11/site-packages/torch/_dynamo/symbolic_convert.py", line 3881, in inline_call_
[rank7]:     self.run()
[rank7]:   File "/opt/conda/lib/python3.11/site-packages/torch/_dynamo/symbolic_convert.py", line 1359, in run
[rank7]:     while self.step():
[rank7]:           ^^^^^^^^^^^
[rank7]:   File "/opt/conda/lib/python3.11/site-packages/torch/_dynamo/symbolic_convert.py", line 1263, in step
[rank7]:     self.dispatch_table[inst.opcode](self, inst)
[rank7]:   File "/opt/conda/lib/python3.11/site-packages/torch/_dynamo/symbolic_convert.py", line 831, in wrapper
[rank7]:     return inner_fn(self, inst)
[rank7]:            ^^^^^^^^^^^^^^^^^^^^
[rank7]:   File "/opt/conda/lib/python3.11/site-packages/torch/_dynamo/symbolic_convert.py", line 2221, in CALL_FUNCTION_EX
[rank7]:     self.call_function(fn, argsvars.items, kwargsvars)
[rank7]:   File "/opt/conda/lib/python3.11/site-packages/torch/_dynamo/symbolic_convert.py", line 1189, in call_function
[rank7]:     self.push(fn.call_function(self, args, kwargs))  # type: ignore[arg-type]
[rank7]:               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank7]:   File "/opt/conda/lib/python3.11/site-packages/torch/_dynamo/variables/functions.py", line 1477, in call_function
[rank7]:     unimplemented_v2(
[rank7]:   File "/opt/conda/lib/python3.11/site-packages/torch/_dynamo/exc.py", line 521, in unimplemented_v2
[rank7]:     raise Unsupported(msg)
[rank7]: torch._dynamo.exc.Unsupported: Attempted to call function marked as skipped
[rank7]:   Explanation: Dynamo developers have intentionally marked that the function `_checkpoint_hook.__init__.<locals>.unpack_hook` in file `/opt/conda/lib/python3.11/site-packages/torch/utils/checkpoint.py` should not be traced.
[rank7]:   Hint: Avoid calling the function `_checkpoint_hook.__init__.<locals>.unpack_hook`.
[rank7]:   Hint: Apply `@torch._dynamo.dont_skip_tracing` to the function `_checkpoint_hook.__init__.<locals>.unpack_hook` to force tracing into the function. More graph breaks may occur as a result of attempting to trace into the function.
[rank7]:   Hint: Please file an issue to PyTorch.

[rank7]:   Developer debug context: module: torch.utils.checkpoint, qualname: _checkpoint_hook.__init__.<locals>.unpack_hook, skip reason: <missing reason>


[rank7]: from user code:
[rank7]:    File "<eval_with_key>.35", line 1818, in forward
[rank7]:     call_hook_2 = torch__dynamo_external_utils_call_hook(getitem_13144, getitem_13145, hook_type = 'unpack_hook');  getitem_13144 = getitem_13145 = None
[rank7]:   File "/opt/conda/lib/python3.11/site-packages/torch/_dynamo/external_utils.py", line 81, in call_hook
[rank7]:     result = hook(*args)
```

</details>

#### Test plan

#### UX

### DIFF
--- a/recipes/full_finetune_distributed.py
+++ b/recipes/full_finetune_distributed.py
@@ -341,13 +341,11 @@ class FullFinetuneRecipeDistributed(FTRecipeInterface):
         self._compile_loss = compile_bool
         self._compile_optimizer_step = compile_bool
         self._compile_scale_grads = compile_bool
-        self._compile_autograd = False  # TODO: work out why this isn't working. We leave it parameterisable for debug purposes.
         if isinstance(compile, DictConfig):
             self._compile_model = compile.get("model", True)
             self._compile_loss = compile.get("loss", True)
             self._compile_optimizer_step = compile.get("optimizer_step", False)
             self._compile_scale_grads = compile.get("scale_grads", True)
-            self._compile_autograd = compile.get("autograd", False)
         if self._compile_model:
             # Capture scalar outputs is required to compile MoE
             torch._dynamo.config.capture_scalar_outputs = True
@@ -706,7 +704,6 @@ class FullFinetuneRecipeDistributed(FTRecipeInterface):
         # remaining context managers for fwd/bwd
         self.train_context = training.get_train_context(
             enable_loss_parallel=self.parallel_dims.loss_parallel_enabled,
-            enable_compiled_autograd=self._compile_autograd,
         )
 
         # Ensure no params and buffers are on meta device
@@ -1024,6 +1021,7 @@ class FullFinetuneRecipeDistributed(FTRecipeInterface):
                                 num_tokens / self.parallel_dims.non_data_parallel_size
                             )
                             / (time_per_step * self.world_size),
+                            "num_tokens": num_tokens,
                         }
                         if self._log_peak_memory_stats:
                             log_dict.update(
--- a/recipes/full_finetune_distributed.py
+++ b/recipes/full_finetune_distributed.py
@@ -357,6 +357,24 @@ class FullFinetuneRecipeDistributed(FTRecipeInterface):
                 self._grad_scaler, backend=self._compile_backend
             )
 
+        # initialize loss
+        self._loss_fn = config.instantiate(cfg.loss)
+        if isinstance(self._loss_fn, SFTLoss):
+            self._loss_fn.enable_loss_parallel = (
+                self.parallel_dims.loss_parallel_enabled
+            )
+
+        # Whether to use the ctx manager. If the loss fn has the property, use that. Otherwise, assume it is supported.
+        # Useful if, for example, user opts to use the basic CrossEntropyLoss() instead of an SFTLoss subclass.
+        self.use_loss_parallel_ctx_manager = (
+            self.parallel_dims.loss_parallel_enabled
+            and getattr(
+                self._loss_fn,
+                "use_loss_parallel_ctx_manager",
+                True,
+            )
+        )
+
         self._model = self._setup_model(
             cfg_model=cfg.model,
             enable_activation_checkpointing=self._enable_activation_checkpointing,
@@ -417,11 +435,6 @@ class FullFinetuneRecipeDistributed(FTRecipeInterface):
 
             # Update the recipe state from the checkpoint state dict.
             self._update_recipe_state(checkpoint_dict)
-
-        # initialize loss
-        self._loss_fn = config.instantiate(cfg.loss)
-        if hasattr(self._loss_fn, "loss_parallel"):
-            self._loss_fn.loss_parallel = self.parallel_dims.loss_parallel_enabled
 
         if isinstance(self._loss_fn, SFTLoss):
             self._loss_fn.set_model_output(self._model)
@@ -626,9 +639,11 @@ class FullFinetuneRecipeDistributed(FTRecipeInterface):
                 self.tp_plan = config.instantiate(
                     self.tp_plan,
                     model=model,
-                    loss_parallel=self.parallel_dims.loss_parallel_enabled,
                     enable_fp8_training=self._enable_fp8_training,
                 )
+                if isinstance(self._loss_fn, SFTLoss):
+                    self.tp_plan = self._loss_fn.patch_tp_plan(self.tp_plan)
+
             parallelize_module(
                 model,
                 self.world_mesh["tp"],
@@ -704,7 +719,7 @@ class FullFinetuneRecipeDistributed(FTRecipeInterface):
         )
         # remaining context managers for fwd/bwd
         self.train_context = training.get_train_context(
-            enable_loss_parallel=self.parallel_dims.loss_parallel_enabled,
+            enable_loss_parallel=self.use_loss_parallel_ctx_manager,
         )
 
         # Ensure no params and buffers are on meta device
--- a/recipes/full_finetune_distributed.py
+++ b/recipes/full_finetune_distributed.py
@@ -1037,7 +1037,6 @@ class FullFinetuneRecipeDistributed(FTRecipeInterface):
                                 num_tokens / self.parallel_dims.non_data_parallel_size
                             )
                             / (time_per_step * self.world_size),
-                            "num_tokens": num_tokens,
                         }
                         if self._log_peak_memory_stats:
                             log_dict.update(
--- a/recipes/full_finetune_distributed.py
+++ b/recipes/full_finetune_distributed.py
@@ -161,7 +161,6 @@ class FullFinetuneRecipeDistributed(FTRecipeInterface):
         self.cp_degree = cfg.get("context_parallel_dim", 1)
         data_shard = cfg.get("data_parallel_shard_dim", -1)  # -1 means to infer
         data_replicate = cfg.get("data_parallel_replicate_dim", 1)
-        enable_loss_parallel = cfg.get("enable_loss_parallel", True)
 
         # Set up n-d device mesh
         self.parallel_dims = training.ParallelDims(
@@ -170,7 +169,6 @@ class FullFinetuneRecipeDistributed(FTRecipeInterface):
             tp=self.tp_degree,
             cp=self.cp_degree,
             world_size=self.world_size,
-            enable_loss_parallel=enable_loss_parallel,
         )
         self.world_mesh = self.parallel_dims.build_mesh(device_type=device_type)
         if self.parallel_dims.dp_enabled:
@@ -359,21 +357,14 @@ class FullFinetuneRecipeDistributed(FTRecipeInterface):
 
         # initialize loss
         self._loss_fn = config.instantiate(cfg.loss)
-        if isinstance(self._loss_fn, SFTLoss):
-            self._loss_fn.enable_loss_parallel = (
-                self.parallel_dims.loss_parallel_enabled
-            )
 
         # Whether to use the ctx manager. If the loss fn has the property, use that. Otherwise, assume it is not supported.
         # Currently our TP plans assume replicating on the output of `output` so without a custom loss class patching the
         # TP plan, there is no memory benefit to the ctx manager
-        self.use_loss_parallel_ctx_manager = (
-            self.parallel_dims.loss_parallel_enabled
-            and getattr(
-                self._loss_fn,
-                "use_loss_parallel_ctx_manager",
-                False,
-            )
+        self.use_loss_parallel_ctx_manager = self.parallel_dims.tp_enabled and getattr(
+            self._loss_fn,
+            "tp_requires_loss_parallel_ctx_manager",
+            False,
         )
 
         self._model = self._setup_model(
@@ -643,6 +634,7 @@ class FullFinetuneRecipeDistributed(FTRecipeInterface):
                     enable_fp8_training=self._enable_fp8_training,
                 )
                 if isinstance(self._loss_fn, SFTLoss):
+                    self._loss_fn.tp_enabled = True
                     self.tp_plan = self._loss_fn.patch_tp_plan(self.tp_plan)
 
             parallelize_module(
--- a/recipes/full_finetune_distributed.py
+++ b/recipes/full_finetune_distributed.py
@@ -627,6 +627,7 @@ class FullFinetuneRecipeDistributed(FTRecipeInterface):
                     self.tp_plan,
                     model=model,
                     loss_parallel=self.parallel_dims.loss_parallel_enabled,
+                    enable_fp8_training=self._enable_fp8_training,
                 )
             parallelize_module(
                 model,
--- a/recipes/full_finetune_distributed.py
+++ b/recipes/full_finetune_distributed.py
@@ -365,8 +365,8 @@ class FullFinetuneRecipeDistributed(FTRecipeInterface):
             )
 
         # Whether to use the ctx manager. If the loss fn has the property, use that. Otherwise, assume it is not supported.
-        # Currently our TP plans assume replicating on the output of `output` so without a custom loss class, there is
-        # no memory benefit to the ctx manager
+        # Currently our TP plans assume replicating on the output of `output` so without a custom loss class patching the
+        # TP plan, there is no memory benefit to the ctx manager
         self.use_loss_parallel_ctx_manager = (
             self.parallel_dims.loss_parallel_enabled
             and getattr(
--- a/recipes/full_finetune_distributed.py
+++ b/recipes/full_finetune_distributed.py
@@ -357,10 +357,6 @@ class FullFinetuneRecipeDistributed(FTRecipeInterface):
 
         # initialize loss
         self._loss_fn = config.instantiate(cfg.loss)
-
-        # Whether to use the ctx manager. If the loss fn has the property, use that. Otherwise, assume it is not supported.
-        # Currently our TP plans assume replicating on the output of `output` so without a custom loss class patching the
-        # TP plan, there is no memory benefit to the ctx manager
         self.use_loss_parallel_ctx_manager = self.parallel_dims.tp_enabled and getattr(
             self._loss_fn,
             "tp_requires_loss_parallel_ctx_manager",
--- a/recipes/full_finetune_distributed.py
+++ b/recipes/full_finetune_distributed.py
@@ -364,14 +364,15 @@ class FullFinetuneRecipeDistributed(FTRecipeInterface):
                 self.parallel_dims.loss_parallel_enabled
             )
 
-        # Whether to use the ctx manager. If the loss fn has the property, use that. Otherwise, assume it is supported.
-        # Useful if, for example, user opts to use the basic CrossEntropyLoss() instead of an SFTLoss subclass.
+        # Whether to use the ctx manager. If the loss fn has the property, use that. Otherwise, assume it is not supported.
+        # Currently our TP plans assume replicating on the output of `output` so without a custom loss class, there is
+        # no memory benefit to the ctx manager
         self.use_loss_parallel_ctx_manager = (
             self.parallel_dims.loss_parallel_enabled
             and getattr(
                 self._loss_fn,
                 "use_loss_parallel_ctx_manager",
-                True,
+                False,
             )
         )
 
--- a/recipes/full_finetune_distributed.py
+++ b/recipes/full_finetune_distributed.py
@@ -161,6 +161,7 @@ class FullFinetuneRecipeDistributed(FTRecipeInterface):
         self.cp_degree = cfg.get("context_parallel_dim", 1)
         data_shard = cfg.get("data_parallel_shard_dim", -1)  # -1 means to infer
         data_replicate = cfg.get("data_parallel_replicate_dim", 1)
+        enable_loss_parallel = cfg.get("enable_loss_parallel", True)
 
         # Set up n-d device mesh
         self.parallel_dims = training.ParallelDims(
@@ -169,6 +170,7 @@ class FullFinetuneRecipeDistributed(FTRecipeInterface):
             tp=self.tp_degree,
             cp=self.cp_degree,
             world_size=self.world_size,
+            enable_loss_parallel=enable_loss_parallel,
         )
         self.world_mesh = self.parallel_dims.build_mesh(device_type=device_type)
         if self.parallel_dims.dp_enabled:
@@ -339,6 +341,7 @@ class FullFinetuneRecipeDistributed(FTRecipeInterface):
         self._compile_loss = compile_bool
         self._compile_optimizer_step = compile_bool
         self._compile_scale_grads = compile_bool
+        self._compile_autograd = False  # TODO: work out why this isn't working. We leave it parameterisable for debug purposes.
         if isinstance(compile, DictConfig):
             self._compile_model = compile.get("model", True)
             self._compile_loss = compile.get("loss", True)
@@ -418,6 +421,9 @@ class FullFinetuneRecipeDistributed(FTRecipeInterface):
 
         # initialize loss
         self._loss_fn = config.instantiate(cfg.loss)
+        if hasattr(self._loss_fn, "loss_parallel"):
+            self._loss_fn.loss_parallel = self.parallel_dims.loss_parallel_enabled
+
         if isinstance(self._loss_fn, SFTLoss):
             self._loss_fn.set_model_output(self._model)
 
@@ -603,11 +609,6 @@ class FullFinetuneRecipeDistributed(FTRecipeInterface):
                 raise RuntimeError(
                     "Float8 fine-tuning requires PyTorch 2.8.0.dev20250318 or later."
                 )
-            if self.tp_plan is not None:
-                raise ValueError(
-                    "FP8 training does not support tensor parallelism yet. "
-                    "This will be enabled in the near future."
-                )
             if self.cp_degree > 1:
                 raise ValueError(
                     "Context Parallel for fp8 training is not currently supported"
@@ -626,6 +627,7 @@ class FullFinetuneRecipeDistributed(FTRecipeInterface):
                 self.tp_plan = config.instantiate(
                     self.tp_plan,
                     model=model,
+                    loss_parallel=self.parallel_dims.loss_parallel_enabled,
                 )
             parallelize_module(
                 model,
@@ -674,13 +676,6 @@ class FullFinetuneRecipeDistributed(FTRecipeInterface):
                 dp_mesh=self.world_mesh[dp_mesh_dim_names],
             )
 
-        # Define context manager for context parallelism
-        self.context_parallel_manager = training.get_context_parallel_manager(
-            enabled=self.cp_degree > 1,
-            world_mesh=self.world_mesh,
-            model=model,
-        )
-
         with training.set_default_dtype(self._dtype), self._device:
             for m in model.modules():
                 # RoPE is not covered in state dict
@@ -700,6 +695,17 @@ class FullFinetuneRecipeDistributed(FTRecipeInterface):
         # activation offloading
         self.activations_handling_ctx = training.get_act_offloading_ctx_manager(
             model, enable_activation_offloading, activation_offloading_use_streams
+        )
+        # context parallel
+        self.optional_context_parallel_manager = training.get_context_parallel_manager(
+            enabled=self.cp_degree > 1,
+            world_mesh=self.world_mesh,
+            model=model,
+        )
+        # remaining context managers for fwd/bwd
+        self.train_context = training.get_train_context(
+            enable_loss_parallel=self.parallel_dims.loss_parallel_enabled,
+            enable_compiled_autograd=self._compile_autograd,
         )
 
         # Ensure no params and buffers are on meta device
@@ -934,9 +940,11 @@ class FullFinetuneRecipeDistributed(FTRecipeInterface):
                 ).sum()
                 num_tokens += current_num_tokens
 
-                # Loss is normalized by default so we multiply by the number of tokens
-                # This way we can normalize by the total number of tokens if we're accumulating gradients
-                with self.context_parallel_manager(list(batch.values())):
+                with self.train_context(
+                    self.optional_context_parallel_manager(list(batch.values()))
+                ):
+                    # Loss is normalized by default so we multiply by the number of tokens
+                    # This way we can normalize by the total number of tokens if we're accumulating gradients
                     current_loss = self._loss_step(batch) * current_num_tokens
                     running_loss += current_loss
                     # For optimizer in backward, we need to normalize before calling backward
--- a/recipes/full_finetune_distributed.py
+++ b/recipes/full_finetune_distributed.py
@@ -701,7 +701,7 @@ class FullFinetuneRecipeDistributed(FTRecipeInterface):
             model, enable_activation_offloading, activation_offloading_use_streams
         )
         # context parallel
-        self.optional_context_parallel_manager = training.get_context_parallel_manager(
+        self.context_parallel_manager = training.get_context_parallel_manager(
             enabled=self.cp_degree > 1,
             world_mesh=self.world_mesh,
             model=model,
@@ -944,7 +944,7 @@ class FullFinetuneRecipeDistributed(FTRecipeInterface):
                 num_tokens += current_num_tokens
 
                 with self.train_context(
-                    self.optional_context_parallel_manager(list(batch.values()))
+                    self.context_parallel_manager(list(batch.values()))
                 ):
                     # Loss is normalized by default so we multiply by the number of tokens
                     # This way we can normalize by the total number of tokens if we're accumulating gradients
--- a/recipes/full_finetune_distributed.py
+++ b/recipes/full_finetune_distributed.py
@@ -347,6 +347,7 @@ class FullFinetuneRecipeDistributed(FTRecipeInterface):
             self._compile_loss = compile.get("loss", True)
             self._compile_optimizer_step = compile.get("optimizer_step", False)
             self._compile_scale_grads = compile.get("scale_grads", True)
+            self._compile_autograd = compile.get("autograd", False)
         if self._compile_model:
             # Capture scalar outputs is required to compile MoE
             torch._dynamo.config.capture_scalar_outputs = True
--- a/torchtune/models/llama3/_parallelism.py
+++ b/torchtune/models/llama3/_parallelism.py
@@ -26,6 +26,7 @@ def _get_base_llama_tp_training_plan(
     layerwise_colwise_parallel_cls: type[ParallelStyle] = ColwiseParallel,
     layerwise_rowwise_parallel_cls: type[ParallelStyle] = RowwiseParallel,
     layerwise_prepare_module_input_cls: type[ParallelStyle] = PrepareModuleInput,
+    loss_parallel: bool = False,
 ) -> dict[str, ParallelStyle]:
     """
     Define the Tensor Parallel plan for Llama3 model, which will also be shared with 3.1, 3.2, and 3.3 models.
@@ -35,7 +36,11 @@ def _get_base_llama_tp_training_plan(
             input_layouts=Replicate(), output_layouts=Shard(1)
         ),
         "norm": SequenceParallel(),
-        "output": ColwiseParallel(input_layouts=Shard(1), output_layouts=Replicate()),
+        "output": ColwiseParallel(
+            input_layouts=Shard(1),
+            output_layouts=Shard(-1) if loss_parallel else Replicate(),
+            use_local_output=not loss_parallel,
+        ),
         "layers.*.attn": layerwise_prepare_module_input_cls(
             input_layouts=(Shard(1), Shard(1)),
             desired_input_layouts=(Replicate(), Replicate()),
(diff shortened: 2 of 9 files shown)


torchtune/models/llama3/_parallelism.py
```diff
@@ -58,51 +63,59 @@ def _get_base_llama_tp_training_plan(
...
+        loss_parallel (bool): Whether to use loss parallelism after the output layer
+
     Returns:
         dict[str, Any]: The float8-enabled tensor parallel plan for Llama3 model.
     """
-    return FP8_LLAMA_TP_TRAINING_PLAN
+    return _get_base_llama_tp_training_plan(
+        layerwise_colwise_parallel_cls=Float8ColwiseParallel,
+        layerwise_rowwise_parallel_cls=Float8RowwiseParallel,
+        layerwise_prepare_module_input_cls=PrepareFloat8ModuleInput,
```
ebsmothers: nit: why do we prepend all of these with `layerwise`? E.g. `layerwise_colwise` sounds redundant to me

torchtune/training/_distributed.py
```diff
@@ -879,3 +884,26 @@ def context(model_inputs: list[torch.Tensor]):
             yield
 
     return context
+
+
+def get_train_context(
```
ebsmothers: Hmm now I am slightly regretting folding all of the context parallel stuff into a single context manager, as we now diverge slightly from how [titan is doing things](https://github.com/pytorch/torchtitan/blob/b7c7ed7167fa72485d987fa69af53538ee5be900/torchtitan/distributed/utils.py#L192-L198). It's a minor thing, but maybe we should just converge to that implementation so that we have the same contract across our generic training context manager and our CP context manager. (Can do it in a follow-up too, you don't have to pay for my sins here)

torchtune/models/llama3/_parallelism.py
```diff
@@ -58,51 +63,59 @@ def _get_base_llama_tp_training_plan(
...
+        "layers.*.attn.output_proj": RowwiseParallel(),
+        "layers.*.mlp.w1": ColwiseParallel(),
+        "layers.*.mlp.w2": RowwiseParallel(),
+        "layers.*.mlp.w3": ColwiseParallel(),
+    }
 
 
 def base_llama_tp_plan(
-    model: nn.Module, inference: bool = False
+    model: nn.Module, *, inference: bool = False, loss_parallel: bool = False
```
ebsmothers: Thanks for making these kwargs-only

recipes/full_finetune_distributed.py
```diff
@@ -339,11 +341,13 @@ def setup(self, cfg: DictConfig) -> None:
         self._compile_loss = compile_bool
         self._compile_optimizer_step = compile_bool
         self._compile_scale_grads = compile_bool
+        self._compile_autograd = False  # TODO: work out why this isn't working. We leave it parameterisable for debug purposes.
```
ebsmothers: If it doesn't work yet, personally I would lean towards leaving it out of this PR just so that someone doesn't accidentally try to use it. Plus iiuc it's relatively decoupled from loss parallel in terms of our implementation, right? But let's create an issue and can tag in compile/distributed as needed. Is the stack trace you shared in the test plan from running with activation checkpointing and/or offloading?

torchtune/models/llama3/_parallelism.py
```diff
@@ -58,51 +63,59 @@ def _get_base_llama_tp_training_plan(
...
+    return (
+        _get_base_llama_tp_inference_plan()
+        if inference
+        else _get_base_llama_tp_training_plan(loss_parallel=loss_parallel)
+    )
 
 
-# TODO: expose this once tested
-def _fp8_llama_tp_plan() -> dict[str, ParallelStyle]:
+def fp8_llama_tp_plan(
```
ebsmothers: I might be missing something here, but are we actually enabling this now? Like if I set `enable_fp8_training=True` and `tensor_parallel_plan=base_llama_tp_plan`, where do we actually hook this up?

torchtune/modules/loss/cross_entropy_loss.py
```diff
@@ -92,22 +94,27 @@ def compute_cross_entropy(
...
             logits = logits.full_tensor()
 
-        return F.cross_entropy(
+        loss = F.cross_entropy(
             logits.float(),
             target_chunk,
             reduction="sum",
             ignore_index=self.ignore_index,
         )
+        # the all-reduce later complains if a DTensor is returned
```
ebsmothers: Is this comment referring to [this](https://github.com/pytorch/torchtune/blob/23b3f7b421ff891c782d021021fed328c6509adc/recipes/full_finetune_distributed.py#L956) all-reduce? Iiuc loss parallel with vanilla cross-entropy returns a DTensor ([e.g.](https://github.com/pytorch/pytorch/blob/c8d39a10457ea5d65184c6e8f037f46c5525d869/torch/distributed/tensor/parallel/loss.py#L329-L336)), so want to make sure we are not doing anything too funky here.

ebsmothers (commented):
Thanks for the PR @nathan-az, these memory savings are really impressive! Lmk if any of the comments are unclear

torchtune/models/llama3/_parallelism.py
```diff
@@ -26,6 +26,7 @@ def _get_base_llama_tp_training_plan(
     layerwise_colwise_parallel_cls: type[ParallelStyle] = ColwiseParallel,
     layerwise_rowwise_parallel_cls: type[ParallelStyle] = RowwiseParallel,
     layerwise_prepare_module_input_cls: type[ParallelStyle] = PrepareModuleInput,
+    loss_parallel: bool = False,
```
pbontrager: I've been really trying to avoid this with the new loss functions and instead been having the loss functions modify the models directly. Otherwise we have to maintain a special TP plan for every kind of loss (see [ligerloss](https://github.com/pytorch/torchtune/pull/2741)). For the liger loss case you can just call full_tensor but I believe you can modify a model's output_layout after the fact too.

torchtune/modules/loss/cross_entropy_loss.py
```diff
@@ -92,22 +94,27 @@ def compute_cross_entropy(
...
+
         else:
             hidden_chunk = hidden_chunk[mask_chunk]  # [num_valid, embed_dim]
 
         # [num_valid, embed_dim] @ [embed_dim, vocab_size]
         if self.linear_projection is None:
             raise AttributeError("forward called before update_model")
         logits = self.linear_projection(hidden_chunk)  # [num_valid, vocab_size]
-        if isinstance(logits, DTensor):
+        if isinstance(logits, DTensor) and not self.loss_parallel:
```
pbontrager: I think we can just remove this check all together if we have the one on line 115

recipes/full_finetune_distributed.py
nathan-az: I don't recall the exact configuration for that stacktrace. Will remove autograd complile for now. Once we merge with loss parallel I can run some quick tests.

torchtune/models/llama3/_parallelism.py
nathan-az: I've been manually changing `tensor_parallel_plan` to `torchtune.models.llama3.fp8_llama_tp_plan`. Good call though - that's bad UX and we can just select the correct plan based on if fp8 is enabled. Will fix.

torchtune/models/llama3/_parallelism.py
nathan-az: This was a copy-paste, agree, happy to remove.

torchtune/models/llama3/_parallelism.py
nathan-az: Mm yes this is a reasonable point. The current pattern was taken from `torchtitan` but we will soon support more loss functions than them.

We have two slightly tricky requirements:
1. Our training step needs to be aware of whether to enter the loss parallel context manager. For ones that use the standard pytorch cross entropy loss, this is the case. For ones like liger, probably not, even if they support loss parallelism
2. We should probably modify the TP plan based on loss parallelism rather than using `redistribute` in the forward step, since this would introduce [additional collective](https://docs.pytorch.org/docs/stable/distributed.tensor.html#torch.distributed.tensor.DTensor.redistribute) (pytorch might be smart enough to optimise out, consecutive redistributes, but I'm not sure).

I think this is antithesis to what you're suggesting, but I wonder about relying _more_ on the loss function here, using:

* a method `patch_tp_plan` which makes any required modifications from the base tp plan prior to parallelisation in the model setup (as simple as a `dict.update`), which we default to a noop
* a property `use_loss_parallel_context_manager`, defaulted to false, which indicates whether to use the context manager in the loss parallel case or not, which we use in the training loop

I'll address simple comments first, then implement this so it's easy to review, and simple to revert if you don't like it. 

My rationale is that it turns loss parallelism into a first class citizen, and centralises any loss parallel-specific functionality to the loss class itself. Modifying the model I think will end up less straightforward depending on how loss is parallelised.

torchtune/modules/loss/cross_entropy_loss.py
nathan-az: Agreed, I don't think this is required since the `ParallelStyle`s default to `use_local_output=True`. I'll remove, but there may have been a good reason for this originally. Easy to add back if we need.

torchtune/modules/loss/cross_entropy_loss.py
nathan-az: That's exactly the allreduce I'm referring to. And yes, the loss parallel CE loss does return a Replicate DTensor. Without the `full_tensor`, when we reach the all-reduce in the recipe, we get:

```
[rank0]: AssertionError: found no DeviceMesh from dtensor args for c10d.allreduce_.default!
```

The loss that comes out is a DTensor, but is a `Replicate` so I saw no difference between using `full_tensor` and using `to_local` in the TP case, versus not using loss parallel at all. Differences to DP (I use as a baseline) are very small. This is without packing, and adjusting effective batch size.

image

I did notice that DP+TP+CP appears to have more drastic differences between grad norms **and** losses versus DP. I expect this is probably due to the grad norm scaling, not the loss parallelism, although CP with/without loss parallel does exhibit slight differences.

image

**note:** I tried to keep the run names short. `dp` refers to dp8, `cp` or `tp` use both dim `2`, with shard dim and batch size adjusted accordingly.

**Suggestion:** Unless you have a clearer idea, I'd lean towards leaving this `full_tensor` in, then follow up to investigate the CP difference.

I won't resolve this comment yet, but please do so if you're happy to leave this as-is for now.

torchtune/models/llama3/_parallelism.py
nathan-az: This [commit](https://github.com/pytorch/torchtune/pull/2782/commits/c4abec6fd7a425928e81a06e79d8d9bfa0209f91) addresses this. Required a bit of a refactor. Have also applied to LLaMA-4 so it raises an error more elegantly if someone tries FP8 + LLaMA-4 training.

Resolve this comment if you're happy with the solution :)

torchtune/training/_distributed.py
nathan-az: Will leave this for a follow-up to make this PR leaner, I'm happy to pick it up after this merges.

nathan-az:
Converting to draft, will update according to https://github.com/pytorch/torchtune/pull/2824 once it merges then request review again.

nathan-az:
I've made some changes to enable loss compilation in the non-loss parallel case. This did require taking some logic out of the function that is compiled, but it is only masking and slicing which I believe should be pretty trivial anyway, from both the memory and compute perspectives. I've confirmed the gains in memory utilisation against the standard PyTorch cross-entropy loss, as well as aligning loss curves.

image

Above is testing with LLaMA 3.1 8B, seq len 2**13, adjusting batch size for DP size, with 16 chunks (except "baseline" which does not use chunking). Very happy to see notable improvements in memory usage with the chunking, and more with loss parallelism, as well as minor tokens-per-second improvements from compile (in the DP case) and loss parallelism in the TP case.

We can experiment in the future with trying to get compile working in the TP case, but at least for now this is improved from the current state where it is fully disabled (in the DP case, I see 20% reduced peak memory usage currently).

p.s. this PR now takes heavily from @felipemello1's work https://github.com/pytorch/torchtune/pull/2824. Would be good to get felipe's thoughts too.

felipemello1:
hey @nathan-az , good stuff! I will review the PR later today. Meanwhile, can you help me understand the results? I am more interested in the memory portion. It seems to me that dp8-tp1-compile is the best case.

Is there a benefit from doing TP **and** not compiling? If the answer is no, should we just not do TP on the loss for now?

felipemello1 (commented):
hey Nathan, thanks for the PR. I will leave the parallelism implementation to @ebsmothers  and @pbontrager, since they were already on top of it. Regarding the loss, i like that you were able to port the other PR! Thanks for making it more modular and enable compile. My two cents:

1. In which scenarios would we want to have `mask_ignored_tokens=False`? If there isnt a strong one, maybe we should remove the flag
2. I personally dont like how SFTLoss has many args related to parallelism:
 a) 'supports_loss_parallel', 
 b) 'loss_parallel_enabled', 
 c) 'loss_paralell_requires_ctx_manager'. 
 d) 'use_loss_parallel_ctx_manager'

Maybe there is a way to simplify it? e.g. merge a) and c) and make b) and d) always True if input is TP?  I am ok with having only way of doing things if it works for most of the cases

nathan-az:
@felipemello1 fair, I erred on the side of leaving options to the users.

> In which scenarios would we want to have mask_ignored_tokens=False? If there isnt a strong one, maybe we should remove the flag

Basically to have an option to enable compile, in cases where very few tokens are masked on average. 

When packing to very high seq len, if samples aren't very long, packing should be very effective, and very few tokens could be masked. Thus, the user may see more gain by enabling compile. Once we have `compile` working in all case I support removing this option.

> I personally dont like how SFTLoss has many args related to parallelism:

Agreed, this is a lot, again, for the sake of adding options for users. You're right - there's no clear downside to LP, we can just default to it with TP.

I've just pushed another commit. It reduces the above to only one class-level property: `tp_requires_loss_parallel_manager`, and an attribute to let the loss funciton be aware of when TP is enabled. Useful for selective compiling, but I think also just generally reasonable since behaviour can change between the two cases. I think it's quite a bit simpler, let me know your thoughts.

recipes/full_finetune_distributed.py
```diff
@@ -355,6 +355,14 @@ def setup(self, cfg: DictConfig) -> None:
                 self._grad_scaler, backend=self._compile_backend
             )
 
+        # initialize loss
+        self._loss_fn = config.instantiate(cfg.loss)
+        self.use_loss_parallel_ctx_manager = self.parallel_dims.tp_enabled and getattr(
+            self._loss_fn,
+            "tp_requires_loss_parallel_ctx_manager",
```
felipemello1, line 366: n00b question: What if we removed this arg and just checked if 'self._loss_fn.patch_tp_plan is not None'?

Would that make sense?

nathan-az: That would throw an `AttributeError` for non-`SFTLoss` loss functions (e.g. the standard pytorch cross entropy loss). Not likely that a user opts into these, but I don't think the `getattr` is particularly gross.

In addition, just because the TP plan is patched in a custom loss function doesn't necessarily mean we want to use the context manager (e.g. your implementation parallelised on the sequence dimension wouldn't have used the context manager).

EDIT: feel free to resolve this thread if that makes sense and you're happy to keep the code as-is. If not let me know.

torchtune/modules/loss/cross_entropy_loss.py
```diff
@@ -34,32 +35,56 @@ def __init__(
...
-        log.warning("Skipping compile loss, as it is not supported at this time")
-        # TODO fix compile and re-enable
-        # self.compute_cross_entropy = torch.compile(
-        #     self.compute_cross_entropy, *args, **kwargs
-        # )
+        # compiling with loss parallelism appears to work without masking
+        if not self.tp_enabled or not self.mask_ignored_tokens:
+            self.compute_cross_entropy = torch.compile(
+                self.compute_cross_entropy, *args, **kwargs
+            )
```
felipemello1: sorry, can you clarify?

1) the comment says: "compiling with loss parallelism appears to work without masking"

So the condition should be:
```
if self.tp_enabled or not self.mask_ignored_tokens:
       compile
elif not self.tp_enabled:
        compile
else:
	do not compile
```



Is that it? If not, and we can ONLY compile if mask=False, then lets have default mask = False. compile >>> mask



2) Can we remove the comment?

torchtune/modules/loss/cross_entropy_loss.py
```diff
@@ -122,24 +156,33 @@ def forward(
...
-        if total_elements == 0:
-            # must return after calling compute_cross_entropy to not hang during data parallel training
-            return total_loss
-        else:
-            return total_loss / total_elements
+        targets = targets.reshape(-1)
+        outputs = outputs.reshape(-1, outputs.shape[-1])
+
+        if self.mask_ignored_tokens:
+            indices = torch.where(targets != self.ignore_index)[0]
```
felipemello1: since indices is not used anywhere else, we can calculate it inside of self.mask_inputs

torchtune/modules/loss/loss_types.py
```diff
@@ -36,6 +44,18 @@ def forward(self, outputs: torch.Tensor, targets: torch.Tensor) -> torch.Tensor:
         """
         pass
 
+    @property
+    def tp_requires_loss_parallel_ctx_manager(self) -> bool:
+        """
+        Whether to use the loss parallel context manager for loss parallelism. Can be
+        used if the function relies on the standard cross_entropy() or CrossEntropyLoss.
```
felipemello1: maybe point to our implementation instead, as an example? something like 'torchtune.modules.loss.LossName'

torchtune/modules/loss/cross_entropy_loss.py
nathan-az: It actually works with masking in the DP only case - something about TP + masking is problematic. Removed the comment and made the condition clearer.

torchtune/modules/loss/loss_types.py
nathan-az: This is just an indicator of whether the forward/backward pass needs the pytorch loss parallelism context manager. In this PR the CE loss actually still uses this (note how in the subclass this is overridden and returns `True`). This is how it is done in `torchtitan` as well although they don't use chunking.

I've removed the sentence and linked the docs to be clearer.

As discussed, we can refactor to parallelise loss on the sequence dimension rather than using this in a follow-up, but the code is a bit less simple and gains are minimal.

recipes/full_finetune_distributed.py
```diff
@@ -701,6 +700,16 @@ def _setup_model(
         self.activations_handling_ctx = training.get_act_offloading_ctx_manager(
             model, enable_activation_offloading, activation_offloading_use_streams
         )
+        # context parallel
+        self.optional_context_parallel_manager = training.get_context_parallel_manager(
```
ebsmothers: Commenting on @pbontrager's behalf.. let's keep it as `self.context_parallel_manager` here (he yelled at me because the "optional" naming is inconsistent with our activation offloading context manager, which is also optional)

torchtune/modules/loss/loss_types.py
```diff
@@ -36,6 +44,18 @@ def forward(self, outputs: torch.Tensor, targets: torch.Tensor) -> torch.Tensor:
...
 
+    @property
+    def tp_requires_loss_parallel_ctx_manager(self) -> bool:
+        """
+        Whether to use the loss parallel context manager for loss parallelism.
+        https://docs.pytorch.org/docs/stable/distributed.tensor.parallel.html#torch.distributed.tensor.parallel.loss_parallel
+        """
+        return False
+
+    def patch_tp_plan(self, tp_plan) -> bool:
```
ebsmothers: This isn't returning a bool, right?

recipes/full_finetune_distributed.py
```diff
@@ -355,6 +355,14 @@ def setup(self, cfg: DictConfig) -> None:
                 self._grad_scaler, backend=self._compile_backend
             )
 
+        # initialize loss
+        self._loss_fn = config.instantiate(cfg.loss)
+        self.use_loss_parallel_ctx_manager = self.parallel_dims.tp_enabled and getattr(
```
ebsmothers, line 364: Maybe a dumb q but does this imply that for vanilla CE loss we do not need a context manager for loss parallel? I had assumed we would still need it in that case

torchtune/modules/loss/cross_entropy_loss.py
```diff
@@ -34,32 +35,55 @@ def __init__(
         self,
         num_output_chunks: int = 8,
         ignore_index: int = -100,
+        tp_enabled: bool = False,
+        mask_ignored_tokens: bool = True,
     ):
-        super().__init__()
+        super().__init__(tp_enabled=tp_enabled)
```
ebsmothers: Hmm tbh this is not my favorite. I don't like relying on inheritance resolution order like this. I know we are gradually tightening the contract on SFTLoss but imo we are putting too much into an ABC here. Do we 100% need so set `tp_enabled` right on `__init__`?

torchtune/modules/loss/cross_entropy_loss.py
```diff
@@ -122,24 +157,32 @@ def forward(
...
-        hidden_chunks = outputs.tensor_split(self.num_output_chunks, dim=1)
-        target_chunks = targets.tensor_split(self.num_output_chunks, dim=1)
-
-        # Compute cross-entropy loss for the chunks
-        total_loss = 0.0
-        for idx in range(len(hidden_chunks)):
-            total_loss += self.compute_cross_entropy(
-                hidden_chunks[idx],
-                target_chunks[idx],
+        total_valid_tokens = torch.where(targets != self.ignore_index)[0].numel()
```
ebsmothers, line 160: Just a sanity check: we are not adding a new sync point here, right? (Based on the perf in the test plan I suspect we aren't, just want to confirm)

torchtune/modules/loss/cross_entropy_loss.py
```diff
@@ -34,32 +35,55 @@ def __init__(
...
+                self.compute_cross_entropy, *args, **kwargs
+            )
         return self
 
     def set_model_output(self, model: nn.Module) -> None:
         """Modify model output to match the expected input for the loss function."""
         model.skip_output_layer = True
         self.linear_projection = model.output
 
+    def patch_tp_plan(self, tp_plan) -> bool:
```
ebsmothers: Same comment here about not returning a bool

recipes/full_finetune_distributed.py
```diff
@@ -355,6 +355,14 @@ def setup(self, cfg: DictConfig) -> None:
...
+            self._loss_fn,
+            "tp_requires_loss_parallel_ctx_manager",
+            False,
+        )
```
ebsmothers, line 368: Also one question about how people will extend `SFTLoss`: do we actually think there are cases where someone will want really fine-grained control over these details? Because I think a logically simpler (but slightly more restrictive/annoying) way to do things is just say e.g.

```
if isinstance(self._loss_fn, SFTLoss) and self.parallel_dims.tp_enabled  and not use_loss_parallel
	raise ValueError("don't do this")
```

torchtune/modules/loss/cross_entropy_loss.py
```diff
@@ -122,24 +157,32 @@ def forward(
...
-        total_loss = 0.0
-        for idx in range(len(hidden_chunks)):
-            total_loss += self.compute_cross_entropy(
-                hidden_chunks[idx],
-                target_chunks[idx],
+        total_valid_tokens = torch.where(targets != self.ignore_index)[0].numel()
+        if total_valid_tokens == 0:
+            return torch.tensor(0.0, device=targets.device)
+
+        # this redistribute allows tensor spitting without replication
```
ebsmothers, line 164: this is really nice

recipes/full_finetune_distributed.py
nathan-az: @ebsmothers Not a dumb q - we absolutely would if a user explicitly wanted to use the vanilla loss with loss parallel.

I chose not to support it (in an early commit of this PR I defaulted to `True`). I realised we don't get any memory savings from it because the TP plans by default Replicate the outputs of the `output` layer.

The only way I could think to support the vanilla CE loss was to follow the `torchtitan` pattern of passing `loss_parallel_enabled` to the TP plan factories, but that felt messier, and it seems like we're leaning more heavily into custom loss classes.

Resolve if that makes sense :)

recipes/full_finetune_distributed.py
nathan-az: I figured the people developing the `SFTLoss` subclasses would be more technical than those running training. Not to mention it may not be clear to users whether a specific context manager is required or not (but the loss function developer will know).

So I opted to the developer declaring whether the class itself requires the context manager with `tp_requires_loss_parallel_ctx_manager`, declared once, in their class. (Plus, this defaults to `false` anyway).

So as-is, loss parallelism is parameterised at the class level and either is or isn't enabled with tensor parallelism, as opposed to being configured by the user.

torchtune/modules/loss/cross_entropy_loss.py
nathan-az: Shouldn't be, since the targets are are always local currently, no collective/sync should be required (assuming that's what you mean by sync).

torchtune/modules/loss/loss_types.py
nathan-az: 🤦

torchtune/modules/loss/cross_entropy_loss.py
nathan-az: Very fair, I've removed this requirement from `SFTLoss` and just have it in the subclass. Setting the field in the recipe won't cause any issue for classes (or functions) that don't expect it.

recipes/full_finetune_distributed.py
ebsmothers: So I would lean towards still supporting torch.nn.CrossEntropyLoss if it's possible without too much extra messiness. Mainly because it's still the canonical loss that most people are familiar with (rather than something custom that we've implemented in torchtune). I think passing `loss_parallel_enabled` to TP plan (a la torchtitan) isn't the worst thing in the world, but do we then need to also make our TP plan loss-aware (i.e. should we only set `use_local_output=False` in the last linear layer if both loss parallel is enabled and we are using vanilla CE loss?). If that's the case then I agree, it may be too much of a headache.

torchtune/modules/loss/cross_entropy_loss.py
ebsmothers: Oh yeah, what I mean is that in my past experience `.numel()` can introduce a CUDA synchronization (something like `torch.cuda.synchronize()`). After looking a bit more I suspect it should be a non-issue here, so going to resolve this comment.

recipes/full_finetune_distributed.py
ebsmothers: Ah I see, I may have misunderstood slightly then. So basically if a user runs with TP degree N and our linear CE loss, we assume they also want loss parallel degree N. If they want to use other losses, they need to specify the expected behavior themselves using this bool. Is that right?

nathan-az: > a user runs with TP degree N and our linear CE loss, we assume they also want loss parallel degree N

Exactly!

> If they want to use other losses, they need to specify the expected behavior themselves using this bool. Is that right?

The user (person running training) won't ever need to specify this bool - it would be specified by the "other loss" class in the `@property` (defaulting to false). So this is zero-touch for users - it will never be in a config. Developers of new loss functions specify it once in-code if they want to override the `false` default.

recipes/full_finetune_distributed.py
nathan-az: Yeah it's tricky since different custom loss functions may desire different TP plans, so I think locating them there makes sense. e.g. there's an alternative pattern to what I have here. Instead of using the torch context manager, we have the output layer's layout be `Shard(0), use_local_output=True`, then compute local losses and all-reduce (I'll try this properly in future).

We can support `torch.nn.CrossEntropyLoss` specifically if we don't mind some adhoc code e.g. directly in the recipe:

```py
if isinstance(self._loss_fn, SFTLoss):
    self._loss_fn.tp_enabled = True
    self.tp_plan = self._loss_fn.patch_tp_plan(self.tp_plan)

# we add one more condition here
elif isinstance(self._loss_fn, torch.nn.CrossEntropyLoss):
    self.tp_plan["output"] = ColwiseParallel(
        input_layouts=Shard(1),
        output_layouts=Shard(-1),
        use_local_output=False,
    )
```
Then later on in the recipe, we have:
```py
self.use_loss_parallel_ctx_manager = self.parallel_dims.tp_enabled and (
    getattr(
        self._loss_fn,
        "tp_requires_loss_parallel_ctx_manager",
        False,
    ) or
    # and one more condition here
    isinstance(self._loss_fn, torch.nn.CrossEntropyLoss)
)
```

ebsmothers: OK thanks for spelling all this out. Then I agree.. it's probably not worth the additional cognitive load to support this. Also, separately, given the discussion on #2741 I am hopeful that in the long run we will have a fused CE loss in core matching the signature of SFTLoss. So if we have to sacrifice a bit on flexibility until then to keep the code cleaner I am OK with it.

ebsmothers (approved):
OK no remaining concerns from my side, this looks good to me. We need to get CI fixed (@felipemello1 is working on this in #2841 and elsewhere), once we can get a clean CI run in this should be good to merge.

recipes/full_finetune_distributed.py
nathan-az: 100%. I'm keen to deprecate this once there's a fused linear loss kernel in core. In the meantime this gives us a nice stopgap. Once https://github.com/pytorch/torchtune/pull/2741 merges as well I can see about getting it to work with loss parallelism (across the sequence) as well, since we should get some time and memory savings there following this pattern - I just don't want to increase the scope of that PR.

felipemello1:
please merge main so tests can pass

PR #2794, opened 2025-06-06
[WIP] Tool calling support in the torchtune

krammnic:
#### Context
- [X] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

#### Changelog
* tool calling support + modifications for the HFBase/HFModel tokenizers

#### Test plan

#### UX

### DIFF
--- a/torchtune/data/_messages.py
+++ b/torchtune/data/_messages.py
@@ -52,6 +52,7 @@ class Message:
         masked (bool): whether the message is masked in the sample. If True, do not use
             in loss calculation. Default: False
         ipython (bool): whether the message is a tool call. Default: False
+        tool_calls (Optional[list]): list of tool calls related to this message. Default: None
         eot (bool): whether the message corresponds to the end of a turn, where control is handed over
             to the assistant from the user or the user from the assistant. Default: True. Should be true
             in most cases except for:
@@ -71,12 +72,14 @@ class Message:
         content: Union[str, list[dict[str, Any]]],
         masked: bool = False,
         ipython: bool = False,
+        tool_calls: Optional[list] = None,
         eot: bool = True,
     ):
         self.role = role
         self.content = self._convert_to_list_of_dict(content)
         self.masked = masked
         self.ipython = ipython
+        self.tool_calls = tool_calls
         self.eot = eot
 
         self._validate_message()
--- a/torchtune/modules/transforms/tokenizers/_hf_tokenizer.py
+++ b/torchtune/modules/transforms/tokenizers/_hf_tokenizer.py
@@ -90,8 +90,11 @@ class HuggingFaceBaseTokenizer(BaseTokenizer):
         self.eos_token = "<eos>"
 
         if self.config:
-            self.bos_token = self._get_token_from_config(self.config, "bos_token")
-            self.eos_token = self._get_token_from_config(self.config, "eos_token")
+            try:
+                self.bos_token = self._get_token_from_config(self.config, "bos_token")
+                self.eos_token = self._get_token_from_config(self.config, "eos_token")
+            except ValueError:
+                pass
             if self.bos_token is not None:
                 self.bos_id = self.tokenizer.token_to_id(self.bos_token)
             if self.eos_token is not None:
@@ -102,9 +105,6 @@ class HuggingFaceBaseTokenizer(BaseTokenizer):
                 self.bos_id = self.generation_config.get("bos_token_id")
             if self.eos_id is None:
                 self.eos_id = self.generation_config.get("eos_token_id")
-
-        if self.bos_id is None or self.eos_id is None:
-            raise ValueError("Could not infer BOS and EOS token IDs from config")
 
     def _infer_should_add_bos_eos(self):
         """
@@ -136,9 +136,11 @@ class HuggingFaceBaseTokenizer(BaseTokenizer):
             list[int]: The list of token ids.
         """
         token_ids = self.tokenizer.encode(text).ids
-        if add_bos and not self.hf_adds_bos and self.bos_token not in text:
+
+        # Both bos_id and eos_id might be None (null). Therefore, we need an additional check.
+        if add_bos and not self.hf_adds_bos and self.bos_token not in text and self.bos_id:
             token_ids.insert(0, self.bos_id)
-        if add_eos and not self.hf_adds_eos:
+        if add_eos and not self.hf_adds_eos and self.eos_id:
             token_ids.append(self.eos_id)
         return token_ids
 
@@ -269,7 +271,7 @@ class HuggingFaceModelTokenizer(ModelTokenizer):
 
         for i, message in enumerate(messages):
             current_messages = [
-                {"role": m.role, "content": m.content[0]["content"]}
+                {"role": m.role, "content": m.content[0]["content"], "tool_calls": m.tool_calls}
                 for m in messages[: i + 1]
             ]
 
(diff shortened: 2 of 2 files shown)


kamalojasv181:
Hey @krammnic, does this support tool calls for all formats (like openai, sharegpt etc)?

krammnic:
It still WIP, but yes, it will

torchtune/modules/transforms/tokenizers/_hf_tokenizer.py
```diff
@@ -136,9 +136,16 @@ def encode(
             list[int]: The list of token ids.
         """
         token_ids = self.tokenizer.encode(text).ids
-        if add_bos and not self.hf_adds_bos and self.bos_token not in text:
+
+        # Both bos_id and eos_id might be None (null). Therefore, we need an additional check.
```
ebsmothers, line 145: Is this related to tool-calling? Or a separate issue?

torchtune/modules/transforms/tokenizers/_hf_tokenizer.py
```diff
@@ -90,8 +90,11 @@ def _infer_bos_eos_tokens(self):
         self.eos_token = "<eos>"
 
         if self.config:
-            self.bos_token = self._get_token_from_config(self.config, "bos_token")
-            self.eos_token = self._get_token_from_config(self.config, "eos_token")
+            try:
+                self.bos_token = self._get_token_from_config(self.config, "bos_token")
+                self.eos_token = self._get_token_from_config(self.config, "eos_token")
+            except ValueError:
+                pass
```
ebsmothers: In this case I wonder whether we should just modify `_get_token_from_config` to directly return None (possibly logging a warning) rather than use this try/except

torchtune/data/_messages.py
```diff
@@ -52,6 +52,7 @@ class Message:
         masked (bool): whether the message is masked in the sample. If True, do not use
             in loss calculation. Default: False
         ipython (bool): whether the message is a tool call. Default: False
+        tool_calls (Optional[list]): list of tool calls related to this message. Default: None
```
ebsmothers, line 55: Should we also update the role "ipython" to "tool" to match what's done by Hugging Face?

ebsmothers (commented):
Thanks for working on this! I left a few small comments. We should also add a test to ensure that this actually works and generates the expected outputs on a tool-calling dataset

torchtune/data/_messages.py
krammnic: Yes, good catch, that argument seemed to me weird.

torchtune/modules/transforms/tokenizers/_hf_tokenizer.py
krammnic: It is caused by separate issue in HuggingfaceBaseTokenizer.

pyproject.toml
```diff
@@ -48,7 +48,7 @@ tune = "torchtune._cli.tune:main"
 
 [project.optional-dependencies]
 dev = [
-    "bitsandbytes>=0.43.0",
+#    "bitsandbytes>=0.43.0",
```
nathan-az, line 51: Was this an intentional removal?

torchtune/data/_messages.py
```diff
@@ -108,7 +111,8 @@ def from_dict(cls, d: dict) -> "Message":
             role=d["role"],
             content=d["content"],
             masked=d.get("masked", False),
-            ipython=d.get("ipython", False),
+            tool_calls=d.get("tool_calls", []),
+            tool=d.get("tool", False),
```
nathan-az, line 115: While I agree with this change, it currently breaks existing tokenizers. Repro:
```py
from torchtune.datasets import alpaca_cleaned_dataset
from torchtune.models.qwen2_5 import qwen2_5_tokenizer

vocab_path = "/tmp/Qwen2.5-14B-Instruct/vocab.json"
merges_path = "/tmp/Qwen2.5-14B-Instruct/merges.txt"
tokenizer_json_path = "/tmp/Qwen2.5-14B-Instruct/tokenizer.json"
tokenizer_config_path = "/tmp/Qwen2.5-14B-Instruct/tokenizer_config.json"

tokenizer_qwen = qwen2_5_tokenizer(
    path=vocab_path,
    merges_file=merges_path,
    max_seq_len=512
)
dataset_qwen = alpaca_cleaned_dataset(tokenizer=tokenizer_qwen, packed=False)
```

nathan-az:
@krammnic This is great progress. It looks like a bit of work needs to be done on BC with existing tokenizers (unless the plan is to fully deprecate them). In addition I'm seeing some issues with the jinja rendering - I think it may require explicitly passing tools to the renderer ([hf ref](https://github.com/huggingface/transformers/blob/main/src/transformers/tokenization_utils_base.py#L1641)).

Repro:

<details>

```py
from torchtune.datasets import alpaca_cleaned_dataset
from torchtune.modules.transforms.tokenizers import HuggingFaceModelTokenizer


tokenizer_json_path = "/tmp/Qwen2.5-14B-Instruct/tokenizer.json"
tokenizer_config_path = "/tmp/Qwen2.5-14B-Instruct/tokenizer_config.json"

tokenizer_hf = HuggingFaceModelTokenizer(
    tokenizer_json_path=tokenizer_json_path,
    tokenizer_config_json_path=tokenizer_config_path,
    max_seq_len=512,
)
dataset_hf = alpaca_cleaned_dataset(tokenizer=tokenizer_hf, packed=True)
```

</details>

Basically optionally propagating `tools` might solve this ([basic example](https://github.com/nathan-az/torchtune/commit/c871557c790ee7bbe31925d1a5b279ddba560f2c)). I haven't tested beyond seeing that the above repro _runs_ but haven't checked correctness.

pyproject.toml
krammnic: Nope :/ I don't like to do it on Mac (in another case it will not install) and then remove the comment. Will open separate PR to address this

torchtune/data/_messages.py
krammnic: Hmm, yep. We need to remove ipython everywhere.

torchtune/modules/transforms/tokenizers/_hf_tokenizer.py
```diff
@@ -272,15 +288,18 @@ def tokenize_messages(
...
         previous_tokens = []
 
         for i, message in enumerate(messages):
             current_messages = [
-                {"role": m.role, "content": m.content[0]["content"]}
+                {
+                    "role": m.role,
+                    "content": m.content[0]["content"],
+                    "tool_calls": m.tool_calls,
+                }
```
nathan-az, line 302: I had some issues specifically with the LLaMA 3(.3) tokenizer here, which didn't play nicely with empty tool calls `[]` or None. Ended up replacing with:
```py
            current_messages = [
                {
                    "role": m.role,
                    "content": m.content[0]["content"],
                    **({"tool_calls": m.tool_calls} if m.tool_calls is not None else {})
                }
                for m in messages[: i + 1]
            ]
```
I'm not sure if this is the correct logic though (or if this plays nicely with all tokenizers)

krammnic: This works fine with other tokenizers! Thanks

nathan-az: For me this broke qwen coder (sigh). I think removing strict undefined may have fixed it? I don't recall exactly. I kept trying components from the transformers jinja compile until I found something, but I don't know much about how jinja works - there may be a better solution.

krammnic:
@nathan-az Hey! We might want to merge this. I will introduce some final changes tomorrow and then we will be able to merge. Thanks for the patience and comprehensive review.

nathan-az:
Sounds good. Given the issues I had with other tokenizers I think it would be good to add tests with a couple of the popular models' tokenizers to confirm that:

1. there are no issues with their templates, and
1. the tokenization is correct (or correct enough)

Up to you whether we want to put their tokenizers directly in the torchtune source or if we add `transformers` as a test dependency only. I'm not sure if we have a precedent for this (but it would also cut down the size of `torchtune` which may be good, and only slightly slow the CI).

PR #2799, opened 2025-06-06
Fix tests for full DPO distributed

joecummings:
## Problem

In our ``test_full_dpo_distributed.py`` tests, we set the ``max_seq_len=256``. The prompt for our example
stack exchange paired datasets were always longer than 256, therefore the model never learned anything. Because it
never learned anything, our ``resume_from_checkpoint`` tests ALWAYS passed b/c the loss was the same every single time.

## Solution

1. Actually compare to loss values
2. Increase max seq len so that the model learns
3. Fix the tests now that they correctly fail

## Open questions

* You'll notice that the first step does not update the loss value. I don't exactly know why this is
except that maybe the gradient updates from the first batch aren't enough to meaningfully change the loss calculation
for batch two, but that feels a little flimsy. I ran the full recipe and loss seemed to decrease in a normal fashion, 
so I'm timeboxing this Issue for now. 
* Could this be happening with other tests? The other DPO tests also do not compare to loss values and have a small max_seq_len so those would be prime candidates for this kind of bug.

### DIFF
--- a/tests/recipes/test_full_dpo_distributed.py
+++ b/tests/recipes/test_full_dpo_distributed.py
@@ -26,9 +26,12 @@ from tests.test_utils import (
 
 
 class TestFullDPODistributedRecipe:
+    def expected_loss_values(self):
+        return [0.69315, 0.69315, 0.69301, 0.69241]
+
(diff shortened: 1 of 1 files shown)


ebsmothers (approved):
you're a hero

Issue #2725, opened 2025-05-13
Integrate Muon optimizer


joecummings:
The Muon optimizer has shown to be an efficient optimizer, potentially outpacing AdamW for LLM training. To quote [Essential AI](https://arxiv.org/abs/2505.02222) “Muon requires 10–15 % fewer tokens than AdamW to reach an identical loss and converts these savings into faster wall-clock convergence, with the advantage staying constant or growing as the batch size increases… These results establish Muon as a drop-in successor to AdamW for second-order optimization at scale.”
 
We'd love to accept a contribution of a canonical example of Muon in the torchtune library, specifically for our full SFT recipes (single device and multi GPU). 

## Artifacts

* An implementation of the Muon optimizer as a Pytorch Optimizer
* Any changes needed to the recipes to support the change for our feature set

## Acceptance Criteria

* Clean, well documented code with proper citations 
* Tests
* Logs comparing Muon to AdamW for text training
* Logs comparing Muon to AdamW for multimodal (image + text) training

## Resources

* [PyTorch implementation](https://medium.com/@kyeg/building-the-muon-optimizer-in-pytorch-a-geometric-approach-to-neural-network-optimization-17f4601be548)
* [Practical Efficiency of Muon for Pretraining](https://arxiv.org/abs/2505.02222)
* [Blog explaning Muon](https://kellerjordan.github.io/posts/muon/)

Saurabh750:
Hi @joecummings , I would like to work on this issue. Can you please assign it to me? Thanks!

PR #2803, opened 2025-06-08
Integrate Muon optimizer (2725)

Saurabh750:
#### Context
- [ ] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

#2725 

#### Changelog
* Integrating Muon optimizer as a pytorch implementation in torchtune.
* Modify recipes accordingly.

#### Test plan

#### UX

### DIFF
--- a/torchtune/modules/muon.py
+++ b/torchtune/modules/muon.py
@@ -1,0 +1,238 @@
+import torch
+import torch.distributed as dist
+
+
+def zeropower_via_newtonschulz5(G, steps: int):
+    """
+    Newton-Schulz iteration to compute the zeroth power / orthogonalization of G. We opt to use a
+    quintic iteration whose coefficients are selected to maximize the slope at zero. For the purpose
+    of minimizing steps, it turns out to be empirically effective to keep increasing the slope at
+    zero even beyond the point where the iteration no longer converges all the way to one everywhere
+    on the interval. This iteration therefore does not produce UV^T but rather something like US'V^T
+    where S' is diagonal with S_{ii}' ~ Uniform(0.5, 1.5), which turns out not to hurt model
+    performance at all relative to UV^T, where USV^T = G is the SVD.
+    """
+    assert G.ndim >= 2 # batched Muon implementation by @scottjmaddox, and put into practice in the record by @YouJiacheng
+    a, b, c = (3.4445, -4.7750,  2.0315)
+    X = G.bfloat16()
+    if G.size(-2) > G.size(-1):
+        X = X.mT
+
+    # Ensure spectral norm is at most 1
+    X = X / (X.norm(dim=(-2, -1), keepdim=True) + 1e-7)
+    # Perform the NS iterations
+    for _ in range(steps):
+        A = X @ X.mT
+        B = b * A + c * A @ A # quintic computation strategy adapted from suggestion by @jxbz, @leloykun, and @YouJiacheng
+        X = a * X + B @ X
+    
+    if G.size(-2) > G.size(-1):
+        X = X.mT
+    return X
+
+
+def muon_update(grad, momentum, beta=0.95, ns_steps=5, nesterov=True):
+    momentum.lerp_(grad, 1 - beta)
+    update = grad.lerp_(momentum, beta) if nesterov else momentum
+    if update.ndim == 4: # for the case of conv filters
+        update = update.view(len(update), -1)
+    update = zeropower_via_newtonschulz5(update, steps=ns_steps)
+    update *= max(1, grad.size(-2) / grad.size(-1))**0.5
+    return update
+
+
+class Muon(torch.optim.Optimizer):
+    """
+    Muon - MomentUm Orthogonalized by Newton-schulz
+
+    https://kellerjordan.github.io/posts/muon/
+
+    Muon internally runs standard SGD-momentum, and then performs an orthogonalization post-
+    processing step, in which each 2D parameter's update is replaced with the nearest orthogonal
+    matrix. For efficient orthogonalization we use a Newton-Schulz iteration, which has the
+    advantage that it can be stably run in bfloat16 on the GPU.
+
+    Muon should only be used for hidden weight layers. The input embedding, final output layer,
+    and any internal gains or biases should be optimized using a standard method such as AdamW.
+    Hidden convolutional weights can be trained using Muon by viewing them as 2D and then
+    collapsing their last 3 dimensions.
+
+    Arguments:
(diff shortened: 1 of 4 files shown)


krammnic:
Thanks for the first pass! Let's split single device and distributed versions of Muon in 2 separate files to improve readability. Speaking about plots: We will need some comparison performance plots against AdamW and general Wandb plots (loss) and  results on evaluation recipe.

krammnic:
I've added few comments, but it looks great! There are 2 things on which we might need to think though:

1. Can we reduce the amount of "muon checks"? Maybe some special wrapper similarly to a fused optimizer?
2. Maybe we need to implement it in a little bit cleaner way and support through builders?

Saurabh750:
Based on above comments, 2 things came to my mind:
1. As @krammnic suggested, I can implement a fused optimizer - a wrapper around Muon where the 2nd optimizer for linear layers will be of the choice of the user. This will eliminate the muon checks and will be a cleaner way.
2. If we want to provide more flexibility while assigning optimizer, we can do it on parameter level. 
Eg: Inside config file: 
```
optimizer:
  muon: [param1, param2]
  AdamW: []
muon:
  _component_: torchtune.modules.SingleDeviceMuon
  momentum: 0.95
  lr: 0.02
  weight_decay: 0
AdamW:
  _component_: bitsandbytes.optim.PagedAdamW
  lr: 1e-5
```

Muon will be assigned to only param1, param2 while AdamW will be assigned for all remaining ones.
Please do let me know your views, I'll implement whatever is suitable!

torchtune/modules/muon.py
```diff
@@ -0,0 +1,253 @@
+######################################################
```
joecummings: Can you move this to our ``optim.py`` file where ``OptimizerInBackward`` lives?

torchtune/modules/muon.py
```diff
@@ -0,0 +1,253 @@
...
+
+def adam_update(grad, buf1, buf2, step, betas, eps):
+    buf1.lerp_(grad, 1 - betas[0])
+    buf2.lerp_(grad.square(), 1 - betas[1])
+    buf1c = buf1 / (1 - betas[0]**step)
+    buf2c = buf2 / (1 - betas[1]**step)
+    return buf1c / (buf2c.sqrt() + eps)
+
+
+class MuonWithAuxAdam(torch.optim.Optimizer):
```
joecummings: I think we should optimize for this interface so that users don't need to specify multiple optimizers in their code and we can remove the messy branching logic. 

My rough proposal is:

```python
class MuonWithAuxAdam(torch.optim.Optimizer):
    def __init__(
        self,
        params,
        *,
        muon_selector=None,
        muon_lr: float = 0.02,
        muon_momentum: float = 0.95,
        adam_lr: float = 3e-4,
        adam_betas=(0.9, 0.95),
        adam_eps: float = 1e-10,
        weight_decay: float = 0.0,
    ):        
       if muon_selector is None:
            muon_selector = (
                lambda name, p: p.ndim >= 2
                and "tok_embeddings" not in name
                and "output"        not in name
            )

        muon_params = [p for n, p in named_params if muon_selector(n, p)]
        adam_params = [p for n, p in named_params if not muon_selector(n, p)]

        muon_params.sort(key=lambda p: p.size(), reverse=True)

        super().__init__(
            [
                dict(params=muon_params,
                     lr=muon_lr,
                     momentum=muon_momentum,
                     weight_decay=weight_decay,
                     use_muon=True),
                dict(params=adam_params,
                     lr=adam_lr,
                     betas=adam_betas,
                     eps=adam_eps,
                     weight_decay=weight_decay,
                     use_muon=False),
            ],
            defaults={}
        )
```

What do you think?

Also cc @ebsmothers  for thoughts.

joecummings (commented):
This is awesome, thanks for porting this over!

The TL;DR of my thoughts is that we should optimize for the AuxAdam varities of the Muon optimizer so that our code stays clean and organized. I made some suggestions so that the instantiation of these classes is easy from code or config.

torchtune/modules/muon.py
joecummings: cc @krammnic, I think this is the easiest way to go. WDYT?

intervitens: Sounds good, I've seen a similar approach in other implementations, like https://github.com/ethansmith2000/fsdp_optimizers/blob/main/fsdp_optimizers/muon.py

torchtune/modules/muon.py
joecummings: Ah I see Mark maybe suggested two files, but I'd prefer to limit everything to `optim.py` until we see the need for new files. Our library has also gotten a little bloated.

Saurabh750:
@joecummings @krammnic : I have added Muon to optim.py
I tried finetuning on alpaca dataset using qwen2-0.5B and compared AdamW and Muon.
Trained for 5, 10, 20 epochs for batchsize 5 and 10. 
For all the experiments, AdamW performed better than Muon. I am attaching snippet of one of the experiments with batch-size 10 and 20 epochs:
adamwvsMuon

- Should I try running the same experiments on a different model? According to [this](https://arxiv.org/abs/2502.16982v1), the results we are getting are on similar lines. AdamW performs better than Muon as the model was not pretrained using Muon.

There is a custom implementation of Adam in Muon class. I tried using the existing pytorch implementation with the view to use any pre-existing implementations of optimizer for linear layers. But this will not be possible due to load_state_dict() which supports only single optimizer to be stored. But, I believe load_state_dict() in OptimizerInBackward supports multiple optimizers. Please correct me if I am wrong.

I have tried reducing the muon checks, but still there is a muon check in the main file. Also, I have updated the get_lr() method for returning the lr of Muon only. Please suggest if this is correct.

krammnic:
Then, something is wrong, I don't like the fact that we have worse performance, because loss might be fixable with some HPO. Will review your changes tomorrow, to maybe find a bottleneck...

Saurabh750:
Hi @joecummings @krammnic , I have updated the Muon optimizer implementation.
In below image: 
Green -> Muon optimizer for first 20 epochs with 5e-4 lr
Blue -> Muon optimizer for epochs 20 to 40 with 5e-5 lr
image

For below image:
Blue -> Muon optimizer for epochs 20 to 40 with 5e-5 lr
Red -> AdamW optimizer for epochs 0 to 20 with 2e-5 lr

image

As suggested earlier, switching to another implementation and playing around with HPO helped.
In the lr_scheduler, I am only returning the Muon learning rate. Please let me know if anything else needs attention.

PR #2804, opened 2025-06-09
[BE] Remove Llama2 references from recipe tests

joecummings:
### Context

1. Llama2 is an old, unused model so it doesn't really help to test on it
2. Fewer useless tests means faster iteration (these tests run in X)
3. This bubbles up to step-based checkpointing b/c I need to switch over all our tests to use llama3_137M

### FAQs

1. Why did you not change the PPO tests?

I'll open an Issue for this, but I won't be touching that for step-based checkpointing right now anyways and I need a script from Salman to generate a new Llama3-based classifier model.

2. Is it really the right move to add ``scaler_block_size`` to the top level component builders for Llama3?

Fair enough, the other way to handle this would be to generate a completely new test model that was large enough to be properly quantized. I went with this approach because a) this llama3 model will be changed soon anyways and b) it's an optional param so it doesn't really affect the UX.

3. Why did you have to remove ``intermediate_dim`` from the tiny QAT model?

It's calculated if omitted anyways and it causes problems b/c the model config for Llama3 doesn't include it, therefore it throws a size mismatch upon construction.

### DIFF
--- a/recipes/configs/llama2/1B_qat_single_device.yaml
+++ b/recipes/configs/llama2/1B_qat_single_device.yaml
@@ -32,7 +32,6 @@ model:
   num_kv_heads: 4
   embed_dim: 2048
   max_seq_len: 2048
-  intermediate_dim: 5632
   attn_dropout: 0.0
   norm_eps: 1e-5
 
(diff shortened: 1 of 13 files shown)


pbontrager (approved):
Thank you for doing this

PR #2808, opened 2025-06-09
Update PPO tests to Llama3

### LINKED ISSUE #2806: [BE] Update PPO recipe test to utilize Llama3 model instead of Llama2
**Context**: In #2804, we removed all references to Llama2 in our recipe tests with the exception of PPO recipe test. We need to do that

**Steps**:

* Create new dummy classifier model w/ Llama3 base
* Use this new dummy classifier model in the PPO tests
* Update loss values to match new arch
* Profit?

salmanmohammadi:
Closes #2806 

Why did I remove the loss values for the other GPU runners? 

Because we don't use them in our CI. I'll add them back if this changes. I think this does mean that this test won't run on some local GPUs, but I'm not sure who out there is running full recipe tests locally. I want to meet this person if they exist and shake their hand.

### DIFF
--- a/tests/recipes/test_ppo_full_finetune_single_device.py
+++ b/tests/recipes/test_ppo_full_finetune_single_device.py
@@ -4,7 +4,6 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
-import logging
 import os
 import runpy
 import sys
@@ -32,9 +31,6 @@ from torchtune.training.checkpointing._utils import (
     RECIPE_STATE_DIRNAME,
     SHARD_FNAME,
 )
-
-
-logger = logging.getLogger(__name__)
 
 
 class TestPPOFullFinetuneSingleDeviceRecipe:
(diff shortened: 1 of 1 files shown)


tests/cache_artifacts.sh
```diff
@@ -17,7 +17,7 @@ SMALL_MODEL_URLS=(
     "https://ossci-datasets.s3.amazonaws.com/torchtune/small-ckpt-meta-03082024.pt"
     "https://ossci-datasets.s3.amazonaws.com/torchtune/small-ckpt-hf-03082024.pt"
     "https://ossci-datasets.s3.amazonaws.com/torchtune/small-ckpt-tune-llama3-05052024.pt"
-    "https://ossci-datasets.s3.amazonaws.com/torchtune/small-ckpt-hf-reward-07122024.pt"
+    "https://ossci-datasets.s3.amazonaws.com/torchtune/small-ckpt-hf-reward-06092025.pt"
```
joecummings, line 20: cc @ebsmothers for uploading the script here

PR #2814, opened 2025-06-11
[BE] Remove unused vars in recipes

joecummings:
ez

Removed fully in #2613, but snuck back in via a bad merge I think.


pbontrager (approved):
:heart:

PR #2820, opened 2025-06-12
[WIP] Qwen3 MoE support

intervitens:
This PR adds support for Qwen3 MoE (30B-A3B and 235B-A22B) models. Loss looked reasonable from a simple test with 30B-A3B on the Alpaca dataset.

TODO:
- [ ] Tensor/Expert parallel
- [x] Test 235B model
- [x] Verify loss curves against HF implementation
- [x] LoRA support
- [ ] Documentation
- [ ] Tests

### DIFF
--- a/torchtune/models/qwen3/_parallelism.py
+++ b/torchtune/models/qwen3/_parallelism.py
@@ -37,21 +37,6 @@ def tp_training_plan(model: nn.Module) -> dict[str, ParallelStyle]:
         dict[str, Any]: The tensor parallel plan for Llama4 model.
     """
     plan = {
-        # "decoder": PrepareModuleInput(
-        #     input_kwarg_layouts={
-        #         "tokens": None,
-        #         "mask": None,
-        #         "input_pos": None,
-        #         "input_embeds": Replicate(),
-        #     },
-        #     desired_input_kwarg_layouts={
-        #         "tokens": None,
-        #         "mask": None,
-        #         "input_pos": None,
-        #         "input_embeds": Shard(1),
-        #     },
-        #     use_local_output=True,
-        # ),
         "norm": SequenceParallel(),
         "output": ColwiseParallel(
             input_layouts=Shard(1),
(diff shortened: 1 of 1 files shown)


torchtune/models/qwen3/_convert_weights.py
```diff
@@ -125,3 +130,149 @@ def qwen3_tune_to_hf(
...
+        if "experts.0" in key:
+            new_key = get_mapped_key_moe(key, _FROM_HF)
+            converted_state_dict[new_key] = torch.stack(
+                [
+                    state_dict[str(i).join(key.rsplit("0", 1))].T
+                    for i in range(num_experts)
+                ]
+            )
+        elif "experts" in key:
+            continue
```
intervitens, line 226: This is kinda hacky and slow, but stacking experts into a single tensor lets us use the existing MoE implementation and _grouped_mm for a very significant speed boost.

salmanmohammadi: I found this to be prohibitively slow for larger models when testing with DeepSeek. I ended up following the torchtitan approach which uses a `nn.ModuleDict` for storing expert weights and grouping activated experts on-the-fly for grouped gemm.

torchtune/modules/moe/moe.py
```diff
@@ -56,14 +62,18 @@ def forward(
...
+            scores = torch.sigmoid(scores.to(torch.float32)).to(x.dtype)
 
         # top scores shape (bs*slen, top_k)
         top_scores, selected_experts_indices = torch.topk(
             scores, k=self.experts_per_token, dim=1
         )
         self.selected_experts_indices = selected_experts_indices
-        # top_scores /= top_scores.sum(dim=-1, keep_dim=True).to(x.dtype)
+        if self.norm_topk_prob:
+            top_scores /= top_scores.sum(dim=-1, keepdim=True).to(x.dtype)
```
salmanmohammadi, line 78: YMMV but I ended up having to do something like
```
            denominator = top_scores.sum(dim=-1, keepdim=True) + 1e-20
            top_scores /= denominator
```

torchtune/models/qwen3/_convert_weights.py
intervitens: As far as I can see, torchtitan also needs to stack expert weights to use grouped gemm, they just do it [at a slightly later point](https://github.com/pytorch/torchtitan/blob/e7c0cae934df78d6e9c2835f42ff1f757dc3fddc/torchtitan/experiments/deepseek_v3/group_gemms.py#L149-L150). They also [don't support](https://github.com/pytorch/torchtitan/blob/e7c0cae934df78d6e9c2835f42ff1f757dc3fddc/torchtitan/experiments/deepseek_v3/train_ds_real.py#L143-L144) grouped gemm for training for now.

salmanmohammadi: Oh you're totally right - I was thinking of EP.
If this doesn't take ages to convert the state dict it may be fine, otherwise only stacking the activated experts at runtime does make more sense to me.

dz1iang:
@intervitens Hi，Does it support training with fp8-compatible checkpoints?

dz1iang:
Hello, I see the training process prompts:` "Saving Qwen3 MoE adapter weights to PEFT format is not supported, saving to torchtune format instead."` 
May I ask how to obtain the Hugging Face (hf) checkpoint? Is there any code example for reference? 
Additionally, I only set lora_attn_modules: `['q_proj', 'v_proj', 'output_proj'], apply_lora_to_mlp: False, and apply_lora_to_output: False`—can the tune_to_peft_adapter_weights logic be used in this case? Thank you.

cinjon:
Why not merge this in?

cinjon:
Has anyone had success compiling the moe here and training?

Issue #2694, opened 2025-05-07
Fix Compile LinearCrossEntropyLoss


pbontrager:
The setup to the problem is that we have bunch of tokens and a mask ([False, False, ..., True, True]) of which tokens are trainable. Instead of computing the logits for all of the tokens and then computing cross entropy loss, we instead want to mask the tokens to only compute the logits (final layer of the LLM) for the trainable tokens. This looks like:

```python
loss = LinearCrossEntropy(model.logits_layer)
model.skip_logits_layer = True

...

for tokens, labels in dataset:
    h = model(tokens)
    l = loss(h, labels)
```

A simplified version of what happens inside of the loss is:

```python
def forward(self, hidden, labels):
    mask = labels != -100
    if isinstance(hidden, DTensor):
        mesh, placements = hidden.device_mesh, hidden.placements
        local_hidden = hidden.to_local()[mask]
        hidden = DTensor.from_local(local_hidden, mesh, placements)
    else:
        hidden = hidden[mask]

    logits = self.logits_layer(hidden)
    if isinstance(logits, DTensor):
        logits = logits.full_tensor()

    return F.cross_entropy(logits.float(), labels)
```

This solution works with TP but fails when the loss module is compiled and it seems to fail on the isinstance check. I'm not sure how to further debug this so I'm putting up for help.

Notes:
- If you remove the masking and logits_layer from the loss, the loss compiles for most models but still fails on models with tied embeddings
- The error message for the above example (no tied embeddings) `torch._dynamo.exc.InternalTorchDynamoError: TypeError: unhashable type: non-nested SymInt`. Seems to error on isinstance.
- The provided code doesn't take advantage of loss_parallel which could potentially help to get rid of the DTensor checks. The masking check could be done and then we'd only compile the logits_layer call and cross_entropy call together. The loss_parallel solution might not work for RL linear losses though.

PR #2824, opened 2025-06-13
fix loss compile for dtensor

felipemello1:
#### Context
- [ ] add a new feature
- [x] fix a bug
- [ ] other (please add here)

fixes https://github.com/pytorch/torchtune/issues/2694

#### Changelog
1) Fix TLDR: move .to_local() out of the compile region

2) minor adjustments for efficiency:
- replace masking with where + select_index
- replaced 'loss == 0', which calls .item(), with torch.where
- replaced initialization 'sum=0' with a tensor

3) Minor naming/comments/docs changes

#### Test plan
8B + FSDP + TP
```
PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True tune run --nproc_per_node 2 full_finetune_distributed --config llama3_1/8B_full tokenizer.max_seq_len=16384 metric_logger=torchtune.training.metric_logging.WandBLogger batch_size=2 compile=True optimizer_in_bwd=True gradient_accumulation_steps=1 tensor_parallel_dim=1 tensor_parallel_plan._component_=torchtune.models.llama3.base_llama_tp_plan dataset.packed=True
```
image

3B + FSDP
image


felipemello1:
cc: @nathan-az

nathan-az:
I like the separation between the distribution logic and the CE calculation.

This implementation passively includes loss parallelism, just on the sequence rather than feature dim that the ctx manager requires. If this all works, it won't benefit (IIUC) from also using the loss parallel context manager.

The trade-off between this method and the context manager, is that since the context manager distributes the CE calculation over the _feature_ dim, if there's a lot of masking on the sequence dim, then the feature dim sharding distributes the work better (but might require more collectives - I'm not sure).

Still, any parallelism of the loss will see an improvement. That said, I had issues running my test config on your branch. Have I made any obvious blunders?

<details>

```yaml
batch_size: 8
checkpointer:
  checkpoint_files:
    max_filename: '00004'
  model_type: LLAMA3
  output_dir: ${output_dir}
  recipe_checkpoint: null
  checkpoint_dir: models/llama3_1_8b
clip_grad_norm: null
compile:
  model: true
  loss: true
  optimizer_step: false
  scale_grads: true
custom_sharded_layers: []
data_parallel_shard_dim: 1
data_parallel_replicate_dim: 1
tensor_parallel_dim: 8
context_parallel_dim: 1
enable_fp8_training: false
tensor_parallel_plan:
  _component_: torchtune.models.llama3.base_llama_tp_plan
dataset:
  packed: true
  train_on_input: true
  split: train[:10%]
device: cuda
dtype: bf16
activation_offloading_use_streams: false
epochs: 1
fsdp_cpu_offload: false
log_every_n_steps: 1
log_level: INFO
log_peak_memory_stats: true
loss:
  _component_: torchtune.modules.loss.LinearCrossEntropyLoss
  num_output_chunks: 16
max_steps_per_epoch: 5
metric_logger:
  _component_: torchtune.training.metric_logging.MLFlowLogger
  experiment_name: benchmark_curr
  run_name: delete_me
optimizer:
  _component_: torchao.optim.AdamW8bit
  lr: 4.0e-05
optimizer_in_bwd: true
output_dir: outputs
resume_from_checkpoint: false
seed: 100
shuffle: true
tokenizer:
  max_seq_len: 8192 # 8192 16384 32768 65536 131072
  path: models/llama3_1_8b/original/tokenizer.model
model:
```

</details>

The above doesn't run whether loss is or isn't compiled. Uncompiled the stderr includes a lot of spam from `vectorized_gather_kernel`. The compiled error is probably more informative:

```
[rank0]: torch._dynamo.exc.TorchRuntimeError: Dynamo failed to run FX node with fake tensors: call_function <built-in function linear>(*(DTensor(local_tensor=FakeTensor(..., device='cuda:0', size=(4026, 32768), dtype=torch.bfloat16), device_mesh=DeviceMesh('cuda', [0, 1, 2, 3, 4, 5, 6, 7], mesh_dim_names=('tp',)), placements=(Replicate(),)), DTensor(local_tensor=FakeTensor(..., device='cuda:0', size=(16032, 4096), dtype=torch.bfloat16), device_mesh=DeviceMesh('cuda', [0, 1, 2, 3, 4, 5, 6, 7], mesh_dim_names=('tp',)), placements=(Shard(dim=0),)), None), **{}): got RuntimeError('a and b must have same reduction dim, but got [4026, 32768] X [4096, 128256].')

[rank0]: from user code:
[rank0]:    File "/workspace/torchtune/torchtune/modules/loss/cross_entropy_loss.py", line 106, in torch_dynamo_resume_in_compute_cross_entropy_at_87
[rank0]:     logits = self.linear_projection(valid_hidden)  # [num_valid, vocab_size]
[rank0]:   File "/opt/conda/lib/python3.11/site-packages/torch/nn/modules/module.py", line 1870, in _call_impl
[rank0]:     return inner()
[rank0]:   File "/opt/conda/lib/python3.11/site-packages/torch/nn/modules/module.py", line 1821, in inner
[rank0]:     result = forward_call(*args, **kwargs)
[rank0]:   File "/opt/conda/lib/python3.11/site-packages/torch/nn/modules/linear.py", line 125, in forward
[rank0]:     return F.linear(input, self.weight, self.bias)
```

EDIT: Does any adjustment need to be made due to the TP plan including `input_layouts=Shard(1),` for the `output` module?

felipemello1:
you are correct @nathan-az , in the config i tried, i had tp=1 -.-` My bad

felipemello1:
@nathan-az, not sure how to fix it. Do you have any ideas?

If we take a step back, i though that maybe we could:
1. not do masking, but i dont like that constraint too much. I tried it, and there are still issues. The targets.shape != logits.shape. I tried to distribute the targets, it ran, but got bad loss.
2. Do not do TP on the linear projection
3. Find some other way

nathan-az:
No luck, I had similar thoughts and tried distributing the targets as well. `compile` really doesn't like the shape changing from the masking. I doubt compile offers enough optimisation to justify removing the masking. I'd be happy to proceed without using compile, and just parallelising the loss either in your sequence parallel fashion or using the ctx manager in my PR.

The most promising method if we want compile to work is probably to not do TP on the linear projection, but that is the most memory intensive layer in the model so users who are more hardware constrained may start seeing memory issues.

felipemello1:
"The most promising method if we want compile to work is probably to not do TP on the linear projection, but that is the most memory intensive layer in the model so users who are more hardware constrained may start seeing memory issues."

if we can produce a small example, i think we can ask compile folks for help @nathan-az 

We could also do:
1) if tp=1, keep the way in this PR
2) if tp>1, do the way you are coding. Then we can benchmark memory from masking/chunking + compile vs Parallelism

nathan-az:
Hmm in this case I don't want to create more duplicate code. Let me see if I can work out a modified version of your code which will compile for non-dtensor and still works distributed.

nathan-az:
@felipemello1 take a look at the last commit: https://github.com/pytorch/torchtune/pull/2782/commits/40df86da0e5314016739996d218ffefeeb4ad0a2

Some interesting takeaways:

* DTensor slicing (or `index_select`) actually works and kept the tensor sharded in forward, but raised errors in backward. Thus I returned to the `.to_local()` and `from_local` pattern
* As a result of the above, I pulled the slicing/masking logic out of the compiled function. This was to remove the "if dtensor" branch, so now `compile` works in the non-loss parallel case. (In TP, compile also works if loss parallel is not enabled - the context manager throws an error unfortunately)

I think the pattern and the way the classes are structured now should help from an engineering perspective with trying new methods. I suspect that it should be possible to parallelise the loss calculation across the sequence as you did, and have compile work. But for now I'm quite happy that the solution:

* compiles in the non-loss parallel case
* has working loss parallelism 
* reduces memory usage in all cases

I'll mark that PR as "ready for review" again - feel free to leave comments! I think since you're a maintainer you can also edit the PR directly, if you want to consolidate our changes there I'm happy for you to do so.

PR #2827, opened 2025-06-13
skip compiling opt step instead of erroring if opt_in_bwd=True

felipemello1:
#### Context
- [ ] add a new feature
- [ ] fix a bug
- [x] other (please add here)

Changelog: as in the title


felipemello1:
> Why? This doesn't happen that far into the process and it's a simple solution to just change compile optimizer step to false.
> 
> If anything, you could provide a better error message like "If you would still like to compile the model and loss, you can specify everything in the config like
> 
> blah blah blah"

I was doing llama 70B. It takes several minutes for it to start (not sure why). And then it errored. So for me, it felt far into the process. If you dont like skipping and prefer it to just error, i could do this logic in the __init__ instead.

PR #2834, opened 2025-06-18
Update get_merged_lora_ckpt  for dist checkpoints

ankitageorge:
#### Context
- [ ] add a new feature
- [x] fix a bug
- [ ] other (please add here)

#### Changelog
* There was a bug where the get_merged_lora_ckpt method doesn't work in distributed recipes when async checkpointing was enabled, because the existing method expects the data to be all on rank-0. This PR adds a new method get_merged_lora_dist_ckpt to be used for dist checkpoints and appropriately calls it when needed
* note that this doesn't work for nf4 tensors, I've added a comment in the code as well
* When testing on 70B model with the lora_finetune_distributed recipe, it took ~20s, compared to 63s for gathering and ~250s for saving (~310s total), saving almost 5 minutes
* since we save the adapter weights separately, change the order to save that first, so that the trainer is unblocked faster. This because dist checkpointer blocks on itself for two separate calls, so if we save adapter checkpoint after normal, then we are blocking on the normal checkpoint saving time, which is longer than adapter. 

#### Test plan

- [x ] manually run any new or modified recipes with sufficient proof of correctness

#### UX

- [ x] I did not change any public API

### DIFF
--- a/torchtune/modules/peft/_utils.py
+++ b/torchtune/modules/peft/_utils.py
@@ -195,6 +195,7 @@ def get_merged_lora_ckpt(
     state_dict: dict[str, Any],
     rank: int,
     alpha: float,
+    use_distributed_barriers: bool = False,
 ) -> dict[str, Any]:
     """
     Merge LoRA weights into the base model format for efficient inference.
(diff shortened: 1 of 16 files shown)


torchtune/modules/peft/_utils.py
```diff
@@ -256,6 +257,123 @@ def get_merged_lora_ckpt(
...
+        alpha (float): The alpha value used for scaling LoRA decompositions.
+    Returns:
+        dict[str, Any]: The merged state dict.
+    """
+
+    lora_modules = _get_lora_modules(state_dict)
+    lora_moe_modules = _get_lora_moe_modules(state_dict)
+
+    # Create a simple module for matrix multiplication
+    class MatMulModule(torch.nn.Module):
```
joecummings: Why do we need this instead of just calling matmul directly?

torchtune/modules/peft/_utils.py
```diff
@@ -256,6 +257,123 @@ def get_merged_lora_ckpt(
...
+
+    for module in sorted(lora_modules.union(lora_moe_modules)):
+        # TODO: we don't currently support DoRA for MoE layers
+        if "experts" in module:
+            for param in ["gate", "up", "down"]:
+                lora_a_weight = state_dict[f"{module}.lora_{param}_a"]
+                lora_b_weight = state_dict[f"{module}.lora_{param}_b"]
+
+                # Create a simple module for transpose operation
+                class TransposeModule(torch.nn.Module):
```
joecummings: Same here: why does this need to be a transpose module?

torchtune/modules/peft/_utils.py
ankitageorge: these operations don't work properly on d-tensors. it's what was causing the hangs

joecummings (approved):

torchtune/modules/peft/_utils.py
joecummings: "doesn't work properly" - can you expand on that?

ankitageorge: ok actually, I think it's not needed. Good catch. I thought it was causing problems, but I just re-tested without it and it still works. I'll get rid of them and just add barriers to the existing method.

PR #2852, opened 2025-06-26
Iterable Dataset

felipemello1:
#### Context
- [x] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

Enable Iterable datasets in torchtune.

CONTEXT: built on top of ongoing PR step-based-ckpt: https://github.com/pytorch/torchtune/pull/2384

#### TIps when reviewing this pr

Follow this order:

0. recipes/configs/llama3_2/3B_full.yaml: see the configs
1. torchtune/datasets/_iterable_base.py: base class for iterable dataset
2. torchtune/datasets/_hf_iterable.py: ds based on HF -- Can be replaced easily. Downstream does not expect HF.
3. torchtune/datasets/_interleaved.py: interleave the datasets
4. torchtune/data/_metrics.py: metrics transform to create the metrics
5. torchtune/data/_aggregator.py: aggregate the metrics at the recipe level
6. recipes/full_finetune_distributed.py: everything put together
7. unit tests

torchtune/datasets/_hf_iterable.py
#### Changelog

1. Datasets are infinite
2. User doesn't define epochs anymore, but training steps (how many times we update the optimizer)
3. Support for dataset mixing -- follow up PRs is to enable curriculum learning
4. Support for dataset metric logging -- User can understand epoch per dataset, distribution of token lens, etc. Easy to add new metrics.
5. HF agnostic. Even though the current dataset is HF, the dataloader, packed, datamixing, metric logging is agnostic to it
6. Well tested in distributed setting -- WARNING: need better testing for multiprocess dataloader. It doesnt guarantee determinism, so I postponed testing this setting

Config and builder design based on the discussions after this RFC: https://github.com/pytorch/torchtune/pull/2785

Next steps:
7. Gather feedback on metric logging. E.g. we can add more aggregation types.
8. Polish the code a little bit
9. Add packing from this RFC: https://github.com/pytorch/torchtune/pull/2819
10. Add curriculum learning
11. Docs?

#### Test plan
image

image
image

UNTESTED: resume from ckpt in the recipe. However, we have plenty of tests showing that resuming works for these iterable datasets.


torchtune/datasets/_slimorca.py
```diff
@@ -94,3 +95,72 @@ def slimorca_dataset(
             )
         return PackedDataset(ds, max_seq_len=tokenizer.max_seq_len)
     return ds
+
+
+def slimorca_iterable_dataset(
```
felipemello1, line 100: added here to demonstrate datamix iterable dataset with this example. Personally, i dislike exposing all of the args and defaults. I would prefer to expose only whats specific to this builder.

torchtune/datasets/_sft.py
```diff
@@ -178,3 +180,99 @@ def __call__(self, sample: Mapping[str, Any]) -> dict[str, Any]:
...
+                tokenized_dict["tokens"][1:],
+            )
+        )
+        tokenized_dict["labels"].append(CROSS_ENTROPY_IGNORE_IDX)
+        assert len(tokenized_dict["tokens"]) == len(tokenized_dict["labels"])
+
+        return tokenized_dict
+
+
+def sft_iterable_dataset(
```
felipemello1, line 233: only purpose is to hardcode the 'output_transform'

torchtune/datasets/_interleaved.py
```diff
@@ -0,0 +1,125 @@
...
+                sample = next(child_iters[ds_name])
+                yield sample
+            except StopIteration:
+                # Per the design, child datasets must be infinite.
+                # We re-initialize to allow for continuous operation but warn loudly
+                # as this may indicate a design problem in the child dataset.
+                logger.warning(
+                    f"Child dataset {self._datasets[ds_name].dataset_name} was exhausted. "
+                    "This is unexpected for an infinite dataset. Re-initializing its iterator."
+                )
```
felipemello1: not 100% sure i like this

torchtune/datasets/_alpaca.py
```diff
@@ -101,3 +102,64 @@ def alpaca_dataset(
 original Alpaca dataset, `yahma/alpaca-cleaned <https://huggingface.co/datasets/yahma/alpaca-cleaned>`_.
 See the dataset page and :func:`~torchtune.datasets.alpaca_dataset` for more details.
 """
+
+
+def alpaca_iterable_dataset(
```
felipemello1, line 107: added here to demonstrate datamix iterable dataset with this example. Personally, i dislike exposing all of the args and defaults. I would prefer to expose only whats specific to this builder.

torchtune/data/_aggregator.py
```diff
@@ -0,0 +1,342 @@
...
+                # queue -> list
+                values = list(state["values"])
+
+                # Sort to get percentiles efficiently
+                sorted_values = sorted(values)
+                n = len(sorted_values)
+
+                # Each stat becomes its own metric
+                # For percentiles, it is an approximattion by computing avg of averages
+                metrics[(ds_name, f"{metric_name}_mean")] = {
```
Darktex: Nah, just use `torch.quantile` ;)

https://docs.pytorch.org/docs/stable/generated/torch.quantile.html

torchtune/data/_aggregator.py
```diff
@@ -0,0 +1,342 @@
...
+                for key, metric_dict in rank_metrics.items():
+                    # A key is a tuple (dataset, metric)
+                    grouped[key].append(metric_dict)
+
+        # Reduce based on aggregation type
+        reduced = {}
+        if not grouped:
+            return reduced
+
+        for key, metric_dicts in grouped.items():
```
Darktex: Seems weird to write the code below twice. Can you factor it out and simply call it after reducing on rank zero?

torchtune/data/_metrics.py
```diff
@@ -0,0 +1,98 @@
...
+class Metric:
+    """A self-describing metric object."""
+
+    dataset_name: str
+    name: str
+    value: Union[int, float, str]
+    agg_type: AggregationType
+
+
+class MetricTransform(Protocol):
```
Darktex: Can you explain a bit more about what's the API? It's not clear to me because we have `__call__` which doesn't have a name that tells me what it does :D

torchtune/data/_metrics.py
```diff
@@ -0,0 +1,98 @@
...
+    """Protocol for metric transforms."""
+
+    def set_dataset_name(self, dataset_name: str) -> None:
+        ...
+
+    def __call__(self, sample: dict[str, Any]) -> dict[str, Any]:
+        ...
+
+
+class StandardMetricTransform(MetricTransform):
```
Darktex: Hmmm, the name makes me think this is an ABC. Maybe `DefaultTrainingMetricTransform`?

torchtune/datasets/_alpaca.py
Darktex: But you are doing this with ``**load_dataset_kwargs**, right? Or did you mean something else?

torchtune/datasets/_alpaca.py
```diff
@@ -101,3 +102,64 @@ def alpaca_dataset(
...
+        >>> dataloader = StatefulDataLoader(alpaca_ds, batch_size=8)
+        >>> for batch in dataloader:
+        >>>     print(f"Batch size: {len(batch)}")
+        >>> Batch size: 8
+    """
+    message_transform = AlpacaToMessages(
+        train_on_input=train_on_input, column_map=column_map
+    )
+
+    return sft_iterable_dataset(
```
Darktex, line 155: This looks clean to me

torchtune/datasets/_alpaca.py
Darktex: nit: it's a function, so... `get_alpaca_iterable_dataset`?

torchtune/datasets/_iterable_base.py
```diff
@@ -0,0 +1,43 @@
...
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+from abc import ABC, abstractmethod
+from typing import Any, Iterator
+
+from torch.utils.data import IterableDataset
+
+
+class TuneIterableDataset(IterableDataset, ABC):
```
Darktex: API looks good. Nit: if we are overriding `__iter__` (which was coming from `IterableDataset` anyway) then I would also just put an `__init__` here.

Also: should we enforce having a `transforms` property, and a `tokenizer` property?

torchtune/datasets/_sft.py
```diff
@@ -178,3 +180,99 @@ def __call__(self, sample: Mapping[str, Any]) -> dict[str, Any]:
             tokenized_dict = transformed_sample
 
         return tokenized_dict
+
+
+class SFTOutputTransform(Transform):
```
Darktex, line 186: This transform is critical for end to end performance, so this is a piece where we need to spend cycles to optimize.

I see several ways to find more performance in this code:

1. No numpy. Even if we haven't moved to GPU yet, doing everything in torch ensures that the code will work regardless and makes it more robust
2. I think `np.where` allocates?
3. Asserts are debug-only statements that get disabled if you pass the -O flag, so this check feels more of a runtime check

I asked a LLM to rewrite given these contraints and it gave me this, which looks reasonable on the surface:

```python
import torch

CROSS_ENTROPY_IGNORE_IDX = -100          # set to whatever you use

class SFTOutputTransform(Transform):
    """
    Build the `"labels"` tensor for causal-LM SFT.

    Expects each dataset element to contain **1-D** torch tensors
    * ``"tokens"`` – token IDs, dtype=torch.long
    * ``"mask"``   – bool/int where **True** marks positions to ignore

    Produces ``"labels"`` of the same shape such that

        labels[t] =  tokens[t+1]                # shift left
        labels[t] =  IGNORE_IDX  if mask[t+1]   # respect mask
        labels[-1] = IGNORE_IDX                 # last token has no target

    All ops are vectorised; only one fresh tensor (`labels`) is allocated.
    """

    __slots__ = ()      # tiny perf win (avoids per-instance __dict__)

    def __call__(self, sample: Mapping[str, Any]) -> Mapping[str, Any]:
        try:
            tokens: torch.Tensor = sample["tokens"]
            mask:   torch.Tensor = sample["mask"]
        except KeyError:
            raise ValueError(
                "SFTOutputTransform expects 'tokens' and 'mask' in the sample."
            )

        if tokens.ndim != 1 or mask.ndim != 1:
            raise ValueError("Both 'tokens' and 'mask' must be 1-D tensors.")

        # ── build labels ────────────────────────────────────────────────
        # 1. pre-fill with IGNORE so we don’t need extra assignments later
        labels = tokens.new_full(tokens.shape, CROSS_ENTROPY_IGNORE_IDX)

        # 2. left-shift via cheap views (no copy)
        labels[:-1].copy_(tokens[1:])

        # 3. apply mask in-place (single fused kernel on GPU/CPU)
        labels[:-1].masked_fill_(mask[1:].bool(), CROSS_ENTROPY_IGNORE_IDX)

        # (labels[-1] is already IGNORE_IDX from the new_full above)

        # ── return a shallow-copied mapping so the original sample stays intact
        out = dict(sample)
        out["labels"] = labels
        return out
```

torchtune/datasets/_interleaved.py
```diff
@@ -0,0 +1,125 @@
...
+            ]
+            raise ValueError(
+                f"Duplicate dataset names detected: {duplicates}. All {names=}"
+                f"Please provide a unique 'dataset_name' for each dataset in the interleaved list."
+            )
+
+        self._sampling_generator = torch.Generator().manual_seed(seed)
+
+        # Normalize weights to sum to 1
+        # TODO: make it a property? rely on ds.weight?
```
Darktex: I'd make it a property so it remains visible. Very cheap anyway

torchtune/datasets/_interleaved.py
```diff
@@ -0,0 +1,125 @@
...
+    def dataset_name(self) -> str:
+        return self._dataset_name
+
+    def __iter__(self) -> Iterator[dict[str, Any]]:
+        """Interleave samples from child infinite datasets"""
+        child_iters = {name: iter(ds) for name, ds in self._datasets.items()}
+
+        while True:
+            # Sample which dataset to use
+            ds_idx = torch.multinomial(
```
Darktex: Hmmmmm, I think we should also log what we did. It's reasonable and cheap to accumulate a list that maps `iteration_id` to `dataset_id` in `self.datasets`. When this guy dumps its state, this log should be part of it

torchtune/datasets/_interleaved.py
Darktex: Let's do this: simply have a subclass for `InfiniteIterable` so this is super explicit

torchtune/datasets/_iterable_base.py
Darktex: Also: let's make a subclass for `InfiniteIterable` so that this is very explicit (we expect to use this all the time) and another for `FiniteIterable` with explicit semantics that we always drop last

Darktex: We need this guy to interact with packing and IIUC I don't believe this is currently happening?

The algo we should implement is this:

1. One batch can be made of *multiple* calls to next. We keep taking until we exceed the max seq len. When we do, we put the last one aside (we'll use it to start the next batch), pad the current one to max len and return.
2. The calls to next will go to the *interleaved* dataset, therefore we automatically construct mixed batches from multiple datasets without much effort
3. Also, every time we call next we should make space for logging transforms (which we are, you already wrote them). I think it's ok to make your metrics transforms and aggregators an optional property here so the semantics are clearer

Darktex (changes requested):
Great PR! I mainly had a question on the interaction with packing and on the SFT transform

torchtune/data/_aggregator.py
felipemello1: replaced it with handlers, it should look better now

torchtune/data/_metrics.py
felipemello1: done

torchtune/datasets/_iterable_base.py
felipemello1: I thought about it and i dont think that it would work. This is the base class also for packedDataset and InterleavedDatset. Concepts like transform and tokenizer dont apply. The __init__ would only noise IMO. We could standardize that every dataset needs a name and weights, but then we have to call super.__init__. I didnt see how this would benefit the implementation. Let me know if you disagree

felipemello1: we have packing here: https://github.com/pytorch/torchtune/pull/2819

torchtune/datasets/_alpaca.py
felipemello1: the get makes sense, but its not the pattern we have in tune :/

torchtune/datasets/_hf_iterable.py
```diff
@@ -0,0 +1,281 @@
...
+        # Distributed setup
+        world_size, rank = 1, 0
+        if dist.is_initialized():
+            world_size = dist.get_world_size()
+            rank = dist.get_rank()
+
+        # Load and shard dataset
+        ds = load_dataset(**load_dataset_kwargs)
+
+        # Use to_iterable_dataset for streaming datasets
```
ebsmothers: I don't understand this comment. Based on L185 it seems you're using to_iterable_dataset for non-streaming datasets

torchtune/datasets/_hf_iterable.py
```diff
@@ -0,0 +1,281 @@
...
+            if desired_shards % total_workers == 0:
+                num_shards = desired_shards
+            else:
+                num_shards = total_workers * (
+                    (desired_shards + total_workers - 1) // total_workers
+                )
+
+            # If the dataset is not streaming and has a defined length,
+            # we cannot have num_shards > dataset_size.
+            if not load_dataset_kwargs.get("streaming", False) and hasattr(
```
ebsmothers: This is redundant inside the current if statement

torchtune/datasets/_hf_iterable.py
```diff
@@ -0,0 +1,281 @@
...
+        # Load and shard dataset
+        ds = load_dataset(**load_dataset_kwargs)
+
+        # Use to_iterable_dataset for streaming datasets
+        if not load_dataset_kwargs.get("streaming", False):
+
+            # Define number of shards based on (world_size, num of shards per GPU, dataloader workers)
+            # E.g. world_size=2, num_shards_per_rank=16, dataloader_workers=3
+            # we will try 2*16 = 32 shards. Since 32 is not a multiple of 3, we will do 36 shards.
+            # Each rank gets 16 shards, each dataloader worker in that rankgets 6 shards.
```
ebsmothers: this math doesn't seem quite right
```suggestion
            # we will try 2*16 = 32 shards. Since 32 is not a multiple of 6, we will do 36 shards.
            # Each rank gets 18 shards, each dataloader worker in that rank gets 6 shards.
```

torchtune/datasets/_hf_iterable.py
```diff
@@ -0,0 +1,281 @@
...
+            # If the dataset is not streaming and has a defined length,
+            # we cannot have num_shards > dataset_size.
+            if not load_dataset_kwargs.get("streaming", False) and hasattr(
+                ds, "__len__"
+            ):
+                dataset_size = len(ds)
+                if num_shards > dataset_size:
+                    raise ValueError(
+                        f"Number of shards ({num_shards}) is greater than the dataset size ({dataset_size})."
+                        f"Please decrease num_shards_per_rank."
```
ebsmothers: nit but this can also happen due to total_workers > dataset_size (since we round up). Probably an edge case but if e.g. `torch.utils.data.get_worker_info()` returns a really large number then this is not actually the right fix

torchtune/datasets/_hf_iterable.py
```diff
@@ -0,0 +1,281 @@
...
+
+        If shuffle_buffer_size is set, it will shuffle the dataset at the beginning of each epoch
+        when set_epoch is called.
+
+        An additional metric "num_epochs" is added to the sample.
+        """
+
+        while True:  # Infinite iteration
+            epoch_seed = self._seed + self._num_epochs
+            self._ds.set_epoch(epoch_seed)
```
ebsmothers: I'm confused.. isn't this bumping epoch by `self._seed` each epoch (when it should be bumping by 1)?

torchtune/datasets/_hf_iterable.py
```diff
@@ -0,0 +1,281 @@
...
+            self._num_epochs += 1
+
+    def state_dict(self) -> dict[str, Any]:
+        """
+        The dataset returns its own state directly, without namespacing.
+        """
+        hf_state = self._ds.state_dict()
+        state = {
+            "num_epochs": self._num_epochs,
+            "seed": self._seed,
```
ebsmothers: Why do we save seed but not restore on load?

torchtune/datasets/_interleaved.py
ebsmothers: Where did this one land? I don't see `InfiniteIterable` anywhere (personally I don't know enough yet to have a strong preference here, just wanna understand where things currently stand)

torchtune/datasets/_interleaved.py
```diff
@@ -0,0 +1,158 @@
...
+
+        # Extract weights from child datasets
+        weights = []
+        for ds in datasets:
+            weight = ds.sampling_weight
+            if isinstance(weight, dict):
+                raise ValueError(
+                    f"Child dataset '{ds.dataset_name}' returned a dict for sampling_weight, "
+                    f"indicating it's a composite dataset (likely InterleavedDataset). "
+                    f"Nested interleaving is not supported. Please flatten the dataset hierarchy."
```
ebsmothers: Is this something we want to support eventually? I thought this was one of the things mentioned by @Darktex as being needed for proper datamixing

torchtune/data/metrics/readme.md
```diff
@@ -0,0 +1,176 @@
...
+
+## Overview
+
+The metrics module provides a robust system for tracking and aggregating training metrics across multiple datasets and distributed environments. It follows a **strategy pattern** design with pluggable aggregation handlers to efficiently handle different types of metrics.
+
+## Architecture Overview
+
+```
+┌────────────────────────────────────────────────────┐
+│                Training Loop                       │
```
ebsmothers, line 11: Nice-looking diagram

torchtune/data/metrics/readme.md
```diff
@@ -0,0 +1,176 @@
...
+│  • Examples: tokens_seen, seq_len, samples_seen    │
+└─────────────────────┬──────────────────────────────┘
+                      │ list[Metric]
+┌─────────────────────▼──────────────────────────────┐
+│                 MetricsAggregator                  │
+│  • Aggregates metrics across samples and ranks     │
+│  • Uses pluggable AggregationHandlers              │
+│  • Handles distributed reduction                   │
+└─────────────────────┬──────────────────────────────┘
+                      │ {prefix_dataset/metric: value}
```
ebsmothers: What does "prefix_dataset/metric" mean here?

torchtune/data/metrics/_metric_transform.py
```diff
@@ -0,0 +1,133 @@
...
+    """Applied to each dataset sample to generate per-sample metrics for training tracking.
+
+    Creates Metric objects that are later aggregated by 'MetricsAggregator'. This separation
+    of concerns ensures metrics are correctly aggregated even with multiple dataloader
+    workers and in distributed settings."""
+
+    def __init__(self):
+        # dataset_name is set by the dataset using set_dataset_name
+        self.dataset_name: Optional[str] = None
+        self.new_metric: Optional[Callable] = None
```
ebsmothers: I don't really understand.. why is this called "new_metric"?

torchtune/data/metrics/_metric_transform.py
```diff
@@ -0,0 +1,133 @@
...
+
+        # Create metrics for this sample
+        return [
+            self.new_metric(name="samples_seen", value=1, agg_type=AggregationType.SUM),
+            self.new_metric(
+                name="tokens_seen", value=token_len, agg_type=AggregationType.SUM
+            ),
+            self.new_metric(
+                name="seq_len", value=token_len, agg_type=AggregationType.DISTRIBUTION
+            ),
```
ebsmothers: A minor thing, but to me metrics having the same value but different aggregation types should not actually be represented as distinct metrics. Like I should be able to just define how a metric is computed for a given sample, then separately choose different types of aggregation as needed

torchtune/data/metrics/_metric_aggregator.py
```diff
@@ -0,0 +1,290 @@
...
+    MaxAggHandler,
+    MeanAggHandler,
+    MetricState,
+    MinAggHandler,
+    SumAggHandler,
+)
+from torchtune.data.metrics._metric_transform import AggregationType, Metric
+
+
+class MetricsAggregator:
```
ebsmothers, line 29: A high level comment: the relationship between this and the agg handlers is not super clear to me. It seems like we are using a registry pattern where the handlers are responsible for defining the actual aggregation logic. But then the all-gather happens in here. (Separately I stand by my claim that it would be better to hold off on more complex cases like distribution aggregators so as not to boil the ocean here.)

recipes/full_finetune_distributed.py
```diff
@@ -324,6 +309,11 @@ def setup(self, cfg: DictConfig) -> None:
         Setup the recipe. This includes training state (if resume_from_checkpoint is True),
         model, tokenizer, loss, optimizer, lr scheduler, sampler, and dataloader.
         """
+        if cfg.get("dataset_val") is not None:
+            raise NotImplementedError(
+                "Validation is not supported yet with iterable datasets."
+            )
```
ebsmothers: Is there a specific technical reason here? Or we just haven't gotten to it yet

recipes/full_finetune_distributed.py
```diff
@@ -273,18 +273,20 @@ def __init__(self, cfg: DictConfig) -> None:
...
             seed=cfg.seed, debug_mode=cfg.get("cudnn_deterministic_mode", None)
         )
-        self.epochs_run = 0
-        self.total_epochs = cfg.epochs
-        self.max_steps_per_epoch = cfg.max_steps_per_epoch
         self.global_step = 0
 
+        # Step-based training support
+        self.num_training_steps = cfg.num_training_steps
+        self._dataset_metrics_log_freq = cfg.get("dataset_metrics_log_freq", 100)
```
ebsmothers: Very minor but does this really need to be configured separately from log_every_n_steps? (Edit: I see you already added a TODO about this below, I guess you know my vote)

recipes/full_finetune_distributed.py
```diff
@@ -968,180 +946,176 @@ def train(self) -> None:
...
+                        / (time_per_step * self.world_size),
+                    }
+                    if self._log_peak_memory_stats:
+                        log_dict.update(training.get_memory_stats(device=self._device))
+                    if self._clip_grad_norm is not None:
+                        log_dict.update({"grad_norm": grad_norm})
+                    self._metric_logger.log_dict(log_dict, step=self.global_step)
+
+                # Log dataset metrics
+                # #TODO: it requires all_gather. Should we keep a separate log_freq for this?
```
ebsmothers: Worth benchmarking, but if it's just a few floats I think it shouldn't matter

ebsmothers (commented):
I read every line of this PR. (Kidding but I tried to at least look at most of the important stuff.) Thanks for taking on this massive set of changes, I think the dataset classes are a big improvement

torchtune/datasets/_interleaved.py
felipemello1: i made changes but didnt push them yet. I added a dummy class that does nothing:

```
class InfiniteTuneIterableDataset(TuneIterableDataset):
    """Abstract base class for infinite datasets, which yield samples indefinitely.
    It only purpose is to make it explicit that the dataset is expected to be infinite, i.e. 
    it never exhausts. This is helpful to avoid complexity due to some rank hanging because
    of lack of data""
    pass
```

and replaced this logger.warning with raise ValueError.

I think its better to have zero tolerance. Datasets that are not infinite need work to make sure no rank hangs.

torchtune/datasets/_interleaved.py
felipemello1: i have made the changes. We can do nested now

torchtune/datasets/_hf_iterable.py
felipemello1: just checking if you were paying attention. Will add it to load_state_dict :)

thanks for catching it

torchtune/datasets/_hf_iterable.py
felipemello1: Yeah, from their docs, it should be just: ```dataset.set_epoch(epoch)```

I will make the change

torchtune/data/metrics/readme.md
felipemello1: will update it. It should be: {prefix}_{dataset_name}/{metric_name}

prefix is given in the recipe when we call `aggregator.get_metrics_for_logging(prefix="train")`, so we can pass "train" and "valid" (or actor type, for example).

torchtune/data/metrics/_metric_transform.py
felipemello1: i can rename/do something better here. Its just a placeholder for the partial.

```
self.new_metric = partial(Metric, dataset_name=dataset_name)
```

torchtune/data/metrics/_metric_transform.py
felipemello1: hmm, so agg_type being a List[AggregationType]?

```
self.new_metric(
                name="tokens_seen", value=token_len, 
                agg_type=[AggregationType.SUM, AggregationType.MEAN
            ),
```

I dont know if the extra complexity is worth it. Adding two metrics is cheap. wdyt?

```
Today the user can just do:
self.new_metric(
                name="tokens_seen_sum", value=token_len, 
                agg_type=AggregationType.SUM
            ), 
self.new_metric(
                name="tokens_seen_mean", value=token_len, 
                agg_type=AggregationType.MEAN
            ), 
 ```

File: torchtune/data/metrics/_metric_aggregator.py (hunk 25 above)
Comment (felipemello1) [reply torchtune/data/metrics/_metric_aggregator.py:29]:
```
the handlers are responsible for defining the actual aggregation logic. But then the all-gather happens in here.
```

why is that a contradiction?

1. The MetricsAggregator calls the handler.finalize_local_agg
2. then does a single all_gather to get the results from all ranks for all metrics
3. Then calls handler._finalize_dist_agg([aggregated_results_per_rank]*n_ranks)

Do you wanna suggest a different way of doing it? Or is it hard to spot this pattern in the code?

Comment (felipemello1) [reply torchtune/data/metrics/_metric_aggregator.py:29]: > my claim that it would be better to hold off on more complex cases like distribution aggregators

If we dont do aggregation across ranks, we wouldnt be able to count things like "tokens_seen", right? :/

Or do you mean that we should delete `DistributionAggHandler`? To clarify, this distribution has nothing to do with multiple gpus. Its just stats, e.g. std, percentiles, max, min, etc. Maybe i should rename if its causing confusion.

File: recipes/full_finetune_distributed.py (hunk 26 above)
Comment (felipemello1) [reply recipes/full_finetune_distributed.py]: validation datasets are not infinite!!! Need to figure out how to solve this one, but it wont be on this PR

File: torchtune/datasets/_hf_iterable.py [hunk 29]
```diff
@@ -0,0 +1,295 @@
...
+            epoch_iterator = iter(self._ds)
+            samples_yielded = 0
+
+            try:
+                for sample in epoch_iterator:
+                    # NOTE: We apply transforms here instead of using .map() to work around
+                    # HuggingFace datasets bug where .map() causes incorrect checkpoint resumption.
+                    # See: https://github.com/huggingface/datasets/issues/7630
+                    # This ensures transforms are applied fresh on each sample during iteration.
+                    sample = self._apply_transforms(sample)
```
divyanshk, line 240: Applying transformations inside the dataset before returning every sample would mean no possibility of parallelizing them (either within every dataset or across datasets). Is that expected?

divyanshk: Synced offline. I had the wrong impression. We shard the dataset so if a user turned on num_workers>0 it would lead to multiple processes all reading the same dataset but different shards of it; and so apply transformations in each process.

torchtune/datasets/_hf_iterable.py
```diff
@@ -0,0 +1,295 @@
...
+                    raise ValueError(
+                        f"Number of shards ({num_shards}) is greater than the dataset size ({dataset_size})."
+                        f"Please decrease one of {num_shards_per_rank=} or {num_dataloader_workers=} or {world_size=}."
+                    )
+
+            ds = ds.to_iterable_dataset(num_shards=num_shards)
+
+        # Shuffle the dataset
+        if self._shuffle_buffer_size and self._shuffle_buffer_size > 0:
+            ds = ds.shuffle(seed=self._seed, buffer_size=self._shuffle_buffer_size)
```
divyanshk, line 205: Would it be better if we shuffled before sharding ?

felipemello1: i believe the sharding is happening when we call .to_iterable_dataset. If we do after split_by_node, then i guess the shuffle would only happen inside of the node, and not across nodes.

e.g.
shuffling before:
[0,1,2,3,4,5] -> [4,1,0,5,3,2]

shuffling after
[0,1], [2,3], [4,5] -> [1,0], [2,3], [5,4]

divyanshk: Right. So we should shuffle first, then shard with .to_iterable_dataset call, then split_by_node.

felipemello1: hmmm, why not this?
1. .to_iterable_dataset(num_shards)
2. shuffle
3. split_by_node

sharding happens at 1. Shouldnt we shuffle after sharding?

felipemello1: btw, thanks for helping me double check this

divyanshk: We can shuffle after sharding, but then the shuffle will only be within the shard (unless HF's `ds.shuffle` is doing something non-obvious here). The example you had above is clearly depicting that. Smaller the sample size, less optimal the shuffle. But obv this is something that can be tweaked based on perf.

felipemello1: > We can shuffle after sharding, but then the shuffle will only be within the shard (

oh, i see. I guess the shuffle will still be across all shards, because the shards are not assigned to any rank yet. Just need to make sure that every rank uses the same seed. The issue is if we shuffle after split_by_node. But i need to double check that in their docs/forum. Last time i saw that was a few weeks ago.

PR #2822, opened 2025-06-12
Fix #2809, modify attention

krammnic:
#### Context
What is the purpose of this PR? Is it to
- [ ] add a new feature
- [X] fix a bug
- [ ] update tests and/or documentation
- [ ] other (please add here)

Please link to any issues this PR addresses.

#### Changelog
What are the changes made in this PR?
* #2809

#### Test plan
Please make sure to do each of the following if applicable to your PR. If you're unsure about any one of these just ask and we will happily help. We also have a [contributing page](https://github.com/pytorch/torchtune/blob/main/CONTRIBUTING.md) for some guidance on contributing.

- [X] run pre-commit hooks and linters (make sure you've first installed via `pre-commit install`)
- [ ] add [unit tests](https://github.com/pytorch/torchtune/tree/main/tests/torchtune) for any new functionality
- [ ] update [docstrings](https://github.com/pytorch/torchtune/tree/main/docs/source) for any new or updated methods or classes
- [ ] run unit tests via `pytest tests`
- [ ] run recipe tests via `pytest tests -m integration_test`
- [ ] manually run any new or modified recipes with sufficient proof of correctness
- [ ] include relevant commands and any other artifacts in this summary (pastes of loss curves, eval results, etc.)

#### UX
If your function changed a public API, please add a dummy example of what the user experience will look like when calling it.
Here is a [docstring example](https://github.com/pytorch/torchtune/blob/6a7951f1cdd0b56a9746ef5935106989415f50e3/torchtune/modules/vision_transformer.py#L285)
and a [tutorial example](https://pytorch.org/torchtune/main/tutorials/qat_finetune.html#applying-qat-to-llama3-models)

- [ ] I did not change any public API
- [ ] I have added an example to docs or docstrings

PR #2860, opened 2025-06-30
[fix] var name change

felipemello1:
#### Context
- [ ] add a new feature
- [x] fix a bug
- [ ] other (please add here)

in generate this var name was changed: https://github.com/pytorch/torchtune/blob/3d735916bd9efca600f79fe8d77a757c1160279a/torchtune/generation/_generation.py#L209

without this change, i get the error "generate() got an unexpected keyword argument 'custom_generate_next_token'""

tested with llama2
```
tune run generate --config generation

INFO:torchtune.utils._logging:Tell me a joke. And I mean a real one, not some punchline that doesn’t make any sense.
If you’re not a comedian, it’s going to be hard to come up with something that’s funny. But you can still try.
I’ve heard a lot of jokes that don’t work. They’re either too long, or they’re too short. Or they just don’t tell a joke.
```

### DIFF
--- a/recipes/generate.py
+++ b/recipes/generate.py
@@ -126,7 +126,7 @@ class InferenceRecipe:
         )
         prompt = torch.tensor(tokens, dtype=torch.int, device=self._device)
 
-        custom_generate_next_token = None
+        compiled_generate_next_token = None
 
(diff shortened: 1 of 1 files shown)


krammnic:
Included this in #2822 oops

Issue #2857, opened 2025-06-30
`KeyError: 'lm_head.weight'` when saving checkpoint with Qwen3-*B-Base (Instruct) model


iqiancheng:
Hi,

I'm encountering a `KeyError: 'lm_head.weight'` when fine-tuning the Qwen3-\*B-Base (Instruct) model using `torchtune`. The error occurs during checkpoint saving:

```
INFO:torchtune.utils._logging:Saving checkpoint. This may take some time. Retrieving full model state dict...
INFO:torchtune.utils._logging:Getting full model state dict took 2.71 secs
[rank0]: Traceback (most recent call last):
[rank0]:   File "/path/to/torchtune/recipes/full_finetune_distributed.py", line 1098, in <module>
[rank0]:     sys.exit(recipe_main())
[rank0]:              ^^^^^^^^^^^^^
[rank0]:   File "/path/to/torchtune/torchtune/config/_parse.py", line 99, in wrapper
[rank0]:     sys.exit(recipe_main(conf))
[rank0]:              ^^^^^^^^^^^^^^^^^
[rank0]:   File "/path/to/torchtune/recipes/full_finetune_distributed.py", line 1093, in recipe_main
[rank0]:     recipe.train()
[rank0]:   File "/path/to/torchtune/recipes/full_finetune_distributed.py", line 1056, in train
[rank0]:     self._checkpoint_client.save_checkpoint(
[rank0]:   File "/path/to/torchtune/torchtune/training/checkpointing/_checkpoint_client.py", line 369, in save_checkpoint
[rank0]:     self._save_checkpoint_sync(
[rank0]:   File "/path/to/torchtune/torchtune/training/checkpointing/_checkpoint_client.py", line 330, in _save_checkpoint_sync
[rank0]:     _save_checkpoint_helper()
[rank0]:   File "/path/to/torchtune/torchtune/training/checkpointing/_checkpoint_client.py", line 314, in _save_checkpoint_helper
[rank0]:     self._get_checkpointer().save_checkpoint(
[rank0]:   File "/path/to/torchtune/torchtune/training/checkpointing/_checkpointer.py", line 844, in save_checkpoint
[rank0]:     cpt_idx = self._weight_map[key]
[rank0]:               ~~~~~~~~~~~~~~~~^^^^^
[rank0]: KeyError: 'lm_head.weight'
[...]

PR #2861, opened 2025-07-01
Fix KeyError when saving checkpoint for Qwen3-*B-Base (Instruct) models

pengyanai:
**Issue:**
Fixes [[#2857](https://github.com/pytorch/torchtune/issues/2857)] — `KeyError: 'lm_head.weight'` occurs during checkpoint saving when fine-tuning Qwen3-\*B-Base (Instruct) models, which lack the `lm_head.weight` entry in the original weight map.

**Cause:**
Unlike Qwen3-\*B models, the \*Base versions (e.g., `Qwen3-7B-Base`) do not include `lm_head.weight` in their safetensors index, leading to a missing key error when saving.

**Fix:**
In `_checkpointer.py`, we now detect if `lm_head.weight` is missing from `_weight_map`.

* If it's `lm_head.weight` and `model.embed_tokens.weight` exists, we tie them to the same shard index.
* Otherwise, we fall back to assigning it to `"0001"`.

**Why this matters:**
This ensures compatibility with Base models that use tied embeddings or omit `lm_head.weight` in the original checkpoint, allowing training and checkpointing to proceed without failure.

### DIFF
--- a/torchtune/training/checkpointing/_checkpointer.py
+++ b/torchtune/training/checkpointing/_checkpointer.py
@@ -833,7 +833,20 @@ class FullModelHFCheckpointer(_CheckpointerInterface):
                 split_state_dicts: dict[str, dict[str, torch.Tensor]] = {}
                 total_size = 0
                 for key, weight in state_dict[training.MODEL_KEY].items():
-                    cpt_idx = self._weight_map[key]
+                    # Handle cases where a key (like lm_head.weight for tied embeddings) 
+                    # was not in the original checkpoint but is added during conversion
+                    if key not in self._weight_map:
+                        # For tied embeddings, assign lm_head.weight to the same shard as model.embed_tokens.weight
+                        if key == "lm_head.weight" and "model.embed_tokens.weight" in self._weight_map:
+                            cpt_idx = self._weight_map["model.embed_tokens.weight"]
+                            # Also update the weight map for consistency
+                            self._weight_map[key] = cpt_idx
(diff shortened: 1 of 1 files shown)


felipemello1:
hey @pengyanai , thanks for the PR. Before i review it, can you double check if this flag is enough: https://github.com/pytorch/torchtune/blob/38edb21f4926c4b40b93d9e8b4080f05ddf2da1a/torchtune/models/qwen3/_model_builders.py#L43

If True, then it would create a new lm_head. It will just use the embeddings weight.

If this flag is enough, maybe instead of adding a potentially misleading behavior to the ckpt, we could add some sort of ValueError with instructions to change the flag if this is the desired behavior.

pengyanai:
@felipemello1 Thanks for the quick response !

I double-checked `torchtune.models.qwen3.qwen3_4b_instruct` — the `tie_word_embeddings` flag is indeed always set to `True` by default.

After digging further, I also reviewed the `4B_full.yaml` recipe and noticed the following config under `checkpointer`:

```yaml
# NOTE: this is necessary because Qwen3 4B is missing the lm_head.weight parameter on HF
# Other Qwen3 models contain two copies of the tied param, but Qwen3 4B does not
model_type: QWEN2
```

I had overlooked this note and had changed `model_type` from `QWEN2` to `QWEN3`, which led to the `KeyError`. Hopefully, sharing this helps others who might run into the same issue.

Appreciate the guidance!

PR #2879, opened 2025-07-14
Update Context Parallel Integration

SKRohit:
#### Context
- [x] add a new feature
- [ ] fix a bug
- [ ] other (please add here)

#### What are the changes made in this PR?
* This PR updates Context Parallelism (CP) support to TorchTune, by adding fsdp sharding support to both fsdp and cp dimension and avoids sharding rope cache buffers. 

#### Context Parallelism Changes

- Applied FSDP shard on `cp + dp_shard` dimension not just `dp_shard`
- Added option to pass rotate method for CP
- Removed sharding of rope_cache buffers as input sequence length must always be same with `model.max_seq_length` to work it correctly. With current model initialization API for some models providing `max_seq_len` is not possible from recipes.
- Updated `_collate.py` to add `input_pos` in the batch in order to apply correct rope embedding for cp

#### Test plan

#### UX

### DIFF
--- a/recipes/full_finetune_distributed.py
+++ b/recipes/full_finetune_distributed.py
@@ -163,6 +163,7 @@ class FullFinetuneRecipeDistributed(FTRecipeInterface):
             # DTensor does not support grouped_mm yet
             moe_utils.use_grouped_mm = False
         self.cp_degree = cfg.get("context_parallel_dim", 1)
+        self.context_parallel_rotate_method = cfg.get("context_parallel_rotate_method", None)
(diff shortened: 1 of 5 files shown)


SKRohit:
@ebsmothers please take  a look at the changes

SKRohit:
@ebsmothers @joecummings @pbontrager  Could you please review these changes and provide feedback?

ebsmothers (approved):
Thank you @SKRohit for finding and fixing these bugs! The RoPE buffers issue was quite tricky and your fix is really nice.

PR #2887, opened 2025-07-17
add smollm to torchtune

zhangtemplar:
Summary: add smolLM2 family to torchtune. SmolLM2 is a family of compact language models available in three size: 135M, 360M, and 1.7B parameters. They architecture is the same as LLaMA 3.

Differential Revision: D78495904


krammnic:
Hmm, I'm not sure whether we want to see some models that are based on existing ones, but personally I'm not against it. But we need to update `README` models section. After that, it will be LGTM

zhangtemplar:
could some one approve it? Thanks!

krammnic (approved):

PR #2897, opened 2025-08-01
actually patch tp plan

ebsmothers:
Follow-up to #2896: we need to actually update how we patch the TP plan for multimodal models as well


torchtune/modules/loss/cross_entropy_loss.py
```diff
@@ -72,12 +72,18 @@ def set_model_output(self, model: nn.Module) -> None:
     def patch_tp_plan(self, tp_plan) -> dict:
         if "output" not in tp_plan and "decoder.output" not in tp_plan:
             raise KeyError("`tp_plan` requires `output` key")
-
-        tp_plan["output"] = ColwiseParallel(
-            input_layouts=Shard(1),
-            output_layouts=Shard(-1),
-            use_local_output=False,
-        )
+        if "output" in tp_plan:
```
pbontrager: nit: could we combine these back into just
```
key = "output" if "output" in tp_plan else "decoder.output"
tp_plan[key] = ...
```

pbontrager (approved):

PR #2900, opened 2025-08-04
Add validation for LinearCrossEntropyLoss with custom_sharded_layers

### LINKED ISSUE #2856: Bug Report: DTensor/torch.Tensor Mixed Type Error in Llama4 LoRA Fine-tuning
I follow this cookbook in order to finetune LLaMa4 (https://github.com/meta-llama/llama-cookbook/blob/main/getting-started/finetuning/finetune_llama4.md)
### Bug Description
Getting RuntimeError: aten.mm.default: got mixed torch.Tensor and DTensor, need to convert all torch.Tensor to DTensor before calling distributed operators! when running distributed LoRA fine-tuning on Llama4-Scout-17B-16E model.

### Error Message
RuntimeError: aten.mm.default: got mixed torch.Tensor and DTensor, need to convert all torch.Tensor to DTensor before calling distributed operators!
### To Reproduce
**Command**:
tune run --nproc_per_node 8 lora_finetune_distributed --config llama4/scout_17B_16E_lora

**Configuration (scout_17B_16E_lora.yaml)**

```yaml
# Config for multi-device full finetuning in full_finetune_distributed.py
# using a Llama4 17Bx16E MoE model
#
# This config assumes that you've run the following command before launching:
#   tune download meta-llama/Llama-4-Scout-17B-16E-Instruct
#
# To launch on 8 devices, run the following command from root:
#   tune run --nproc_per_node 8 lora_finetune_distributed --config llama4/scout_17B_16E_lora
#
# You can add specific overrides through the command line. For example, to use a larger bsz:
#   tune run --nproc_per_node 8 lora_finetune_distributed --config llama4/scout_17B_16E_lora batch_size=8
#
# This config was only tested on 8xA100 machine.

output_dir: /tmp/torchtune/llama4_17Bx16E/lora

# Modeling Arguments
model:
  _component_: torchtune.models.llama4.lora_llama4_scout_17b_16e
  decoder_trainable: "lora"
  encoder_trainable: "frozen"
  fusion_trainable: "lora"
  lora_attn_modules: ['q_proj', 'v_proj', 'output_proj']
  apply_lora_to_mlp: True
  apply_lora_to_output: False
  lora_rank: 16  # higher increases accuracy and memory
  lora_alpha: 32  # usually alpha=2*rank
  lora_dropout: 0.0

tokenizer:
  _component_: torchtune.models.llama4.llama4_transform
  path: /tmp/Llama-4-Scout-17B-16E-Instruct/tokenizer.model
  max_seq_len: null
  max_num_tiles: 16

checkpointer:
  checkpoint_dir: /tmp/Llama-4-Scout-17B-16E-Instruct
  checkpoint_files:
    max_filename: "00050"
  recipe_checkpoint: null
  output_dir: ${output_dir}
  model_type: LLAMA4
save_adapter_weights_only: True # use this for faster checkpoint save
resume_from_checkpoint: False

# Dataset and Sampler
dataset:
  packed: False  # True increases speed
  split: train[:95%]
seed: null
shuffle: True

# Validation
run_val_every_n_steps: null  # Change to an integer to enable validation every N steps
dataset_val:
  split: train[95%:]
batch_size_val: ${batch_size}

# Training arguments
epochs: 1
batch_size: 2
max_steps_per_epoch: null
gradient_accumulation_steps: 1 # Use to increase effective batch size
optimizer:
  lr: 2e-5
  fused: False
optimizer_in_bwd: False
lr_scheduler:
  num_warmup_steps: 100
loss:
  _component_: torchtune.modules.loss.LinearCrossEntropyLoss
clip_grad_norm: null

# cuda, cpu, rocm, xpu...
device: cuda

# Memory management / performance
enable_activation_offloading: False
custom_sharded_layers: ['tok_embeddings']
fsdp_cpu_offload: False
compile: False # torch.compile, set to true for perf/memory improvement

# Reduced precision
dtype: bf16

# Log metrics during training
metric_logger:
  log_dir: ${output_dir}/logs
log_every_n_steps: 1
log_peak_memory_stats: True
log_level: INFO  # DEBUG, WARN, etc.

# Useful for understanding how to optimize memory and performance
profiler:
  _component_: torchtune.training.setup_torch_profiler
  enabled: False
```

**Environment Information**
pytorch
PyTorch: 2.8.0.dev20250625+cu126
cdua version
CUDA: 12.6
CUDA devices: 8
python --version
Python 3.10.18
OS system
Linux ip-10-207-79-120 6.8.0-1030-aws #32-Ubuntu SMP Wed May 28 19:48:56 UTC 2025 x86_64 x86_64 x86_64 GNU/Linux
nvidia-smi 
```bash
py310) ubuntu@ip-10-207-79-120:~/torchtune/recipes/configs/llama4$ nvidia-smi
Sun Jun 29 10:56:45 2025
+-----------------------------------------------------------------------------------------+
| NVIDIA-SMI 570.133.20             Driver Version: 570.133.20     CUDA Version: 12.8     |
|-----------------------------------------+------------------------+----------------------+
| GPU  Name                 Persistence-M | Bus-Id          Disp.A | Volatile Uncorr. ECC |
| Fan  Temp   Perf          Pwr:Usage/Cap |           Memory-Usage | GPU-Util  Compute M. |
|                                         |                        |               MIG M. |
|=========================================+========================+======================|
|   0  NVIDIA A100-SXM4-80GB          On  |   00000000:10:1C.0 Off |                    0 |
| N/A   49C    P0             69W /  400W |       0MiB /  81920MiB |      0%      Default |
|                                         |                        |             Disabled |
+-----------------------------------------+------------------------+----------------------+
|   1  NVIDIA A100-SXM4-80GB          On  |   00000000:10:1D.0 Off |                    0 |
| N/A   46C    P0             68W /  400W |       0MiB /  81920MiB |      0%      Default |
|                                         |                        |             Disabled |
+-----------------------------------------+------------------------+----------------------+
|   2  NVIDIA A100-SXM4-80GB          On  |   00000000:20:1C.0 Off |                    0 |
| N/A   49C    P0             72W /  400W |       0MiB /  81920MiB |      0%      Default |
|                                         |                        |             Disabled |
+-----------------------------------------+------------------------+----------------------+
|   3  NVIDIA A100-SXM4-80GB          On  |   00000000:20:1D.0 Off |                    0 |
| N/A   43C    P0             65W /  400W |       0MiB /  81920MiB |      0%      Default |
|                                         |                        |             Disabled |
+-----------------------------------------+------------------------+----------------------+
|   4  NVIDIA A100-SXM4-80GB          On  |   00000000:90:1C.0 Off |                    0 |
| N/A   47C    P0             66W /  400W |       0MiB /  81920MiB |      0%      Default |
|                                         |                        |             Disabled |
+-----------------------------------------+------------------------+----------------------+
|   5  NVIDIA A100-SXM4-80GB          On  |   00000000:90:1D.0 Off |                    0 |
| N/A   44C    P0             70W /  400W |       0MiB /  81920MiB |      0%      Default |
|                                         |                        |             Disabled |
+-----------------------------------------+------------------------+----------------------+
|   6  NVIDIA A100-SXM4-80GB          On  |   00000000:A0:1C.0 Off |                    0 |
| N/A   45C    P0             69W /  400W |       0MiB /  81920MiB |      0%      Default |
|                                         |                        |             Disabled |
+-----------------------------------------+------------------------+----------------------+
|   7  NVIDIA A100-SXM4-80GB          On  |   00000000:A0:1D.0 Off |                    0 |
| N/A   42C    P0             68W /  400W |       0MiB /  81920MiB |      0%      Default |
|                                         |                        |             Disabled |
+-----------------------------------------+------------------------+----------------------+

+-----------------------------------------------------------------------------------------+
| Processes:                                                                              |
|  GPU   GI   CI              PID   Type   Process name                        GPU Memory |
|        ID   ID                                                               Usage      |
|=========================================================================================|
|  No running processes found                                                             |
+-----------------------------------------------------------------------------------------+
```

jscaldwell55:
## Summary
Fixes #2856 - DTensor/torch.Tensor mixed type error in Llama4 LoRA fine-tuning

## Problem
When using LoRA fine-tuning with `LinearCrossEntropyLoss` and `custom_sharded_layers`, users encounter a tensor type mismatch error:
RuntimeError: aten.mm.default: got mixed torch.Tensor and DTensor, need to convert all torch.Tensor to DTensor before calling distributed operators!

This happens because:
- `LinearCrossEntropyLoss` uses `model.output` for the final projection
- LoRA configs typically set `custom_sharded_layers = ['tok_embeddings']` without including `'output'`
- FSDP wraps only the layers listed in `custom_sharded_layers` as DTensors
- This creates a mismatch when computing loss (DTensor hidden states × regular Tensor output weights)

## Solution
Added validation that checks if `LinearCrossEntropyLoss` is used with `custom_sharded_layers` and ensures `'output'` is included in the list. This provides a clear, actionable error message at setup time rather than a cryptic error during training.

## Implementation
- Created shared validation module `recipes/validation.py` to avoid code duplication
- Added validation to both `full_finetune_distributed.py` and `lora_finetune_distributed.py` recipes
- Validation is called in `_setup_model` before FSDP wrapping occurs
- Added comprehensive unit tests covering various edge cases

## Testing
- Unit tests added in `tests/recipes/test_validation.py`
- Tests cover: missing output, correct config, None/empty layers, disabled parallelism, non-LinearCrossEntropyLoss
- No changes to existing functionality - only adds validation

## Example Error Message
When misconfigured, users will now see:
ValueError: When using LinearCrossEntropyLoss with custom_sharded_layers, 'output' must be included to ensure tensor compatibility. Example: custom_sharded_layers = ['tok_embeddings', 'output'].

This guides users to the correct configuration immediately.

### DIFF
--- a/recipes/full_finetune_distributed.py
+++ b/recipes/full_finetune_distributed.py
@@ -29,6 +29,7 @@ from torchtune.data import padded_collate_packed
 from torchtune.datasets import ConcatDataset
 from torchtune.modules.embedding_utils import resize_token_embeddings
 from torchtune.modules.loss import SFTLoss
+from recipes.validation import validate_custom_sharding_config
 from torchtune.modules.moe import utils as moe_utils
 from torchtune.recipe_interfaces import FTRecipeInterface
 from torchtune.training import (
@@ -669,6 +670,12 @@ class FullFinetuneRecipeDistributed(FTRecipeInterface):
                 model, auto_wrap_policy={modules.TransformerSelfAttentionLayer}
             )
 
+	# Validate custom_sharded_layers configuration
+        validate_custom_sharding_config(
+            self._loss_fn,
+            custom_sharded_layers,
+            parallelism_enabled=self.parallel_dims.dp_shard_enabled or self.parallel_dims.cp_enabled
+        )
         # Apply Fully Sharded Data Parallelism to the model
         if self.parallel_dims.dp_shard_enabled or self.parallel_dims.cp_enabled:
             fsdp_shard_conditions = [
--- a/recipes/lora_finetune_distributed.py
+++ b/recipes/lora_finetune_distributed.py
@@ -26,6 +26,7 @@ from torchtune.config._utils import _get_component_from_path
 from torchtune.data import padded_collate_packed
 from torchtune.datasets import ConcatDataset
 from torchtune.modules.loss import SFTLoss
+from recipes.validation import validate_custom_sharding_config
 from torchtune.modules.peft import (
     AdapterModule,
     get_adapter_params,
@@ -530,6 +531,14 @@ class LoRAFinetuneRecipeDistributed(FTRecipeInterface):
                 model, auto_wrap_policy={modules.TransformerSelfAttentionLayer}
             )
 
+	# Validate custom_sharded_layers configuration
+        validate_custom_sharding_config(
+            self._loss_fn,
+            custom_sharded_layers,
+            parallelism_enabled=self.parallel_dims.dp_shard_enabled or self.parallel_dims.cp_enabled
+        
+	)
+	
         # Apply Fully Sharded Data Parallelism to the model
         if self.parallel_dims.dp_shard_enabled or self.parallel_dims.cp_enabled:
             # For FSDP sharding
--- a/recipes/validation.py
+++ b/recipes/validation.py
@@ -1,0 +1,57 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+"""Shared validation utilities for recipes."""
+
+from typing import Optional, Set
+from torchtune.modules.loss import LinearCrossEntropyLoss
+
+
+def validate_custom_sharding_config(
+    loss_fn,
+    custom_sharded_layers: Optional[list[str]],
+    required_layer: str = "output",
+    parallelism_enabled: Optional[bool] = None,
+    available_layers: Optional[Set[str]] = None,
+) -> None:
+    """
+    Validates custom_sharded_layers configuration for specific loss functions.
+    
+    Args:
+        loss_fn: The loss function instance
+        custom_sharded_layers: List of layer names to shard, or None
+        required_layer: The layer name that must be included (default: "output")
+        parallelism_enabled: If False, skip validation (default: None)
+        available_layers: Optional set of valid layer names for typo checking
+        
+    Raises:
+        ValueError: If validation fails
+    """
+    # Skip when nothing to validate
+    if not custom_sharded_layers:
+        return
+        
+    # Skip validation if parallelism is explicitly disabled
+    if parallelism_enabled is False:
+        return
+    
+    # Only enforce when the loss needs the output projection
+    needs_output_proj = isinstance(loss_fn, LinearCrossEntropyLoss)
+    
+    if needs_output_proj and required_layer not in custom_sharded_layers:
+        raise ValueError(
+            f"When using {type(loss_fn).__name__} with custom_sharded_layers, "
+            f"'{required_layer}' must be included to ensure tensor compatibility. "
+            f"Example: custom_sharded_layers = ['tok_embeddings', '{required_layer}']."
+        )
+    
+    # Optional: catch typos early
+    if available_layers is not None:
+        unknown = set(custom_sharded_layers) - set(available_layers)
+        if unknown:
+            raise ValueError(
+                f"Unknown layer(s) in custom_sharded_layers: {sorted(unknown)}"
+            )
--- a/tests/recipes/test_validation.py
+++ b/tests/recipes/test_validation.py
@@ -1,0 +1,64 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+import pytest
+from unittest.mock import Mock
+from torchtune.modules.loss import LinearCrossEntropyLoss
+from recipes.validation import validate_custom_sharding_config
+
+
+class TestValidateCustomShardingConfig:
+    """Unit tests for validate_custom_sharding_config"""
+    
+    def test_missing_output_raises_error(self):
+        """Test that validation raises error when output is missing"""
+        loss_fn = LinearCrossEntropyLoss()
+        custom_sharded_layers = ['tok_embeddings']
+        
+        with pytest.raises(ValueError, match="'output' must be included"):
+            validate_custom_sharding_config(loss_fn, custom_sharded_layers)
+    
+    def test_with_output_passes(self):
+        """Test that validation passes when output is included"""
+        loss_fn = LinearCrossEntropyLoss()
+        custom_sharded_layers = ['tok_embeddings', 'output']
+        
+        # Should not raise
+        validate_custom_sharding_config(loss_fn, custom_sharded_layers)
+    
+    def test_none_layers_passes(self):
+        """Test that validation passes when custom_sharded_layers is None"""
+        loss_fn = LinearCrossEntropyLoss()
+        
+        # Should not raise
+        validate_custom_sharding_config(loss_fn, None)
+    
+    def test_empty_layers_passes(self):
+        """Test that validation passes when custom_sharded_layers is empty"""
+        loss_fn = LinearCrossEntropyLoss()
+        
+        # Should not raise
+        validate_custom_sharding_config(loss_fn, [])
+    
+    def test_parallelism_disabled_skips_validation(self):
+        """Test that validation is skipped when parallelism is disabled"""
+        loss_fn = LinearCrossEntropyLoss()
+        custom_sharded_layers = ['tok_embeddings']  # Missing output
+        
+        # Should not raise because parallelism_enabled=False
+        validate_custom_sharding_config(
+            loss_fn, 
+            custom_sharded_layers,
+            parallelism_enabled=False
+        )
+    
+    def test_non_linear_ce_loss_passes(self):
+        """Test that non-LinearCrossEntropyLoss doesn't require output"""
+        loss_fn = Mock()  # Some other loss function
+        custom_sharded_layers = ['tok_embeddings']  # Missing output
+        
+        # Should not raise
+        validate_custom_sharding_config(loss_fn, custom_sharded_layers)


jscaldwell55:
@nathan-az Hey Nathan, apologies for confusion. I had meant to revise current PR but created a new branch without thinking. Wanted to give you a little context for this. After looking into it, here's what I found:

- The full fine-tuning config correctly includes `custom_sharded_layers: ['tok_embeddings', 'output']` 
- But, LoRA fine-tuning configs typically only include `['tok_embeddings']`
- This makes sense for LoRA's use case, but causes issues with `LinearCrossEntropyLoss` which needs the output layer for its projection

Key changes from the previous attempt:

- No modifications to `cross_entropy_loss.py` 
- Validation logic in a shared module to avoid duplication
- Error message that guides users to the fix
- Comprehensive unit tests
- Validation placed in `_setup_model` where both variables are in scope

I guess my "fix" here is just making sure users get a descriptive error message rather than runtime error during training. Particularly for LoRA users who might not realize they need to include 'output' when using `LinearCrossEntropyLoss`.

Thanks again for your review. Your notes were super helpful, and I learned some cool stuff working through this :) 
If you think further revision is needed just let me know.

nathan-az:
Hey @jscaldwell55 - thanks again for your work here! 2 request if you have the time.
Could you confirm you're using the latest (or near latest) nightlies? There were some [changes made to the model root resharding logic during FSDP](https://github.com/pytorch/torchtune/pull/2800) in response to some recent PyTorch changes, and this error is reminiscent of those.

In addition, could you provide an updated traceback/error logs showing what happens when you run your config on `main`, either here or in the original issue? The reason I ask is that a lot of the loss logic [was updated after that bug was reported](https://github.com/pytorch/torchtune/pull/2782), where we stopped extracting the `weight` directly, so the old logs don't help much now. It would be good to see where the error actually reports now.

It's not immediately obvious to me why having `tok_embeddings` sharded separately (i.e. in `custom_sharded_layers`) would necessitate also including `output`, but having updated logs may make it clearer.

jscaldwell55:
@nathan-az Appreciate the continued review and your feedback! This totally makes sense, I may have missed this and need to check. I'll verify against latest nightlies and do the following:
1) update to latest torchtune nightly build
2) test with a LoRA config using `custom_sharded_layers =['tok_embeddings']` (without 'output')
3) provide updated error logs if issue persists

As you said, may no longer be an actual requirement if resharding logic changes already resolved

Will let you know what I find!

jscaldwell55:
@nathan-az Got to work on this some this morning and completed the testing you requested. Here are findings:

## Test Environment
- **PyTorch:** 2.6.0+cu124 (latest available in Colab)
- **Torchtune:** 0.6.1
- **Model Used:** meta-llama/Llama-2-7b-hf with LoRA

## Test Results

### Single-GPU Testing
✅ Training runs successfully without DTensor errors in single-GPU mode with the problematic configuration:

```yaml
loss:
  _component_: torchtune.modules.loss.LinearCrossEntropyLoss
custom_sharded_layers: ['tok_embeddings']  # Without 'output'
```

However, single-GPU doesn't trigger the FSDP DTensor wrapping that causes the issue.


**Multi-GPU Evidence**
The original issue #2856 from June 29, 2025 shows the error still occurring with:

PyTorch 2.8.0.dev20250625
8xA100 GPUs
Same configuration causing the exact DTensor/Tensor mismatch

**Key Finding:** The error explicitly occurs at cross_entropy_loss.py:71:
pythonlogits = F.linear(hidden_chunk, weight)  # DTensor × regular Tensor = Error


**Analysis**
The issue manifests only in multi-GPU FSDP scenarios where:

- tok_embeddings in custom_sharded_layers → wrapped as DTensor
- output NOT in custom_sharded_layers → remains regular Tensor
- LinearCrossEntropyLoss uses output.weight for projection
- Mixed tensor types cause the runtime error

**Note on Multi-GPU Testing**


I wasn't able to generate updated error logs from multi-GPU testing due to resource constraints (only have access to single GPU via Colab). The single-GPU test doesn't trigger the FSDP DTensor wrapping, so it runs without errors.


The most recent multi-GPU error logs I could find are from issue #2856, which show the error occurring at the same location in the code (cross_entropy_loss.py:71). While these logs are from PyTorch 2.8.0.dev20250625, the error mechanism appears unchanged based on:

The code structure in current main still has the same F.linear call
The FSDP wrapping logic with custom_sharded_layers works the same way
The loss function still extracts model.output.weight for projection


For next steps, would you like me to:

Try to find a multi-GPU environment for definitive testing?
Or proceed with the validation in this PR as a defensive measure given the evidence?

Please let me know what you're thinking re best path forward. Really appreciate your patience and feedback as I work through this :)

nathan-az:
Hey mate, thanks for checking in on this. I have access to hardware again so ran some tests.

I ran two configs - the LLaMA 3.1 8B lora config, and the LLaMA 3.2 1B lora config. Both with `custom_sharded_layers: ['tok_embeddings']`

I was not able to replicate the issue with 8B, but I was with 1B. I believe the key difference in architecture between these two is that the 1B model uses tied weights between the token embeddings and final output projection. I think this is likely why we see issues when one is sharded and the other is not.

If this is correct, I think adding a layer of validation would be useful, but that it should factor this in.

jscaldwell55:
@nathan-az Awesome, thanks for running those tests!

This definitely makes sense to me - models with tied embeddings (where `tok_embeddings.weight` is the same tensor as `output.weight`) would have this issue when only one is in `custom_sharded_layers`. The 1B model uses weight tying for efficiency, while the 8B model has separate weights.

I'll update the validation to be more precise:

1. **Check if the model uses tied weights** (when `model.tok_embeddings.weight is model.output.weight`)
2. **Only enforce the validation for tied-weight models** using `LinearCrossEntropyLoss`
3. **Update the error message** to explain the tied weights issue

This way we can avoid unnecessary restrictions on models that don't have this architectural constraint.

Quick question: Should I also check for the reverse case (where `output` is in `custom_sharded_layers` but `tok_embeddings` isn't)? Or would FSDP handle that differently?

nathan-az:
No worries! So - one thing you reported that I haven't been able to replicate is that custom sharding of **_both_** also throws the type mismatch error when the layers are tied for me.

I'm not an expert in how FSDP (and FSDP2) work, but I see that the `TiedLinear` class's forward directly accesses the `weight` attribute from the underlying embedding via `return self.linear(x, self.tied_module.weight)`. My best theory is that since the weight is being accessed directly, the FSDP hook to unshard is not running (i.e. by the time the matmul is done, I think the weights should just be a `Tensor`, not `DTensor`).

@ebsmothers @joecummings sorry for the direct mention - if anything is immediately obvious to either of you (e.g. does my theory have merit, is there an obvious + is there an obvious/clean fix), that would be great. No pressure to look deeper, I just don't have that much experience with how FSDP does sharding.

IMO - if sharding tied weights doesn't currently work, it's not a huge deal and we can just add validation ot confirm. By default, the transformer layers are still sharded when FSDP is used (this can be confirmed in the sharding code), and because these weights are tied, they consume half as much memory as they would by default.

jscaldwell55:
@nathan-az I'll hold off on implementing anything until we hear from the other reviewers. Happy to go with whatever makes the most sense to y'all. 

Really appreciate you digging into this with me; I learn so much getting into these distributed training edge cases.

PR #2901, opened 2025-08-05
properly extend the collective timeout #2093

pengyanai:
#### Context
- [ x ] fix a bug

to fix : https://github.com/pytorch/torchtune/issues/2093

### DIFF
--- a/recipes/full_dpo_distributed.py
+++ b/recipes/full_dpo_distributed.py
@@ -36,6 +36,7 @@ from torchtune.training.lr_schedulers import get_lr
 from torchtune.utils import get_world_size_and_rank
 from tqdm import tqdm
 
+from datetime import timedelta
 
 class FullDPORecipeDistributed(FTRecipeInterface):
     """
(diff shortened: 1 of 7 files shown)


nathan-az:
I've done a similar thing on my personal branch. I recommend reading a new config option `collective_timeout`, defaulting to the standard `600` with users having the option to increase.

I also added a trivially small collective after the init (summing a random tensor on each rank) because I came across [this comment](https://github.com/pytorch/pytorch/issues/104506#issuecomment-1679762880). I'm not certain that I'm interpreting it correctly or that it's required, but it's trivially cheap. ([Specific issue](https://github.com/pytorch/pytorch/issues/124950))

pengyanai:
@nathan-az Thanks for sharing your approach, Nathan. Adding a configurable `collective_timeout` option makes sense, and I agree that giving users flexibility is useful. Including a small collective operation after init also seems like a good safeguard, even if it's not strictly required. I'll look into the linked issues and comments—appreciate the reference!

PR #2937, opened 2025-11-07
Remove old imports to unblock code cleanup in torchao

andrewor14:
Specifically, this unblocks:
https://github.com/pytorch/ao/pull/3308
https://github.com/pytorch/ao/pull/3146


andrewor14:
@ebsmothers @joecummings wondering if we can still land this PR/diff to unblock torchao BE efforts?

PR #2964, opened 2026-05-22
fix(logging): use RANK env var in log_rank_zero before dist init

### LINKED ISSUE #2700: Log config to output only on rank zero
Currently, we call `config.log_config` in the recipe_main which means this logs to every device.

We really only need this to happen once. See callsite here: https://github.com/pytorch/torchtune/blob/dd6764189d7e6ecc0f59902afb657dd60ee74b13/recipes/full_finetune_distributed.py#L1093

kratos0718:
Fixes #2700

## Summary

`log_rank_zero` fell back to `rank=0` for all processes when
`dist.is_initialized()` was `False` (i.e. before `setup()` calls
`init_process_group`). This meant `config.log_config` — called in
`recipe_main` before the recipe is constructed — printed the full
config on **every rank** in a multi-GPU torchrun launch.

`torchrun` always sets `RANK` in the environment before spawning
workers, so we use it as the fallback rank when the process group is
not yet initialized.

## What changed
- `torchtune/utils/_logging.py`: `log_rank_zero` now reads `os.environ["RANK"]` (default `"0"`) as fallback instead of assuming 0
- `tests/torchtune/utils/test_logging.py`: 3 new test cases covering pre-init RANK=0, RANK=1, and no-RANK (single device)

## Impact
Fixes redundant config logging in all distributed recipes
(`full_finetune_distributed`, `lora_finetune_distributed`,
`knowledge_distillation_distributed`, etc.) with a single change.

### DIFF
--- a/tests/torchtune/utils/test_logging.py
+++ b/tests/torchtune/utils/test_logging.py
@@ -5,6 +5,7 @@
 # LICENSE file in the root directory of this source tree.
 
 import logging
+import os
 from io import StringIO
 from unittest import mock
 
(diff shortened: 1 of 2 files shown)


kratos0718:
̑

kratos0718:
Hi! Following up — the Meta CLA is signed and all checks are passing. Happy to address any feedback or requested changes. Thanks!

PR #2465, opened 2025-03-06
Reference-free DPO losses in torchtune.

krammnic:
#### Context
What is the purpose of this PR? Is it to
- [X] add a new feature
- [ ] fix a bug
- [X] update tests and/or documentation
- [ ] other (please add here)

Please link to any issues this PR addresses.

#### Changelog
What are the changes made in this PR?
* #1223 #2292 #2081 

#### Test plan
Please make sure to do each of the following if applicable to your PR. If you're unsure about any one of these just ask and we will happily help. We also have a [contributing page](https://github.com/pytorch/torchtune/blob/main/CONTRIBUTING.md) for some guidance on contributing.

- [X] run pre-commit hooks and linters (make sure you've first installed via `pre-commit install`)
- [ ] add [unit tests](https://github.com/pytorch/torchtune/tree/main/tests/torchtune) for any new functionality
- [ ] update [docstrings](https://github.com/pytorch/torchtune/tree/main/docs/source) for any new or updated methods or classes
- [ ] run unit tests via `pytest tests`
- [ ] run recipe tests via `pytest tests -m integration_test`
- [ ] manually run any new or modified recipes with sufficient proof of correctness
- [ ] include relevant commands and any other artifacts in this summary (pastes of loss curves, eval results, etc.)

#### UX
If your function changed a public API, please add a dummy example of what the user experience will look like when calling it.
Here is a [docstring example](https://github.com/pytorch/torchtune/blob/6a7951f1cdd0b56a9746ef5935106989415f50e3/torchtune/modules/vision_transformer.py#L285)
and a [tutorial example](https://pytorch.org/torchtune/main/tutorials/qat_finetune.html#applying-qat-to-llama3-models)

- [ ] I did not change any public API
- [ ] I have added an example to docs or docstrings

PR #2966, opened 2026-06-17
Fix NameError in torchtune rlhf/loss/dpo.py from missing imports

wjh70301-meta:
Summary:
`torchtune/rlhf/loss/dpo.py` referenced `dataclass` and `TypeVar` (line `T = TypeVar("T", bound=dataclass)`) without importing them, raising `NameError: name 'TypeVar' is not defined` at import time. This blocked every Mitra DPO unit that imports `torchtune.rlhf.loss.DPOLoss` — including the IG Reels interest-judge DPO work.

Adds the missing `from dataclasses import dataclass` and `from typing import Optional, Tuple, TypeVar`.

___

Differential Revision: D108715300

### DIFF
--- a/torchtune/rlhf/loss/dpo.py
+++ b/torchtune/rlhf/loss/dpo.py
@@ -4,6 +4,9 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
+from dataclasses import dataclass
+from typing import Optional, Tuple, TypeVar
+
 import torch
 import torch.nn as nn
 import torch.nn.functional as F


ErenAta16 (commented):
This is worse than a lint fix and I think the description undersells it. The four names are used at module scope, not inside a function, so the file cannot be imported at all:

```python
# torchtune/rlhf/loss/dpo.py on main, line 14
T = TypeVar("T", bound=dataclass)
```

with the imports being exactly `torch`, `torch.nn`, `torch.nn.functional`, `ChosenRejectedOutputs` and `deprecated`. Neither `TypeVar` nor `dataclass` is among them, and that line runs at import time.

It is not a leaf module either. `torchtune/rlhf/loss/__init__.py` does:

```python
from .dpo import DPOLoss, RSOLoss
```

so `import torchtune.rlhf.loss` raises `NameError: name 'TypeVar' is not defined` before anything else happens, and every consumer of `DPOLoss`, `RSOLoss` or `PPOLoss` goes with it. Worth saying plainly in the PR title or body, because "missing imports" reads like a typing cleanup and reviewers triage it accordingly.

The history suggests how it slipped through. `0a08c2f5` ("refactor: migrate type hints to PEP 585") removed the `typing` imports when it moved annotations to builtin generics, and the file still carries both spellings today, `Tuple[torch.Tensor, ...]` on one method and `tuple[torch.Tensor, ...]` on others. The later `a9c32c09` ("Reference-free DPO losses", #2465) then reintroduced `Optional[T]`, `Tuple[...]` and the `TypeVar` line without bringing the imports back. Nothing to fix there, it just explains the shape.

**One thing I would change while you are in here.** `bound=dataclass` is not a meaningful bound. `dataclass` is the decorator function, not a type:

```python
T = TypeVar("T", bound=dataclass)
T.__bound__        # <function dataclass at 0x...>
isinstance(dataclass, type)   # False
```

It constructs without complaint at runtime, so this will not break anything, but a type checker cannot do anything useful with a function as a bound and the annotation is effectively unbounded. If the intent is "any dataclass instance", a `Protocol` with `__dataclass_fields__` expresses that, and if the intent was just "the concrete output type", dropping the bound entirely is more honest than one that does not check.

Since the whole point of this PR is to make the module importable, and adding `from dataclasses import dataclass` is what makes that bound resolvable, it seems worth settling now rather than importing a symbol solely to satisfy a line that does not do what it looks like it does.

Given the package is currently unimportable, this is worth landing quickly either way.

Issue #2792, opened 2025-06-05
Qwen2/2.5/3 max_seq_len and EOS Masking


haydn-jones:
I've been trying to debug why I can't get my model to consistently stop its output, and it turns out that passing `max_seq_len` changes whether the EOS token is masked or not in SFT:
```python
from torchtune.data import InputOutputToMessages
from torchtune.models.qwen3 import qwen3_tokenizer

msg_fmt = InputOutputToMessages(
    masking_strategy="train_on_assistant",
)
inp = {
    "input": "A",
    "output": "B",
}

tokenizer = qwen3_tokenizer(
    path="./model/Qwen3-0.6B/vocab.json",
    merges_file="./model/Qwen3-0.6B/merges.txt",
)
out = tokenizer(msg_fmt(inp))
print(out["tokens"])
print(out["mask"])

# [151644, 872, 198, 32, 151645, 198, 151644, 77091, 198, 33, 151645]
# [True, True, True, True, True, True, False, False, False, False, False] <-- False

# Same tokenizer but set max_seq_len
tokenizer = qwen3_tokenizer(
    path="./model/Qwen3-0.6B/vocab.json",
    merges_file="./model/Qwen3-0.6B/merges.txt",
    max_seq_len=8192,
)
out = tokenizer(msg_fmt(inp))
print(out["tokens"])
print(out["mask"])

# [151644, 872, 198, 32, 151645, 198, 151644, 77091, 198, 33, 151645]
# [True, True, True, True, True, True, False, False, False, False, True]  <-- Now True
```

It gets flipped here: https://github.com/pytorch/torchtune/blob/c7a92e49935c40da088a9f6ae2b5b7b637f3f3a1/torchtune/data/_utils.py#L60

Because the Qwen2 tokenizer calls truncate with eos_id=True here:
https://github.com/pytorch/torchtune/blob/c7a92e49935c40da088a9f6ae2b5b7b637f3f3a1/torchtune/models/qwen2/_tokenizer.py#L416

This of course means that the SFT transform will use the cross entropy ignore ID as the target for the token before EOS rather than the EOS  ID. I assume this is not intended behavior?

krammnic:
Hey @haydn-jones, I will definitely take a look, but we are planning to deprecate model-specific tokenizers, because we proposed a general one #2723. However, this problem is related to a truncation, so this is a good catch. Thanks for the issue!

PR #2967, opened 2026-06-28
[Fix] Qwen tokenizers mask trained EOS out of loss when max_seq_len is set (#2792)

yushuosun:
## Motivation

When `max_seq_len` is set, the Qwen2 / 2.5 / 3 tokenizers mask the trailing EOS token out of the loss, so the model is never trained to emit EOS and won't reliably stop generating (#2792).

## Root cause

In `tokenize_messages`, after appending EOS its mask is correctly set to `mask[-1]` (the final message's mask — `False`/trained for an assistant turn):

```python
if add_end_tokens:
    tokenized_messages.append(self.eos_id)
    mask.append(mask[-1])
```

But the subsequent `truncate(...)` of the mask hardcodes the forced last element to `True`:

```python
mask = truncate(
    tokens=mask,
    max_seq_len=self.max_seq_len,
    eos_id=True if add_end_tokens else None,   # forces mask[-1] = True even with no truncation
    truncation_type=self.truncation_type,
)
```

`truncate` sets the last element to `eos_id` whenever it is not `None`, so even when no truncation occurs the EOS mask flips `False -> True`, and `mask == True` maps to `CROSS_ENTROPY_IGNORE_IDX`. Net effect: with `max_seq_len` set, EOS is never a training target. Qwen3 subclasses Qwen2.5 and inherits the bug. (Llama3 intentionally always-masks EOS, so its identical line is correct there and is left unchanged.)

## Modifications

`torchtune/models/qwen2/_tokenizer.py` and `torchtune/models/qwen2_5/_tokenizer.py`: pass the EOS token's actual mask (`mask[-1]`) instead of hardcoded `True` to the mask `truncate(...)`, so the forced last element matches how the appended EOS was masked. `mask` is unmodified between the append and this call, and the token list already forces `self.eos_id` symmetrically.

## Duplicate-check

- Track A — the search API returned HTTP 422 for this token on this repo, so per fail-closed I enumerated all open PRs via the core API (`repos/pytorch/torchtune/pulls?state=open --paginate`) and grepped bodies for `2792` → no references. (The stalled PR #2512 edits the same Qwen EOS area but addresses a different root cause — whether EOS is appended at all — and does not reference #2792.)
- Track B — issue #2792 (state OPEN): the maintainer acknowledged it and labeled it a bug; no "opened a PR / taking this / working on this" claim.

### DIFF
--- a/torchtune/models/qwen2/_tokenizer.py
+++ b/torchtune/models/qwen2/_tokenizer.py
@@ -413,7 +413,7 @@ class Qwen2Tokenizer(ModelTokenizer):
             mask = truncate(
                 tokens=mask,
                 max_seq_len=self.max_seq_len,
-                eos_id=True if add_end_tokens else None,
+                eos_id=mask[-1] if add_end_tokens else None,
                 truncation_type=self.truncation_type,
             )
 
--- a/torchtune/models/qwen2_5/_tokenizer.py
+++ b/torchtune/models/qwen2_5/_tokenizer.py
@@ -198,7 +198,7 @@ class Qwen2_5Tokenizer(Qwen2Tokenizer):  # noqa: N801
             mask = truncate(
                 tokens=mask,
                 max_seq_len=self.max_seq_len,
-                eos_id=True if add_end_tokens else None,
+                eos_id=mask[-1] if add_end_tokens else None,
                 truncation_type=self.truncation_type,
             )
 


ErenAta16 (commented):
The diagnosis holds and the asymmetry is visible in the surrounding four lines, which is what makes this convincing. On `main`:

```python
    mask.append(mask[-1])          # untruncated path: EOS mask follows the final message

    if self.max_seq_len:
            tokens=tokenized_messages,
            eos_id=self.eos_id if add_end_tokens else None,   # real EOS token
            ...
        )
        mask = truncate(
            tokens=mask,
            eos_id=True if add_end_tokens else None,          # always masked out
            ...
        )
```

The token stream gets a genuine EOS appended at the truncation boundary, and the mask gets a hardcoded `True` at the same index. `True` means excluded from the loss, so the EOS is present in the input but contributes no gradient. Four lines above, the same position is given `mask[-1]`, which for a trained assistant turn is `False`. The two paths disagree about the same token.

Consequence matches #2792: with `max_seq_len` set and truncation actually firing, the model sees EOS but is never trained to produce it, so it does not learn to stop. It is quiet because nothing errors and the token is right there in the sequence; only the mask is wrong, and only on truncated samples, so it degrades with sequence length rather than failing outright.

`eos_id=mask[-1] if add_end_tokens else None` restores the symmetry, and reading `mask[-1]` after the `append` above is correct: that element is the appended EOS mask, which already equals the final message's mask. So the truncated and untruncated paths now agree by construction rather than by coincidence.

**One case I would want the PR to take a position on.** When truncation cuts far enough back that the last retained token belongs to a prompt message rather than the final assistant turn, `mask[-1]` still carries the final message's mask, so a forced EOS sitting after prompt content can now be trained. Whether that is desirable is arguable, and I do not think it is a reason to hold this: matching the untruncated path is the defensible default and the previous behaviour was wrong for the common case rather than deliberately right for this one. But a sentence in the description saying the mask follows the final message regardless of where the cut lands would stop this being rediscovered as a bug later.

Two-character change across the three tokenizers, consistent in all of them. Looks right.

yushuosun:
Closing after noticing that torchtune development has formally wound down. The root-cause analysis remains in #2792 if maintenance resumes, but keeping an unlandable PR open does not help the queue.

Issue #2883, opened 2025-07-15
[IMPORTANT] The future of torchtune


ebsmothers:
Dear Tune Community,

We started torchtune in late 2023 with the goal of increasing the accessibility of finetuning in a world where cloud GPUs were prohibitively expensive for many ML practitioners. Thanks to over 150 amazing contributors, today we have 21 recipes spanning the whole post-training stack from SFT to RLHF to QAT. 

During this time the AI landscape has rapidly evolved, with ever-increasing scale, an emphasis on agents, and a reinforcement learning renaissance. We have heard your requests on these fronts loud and clear, and in line with our mission at PyTorch to democratize state-of-the-art AI, we are developing a new product - in a new repo - that captures this evolution of torchtune. The product will provide a simple native PyTorch solution for end-to-end post-training with scale as a first-class citizen, while bringing over what you loved from torchtune (hackable recipes, minimal abstraction, strong integrations across the ecosystem). This is an exciting opportunity and we expect to have more to share in the upcoming weeks.

As a result of this new strategy, **we are stopping active development on torchtune, effective immediately**. We remain committed to supporting our partners and the community to provide a smooth transition. Concretely, this means:

- Torchtune will continue to receive critical bug fixes and security patches during 2025
- Discord and Github issues will remain open for support
- No new features will be added to the library
- We would love to keep working with you to assist with a smooth transition and build this new solution together!

If you have any questions or concerns, please reach out on Discord or comment here. Thank you all for your support and all that you've done to build this LLM post-training community with us.

From all of us on the torchtune team,
@joecummings @ebsmothers @pbontrager @felipemello1 @RdoubleA @kartikayk @rohan-varma 

Our heartfelt thanks to our many contributors, including 
@SalmanMohammadi @krammnic @daniellepintz @andrewor14 @SLR722 @ankitageorge @svekars @NicolasHug @gokulavasan @janeyx99 @jerryzh168 @thomasjpfan @Ankur-singh @acisseJZhong @calvinpelletier @mirceamironenco @weifengpy @nathan-az @solitude-alive @songhappy @Optimox @RedTachyon @lindawangg @lucylq @hardikjshah @msaroufim @yechenzhi @mreso @IvanKobzarev @tcapelle @gau-nernst @tambulkar @bogdansalyp @albert-inflection

init27:
Thanks team for all the monumental effort and the gift of Tune. Excited for what's next!

iqiancheng:
[...]

PR #2969, opened 2026-07-09
fix: respect Gemma tokenizer inference mode.

### LINKED ISSUE #2478: Add add_end_token to gemma
Llama3 Tokenizer was updated to support add_end_token in tokenize_messages to support correct generation https://github.com/pytorch/torchtune/pull/1494. These changes need to be made for gemma.

Changes:
- Update tokenize_message to use add_start_tokens and add_end_tokens like in https://github.com/pytorch/torchtune/pull/1494
- Replace add_eos with add_end_tokens and update tokenize_messages as in https://github.com/pytorch/torchtune/pull/1494
- In call update tokens, mask = self.tokenize_messages(messages) to tokens, mask = self.tokenize_messages(messages, add_end_tokens=not inference)

ofeliacode:
#### Context
  - [ ] add a new feature
  - [x] fix a bug
  - [ ] other (please add here)

  Fixes #2478.

  `GemmaTokenizer.tokenize_messages(..., add_eos=False)` already supports omitting EOS, but
  `GemmaTokenizer.__call__(..., inference=True)` was not passing the inference mode through. As a result,
  the callable tokenizer path always appended EOS.

  #### Changelog
  * Pass `add_eos=not inference` from `GemmaTokenizer.__call__`.
  * Add a regression test covering `inference=True`.

  #### Test plan
  Please make sure to do each of the following if applicable to your PR. If you're unsure about any one of
  these just ask and we will happily help. We also have a [contributing page](https://github.com/pytorch/
  torchtune/blob/main/CONTRIBUTING.md) for some guidance on contributing.

  - [x] add [unit tests](https://github.com/pytorch/torchtune/tree/main/tests/torchtune) for any new
  functionality
  - [ ] update [docstrings](https://github.com/pytorch/torchtune/tree/main/docs/source) for any new or
  updated methods or classes
  - [x] include relevant commands and any other artifacts in this summary (pastes of loss curves, eval
  results, etc.)

  Commands run:
  * `git diff --check`
  * `python -m py_compile torchtune/models/gemma/_tokenizer.py tests/torchtune/models/gemma/
  test_gemma_tokenizer.py`

  I attempted to run:
  * `pytest tests/torchtune/models/gemma/test_gemma_tokenizer.py -q`

  but my local environment failed during torchtune test bootstrap because of a PyTorch/torchao compatibility
  issue around `torchao.quantization.NF4Tensor`.

  #### UX
  If your function changed a public API, please add a dummy example of what the user experience will look
  like when calling it.
  Here is a [docstring example](https://github.com/pytorch/torchtune/
  blob/6a7951f1cdd0b56a9746ef5935106989415f50e3/torchtune/modules/vision_transformer.py#L285)
  and a [tutorial example](https://pytorch.org/torchtune/main/tutorials/qat_finetune.html#applying-qat-to-
  llama3-models)

### DIFF
--- a/tests/torchtune/models/gemma/test_gemma_tokenizer.py
+++ b/tests/torchtune/models/gemma/test_gemma_tokenizer.py
@@ -81,3 +81,31 @@ class TestGemmaTokenizer:
         expected_mask = [True] * 75 + [False] * 124
         assert expected_tokens == tokens
         assert expected_mask == mask
+
+    def test_call_drops_eos_for_inference(self, tokenizer, expected_tokens):
+        messages = [
+            Message(
+                role="user",
+                content="Below is an instruction that describes a task. Write a response "
+                "that appropriately completes the request.\n\n### Instruction:\nGenerate "
+                "a realistic dating profile bio.\n\n### Response:\n",
+                masked=True,
+            ),
+            Message(
+                role="assistant",
+                content="I'm an outgoing and friendly person who loves spending time with "
+                "friends and family. I'm also a big-time foodie and love trying out new "
+                "restaurants and different cuisines. I'm a big fan of the arts and enjoy "
+                "going to museums and galleries. I'm looking for someone who shares my "
+                "interest in exploring new places, as well as someone who appreciates a "
+                "good conversation over coffee.",
+            ),
+        ]
+
+        train_sample = tokenizer({"messages": messages.copy()})
+        inference_sample = tokenizer({"messages": messages.copy()}, inference=True)
+
+        assert train_sample["tokens"] == expected_tokens
+        assert train_sample["mask"] == [True] * 75 + [False] * 125
+        assert inference_sample["tokens"] == expected_tokens[:-1]
+        assert inference_sample["mask"] == [True] * 75 + [False] * 124
--- a/tests/torchtune/models/mistral/test_mistral_tokenizer.py
+++ b/tests/torchtune/models/mistral/test_mistral_tokenizer.py
@@ -96,3 +96,14 @@ class TestMistralTokenizer:
         expected_mask = [True] * 75 + [False] * 124
         assert expected_tokens == tokens
         assert expected_mask == mask
+
+    def test_call_drops_eos_for_inference(self, messages, expected_tokens):
+        tokenizer = self.tokenizer(template=False)
+
+        train_sample = tokenizer({"messages": messages.copy()})
+        inference_sample = tokenizer({"messages": messages.copy()}, inference=True)
+
+        assert train_sample["tokens"] == expected_tokens
+        assert train_sample["mask"] == [True] * 75 + [False] * 125
+        assert inference_sample["tokens"] == expected_tokens[:-1]
+        assert inference_sample["mask"] == [True] * 75 + [False] * 124
--- a/torchtune/models/gemma/_tokenizer.py
+++ b/torchtune/models/gemma/_tokenizer.py
@@ -165,7 +165,7 @@ class GemmaTokenizer(ModelTokenizer, Transform):
                 and the "messages" field removed.
         """
         messages = sample.pop("messages")
-        tokens, mask = self.tokenize_messages(messages)
+        tokens, mask = self.tokenize_messages(messages, add_eos=not inference)
         sample["tokens"] = tokens
         sample["mask"] = mask
         return sample
--- a/torchtune/models/mistral/_tokenizer.py
+++ b/torchtune/models/mistral/_tokenizer.py
@@ -196,7 +196,7 @@ class MistralTokenizer(ModelTokenizer, Transform):
             inference (bool): Whether the template is being used for inference or not.
         """
         messages = sample.pop("messages")
-        tokens, mask = self.tokenize_messages(messages)
+        tokens, mask = self.tokenize_messages(messages, add_eos=not inference)
         sample["tokens"] = tokens
         sample["mask"] = mask
         return sample


ErenAta16 (commented):
Fix is correct. `__call__` takes `inference` and `tokenize_messages` takes `add_eos`, and the call between them dropped the connection, so `add_eos` fell back to its default of `True` and the prompt got a trailing EOS at inference time. Forwarding it is the whole fix.

**The same defect is in the Mistral tokenizer and I would fold it into this PR.** I checked all three tokenizers that make a bare `self.tokenize_messages(messages)` call from `__call__`:

```
                 __call__   has `inference` param   tokenize_messages has add_eos   bare call
gemma            L152       yes (L153)              yes (L107)                      L168   <- this PR
mistral          L182       yes (L183)              yes (L132)                      L199   <- same bug
phi4             L172       no                      -                               L177   <- not applicable
```

Mistral is the identical shape: `__call__(self, sample, inference: bool = False)` accepts the flag, `tokenize_messages(self, messages, add_eos: bool = True)` accepts the knob, and line 199 connects neither. Same one-line change, `add_eos=not inference`.

Phi4 is a false match. Its `__call__` has no `inference` parameter at all, so there is nothing to forward and the bare call is correct there. Worth saying so explicitly, otherwise the next person doing this sweep will "fix" it and change behaviour.

Fixing Gemma alone leaves users of Mistral with the same silent problem, and since the fix is byte-identical it is cheaper to do both here than to open a second PR and a second test. If you would rather keep this PR scoped to #2478, I am happy to open the Mistral one separately, just say which you prefer.

On the test: `test_call_drops_eos_for_inference` going through `__call__` rather than `tokenize_messages` is the right level, since `tokenize_messages(add_eos=False)` already worked and testing it again would not have caught this. Asserting on the tokens from the public entry point is what pins the wiring.

ErenAta16:
Flagging a collision rather than a change request. #2977 fixes the same `inference` wiring gap this PR does, plus Mistral, and touches the same two files:

```
this PR  torchtune/models/gemma/_tokenizer.py       +1/-1
         tests/.../test_gemma_tokenizer.py          +28/-0
#2977    torchtune/models/gemma/_tokenizer.py       +9/-5
         tests/.../test_gemma_tokenizer.py          +22/-6
         torchtune/models/mistral/_tokenizer.py     +9/-5
         tests/.../test_mistral_tokenizer.py        +30/-6
```

It also renames `add_eos` to `add_end_tokens` and adds `add_start_tokens`, bringing Gemma and Mistral in line with llama2, llama3 and phi3, which already use that naming. So the two PRs rewrite the same test in incompatible ways and whichever merges second needs real rework rather than a rebase.

My read is that #2977 subsumes this one, since it covers Gemma and the identical Mistral gap I mentioned in my earlier review here. That is a maintainer call rather than mine, so I have noted the same thing on that thread; neither of you should find out from a conflict.

If the maintainers prefer the smaller change, this PR is still the more surgical fix and lands the Gemma half without the API rename.

ofeliacode:
@ErenAta16 Hi! Applied the same inference-mode fix to the Mistral tokenizer and added a regression test verifying that EOS is omitted when inference=True

ErenAta16:
Before this goes further, you and #2977 are editing the same four files, and I
should flag it since I have been reviewing that one and only noticed the overlap
now.

```
#2969  torchtune/models/gemma/_tokenizer.py     +1/-1     tests +28/-0
       torchtune/models/mistral/_tokenizer.py   +1/-1     tests +11/-0

#2977  torchtune/models/gemma/_tokenizer.py     +22/-5    tests +35/-6
       torchtune/models/mistral/_tokenizer.py   +22/-5    tests +43/-6
```

Same files, and the behaviour overlaps rather than sitting side by side. #2977
aligns both tokenizers on the `add_start_tokens` / `add_end_tokens` API and
routes `__call__` through `add_end_tokens=not inference`, which is the same
inference-mode behaviour this PR fixes directly. So whichever lands second gets a
conflict in all four files, and the second author rewrites work that is already
merged.

You opened first, which normally settles it. What complicates it here is that
#2977 is the wider change and has been through three rounds of review, including
a deprecation path for the old `add_eos` keyword and regression tests for both
the legacy and canonical spellings. Landing this one first would leave that PR
rebasing onto a partial version of its own fix.

I am not going to pick between them, and it is not mine to pick. This needs a
maintainer to say which shape they want. Worth raising it in both threads so the
decision happens before one of you spends the rebase.

@patrickswedish flagging you here as well.

ErenAta16:
Following up on the overlap I raised, because it turns out not to matter and the
reason is worth knowing before either of you spends more time here.

torchtune is not maintained any more. The README carries it at the top:

> Torchtune is no longer actively maintained: torchtune development wound down in
> 2025 — see "The future of torchtune" (#2883).

and the tree matches. The last commit on `main` is `bd2a0fc7`, dated 2026-04-23,
which is the commit that added that README notice. Nothing has landed in the four
months since, against 457 open issues.

So the collision question I raised is moot. Neither PR is going to be merged
ahead of the other, because neither is going to be merged. I would rather say
that plainly than leave you both rebasing around each other.

For what it is worth, the work itself held up. The deprecation path, the negative
assertion on the canonical call, and the equivalence assertions between the
legacy and canonical spellings are all correct, and the reasoning is on the
record here if torchtune is ever picked up again or if the same pattern needs
fixing in a fork. That is not nothing, it is just not a merge.

Apologies for the noise on the overlap. I should have checked whether the repo
was alive before worrying about which of two PRs had seniority in its queue.

PR #2973, opened 2026-07-21
fix(security): restrict recipe _component_ imports to trusted roots

### LINKED ISSUE #2971: Arbitrary code execution via the _component_ field in an untrusted recipe config
reported via https://github.com/meta-pytorch/torchtune/security/advisories/GHSA-gw63-xxf2-xxcg on 10 June 2026 - no response so trying here:

### Summary

Loading or validating an untrusted torchtune recipe config (a .yaml data file) executes arbitrary code. A config node names a Python callable in its `_component_` field, and torchtune resolves it by importing the named module and calling it with the config arguments, with no allowlist. Loading a recipe config imports an attacker-named module (running its top-level code) and can call a dangerous callable directly. Recipe configs are explicitly shared and distributed, so loading one is a deliberate action.

### Details

In `torchtune/config/_utils.py`, `_get_component_from_path` (line 39) resolves the `_component_` string with `import_module(module_path)` then `getattr(...)` (single-part branch around line 85, dotted branch around lines 108 to 110). There is no allowlist, no trusted-module set, and no sandbox.

Two reachable sinks:
- `torchtune/config/_instantiate.py:54` `_instantiate_node` calls `_get_component_from_path(obj["_component_"])` then calls the result with kwargs taken from the YAML. Every recipe instantiates attacker-controlled config nodes (for example `recipes/full_finetune_distributed.py` lines 336, 368, 387, 612, 784 for metric_logger, loss, tokenizer, model, optimizer).
- `torchtune/config/_validate.py:28` `validate` (invoked by `tune validate <config>`) calls `_get_component_from_path` for every node. The surprise is that a command documented as only checking that the config is well-formed still imports the attacker-named module, running its top-level code.

So a config with `_component_: os.system` and `command: "..."` calls a dangerous stdlib callable directly, and a config naming any module imports it (running module-body code) even on the nominally read-only validate path. This is the same data-to-code pattern as other config frameworks, shipped with no restriction. No public advisory covers this class on torchtune.

### PoC

```
pip install torchtune torchao
python poc_torchtune_component_rce.py
```

`poc_torchtune_component_rce.py`:

```python
import os, tempfile
from torchtune import config

marker = os.path.join(tempfile.gettempdir(), "TT_PWNED")
if os.path.exists(marker): os.remove(marker)

# A DATA file: a shared torchtune fine-tuning recipe (.yaml). No .py shipped.
EVIL = f"""
loss:
  _component_: os.system
  command: "touch {marker}"
"""
cfg = OmegaConf.create(EVIL)
config.instantiate(cfg.loss)   # exactly recipes/full_finetune_distributed.py:368
print("code executed:", os.path.exists(marker))
```

Observed: `config.instantiate(cfg.loss)` runs the command and the marker is created. A `tune validate` variant (a yaml `model: { _component_: evilmod.Trainer }` next to an importable `evilmod.py` with code in its module body) executes the module body during validation, with no instantiation.

### Impact

Arbitrary code execution as the user when a victim loads, validates, or runs an attacker-supplied recipe config. Recipe configs are shared via `tune cp`, the configs registry, and community recipes, so a malicious config is a realistic distribution vector. Running a recipe runs training code by design; the vulnerability is that a data YAML imports and calls an arbitrary module and callable with no allowlist, including on the read-only validate path.

Affected version: 0.6.1 and 0.7.0

Solaris-star:
## Summary
Recipe configs resolve `_component_` via `import_module` with no allowlist. Loading or validating untrusted YAML can therefore import arbitrary modules and run top-level code, or call dangerous callables such as `os.system`.

Restrict multi-part (and non-local single-part) component paths to the package roots used by official recipes/tests:

- `torchtune.*`
- `torch.*`
- `bitsandbytes.*`

Caller-local names (globals) remain resolvable for tests/helpers.

## Testing
- Updated `tests/torchtune/config/test_config_utils.py` (allowed roots + blocks `os.system` / `subprocess.run`)
- Lightweight local import checks for allowed/disallowed paths

Fixes #2971

### DIFF
--- a/tests/torchtune/config/test_config_utils.py
+++ b/tests/torchtune/config/test_config_utils.py
@@ -43,7 +43,7 @@ class TestUtils:
         # Test valid paths with three components
         valid_paths = [
             "torchtune",
-            "os.path.join",
+            "torch.nn.Linear",
         ]
         for path in valid_paths:
             result = _get_component_from_path(path)
@@ -90,16 +90,29 @@ class TestUtils:
 
         # Multi-part path with import failure
         with pytest.raises(
-            InstantiationError, match=r"Could not import module 'os\.nonexistent': .*"
+            InstantiationError, match=r"Could not import module 'torch\.nonexistent': .*"
         ):
-            _get_component_from_path("os.nonexistent.attr")
+            _get_component_from_path("torch.nonexistent.attr")
 
         # Multi-part path with attribute error
         with pytest.raises(
             InstantiationError,
-            match=r"Module 'os\.path' has no attribute 'nonexistent'",
+            match=r"Module 'torch\.nn' has no attribute 'nonexistent'",
         ):
-            _get_component_from_path("os.path.nonexistent")
+            _get_component_from_path("torch.nn.nonexistent")
+
+        # Disallowed roots must not be imported (security: untrusted recipe configs)
+        with pytest.raises(
+            InstantiationError,
+            match=r"Component path 'os\.system' is not allowed",
+        ):
+            _get_component_from_path("os.system")
+
+        with pytest.raises(
+            InstantiationError,
+            match=r"Component path 'subprocess\.run' is not allowed",
+        ):
+            _get_component_from_path("subprocess.run")
 
     @mock.patch(
         "torchtune.config._parse.OmegaConf.load", return_value=OmegaConf.create(_CONFIG)
--- a/torchtune/config/_utils.py
+++ b/torchtune/config/_utils.py
@@ -14,6 +14,17 @@ from omegaconf import DictConfig, OmegaConf
 from torchtune.config._errors import InstantiationError
 from torchtune.utils._logging import get_logger, log_rank_zero
 
+# Recipe configs resolve `_component_` paths via importlib. Only allow known
+# package roots used by torchtune recipes/tests so untrusted YAML cannot import
+# arbitrary modules (e.g. os.system) or run top-level import side effects.
+_ALLOWED_COMPONENT_ROOTS = frozenset(
+    {
+        "torchtune",
+        "torch",
+        "bitsandbytes",
+    }
+)
+
 
 def log_config(recipe_name: str, cfg: DictConfig) -> None:
     """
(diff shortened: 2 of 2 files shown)


Solaris-star:
recheck

ErenAta16 (commented):
The enforcement itself is written correctly, which is worth saying because this is where allowlists usually go wrong. It keys on `parts[0]` with exact set membership rather than a string prefix, so `torchtune_evil.payload` is rejected rather than sailing through on `startswith("torchtune")`. Good.

Two problems with the approach, though, and I think they are both blocking.

**The allowlist does not meet its own threat model.** The stated goal is that untrusted YAML cannot run arbitrary code. But `torch` is an allowed root, and several arbitrary-code primitives live under it:

```
torch.hub.load            resolves -> function   (root 'torch' allowlisted)
torch.load                resolves -> function
torch.jit.load            resolves -> function
torch.serialization.load  resolves -> function
```

`torch.hub.load` fetches a repository and executes its `hubconf.py`, so `_component_: torch.hub.load` with attacker-chosen arguments is a remote code execution path that survives this change untouched. `torch.load` on an attacker-supplied path is the classic pickle deserialisation issue. A config author who can set `_component_` can still reach all of them.

So the change moves the boundary from "any module" to "any module under three roots", and the three roots still contain what an attacker needs. Blocking `os.system` and `subprocess.run` makes the tests pass without making the property true, and I would rather see no allowlist than one that reads as a guarantee it does not provide. If the threat model really is untrusted YAML, the enforcement has to be on the resolved callable, not on the import root, and that is a much larger design question.

**It breaks a documented feature.** Custom components are a supported workflow with their own docs page, `docs/source/basics/custom_components.rst`, linked from the index, and `tune_cli.rst` explicitly tells users to mix custom recipes and configs. Every one of those points `_component_` at a user module, none of which is `torchtune`, `torch` or `bitsandbytes`. This change makes all of them fail with "is not allowed" the moment it lands.

That is the part I would expect to generate issues immediately after release. Anyone with a custom dataset, message transform or optimiser has a config that stops working, with an error message that reads like a bug rather than a policy.

**On the test changes.** Rewriting `os.path.join` to `torch.nn.Linear` and `os.nonexistent` to `torch.nonexistent` throughout `test_get_component_from_path` is necessary given the new rule, but it does mean the existing tests no longer exercise the "arbitrary stdlib module" case they were originally written for. Worth noting in the description so it does not look like unrelated churn.

**What I would suggest instead.** If the concern is running someone else's config, the honest framing is that a torchtune config is executable input, in the same way a `conftest.py` or a `Makefile` is, and the fix is documentation plus not running configs you did not write. If you want a mechanism, an opt-in strict mode that users enable when processing untrusted configs would give the property without breaking the documented custom-component path for everyone else.

Happy to be wrong about the custom-component impact if there is a fallback path I missed, but I could not find one in `_get_component_from_path`: the multi-part branch raises before `import_module` is reached.

Solaris-star:
Thank you for the detailed review. You are right on both blocking points: allowing the `torch` root does not establish an untrusted-YAML boundary because callables such as `torch.hub.load` and `torch.load` remain reachable, and enforcing the allowlist unconditionally breaks the documented custom-components workflow.

I am withdrawing this implementation rather than adding a blacklist that would create a false security guarantee. A useful follow-up would need a separately designed opt-in strict mode with an explicit callable/argument threat model, while preserving the default custom-component behavior. I will close this PR so the current unsound policy is not merged.

Solaris-star:
Withdrawing this implementation after review: the unconditional root allowlist is incomplete for the stated threat model and breaks supported custom components. A future opt-in strict mode should be designed separately.

ErenAta16:
Withdrawing this was the right call, and I would rather say that plainly than let a closed PR stand as a rebuke. Adding a blacklist on top of the `torch` root would have been the worse outcome: the policy would have read as a security boundary while `torch.hub.load` and `torch.load` stayed reachable, and a guarantee that does not hold is more dangerous than no guarantee, because people stop checking.

One concrete thing for whoever picks up the follow-up, so the analysis in this thread is not lost. The reason a root allowlist cannot work here is that `_component_` resolution is import-path based while the threat is callable-behaviour based, and those do not line up: the dangerous callables are inside roots you have to allow for the library to function at all. So a workable strict mode has to enumerate at the callable level rather than the module level, and it has to be opt-in, because the documented custom-components workflow requires resolving paths that no fixed allowlist can know in advance.

That also sets the acceptance test for the follow-up: it is not "does the allowlist block a bad path", it is "with strict mode on, does the documented custom-component config still load, and is there a reachable callable inside an allowed root that executes attacker-controlled input". The second half is the one that decides whether the boundary is real.

PR #2977, opened 2026-07-26
Fix Mistral and Gemma end-token control

### LINKED ISSUE #2479: Add add_end_token to Mistral tokenizer
Llama3 Tokenizer was updated to support add_end_token in tokenize_messages to support correct generation https://github.com/pytorch/torchtune/pull/1494. These changes need to be made to Mistral.

Changes:
- Update tokenize_message to use add_start_tokens and add_end_tokens like in https://github.com/pytorch/torchtune/pull/1494
- Replace add_eos with add_end_tokens and update tokenize_messages as in https://github.com/pytorch/torchtune/pull/1494
- In call update tokens, mask = self.tokenize_messages(messages) to tokens, mask = self.tokenize_messages(messages, add_end_tokens=not inference)

patrickswedish:
## What

- align Mistral and Gemma tokenizers with the existing Llama 2 `add_start_tokens` / `add_end_tokens` API
- omit the final EOS token when either tokenizer is used for inference
- add focused coverage for BOS/EOS control and Mistral inference behavior

## Why

Mistral and Gemma still exposed the older `add_eos` switch and their transform path ignored the `inference` flag. That left generation inputs ending in EOS, unlike the corrected Llama tokenizer behavior, and provided no way to suppress the initial BOS token.

Fixes #2479.

## Impact & Compatibility Note

- **Compatibility note**: `add_end_tokens` is now the canonical keyword for Mistral and Gemma, aligning with the broader tokenizer API. The previous `add_eos` keyword remains supported as a compatibility alias, so existing direct callers do not need to migrate immediately.
- Default training behavior is unchanged.
- Inference transforms now omit the terminal EOS token so generation can continue, and callers can control boundary tokens consistently across tokenizers.

## Checks

- `ruff check` on all four changed files
- `ruff format --check` on all four changed files
- Python bytecode compilation on all four changed files

### DIFF
--- a/tests/torchtune/models/gemma/test_gemma_tokenizer.py
+++ b/tests/torchtune/models/gemma/test_gemma_tokenizer.py
@@ -55,7 +55,16 @@ class TestGemmaTokenizer:
         assert expected_tokens == tokens
         assert expected_mask == mask
 
-    def test_tokenize_messages_drop_eos(self, tokenizer, expected_tokens):
+    @pytest.mark.parametrize(
+        "add_start_tokens, add_end_tokens",
+        [
+            (True, True),
+            (False, False),
+        ],
+    )
+    def test_tokenize_messages_special_token_control(
+        self, tokenizer, expected_tokens, add_start_tokens, add_end_tokens
+    ):
         messages = [
             Message(
                 role="user",
@@ -74,10 +83,17 @@ class TestGemmaTokenizer:
                 "good conversation over coffee.",
             ),
         ]
-        tokens, mask = tokenizer.tokenize_messages(messages, add_eos=False)
-        # Drop eos token.
-        expected_tokens = expected_tokens[:-1]
-        # On 1 less then with eos
-        expected_mask = [True] * 75 + [False] * 124
+        tokens, mask = tokenizer.tokenize_messages(
+            messages,
+            add_start_tokens=add_start_tokens,
+            add_end_tokens=add_end_tokens,
+        )
+        if not add_start_tokens:
+            expected_tokens = expected_tokens[1:]
+        if not add_end_tokens:
+            expected_tokens = expected_tokens[:-1]
+        expected_mask = [True] * (75 if add_start_tokens else 74) + [False] * (
+            125 if add_end_tokens else 124
+        )
         assert expected_tokens == tokens
         assert expected_mask == mask
--- a/tests/torchtune/models/gemma/test_gemma_tokenizer.py
+++ b/tests/torchtune/models/gemma/test_gemma_tokenizer.py
@@ -4,6 +4,8 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
+import warnings
+
 import pytest
 from tests.common import ASSETS
 from torchtune.data import Message
@@ -83,11 +85,13 @@ class TestGemmaTokenizer:
                 "good conversation over coffee.",
             ),
         ]
-        tokens, mask = tokenizer.tokenize_messages(
-            messages,
-            add_start_tokens=add_start_tokens,
-            add_end_tokens=add_end_tokens,
-        )
+        with warnings.catch_warnings():
+            warnings.simplefilter("error", DeprecationWarning)
+            tokens, mask = tokenizer.tokenize_messages(
+                messages,
+                add_start_tokens=add_start_tokens,
+                add_end_tokens=add_end_tokens,
+            )
         if not add_start_tokens:
             expected_tokens = expected_tokens[1:]
         if not add_end_tokens:
@@ -98,12 +102,11 @@ class TestGemmaTokenizer:
         assert expected_tokens == tokens
         assert expected_mask == mask
 
-        if not add_end_tokens:
-            with pytest.deprecated_call():
-                legacy_tokens, legacy_mask = tokenizer.tokenize_messages(
-                    messages,
-                    add_start_tokens=add_start_tokens,
-                    add_eos=False,
-                )
-            assert legacy_tokens == tokens
-            assert legacy_mask == mask
+        with pytest.deprecated_call():
+            legacy_tokens, legacy_mask = tokenizer.tokenize_messages(
+                messages,
+                add_start_tokens=add_start_tokens,
+                add_eos=add_end_tokens,
+            )
+        assert legacy_tokens == tokens
+        assert legacy_mask == mask
--- a/tests/torchtune/models/gemma/test_gemma_tokenizer.py
+++ b/tests/torchtune/models/gemma/test_gemma_tokenizer.py
@@ -55,7 +55,16 @@ class TestGemmaTokenizer:
         assert expected_tokens == tokens
         assert expected_mask == mask
 
-    def test_tokenize_messages_drop_eos(self, tokenizer, expected_tokens):
+    @pytest.mark.parametrize(
+        "add_start_tokens, add_end_tokens",
+        [
+            (True, True),
+            (False, False),
+        ],
+    )
+    def test_tokenize_messages_special_token_control(
+        self, tokenizer, expected_tokens, add_start_tokens, add_end_tokens
+    ):
         messages = [
             Message(
                 role="user",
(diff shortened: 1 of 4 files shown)


ErenAta16 (commented):
The alignment argument holds up when you count the tokenizers. Mistral and Gemma really are the outliers:

```
llama2   add_start_tokens: bool = True,  add_end_tokens: bool = True
llama3   add_end_tokens: bool = True
phi3     add_end_tokens: bool = False
mistral  add_eos: bool = True        <- renamed here
gemma    add_eos: bool = True        <- renamed here
```

so this is bringing the last two into line rather than introducing a third convention, and adding `add_start_tokens` gives Gemma and Mistral the BOS control the Llama 2 tokenizer already had.

**On the rename being breaking.** It is, for anyone calling `tokenize_messages(messages, add_eos=False)` on these two tokenizers, and there is no alias. I went looking for in-tree fallout and it is smaller than it first appears: the `add_eos=` occurrences in `recipes/eleuther_eval.py` (lines 147, 361, 363) are on `tokenizer.encode`, a different method this PR does not touch, so they are unaffected. Inside the repo only the two test files are affected and both are updated here. So the exposure is external callers, and given three of the five tokenizers already use the new name, an external caller passing `add_eos` was already writing tokenizer-specific code. Worth a line in the description saying so explicitly, because "renamed a public keyword argument" is the kind of thing that wants to be a deliberate note rather than something a user discovers from a `TypeError`.

**This collides with #2969, and I think it supersedes it.** That PR is Gemma-only and touches the same two files:

```
#2969  ofeliacode      07-09   torchtune/models/gemma/_tokenizer.py  +1/-1
                               tests/.../test_gemma_tokenizer.py     +28/-0
#2977  patrickswedish  07-26   torchtune/models/gemma/_tokenizer.py  +9/-5
                               tests/.../test_gemma_tokenizer.py     +22/-6
                               torchtune/models/mistral/_tokenizer.py +9/-5
                               tests/.../test_mistral_tokenizer.py   +30/-6
```

#2969 fixes the same defect, `__call__` not forwarding `inference` into the tokenize call, with the one-line `add_eos=not inference`. This PR does that and Mistral as well, which is exactly what I suggested on #2969 when I reviewed it: the Mistral tokenizer has the identical wiring gap and the fix is byte-identical.

Since #2969 rewrites the same test that this PR rewrites differently, whichever lands second needs more than a trivial rebase. Worth the maintainers deciding between them rather than reviewing both: #2969 is older and smaller, this one is complete and also does the API alignment. My read is that this one makes #2969 redundant, but that is a call for someone with commit rights, and I have said the same on that thread so neither author is surprised.

patrickswedish:
@ErenAta16 Thanks for the thorough review!

I have updated the PR description to explicitly document the compatibility note regarding the \dd_eos\ -> \dd_end_tokens\ keyword parameter alignment for external callers.

I also reviewed the overlap with #2969 and agree that maintainers can decide whether to merge the broader cross-tokenizer alignment in this PR or take #2969 first. Happy to follow maintainer direction on precedence.

ErenAta16 (commented):
Thanks for updating the description, and for going and reading #2969 instead of just asserting precedence over it. Both were the right calls.

I checked the compatibility note against the diff rather than taking it on trust, because a description promising an alias that is not actually there would be worse than no note at all. It is there, and it does what the note says:

```python
add_eos: bool | None = None,
...
if add_eos is not None:
    add_end_tokens = add_eos
```

The docstring stating "when provided, it takes precedence" is the part that makes it safe to rely on, since that ordering is otherwise the kind of thing each caller would have to discover.

Something useful you may not have seen: torchtune already has this exact shape, and yours matches it. `torchtune/data/_messages.py` deprecates `train_on_input` in favour of `masking_strategy` using the same `None` sentinel and the same "old parameter wins" precedence, in five transforms (`_messages.py:203`, `354`, `484`, `650`, `782`).

The one element that precedent has and this does not is the warning:

```python
if train_on_input is not None:
    warn(
        "train_on_input is deprecated and will be removed in a future release. "
        ...
        "the value of masking_strategy will be ignored until torchtune 0.7. ",
        DeprecationWarning,
        stacklevel=2,
    )
```

Without it, "deprecated" exists only in the docstring, so a caller still passing `add_eos` has no way to find out they should move, and the alias has no exit condition. A `warn(..., DeprecationWarning, stacklevel=2)` in both tokenizers, naming a release the way the existing message does, would turn it from a permanent second spelling into an actual deprecation. Small change, and it is copying a pattern the repo already uses in five places rather than introducing one.

Agree that precedence against #2969 is a maintainer call.

patrickswedish:
@ErenAta16 Thanks for pointing out the deprecation warning precedent in `torchtune/data/_messages.py`!

I have updated both `GemmaTokenizer.tokenize_messages` and `MistralTokenizer.tokenize_messages` to emit a `warn("add_eos is deprecated and will be removed in a future release. Please use add_end_tokens instead.", DeprecationWarning, stacklevel=2)` when the legacy `add_eos` parameter is provided. I also updated the test suite to verify deprecation warning emission via `pytest.deprecated_call()`.

Rebased and squashed to 1 clean atomic commit on latest `upstream/main`.

ErenAta16 (commented):
That is the right shape, and the part I did not ask for is the part I like most: asserting `legacy_tokens == tokens` and `legacy_mask == mask` rather than only that a warning fired. That turns it from a warning test into an equivalence test, which is what actually protects callers on the alias.

Checked the diff rather than assuming:

- `from warnings import warn` is added in both `_tokenizer.py` files, so the deprecated path does not `NameError` the first time someone hits it. This is the failure mode I was most worried about, since nothing would have caught it until a real legacy caller came along.
- The guard is `if add_eos is not None:`, so it fires on explicit `False` as well as `True`, and stays quiet when the parameter is omitted.
- `stacklevel=2` points at the caller rather than at `tokenize_messages`, which matches the precedent.

Two gaps, both small.

**Nothing asserts the new path is warning-free.** `pyproject.toml` has no `filterwarnings` under `[tool.pytest.ini_options]` (only `--showlocals`, `--import-mode=prepend`, and the two integration flags), so warnings are not errors here. That means if the `if add_eos is not None:` guard were later dropped or the `warn` moved above it, every existing caller would get a `DeprecationWarning` on every `tokenize_messages` call and the suite would stay green. A `warnings.catch_warnings()` block with `simplefilter("error")` around the non-legacy call, or an assertion on `recwarn`, closes it in a line or two.

**`add_eos=True` is never exercised.** The legacy call sits under `if not add_end_tokens:` and always passes `add_eos=False`, so the tests only cover the alias in its false form. `add_eos=True` is the more common legacy call, since it was the old default. It would almost certainly pass, but it is the branch real callers are on.

One leftover from my earlier note: the message says "in a future release" where the `_messages.py` precedent names one ("the value of `masking_strategy` will be ignored until torchtune 0.7"). Naming a version is what gives a deprecation an end, and it is the difference between a plan and a permanent second spelling. Entirely your call, and not worth another round on its own.

patrickswedish:
@ErenAta16 Thanks — both requested coverage gaps are now included in 4e31b8b2.

- canonical add_end_tokens usage now treats any unexpected DeprecationWarning as a test failure;
- legacy add_eos=True and add_eos=False are both covered for Gemma and Mistral;
- each legacy path asserts exact token and mask equivalence with its canonical add_end_tokens counterpart.

Validation:
- focused tokenizer tests: 8 passed
- models/datasets/generation: 280 passed; the 3 Llama3ScaledRoPE CPU precision failures reproduce identically on clean current main
- full pytest tests/: the same 6 collection errors reproduce on clean current main
- pre-commit run --all-files: PASS
- Ruff check/format: PASS
- git diff --check: PASS

ErenAta16 (commented):
Checked `4e31b8b2`. Both gaps are closed, and the negative assertion is built the
way that does not depend on repo config:

```python
with warnings.catch_warnings():
    warnings.simplefilter("error", DeprecationWarning)
```

Scoped to the canonical call, so if the `if add_eos is not None:` guard were ever
dropped, this raises instead of passing quietly. That was the failure mode I was
worried about, and `pyproject.toml` carrying no `filterwarnings` no longer
matters for it.

`add_eos=add_end_tokens` on the legacy call covers both values now, and removing
the `if not add_end_tokens:` wrapper is what made that possible.

One small blind spot, and I am not asking you to hold the PR for it. The
parametrisation is `[(True, True), (False, False)]`, so the two flags are equal
in every case tested. If `add_start_tokens` and `add_end_tokens` were ever
transposed in the implementation, all of these tests would still pass, because
swapping two equal values changes nothing. A single off-diagonal case such as
`(True, False)` closes it. Both flags are new in this PR, which is when a
transposition is most likely to slip in unnoticed.

ErenAta16:
Separate from the review, and it is my miss for not spotting it earlier: #2969 by
@ofeliacode touches the same four files as this PR.

```
#2969   gemma/_tokenizer.py  +1/-1     mistral/_tokenizer.py  +1/-1   (+ tests)
#2977   gemma/_tokenizer.py  +22/-5    mistral/_tokenizer.py  +22/-5  (+ tests)
```

It fixes the inference-mode behaviour directly, which this PR also produces
through `__call__` passing `add_end_tokens=not inference`. So the two overlap
rather than sitting alongside each other, and whichever lands second conflicts in
all four files.

#2969 was opened first, which normally decides it. What makes this one awkward is
that your PR is the wider change and has been through review, so landing the
narrow one first would leave you rebasing onto a partial version of your own fix.

I have raised it there as well and I am not going to pick between them. It needs
a maintainer to say which shape they want, and it is better decided now than
after one of you does the rebase.

ErenAta16:
Following up on the overlap I raised, because it turns out not to matter and the
reason is worth knowing before either of you spends more time here.

torchtune is not maintained any more. The README carries it at the top:

> Torchtune is no longer actively maintained: torchtune development wound down in
> 2025 — see "The future of torchtune" (#2883).

and the tree matches. The last commit on `main` is `bd2a0fc7`, dated 2026-04-23,
which is the commit that added that README notice. Nothing has landed in the four
months since, against 457 open issues.

So the collision question I raised is moot. Neither PR is going to be merged
ahead of the other, because neither is going to be merged. I would rather say
that plainly than leave you both rebasing around each other.

For what it is worth, the work itself held up. The deprecation path, the negative
assertion on the canonical call, and the equivalence assertions between the
legacy and canonical spellings are all correct, and the reasoning is on the
record here if torchtune is ever picked up again or if the same pattern needs
fixing in a fork. That is not nothing, it is just not a merge.

Apologies for the noise on the overlap. I should have checked whether the repo
was alive before worrying about which of two PRs had seniority in its queue.

